PR #4, opened 2020-03-08
Fixes PUSH with register encoded in op

Sonicadvance1:
Unittest to ensure this keeps working is forthcoming.


skmp:
@Sonicadvance1 want to merge this now, or once the unit test is also around?

Sonicadvance1:
Should merge it now since it blocks real code running very early. The unit test for it will arrive in a day or two

PR #5, opened 2020-03-09
Fixes found from the rest of Two Byte table unit tests

Sonicadvance1:


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1165,14 +1165,9 @@ void OpDispatchBuilder::SHLDOp(OpcodeArgs) {
...
   OrderedNode *Res{};
-  if (Size == 16) {
-    LogMan::Msg::A("Unsupported Op");
-  }
-  else {
-    auto Tmp1 = _Lshl(Dest, Shift);
-    auto Tmp2 = _Lshr(Src, ShiftRight);
-    Res = _Or(Tmp1, Tmp2);
-  }
+  auto Tmp1 = _Lshl(Dest, Shift);
```
skmp: does it take care of Size here automagically?

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1200,17 +1195,12 @@ void OpDispatchBuilder::SHRDOp(OpcodeArgs) {
   else
     Shift = _And(Shift, _Constant(0x1F));
```
skmp: What about size==16? don't we need extra masking there?

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Yes, the sizing magic comes from the ops operating at the correct sizes (32bit/64bit) and the sizes being baked in to the instruction encodings that the frontend parse for us.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2193,12 +2184,13 @@ void OpDispatchBuilder::BSFOp(OpcodeArgs) {
       Src, ZeroConst,
       OneConst, ZeroConst);
 
-  StoreResult(GPRClass, Op, SelectOp, -1);
+  StoreResult_WithOpSize(GPRClass, Op, Op->Dest, SelectOp, DstSize, -1);
   SetRFLAG<FEXCore::X86State::RFLAG_ZF_LOC>(ZFSelectOp);
 }
 
 void OpDispatchBuilder::BSROp(OpcodeArgs) {
-  OrderedNode *Dest = LoadSource(GPRClass, Op, Op->Dest, Op->Flags, -1);
```
skmp: what about size==4 here?

External/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
```diff
@@ -182,14 +182,14 @@ void InitializeSecondaryTables() {
     {0xB8, 1, X86InstInfo{"",        TYPE_INVALID, FLAGS_NONE,                                                                                      0, nullptr}},
     {0xB9, 1, X86InstInfo{"",        TYPE_GROUP_10, FLAGS_NONE,                                                                                     0, nullptr}},
     {0xBA, 1, X86InstInfo{"",        TYPE_GROUP_8, FLAGS_NONE,                                                                                      0, nullptr}},
-    {0xBB, 1, X86InstInfo{"BTC",     TYPE_INST, FLAGS_DEBUG_MEM_ACCESS | FLAGS_MODRM | FLAGS_SF_MOD_DST,                                            0, nullptr}},
-    {0xBC, 1, X86InstInfo{"BSF",     TYPE_INST, FLAGS_MODRM,                                                                                        0, nullptr}},
-    {0xBD, 1, X86InstInfo{"BSR",     TYPE_INST, FLAGS_MODRM,                                                                                        0, nullptr}},
+    {0xBB, 1, X86InstInfo{"BTC",     TYPE_INST, FLAGS_DEBUG_MEM_ACCESS | FLAGS_MODRM | FLAGS_SF_MOD_DST | FLAGS_NO_OVERLAY,                         0, nullptr}},
```
skmp: what does FLAGS_NO_OVERLAY mean?

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Nope. x86-64 masks 0x3F for 64bit ops and 0x1F for anything lower.
This allows you to shift things out of the register in those instances

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: That's actually explicitly the operation that this change is working around.
When size == 4, we need to be careful to not do a partial writeback of the register when the incoming source has zero bits set, so we load the full 8byte register in that case to ensure zext magic doesn't occur in StoreResult

External/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
Sonicadvance1: FLAGS_NO_OVERLAY means that the instruction op should ignore instruction prefixes (specifically 0x66, 0xF2, and 0xF3) which could push the instruction down to another table.
This is because some of these instructions actually consume these prefixes rather than typically going to a new table.
BSF and TZCNT are good examples of this interaction.
BSF can use the 0x66 prefix (which drops it to 16bits), but the 0xF3 prefix can push it down to TZCNT in the other table (Which I have an upcoming change to make NO_OVERLAY a bit more granular)

skmp: :+1:

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
skmp: Cool beans

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
skmp: Oh okay. Such consistency much x86...

Issue #14, opened 2020-03-12
Clean up CVT IR Ops and OpDispatcher function names.


Sonicadvance1:
Both the IR ops and x86 functions for CVT can end up being confusing and prone to bugs.
Clean it up and make it more apparent what they are supposed to do.

PR #17, opened 2020-03-12
Cleans up and renames the CVT ops

Sonicadvance1:
Fixes issue #14.

A lot of these were named in very confusing ways.
Makes them very explicit about what the operation is doing

### DIFF
--- a/External/FEXCore/Source/Interface/IR/IR.json
+++ b/External/FEXCore/Source/Interface/IR/IR.json
@@ -544,7 +544,7 @@
       ]
     },
 
-    "FCVTZU": {
+    "Float_ToGPR_ZU": {
       "HasDest": true,
       "DestSize": "ElementSize",
       "SSAArgs": "1",
@@ -553,21 +553,12 @@
       ]
     },
 
-    "FCVTZS": {
+    "Float_ToGPR_ZS": {
       "HasDest": true,
       "DestSize": "ElementSize",
       "SSAArgs": "1",
       "Args": [
         "uint8_t", "ElementSize"
-      ]
-    },
-
-    "ExtractElement": {
-      "HasDest": true,
-      "DestSize": "GetOpSize(ssa0)",
-      "SSAArgs": "1",
-      "Args": [
-        "uint8_t", "Idx"
       ]
     },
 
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/IR/IR.json
```diff
@@ -1109,31 +1100,46 @@
...
+      "Desc": "Scalar op: Converts signed GPR to Scalar float",
       "HasDest": true,
       "SSAArgs": "1",
       "Args": [
         "uint8_t", "ElementSize"
       ]
     },
 
-    "VUCVTF": {
+    "FCVTFloatToFloat": {
```
phire: This is inconsistent in multiple ways.

 1. It really should be renamed to `FCVT_FToF` for consistency with `VCVT_UToF` and friends. Or maybe those should be renamed to `VCVT_UnsignedToFloat` for more clarity. 
 2. Scalar has one operation that switches between multiple sizes. Vector has two operations `VCVT_FNarrow`/`VCVT_FExtend`.
The naming of these be uniform.

External/FEXCore/Source/Interface/IR/IR.json
```diff
@@ -1109,31 +1100,46 @@
...
+      "Desc": "Scalar op: Converts float from one size to another",
+      "HasDest": true,
+      "SSAArgs": "1",
+      "Args": [
+        "uint8_t", "DstElementSize",
+        "uint8_t", "SrcElementSize"
+      ]
+    },
+
+    "VCVT_UToF": {
```
phire: VCVT is kind of opaque for a casual reader.

`Vector_UToF` would be infinity clearer and is only two chars longer. The `To` already implies conversion. 

Or perhaps Vector_UnsignedToFloat to go with the point above.

External/FEXCore/Source/Interface/IR/IR.json
```diff
@@ -1109,31 +1100,46 @@
       ]
     },
 
-    "UCVTF": {
+    "FCVTFromGPR_U": {
```
phire: To go with the VCVT comment below, FCVT is also opaque.

Perhaps `Float_FromGPR_U`

PR #23, opened 2020-03-13
Fix Neg IR_OP, Add Not IR_OP

phire:
In some places, Neg meant logical one's complement negation.
In others it meant arithmetic two's complement negation.

As both x87 and aarch64 have neg instructions that do two's complement negations, we will standardise on that.

Additionally we will add a Not_IR op for one's complement.


alyssarosenzweig:
Off-by-ones! :tada:

PR #41, opened 2020-03-17
LLVM 9 fixes

Sonicadvance1:
Moving from LLVM 8 to LLVM 9 caused unit tests to fail.
I didn't notice since my desktop was on LLVM 8 while the unit test runner was on LLVM 9.
We will want to keep an eye on LLVM 10 once it releases soon as well.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
+++ b/External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
@@ -1267,7 +1267,7 @@ void LLVMJITCore::HandleIR(FEXCore::IR::IRListView<true> const *IR, IR::NodeWrap
       // Our IR assumes defined behaviour for shifting all the bits out of the value
       // So we need to ZEXT to the next size up and then trunc
       auto OriginalType = Src1->getType();
-      auto BiggerType = Type::getIntNTy(*Con, 128);
+      auto BiggerType = Type::getIntNTy(*Con, OriginalType->getPrimitiveSizeInBits() * 2);
       Src1 = JITState.IRBuilder->CreateZExt(Src1, BiggerType);
       Src2 = JITState.IRBuilder->CreateZExtOrTrunc(Src2, BiggerType);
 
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
```diff
@@ -85,6 +94,9 @@ class LLVMJITCore final : public CPUBackend {
     llvm::Function *CPUIDFunction;
     llvm::Function *ExitVMFunction;
     llvm::Function *ValuePrinter;
+#ifdef _M_ARM_64
+    llvm::Function *AArch64ReadCycleCounterFunction;
```
skmp: Doesn't llvm have some intrinsic for this?

External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
```diff
@@ -184,6 +197,10 @@ class LLVMJITCore final : public CPUBackend {
     Result = JITState.IRBuilder->CreateShuffleVector(Result, ZeroVector, VectorMaskConstant);
 
 #elif _M_ARM_64
```
skmp: I'm skeptical of these defines - why are they needed?

External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
Sonicadvance1: Yea. It's actually explained in a comment down on line 274.

External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
Sonicadvance1: These are needed because we are actually using architecture specific intrinsics that behave slightly differently depending on x86 or aarch64.
Sadly LLVM doesn't have any idiom detection for automatically choosing the `packusdw` or `sqxtun{2}` instructions.

PR #57, opened 2020-03-21
Fixes a couple bugs in the futex syscall

Sonicadvance1:
futex command 9 is actually WAIT_BITSET, not WAKE_BITSET
Implements futexes with timeouts as well, necessary for anything using
std::condition_variable and wait_for/wait_until


Sonicadvance1:
There are some futexes tied to the TID that might end up being munged.
Haven't checked if there is a way to work around it yet

PR #62, opened 2020-03-22
Implement vector integer multiply IR ops

Sonicadvance1:


skmp:
Tests for this one?

Sonicadvance1:
Unit tests will arrive for this one once I push the instructions that use them

PR #70, opened 2020-03-28
Adds MMX decoding support to the frontend

Sonicadvance1:


External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -71,10 +71,32 @@ static uint32_t MapModRMToReg(uint8_t REX, uint8_t bits, bool HighBits, bool Has
...
+    ~0ULL,
+    ~0ULL
+  };
+
   const std::array<uint64_t, 16> *GPRs = &GPRIndexes;
   if (HasXMM) {
     GPRs = &XMMIndexes;
   }
+  else if (HasMM) {
+    GPRs = &MMIndexes;
```
skmp: maybe GPRs is best named "Registers" here? :P

PR #71, opened 2020-03-28
Removes debug logs from the frontend ELF loader

Sonicadvance1:


External/SonicUtils/Source/ELFLoader.cpp
```diff
@@ -550,7 +547,6 @@ void ELFContainer::FixupRelocations(void *ELFBase, uint64_t GuestELFBase, Symbol
               *Location = ELFSym->Address + Entry->r_addend;
             }
             else {
-              LogMan::Msg::D("Could not find symbol for x86_64_64 '%s'", EntrySymbolName);
```
skmp: isn't this one quite fatal?

Sonicadvance1: It can be if we are actually going to be using the custom dynamic linker path which this lives in.
Since that path is completely knackered it isn't worth keeping the debug logs around until someone looks at it.

PR #73, opened 2020-03-28
Implements push and pop flags instruction

Sonicadvance1:
This allows applications to push the eflags to and from the stack.
Noticed when running some GL games


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2345,6 +2345,42 @@ void OpDispatchBuilder::BSWAPOp(OpcodeArgs) {
   StoreResult(GPRClass, Op, Dest, -1);
 }
 
+void OpDispatchBuilder::PUSHFOp(OpcodeArgs) {
+  uint8_t Size = GetSrcSize(Op);
+  OrderedNode *Src = GetPackedRFLAG(false);
+  if (Size != 8) {
+    Src = _Bfe(Size * 8, 0, Src);
```
skmp: what does `_Bfe` stand for?

PR #76, opened 2020-04-05
Syscall cleanup

### LINKED ISSUE #12: Syscall emulation has incorrect global mutex
There is a scoped mutex at the top of the syscall handling function.
This is incorrect and causes the application to stall once threading kicks in.
Should be removed and only have a mutex on things that actually need it.
Can be done alongside syscall cleanup.

I think skmp wanted to have a go at this?

Sonicadvance1:
Splits out syscalls to more files and changes over to a jump table for handling the syscalls
Fixes a few bugs where we were missing checking errno returns.
Minor optimization where since we have syscall definitions we can optimize away the argument loading.
Until we have function calls in the IR we can't jump directly to the handler yet.
Fixes #12


External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -126,7 +126,7 @@ namespace DefaultFallbackCore {
 namespace FEXCore::Context {
   Context::Context()
     : FrontendDecoder {this}
-    , SyscallHandler {this} {
+    , SyscallHandler {FEXCore::CreateHandler(64, this)} {
```
skmp, line 129: what does `64` mean here?

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -422,8 +420,10 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
         uint64_t SPOffset = AlignUp((RA64.size() + 7 + 1) * 8, 16);
 
         sub(sp, sp, SPOffset);
-        for (uint32_t i = 0; i < 7; ++i)
+        for (uint32_t i = 0; i < 7; ++i) {
```
skmp, line 423: Why not `FEXCore::HLE::SyscallArguments::MAX_ARGS`?

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -1359,14 +1351,18 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
         auto Desired = GetSrc<RA_64>(Op->Header.Args[1].ID());
         auto MemSrc = GetSrc<RA_64>(Op->Header.Args[2].ID());
 
-        add(TMP1, MEM_BASE, MemSrc);
+        if (!CTX->Config.UnifiedMemory) {
```
skmp, line 1354: Do we support non-unified memory?

Design wise, wouldn't it be simpler to have membase as IR, and not in the backend?

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -1781,32 +1784,31 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
           // Result: RAX
 
           // These are pushed in reverse order because stacks
-          for (uint32_t i = 7; i > 0; --i)
+          for (uint32_t i = 7; i > 0; --i) {
```
skmp, line 1787: `FEXCore::HLE::SyscallArguments::MAX_ARGS`?

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
Sonicadvance1: Currently non-unified is required for the unit tests since they use a bunch of absolute addressing that causes issues with unified. Otherwise non-unified is effectively not implemented (Especially after this PR which strips it out of syscalls entirely)

Sonicadvance1:
This now has a requirement on #84 to pass the unit tests

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
Sonicadvance1: Did it

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
Sonicadvance1: Did it

External/FEXCore/Source/Interface/Core/Core.cpp
Sonicadvance1: You're right. Was difficult to tell what it was doing there.
Switched over to an enum to know if operating in 32bit or 64bit mode.
32bit mode not implemented yet.

PR #79, opened 2020-04-05
Fixes bug in CBW op

Sonicadvance1:
It was managing to Sext and clear the entire GPR when it was only meant to write to the lower 16bits


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -834,14 +834,16 @@ void OpDispatchBuilder::XCHGOp(OpcodeArgs) {
...
   OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags, -1);
-  uint8_t SrcSize = GetSrcSize(Op);
+  uint8_t DstSize = GetDstSize(Op);
+  uint8_t SrcSize = DstSize >> 1;
 
-  Src = _Sext(SrcSize * 4, Src);
-  if (SrcSize == 4) {
-    Src = _Zext(SrcSize * 8, Src);
-    SrcSize *= 2;
+  Src = _Sext(SrcSize * 8, Src);
```
skmp, line 840: This still feels dirty to me :/

PR #82, opened 2020-04-05
Adds RootFS search path in ELFLoader

### LINKED ISSUE #69: Fix ELFLoader not searching in rootfs for dynamic linker
Currently if you're on a host that doesn't have a `/lib64/ld-linux-x86-64.so.2` then you need to explicitly point to the dynamic linker location.
Let it search in the rootfs first.

Sonicadvance1:
Fixes #69, makes it less annoying to run things on non x86 hosts

### DIFF
--- a/External/SonicUtils/Source/ELFLoader.cpp
+++ b/External/SonicUtils/Source/ELFLoader.cpp
@@ -8,7 +8,7 @@
 
 namespace ELFLoader {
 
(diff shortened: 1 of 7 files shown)


skmp:
@Sonicadvance1 is this propagated correctly on fork/clone?

Sonicadvance1:
Yes, this'll propagate correctly through forks and clone.

PR #89, opened 2020-04-08
CMPXCHG8B and CMPXCHG16B support

Sonicadvance1:
This implements CMPXCHG8B and CMPXCHG16B instructions in the CPU emulation.
CPUID was already exposing this as "supported" but now it actually is.
There are a few extraneous moves (4-6 extra depending) that will end up getting resolved once we have RA constraints on instructions and potentially a move elimination pass.


CMakeLists.txt
```diff
@@ -48,6 +48,7 @@ set (CMAKE_LINKER_FLAGS_RELEASE "${CMAKE_LINKER_FLAGS_RELEASE} -fomit-frame-poin
 
 if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
   set(_M_X86_64 1)
+  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcx16")
```
skmp, line 51: what does `mcx16` do?

Sonicadvance1: This is the flag that enables cmpxchg16b on x86.
This force enables it so we technically lose some host x86-64 support for some Athlon 64 CPUs from 2009 but I'm fine with losing those on our non-primary architecture.

PR #95, opened 2020-04-09
BlockCache: Remove MemoryBase logic, check full address for match

skmp:
A slight modification of the original aliasing logic, that avoids aliases.


Sonicadvance1:
Looks good as a stop gap until the block cache is more extensive. Just needs to fix the merge conflict.

PR #110, opened 2020-04-12
Fixes clone syscall when VFORK flag is used

Sonicadvance1:
The VFORK flag lets clone behaviour similarly to a fork.
The parent will be suspended until the resulting thread is exited


Sonicadvance1:
Oop, I need to unlock the mutex in this before joining that thread.

Sonicadvance1:
Turns out the clone mutex isn't needed.
Looked at the relevant rest of the code and it already has mutexes in the correct locations.
Also added a check to throw a log message when something tries to clone without the CLONE_VM flag.
Since we don't support that yet.

PR #112, opened 2020-04-13
 Lets the IR validation pass validate RA

Sonicadvance1:
Adds some additional validation that an IR op is generating RA for ops
that set a destination.
Ensuring correct register class and that there is a physical register
assigned

Relies on #106 to be merged first!


skmp:
@Sonicadvance1 is the build really failing on this one?

PR #124, opened 2020-04-17
Fixes crashes when code cache is cleared

Sonicadvance1:
If the code cache is cleared then we would have the case where the IR data would remain.
Which could result in stale IR data when aliasing.
And if the IR existed then we wouldn't grab DebugData for that code block which would cause the interpreter and x86 JIT to crash.


Sonicadvance1:
Ping. Seven days old now.

PR #133, opened 2020-04-20
[WIP] Initial RA constraint codegen work

Sonicadvance1:
This isn't up for merging yet, just dumping it in to a PR while in progress.

`External/FEXCore/Source/Interface/IR/IR_Constraints_x86.json` has a bunch of comments in it about what the needs are.

This is currently just json codegen.
Nothing actually wired up yet.

- [ ] Rewrite RA to work bottom-up rather than top-down
- [ ] Rewrite JITs to pull register information from json
- [ ] Write all the constraints for each architecture in the json files
- [ ] Remove documentation from the x86 file
- [ ] RA improvements
  - [x] Support Dest = Src0
  - [x] Support Dest = Physical
  - [ ] Support Src<N> == Physical (x86 CMPXCHG16B limitation)
  - [x] Support that an op needs <N> temps
  - [x] Support that an op needs specific physical register temps (x86 only limitation?)
  - [ ] Support late-kill constraints (SrcN != dest) (x86 only limitation?)
  - [ ] Support Callee-Caller register designation so RA can return live registers that are caller saved at node location
  - [x] Support getting live registers at node location
- [ ] Ensure performance is same or better than current
- [ ] Ensure PHI nodes still work (Even though they are unused atm)
- [ ] Ensure spilling works


skmp:
@Sonicadvance1 is this still relevant?

phire:
I've done an initial rebase of this

 https://github.com/phire/FEX/tree/ra_constraints

Sonicadvance1:
Closing this. I won't delete the branch yet if anyone wants the old branch still.

PR #135, opened 2020-04-23
Added include of <llvm/IR/IntrinsicsX86.h> to LLVMCore.cpp

dmiller423:


External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
```diff
@@ -13,6 +13,7 @@
 #include <llvm/IR/LLVMContext.h>
 #include <llvm/IR/LegacyPassManager.h>
 #include <llvm/IR/Verifier.h>
+#include <llvm/IR/IntrinsicsX86.h>
```
Sonicadvance1, line 16: This include needs to be wrapped with an `#ifdef _M_X86_64` check since LLVM is used on both AArch64 and x86-64

Sonicadvance1:
Not sure why github's actions failed to run here. Looks like something in their frontend flaked out?

External/FEXCore/Source/Interface/Core/LLVMJIT/LLVMCore.cpp
dmiller423: I thought all of the Intrinsics were available for code generation regardless of the host platform.  I'll add the guard.

PR #157, opened 2020-04-25
Implements support for rcpps/rcpss

Sonicadvance1:


phire:
I'm confused. Why does rcpps/rcpss have insignificant precision to implement rcpps/rcpss on jit64?

Sonicadvance1:
Once again x86 has absolutely abysmal precision in its transcendental instructions.
The rcp instruction offers the precision of: "The maximum error is less than or equal to 1.5 * 2 –12 times the true reciprocal."
Which once again is off enough to make rcp of 1.0 turn in to `0x3f7ff000` or `0.999756`
Annoying enough to make the unit tests differ between the backends.
Since we aren't offering bit-accurate float implementations to a known x86 CPU yet, we can at least be more precise and be accurate to each backend.

phire:
Makes sense, I guess.

Can we get a issue ticket for this topic (bit accurate transcendental)

Sonicadvance1:
Tracking issue #173 opened for this.

PR #158, opened 2020-04-25
Default UnifiedMemory,  ensure TestHarnessRunner doesn't enable it.

dmiller423:


dmiller423:
The first commit is deceptive,  it only changes the typo in the commit message. Evidently the force push didn't go through until It was merged.

Source/Common/EnvironmentLoader.cpp
```diff
@@ -87,6 +87,9 @@ namespace FEX::EnvLoader {
       if ((Value = GetVar("FEX_UNIFIED_MEM")).size()) {
         if (isdigit(Value[0])) Config::Add("UnifiedMemory", Value);
       }
+      else {  // Most convenient place for a default as it's run before ArgLoader
+        Config::Add("UnifiedMemory", string_view("1"));
```
Sonicadvance1, line 91: This seems unnecessary.
The "Default" config sits in `External/FEXCore/Source/Interface/Core/Context.h` at line 52 there.
`bool UnifiedMemory {false};` Can just change that to true there and it'll default to enabled.

PR #162, opened 2020-04-26
Implements support for the three loop instructions

Sonicadvance1:
loop, loope, loopne


Sonicadvance1:
Oof. Yep, I missed that part of the documentation.
I'll fix that in a bit.

Sonicadvance1:
Concerns addressed.
Since you can't encode 16bit CX usage until running in 32bit mode, that'll be saved for when 32bit mode operation gets merged in.

PR #163, opened 2020-04-26
Fixes GS, implements segment reading and writing instructions, and implements XLAT

Sonicadvance1:
Noticed GS was wrong when implementing XLAT which led me down implementing the segment read/write registers for handling unit tests for this.


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1861,6 +1861,49 @@ void OpDispatchBuilder::PopcountOp(OpcodeArgs) {
...
+    Src = _Add(Src, _LoadContext(8, offsetof(FEXCore::Core::CPUState, gs), GPRClass));
+  }
+  Src = _Add(Src, Offset);
+
+  auto Res = _LoadMem(GPRClass, 1, Src, 1);
+
+  _StoreContext(GPRClass, 1, offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RAX]), Res);
+}
+
+template<bool FS>
```
phire, line 1881: For code readability purposes, could it be possible to use an Enum here?

That way we would end up with something like this below: 

```c++
    {OPD(FEXCore::X86Tables::TYPE_GROUP_15, PF_F3, 0), 1, &OpDispatchBuilder::ReadSegmentReg<Segment_FS)>},
    {OPD(FEXCore::X86Tables::TYPE_GROUP_15, PF_F3, 1), 1, &OpDispatchBuilder::ReadSegmentReg<Segment_GS>},
    {OPD(FEXCore::X86Tables::TYPE_GROUP_15, PF_F3, 2), 1, &OpDispatchBuilder::WriteSegmentReg<Segment_FS>},
    {OPD(FEXCore::X86Tables::TYPE_GROUP_15, PF_F3, 3), 1, &OpDispatchBuilder::WriteSegmentReg<Segment_GS>},
```

Sonicadvance1: Concern addressed

Sonicadvance1:
Requests addressed on this one

PR #172, opened 2020-04-29
Initial 32bit support

Sonicadvance1:
This is about 2500 lines to bring up initial support so I'm splitting it between multiple pull requests to make it more easy to digest.
This is the initial code that allows the ELF loader to load an ELF, and passes the running mode through to FEXCore.
Further implementations will be coming up after this PR is merged.

### DIFF
--- a/External/SonicUtils/include/SonicUtils/ELFLoader.h
+++ b/External/SonicUtils/include/SonicUtils/ELFLoader.h
@@ -24,7 +24,14 @@ public:
   ELFContainer(std::string const &Filename, std::string const &RootFS, bool CustomInterpreter);
   ~ELFContainer();
 
-  uint64_t GetEntryPoint() const { return Header.e_entry; }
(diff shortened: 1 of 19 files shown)


External/SonicUtils/include/SonicUtils/ELFLoader.h
```diff
@@ -59,11 +66,20 @@ class ELFContainer {
 
   void GetInitLocations(uint64_t GuestELFBase, std::vector<uint64_t> *Locations);
 
-  bool HasTLS() const { return TLSHeader != nullptr; }
+  bool HasTLS() const { return TLSHeader._64 != nullptr; }
```
skmp, line 69: Shouldn't this check 32/64 bit based on elf type?

Sonicadvance1: Since this is just a pointer check, doesn't matter here.

PR #174, opened 2020-04-30
[WIP] Enter inst

Sonicadvance1:
This one implements the ENTER instruction but I still need to do some verifying on it.
The AMD documentation claims that there is an additional memory write at the end of the instruction, but all other documentation I see doesn't have that at all.

- [ ] Need to do more hardware tests
- [ ] Need to write more unit tests


skmp:
@Sonicadvance1 I guess this is on pause for now?

Sonicadvance1:
Alright, this is now cleared for merging.
I did some hardware test verification as to its behaviour to ensure correctness, and added another unit test to ensure it tests that the rbp value is being pushed correctly.

External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -140,6 +140,7 @@ uint64_t Decoder::ReadData(uint8_t Size) {
   case 0: return 0;
   READ_DATA(1, uint8_t);
   READ_DATA(2, uint16_t);
+  case 3: memcpy(&Res, &InstStream[InstructionSize], Size);
```
skmp, line 143: Hum, what is this needed for?

skmp:
From what i understand this does not implement operand size / address size prefixes?

Also, is the logic for lock and such handled on the frontend?

PR #178, opened 2020-05-01
Ensure local pushes to a branch don't run through CI

Sonicadvance1:
If someone is pushing to a local branch on the repo, then don't run CI.
Only run once someone pushes a PR if local.
This makes it so we don't run CI for a branch twice over.


skmp:
@Sonicadvance1 while this looks good, having the CI run on every push really helped me debug the CI issues I had in posix-tests. I'll open a WIP PR from now on and hopefully that works

PR #181, opened 2020-05-08
Adds IR documentation generator and unit test runner

Sonicadvance1:
This is two fold since the unit test runner relies on changes from the documentation generator.
First thing that this does is adds a bunch of documentation labels to the IR.
It's not quite full coverage, a lot of the vector ops still don't have a description but I'll be filling those out over time as we go.
Secondly it takes all these new data and generates a markdown file for easing searching.
Example of generation: https://gist.github.com/Sonicadvance1/a97a58d6f9263910fb6bbebed3e1cff3

Then there is an IR loader application that implements an IR lexer, converts it to binary IR, and loads and runs the code.
This is needed for hand written IR unit tests.
This isn't fully fleshed out but I need to figure out what all it needs to support before I push those.


Source/Tests/IRLoader/Loader.cpp
```diff
@@ -0,0 +1,498 @@
...
+      return {DecodeFailure::DECODE_INVALID, {}};
+    }
+
+    return {DecodeFailure::DECODE_OKAY, FEXCore::IR::TypeDefinition::Create(Size / 8, Elements)};
+  }
+
+  template<>
+  std::pair<DecodeFailure, FEXCore::IR::CondClassType> Loader::DecodeValue(std::string &Arg) {
+    std::array<std::string, 14> CondNames = {
+      "EQ",
```
phire, line 116: These strings really should be defined somewhere near the CondClassType enum.

Sonicadvance1: I'll make a followup PR for this, since currently the CondClassType definition is defined in JSON. Needs a bit of JSON mangling to support it correctly (Or just move it outside of JSON)

PR #190, opened 2020-05-24
Fixes 128bit divisions on AArch64 

Sonicadvance1:
These were only doing 64bit divisions.
Function calls are a bit ugly but once the new RA lands we can clean them up

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -84,6 +84,30 @@ static uint64_t SyscallThunk(FEXCore::SyscallHandler *Handler, FEXCore::Core::In
   return FEXCore::HandleSyscall(Handler, Thread, Args);
 }
 
+static uint64_t LUDIV(uint64_t SrcHigh, uint64_t SrcLow, uint64_t Divisor) {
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -1987,7 +2011,41 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
...
+            str(RA, MemOperand(sp, i * 8));
+            i++;
+          }
+          str(lr,       MemOperand(sp, RA64.size() * 8 + 0 * 8));
+
+          mov(x0, GetSrc<RA_64>(Op->Header.Args[1].ID()));
+          mov(x1, GetSrc<RA_64>(Op->Header.Args[0].ID()));
+          mov(x2, GetSrc<RA_64>(Op->Header.Args[2].ID()));
+
+#if _M_X86_64
```
skmp, line 2028: Why is this ifdef needed?

Sonicadvance1: This ifdef is needed for running the vixl simulator for the AArch64 JIT on an x86 host.

PR #191, opened 2020-05-24
Fixes ASHR on AArch64

Sonicadvance1:
Smaller than 8byte wasn't sexting to the full register so it was causing issues


External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -1182,8 +1182,10 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
         auto Op = IROp->C<IR::IROp_Ashr>();
         if (OpSize == 8)
           asrv(GetDst<RA_64>(Node), GetSrc<RA_64>(Op->Header.Args[0].ID()), GetSrc<RA_64>(Op->Header.Args[1].ID()));
-        else
-          asrv(GetDst<RA_32>(Node), GetSrc<RA_32>(Op->Header.Args[0].ID()), GetSrc<RA_32>(Op->Header.Args[1].ID()));
+        else {
```
skmp, line 1185: How about in the future we do `else if (size == XX || YY || ZZ) { } else { assert(bad stuff); }` so the range of supported values is visible?

PR #193, opened 2020-05-24
Ensure only absolute paths fall down the rootfs path

Sonicadvance1:
Fixes ls


skmp:
Hmm, why does this only have to apply for absolute paths?

Sonicadvance1:
This is actually wrong. #201 is the correct fix for this.

PR #201, opened 2020-05-26
Fixes FD flags mismatch between x86 and ARM64

Sonicadvance1:
AArch64 redefines these four flags so they don't match between
architectures.
This is what was actually causing ls to fail, the x86 side was passing
in O_DIRECTORY which the AArch64 side was interpreting at O_DIRECT.
This made it return EINVAL


skmp:
Peculiar

PR #205, opened 2020-05-26
Makes threaded compilation not rely on a global mutex

Sonicadvance1:
As we run more threaded applications it is becoming apparent that this
needs to be fixed now.
This duplicates the FrontendDecoder and Passmanager per thread so they
no longer block each other while compiling.
Increases a bit of memory usage but completely worth it.


skmp:
Branch isn't on the main repo so can't rebase myself :<

PR #207, opened 2020-05-27
Adds support for interpreter fallback

Sonicadvance1:
The IRHeader itself contains a flag for if a block should fallback to
the interpreter.
This is necessary for the case that an IR is loaded and the Frontend
object no longer has the data necessary to know if it should do an
interpreter fallback.
This is super useful for bisecting JIT versus Intepreter failures, and
also cases where falling back to intepreter for compatibility is a lot
more simple than wiring things through the JIT (ala x87)

This isn't currently utilized, but will be once x87 work lands

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
@@ -1220,6 +1220,21 @@ namespace FEXCore::IR {
   bool ConstrainedRAPass::Run(IREmitter *IREmit) {
     bool Changed = false;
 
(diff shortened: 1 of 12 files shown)


External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -1220,6 +1220,21 @@ namespace FEXCore::IR {
   bool ConstrainedRAPass::Run(IREmitter *IREmit) {
     bool Changed = false;
 
+    auto IR = IREmit->ViewIR();
```
skmp, line 1223: I think we should split these checks to an "IRValidationPass" eventually

PR #208, opened 2020-05-27
Implements fence ops

Sonicadvance1:
These map directly to x86 fence instructions, which have equivalent ARM
versions as well.


skmp:
It's nice when things map cleanly, isn't it?

PR #211, opened 2020-05-27
Implements softfloat library for x87

Sonicadvance1:
Comes from http://www.jhauser.us/arithmetic/SoftFloat.html which is under a permissive license that we can use.
Gets most of the x87 implementation out of the way.

### DIFF
--- a/External/FEXCore/Source/Common/SoftFloat.h
+++ b/External/FEXCore/Source/Common/SoftFloat.h
@@ -1,0 +1,253 @@
+#pragma once
+#include "LogManager.h"
+
+#include <cmath>
+#include <cstring>
+#include <stdint.h>
+#include <string>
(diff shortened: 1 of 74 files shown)


External/FEXCore/Source/Common/SoftFloat.h
```diff
@@ -0,0 +1,253 @@
...
+#include <string>
+#include <sstream>
+
+extern "C" {
+#include "SoftFloat-3e/platform.h"
+#include "SoftFloat-3e/softfloat.h"
+}
+
+struct X80SoftFloat {
+#ifdef _M_X86_64
```
skmp, line 16: Maybe we should use some more descriptive and less magic defines here? It's kinda hard to keep track of all those esoteric names imo

PR #217, opened 2020-05-30
Forces LLVM backend to use the native CPU architecture

Sonicadvance1:
Better than relying on a hardcoded CPU target
Should fix a crash on x86 host cpus that don't support BMI1


skmp:
<3

PR #222, opened 2020-05-31
syscalls/x64: Add pselect6

skmp:
Fixes python REPL

### DIFF
--- a/External/FEXCore/Source/Interface/HLE/Syscalls/FD.cpp
+++ b/External/FEXCore/Source/Interface/HLE/Syscalls/FD.cpp
@@ -293,6 +293,11 @@ namespace FEXCore::HLE {
     SYSCALL_ERRNO();
   }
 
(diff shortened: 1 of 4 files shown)


External/FEXCore/Source/Interface/HLE/Syscalls/FD.cpp
```diff
@@ -293,6 +293,12 @@ namespace FEXCore::HLE {
     SYSCALL_ERRNO();
   }
 
+  uint64_t Pselect6(FEXCore::Core::InternalThreadState *Thread, int nfds, fd_set *readfds, fd_set *writefds, fd_set *exceptfds, const struct timespec *timeout, const sigset_t *sigmask) {
+    // sig is ignored here? -> need to validate w/ glibc source
```
Sonicadvance1, line 297: What sig here?

skmp: oh, the comment is outdated, will remove it.

PR #226, opened 2020-06-02
Dynamic Syscall Regsistration, Automatic Strace generation

skmp:
Second attempt at this

- Dynamic registration of syscalls instead of look up table
- Automatic parameter number deduction, as well as strace generation
- Added `NotImplemented.cpp` for syscalls we don't plan to implement
- Added `Stubs.cpp` with asserting stubs for all syscalls, to make future implementation much easier

Advantages
- No header files required for syscalls
- Lookup table doesn't need to be filled in
- Strace prints don't need to be manually written
- For simple syscall forwarding, an one liner can be used

Disadvantages
- Requires syscall implementations to be lambas / somewhat funky syntax
- Requires registration fuctions - constructor based registration is buggy on clang
- No control over how strace prints (always `::D` in)

Follow up work
- Double check that the interfaces actually match the kernel ABI (existing and stubs)
- Make a plan on how to implement, test, and double check syscalls
- Validate existing implementations (I've fixed a couple like syslog, possibly a few more bugs lurking)
- Make strace prints also configurable, when and if we need it


skmp:
@Sonicadvance1 I've disabled `strftime` (unknown why it fails, I can't repro even w/ locale locally) and more of the signal tests (tgkill/tkill make them fail). The rest of the tests pass.

PR #233, opened 2020-06-12
Implements atomic ops for ARMv8.0 devices

Sonicadvance1:
These can only be implemented as acquire/release exclusive loops on
ARMv8.0 so they end up looking pretty disgusting.
This is more for compatibility for testing than anything. I'm not
expecting ANY ARMv8.0 device to get decent performance.

I'll be buying an ARMv8.0 device that can live in CI for ensuring these stay working.
That will be wired up once it arrives.


Sonicadvance1:
Oops, forgot about the loadstore exclusive pair ops. cmpxchg can still happen. I'll get on that.

PR #240, opened 2020-06-16
Implements two missing timerfd syscalls

Sonicadvance1:


Sonicadvance1:
Oh right. Necessary for FTL for when Pulseaudio is correctly setup in the rootfs

PR #244, opened 2020-06-25
Only run conformance-interfaces-strftime-2-1.test

skmp:


skmp:
The `strftime` is buggy, and doesn't correctly set LANG, and also allocates a 20 byte buffer but tells strftime that it is a 256 byte buffer.

The reason it passed before, is that `tgkill` was nop'd in the syscalls, leading to a normal exit, even though the stack smashing was detected.

PR #247, opened 2020-06-26
Fixes LLVM multiblock

Sonicadvance1:
The break ir op is actually a block ender and it happened that in
regular code paths they were already ending the blocks correctly.

This enforces block ending in the Unimplemented op so multiblock still
generates the invalid code path correctly and it compiles.
We hit these a few times in multiblock for times in the code where it is
checking for CPU features and choosing a code path depending on CPUID
results.

Potentially these could be made in to no-ops since we'll never hit these
codepaths in practice, but it is nicer to break in case we actually do
hit one of these paths.


skmp:
I guess this also happens when multiblock leads us to non-code, right?

Sonicadvance1:
Multiblock won't ever lead to non-code unless it is malformed (Guest code patching a branch target and setting conditional branch comparison to go down that path I guess?)
OpDispatcher will early exit on an instruction it doesn't understand instead

PR #256, opened 2020-07-02
syscalls: Implement a few easy ones

skmp:
truncate,creat,chown,fchown,lchown,setpgid,getpgid,setfsuid,setfsgid,getsid,sync,syncfs
*get/setxattr


skmp:
Yeah, I let them to do a validation pass later on for the params

PR #262, opened 2020-07-04
[RFC] thunks: initial POC

skmp:
This is a PoC to get feedback

Basic Idea
- New op, OP_THUNK that takes `const char *name` (imm, host address space), `ThunkFunction *fp` (imm, host address space), `void *args` (passed from guest code, guest address space)
- Repurpose `int 0x7F` to have "thunk" meaning. After it, an asciz follows with the thunk name. It also performs a return. This is used as a "magic opcode" to trampoline from guest to host code, similar to Syscalls. `RDI` holds the arg structure, which includes args and rv (if any)
- Thunking helpers generated by python, both guest and host. The guest helpers collect args and rv in a stack-stored struct, the host helpers uncollect them from the struct to params/rv. This is done so that we don't have to worry about ABI/arguments on top of everything else
- ThunkHandler interface (also part of context) that contains similar logic to SyscallHandler.


skmp:
Update: X11/GLX/GL thunking example

Almost 100% generated.
Manual points
- lib init calls in `ThunkHandler_impl()`
- `glXGetProcAddress` hackplementation

This doesn't generate `.so`s in the rootfs, instead linking directly with the thunks (see `unittests/Thunks/build.sh` for details)
```
$> cd unittests/Thunks
$> ./build.sh
$> cd ../../build
$> ninja Bin/ELFLoader
$> Bin/ELFLoader -- ./thunk-test.elf
$> Bin/ELFLoader -- ./gl3thunked.elf
```

skmp:
Update: Rootfs Library for thunking

- Compiles `glx-thunks.cpp` to .so, links to `libX11`, `libGL`, `libGLX` and `libGLdispatcher`
- More thunks to be able to run glxgears
- Updated build scripts

skmp:
Update: More of GL/GLX/X11, more complete `glXGetProcAddress`
- Enough thunks to load FTL
- Gets stuck on `XIfEvent`, likely needs callback support

skmp:
(Closing as #263 is a much improved implementation)

PR #263, opened 2020-07-07
[RFC] Thunking system

skmp:
# What is this?
This implements a semi-automated framework for guest -> host thunking, to allow us to quickly write thunks.

# How does it work?

## Runtime
Building on the `int 0x7F;<aciiz symbol>` pattern from #262 this goes a step further by moving all of the thunking implementations to `guest` and `host` libs.

This depends on FEX using a rootfs where the `guest` libs (`libNAME-guest.so`) are in the rootfs with the expected linker names, and it also expects the `host` libs to be in a special folder (hardcoded for now, soon to change).

When a `guest` lib is loaded, a special thunk, `fex:loadlib`, is called to notify FEX. FEX then loads the matching `host` lib via dlopen, calls `fexthunks_exports_libNAME` which loads the native lib/imports via `dlopen`/`dlsym` and returns to get a list of the host thunks. FEX then and inserts the host thunks to the thunk map.

The guest lib can then use the new thunks (which are named `libName:export`). It's up to the `guest` and `host` libs to work together to implement thunking, especially in tricky cases like variadic functions.

## Compile
Libraries are defined in `ThunkGen/libName-thunks.py`. `ThunkGen/generate-thunks.sh` needs to be run before compilation to generate several .inl files used in `ThunkLibs`. The python helpers are used to generate the bulk of the thunks that follow a the standard pattern.

Supporting code for the guest libraries is in `ThunkLibs/lib*/_Guest.cpp`
Supporting code for the host libraries is in `ThunkLibs/lib*/_Host.cpp`
There is some boilerplate there, as well "tricky/custom" implementations (eg, glXGetProcAddress, XGetICValues, XCreateIC)

`ThunkLibs/build.sh` will build the libraries. It's up to you to move them to the expected locations (RootFS / magic folder for host libs)

# File Structure
- `ThunkGen/*` contains scripts that define librarties
  - `ThunkGen/ThunkHelpers.py` contains the generator and DSL definition
- `ThunkLibs/lib*/lib*_Guest.cpp` contains the guest.so implementation
- `ThunkLibs/lib*/lib*_Host.cpp` contains the guest.so implementation

# TODO
- Integrate with CMAKE
- Figure out where to put the host libs, and how to integrate with RootFSes in an easy way.
- Investigate callbacks. x64 will silently work for now since the passed code is native code.

# That's nice, but does it work?
`glxgears` and `FTL` work. Not much else.


skmp:
This is not actually part of FEXCore, the thunks are completely independent and build in a separate build step (apart from the small loader in [Thunks.cpp](https://github.com/FEX-Emu/FEX/pull/263/files#diff-9c3741c863e3b6f1ffe6db667a5607a5)). Syscalls are more integrated.

32bit software I don't have a plan yet. I think the general idea is to provide an x86 rootfs and run the emulator in a window in a 64-bit process. We can do some mmap magic to force all allocations to be in the lower 32-bits, to ease thunking and syscalls, but structures will have to be converted a lot.

I'm waiting on @Sonicadvance1's work on re-entrance before implementing callbacks. As testing happens on x86-64 atm, we can pass the "guest" code as callbacks and it works. In the future, we'll need reverse thunks/trampolines. 

The general idea is that in a per-lib case, we'll likely have to re-implement quite a bit of the logic. There's some other difficult cases, eg X11 exports data symbols for locks, and it'll be impossible to interwork guest and host locks there.

The generator can be configured to only generate some parts (thunk itself, handlers for argument pack/unpack, dlsym loader, etc) already, and I plan to keep extending this to integrate it as cleanly as possible while reducing code duplication. It'll be a complex balancing act. Hope it doesn't end up looking like lex/yacc :|

skmp:
Split to #281 and #292

PR #270, opened 2020-07-12
AArch64 TSO loadstore ops

Sonicadvance1:
Relies on #259 to be merged first!

This implements new TSO IR ops for x86 loadstores.
This /improves/ the situation for implementing x86 TSO memory semantics, but doesn't cover it 100%.
A followup PR will fix the issues with unaligned atomicity but that needs to wait a moment


Sonicadvance1:
New PR opened for this.

PR #286, opened 2020-08-03
Creates the interpreter core just before the IR jit

Sonicadvance1:
Arm64 IR JIT needs the interpreter to be around first to get a pointer
to it


phire:
Can we get a comment saying the initialisation order is important?

Sonicadvance1:
Comment added about why we want to initialize the interpreter before the IRJIT.

PR #288, opened 2020-08-03
Declares new sigret x86 instruction and IR op

Sonicadvance1:
We are rebranding an x86-64 illegal instruction near VIA's old ALTINST
instruction as an instruction to handle returning from guest signals
cleanly.

If this ever collides with a new x86-64 (unlikely) then we will need to
change the instruction.
This ends up looking similar to what we will do for thunk callbacks


External/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
```diff
@@ -251,7 +251,9 @@ void InitializeSecondaryTables() {
 
     // FEX reserved instructions
     // Unused x86 encoding instruction.
-    
+    // Used by FEX to know when to do a signal reteurn
```
skmp, line 254: s/reteurn/return?

PR #302, opened 2020-08-20
Throw an error in CMake if someone attempts to compile with GCC

Sonicadvance1:
FEX doesn't support GCC.


skmp (approved):

pinskia:
What does FEX do that causes it not be able to compile with FEX? Does it use some extension or just don't want folks to even try compiling with GCC?

Sonicadvance1:
> What does FEX do that causes it not be able to compile with FEX? Does it use some extension or just don't want folks to even try compiling with GCC?

Our IR is generated in a way that makes GCC really upset at least. Not sure what else has broken since I last tested years ago.

PR #308, opened 2020-08-20
Extends gdbstub functionality

Sonicadvance1:
There are a significant number of changes here that I don't really
expect much review on.
Adds a bunch of new features and allows us to support GDB, IDA, and
Binja through their remote GDB debugging facilities.
Ghidra may also work but I haven't tried.

This work isn't quite yet complete to do correct thread stepping yet. We
need to add support for stopping threads using signals since threads may
be stuck inside of a syscall.

I have a hack locally that tries pausing all threads and times out after
a few seconds which works well enough for local testing.

### DIFF
--- a/External/FEXCore/Source/Common/NetStream.cpp
+++ b/External/FEXCore/Source/Common/NetStream.cpp
@@ -6,6 +6,7 @@
 #include <sys/socket.h>
 
 #include <stdio.h>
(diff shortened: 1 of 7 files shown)


phire (approved):
You have a few traces left of the NO_REPLY in-band signaling.

Should delete them rather than have people in the future confused.

PR #332, opened 2020-08-29
IR: Iterative RCLSE

skmp:
## Iterative RCLSE
Due to the way RCLSE is implemented, it improves the IR but misses some cases. This will run it up to 5 times. For some reason, running it twice is not good enough.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
@@ -1,3 +1,4 @@
+#include "Interface/IR/Passes.h"
 #include "Interface/IR/PassManager.h"
 #include "Interface/Core/OpcodeDispatcher.h"
 #include <FEXCore/Core/CoreState.h>
@@ -246,9 +247,12 @@ class RCLSE final : public FEXCore::IR::Pass {
 public:
   RCLSE() {
     ClassifyContextStruct(&ClassifiedStruct);
+    DCE.reset(FEXCore::IR::CreatePassDeadCodeElimination());
   }
   bool Run(FEXCore::IR::IREmitter *IREmit) override;
 private:
+  std::unique_ptr<FEXCore::IR::Pass> DCE;
+
   ContextInfo ClassifiedStruct;
   std::unordered_map<FEXCore::IR::OrderedNodeWrapper::NodeOffsetType, BlockInfo> OffsetToBlockMap;
 
@@ -608,7 +612,12 @@ bool RCLSE::Run(FEXCore::IR::IREmitter *IREmit) {
   // XXX: We don't do cross-block optimizations yet
   //CalculateControlFlowInfo(IREmit);
   bool Changed = false;
-  Changed |= RedundantStoreLoadElimination(IREmit);
+  
+  // Run up to 5 times
+  for( int i = 0; i < 5 && RedundantStoreLoadElimination(IREmit); i++) {
+    Changed = true;
+    DCE->Run(IREmit);
+  }
 
   return Changed;
 }


phire:
On my C630, this is good for about a 7% performance gain. 

On my 3900x, this is good for about a 6% performance gain.

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
```diff
@@ -608,7 +619,14 @@ bool RCLSE::Run(FEXCore::IR::IREmitter *IREmit) {
   // XXX: We don't do cross-block optimizations yet
   //CalculateControlFlowInfo(IREmit);
   bool Changed = false;
-  Changed |= RedundantStoreLoadElimination(IREmit);
+  
+  VerboseLog("Start!");
+  while(RedundantStoreLoadElimination(IREmit)) {
+    Changed = true;
+    DCE->Run(IREmit);
+    VerboseLog("%p: Changed, Re-running\n", this);
```
phire, line 627: If I'm right, this should never run more than twice. 

First pass will remove all the LoadContexts it can. Second pass will pickup and remove all the redundant StoreContexts

phire:
And with the extra BFE optimisation, I get another 1.5% on arm and 2.5% on x86

(These are all FTL loading time benchmarks)

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
skmp: In some cases, it runs 4-5 times, removing one or two more ops per iteration.

External/FEXCore/Source/Interface/IR/Passes/DeadCodeElimination.cpp
```diff
@@ -38,12 +38,22 @@ bool DeadCodeElimination::Run(IREmitter *IREmit) {
...
     auto CodeLast = CurrentIR.at(BlockIROp->Last);
+    auto CodeEnd = CurrentIR.at(BlockIROp->Last);
 
     while (1) {
       auto CodeOp = CodeLast();
       OrderedNode *CodeNode = CodeOp->GetNode(ListBegin);
       auto IROp = CodeNode->Op(DataBegin);
 
+      // A BFE that extract all bits is a nop
+      if (IROp->Op == OP_BFE) {
```
phire, line 49: Would make more sense to move this to the Constant Propagation pass.

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
```diff
@@ -482,6 +487,7 @@ bool RCLSE::RedundantStoreLoadElimination(FEXCore::IR::IREmitter *IREmit) {
           // Happens when we store in to a location then store again
           IREmit->Remove(LastStoreNode);
           Changed = true;
+          VerboseLog("WaW");
```
phire, line 490: Might want to remove some of these verbose debug logs

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
```diff
@@ -608,7 +619,14 @@ bool RCLSE::Run(FEXCore::IR::IREmitter *IREmit) {
   // XXX: We don't do cross-block optimizations yet
   //CalculateControlFlowInfo(IREmit);
   bool Changed = false;
-  Changed |= RedundantStoreLoadElimination(IREmit);
+  
+  VerboseLog("Start!");
+  while(RedundantStoreLoadElimination(IREmit)) {
```
phire, line 624: Can we bound this loop to some kind of max iterations. 

During some dev on top of this branch I ran into a case where this looped indefinitely.

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
phire: Weird. Maybe I don't fully understand the flow of this pass.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -38,11 +38,26 @@ bool ConstProp::Run(IREmitter *IREmit) {
...
       auto CodeOp = CodeBegin();
       OrderedNode *CodeNode = CodeOp->GetNode(ListBegin);
       auto IROp = CodeNode->Op(DataBegin);
       switch (IROp->Op) {
+
+      case OP_CONSTANT: {
+        auto Op = IROp->C<IR::IROp_Constant>();
+        if (consts.count(Op->Constant)) {
+          IREmit->ReplaceAllUsesWithInclusive(CodeNode, consts[Op->Constant], CodeBegin, CodeLast);
+          IREmit->Remove(CodeNode);
```
phire, line 53: Can you comment out this `Remove` for now?

There are some weird edge cases with ReplaceAllUsesWithInclusive not removing all possible uses. Best to just let DCE remove the node for now.

phire (changes requested):

External/FEXCore/Source/Interface/IR/Passes/DeadCodeElimination.cpp
skmp: Will make a new pass for these cases - also `AND %v, %v`, `OR %v, %v`, `XOR %v, %v`, `Add/Sub %v, 0`, etc

phire: I think it makes sense to put these in ConstProp; We already optimise away OP_LSHL shift by zero there.

What is best for performance, a single pass which does both, or two passes?

skmp: Done in #338

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
skmp: Done in #339

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
skmp: Removed them for now, will work on a nicer way to log these things in the future

External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
skmp: It'll run a maximum of 5 times now

skmp:
@phire updated and cleaned up. Looks good now?

phire (approved):

PR #339, opened 2020-09-02
ConstProp: Allocate constants once per block

skmp:
This seems to hit some bug on RA, so blocked til @phire has that fix merged


phire:
You should be able to rebase this now.

skmp:
@phire rebased this, still needs fixed RA before merge

skmp:
@Sonicadvance1 This can be reviewed now. I added "panic spilling" that works around the RA issues. We need to revisit our RA spilling logic asap.

Sonicadvance1 (approved):

PR #343, opened 2020-09-05
Has /proc/cpuinfo generated from CPUID

Sonicadvance1:
Instead of having this file mostly just copy and pasted, it'll now be
generated.
We end up losing a lot of supported flags with this change, which is for
the best since we wouldn't have actually supported those things.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/External/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -1,4 +1,7 @@
 #include "Interface/Core/CPUID.h"
+#include "git_version.h"
+
+#include <cstring>
 
 namespace FEXCore {
 //#define CPUID_AMD
@@ -8,15 +11,15 @@ CPUIDEmu::FunctionResults CPUIDEmu::Function_0h() {
 
   // EBX, EDX, ECX become the manufacturer id string
 #ifdef CPUID_AMD
-  Res.Res[0] = 0x0D; // Let's say we are a Zen+
-  Res.Res[1] = CPUID_VENDOR_AMD1;
-  Res.Res[2] = CPUID_VENDOR_AMD2;
-  Res.Res[3] = CPUID_VENDOR_AMD3;
+  Res.eax = 0x0D; // Let's say we are a Zen+
+  Res.ebx = CPUID_VENDOR_AMD1;
+  Res.edx = CPUID_VENDOR_AMD2;
+  Res.ecx = CPUID_VENDOR_AMD3;
 #else
-  Res.Res[0] = 0x16; // Let's say we are a Skylake
-  Res.Res[1] = CPUID_VENDOR_INTEL1;
-  Res.Res[2] = CPUID_VENDOR_INTEL2;
-  Res.Res[3] = CPUID_VENDOR_INTEL3;
+  Res.eax = 0x16; // Let's say we are a Skylake
+  Res.ebx = CPUID_VENDOR_INTEL1;
+  Res.edx = CPUID_VENDOR_INTEL2;
+  Res.ecx = CPUID_VENDOR_INTEL3;
 #endif
   return Res;
 }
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -25,89 +28,321 @@ CPUIDEmu::FunctionResults CPUIDEmu::Function_0h() {
...
+    (1 <<  3) | // MWait
+    (1 <<  4) | // DS-CPL
+    (0 <<  5) | // VMX
+    (0 <<  6) | // SMX
+    (0 <<  7) | // Intel SpeedStep
+    (0 <<  8) | // Thermal Monitor 2
+    (0 <<  9) | // SSSE3
+    (0 << 10) | // L1 context ID
+    (0 << 11) | // Silicon debug
+    (1 << 12) | // FMA
```
phire, line 54: 1. Should be labled FMA3
2. I don't think we support FMA3.

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -25,89 +28,321 @@ CPUIDEmu::FunctionResults CPUIDEmu::Function_0h() {
...
-      (1 << 14)   // Remove AVX512VPOPCNTDQ
-      );
+  Res.eax = 0x0;
+  Res.ebx =
+    (1 <<  0) | // FS/GS support
+    (0 <<  1) | // TSC adjust MSR
+    (0 <<  2) | // SGX
+    (0 <<  3) | // BMI1
+    (0 <<  4) | // Intel Hardware Lock Elison
+    (1 <<  5) | // AVX2 support
```
phire, line 129: AVX2 should be disabled.

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -25,89 +28,321 @@ CPUIDEmu::FunctionResults CPUIDEmu::Function_0h() {
...
+  Res.eax = 0 | // Stepping
+    (0 << 4) | // Model
+    (0 << 8) | // Family ID
+    (0 << 12) | // Processor type
+    (0 << 16) | // Extended model ID
+    (0 << 20); // Extended family ID
+
+  Res.ecx =
+    (1 <<  0) | // LAHF/SAHF
+    (1 <<  1) | // 0 = Single core product, 1 = multi core product
```
phire, line 248: More of a question than a code review: 

Should we be setting this to 0 when we set the number of threads to zero?

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -25,89 +28,321 @@ CPUIDEmu::FunctionResults CPUIDEmu::Function_0h() {
...
+    (1 << 22) | // MMXExt
+    (1 << 23) | // MMX
+    (1 << 24) | // FXSAVE/FXRSTOR
+    (1 << 25) | // FXSAVE/FXRSTOR Optimizations
+    (1 << 26) | // 1 gigabit pages
+    (0 << 27) | // RDTSCP
+    (0 << 28) | // Reserved
+    (1 << 29) | // Long Mode
+    (1 << 30) | // 3DNow! Extensions
+    (1 << 31);  // 3DNow!
```
phire, line 312: Do we really want to claim to support **3DNow!**

phire (approved):

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Fixed

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Fixed

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Disabled

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Without hardware testing this on a real single core product I don't really want to fuss with it.
Apparently the last one released from Intel is a Sandy Bridge Celeron G470 CPU back in 2013.

Issue #318, opened 2020-08-25
Take advantage of ARMv8.4-RCPC


Sonicadvance1:
We aren't currently using this instruction set. We should move over to supporting it to improve memory access performance.
Our Cortex-A76/A55 CPUs support the instructions.

Sonicadvance1:
This has been implemented.

PR #344, opened 2020-09-05
Adds support for RCPC load instructions

Sonicadvance1:
If the host CPU supports RCPC instructions then it'll use those
instruction of atomics for LoadMemTSO ops.

This pushed my worst case FTL test from 16FPS to 17FPS .

There is another extension to RCPC to add short immediate offset support
which the c630 doesn't support any way. For future optimization on
hardware we can use that.


Sonicadvance1:
Fixes #318

skmp (approved):

PR #350, opened 2020-09-06
loopOp: Constant takes size in bits, not bytes

skmp:


Sonicadvance1 (approved):

Sonicadvance1:
What was the situation that this got hit in that unit tests didn't pick it up?
Was it just broken IR that happened to work with i0 constants because they technically end up being treated as 64bit constants in most cases?

skmp:
I don't know - I run into it from the qemu tests - maybe we don't test for size == 4 in our unit tests? Also needs the IR validation pass to show up, and shows up as a divide by zero

Issue #351, opened 2020-09-06
Zext semantics are broken


skmp:
Eliminating supposedly harmless full-extract BFEs breaks CS:GO's shader translation logic

PR #352, opened 2020-09-06
IR/ConstProp: Disable BFE identity as it breaks cs:go

skmp:
See #351


phire (approved):

phire:
Yeah, my bad.

I never finished the changes which fixed all IR OPs to not zext, nor did I finish writing tests.

I should have remembered that before approving this optimisation.

PR #353, opened 2020-09-06
Thunks: Improve SDL2 so that CSGO creates the window and GL context

skmp:


ThunkLibs/libSDL2/libSDL2_Guest.cpp
```diff
@@ -29,4 +33,11 @@ int SDL_SetError(const char*, ...) { return printf("SDL2: SDL_SetError\n"); }
...
 void SDL_LogMessageV(int, SDL_LogPriority, const char*, __va_list_tag*) { printf("SDL2: SDL_LogMessageV\n");}
 int SDL_vsnprintf(char*, size_t, const char*, __va_list_tag*) { return printf("SDL2: SDL_vsnprintf\n");}
-int SDL_vsscanf(const char*, const char*, __va_list_tag*) { return printf("SDL2: SDL_vsscanf\n");}
\ No newline at end of file
+int SDL_vsscanf(const char*, const char*, __va_list_tag*) { return printf("SDL2: SDL_vsscanf\n");}
+
+extern "C" {
+    void* SDL_GL_GetProcAddress(const char* name) {
+		// TODO: Fix this HACK
+		return (void*)glXGetProcAddress((const GLubyte*)name);
```
Sonicadvance1, line 41: We really need to stop relying on glx.
I don't plan on running X11 for my entire life.

Sonicadvance1 (approved):

ThunkLibs/libSDL2/libSDL2_Guest.cpp
skmp: We need an EGL loader, and then to figure out how to load the right lib.

PR #354, opened 2020-09-06
Opcode fixes

skmp:
These were discovered by running some of valgrind's unit tests under FEX


Sonicadvance1:
Before this gets merged, it needs some asm unit tests attached.

skmp:
Split into single PR per fix

PR #365, opened 2020-09-11
Refactor the IR Interface to prevent "foot-shooting"

phire:
This moves most of the internal complexities of our IR inside just a few files:  `IR.h`, `IR.cpp`, `IREmitter.cpp`, `IREmitter.cpp` and `IntrusiveIRList.h`

There are now new iterators for Blocks and Code.

PR #366, opened 2020-09-11
Replace improvements

phire:
* Provide a fast path for "Replace with Constant"
 * Early exit optimisation for ReplaceAllUses. Stop when all uses have been found.

Depends on #365 

Ignore everything except the last two commits.


phire:
Rebased and ready for review

Sonicadvance1 (approved):

PR #374, opened 2020-09-11
Removes a couple of unnecessary variables to improve reentrancy chances

Sonicadvance1:
We still aren't 100% reentrant safe in the case that the cores are
interrupted while running, but this is a step in the right direction


phire:
Hmm... this conflicts with #365

phire (approved):

PR #390, opened 2020-09-15
SMC: Add --smc-full-checks, implement frontend, irint, irjit, unit test

skmp:
This is a brute-force solution that can catch any kind of SMC, including modifications to the next instruction

Adds two new IR ops
- ValidateCode that compares current vs original code and returns 0 on same, non zero on different
- RemoveCodeEntry that removes the code entries (via Context::RemoveCodeEntry)

Adds `--smc-full-checks` in FEXLoader and UnitTestHarness

Adds logic to validate code in `Context::CompileBlock` (Core.cpp) if enabled

Adds a unit test to check that SMC detection works, `--smc-full-checks` is passed only for that test

Notes
- ValidateCode can be implemented as a string of Constant/Cmp/Select IR ops, though codegen would be much worse
- When enabled it disables most optimizations as it emits a host-block per guest-op

### DIFF
--- a/External/FEXCore/Source/Interface/Config/Config.cpp
+++ b/External/FEXCore/Source/Interface/Config/Config.cpp
@@ -36,6 +36,9 @@ namespace FEXCore::Config {
     case FEXCore::Config::CONFIG_TSO_ENABLED:
       CTX->Config.TSOEnabled = Config != 0;
     break;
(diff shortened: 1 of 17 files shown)


Sonicadvance1 (approved):
It's disgusting as expected.
But for something that is meant to bring down the hammer on SMC sure.

PR #391, opened 2020-09-16
Remove unneeded BFEs around flag generation code.

phire:
Depends on #365. Only last few commits are relevant.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -1746,24 +1746,15 @@ void OpDispatchBuilder::ASHROp(OpcodeArgs) {
     Src = LoadSource(GPRClass, Op, Op->Src[1], Op->Flags, -1);
   }
 
-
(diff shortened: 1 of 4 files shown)


phire:
Looks like this has broken Nvidia drivers. 

Things like glxgears and FTL are just showing a black window.

Will track down later.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1772,11 +1788,14 @@ void OpDispatchBuilder::ASHRImmediateOp(OpcodeArgs) {
     Shift &= 0x1F;
 
   OrderedNode *Src = _Constant(Size, Shift);
-  auto ALUOp = _Ashr(Dest, Src);
+  OrderedNode *Result = _Ashr(Dest, Src);
```
phire, line 1791: Also need sign extend

phire:
Rebased and tracked down the nvidia driver issue (INC overflow had incorrect zero flag)

Fixed and added unit tests

Sonicadvance1 (approved):
Changes SHLD behaviour slightly for zero shifts, but it is closer to correct so still a win.

PR #392, opened 2020-09-16
Spam out a wackload more syscalls

Sonicadvance1:
Almost all of these end up being simple passthrough.
This covers most syscalls, the remaining ones are signal related, exec
related, or rseq required.

All of these syscalls exist prior to Linux 5.0. We are using these with
the expectation still that the host Linux implementation is 5.0 or
higher.
Once we implement newer syscalls we will need to start checking for host
Linux version. We don't really have a care about devices shipping older
4.x kernels atm.


skmp (approved):
here's to hoping that all the flags and structs match!

PR #395, opened 2020-09-17
Fixes CASP IR op on interpreter when unified memory is enabled

Sonicadvance1:
Noticed by Stef on 09/09/2020 as a curious code block that was failing
Forgot about it until now since we couldn't see exactly where it was
wrong

Code block was mixing x87 (Which forces a fallback to Interpreter) and
cmpxchg8b, which caused that code block to fail.

Since this op was broken with Unified Memory + Interpreter it wasn't
picked up by unit tests since those don't run in unified memory

Code block from before
```
   0x00007fece80daa10:    push   rbp
   0x00007fece80daa11:    mov    rbp,rsp
   0x00007fece80daa14:    push   rbx
   0x00007fece80daa15:    fild   QWORD PTR [rdi]
   0x00007fece80daa17:    fistp  QWORD PTR [rbp-0x10]
   0x00007fece80daa1a:    mov    rax,QWORD PTR [rbp-0x10]
   0x00007fece80daa1e:    xchg   ax,ax
   0x00007fece80daa20:    lea    rcx,[rax+rsi*1]
   0x00007fece80daa24:    mov    ebx,ecx
   0x00007fece80daa26:    shr    rcx,0x20
   0x00007fece80daa2a:    lock cmpxchg8b QWORD PTR [rdi]
   0x00007fece80daa2e:    cmp    rdx,rax
   0x00007fece80daa31:    je     0x7fece80daa40
   0x00007fece80daa33:    mov    rax,rdx
   0x00007fece80daa36:    jmp    0x7fece80daa20
   0x00007fece80daa38:    nop    DWORD PTR [rax+rax*1+0x0]
   0x00007fece80daa40:    pop    rbx
   0x00007fece80daa41:    pop    rbp
   0x00007fece80daa42:    ret 
```

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
@@ -472,7 +472,7 @@ void InterpreterCore::ExecuteCode(FEXCore::Core::InternalThreadState *Thread) {
               case 4: {
                 std::atomic<uint64_t> *Data = {};
                 if (Thread->CTX->Config.UnifiedMemory) {
-                  Data = GetSrc<std::atomic<uint64_t> *>(SSAData, Op->Header.Args[2]);
+                  Data = *GetSrc<std::atomic<uint64_t> **>(SSAData, Op->Header.Args[2]);
                 }
                 else {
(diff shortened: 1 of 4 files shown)


skmp (approved):

skmp:
nice catch!

PR #396, opened 2020-09-21
syscalls: execve should fail on missing files

skmp:


External/FEXCore/Source/Interface/HLE/Syscalls/Thread.cpp
```diff
@@ -254,6 +255,13 @@ namespace FEXCore::HLE {
     REGISTER_SYSCALL_IMPL(execve, [](FEXCore::Core::InternalThreadState *Thread, const char *pathname, char *const argv[], char *const envp[]) -> uint64_t {
       std::vector<const char*> Args;
 
+      std::error_code ec;
+      bool exists = std::filesystem::exists(pathname, ec);
+      if (ec || !exists) {
+        return -ENOENT;
+      }
+      printf("EXEC: %s\n", pathname);
```
Sonicadvance1, line 263: Debug statement needs to be removed

Sonicadvance1 (approved):

PR #397, opened 2020-09-21
x86-64: Implement IRET for wine

skmp:


Sonicadvance1 (approved):
Needs a unit test which I'll get around to doing.

PR #399, opened 2020-09-22
Allows unit tests registers to have partial register views

Sonicadvance1:
This allows us to only check for the lower 64bits of the MM registers
when doing MMX ops for example.
Top 16bits in that case is undefined and needs to be ignored

### DIFF
--- a/Source/Tests/HarnessHelpers.h
+++ b/Source/Tests/HarnessHelpers.h
@@ -157,52 +157,52 @@ namespace FEX::HarnessHelper {
       }
 
       if (BaseConfig.OptionRegDataCount > 0) {
-        constexpr std::array<std::pair<uint64_t, unsigned>, 45> OffsetArray = {{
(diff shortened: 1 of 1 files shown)


Source/Tests/HarnessHelpers.h
```diff
@@ -228,7 +228,7 @@ namespace FEX::HarnessHelper {
             Matches &= A == B;
           };
 
-          for (unsigned j = 0; j < Offset.second; ++j) {
+          for (unsigned j = 0; j < RegData->RegDataCount; ++j) {
```
skmp, line 231: This depends on the test config file correctly setting this field, right?

skmp (approved):

Source/Tests/HarnessHelpers.h
Sonicadvance1: This is already getting set correctly in the python config parser.
RegCount always counts the number of 64bit registers the test is testing for.
We don't yet support smaller than 64bit here.

PR #406, opened 2020-09-23
Fixes crash when SSA value has size of zero

Sonicadvance1:
In the case of an invalid SSA destination size, the IR validation pass
would catch it and then when printing the IR it would crash with a
divide by zero.
Fixes that.


skmp:
This crash is how I found out about #350 :P

skmp (approved):

PR #408, opened 2020-09-23
Adds new IR ops for supporting host rounding mode conversion

Sonicadvance1:
Interpreter might not be quite correct here (Could end up doing truncate or host flags depending on how it compiles)
But until we have the unit tests in place it's hard to ensure atm.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
@@ -19,6 +19,9 @@
 #include <cmath>
 #include <limits>
 #include <vector>
+#ifdef _M_X86_64
+#include <xmmintrin.h>
+#endif
 
 namespace FEXCore::CPU {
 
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
```diff
@@ -3698,6 +3733,30 @@ void InterpreterCore::ExecuteCode(FEXCore::Core::InternalThreadState *Thread) {
             }
             break;
           }
+          case IR::OP_FLOAT_TOGPR_S: {
```
skmp, line 3736: C/++ defines the cast as a truncation, so these don't actually use the host fpu flags. Also, x86 saturates negative /and/ positive overflows to 0x8000_0000, while arm does 0x8000_0000 and 0x7FFF_FFFF, so we might want to have manual overflow checks here?

Sonicadvance1: I think until we get some unit tests for this. At least until then were aren't actually doing worse than before.
I'll create an issue about it.

skmp (approved):

PR #422, opened 2020-09-23
Adds and enables the x86 host runner when building on x86 host!

Sonicadvance1:
Disables the thunking unit test on host because obviously it won't work
there

### DIFF
--- a/unittests/32Bit_ASM/CMakeLists.txt
+++ b/unittests/32Bit_ASM/CMakeLists.txt
@@ -43,21 +43,23 @@ foreach(ASM_SRC ${ASM_SOURCES})
 
   # Format is "<Test Arguments>" "<Test Name>"
   set(TEST_ARGS
-    "-c irint -n 1"      "int_1"
-    "-c irint -n 500"    "int_500"
-    "-c irint -n 500 -m" "int_500_m"
-    "-c irjit -n 1"      "jit_1"
-    "-c irjit -n 500"    "jit_500"
-    "-c irjit -n 500 -m" "jit_500_m"
+    "-c irint -n 1"      "int_1"     "int"
+    "-c irint -n 500"    "int_500"   "int"
+    "-c irint -n 500 -m" "int_500_m" "int"
+    "-c irjit -n 1"      "jit_1"     "jit"
+    "-c irjit -n 500"    "jit_500"   "jit"
+    "-c irjit -n 500 -m" "jit_500_m" "jit"
     )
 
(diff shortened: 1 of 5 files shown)


unittests/ASM/CMakeLists.txt
```diff
@@ -43,21 +43,28 @@ foreach(ASM_SRC ${ASM_SOURCES})
...
+    "-c irint -n 1"      "int_1"     "int"
+    "-c irint -n 500"    "int_500"   "int"
+    "-c irint -n 500 -m" "int_500_m" "int"
+    "-c irjit -n 1"      "jit_1"     "jit"
+    "-c irjit -n 500"    "jit_500"   "jit"
+    "-c irjit -n 500 -m" "jit_500_m" "jit"
     )
+  if (_M_X86_64)
+    list(APPEND TEST_ARGS
+      "-g -c host"       "host"      "host"
```
skmp, line 55: why not pass -g globally? seems like a good idea anyway

unittests/32Bit_ASM/CMakeLists.txt
```diff
@@ -43,28 +43,31 @@ foreach(ASM_SRC ${ASM_SOURCES})
 
   # Format is "<Test Arguments>" "<Test Name>"
```
skmp, line 44: Add "\<Test Runner\>"?

unittests/ASM/Disabled_Tests
```diff
@@ -23,3 +23,6 @@ Test_Primary/Primary_E9.asm
 Test_X87/D9_F9.asm
 
 Test_X87/D9_F2.asm
+
+# 3DNow! no longer exists on AMD Zen CPUs
```
skmp, line 27: Now with the `Disabled_Tests_host`, wouldn't it make more sense to move this there?

skmp (changes requested):

unittests/ASM/CMakeLists.txt
Sonicadvance1: Sounds good

unittests/32Bit_ASM/CMakeLists.txt
Sonicadvance1: Added

unittests/ASM/Disabled_Tests
Sonicadvance1: Moved

skmp (approved):

PR #445, opened 2020-09-29
Signed multiply fixes

Sonicadvance1:
Flags calculation was wrong in these and for 32bit sext result we were storing the zext result instead of the sext result.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -2581,25 +2581,25 @@ void OpDispatchBuilder::IMULOp(OpcodeArgs) {
   if (Size == 1) {
     // Result is stored in AX
     _StoreContext(GPRClass, 2, offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RAX]), Result);
-    ResultHigh = _Bfe(8, 8, Result);
-    ResultHigh = _Sext(Size * 8, ResultHigh);
+    ResultHigh = _Sbfe(8, 8, Result);
   }
   else if (Size == 2) {
     // 16bits stored in AX
     // 16bits stored in DX
     _StoreContext(GPRClass, Size, offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RAX]), Result);
-    ResultHigh = _Bfe(16, 16, Result);
-    ResultHigh = _Sext(Size * 8, ResultHigh);
+    ResultHigh = _Sbfe(16, 16, Result);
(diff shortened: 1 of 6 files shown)


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2556,23 +2556,21 @@ void OpDispatchBuilder::IMULOp(OpcodeArgs) {
...
+    ResultHigh = _Sbfe(16, 16, Result);
     _StoreContext(GPRClass, Size, offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RDX]), ResultHigh);
   }
   else if (Size == 4) {
     // 32bits stored in EAX
     // 32bits stored in EDX
     // Make sure they get Zext correctly
-    OrderedNode *ResultLow = _Bfe(4, 32, 0, Result);
-    ResultHigh = _Bfe(4, 32, 32, Result);
+    OrderedNode *ResultLow = _Sbfe(32, 0, Result);
```
skmp, line 2572: Shouldn't these remain `_Bfe`s here?

skmp: (I went over the manual and it doesn't say anything about SEXT on imul, so don't the typical ZEXT rules apply?)

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2556,23 +2556,21 @@ void OpDispatchBuilder::IMULOp(OpcodeArgs) {
...
+    ResultHigh = _Sbfe(32, 32, Result);
     _StoreContext(GPRClass, Size, offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RAX]), ResultLow);
```
skmp, line 2574: For correct ZEXT semantics, shouldn't these store `GRPSize` instead of `Size` ?

skmp: (And also with `Bfe` instead of `Sbfe`

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Oops, you're right.
I was abusing the sext for the flags calculation in the 8bit and 16bit flags calculation and accidentally extended in to the 32bit side

Sonicadvance1: Should probably extend the unit test again just to ensure proper zext mechanics here.

skmp (approved):

PR #448, opened 2020-09-30
Splits x86 JIT's IR handling over to a jump table

Sonicadvance1:
This is changes the behaviour to be exactly like the ARM64 side.
No behaviour change beyond that.


skmp (approved):

skmp:
:D

PR #473, opened 2020-10-22
Complex address generation for Load/StoreMem

skmp:
### Overview

Adds a third argument to Load/StoreMem*, `Offset`, which is by default `Invalid()`. Also adds `OffsetType` (SXTX, SXTW, UXTW, supported options vary by backend), and `OffsetScale` (1, 2, 8, 16, supported options also vary by backend).

Modifies OpDispatcher to not generate TSO opcodes if TSO is disabled.

Modifies IRLoader to support `%Invalid`, extends IRWriter and IRLoader to support `MemOffsetType`.

The ever-more-complex constprop pass does
- Fills `Offset` whenever possible, either with reg or constant
- Detects `Offset = RealOffset * scale` or `Offset = Zext/Sext(RealOffset)` and sets `OffsetType` and `OffsetScale` accordingly
- If `Offset` is a constant, and if it can be inlined, it converts it to an `InlineConstant`.

### Interpreter backend
- Supports Offset Reg
- Doesn't support Offset InlineConstant
- Supports any OffsetScale
- Supports SXTX, SXTW, UXTW

### Arm64 backend
- Supports Offset Reg
- Supports Offset InlineConstant (based on arm64 imm rules)
- Supports OffsetScale of 1 or access size
- Supports SXTX, SXTW, UXTW

### x86 backend
- Supports Offset Reg
- Supports OffsetInlineConstant (simm32)
- Supports OffsetScale of 1, 2, 4 or 8
- Supports only SXTX

### Notes
This modifies LoadMem/LoadMemTSO/StoreMem/StoreMemTSO, as the *TSO variants are assumed to have the same struct in some parts of the code. The optimization is only performed for non-TSO, as arm can't do address generation in TSO opcodes.

Arm64 and x86 backends will assert if any unsupported options are used.

### Further work
- Add detection for scaled SXTW and UXTW
- Add support for write backs (Pre/Post increments). This needs some more extensive IR work.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
@@ -350,6 +350,7 @@ private:
   DEF_OP(FillRegister);
   DEF_OP(LoadFlag);
   DEF_OP(StoreFlag);
+  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, uint8_t OffsetType, uint8_t OffsetScale);
   DEF_OP(LoadMem);
   DEF_OP(StoreMem);
   DEF_OP(LoadMemTSO);
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
@@ -350,7 +350,7 @@ private:
   DEF_OP(FillRegister);
   DEF_OP(LoadFlag);
   DEF_OP(StoreFlag);
-  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, uint8_t OffsetType, uint8_t OffsetScale);
+  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, IR::MemOffsetType OffsetType, uint8_t OffsetScale);
   DEF_OP(LoadMem);
   DEF_OP(StoreMem);
   DEF_OP(LoadMemTSO);
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
@@ -148,6 +148,8 @@ private:
   aarch64::VRegister GetSrc(uint32_t Node);
   aarch64::VRegister GetDst(uint32_t Node);
 
+  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, IR::MemOffsetType OffsetType, uint8_t OffsetScale);
+
   bool IsInlineConstant(const IR::OrderedNodeWrapper& Node, uint64_t* Value = nullptr);
 
   struct LiveRange {
@@ -350,7 +352,6 @@ private:
   DEF_OP(FillRegister);
   DEF_OP(LoadFlag);
   DEF_OP(StoreFlag);
-  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, IR::MemOffsetType OffsetType, uint8_t OffsetScale);
   DEF_OP(LoadMem);
   DEF_OP(StoreMem);
   DEF_OP(LoadMemTSO);
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -694,8 +694,8 @@ void JITCore::RegisterMemoryHandlers() {
     REGISTER_OP(STOREMEMTSO,         StoreMemTSO);
   }
   else {
-    REGISTER_OP(LOADMEMTSO,          LoadMem);
-    REGISTER_OP(STOREMEMTSO,         StoreMem);
+    REGISTER_OP(LOADMEMTSO,          Unhandled);
+    REGISTER_OP(STOREMEMTSO,         Unhandled);
   }
   REGISTER_OP(VLOADMEMELEMENT,     VLoadMemElement);
   REGISTER_OP(VSTOREMEMELEMENT,    VStoreMemElement);
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -404,14 +404,18 @@ DEF_OP(StoreFlag) {
 DEF_OP(LoadMem) {
   auto Op = IROp->C<IR::IROp_LoadMem>();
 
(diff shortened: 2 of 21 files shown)


External/FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -438,6 +438,10 @@ friend class FEXCore::IR::PassManager;
     return RealNode->Op(Data.Begin());
   }
 
+  OrderedNode *UnwarpNode(OrderedNodeWrapper ssa) {
```
Sonicadvance1, line 441: UnwrapNode

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -513,6 +513,26 @@ friend class FEXCore::IR::PassManager;
 
   bool Multiblock{};
   uint64_t Entry;
+
+  IRPair<IROp_StoreMemTSO> _StoreMemTSO(FEXCore::IR::RegisterClassType Class, uint8_t Size, OrderedNode *ssa0, OrderedNode *ssa1, uint8_t Align = 1) {
```
Sonicadvance1, line 517: Make these be helpers that don't follow the same naming convention as the IREmitter functions.
It's confusing that an IR "emission" of _StoreMemTSO doesn't always do a _StoreMemTSO Op.
Also should just make the helper return OrderedNode*, doesn't feel good converting the IRPair to TSO, if one happens to change definition then this would immediately be a broken cast.

External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
```diff
@@ -290,6 +290,7 @@ class JITCore final : public CPUBackend, public Xbyak::CodeGenerator {
   DEF_OP(FillRegister);
   DEF_OP(LoadFlag);
   DEF_OP(StoreFlag);
+  Xbyak::RegExp GenerateModRM(Xbyak::Reg Base, IR::OrderedNodeWrapper Offset, IR::MemOffsetType OffsetType, uint8_t OffsetScale);
```
Sonicadvance1, line 293: Move this outside of the definitions of the IR handling ops for cleanliness.

External/FEXCore/Source/Interface/Core/JIT/x86_64/MemoryOps.cpp
```diff
@@ -606,8 +636,14 @@ void JITCore::RegisterMemoryHandlers() {
   REGISTER_OP(STOREFLAG,           StoreFlag);
   REGISTER_OP(LOADMEM,             LoadMem);
   REGISTER_OP(STOREMEM,            StoreMem);
-  REGISTER_OP(LOADMEMTSO,          LoadMem);
-  REGISTER_OP(STOREMEMTSO,         StoreMem);
+  if (CTX->Config.TSOEnabled) {
```
Sonicadvance1, line 639: Why?
The IR op handling should be enabled regardless.
I could see this being useful for your debugging but it shouldn't hit upstream.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -665,8 +705,8 @@ void JITCore::RegisterMemoryHandlers() {
     REGISTER_OP(STOREMEMTSO,         StoreMemTSO);
   }
   else {
-    REGISTER_OP(LOADMEMTSO,          LoadMem);
-    REGISTER_OP(STOREMEMTSO,         StoreMem);
+    REGISTER_OP(LOADMEMTSO,          Unhandled);
```
Sonicadvance1, line 708: Same comment as from the x86 side.
Maybe even further that if you are expecting to have TSO ops entirely eliminated prior to them hitting the backend, then we should have the TSO ops never be conditional selecting the handling and instead always act as TSO; Regardless of the config option.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
```diff
@@ -350,6 +350,7 @@ class JITCore final : public CPUBackend, public vixl::aarch64::Assembler  {
   DEF_OP(FillRegister);
   DEF_OP(LoadFlag);
   DEF_OP(StoreFlag);
+  MemOperand GenerateMemOperand(uint8_t AccessSize, aarch64::Register Base, IR::OrderedNodeWrapper Offset, IR::MemOffsetType OffsetType, uint8_t OffsetScale);
```
Sonicadvance1, line 353: Move this above the IR handler definitions for cleanliness.

Sonicadvance1 (changes requested):

External/FEXCore/include/FEXCore/IR/IREmitter.h
skmp: Fixed in #490 (will rebase once merged)

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
skmp: `_StoreMemAutoTSO`? Maybe also get rid of the `_`?

Yeah, agreed on the return, though we cast the structs to each other all over the code. Will change it

External/FEXCore/Source/Interface/Core/JIT/x86_64/MemoryOps.cpp
skmp: Mostly to make it symmetrical with arm, and for validation. Never got hit though :P

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
skmp: Yeah, the only reason I changed this was because it was conditional before. I do like the validation aspect, but that can be moved to the IR validation passes (if we want, when we have free time). I'll change it for now to not be conditional across all backends ~

External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/x86_64/MemoryOps.cpp
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
skmp: Done

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
skmp: Renamed to `_StoreMemAutoTSO`

External/FEXCore/include/FEXCore/IR/IREmitter.h
skmp: Rebased

Sonicadvance1 (approved):

PR #475, opened 2020-10-25
Fixes 32bit x87 usage

Sonicadvance1:
All of the ops that were relying on a 64bit memory offset were being
downgraded to 32bit.

Abuses the XMM_FLAGS option a bit to indicate that it shouldn't
downgrade the 64bit size flag. Which isn't an issue since all these
operations can't encode a register since modrm_r is used for table
decoding


skmp (approved):

skmp (approved):
Can you create an alias for FLAGS_XMM_FLAGS so that it makes sense textually? Reading the tables is hard enough because x86 w/o extra gotchas

PR #479, opened 2020-10-30
IR: Add DeadGPRStoreElimination

skmp:
This is based on the Flag Elimination pass, and is not optimal - but does eliminate quite a few dead stores.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
@@ -1,6 +1,9 @@
 #include "Interface/IR/PassManager.h"
 #include "Interface/Core/OpcodeDispatcher.h"
 
+// Higher values might result in more stores getting eliminated but will make the optimization take more time
+constexpr int PropagationRounds = 5;
+
 namespace FEXCore::IR {
 
 class DeadGPRStoreElimination final : public FEXCore::IR::Pass {
@@ -92,7 +95,7 @@ bool DeadGPRStoreElimination::Run(IREmitter *IREmit) {
   // Pass 2
   // Compute GPRs that are stored, but always ovewritten in the next blocks
   // Propagate the information a few times to eliminate more
-  for (int i = 0; i < 5; i++)
+  for (int i = 0; i < PropagationRounds; i++)
   {
     for (auto [BlockNode, BlockIROp] : CurrentIR.GetBlocks()) {
       for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
@@ -135,7 +138,6 @@ bool DeadGPRStoreElimination::Run(IREmitter *IREmit) {
           // If this OP_STORECONTEXT is never read, remove it
           if (GPRMap[BlockNode].kill & GPRBit(Op->Offset)) {
             IREmit->Remove(CodeNode);
-            //printf("Removed dead store %d, %d\n", Op->Offset, IROp->Size);  
             Changed = true;
           }
         }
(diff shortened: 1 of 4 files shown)


skmp:
*updated to use TrueBlock/FalseBlock instead of Arg[XX]*

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
```diff
@@ -0,0 +1,154 @@
...
+          auto Op = IROp->CW<IR::IROp_StoreContext>();
+          if (IsFullGPR(Op->Offset, IROp->Size))
+            GPRMap[BlockNode].writes |= GPRBit(Op->Offset);
+          else
+            GPRMap[BlockNode].reads |= GPRBit(Op->Offset);
+        }
+        else if (IROp->Op == OP_STORECONTEXTINDEXED ||
+               IROp->Op == OP_LOADCONTEXTINDEXED ||
+               IROp->Op == OP_LOADCONTEXTPAIR ||
+               IROp->Op == OP_STORECONTEXTPAIR) {
```
Sonicadvance1, line 79: We also can't track through syscall here.
Syscall can modify full context state...
We should also probably move rax setting from syscall to be handled by syscall handler so we are more generic...

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+      }
+    }
+  }
+
+  // Pass 2
+  // Compute GPRs that are stored, but always ovewritten in the next blocks
+  // Propagate the information a few times to eliminate more
+  for (int i = 0; i < 5; i++)
```
Sonicadvance1, line 95: Make this constant a constexpr value towards the top for easy tinkering

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
```diff
@@ -0,0 +1,154 @@
...
+  {
+    for (auto [BlockNode, BlockIROp] : CurrentIR.GetBlocks()) {
+      for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
+
+        if (IROp->Op == OP_STORECONTEXT) {
+          auto Op = IROp->CW<IR::IROp_StoreContext>();
+          // If this OP_STORECONTEXT is never read, remove it
+          if (GPRMap[BlockNode].kill & GPRBit(Op->Offset)) {
+            IREmit->Remove(CodeNode);
+            //printf("Removed dead store %d, %d\n", Op->Offset, IROp->Size);
```
Sonicadvance1, line 138: remove debug printf?

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
skmp: RCLSE, DeadFlags and RA assume that SYSCALLS don't modify the context. Do syscalls really do it?

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
skmp: Done

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
skmp: Done

Sonicadvance1 (approved):

External/FEXCore/Source/Interface/IR/Passes/DeadGPRStoreElimination.cpp
Sonicadvance1: Will resolve later

PR #482, opened 2020-11-05
RCLSE improvements

skmp:
- Handles some FPR cases that were for some reason disabled. A few are left disabled because they break stuff and need further investigation.
- Also handles some other corner cases
- VMOV zext/trunc seem to no longer be causing issues, enabled


skmp:
Hmm, can't replicate the test failure locally. Peculiar...

Sonicadvance1 (approved):

PR #483, opened 2020-11-05
MUL: remove zext/sexts from 32 bit form

skmp:


External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -153,9 +153,10 @@ DEF_OP(Mul) {
       sxth(Dst, Dst);
     break;
     case 4:
-      sxtw(TMP1, GetReg<RA_64>(Op->Header.Args[0].ID()));
-      sxtw(TMP2, GetReg<RA_64>(Op->Header.Args[1].ID()));
-      mul(Dst.W(), TMP1.W(), TMP2.W());
+      //sxtw(TMP1, GetReg<RA_64>(Op->Header.Args[0].ID()));
```
Sonicadvance1, line 156: Commented lines need cleaned up.

Sonicadvance1 (approved):

External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
skmp: Done

PR #486, opened 2020-11-05
Unsafe opts: --unsafe-local-flags, --unsafe-no-pf, --unsafe-no-tso

skmp:
- InvalidateFlags pseudo-op is a hint for the Dead Flags Elimination pass that flags are not needed past that point
- Adds a new option, --unsafe-local-flags, that invalidates the flags on CALLs or RETs. This might break code that doesn't follow the ABI rules.
- Adds a new option, --unsafe-no-pf, that doesn't calculate PF for integer operations. This might break code that actually uses PF.
- Renames --disable-tso to --unsafe-no-tso to group the options nicely


Sonicadvance1 (approved):

Sonicadvance1 (approved):

Sonicadvance1:
Has merge conflicts now.

skmp:
Rebased

PR #487, opened 2020-11-05
OpDisp: More efficient handling of setcc and cmov

skmp:
Instead of reducing eflags to a register and then testing, this tests the needed flags directly


skmp:
I'll close this in favor of the solution in #489

PR #489, opened 2020-11-05
OpDisp: flags bypass

skmp:
This

- Adds special cases for when the flags are a result of a cmp or a test

depends on #488, #491, #499


skmp:
This exposes at least one bug in the reg allocator where

```
test   0xConst,%reg
rep stos // or other MB opcode
je // uses %reg in fastpath
```

Doesn't generate valid code.

%reg, is mapped to the wrong register, on the second usage, and is also overwritten in the test opcode.

skmp:
(Rebased)

Sonicadvance1 (approved):

PR #490, opened 2020-11-05
CondJump: Add support for two operands to match Select

skmp:
Depends on #488

- Renames Cond to Cmp1, adds Cmp2, adds Cond as a CC (to match select), backwards compatible emit that does Cmp2 = 0, CC = NEQ
- Adds support to the backends for Cmp1/Cmp2 & Cond, with special cases for == 0, !=0 and Cmp2 being a const
- Optimization to fold Selects to CondJumps whenever possible

*rebased and re-re-re-reviewed*

### DIFF
--- a/External/FEXCore/include/FEXCore/IR/IREmitter.h
+++ b/External/FEXCore/include/FEXCore/IR/IREmitter.h
@@ -443,6 +443,10 @@ friend class FEXCore::IR::PassManager;
     return RealNode->Op(Data.Begin());
   }
 
+  OrderedNode *UnwarpNode(OrderedNodeWrapper ssa) {
+    return ssa.GetNode(ListData.Begin());
+  }
+
   // Overwrite a node with a constant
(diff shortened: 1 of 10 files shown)


External/FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -438,6 +443,10 @@ friend class FEXCore::IR::PassManager;
     return RealNode->Op(Data.Begin());
   }
 
+  OrderedNode *UnwarpNode(OrderedNodeWrapper ssa) {
```
Sonicadvance1, line 446: UnwrapNode

Sonicadvance1 (changes requested):

External/FEXCore/include/FEXCore/IR/IREmitter.h
skmp: Woops. Fixed.

Sonicadvance1 (approved):

PR #496, opened 2020-11-12
BlockCache: Add a cache for faster lookups

skmp:
## Overview
This adds an direct cache (1M entries) for the block lookups. Lookups in the Cache are /much/ faster than full lookups, leading to overall speedup. The cache structure is also friendly to atomic modifications.

The cache hit rate is > 98% based on my testing, though it depends on the complexity of the guest app.

This also switches to dispatching via jmp instead of call/ret, and does inline dispatch at the end of each block.

## Generic Block Lookup
```
auto& Entry = Cache[Addr & MASK];
if (Entry.Guest == Addr) {
  return Entry.Host;
} else {
  HostAddr = FullLookup();
  Entry.Guest = Addr;
  Entry.Host = HostAddr;
  return HostAddr;
}
```

## X86 Lookup
```
  mov(r13, ThreadState->BlockCache->GetL1Pointer());
  mov(rax, rdx);

  and_(rax, 1 * 1024 * 1024 - 1);
  shl(rax, 1);
  cmp(qword[r13 + rax*8 + 0], rdx);
  jne(FullLookup);
  call(qword[r13 + rax*8 + 8]);
```

## ARM64 Lookup
```
  and_(x3, RipReg, 1 * 1024 * 1024 - 1);
  add(x0, x0, Operand(x3, Shift::LSL, 4));
  ldp(x0, x1, MemOperand(x0));
  cmp(x0, RipReg);
  b(&FullLookup, Condition::ne);
  blr(x1)
```

## Follow ups
- [ ] Embed the pointers to some context param
- [ ] Simplify dispatcher (Compile and such is shouldn't be in assembly)
- [ ] Do distributed dispatching
- [ ] Do block linking

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -981,20 +981,36 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
     }
   };
 
+  aarch64::Label FullLookup{};
+
   // We want to ensure that we are 16 byte aligned at the top of this loop
   Align16B();
 
   bind(&LoopTop);
   AbsoluteLoopTopAddress = GetLabelAddress<uint64_t>(&LoopTop);
 
-  // This is the block cache lookup routine
-  // It matches what is going on it BlockCache.h::FindBlock
-  ldr(x0, &l_PagePtr);
-
   // Load in our RIP
   // Don't modify x2 since it contains our RIP once the block doesn't exist
   ldr(x2, MemOperand(STATE, offsetof(FEXCore::Core::ThreadState, State.rip)));
   auto RipReg = x2;
+
+  // L1 Cache
+  LoadConstant(x0, Thread->BlockCache->GetL1Pointer());
+
+  and_(x3, RipReg, 1 * 1024 * 1024 - 1);
+  add(x0, x0, Operand(x3, Shift::LSL, 4));
+  ldp(x1, x0, MemOperand(x0));
+  cmp(x0, RipReg);
+  b(&FullLookup, Condition::ne);
+  blr(x1);
+  b(&LoopTop);
+  
+  // L1C check failed, do a full lookup
+  bind(&FullLookup);
+
+  // This is the block cache lookup routine
+  // It matches what is going on it BlockCache.h::FindBlock
+  ldr(x0, &l_PagePtr);
 
   // Mask the address by the virtual address size so we can check for aliases
   if (__builtin_popcountl(VirtualMemorySize) == 1) {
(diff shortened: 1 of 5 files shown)


phire:
Overall I think this cache is a good idea. 

I do think the hashing function could be improved though later experimentation. But not for this commit.

At the very least, dropping the bottom two-four or even six bits might be smart. Maybe a multiplication to mix in some of the upper bits would be smart.

skmp:
Updated PR to only include the L1C change, and not also distributed dispatch

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -875,6 +888,14 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
     cmp(rax, 0);
     je(NoBlock);
 
+    // Update L1
+    mov(r13, Thread->BlockCache->GetL1Pointer());
+    mov(rcx, rdx);
+    and_(rcx, 1 * 1024 * 1024 - 1);
```
Sonicadvance1, line 894: Don't use hard coded L1C size.
Pull a constant from the BlockCache class instead

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -836,18 +836,31 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
...
-    mov(r13, Thread->BlockCache->GetPagePointer());
-
     // Load our RIP
     mov(rdx, qword [STATE + offsetof(FEXCore::Core::CPUState, rip)]);
 
+    // L1 Cache
+    mov(r13, Thread->BlockCache->GetL1Pointer());
+    mov(rax, rdx);
+
+    and_(rax, 1 * 1024 * 1024 - 1);
```
Sonicadvance1, line 853: Don't use hard coded L1C size.
Pull a constant from the BlockCache class instead

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -1029,12 +1045,18 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
...
+    ldr(x3, MemOperand(x0, offsetof(FEXCore::BlockCache::BlockCacheEntry, HostCode)));
+    cbz(x3, &NoBlock);
 
     // If we've made it here then we have a real compiled block
     {
-      blr(x0);
+      // update L1 cache
+      LoadConstant(x0, Thread->BlockCache->GetL1Pointer());
+
+      and_(x1, RipReg, 1 * 1024 * 1024 - 1);
```
Sonicadvance1, line 1056: Don't use hard coded L1C size.
Pull a constant from the BlockCache class instead

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -981,21 +981,37 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
...
-
   // Load in our RIP
   // Don't modify x2 since it contains our RIP once the block doesn't exist
   ldr(x2, MemOperand(STATE, offsetof(FEXCore::Core::ThreadState, State.rip)));
   auto RipReg = x2;
 
+  // L1 Cache
+  LoadConstant(x0, Thread->BlockCache->GetL1Pointer());
+
+  and_(x3, RipReg, 1 * 1024 * 1024 - 1);
```
Sonicadvance1, line 1000: Don't use hard coded L1C size.
Pull a constant from the BlockCache class instead

Sonicadvance1 (commented):
The entire constprop change seems to be unrelated?

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
skmp: Done

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
skmp: Done

PR #503, opened 2020-11-17
Remove duplicated compaction pass

Sonicadvance1:
RA Pass and passmanager compaction pass don't need to be independent.
This saves some memory.

In the future we should allow passes to fetch passes by name, that will
be part of the PassManager improvement in the future


skmp:
Instead of reaching out and coupling to the pass manager directly, why not pass the CompactionPass as a pointer on RA's constructor? Seems much cleaner that way.

skmp (changes requested):
Instead of reaching out and coupling to the pass manager directly, why not pass the CompactionPass as a pointer on RA's constructor? Seems much cleaner that way.

skmp (approved):

PR #509, opened 2020-11-17
Fixes cache clear bug on x86_64

Sonicadvance1:
On cache clear this would munmap pages 1.5x past the code cache.
This most certainly leads to spurious bugs


skmp (approved):
Nice catch!

PR #529, opened 2020-11-25
Shebang rootfs prefix improvements

Sonicadvance1:
If we are passed a shebang application then append the rootfs prefix for
absolute paths.
Fixes the issue where bash scripts weren't working on ARM.

Also fixes a crash that can occur if FEX wasn't passed in arguments
or if the argument passed in didn't exist.
It'll now instead exit cleanly and state that the application doesn't
exist respectively.


skmp (approved):
looks good

PR #532, opened 2020-11-25
GdbServer fixes

phire:
Fix start paused and add a hacky segfault signal handling.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/GdbServer.cpp
+++ b/External/FEXCore/Source/Interface/Core/GdbServer.cpp
@@ -25,18 +25,35 @@
 namespace FEXCore
 {
 
-void GdbServer::Break() {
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -829,6 +846,7 @@ GdbServer::HandledPacketType GdbServer::handleBreakpoint(std::string &packet) {
 }
 
 GdbServer::HandledPacketType GdbServer::ProcessPacket(std::string &packet) {
+  LogMan::Msg::D("GdbServer: packet %s\n", packet.c_str());
```
phire, line 849: These debug logs shouldn't be here.

Sonicadvance1 (approved):

PR #533, opened 2020-11-26
Remove 8/16bit ROR and all ROL IR ops as arm doesn't support them

phire:
This eliminates several IR ops that lie about support on the ARM 64 backend and moves the workarounds into Opcode Dispatcher

Alternative to #528


Sonicadvance1 (approved):

skmp:
Nice!

PR #540, opened 2020-11-29
A wackload of new 32bit syscall support

Sonicadvance1:
This implements a bunch of new 32bit syscalls which almost gets FEX to
be able to run Steam.

The main missing syscalls are mremap, shmat, and ioctl for supporting
Steam execution.


Sonicadvance1 (commented):
Needs #539 merged, then this rebased before this will pass tests.

Sonicadvance1:
Also will cause compile failure once #541 is merged. Will need to be rebased at that point.

skmp (approved):
Impressive!

PR #555, opened 2020-12-03
ConstProp: Masking elimination

skmp:
These are several small improvements

- Add RemoveUselessMasking, implement for a few ops
- Documented different sub-passes, remove leftover continue; statements
- Special case BFE, AND in RemoveUselessMasking sub-pass
- SBFE useless masking elim, VMOV useless zexting
- Elim Bfe on ops that zext by default
- Remove some BFEs that do nothing


Sonicadvance1:
Once this is merged, our minimum clang compiler version required will be Clang 10 instead of Clang 9.

Sonicadvance1 (approved):
Should probably stop dumping the world in to ConstProp.

PR #579, opened 2020-12-10
Removes SHM and non-unified memory!

Sonicadvance1:
Removes the SHM region which was only used for purposes that didn't matter.
Removes non-unified memory path since we didn't really support it anyway.
Fixes mlockall unit test since it no longer tries making 64GB of memory resident.


skmp (approved):
NICE!

PR #584, opened 2020-12-14
Fixes I found when running ASAN

Sonicadvance1:
Turns out the past few PRs broke ASAN cleanliness and had a few other warnings.
Fix that.


skmp (approved):
looks good

PR #589, opened 2020-12-20
Removes dead HandleSyscall function

Sonicadvance1:
The virtual function is called directly, so we were missing the strace
functionality

Strace and HandleSyscall just moved to the base implementation class instead and deduplicated.


skmp (approved):
Looks good!

Issue #561, opened 2020-12-03
Make sure that tests that pass don't crash


skmp:
Follow up from #559

Sometimes when tests crash with sigsegv they get treated as if they passed. We need to investigate why and make sure it doesn't happen to any other tests

PR #606, opened 2021-01-06
Have tests that throw a non-HLT SIGSEGV fail

duck-37:
This patch moves HandleSIGSEGV in HostFactory.cpp to a non-frontend host signal handler, then registers its own frontend signal handler to catch unhandled segmentation faults. 

Related issue: #561

Still needs an intentionally failing test to verify, so I'm marking this as a draft for now.

### DIFF
--- a/Source/Tests/TestHarnessRunner.cpp
+++ b/Source/Tests/TestHarnessRunner.cpp
@@ -102,7 +102,7 @@ int main(int argc, char **argv, char **const envp) {
   std::unique_ptr<FEXCore::HLE::SyscallHandler> SyscallHandler{FEX::HLE::CreateHandler(Loader.Is64BitMode() ? FEXCore::Context::OperatingMode::MODE_64BIT : FEXCore::Context::OperatingMode::MODE_32BIT, CTX, SignalDelegation.get())};
 
   bool DidFault = false;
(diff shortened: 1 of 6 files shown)


duck-37:
Marking this as ready for review; currently, the closest thing to an expected-to-fail test case is Known_Failures which just ignores the result. In addition, setting up a system like that is out of the scope of this PR.

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -101,6 +101,12 @@ int main(int argc, char **argv, char **const envp) {
   std::unique_ptr<FEX::HLE::SignalDelegator> SignalDelegation = std::make_unique<FEX::HLE::SignalDelegator>();
   std::unique_ptr<FEXCore::HLE::SyscallHandler> SyscallHandler{FEX::HLE::CreateHandler(Loader.Is64BitMode() ? FEXCore::Context::OperatingMode::MODE_64BIT : FEXCore::Context::OperatingMode::MODE_32BIT, CTX, SignalDelegation.get())};
 
+  bool DidFail = false;
```
skmp, line 104: Probably best to name this `DidFault`

duck-37: Good point.

Sonicadvance1 (approved):

Sonicadvance1:
Very nice first PR! Thanks for the change!

PR #618, opened 2021-01-08
Dispatch: Distributed Dispatch, IR Optimizations

skmp:
## Overview

- [x] Use jumps instead of calls for arm64
- [x] Dispatch on exits instead of centralized dispatch
- [x] Minimize context round trips for branches
- [x] Use jumps instead of calls for x86

## Validation before merge
- [x] Pass tests

## Follow ups
- [ ] Support `InlineConstants` for `ExitFunction`
- [ ] Don't save rip to ctx if not needed
- [ ] Block linking

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -768,11 +768,6 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const 
   uint32_t SSACount = IR->GetSSACount();
 
   auto HeaderOp = IR->GetHeader();
-
-  
-  if (HeaderOp->ShouldInterpret) {
-    return reinterpret_cast<void*>(ThreadSharedData.InterpreterFallbackHelperAddress);
-  }
   
   #ifndef NDEBUG
   LoadConstant(x0, HeaderOp->Entry);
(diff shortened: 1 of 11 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -813,64 +804,90 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
...
+      // Stop the thread
+      LoadConstant(x0, ThreadPauseHandlerAddressSpillSRA);
+      br(x0);
+    }
+    bind(&RunBlock);
   }
 
-  PendingTargetLabel = nullptr;
+  if (HeaderOp->ShouldInterpret) {
+    LoadConstant(x0, ThreadSharedData.InterpreterFallbackHelperAddress);
```
Sonicadvance1, line 832: Why does this change to indirect rather than just passing back `ThreadSharedData.InterpreterFallbackHelperAddress`?
This adds a small trampoline to every interpreter fallback for no reason that I can see.

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -628,137 +625,160 @@ void *JITCore::CompileCode([[maybe_unused]] FEXCore::IR::IRListView<true> const
...
-    mov(rcx, reinterpret_cast<uintptr_t>(SamplingData));
-    rdtsc();
-    shl(rdx, 32);
-    or(rax, rdx);
-    mov(qword [rcx + offsetof(BlockSamplingData::BlockData, Start)], rax);
+    L(RunBlock);
   }
 
-  auto ExitBlock = [&]() {
+  if (HeaderOp->ShouldInterpret) {
```
Sonicadvance1, line 647: Same comment as the AArch64 side.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
skmp: Because of gdb stub header, and later on of unsynchronized (implicit) RIP

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
skmp: Because of gdb stub header, and later on of unsynchronized (implicit) RIP

Sonicadvance1 (approved):

PR #622, opened 2021-01-10
Fixes execve with rootfs

Sonicadvance1:
We need to check to see if the application wanting to be executed is in
the rootfs first. Otherwise we end up in a case where we either lose
tracking of applications in FEX, or applications fail to launch since
they don't exist in the global filesystem


skmp (approved):
Looks good

PR #624, opened 2021-01-10
Have EmulatedFiles support osrelease file

Sonicadvance1:
This was being checked by an application for kernel version. Which was
seeing too new of a kernel and using things that were unsupported.

Also removes the EmulatedMap which is an unnecessary lookup.


skmp (approved):
Looks good

PR #625, opened 2021-01-10
Shebang improvements

Sonicadvance1:
Fixes a problem where a space between the shebang and application
argument would result in us trying to execute a null program.

Additionally it now captures all the arguments passed in the shebang
header rather than just the application and first argument.


skmp (approved):
:+1:

PR #626, opened 2021-01-10
Fixes 16bit ModRM rm decoding

Sonicadvance1:
When we are running in 32bit mode then it is possible to have an
instruction's modrm working as 16bit mode.
All one needs to do is prefix the instruction with an address size
override


skmp:
@Sonicadvance1 Can we have some tests for this one?

skmp (approved):
Looks good

PR #631, opened 2021-01-10
Pass DumpIR to TestHarnessRunner

duck-37:
DumpIR was not passed through the config to CTX in the TestHarnessRunner, leading to confusion when running it independently from the tests.


skmp (approved):
Looks good!

PR #632, opened 2021-01-10
Fix a few branching instructions on x86-32

duck-37:
JCXZ (X86 + address size prefix) was not processed correctly here, potentially leading to infinite loops.


duck-37:
Turns out quite a few rip-relative branching instruction was broken due to sign extension issues.

Sonicadvance1 (approved):
Very nice catch on these. Very surprised it wasn't breaking more things.

PR #639, opened 2021-01-11
Makes sure to not calculate zero sized program headers on 32bit

Sonicadvance1:
This is the same thing as the 64bit side


skmp (approved):
Looks good

PR #640, opened 2021-01-11
Disables killpg tests and reenables a few tests

Sonicadvance1:
killpg has been causing us testing problems for a while now.
As we were running more tests it was becoming more likely to hit a test
doing killpg in the background.

All of the tests run under the same process group so it is a significant
race condition to have these running.
Until we improve CI to somehow work around how killpg works, we just
need to have these disabled.


skmp (commented):
Looks good

skmp (approved):

PR #641, opened 2021-01-11
Uninstalls both the signal delegator and logmanager handlers on shutdown

Sonicadvance1:
Can prevent some crashing.


skmp (approved):
Looks good

PR #642, opened 2021-01-11
Fixes 64bit ELF loading when LLVM ASAN is used

Sonicadvance1:
Instead of hardcoding 64bit dynamic ELF files to 0x1'0000'0000 we
instead calculate the size of the ELF in memory. Then allocate the
ELF base up front.

LLVM ASAN steals a large amount of virtual memory space that intersected
with the original hardcoded range which caused memory allocations to
fail.


skmp (approved):
Looks good

PR #643, opened 2021-01-12
Support one-byte horizontal adds on x64

duck-37:


Sonicadvance1 (approved):
What a ...fun interleave dance.
Thanks for doing this.

PR #644, opened 2021-01-12
Add specialized case for PMOVMSKB with vector ops

duck-37:


duck-37:
Depends on #643.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -4005,6 +4005,24 @@ void OpDispatchBuilder::MOVMSKOp(OpcodeArgs) {
   StoreResult(GPRClass, Op, CurrentVal, -1);
 }
 
+void OpDispatchBuilder::MOVMSKOpOne(OpcodeArgs) {
+  OrderedNode *CurrentVal = _Constant(0);
+  OrderedNode *Src = LoadSource(FPRClass, Op, Op->Src[0], Op->Flags, -1);
+
+  auto M = _Constant(0x80'40'20'10'08'04'02'01ULL);
+  OrderedNode *VMask = _VCastFromGPR(16, 8, M);
```
Sonicadvance1, line 4013: Might be nice to implement a `VDUPFromGPR` IR op to remove this VCastFromGPR+VInsGPR pair.
This would map directly to AArch64's dup instruction and remove another instruction from this emulation.

Either add a comment about this minor opt and open and issue to track it, or do it it in this PR, making sure it is implemented in all three backends.
I'm fine with either route, otherwise looks good.

duck-37: Will add a comment, I think adding an additional opcode is most likely out-of-scope here.

Sonicadvance1 (approved):
Awesome optimization. This was implemented so naively before that it was completely terrible.

PR #646, opened 2021-01-12
Minor additions to constant propagation

duck-37:
1. Invert add/sub instructions with a negative constant to open up more folds into the operation itself
2. Fold X | ~0ULL -> ~0ULL
3. Fold X | 0ULL -> X
4. Fold X ^ 0ULL -> X 
5. Fold constant select operations

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -501,13 +501,22 @@ bool ConstProp::Run(IREmitter *IREmit) {
       uint64_t Constant1{};
       uint64_t Constant2{};
 
-      if (IREmit->IsValueConstant(Op->Header.Args[0], &Constant1) &&
-          IREmit->IsValueConstant(Op->Header.Args[1], &Constant2)) {
+      auto Is1CST = IREmit->IsValueConstant(Op->Header.Args[0], &Constant1);
+      auto Is2CST = IREmit->IsValueConstant(Op->Header.Args[1], &Constant2);
+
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -559,11 +580,22 @@ bool ConstProp::Run(IREmitter *IREmit) {
...
-      if (IREmit->IsValueConstant(Op->Header.Args[0], &Constant1) &&
-          IREmit->IsValueConstant(Op->Header.Args[1], &Constant2)) {
+      auto Is1CST = IREmit->IsValueConstant(Op->Header.Args[0], &Constant1);
+      auto Is2CST = IREmit->IsValueConstant(Op->Header.Args[1], &Constant2);
+
+      if (Is1CST && Is2CST) {
         uint64_t NewConstant = Constant1 | Constant2;
         IREmit->ReplaceWithConstant(CodeNode, NewConstant);
         Changed = true;
+      } /* OR with ~0 is ~0 */ else if ((Is1CST && Constant1 == ~0ULL) || (Is2CST && Constant2 == ~0ULL)) {
```
Sonicadvance1, line 590: These should be checking against a constant generated by `IROp->Size * 8 - 1` So it works for both 32bit and 64bit.
Then replaced with that constant as well.

duck-37: Converting this to a draft PR until I fix this + make sure there aren't any behavior regressions.

Sonicadvance1:
Closing from lack of activity.
If you want to bring this back then you can do this later.

PR #655, opened 2021-01-14
Allows JITSymbols to see central dispatcher

Sonicadvance1:
Will let us see in perf how much time is spent there


skmp (approved):
Looks good

PR #657, opened 2021-01-15
OpDisp: Improved Block Ordering

skmp:
## Overview

When blocks aren't ordered in program order we get global variables that we can't spill. With enough unspillable vars we end up failing to compile the function.

This
- Adds an option to insert blocks right after the current block or at then end of the function
- Modifies OpDisp to insert blocks in program order
- Modifies branches for slightly better perf for fall through edge cases
- Optimizes some conditionals is OpDisp
- Adds a test with loads or REPs that requires this optimization

##
- [x] Add option to insert block after current one
- [x] Update code for it
- [x] Add new test for repeated rep XX use
- [x] Make it clearer after which block blocks are inserted
- [x] Make sure all cases are optimal

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -656,14 +656,14 @@ namespace FEXCore::Context {
 
             auto InvalidateCodeCond = Thread->OpDispatcher->_CondJump(CodeChanged);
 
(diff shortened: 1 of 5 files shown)


skmp:
@Sonicadvance1 can you check my `InsertAfter` logic? Not sure it's 100% correct with the rest of IR machinery.

Sonicadvance1 (approved):

PR #665, opened 2021-01-18
Frontend, OpDisp: Don't assert on invalid instructions

skmp:
## Overview
Allows multiblock to compile blocks where some paths lead to invalid / unsupported instructions.

## Details
- Moves REX handling to `Decoder::DecodeInstruction` from `HandleOpHeader`
- Makes `HandleOpHeader` and `HandleOp` return false for unhandled cases instead of asserting
- Adds `HasInvalidInstruction` to `DecodedBlocks`
- Updates `DecodeInstructionsAtEntry` to set `HasInvalidInstruction` to true for blocks with invalid instructions, or assert if the entry block as an invalid insn.
- Updates `Context::CompileCode` to not compile blocks with invalid instructions


Sonicadvance1 (approved):
Nice! I hadn't gotten around to this.

PR #670, opened 2021-01-19
IR: Optimize runtime of optimization passes

skmp:
## Overview

This does some easy optimizations to the runtime of our optimization passes. In general, there are no big algorithmic changes, this focuses on improving the data structures. RA is substantially changed.

## Details
- Reworked `RA::CalculateNodeInterference` to almost O(N) from O(N!)
- Simplified and optimized data structures & initialization in RA
- Reworked `RCLSE::FindMemberInfo` to be O(1) from O(N)
- Merged DSE passes. Makes them ~ 2.8x as fast.
- Replaced some `std::map`s with `std::unordred_maps`. This was benchmarked, as it's not always a net-win.
- DSE directly looks for OP_JUMP/OP_CONDJUMP instead of looping
- Removed the concept of Virtual Registers from RA
- Optimized RA allocation from O(Regs * Interferences * Nodes) to O(Interferences * Nodes)
- Optimized RA to stop allocating after first spill
- Optimized RA to do compaction after spilling, as the input comes pre-compacted
- Optimized IR Compaction to do fewer lookups, and to only memset for debug builds
- Validated that IR output is exactly the same as before in all of Bytemark

## Data
clang prompt takes 2.3s before this, and 0.90s after (with `-n 4000 -m`)


skmp:
Found a major bug in the RA changes, converting to draft

skmp:
Added some more optimizations, removed a few bugs that crept in, and validated IR output is the same as before. This is ready for review now @Sonicadvance1

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -9,59 +9,173 @@
...
-  constexpr uint32_t DEFAULT_INTERFERENCE_LIST_COUNT = 128;
+  constexpr uint32_t INVALID_REG = 31;
+  constexpr uint32_t INVALID_CLASS = 7;
+  constexpr uint32_t DEFAULT_INTERFERENCE_LIST_COUNT = 120;
   constexpr uint32_t DEFAULT_NODE_COUNT = 8192;
-  constexpr uint32_t DEFAULT_VIRTUAL_REG_COUNT = 1024;
+
+  constexpr uint64_t INVALID_REGCLASS = (((uint64_t)INVALID_CLASS) << 32) | (INVALID_REG);
+
+  template<unsigned _Size = 6, typename T = uint32_t>
```
phire, line 19: Could do with a comment explaining why we have a bucketlist and how the constants have been chosen.

External/FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
```diff
@@ -0,0 +1,329 @@
...
+          if (IsTrackedWriteFPR(Op->Offset, IROp->Size))
+            BlockInfo.fpr.writes |= FPRBit(Op->Offset, IROp->Size);
+          else
+            BlockInfo.fpr.reads |= FPRBit(Op->Offset, IROp->Size);
+        } else if (IROp->Op == OP_STORECONTEXTINDEXED ||
+               IROp->Op == OP_LOADCONTEXTINDEXED) {
+          auto& BlockInfo = InfoMap[BlockNode];
+
+          //// GPR ////
+          // We can't track through these
```
phire, line 176: We probably could if we added a restriction that OP_STORECONTEXTINDEXED was only valid for x87, which is the only place we use it.

skmp: It's also used for the segment registers afaik. @Sonicadvance1 thoughts?

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
skmp: Good point. Will get to it tomorrow

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -300,15 +385,22 @@ namespace FEXCore::IR {
        */
       uint64_t GetNodeRegister(uint32_t Node) override;
     private:
-    bool OptimizeSRA;
+      bool OptimizeSRA;
+      uint32_t SpillPointId;
+      
+      #define INFO_MAKE(id, Class) ((id) | (Class << 24))
```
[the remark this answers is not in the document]
skmp: I initially had a bitfield struct, but for some reason I don't recall I switched over to the macros. I'll investigate the struct again tomorrow, and switch over if there's no blocker

External/FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
```diff
@@ -0,0 +1,322 @@
...
+  else if (Size == 4)
+    return  1UL << (bitn);
+  else
+    LogMan::Throw::A(false, "Unexpected FPR size %d", Size);
+}
+
+struct Info {
+  FlagInfo flag;
+  GPRInfo gpr;
+  FPRInfo fpr;
```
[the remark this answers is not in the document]
skmp: I'd rather note this as a follow up task at this point, as the changeset is quite big already. Sounds good?

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
skmp: https://github.com/FEX-Emu/FEX/issues/675

External/FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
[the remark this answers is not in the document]
skmp: https://github.com/FEX-Emu/FEX/issues/675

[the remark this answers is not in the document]
phire: Yeah, I'm happy with followups.

External/FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
phire: Hmm, we might.

Either way, as long as we can prove it's only used for for a limited set of registers we can avoid this optimisation killer.

Sonicadvance1: It's only used for 32bit and LDT/GDT loading of segment register data. Very minor.

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
skmp: Added the comment and removed default size of 6 (it's explicitly set in all instances anyway)

External/FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
skmp: Created follow up https://github.com/FEX-Emu/FEX/issues/677

phire (approved):

PR #674, opened 2021-01-21
IR: Split parser from IRLoader, add IR::Parse(), Fixes, Tests

skmp:
## Overview
This moves IR parsing to `FEXCore::IR::Parse` from `IRLoader`, so that it works as a reusable component and not part of the IRTestHarness. Then, it has several fixes to the parser, dumper and other places so that we can round trip all blocks. It also enables round trip validation for the asm tests to make sure we don't break this again


External/FEXCore/include/FEXCore/IR/IR.h
```diff
@@ -287,29 +287,29 @@ struct MemOffsetType final {
 };
 
 struct TypeDefinition final {
-  uint8_t Val;
-  operator uint8_t() const {
+  uint16_t Val;
```
Sonicadvance1, line 290: Was this size change just to support vector element counts up to 255 elements in size?

skmp: Yes. The old setup supported vectors up to 3 elements, which is not enough, and packed it on 8 bits.

Sonicadvance1 (approved):

PR #701, opened 2021-01-25
OpDisp: Address most 32-bit rip size issues

skmp:
Some MB calculations are still wrong, See #702 and #703

### DIFF
--- a/External/FEXCore/include/FEXCore/IR/IREmitter.h
+++ b/External/FEXCore/include/FEXCore/IR/IREmitter.h
@@ -40,7 +40,8 @@ friend class FEXCore::IR::PassManager;
 
   IRPair<IROp_Constant> _Constant(uint8_t Size, uint64_t Constant) {
     auto Op = AllocateOp<IROp_Constant, IROps::OP_CONSTANT>();
-    Op.first->Constant = Constant;
+    uint64_t Mask = ~0ULL >> (Size - 64);
+    Op.first->Constant = (Constant & Mask);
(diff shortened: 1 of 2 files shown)


External/FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -40,7 +40,11 @@ friend class FEXCore::IR::PassManager;
 
   IRPair<IROp_Constant> _Constant(uint8_t Size, uint64_t Constant) {
     auto Op = AllocateOp<IROp_Constant, IROps::OP_CONSTANT>();
-    Op.first->Constant = Constant;
+    uint64_t Mask = 1ULL << Size;
+    if (Size == 64)
```
phire, line 44: Branchless mask calculation:

`Mask = ~0ULL >> (Size - 64);`

Works for all sizes except 0, which we don't care about here.

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -601,7 +601,7 @@ namespace FEXCore::Context {
   std::tuple<void *, FEXCore::Core::DebugData *> Context::CompileCode(FEXCore::Core::InternalThreadState *Thread, uint64_t GuestRIP) {
     uint8_t const *GuestCode{};
       GuestCode = reinterpret_cast<uint8_t const*>(GuestRIP);
-
+
```
phire, line 604: ?

phire (approved):
Otherwise, LGTM

PR #704, opened 2021-01-25
Add gcc target tests

skmp:
## Overview
This adds gcc's "gcc.target" tests for i386 and amd64. We mostly pass these, however there are a few failures that should be investigated (#713). The 32-bit tests don't work on CI but work for me locally, so that also needs to be investigated (#714).

There's test cases all the way up to avx512, though I haven't included anything avx*


skmp:
@Sonicadvance1 can you investigate why it fails on the runners? I suspect missing 32-bit libs from the rootfs?

phire (approved):

PR #705, opened 2021-01-25
Flush IR/Code cache on mmap, mmunmap & mprot

skmp:
## Motivation
To get gimp working w/o --smc

## Overview
This flushes the cache on `mmap` or `munmap`, and on `mprotect` w/ `PROT_EXEC` set. The idea is to catch .so load and unload, to avoid the actual need for smc tracking for non-jits.

This is a very naive implementation that uses block-lists per-page, and erases the entire page when touched.

With this, GIMP works at interactive speeds.

## Details
- [x] rebase on top of `skmp/ir-cache`
- [x] Keep blocks-in-page lists
- [x] Insert blocks on lists based on their start/length range
- [x] Invalidate on mmap, munmap, mprotect, 64 and 32 bits

## TODO
- [x] Modify `--smc` to have `--smc-chekcs=none`, `--smc-chekcs=mman` and `--smc-chekcs=full`. Make the default value `mman`

## Follow ups
- [x] Find better structures & algos for tracking and intersecting invalidation ranges (#807)
- [x] Make invalidation remove from all threads (#806)
- [x] Self-invalidation issues with interpreter, rare (#808)

### DIFF
--- a/Source/Tests/LinuxSyscalls/x64/Memory.cpp
+++ b/Source/Tests/LinuxSyscalls/x64/Memory.cpp
@@ -25,6 +25,7 @@ namespace FEX::HLE::x64 {
       uint64_t Result = ::munmap(addr, length);
       if (Result != -1) {
         FEXCore::Context::RemoveNamedRegion(Thread->CTX, (uintptr_t)addr, length);
+        FEXCore::Context::FlushCodeRange(Thread, (uintptr_t)addr, length);
       }
       SYSCALL_ERRNO();
     });
@@ -32,10 +33,13 @@ namespace FEX::HLE::x64 {
     REGISTER_SYSCALL_IMPL_X64(mmap, [](FEXCore::Core::InternalThreadState *Thread, void *addr, size_t length, int prot, int flags, int fd, off_t offset) -> uint64_t {
       static FEXCore::Config::Value<bool> AOTIRLoad(FEXCore::Config::CONFIG_AOTIR_LOAD, false);
       uint64_t Result = reinterpret_cast<uint64_t>(::mmap(addr, length, prot, flags, fd, offset));
-      if (Result != -1 && !(flags & MAP_ANONYMOUS)) {
-        auto filename = get_fdpath(fd);
+      if (Result != -1) {
+        if (!(flags & MAP_ANONYMOUS)) {
+          auto filename = get_fdpath(fd);
 
-        FEXCore::Context::AddNamedRegion(Thread->CTX, Result, length, offset, filename);
+          FEXCore::Context::AddNamedRegion(Thread->CTX, Result, length, offset, filename);
+        }
+        FEXCore::Context::FlushCodeRange(Thread, (uintptr_t)Result, length);
       }
       SYSCALL_ERRNO();
     });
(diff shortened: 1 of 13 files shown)


phire:
Yeah, this looks like the right approach.

As for future data structures, for this case (of gimp unloading libraries, then loading new libraries in their place), we only really need to track which mmap allocation a code block is in. We can just invalid the code for the whole .so file on munmap.

We could make such a cache adaptable. If a region is mmapd to write + execute, we can track just that region on a per-page basis.

Source/Tests/LinuxSyscalls/x64/Memory.cpp
```diff
@@ -25,17 +25,21 @@ namespace FEX::HLE::x64 {
...
-      if (Result != -1 && !(flags & MAP_ANONYMOUS)) {
-        auto filename = get_fdpath(fd);
+      if (Result != -1) {
+        if (!(flags & MAP_ANONYMOUS)) {
+          auto filename = get_fdpath(fd);
 
-        FEXCore::Context::AddNamedRegion(Thread->CTX, Result, length, offset, filename);
+          FEXCore::Context::AddNamedRegion(Thread->CTX, Result, length, offset, filename);
+        }
+        FEXCore::Context::FlushCodeRange(Thread, (uintptr_t)Result, (uintptr_t)Result + length);
```
phire, line 42: Do we really need to flush the code range?

Nothing was previously here, and if something was here earlier, we *should* have flushed it on unmap

Amanieu: Flushing is required if `MAP_FIXED` is used.

Amanieu:
Derek Bruening's [PhD thesis](http://www.burningcutlery.com/derek/docs/phd.pdf) describes a system for invalidating code blocks (called "fragments") across multiple thread. The relevant part starts on page 156.

However it is quite complex: you need to be able to hotpatch branch instructions and update lookup tables while another thread is executing translated code. At that point it might be worth simply switching to a single shared code cache, which has many other benefits. See [this paper](http://www.burningcutlery.com/derek/docs/threadshared-CGO06.pdf) for details on implementing a thread-shared code cache.

Source/Tests/LinuxSyscalls/x64/Memory.cpp
skmp: @phire flushing is fast when there's no backing code, so i've added it just to be sure. We don't catch all mmaps yet (namely, we don't catch mmaps from the elf iinterpreter), though I don't think anything will unload that

skmp:
@Amanieu interesting. I'll create a follow up ticket with that information, as solving that problem looks much more involved.

Source/Tests/LinuxSyscalls/x64/Memory.cpp
```diff
@@ -47,6 +51,9 @@ namespace FEX::HLE::x64 {
 
     REGISTER_SYSCALL_IMPL_X64(mprotect, [](FEXCore::Core::InternalThreadState *Thread, void *addr, size_t len, int prot) -> uint64_t {
       uint64_t Result = ::mprotect(addr, len, prot);
+      if (Result != -1 && prot & PROT_EXEC) {
```
phire, line 54: Thinking about this. Shouldn't we also be flushing when the region was previously PROT_EXEC and the new mprotect is not PROT_EXEC?

Otherwise a program running under fex could theoretically continue executing cached code on a non-executable page. Maybe there is some form of jit out there that relies on segfaults from stale code to make sure they are removed from block-linking? 

Then again, I'm pretty sure our current jit doesn't even check page executablity before emitting code. Should we create an issue and push this down the line?

phire (commented):
Apart from the missing cmd line arguments, LGTM.

Sonicadvance1 (approved):

Sonicadvance1 (approved):

PR #715, opened 2021-01-27
Fixes some asserts from vixl and a bug in ExitSpillSRA 

Sonicadvance1:
Assembler by default doesn't allow you to use the assembler. Needs a
CodeBufferCheckScope.

Labels with binds and not links will throw an assert.
ExitSpillSRA was using b instead of bind incorrectly.
Thus branching to a invalid location


skmp (approved):
nice catch

PR #724, opened 2021-01-29
Core/Context Refactor

skmp:
## Overview
This majorly refactors how code gets compiled, and cleans up a lot of things in the way. Removes the LookupCache, IRLists from the Compile Service, simplifies how caches are cleared, stores RAData so that IRLists can be correctly re-compiled.

## Depends
#722 #723


skmp:
(This can't be separated from #722, so closing this one)

PR #730, opened 2021-01-30
AOTIR: Initial implementation

skmp:
## Overview
This adds IR caching support. In FEXLoader, `--aotir-capture` will catpure ir during execution and generate `~/.fex-emu/aotir/<module>-<pathhash>-<optoptions>` for all elfs loaded by the process, and `--aotir-load` will load them.

Currently, only one process can update the aotir cache, so programs that fork won't work well with`--aotir-capture`

IR Caches hash verify the guest code (usting some fasthash), though we likely want to use native CRC support in arm for faster hashing.

IR Caches are also relocated on use.

I've successfully run clang, clang++, FTL, gnome-calculator and gnome-text-editor. Load time speedup is 2x to 3x vs non-cached.

IR Loading is done via c++ streams, and could be much further optimized.

To append to the cache, both `--aotir-load` and `--aotir-capture` need to be passed to `FEXLoader`.

This also includes a fix for `exit_group`, which GLib programs need to exit cleanly.

## Details
- Adds `--aotir-capture` and `--aotir-load`
- Introduces a new IR Op, EntrypointOffset to handle relocations
- Makes IRListView non-templated
- Adds AOTCache to Context/Core.cpp
- Adds Addr -> File mappings to Context/Core.cpp
- Adds `AddNamedRegion` and `RemoveNamedRegion` to Context/Core.cpp to keep track of addr -> file mappings
- Adds `WriteAOTIRCache` and `SetAOTIRLoader`
- Changes OP_THUNK to use sha256 hashes instead of strings, so that it can be serialized
- Adds `IsShared` to `RAData` and `IRListView`, updates deleters, so that data from AOTIR can be imported directly to the thread local lists
- Refactors `IRLists` and `RALists` and `DebugData` to `LocalIRCache`

## Todo
- [x] Fix memory ownership and double free issues
- [x] Add InlineEntrypointOffset an for ExitFunction
- [x] Support OP_THUNK (needs to be serialized differently)

## Follow ups
- Optimize hasing via 2x CRC32 or XXH3 (#797)
- Hash function chunks only (#798)
- Switch to a flag for ir args w/o dst (#799)

## Future work
- AOTIR static generator
- Optimize cache loading via mmap


Source/Tests/ELFLoader.cpp
```diff
@@ -287,8 +293,44 @@ int main(int argc, char **argv, char **const envp) {
...
   }
 
+  std::string base_filename = Program.substr(Program.find_last_of("/\\") + 1) + ".fex-emu.iraot";
+
+  if (AOTIRLoad() || AOTIRCapture()) {
+    LogMan::Msg::I("Warning: AOTIR is experimental, and might lead to crashes. Capture doesn't work with programs that fork.");
+  }
+
+  FEXCore::Context::SetAOTIRLoader(CTX, [](const std::string &fileid) -> std::unique_ptr<std::istream> {
+    auto filepath = std::string(getenv("HOME")) + "/.fex-emu/aotir/" + fileid;
```
phire, line 303: We really should be following the XDG standard.

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -1013,4 +1241,27 @@ namespace FEXCore::Context {
     return Result;
   }
 
+  void Context::AddNamedRegion(uintptr_t Base, uintptr_t Size, uintptr_t Offset, const std::string &filename) {
+    // TODO: Support overlapping maps and region splitting
+    auto base_filename = filename.substr(filename.find_last_of("/\\") + 1);
```
phire, line 1246: Use `std::filesystem`

Source/Tests/LinuxSyscalls/x64/Memory.cpp
```diff
@@ -1,19 +1,57 @@
...
+#include <FEXCore/Core/Context.h>
+#include <FEXCore/Config/Config.h>
+#include <fstream>
+
+std::string get_fdpath(int fd)
+{
+    std::vector<char> buf(400);
+    ssize_t len;
+
+    std::string fdToName = "/proc/self/fd/" + std::to_string(fd);
```
phire, line 20: Use `std::filestem::path` for path concatenation

Source/Tests/LinuxSyscalls/x64/Memory.cpp
```diff
@@ -1,19 +1,57 @@
...
+
+    do
+    {
+        buf.resize(buf.size() + 100);
+        len = ::readlink(fdToName.c_str(), &(buf[0]), buf.size());
```
phire, line 25: `std::filesystem::absolute()`

skmp: I just want to resolve the softlink there, I don't think ::abosulte() does that

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -27,6 +27,49 @@
...
 
+namespace {
+  // Compression function for Merkle-Damgard construction.
+  // This function is generated using the framework provided.
+  #define mix(h) ({					\
+        (h) ^= (h) >> 23;		\
+        (h) *= 0x2127599bf4325c37ULL;	\
+        (h) ^= (h) >> 47; })
+
+  static uint64_t fasthash64(const void *buf, size_t len, uint64_t seed)
```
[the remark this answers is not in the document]
Sonicadvance1: Also use xxhash's xxh3 hash instead of this one.
Ran xxhash's benchmark comparing the two and xxh3 wins across the board on both x86 and AArch64

Source/Tests/LinuxSyscalls/x64/Memory.cpp
phire: whoops, you want `std::filesystem::canonical()` instead

phire: I don't think readlink guarantees a canonical path, it only follows one simlink

External/FEXCore/Source/Interface/Core/Core.cpp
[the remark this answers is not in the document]
skmp: Will do on follow up, #797

External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
```diff
@@ -209,8 +210,10 @@ class JITCore final : public CPUBackend, public Xbyak::CodeGenerator {
 
   ///< ALU Ops
   DEF_OP(TruncElementPair);
-  DEF_OP(Constant);
+  DEF_OP(Constant);
```
phire, line 213: Trailing whitespace

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -535,6 +537,8 @@ namespace FEXCore::IR {
         for (uint8_t i = 0; i < NumArgs; ++i) {
           if (IROp->Args[i].IsInvalid()) continue;
           if (IR->GetOp<IROp_Header>(IROp->Args[i])->Op == OP_INLINECONSTANT) continue;
+          if (IR->GetOp<IROp_Header>(IROp->Args[i])->Op == OP_INLINEENTRYPOINTOFFSET) continue;
+          if (IR->GetOp<IROp_Header>(IROp->Args[i])->Op == OP_IRHEADER) continue;
```
phire, line 541: This should be based on an IR.json property. Can be done in a follow up ticket

phire (approved):

PR #747, opened 2021-02-01
Android build fixes

phire:
These are all the changes I needed to make while trying to get FEX built under termux on my android phone.

I've put this porting effort on hold for now, as NDK r21 doesn't implement `std::filesystem` and termux are still working on updating to r22. 

Apart from `std::filesystem` issues, I manged to get FEXCore building. The LinuxEmulation layer is going to need some more work. 

I have another PR coming which allows building and running IRLoader without LinuxEmulation (which it doesn't use).

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
@@ -26,6 +26,7 @@
 #ifdef _M_X86_64
 #include <xmmintrin.h>
 #endif
+#include <unistd.h>
 
 namespace FEXCore::CPU {
 
@@ -397,13 +398,19 @@ Res GetSrc(void* SSAData, IR::OrderedNodeWrapper Src) {
 [[noreturn]]
 static void StopThread(FEXCore::Core::InternalThreadState *Thread) {
   Thread->CTX->StopThread(Thread);
-  std::unexpected();
+
+  // If the StopThread signal is delayed for any reason, spin forever
+  for (;;)
+    sleep(1);
 }
 
 [[noreturn]]
 static void SignalReturn(FEXCore::Core::InternalThreadState *Thread) {
   Thread->CTX->SignalThread(Thread, FEXCore::Core::SIGNALEVENT_RETURN);
-  std::unexpected();
+
+  // If the return signal is delayed for any reason, spin forever
+  for (;;)
+    sleep(1);
 }
 
 void InterpreterOps::InterpretIR(FEXCore::Core::InternalThreadState *Thread, FEXCore::IR::IRListView<true> *CurrentIR, FEXCore::Core::DebugData *DebugData) {
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
@@ -352,7 +352,7 @@ bool IsConditionTrue(uint8_t Cond, uint64_t Src1, uint64_t Src2) {
     case FEXCore::IR::COND_ULE:
       CompResult = static_cast<unsigned_type>(Src1) <= static_cast<unsigned_type>(Src2);
       break;
-    
+
     case FEXCore::IR::COND_FLU:
       CompResult = reinterpret_cast<float_type&>(Src1) < reinterpret_cast<float_type&>(Src2) || (std::isnan(reinterpret_cast<float_type&>(Src1)) || std::isnan(reinterpret_cast<float_type&>(Src2)));
       break;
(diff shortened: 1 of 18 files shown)


Source/Tests/CMakeLists.txt
```diff
@@ -1,78 +1,21 @@
...
-  LinuxSyscalls/Syscalls/Signals.cpp
-  LinuxSyscalls/Syscalls/Socket.cpp
-  LinuxSyscalls/Syscalls/Thread.cpp
-  LinuxSyscalls/Syscalls/Time.cpp
-  LinuxSyscalls/Syscalls/Timer.cpp
-  LinuxSyscalls/Syscalls/NotImplemented.cpp
-  LinuxSyscalls/Syscalls/Stubs.cpp
-  )
-set(LIBS FEXCore Common CommonCore pthread)
-set(NAME FEXLoader)
```
phire, line 60: I removed all the ${NAME} stuff because it made this file very hard to follow and know what ${NAME} was referencing at any-one point.

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -397,13 +398,19 @@ Res GetSrc(void* SSAData, IR::OrderedNodeWrapper Src) {
 [[noreturn]]
 static void StopThread(FEXCore::Core::InternalThreadState *Thread) {
   Thread->CTX->StopThread(Thread);
-  std::unexpected();
+
+  // If the StopThread signal is delayed for any reason, spin forever
+  for (;;)
```
skmp, line 403: How about `LogMan::Msg::A` here?

External/FEXCore/Source/Utils/ELFLoader.cpp
```diff
@@ -412,7 +412,7 @@ void ELFContainer::CalculateSymbols() {
       uint64_t offset = SymTabHeader->sh_offset + i * SymTabHeader->sh_entsize;
       Elf32_Sym const *Symbol =
           reinterpret_cast<Elf32_Sym const *>(&RawFile.at(offset));
-      if (ELF32_ST_VISIBILITY(Symbol->st_other) != STV_HIDDEN &&
+      if (Symbol->st_other & 0x3 != STV_HIDDEN &&
```
skmp, line 415: Why was the macro expanded here? It's much cleaner to have it wrapped in a macro

skmp (changes requested):
Looks overall good to me, apart from the noted comments. Probably @Sonicadvance1 should sign off this as well as i'm not too familiar with our cmake files

External/FEXCore/Source/Utils/ELFLoader.cpp
phire: The macro doesn't exist on android's version of `elf.h`

Looks like it's a GNU libc extension to the elf.h library.

Sonicadvance1: Might be nice to define our own macro or lambda for this then :)

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
phire: I'm really not sure that tgkill has any kind of synchronicity guarantee for doing a self kill. 

It might be somewhat common to reach here. 

I'm working on another PR which just stops the thread synchronously.

External/FEXCore/Source/Utils/ELFLoader.cpp
phire: Also, this is wrong. Need brackets around `Symbol->st_other & 0x3`

Source/Tests/CMakeLists.txt
Sonicadvance1: Reasonable enough

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
Sonicadvance1: tgkill to itself will immediately switch the thread to the signal handler. It's abused in the Linux syscall handler for pending signal handling.

phire: Hmm... I checked the linux source code and couldn't really see a code path which guaranteed that.

External/FEXCore/Source/Utils/ELFLoader.cpp
phire: Ok, I'm now redefining the macro if it's not defined.

External/FEXCore/Source/Utils/ELFLoader.cpp
```diff
@@ -435,7 +435,7 @@ void ELFContainer::CalculateSymbols() {
       uint64_t offset = DynSymTabHeader->sh_offset + i * DynSymTabHeader->sh_entsize;
       Elf32_Sym const *Symbol =
           reinterpret_cast<Elf32_Sym const *>(&RawFile.at(offset));
-      if (ELF32_ST_VISIBILITY(Symbol->st_other) != STV_HIDDEN &&
+      if (ELF64_ST_VISIBILITY(Symbol->st_other) & 0x3 != STV_HIDDEN &&
```
Sonicadvance1, line 438: Name flipped from other one and oops on still doing & 3?

phire: oops

Sonicadvance1 (approved):

Sonicadvance1 (approved):

skmp (approved):

PR #749, opened 2021-02-02
Remove dependency on signals for clean exits

phire:
Allows consumers of the FEXCore API to run and exit cleanly without having to implement linux signal handling.

As an example, I've removed the linux emulation dependency on IRLoader tests.

### TODO: 

 * [x] ARM jit implementation.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
@@ -482,4 +482,13 @@ void InterpreterCore::DeleteAsmDispatch() {
   delete Generator;
 }
 
+bool InterpreterCore::IsAddressInJITCode(uint64_t Address, bool IncludeDispatcher) {
+  if (!IncludeDispatcher)
+    return false;
+
+  size_t Start = reinterpret_cast<size_t>(Generator->getCode());
+  size_t End = Start + Generator->getSize();
+
+  return Address >= Start && Address < End;
+}
 }
--- a/External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
@@ -1,5 +1,6 @@
 #include "Common/MathUtils.h"
 #include "Interface/Core/Interpreter/InterpreterClass.h"
+#include "Interface/Core/Interpreter/Dispatcher_asm.h"
 #include "Interface/Context/Context.h"
 #include <FEXCore/Core/X86Enums.h>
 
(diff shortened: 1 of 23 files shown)


External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
```diff
@@ -449,6 +455,15 @@ void InterpreterCore::CreateAsmDispatch(FEXCore::Context::Context *ctx, FEXCore:
   DispatchPtr = Generator->DispatchPtr;
   CallbackPtr = Generator->CallbackPtr;
 
+  using PivotFnType = __attribute__((naked)) void(*)(uint64_t);
+  auto PivotFn = reinterpret_cast<PivotFnType>(Generator->ThreadStopHandlerPivotStackAddress);
+
+// This needs to be on the thread object, because we (currently) create one dispatcher per thread
+  Thread->LongJumpExit = std::make_unique<std::function<void(FEXCore::Core::InternalThreadState *)>>([PivotFn](FEXCore::Core::InternalThreadState *Thread) {
+    PivotFn(Thread->State.ReturningStackLocation);
+    // Does not return
```
Sonicadvance1, line 464: std::unexpected() could go here

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -1076,6 +1079,16 @@ void JITCore::CreateCustomDispatch(FEXCore::Core::InternalThreadState *Thread) {
...
 #endif
 
+  // Provide a way for c++ code to long jump to ThreadStopHandler
+  using PivotFnType = __attribute__((naked)) void(*)(uint64_t);
+  auto PivotFn = reinterpret_cast<PivotFnType>(ThreadStopHandlerPivotStackAddress);
+
+  // This needs to be on the thread object, because we (currently) create one dispatcher per thread
+  Thread->LongJumpExit = std::make_unique<std::function<void(FEXCore::Core::InternalThreadState *)>>([PivotFn](FEXCore::Core::InternalThreadState *Thread) {
+    PivotFn(Thread->State.ReturningStackLocation);
+    // Does not return
```
Sonicadvance1, line 1089: std::unexpected() could go here

Sonicadvance1 (commented):
I think this is completely reasonable.
Setting stack on ARM side is fairly trivial.
`add (rsp, x0, 0);` to do a stack move.

External/FEXCore/Source/Interface/Core/Interpreter/x86_64Dispatcher.cpp
Sonicadvance1: womp womp, std::unexpected removed from C++17 and higher.

phire:
After skmp pointed out that not cleaning up stackframes (which the old implementation didn't do either) was bad, I'm looking into an unwinding based approach.

Current version cleanly unwinds the current frame, running all the deconstructors, until it hits jit or dispatcher. But it can't see previous stackframes of callbacks or signals, so they don't get cleaned up.

skmp:

Sonicadvance1:
Reopen if progress is made.

PR #751, opened 2021-02-02
Syscalls: chmodat, faccessat glibc interface differs from kernel

skmp:


phire (approved):

Sonicadvance1:
Ah cool, looks like this was accidentally implemented as faccessat2 signature. Which is kernel 5.8 and we don't expose it

PR #755, opened 2021-02-05
Pulls uname nodename from host system

### LINKED ISSUE #600: Pull uname nodename from host system.
Currently we just populate it with "FEXCore" on both 32bit and 64bit LInux.
This causes Steam to state "FEXCore is available for streaming".
Let's nab this from the host uname instead.

Sonicadvance1:
Hardcoding FEXCore as the nodename is an annoyance.
Pull the host's nodename instead.

Fixes #600


Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
```diff
@@ -19,6 +19,23 @@
...
+        static_assert(sizeof(Local.nodename) <= 65);
+      }
+      else {
+        strcpy(buf->nodename, "FEXCore");
+      }
+      strcpy(buf->sysname, "Linux");
+      strcpy(buf->release, "5.0.0");
+      strcpy(buf->version, "#" FEXCORE_VERSION);
+      // Tell the guest that we are a 64bit kernel
+      strcpy(buf->machine, "x86_64");
```
skmp, line 35: Should this really be "x86_64" for x32 syscalls?

Sonicadvance1: Yes. This is the exact behaviour of calling uname with a 32bit executable.
it's the machine architecture of the kernel, not the program itself.

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
```diff
@@ -19,6 +19,23 @@
 
 namespace FEX::HLE {
   void RegisterInfo() {
+    REGISTER_SYSCALL_IMPL(uname, [](FEXCore::Core::InternalThreadState *Thread, struct utsname *buf) -> uint64_t {
+      struct utsname Local{};
+      if (::uname(&Local) == 0) {
+        strncpy(buf->nodename, Local.nodename, sizeof(Local.nodename));
```
phire, line 25: From a readability perspective, it might be smarter to just do a memcpy here. 

It will stop people who read this code from wondering if strncpy does the right thing here, or hit's it's edge case.  

It does, but only because the uname syscall will have correctly null-terminated the original string.

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
```diff
@@ -19,6 +19,23 @@
...
+        static_assert(sizeof(Local.nodename) <= 65);
+      }
+      else {
+        strcpy(buf->nodename, "FEXCore");
```
phire, line 29: Should probably `LogMan::Msg::E` here.

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
```diff
@@ -19,6 +19,23 @@
...
+      }
+      strcpy(buf->sysname, "Linux");
+      strcpy(buf->release, "5.0.0");
+      strcpy(buf->version, "#" FEXCORE_VERSION);
```
phire, line 33: Should `static_assert` that `sizeof("#" FEXCORE_VERSION) <= 65`

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
Sonicadvance1: Fixed

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
Sonicadvance1: Added a message about it.

Source/Tests/LinuxSyscalls/Syscalls/Info.cpp
Sonicadvance1: Added an assert

phire (approved):

PR #759, opened 2021-02-07
Fix power of 2 OP_MUL optimisation (Fixes Geekbench result upload) 

### LINKED ISSUE #647: Geekbench fails to upload results
After testing (due to another bug, you currently need to test with `-T 1` geekbench fails to upload results to the server with:

```
Uploading results to the Geekbench Browser. This could take a minute or two 
depending on the speed of your internet connection.

 unknown error (internal code 35)
```

Before exiting.

phire:
The new left-shift that replaced the multiply takes
it's size from arg[0]. If arg[0] was 64bit and the original
OP_MUL was 32bit, then the original code would truncate
upper bits, and the replacement code wouldn't.

This bug broke the SSL cert checking code in geekbench,
causing it to fail to upload results.
Fixes #647


skmp (approved):
Nice catch

PR #764, opened 2021-02-09
Fix the number of physical cpus reported

### LINKED ISSUE #662: Investigate: Geekbench detects invalid CPU topology
When run with `-T 4` to report four threads, geekbench 4 detects the CPU topology as: 

> 4 Processors, 16 Cores

[see full report](https://browser.geekbench.com/v4/cpu/15997950)

I have no idea where it's getting 16 cores from, this is a 12 core system.  
 htop reports that It's only actually running 4 threads of execution, as expected.

phire:
Fixes #662


Sonicadvance1 (approved):
Lol, oops.

PR #769, opened 2021-02-11
x87: Call C++ handlers instead of forcing interpreter

skmp:
## Overview

This adds a "fallback" path for unimplemented opcodes in the JITs, and adds an interface to query the interpreter for opcode "fallbacks". The interface is semi-generic so it can be used to fallback any opcode, however only the x87 opcodes have been implemented. `ShouldInterpret` & All related logic has been removed.

This passes all tests now

### x86-64 jit
- Add `JIT::PushRegs` && `JIT::PopRegs`
- Remove xmm0 from allocated regs as it is the return value
- Add xmm11 as an allocated reg, and use xmm0 as scratch reg in its place
- Add fallbacks for the abis used by x87 handlers
- Implement VBSL
- Remove `ShouldInterpret` logic
### arm64 jit
- Add fallback logic
- Remove empty F80Cmp handler
- Remove `ShouldInterpret` logic
### OpDisp
- Make sure that F80CVTTo ops use FPR/GPR arguments as they should
- Replace VSTOREMEMELEMENT with VEXTRACT
- Mark only BCD opcodes as fallbacks
- Remove `ShouldInterpret` logic
### InterpreterOps
- Add API to query handler / abi info
- Implement handlers (copy paste from interpreter execution for now)

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
@@ -51,8 +51,8 @@ namespace FEXCore::CPU {
 using namespace Xbyak::util;
 const std::array<Xbyak::Reg, 9> RA64 = { rsi, r8, r9, r10, r11, rbp, r12, r13, r15 };
 const std::array<std::pair<Xbyak::Reg, Xbyak::Reg>, 4> RA64Pair = {{ {rsi, r8}, {r9, r10}, {r11, rbp}, {r12, r13} }};
-const std::array<Xbyak::Reg, 11> RAXMM = { xmm0, xmm1, xmm2, xmm3, xmm4, xmm5, xmm6, xmm7, xmm8, xmm9, xmm10 };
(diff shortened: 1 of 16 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
```diff
@@ -51,8 +51,8 @@ namespace FEXCore::CPU {
 using namespace Xbyak::util;
 const std::array<Xbyak::Reg, 9> RA64 = { rsi, r8, r9, r10, r11, rbp, r12, r13, r15 };
 const std::array<std::pair<Xbyak::Reg, Xbyak::Reg>, 4> RA64Pair = {{ {rsi, r8}, {r9, r10}, {r11, rbp}, {r12, r13} }};
-const std::array<Xbyak::Reg, 11> RAXMM = { xmm0, xmm1, xmm2, xmm3, xmm4, xmm5, xmm6, xmm7, xmm8, xmm9, xmm10 };
```
Sonicadvance1, line 54: Why is xmm0 removed here? We have something like xmm11-xmm15 reserved for temps?

skmp: xmm0 is used as the return value in functions, I've added xmm11 to the list and changed the code to use xmm0 as temp.

Sonicadvance1 (approved):

PR #778, opened 2021-02-15
x87: Round, Truncate & Precision control

skmp:
## Overview
Adds rounding mode, precision control and implements fisttp* as truncating conversions.

Fixes rendering in Counter Strike 1.6, fixes valgrind's `insn_fpu` & `insn_sse3`

## Details
- [x] Add FCW to the context
- [x] Add `OP_F80LOADFCW` that sets the F80 rounding mode, precision
- [x] Update F80 Handlers to use the precision specified instead of hardcoding
- [x] Extend `OP_F80CVTINT` to have a `bool Truncate` to support `FISTTP*`
- [x] Make sure all fpu ops to use FCW instead of constants

## Follow Ups
- #779, #780

### DIFF
--- a/External/FEXCore/Source/Interface/IR/IRParser.cpp
+++ b/External/FEXCore/Source/Interface/IR/IRParser.cpp
@@ -82,6 +82,15 @@ class IRParser: public FEXCore::IR::IREmitter {
 	}
 
   template<>
+  std::pair<DecodeFailure, bool> DecodeValue(std::string &Arg) {
+    if (Arg.at(0) != '#') return {DecodeFailure::DECODE_INVALIDCHAR, 0};
+
+    uint8_t Result = strtoul(&Arg.at(1), nullptr, 0);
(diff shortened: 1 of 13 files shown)


External/FEXCore/Source/Interface/IR/IRParser.cpp
```diff
@@ -81,6 +81,15 @@ class IRParser: public FEXCore::IR::IREmitter {
...
 	}
 
+  template<>
+  std::pair<DecodeFailure, bool> DecodeValue(std::string &Arg) {
+    if (Arg.at(0) != '#') return {DecodeFailure::DECODE_INVALIDCHAR, 0};
+
+    uint8_t Result = strtoul(&Arg.at(1), nullptr, 0);
+    if (errno == ERANGE || Result > 1) return {DecodeFailure::DECODE_INVALIDRANGE, 0};
+    return {DecodeFailure::DECODE_OKAY, Result != 0};
+	}
```
phire, line 91: indent wrong

phire (approved):

PR #781, opened 2021-02-16
X87: FPREM needs to set C2 to 0 to indicate finished iteration

skmp:
## Overview
This is needed for Half Life 1 & 2, as well as most Source titles.

This is not an accurate model of how the instructions works

## Follow ups
#782

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -6865,6 +6865,11 @@ void OpDispatchBuilder::X87BinaryOp(OpcodeArgs) {
   // Overwrite the op
   result.first->Header.Op = IROp;
 
+  if (IROp == IR::OP_F80FPREM) {
+    //TODO: Set C0 to Q2, C3 to Q1, C1 to Q0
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -6865,6 +6865,11 @@ void OpDispatchBuilder::X87BinaryOp(OpcodeArgs) {
   // Overwrite the op
   result.first->Header.Op = IROp;
 
+  if (IROp == IR::OP_F80FPREM) {
+    //TODO: Set C0 to Q2, C3 to Q1, C1 to Q0
```
phire, line 6869: Can we expand on this and say something like: "FREP is not correctly implemented at all, but this hack is better than nothing, preventing infinite loops"

phire (approved):

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
skmp: Well, it does give correct results - as long as one doesn't use C0, C1 or C3 and doesn't depend on the iterative nature

PR #789, opened 2021-02-22
CPUID: Add cache information, function 0x2

skmp:
glibc has a bug in the cache size lookup logic and mistakenly calculates the cache size as 0 if no information is given.

memcpy has a further bug that it assumes that the size of cache is > 144 bytes, leading to out of bound memcpys.

This adds a basic cpuid 0x2 function, and also adds a log for unhandled cpuid bits in debug bulds.

Fixes glibc2.33 that uses ssse3 memcpy that has the bug.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/External/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -109,6 +109,31 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_01h() {
   return Res;
 }
 
+// 2: Cache and TLB information
+FEXCore::CPUID::FunctionResults CPUIDEmu::Function_02h() {
+  FEXCore::CPUID::FunctionResults Res{};
+
+  // returns default values from i7 model 1Ah
+  Res.eax = 0x1 | // Number of iterations needed for all descriptors
+    (0x5A << 8) |
+    (0x03 << 16) |
+    (0x55 << 24);
+
+  Res.ebx = 0xE4 |
+    (0xB2 << 8)  |
+    (0xF0 << 16) |
+    (0 << 24);
+
+  Res.ecx = 0; // null descriptors
+
+  Res.edx = 0x2C |
+    (0x21 << 8)  |
+    (0xCA << 16) |
+    (0x09 << 24);
+
+  return Res;
+}
+
 FEXCore::CPUID::FunctionResults CPUIDEmu::Function_06h() {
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -109,6 +109,29 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_01h() {
...
+    (0x3 << 16) | // Data TLB: 4 KByte pages, 4-way set associative, 64 entries
+    (0x6 << 24);  // L1I: 8K
+
+  Res.ebx = 0xC | // L1D: 16K
+    (0x21 << 8) | // L2: 256K
+    (0x22 << 16) | // L3: 512K
+    (0 << 24);  // null descriptor
+
+  Res.ecx = 0 | // null descriptors
+    (1 << 31); // Invalid data
```
phire, line 127: This is "reserved for future use" according to the intel manual

Perhaps better to just fill this this register (and edx) with 4 null descriptors (aka, `0x00000000`)

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -109,6 +109,29 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_01h() {
   return Res;
 }
 
+// 2: Cache and TLB information
+FEXCore::CPUID::FunctionResults CPUIDEmu::Function_02h() {
+  FEXCore::CPUID::FunctionResults Res{};
+
+  Res.eax = 0x1 | // lowest byte always 0x01
```
phire, line 116: Well... the Intel manual actually goes into more detail about how this is the count for the you need to read this cpuid to get every single cache/tlb descriptor byte.

But I think all intel CPUs do actually fit everything inside the 15 bytes supplied here, so it's probably true that it's always 0x1.

phire (approved):

External/FEXCore/Source/Interface/Core/CPUID.cpp
phire: This is also what intel do in their example, register ecx is filled with all zeros.

image

skmp: The docs I was using say otherwise, but they are far older. I'll update to also return this table, because why not.

phire:
Checked, still works

PR #790, opened 2021-02-22
Threading: Workaround exit_group bug

skmp:
## Overview
This works around the issue of us stopping threads that are in stopping state, by only sending the stop signals to non-stopped threads.

Resolves the std::terminate called recursively issue in many multi threaded apps, and also allows for clean shutdown for --aotir-capture


phire (approved):
I think this is an acceptable workaround.

Sonicadvance1 (approved):

PR #793, opened 2021-02-23
Adds assert checks on missing LOCK support

Sonicadvance1:
These usually don't get hit, but Geekbench4 DOES manage to hit LOCK on
BTS.

Which we just don't support right now.


Sonicadvance1:
This will need to be rebased once #795 is merged.

skmp (approved):

PR #801, opened 2021-02-24
Fixes CMPXCHG flags being incorrect aside from ZF

Sonicadvance1:
Almost everything only checks ZF but we had the arguments reversed for
the rest of the comparison flag results.


skmp (approved):
Nice catch!

PR #803, opened 2021-02-24
Implements support for faccessat2 in syscallhandler

Sonicadvance1:
Only exists if the host kernel is >= 5.8.0.


skmp:
Should we add an implementation that returns -ENOSYS if kernel < 5.8.0 so that we don't hit the assert in Unimplemented?

skmp (approved):
😍

PR #812, opened 2021-03-05
Allow both ARM64 and X86_64 jits to be compiled at the same time

phire:
The main goal of this PR is to minimise the amount of code which is only compiled on one platform. This should allow for easier refactoring of code in both jits.

This is not the old Vixel simulator. A secondary goal is that one FEXCore binary for one platform will be able to emit both x86_64 and Arm64 assembly. This might be useful in the future for a debug interface which shows both x86_on_x86 and arm64 code for comparison, or maybe static profiling tools that show differences in instruction counts between various backends. 

### How to use

 * By default, cmake will select the correct JIT for the host architecture. 
 * Setting -DENABLE_JIT_ARM64=on/off will override this decision. 
 * FexCore will always initialise the correct JIT. 
 * If the correct JIT isn't enabled in, FEXCore die at runtime with `"FEXCore has been compiled without a viable JIT core"` when trying to initialise a jit backend.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -29,7 +29,7 @@ static int64_t LREM(int64_t SrcHigh, int64_t SrcLow, int64_t Divisor) {
 
 using namespace vixl;
 using namespace vixl::aarch64;
-#define DEF_OP(x) void JITCore::Op_##x(FEXCore::IR::IROp_Header *IROp, uint32_t Node)
+#define DEF_OP(x) void Arm64JITCore::Op_##x(FEXCore::IR::IROp_Header *IROp, uint32_t Node)
 DEF_OP(TruncElementPair) {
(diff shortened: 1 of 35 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -524,7 +524,7 @@ DEF_OP(LDiv) {
       mov(x1, GetReg<RA_64>(Op->Header.Args[0].ID()));
       mov(x2, GetReg<RA_64>(Op->Header.Args[2].ID()));
 
-#if _M_X86_64
+#if VIXL_SIMULATOR
```
Sonicadvance1, line 527: These are leftover, can just delete anything you've changed over to `#if VIXL_SIMULATOR`. It isn't expected to work anymore.

phire: Removed

phire:
I've noticed that the dispatchers aren't both being built, but I'll do those changes in a later PR.

Sonicadvance1 (approved):

PR #814, opened 2021-03-06
Implements CPUID 0x8000'0006 for cacheline information

Sonicadvance1:
This exposes some cache information including cacheline information.

Fills out this data structure as well just incase we hit another cacheline size bug

### DIFF
--- a/External/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/External/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -375,6 +375,40 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0004h() {
   return Res;
 }
 
+// L2 Cache identifiers
+FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0006h() {
+  FEXCore::CPUID::FunctionResults Res{};
+
+  // L2 TLB Information for 2MB and 4MB pages
+  Res.eax =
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -375,6 +375,47 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0004h() {
...
+    (3 << 28);     // data TLB associativity type
+
+  // TLB Information for 4KB pages
+  Res.ebx =
+    (1024 << 0)  | // Number of TLB instruction entries
+    (6 << 12)    | // instruction TLB associativity type
+    (1536 << 16) | // Number of TLB data entries
+    (5 << 28);     // data TLB associativity type
+
+  // L2 cache identifiers
```
skmp, line 396: L1?

skmp (approved):

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Nope, fn8000'0006 is about L2 in EAX, EBX, and ECX. Then L3 in EDX.
If you want L1 information then you need to use fn8000'0005.

Sonicadvance1:
Fixed a duplicated ecx being set. Oops.

PR #815, opened 2021-03-06
Implements MOVNTDQA

Sonicadvance1:
Found this when running Steamlink under FEX on my laptop.
Turns out the Iris video driver now uses this unconditionally in a couple of locations.
Makes sense there since all the Iris targets are expected to support SSE4.1 atm.


skmp (approved):
👌

PR #818, opened 2021-03-08
Thunk configs path added to configuration

Sonicadvance1:
Completely unused right at this moment.
Just place `FEXCore::Config::Value<std::string> Environment{FEXCore::Config::CONFIG_THUNKSCONFIGPATH, ""};` Wherever you want the config used at


Sonicadvance1:
Nope more.

PR #823, opened 2021-03-09
OpDisp: Add atomic logic for SecondaryALUOp

skmp:
This fixes at least one bug in pthreads, and possibly other things


skmp:
Do we want unit tests for these @Sonicadvance1 ? Can we test for atomics?

Sonicadvance1:
We can do atomic unit tests but they will behave like a single thread touching it.
Which we have some to ensure correct behaviour but not all of them.
I think this is fine without but maybe we should go over all of our instructions that can support LOCK and ensure we are actually implementing it.
Or throwing an assert on the ones we know we don't support, like the few that have that.

Sonicadvance1 (approved):

PR #829, opened 2021-03-10
Commandline: -T 0 will get cpu count from host

skmp:
## Overview
When the threads parameter is set to zero, it will query the host for the number of cpu cores

### DIFF
--- a/Source/Common/ArgumentLoader.cpp
+++ b/Source/Common/ArgumentLoader.cpp
@@ -1,3 +1,5 @@
+#include <sys/sysinfo.h>
+
 #include "Common/ArgumentLoader.h"
 #include "Common/Config.h"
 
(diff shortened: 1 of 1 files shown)


Source/Common/ArgumentLoader.cpp
```diff
@@ -215,6 +217,9 @@ namespace FEX::ArgLoader {
 
       if (Options.is_set_by_user("Threads")) {
         uint64_t Config = Options.get("Threads");
+        if (Config == 0) {
+          Config = get_nprocs();
```
Sonicadvance1, line 221: You probably want `get_nprocs_conf` due to how aggressively ARM devices park CPU cores.
`get_nprocs` will only return the current number of online cores.

PR #830, opened 2021-03-10
Arm64 jit: Remove unneeded branch Link

phire:


phire:
turns out it was needed.

PR #831, opened 2021-03-10
Unify all four dispatchers

phire:
The goal of this PR is to unify all four dispatchers to use as much common code as possible

The four dispatchers are ArmJit, x86Jit, ArmInterpreter, x86Interpreter. These were all originally copy/pasted from one original implementation, so their code mostly matches. Most differences are missing functionality. 

This PR saves 900 lines of code and will hopefully make changes to this code easier. 

### Design overview: 

 * We have a Dispatcher base class: Dispatcher
   * Mostly code which was copy/pasted in both jits
   * Mostly Signal handling
   * Holds all the pointers into the dispatcher
   * Holds a bit of state
   * Deals with `IsAddressInJit`
 * There are two sub-classes for Arm64 and X86_64.
   - These also inherit the two emitters and emit the dispatchers
   - Arm64 also knows hows to spill SRA from a context
 * A new "Arm64Emitter" class was added to contain utility emitting functions shared by the jit emitter and dispatcher emitter
   - Helpful functions like LoadConstant
   - I kind of just moved anything here that wasn't aware of FEXes IR. 
   - Goes into the inheritance chain in-between vixl and Arm64Jit/Arm64Dispatcher
 * Both Jits have been thinned down. 
   - They have a dispatcher as a component. 
   - Gets a little messy, after constructing a dispatcher, they need to copy a few things out. 
     - This would require wider changes to clean up.

Both dispatchers are built by default. This does make vixl a dependency even on x86, but x86 is only a debug platform and vixl is pretty small. LTO will eliminate it.


External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
```diff
@@ -0,0 +1,365 @@
...
+using namespace vixl::aarch64;
+
+static constexpr size_t MAX_DISPATCHER_CODE_SIZE = 4096;
+#define STATE x28
+
+Arm64Dispatcher::Arm64Dispatcher(FEXCore::Context::Context *ctx, FEXCore::Core::InternalThreadState *Thread, DispatcherConfig &config)
+  : Dispatcher(ctx, Thread) {
+  SRAEnabled = config.StaticRegisterAssignment;
+
+  //buffer_ = CodeBuffer(MAX_DISPATCHER_CODE_SIZE);
```
phire, line 26: This shouldn't be commented out

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -0,0 +1,219 @@
...
+
+#include <FEXCore/Utils/LogManager.h>
+#include <FEXCore/Core/CoreState.h>
+
+#include "aarch64/cpu-aarch64.h"
+
+namespace FEXCore::CPU {
+#define STATE x28
+
+Arm64Emitter::Arm64Emitter() : vixl::aarch64::Assembler(vixl::aarch64::PositionDependentCode) {
```
phire, line 11: The intention here is that Assembler will allocate an empty codebuffer, which arm64jit can replace later.

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -0,0 +1,343 @@
...
+      ArchHelpers::Context::SetPc(ucontext, ThreadStopHandlerAddressSpillSRA);
+    } else {
+      if (SRAEnabled) {
+        // We are in non-jit, SRA is already spilled
+        LogMan::Throw::A(!IsAddressInJITCode(ArchHelpers::Context::GetPc(ucontext), true), "Signals in dispatcher have unsynchronized context");
+      }
+      ArchHelpers::Context::SetPc(ucontext, ThreadStopHandlerAddress);
+    }
+
+    ArchHelpers::Context::SetPc(ucontext, ThreadStopHandlerAddress);
```
phire, line 288: BUG

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -0,0 +1,343 @@
...
+  // We can end up getting a signal at any point in our host state
+  // Jump to a handler that saves all state so we can safely return
+  uint64_t OldSP = ArchHelpers::Context::GetSp(ucontext);
+  uintptr_t NewSP = OldSP;
+
+  size_t StackOffset = sizeof(ArchHelpers::Context::ContextBackup);
+
+  // We need to back up behind the host's red zone
+  // We do this on the guest side as well
+  NewSP -= 128;
```
Sonicadvance1, line 33: ARM doesn't have a redzone in Linux, this shouldn't be necessary there?

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: BUG

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.h
```diff
@@ -0,0 +1,84 @@
...
+#include <FEXCore/Core/SignalDelegator.h>
+
+#include "Interface/Context/Context.h"
+
+#include <stack>
+
+namespace FEXCore::CPU {
+
+struct DispatcherConfig {
+  bool Interpeter = false;
```
Sonicadvance1, line 13: Can we rename this to something like `InterpreterBackend`?
Immediately it confused me between Interpreter CPU core or Linux Interpreter

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
```diff
@@ -0,0 +1,325 @@
...
+      mov(rax, rdx);
+
+      and_(rax, LookupCache::L1_ENTRIES_MASK);
+      shl(rax, 4);
+      cmp(qword[r13 + rax + 8], rdx);
+      jne(FullLookup);
+      jmp(qword[r13 + rax + 0]);
+    }
+
+    L(FullLookup);
```
Sonicadvance1, line 93: Might want to test this on a debug build. xbyak might throw an assert if you bind a Label without using it.

Sonicadvance1 (commented):
I believe that's all the comments I have on this

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
phire: It works in debug. 

I'm not seeing any "bound but not used" errors in xbyak.h. Just: 

 * `ERR_LABEL_IS_REDEFINED`
 * `ERR_LABEL_IS_TOO_FAR`
 * `ERR_LABEL_IS_NOT_FOUND`
 * `ERR_LABEL_ISNOT_SET_BY_L`
 * `ERR_LABEL_IS_ALREADY_SET_BY_L`

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.h
phire: Renamed to `ExecuteBlockWithCall` which better describes what this config actually does.

Sonicadvance1 (approved):

PR #836, opened 2021-03-11
Adds config value option constructor without default initializer

Sonicadvance1:
This has the expected behaviour that the configuration option will be available and not default initialized.
To enforce this fact, it will assert if it tries to get a value that doesn't exist yet.

Currently unused but should be used for any configuration option that is guaranteed to have a value.

### DIFF
--- a/External/FEXCore/include/FEXCore/Config/Config.h
+++ b/External/FEXCore/include/FEXCore/Config/Config.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <FEXCore/Core/Context.h>
+#include <FEXCore/Utils/LogManager.h>
 
 #include <list>
 #include <memory>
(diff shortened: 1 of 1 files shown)


External/FEXCore/include/FEXCore/Config/Config.h
```diff
@@ -135,6 +136,29 @@ namespace Type {
       GetListIfExists(Option, &AppendList);
     }
 
+    template <typename TT = T,
+      typename std::enable_if<!std::is_same<TT, std::string>::value, int>::type = 0>
+    Value(FEXCore::Config::ConfigOption _Option)
+      : Option {_Option} {
+      if (!FEXCore::Config::Exists(Option)) {
+        LogMan::Msg::A("FEXCore::Config::Value has no value");
```
phire, line 144: Should these be `ERROR_AND_DIE` instead of `LogMan::Msg::A`, which gets disabled in release builds?

Sonicadvance1: You're right. Let me change over to that

phire (approved):

PR #848, opened 2021-03-13
Remove most warnings from FEX

Sonicadvance1:
This removes a significant number of warnings from FEX.
Avoids changing code that would conflict with #831 though.

Remaining warnings end up being in externals or the `offsetof(FEXCore::Context::Context, Config.RunningMode)` warning


phire:
Is it possible to exclude "unused variables" from -Werror?

Sonicadvance1:
Sure. we could use ` -Wno-error=unused-variable`

phire:
Perhaps we should have two options, `ENABLE_WERROR` which enables most errors (balanced for development) , and `ENABLE_STRICT_WERROR` which enables all errors for CI testing.

We could probally get away with enabling the less strict version by default

phire (approved):

PR #850, opened 2021-03-14
libclang based Struct verifier written in python

Sonicadvance1:
This is a fairly complex script that walks all structs and unions in a header and following some annotations tied to the declarations will compare those struct definitions to different struct "aliases" and across architectures.

This is important for:

- Ensuring our hand defined structs for syscalls matches what the syscall wants through aliasing rules
- Ensuring thunk generated structs match across the architecture boundary

We really need to rely on a script for this, hand checking every struct is prone to failure and won't catch every case.
Additional documentation: https://wiki.fex-emu.org/index.php/Development:StructPackVerifier


skmp (approved):
Very nice!

PR #851, opened 2021-03-15
Dynamic code DWARF unwinding infomation (sane backtraces)

phire:
This PR implements DWARF emission for dynamic code, which results in sane backtraces when examining FEX in gdb. Backtraces that go though the JIT.

For example:

image

This PR only supports one JIT frame. Multiple call frames will end up collapsed into a single [Unknown/Just-In-Time compiled code] block with any intermediary c++ code hidden. This will be fixed in a future PR. 

This PR also doesn't implement any unwinding. Just the info that will allow unwinding (in a future PR)

TODO: 

 * [ ] Implement GAS Dispatcher exit for Arm64
 * [ ] Remove Backtrace function which snuck in here. 
 * [ ] Different callee saved registers for Arm64
 * [ ] General testing on Arm64.

### DIFF
--- a/External/FEXCore/Source/CMakeLists.txt
+++ b/External/FEXCore/Source/CMakeLists.txt
@@ -1,3 +1,5 @@
+enable_language(ASM)
+
 set (SRCS
   Common/Paths.cpp
   Common/JitSymbols.cpp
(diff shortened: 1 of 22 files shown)


skmp:

Sonicadvance1:
Reopen if progress is made.

PR #860, opened 2021-03-17
Switch to using CpuStateFrames 

phire:
This PR Continues the work of #817

Previously we were hiding the fact that our signal handlers were recursive. This change makes it blindingly obvious.
It also makes debugging way easier when combined with #851 

This change also removes most instances of host context modification and fixes a few signal handing bugs.


skmp:

Sonicadvance1:
Reopen if progress is made.

PR #883, opened 2021-03-23
Cleanup threads when they exit

### LINKED ISSUE #760: Clean up thread state on exit
When a thread exits we don't fully clean up its state until process end. Effectively this leaks memory.
Clean it up and test with a fork bomb application.

### LINKED ISSUE #772: GDBServer: Dead threads aren't cleared out
They just stick around forever, with their program counter pointing at a syscall to `sys_exit` in pthread's `start_thread`

Probably related to #760

phire:
Massively reduces memory usage for programs that fire up large number of threads that only run for a short amount of time.  
This makes it viable to run 24 threads of geekbench without AOTIR (peeks at about 7gb of memory)

fixes #772 
fixes #760

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -25,6 +25,7 @@
 #include <fstream>
 #include <unistd.h>
 #include <filesystem>
+#include <algorithm>
 
 #include "Interface/Core/GdbServer.h"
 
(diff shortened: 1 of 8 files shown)


External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -621,6 +622,47 @@ namespace FEXCore::Context {
...
+    for (auto &DeadThread : Threads) {
+      if (DeadThread == ExceptForThread) {
+        continue;
+      }
+
+      // Setting running to false ensures that when they are shutdown we won't send signals to kill them
+      DeadThread->RunningEvents.Running = false;
+
+      // Clean up thread state
+      delete DeadThread;
```
phire, line 655: I'm really not sure about this. None of our tests actually hit this.

Assuming any program is evil enough to fork while it has threads, the `std::thread` thread is going to be desynced with reality. The threads should still be marked as `joinable()` which means deleting the object will `std::terminate()`

But they can't actually be joined or detached.

I suspect the only sane thing to do here is to do something evil to deallocate the std::thread without actually deconstructing it.

phire: It's fine, as long as we join. See comments in followup commit.

Sonicadvance1 (approved):
Looks very nice!
I think the compromises of still having leaks under fork are reasonable now.
Much smaller amounts of data are leaking rather than full thread states which is good.

Can you create an issue to track further forking leaks?

phire:
Ticket created.

PR #902, opened 2021-03-26
Dockerfile add arm64 compatibility

Azkali:
Python3, Python3-dev and linux-headers prevent from building on arm64

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,8 @@ FROM ubuntu:20.04 as builder
 RUN DEBIAN_FRONTEND="noninteractive" apt-get update
 RUN DEBIAN_FRONTEND="noninteractive" apt install -y cmake \
 clang-10 llvm-10 nasm ninja-build libnuma-dev \
-libcap-dev libglfw3-dev libepoxy-dev
+libcap-dev libglfw3-dev libepoxy-dev python3-dev \
+python3 linux-headers-5.8.0-23-generic
 
 COPY . /opt/FEX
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -4,7 +4,8 @@ FROM ubuntu:20.04 as builder
 RUN DEBIAN_FRONTEND="noninteractive" apt-get update
 RUN DEBIAN_FRONTEND="noninteractive" apt install -y cmake \
 clang-10 llvm-10 nasm ninja-build libnuma-dev \
-libcap-dev libglfw3-dev libepoxy-dev
+libcap-dev libglfw3-dev libepoxy-dev python3-dev \
+python3 linux-headers-5.8.0-23-generic
```
Sonicadvance1, line 8: Good catch.
Instead of explicitly defining the version number in the package, you should use Ubuntu's meta package of `linux-headers-generic`.
This will ensure that future updates from Canonical's repos don't break these rules.

Azkali: Thank you :)
Just updated the PR to use the meta package for headers.

Sonicadvance1 (approved):
Thanks for the patch!
Our Dockerfile is definitely lacking but always good to see contributions.

Azkali:
Would you like me to help with your Dockerfile, could you explain me why you are splitting it into stages and what are your goals with it ? I already have multiple ideas that you may or may not want

Sonicadvance1:
If you would like to discuss more, it maybe be good to join https://discord.gg/fexemu
I'm not super familiar with Docker myself but I believe the use case for Docker is reproducible build environments?
I'm not sure what the point of splitting in to stages is.

PR #911, opened 2021-03-28
Replaces FEX's usage of classic C array element count with std::size

Sonicadvance1:
Just learned that this is a c++17 feature and it's a nice little
cleanup.

Only one remains in our codebase which can't easily be replaced


skmp (approved):

skmp:
Nice!

PR #913, opened 2021-03-29
Adds a 32-bit mask for multiblock RIP calculation

### LINKED ISSUE #703: OpDisp: MB for 32-bit is broken
MB Calculations always do `uint64_t` math on RIP

Sonicadvance1:
In the case of overflow then it'll just mask this result.
The rest of the logic already matches what 32-bit expects.

Fixes #703


skmp (approved):
Hope that's the last one of these

skmp (approved):

PR #914, opened 2021-03-29
Context: Make use of std::optional with GetFilenameHash() 

lioncash:
Removes the need for an out-variable.

While we're in the same area, some minor other tidying up can be done.

Made just to get my feet wet with contributing :)


Sonicadvance1 (approved):
Yay! Thanks for the cleanup!

PR #915, opened 2021-03-30
OpcodeDispatcher: Add missing break for UD2 in INTOp()

lioncash:
Fixes a case where the reason gets overwritten.


Sonicadvance1 (approved):
Oops! Nice catch!

PR #918, opened 2021-03-30
CMakeLists: Flag unannotated implicit fallthrough as an error

lioncash:
Prevents a class of sneaky logic bugs from slipping through into the codebase.

This also resolves a case of such a bug within the Decorder's ReadData() where all 3 byte reads would be performed as if they were a 4 byte read.

### DIFF
--- a/External/FEXCore/Source/Common/SoftFloat-3e/extF80_roundToInt.c
+++ b/External/FEXCore/Source/Common/SoftFloat-3e/extF80_roundToInt.c
@@ -96,6 +96,7 @@ extFloat80_t
         switch ( roundingMode ) {
          case softfloat_round_near_even:
             if ( !(sigA & UINT64_C( 0x7FFFFFFFFFFFFFFF )) ) break;
+            __attribute__((fallthrough));
          case softfloat_round_near_maxMag:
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Common/SoftFloat-3e/extF80_roundToInt.c
```diff
@@ -96,6 +96,7 @@ extFloat80_t
         switch ( roundingMode ) {
          case softfloat_round_near_even:
             if ( !(sigA & UINT64_C( 0x7FFFFFFFFFFFFFFF )) ) break;
+            __attribute__((fallthrough));
```
Sonicadvance1, line 99: Since we are C++20 shouldn't this be `[[fallthrough]];`?

lioncash: I'd have used that, but this is in a C file

Sonicadvance1: Oh right. SoftFloat-3e problems...

Sonicadvance1 (approved):
Very nice!
Implicit fallthrough is so painful.

PR #922, opened 2021-03-30
Disables gvisor test proc_pid_oomscore_test

Sonicadvance1:
Depending on runner this passes or fails.
The test expects to be able to open `/proc/self/oom_score_adj` as writable to adjust the oom score.
This is expected to fail on all three of our runners, but sometimes the solidrun board manages to open it.
Disable as it is a flake. Could be a kernel bug on the solid run board


phire (approved):
Kill it with fire!

PR #923, opened 2021-03-31
Adds support for named rootfs configurations

### LINKED ISSUE #684: Default rootfs(s) location
.fex-emu/<rootfsname>/.... or some other XDG correct path

Sonicadvance1:
If the relative or absolute folder doesn't exist then FEX will search in the data folder for a rootFS named the same thing
If the path exists then it is used.

For example if my data folder contains `$HOME/.fex-emu/RootFS/Ubuntu_main/` and I set the RootFS option to `Ubuntu_main`
Then this rootfs will be used. Allows you to easily select a rootfs in that folder without having the full file path.

Relies on #916 being merged first
Fixes #684

### DIFF
--- a/External/FEXCore/Source/Interface/Config/Config.cpp
+++ b/External/FEXCore/Source/Interface/Config/Config.cpp
@@ -237,7 +237,12 @@ namespace FEXCore::Config {
       }
 
       // Expand relative path to absolute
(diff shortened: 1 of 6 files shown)


skmp (approved):
LGTM

PR #928, opened 2021-04-01
Switches FEXCore over to pthreads implementation 

Sonicadvance1:
This defaults to a pthread implementation but it can be switched over to a custom thread
handler if the frontend desires.


skmp (approved):
We can fork without crashes at last :dancers:

PR #939, opened 2021-04-02
Default hidden visibility and strip symbols

### LINKED ISSUE #934: Change default visibility to hidden
We expose too many symbols. Change everything to hidden

Sonicadvance1:
Set our default visibility to hidden. There is no need to have anything default visibility.

Additionally on release builds strip unused sections and symbols.

This shaves the binary down to 2MB on x86-64 host
1.9MB on an ARMv8.2 host

Fixes #934


Sonicadvance1:
Oops, need to fix symbol exports on shared library.
Magically worked before.

Sonicadvance1:
Fixed

phire (approved):

PR #945, opened 2021-04-03
Fixes truncate and creat

Sonicadvance1:
These were using the passthrough helper which doesn't was breaking the result value


skmp (approved):

skmp:
What was the exact cause of this? Maybe we should move away from the auto forwards?

PR #946, opened 2021-04-03
Fixes eventfd and eventfd2

Sonicadvance1:
glibc function signature doesn't match kernel


skmp (approved):

skmp:
Arm seems to only have SYS_eventfd2

Sonicadvance1:
Easy enough to fix

PR #961, opened 2021-04-06
Fixes segment register override on string instructions. 

### LINKED ISSUE #845: Go through every instruction and ensure segment prefixing supporting
We may be missing segment prefixing on instructions.
Go through all of them and ensure we haven't missed any.

Sonicadvance1:
The segment prefix only overrides one side of the string instruction.
Passing in the flags was causing the instruction to have the segment override on both sides of the copy.
So something like `movsb es:[rdi], fs:[rsi]` was being interpreted as `movsb fs:[rdi], fs:[rsi]`
Same for all the string ops which is now fixed.

We can't unit test these currently since we can't set host segment register for confirmed host behaviour.

Fixes #845


skmp (approved):
Nice catch!

PR #978, opened 2021-04-21
64-bit allocator and ioctl emulation

Sonicadvance1:


Sonicadvance1:
Needs an option introduced to make the Jetson's run times not be absolutely abysmal due to old kernel.
Will introduce that tomorrow.

Sonicadvance1:
This is now ready to get merged. Kernels prior to 4.17 have the fastest startup that they can possibly have now.

Sonicadvance1:
Xavier has an interesting failure case that only reproduces in the environment of old kernel + github-actions runner + tests are stealing the full VMA region.
I'll need to take the Xaviers down to update the kernel which will resolve the problem.

External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -345,7 +344,7 @@ DEF_OP(MulH) {
       sxtw(TMP1, GetReg<RA_64>(Op->Header.Args[0].ID()));
       sxtw(TMP2, GetReg<RA_64>(Op->Header.Args[1].ID()));
       mul(TMP1, TMP1, TMP2);
-      sbfx(GetReg<RA_64>(Node), TMP1, 32, 32);
+      ubfx(GetReg<RA_64>(Node), TMP1, 32, 32);
```
skmp, line 347: nice catch

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -262,7 +263,7 @@ namespace {
     Graph->Nodes.resize(NodeCount);
     Graph->VisitedNodePredecessors.clear();
     Graph->AllocData.reset();
-    Graph->AllocData.reset((FEXCore::IR::RegisterAllocationData*)malloc(FEXCore::IR::RegisterAllocationData::Size(NodeCount)));
+    Graph->AllocData.reset((FEXCore::IR::RegisterAllocationData*)FEXCore::Allocator::malloc(FEXCore::IR::RegisterAllocationData::Size(NodeCount * 4)));
```
skmp, line 266: `* 4`? Why?

Source/Tests/ELFCodeLoader2.h
```diff
@@ -111,7 +111,7 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
     if (Elf.ehdr.e_type == ET_DYN) {
       // needs base address
       auto TotalSize  = CalculateTotalElfSize(Elf.phdrs) + (BrkBase ? BRK_SIZE : 0);
-      LoadBase = (uintptr_t)Mapper(0, TotalSize, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0);
+      LoadBase = (uintptr_t)Mapper((void*)0x1'0000, TotalSize, PROT_NONE, MAP_ANONYMOUS | MAP_PRIVATE, 0, 0);
```
skmp, line 114: why is this needed?

Source/Tests/LinuxSyscalls/x32/Ioctl/steams.inl
```diff
@@ -0,0 +1 @@
+
```
skmp, line 1: steams? empty? erratic file?

skmp (approved):
Looks good overall, see comments

Source/Tests/LinuxSyscalls/x32/Ioctl/steams.inl
Sonicadvance1: Oops, erratic file.

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
Sonicadvance1: Oops, debugging left over.

Source/Tests/ELFCodeLoader2.h
Sonicadvance1: Was when I was still tinkering, oops.

PR #987, opened 2021-04-26
Changes LogMan assert to defines

Sonicadvance1:
C++ no-op functions can't optimize out the predicate arguments in all cases.
This was causing a problem where zero cost assertions weren't actually zero cost.

The only way to resolve this is to actually use macros sadly enough.

This will give a fairly hefty performance uplift with anything operating on IR.


skmp (approved):
Macros to the rescue

PR #990, opened 2021-04-27
Work around M1 Parallels hypervisor not showing CPU type

Sonicadvance1:
Parallels claims that the CPU is of implementor 0x41 and part number 0.
Easy enough to work around before real numbers get exposed

### DIFF
--- a/Scripts/aarch64_fit_native.py
+++ b/Scripts/aarch64_fit_native.py
@@ -2,6 +2,7 @@
 import re
 import sys
 import subprocess
+from distutils.version import LooseVersion, StrictVersion
 
 # Order this list from oldest to newest
 # try not to list something newer than our minimum compiler supported version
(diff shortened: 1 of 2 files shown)


Scripts/aarch64_fit_native.py
```diff
@@ -23,6 +23,8 @@
         tuple([0x51, 0x800]): "cortex-a73", # Kryo 2xx Gold
         tuple([0x51, 0x802]): "cortex-a75", # Kryo 3xx Gold
         tuple([0x51, 0x804]): "cortex-a76", # Kryo 4xx Gold
+        # Apple M1 Parallels hypervisor
+        tuple([0x41, 0x0]): "cortex-a76", # Placeholder CPU that is atleast ARMv8.2
```
Sonicadvance1, line 27: XXX: This should check the clang version used. If it is clang-12 or newer then `-mcpu=apple-a14` is available. Otherwise should choose A13 as the next best featureset

Sonicadvance1: Fixed

skmp (approved):

PR #996, opened 2021-04-29
AOTIR: Offline Generation

skmp:
## Overview
- Adds new --aotirgenerate / FEX_AOTIRGENERATE option that will generate aotir files for a given executable.
- Collects all modules with code in $AOTIR/<fileid>.path files, so we can generate them later on
- Renames aotir files to have .aotir postfix
- Includes a script that scans for .path files without matching .aotir file, and generates them

Depends on #976

## Using AOTIR Generation
- Install FEX
- Run an application via FEX to collect used libraries
- Run the FEXUpdateAOTIRCache.sh script
- Run the same application using `--aotirload` or FEX_AOTIRLOAD=1

## Details
- Converted Frontend Asserts to decode failures
- Converted OpDispatcher Asserts to OpDispatcher failures
- Frontend changes to collect ExternalBranches, Max decode address, w/ Context api
- AOTIRGenerate option

## Follow Ups
- Make .aotir files streamable so we don't have to buffer them in memory before writeout. They can be up to 10 gigs.
- Include parallel version of generation script / paralellize front end


External/FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -61,6 +61,7 @@ friend class FEXCore::IR::PassManager;
     return _LoadContext(Offset, Class, Size);
   }
   IRPair<IROp_StoreContext> _StoreContext(RegisterClassType Class, uint8_t Size, uint32_t Offset, OrderedNode *ssa0) {
+    assert(Offset != 4048);
```
Sonicadvance1, line 64: Magic number?

Sonicadvance1 (approved):

External/FEXCore/include/FEXCore/IR/IREmitter.h
skmp: Woops, that somehow made it past my self review. That's the value we get when register is invalid (=255), which sometimes happens from the frontend decoder. I'll remove the assert

PR #1009, opened 2021-05-04
Handle more cases of an application pinging self

Sonicadvance1:
AppImage uses this to inspect its own executable to ensure it is an AppImage file.


Sonicadvance1:
Before pushing need to handle nullptr pathname.

skmp (approved):

PR #1013, opened 2021-05-05
Handles symlinks in rootfs in openat 

Sonicadvance1:
This fixes an issue where rootfs has symlinks to other things in the rootfs so we need to track it through.

Relies on #1009 to be merged first.
Fixes any application that relies on libblas, mpv for example.


skmp:
Would be nice to have a more generic solution for all syscalls

PR #1016, opened 2021-05-06
Microoptimization for the DRM ioctls

Sonicadvance1:
For DRM applications have a three FD deep MRU cached for faster lookups of FD to DRM handlers.
In a completely DRM ioctl bound situation like es2gears or GL application without threaded context
Then this puts us /nearly/ at the performance of calling the ioctl32 handler directly.
Sadly there is overhead that can't be overcome so this is the best that can be done from userland

Also moves ioctl_32 handler to FD.cpp where it can be used directly from the handler in the future when we support using it directly.


Sonicadvance1:
oop, caught a crash that needs to be checked first.

Sonicadvance1:
Fixed. Was accidentally overwriting the final element in my LRU cache which is a helper loader.

skmp (approved):

PR #1025, opened 2021-05-12
Ignore non-canonical addresses in FS/GS setting

Sonicadvance1:


Source/Tests/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -306,10 +306,18 @@ namespace FEX::HLE {
       uint64_t Result{};
       switch (code) {
         case 0x1001: // ARCH_SET_GS
+          if (addr >= (1ULL << 48)) {
```
skmp, line 309: I assume this depends on the presented guest kernel config. Isn't this better off as a const?

skmp (approved):

Source/Tests/LinuxSyscalls/Syscalls/Thread.cpp
Sonicadvance1: Guest kernel config on x86-64 is always 48bit (or 52bit on Ice Lake-SP but there is some special sauce there).
Once we get an interface on ARM to legitimately query the VA size then it should be changed to check that.

PR #1027, opened 2021-05-12
ppoll fixes

Sonicadvance1:
glibc implementation makes a copy of the timeout and the kernel is expected to update it
Can't use the glibc implementation because of this.


skmp:
sigmask matches in this case? it matches in every case?

Sonicadvance1:
Yes. Sigmask matches in every case since x86 has always used a 64bit sigmask.
This is retained for both x86 and x86-64

skmp (approved):

PR #1041, opened 2021-06-02
DecodedOperand: Add helper functions for type testing 

lioncash:
Shortens the length of code necessary for testing the type of a decoded operand.

Also relocates the type field out of all the union types to have it in a central location.

### DIFF
--- a/External/FEXCore/include/FEXCore/Debug/X86Tables.h
+++ b/External/FEXCore/include/FEXCore/Debug/X86Tables.h
@@ -418,6 +418,9 @@ struct X86InstInfo {
     // We don't care if the opcode dispatcher differs
     return true;
   }
+  bool operator!=(const X86InstInfo &b) const {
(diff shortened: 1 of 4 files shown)


External/FEXCore/include/FEXCore/Debug/X86Tables.h
```diff
@@ -418,6 +436,9 @@ struct X86InstInfo {
     // We don't care if the opcode dispatcher differs
     return true;
   }
+  bool operator!=(const X86InstInfo *b) const {
```
Sonicadvance1, line 439: Looks like a good cleanup, just the buildbots don't like converting this pointer to a reference.

Sonicadvance1:
Looks like this still needs some more build testing :)

lioncash:
Amended

Sonicadvance1 (approved):
Awesome. thanks!

PR #1042, opened 2021-06-02
DecodedOperand: Convert operand type into an enum class

lioncash:
Now possible in a less messy way, since the type is now centralized in one location.

Makes it strongly typed and prevents any potential accidental implicit assignments to Type instead of something intended for the Data members.


Sonicadvance1 (approved):
Nice cleanup

PR #1043, opened 2021-06-04
General: Add missing const specifiers where applicable

lioncash:
A minor change that adds missing const specifiers to getters that don't modify the internal class state. Makes the API a little more explicit (and in some locations, a little more consistent) about non-modifying operations.


Sonicadvance1 (approved):
Nice!

PR #1044, opened 2021-06-04
Config: Move strings where applicable 

lioncash:
Noticed when adding amending missing const qualifiers on interfaces. We can make use of std::move here to avoid potential allocation churn a little.

While we're at it, we can implement EraseSet in terms of, well, Erase() and Set().


Sonicadvance1 (approved):
Ah nice catch on the silly GetIfExists. I believe this was necessary back when we had another function in the config namespace with the same name.

PR #1045, opened 2021-06-04
IR: Make use of defaulted operator==

lioncash:
Same behavior, but allows both operator== and operator!= to be automatically generated with a single declaration.


External/FEXCore/include/FEXCore/IR/IR.h
```diff
@@ -252,76 +251,73 @@ class OrderedNode final {
...
+  friend constexpr bool operator==(const CondClassType&, const CondClassType&) = default;
 };
 
 struct MemOffsetType final {
   uint8_t Val;
-  operator uint8_t() const {
+  constexpr operator uint8_t() const {
     return Val;
   }
-  int operator ==(const MemOffsetType other) const {
```
lioncash, line 281: I don't know if it was intentional to return int from these, and since there was no comment, I just assumed it was a mistake.

Sonicadvance1: I believe it was just a mistake.

Sonicadvance1 (approved):

PR #1050, opened 2021-06-05
Fixes semctl and msgctl

### LINKED ISSUE #744: 32bit msgctl broken on ARM
msgrcv and msgsnd might also be a part of this. Needs to be double checked.

### LINKED ISSUE #745: 32bit semctl is broken on ARM

Sonicadvance1:
There were some problems in both of these implementations.

Fixes #744
Fixes #745


skmp (approved):
👌

PR #1051, opened 2021-06-05
IRParser: Convert array of std::string over to array of std::string_view 

lioncash:
Same behavior, but allows the arrays to be constexpr (and use a little bit less space; 16 bytes vs 32 bytes per element*).

\* on clang, GCC, and MSVC (MSVC isn't supported, but included for completeness sake)

While in the area, I noticed some minor other locations that could have some cleanup/improvements (which I've separated into their own commits).


skmp (approved):
Nice!

Issue #146, opened 2020-04-24
Convert all string formatting to {fmt}


phire:
Rather than `printf` and `cout` everywhere, [{fmt}](https://fmt.dev/latest/index.html) is the future

It's an implementation of std::format that works in versions of c++ before c++20. By standardising on {fmt} now, we can easily switch to `std::format` later.

neobrain:
Is there much left to do here (cc @lioncash)?

lioncash:
I think for the most part, this can be considered resolved

PR #1052, opened 2021-06-05
LogManager: Add fmt-capable logging functions

lioncash:
Addresses #146 a little more by providing an interface to perform fmt-compatible logging, allowing migration over to it to begin.

No more, will people on the project be tormented by classic printf features like:

- Accidentally passing in a non-trivial type
- Goofy looking PRI macros
- Not being able to add support for custom types
- Mixing up signed/unsigned printf formatting specifiers accidentally

fmt-capable versions of the logging functions are named the same as the existing functions, just with a `Fmt` or `_FMT` suffix (depending on whether or not it's a function being used or a macro, respectively).


Sonicadvance1 (approved):
Niiiice. It begins.

PR #1053, opened 2021-06-05
CodeLoader: Add type aliases for mapper and unmapper functions

lioncash:
Adds two aliases to turn the signatures of MapMemory from long bois to short bois.

While we're at it, we can also give CodeLoader a virtual destructor to ensure destruction behavior is always consistent, no matter the scenario they're used in.


Sonicadvance1 (approved):
Nice. I kept seeing that then forgetting to clean it up.

PR #1054, opened 2021-06-05
GdbServer: Migrate logging/string handling to fmt where applicable

lioncash:
Simplifies a bunch of string manipulation code.

Much nicer to grok than stream formatting in many cases.


Sonicadvance1 (approved):
whew, that one is dense.

PR #1057, opened 2021-06-05
ConstProp: Separate out constituent chunks of const prop

lioncash:
The Run() function was quite large which makes it easy to get lost reading it, so this can be broken up into constituent functions that are then called in the Run() function, making it easy to see how the pass functions at a high level.

This is purely code movement. No functional changes are made.

This change was mainly made because I was seeing if I could add some basic additions to constant propagation, but wanted to move things into their own relevant chunks so that there were fewer moving parts to worry about.


Sonicadvance1 (approved):
Lots of good cleanups.
Let's hope perf characteristics haven't changed.

PR #1058, opened 2021-06-07
General: Place more compiler specifics into CompilerDefs.h

lioncash:
We can also make use of the `<bit>` header to remove some compiler builtins, since many of them were standardized.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
@@ -1867,7 +1867,7 @@ void InterpreterOps::InterpretIR(FEXCore::Core::InternalThreadState *Thread, uin
           case IR::OP_POPCOUNT: {
             auto Op = IROp->C<IR::IROp_Popcount>();
             uint64_t Src = *GetSrc<uint64_t*>(SSAData, Op->Header.Args[0]);
-            GD = __builtin_popcountl(Src);
+            GD = std::popcount(Src);
             break;
           }
           case IR::OP_FINDLSB: {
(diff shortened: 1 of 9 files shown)


External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -1943,7 +1944,7 @@ void InterpreterOps::InterpretIR(FEXCore::Core::InternalThreadState *Thread, uin
                 uint32_t Src = *GetSrc<uint8_t*>(SSAData, Op->Header.Args[0]);
                 Src <<= 24;
```
Sonicadvance1, line 1945: Since std::countl_zero supports 8-bit and 16 bit types. The shifts here can probably be removed?

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -1952,23 +1953,23 @@ void InterpreterOps::InterpretIR(FEXCore::Core::InternalThreadState *Thread, uin
                 uint32_t Src = *GetSrc<uint16_t*>(SSAData, Op->Header.Args[0]);
                 Src <<= 16;
```
Sonicadvance1, line 1954: Same here.

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
lioncash: Good catch. Amended. Since these have defined behaviors on zero, we can also collapse the if statements too.

Sonicadvance1 (approved):

PR #1067, opened 2021-06-10
Fixes 32-bit applications crashing

Sonicadvance1:
If you forced the 32-bit allocator debug option then 32-bit applications would work.
This is only meant for a debug and CI option though since some of our CI machines don't have a new enough kernel.

A major problem of this was that Steam was crashing during initialization.
Depending on what was happening it would either crash on some threads shutting down, or it would run out of memory since FEX was leaking allocations in to the 32-bit VA.

We now hook glibc more aggressively so it stops allocating TLS regions and who knows what else in the 32-bit VA space.
It only worked with the 32-bit allocator because we had allocated the full lower 32-bit VA space. Meaning glibc couldn't possibly allocate in that space.

With this and the other 32-bit crash PR, Steam is now consistently running again on my PC.
There's a bit of dirtiness in this around allocation lines, but we need a deeper investigation to determine where we are leaking state across when we change allocators.


Source/Tests/FEXLoader.cpp
```diff
@@ -70,15 +70,17 @@ void MsgHandler(LogMan::DebugLevels Level, char const *Message) {
   }
 
   if (!SilentLog) {
-    fprintf(OutputFD, "[%s] %s\n", CharLevel, Message);
-    fflush(OutputFD);
+    std::ostringstream Output;
+    Output << "[" << CharLevel << "] " << Message;
```
Sonicadvance1, line 74: Oops, need to put std::endl on the end of these.

phire (approved):

PR #1074, opened 2021-06-11
Syscalls/Signals: Add missing arguments to pidfd_send_signal syscall

lioncash:
These were accidentally omitted from the call.


Sonicadvance1 (approved):
Oops!

PR #1077, opened 2021-06-11
HostAllocator: Make Create64BitAllocator() return a unique_ptr 

lioncash:
Communicates the ownership intent in the interface.


Sonicadvance1:
`8247 - conformance-interfaces-sigtimedwait-1-1.test.int.posix (Failed)` Broke a posix test...and only on Interpreter?

Sonicadvance1:
Okay. It's a flake. That posix test sleeps for 1 second and the process needs to wake up within a 10ms margin to pass.
Just a flaky test...

Sonicadvance1 (approved):

PR #1079, opened 2021-06-12
Passes: Replace <iostream> header with narrower equivalents

lioncash:
`<iostream>` injects a static constructor in translation units that include it, even if its facilities aren't used. We can make use of `<sstream>` to avoid needing to execute those on startup.

While we're in the area, we can pass some logging strings as arguments to a format string instead of directly passing the output as a format string, so that the strings avoid being mangled (especially given we have things like `%ssa` in the strings which can unintentionally be interpreted as a null terminated string specifier).


Sonicadvance1 (approved):
haha oops.

PR #1091, opened 2021-06-17
Core: Remove unused DefaultFallbackCore

lioncash:
This doesn't seem to be hooked up to anything.


Sonicadvance1 (approved):
Oh right. This was passed off as a responsibility to the frontend. Which is why this wasn't used anymore.

PR #1101, opened 2021-06-20
Fixes fixed stack offset for 32-bit applications

### LINKED ISSUE #317: Guest thread fixed offset stack allocation needs fixed
Currently the initial guest thread's stack allocation is allocated at a fixed memory offset.
This should be fixed to be allocated at "any" location.
This would be within the lower 32bits for x86, and "anywhere" for x86-64.
There might need to be some special bits here for executable stack but that can be worked out.

Sonicadvance1:
Instead of forcing a fixed offset for the stack. Allow it to get placed automatically.
64-bit was already doing this; Now we can also do it in 32-bit.
This is possible because the mapper that is mapping the code will always map in the 32-bit space
for a 32-bit guest.

Fixes #317

### DIFF
--- a/Source/Tests/ELFCodeLoader.h
+++ b/Source/Tests/ELFCodeLoader.h
@@ -194,13 +194,7 @@ public:
 
   uint64_t GetStackPointer() override {
     uintptr_t StackPointer{};
-    if (File.GetMode() == ::ELFLoader::ELFContainer::MODE_64BIT) {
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Oop, need to look at this more. The Jetson is still really angry even though I though I had it working on my local machine.

skmp (approved):

PR #1125, opened 2021-06-25
FEXLoader: Make use of unique_ptr for syscall handlers

lioncash:
Makes the ownership requirements explicit in the interface.

Also makes it harder to unintentionally/accidentally leak memory.


Sonicadvance1:
I'll need to double check this one. This might hit a different bug in 32-bit application execution that CI isn't hitting.

Sonicadvance1:
Looks like we fixed the memory leak that was causing this to crash last time I tried.

Sonicadvance1 (approved):

PR #1130, opened 2021-06-27
x32: Fixes fcntl OP_GETLK64_32

Sonicadvance1:
This was overwriting the cmd argument and then being checked in the switch statement
after the call.
Since it was overwritten, it wasn't falling down the correct path, returning a flock_32 instead of a flock64_32


skmp (approved):
Good catch

PR #1135, opened 2021-06-30
Context: Place return value from GenerateIR into a struct

lioncash:
This makes it a little more straightforward to see what these return values mean without needing to look at the implementation, since a tuple's declaration is kind of opaque.

These tuples were also getting a little bit large enough to make the signature a little lengthy. Since the types are still aggregates, it's still possible to use structured bindings with them.


Sonicadvance1 (approved):
Oh thank goodness. These were a nightmare to look at when merged.

PR #1136, opened 2021-06-30
Context: Add alias for exit handler

lioncash:
Places the definition in one place so it doesn't need to be written several times.


Sonicadvance1 (approved):
Good cleanup

PR #1137, opened 2021-07-01
OpcodeDispatcher: Make use of if constexpr

lioncash:
Given these arguments are template arguments, we can use if constexpr with these branches to guarantee their elision.


Sonicadvance1 (approved):
Very reasonable.
Love that it can be forced with constexpr.

PR #1138, opened 2021-07-01
Arm64: Use regular function pointers with CAS handling functions

lioncash:
Given these are called in a loop repeatedly, where we know we'll always have a set function to call, std::function adds a little bit of
overhead.

e.g. for the uint32_t case:

<details>
<summary>Previous</summary>

```asm
        sub     sp, sp, #32
        stp     x29, x30, [sp, #16]
        add     x29, sp, #16
        ldr     x8, [x0, #16]
        stur    w1, [x29, #-4]
        str     w2, [sp, #8]
        cbz     x8, .LBB2_2
        ldr     x8, [x0, #24]
        sub     x1, x29, #4
        add     x2, sp, #8
        blr     x8
        ldp     x29, x30, [sp, #16]
        add     sp, sp, #32
        ret
.LBB2_2:
        bl      std::__throw_bad_function_call()
```
</details>

<details>
<summary>After</summary>

```asm
        mov     x3, x0
        mov     w0, w1
        mov     w1, w2
        br      x3
```
</details>

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
@@ -221,13 +221,18 @@ bool HandleCASPAL(void *_ucontext, void *_info, uint32_t Instr) {
   return false;
 }
 
+template <typename T>
+using CASExpectedFn = T (*)(T Src, T Expected);
+template <typename T>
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Nice. I have some additional changes to this that I'll be pushing today as well.

PR #1139, opened 2021-07-01
ARM64: Fixes bugs in unaligned atomic signal handlers

Sonicadvance1:
There were two bugs in here.

The first bug here is with with the CMPXCHG emulation.
1) If the *Expected* value did *NOT* match what was in memory
2) *AND* The *Desired* value matched the memory value
3) The CAS would incorrectly return success for this CMPXCHG
4) Thus setting ZF incorrectly

The second bug comes from atomic memory operations (Add, CLR, EOR, SET, SWAP).
This operation is a Load + &lt;Op&gt; + CAS
1) If the memory backing between the Load and CAS changes
2) The CAS would then fail
3) On Atomic memory operations this should then retry to ensure it completes successfully
4) We were not retrying on failure in this case
5) Thus something like `LOCK INC` would have never atomically incremented correctly
6) We can't use our ASM unit tests to test this, since it needs thread contention.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
@@ -157,6 +157,13 @@ bool HandleCASPAL(void *_ucontext, void *_info, uint32_t Instr) {
           mcontext->regs[ExpectedReg2] = FailedResult >> 32;
           return true;
         }
+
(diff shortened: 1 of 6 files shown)


phire (approved):
Really ugly. But I don't see a better option

Sonicadvance1:
I agree

skmp (commented):
urf

PR #1141, opened 2021-07-01
CPack: Update package name to remove conflict

Sonicadvance1:
'fex' is already taken by another projects. Use 'fex-emu' instead; Similar to other emulator package names.


skmp (approved):
:+1:

PR #1144, opened 2021-07-02
JIT: Add an option to disable SRA

skmp:
Adds --sra=false, only has effect on arm64 as x86 doesn't implement SRA


Sonicadvance1:
Supersceded by #1197

PR #1145, opened 2021-07-02
GdbStub: Fix memory leak in GdbServerLoop()

lioncash:
Previously the opened socket stream would be leaked.


Sonicadvance1 (approved):
The number of times we accidentally release instead of reset. oops!

PR #1147, opened 2021-07-04
OpDispatcher: Fixes imul flags calculations

Sonicadvance1:
We were calculating the high bits incorrectly in a couple variants


skmp (approved):
Nice catch!

PR #1155, opened 2021-07-08
Switch some memory allocations over to mmap

Sonicadvance1:
This cuts start up time from 8.962ms to 3.949ms and lets us use less physical memory.
We pay a small amortized cost from the faulting to populate pages which will be fairly low

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
@@ -12,6 +12,7 @@ $end_info$
 
 #include <iterator>
 #include <unordered_set>
+#include <sys/mman.h>
 
 #define SRA_DEBUG(...) // printf(__VA_ARGS__)
 
@@ -177,6 +178,7 @@ namespace {
   };
 
   static_assert(sizeof(RegisterNode) == 128 * 4);
+  constexpr size_t REGISTER_NODES_PER_PAGE = FEXCore::Core::PAGE_SIZE / sizeof(RegisterNode);
 
   struct RegisterSet {
     std::vector<RegisterClass> Classes;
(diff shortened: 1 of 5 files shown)


External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -254,13 +255,18 @@ namespace {
...
+    }
     delete Graph;
   }
 
   void ResetRegisterGraph(RegisterGraph *Graph, uint64_t NodeCount) {
     NodeCount = AlignUp(NodeCount, sizeof(uint64_t));
-    Graph->Nodes.clear();
-    Graph->Nodes.resize(NodeCount);
+    if (Graph->Nodes) {
+      ::munmap(Graph->Nodes, Graph->NodeCount * sizeof(RegisterNode));
```
phire, line 267: This is pretty hot code.

Wouldn't it be much faster to just memset the memory if the allocated space is already large enough?

Sonicadvance1: You're right. I thought the check for that case was elsewhere which is why I didn't handle it there.

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -254,15 +255,30 @@ namespace {
...
+    }
     delete Graph;
   }
 
   void ResetRegisterGraph(RegisterGraph *Graph, uint64_t NodeCount) {
     NodeCount = AlignUp(NodeCount, sizeof(uint64_t));
-    Graph->Nodes.clear();
-    Graph->Nodes.resize(NodeCount);
+    if (Graph->NodeCount < NodeCount) {
```
phire, line 266: Graph->NodeCount really should be aligned up to the number of nodes that can fit on a page. 

Might prevent a few unnecessary resizes during startup.

Sonicadvance1: Good idea. It now rounds up to how many nodes can fit in to a page.

phire (approved):

PR #1179, opened 2021-07-22
OpcodeDispatcher: Handle constant pooling directly

Sonicadvance1:
Instead of having an optimization pass handle block level constant pooling.
Let the opcodedispatcher do it directly.
This shuffles our handling from the optimization passes to the opcodedispatcher,
making us emit less terrible IR up front instead of being cleaned up later.

In most cases the lookup will only have 1-4 constants in the table, we typically use a very small number per block.
Additionally every conditional block will always have constant 0 and constant 1 in the lookup.

In a microbench I was testing, this improved total compilation time from 150ms down to 140ms, nearly a 7% reduction.
This is a tangible win by just emitting smarter IR.


Sonicadvance1:
Oop, this introduces crashes on AArch64.

Sonicadvance1:
This isn't a clear win with our current implicit sized ops now.

PR #1180, opened 2021-07-23
 Linux: Implements support for signals 32 and 33 

Sonicadvance1:
When a guest tries to use the setxid syscalls, the guest glibc has
a mechanism in place to ensure that the process wide setxid is handled.
The mechanism is fairly complex but it uses signal 33 and sends the signal to all
active threads to ensure every thread sets the correct state here.

We need to intercept this and pass the context information correctly to the guest instead.
Otherwise FEX just crashes when the glibc HOST handler tries handling the guest applications signal.

Fixes the game SOMA https://store.steampowered.com/app/282140/SOMA/
Weirdly this game uses some Autodesk SDK `FBX SDK 2012` which part of its thread initialization code tries to do a `setuid(0)`

### DIFF
--- a/Source/Tests/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tests/LinuxSyscalls/SignalDelegator.cpp
@@ -23,6 +23,15 @@ $end_info$
 #include <unistd.h>
 
 namespace FEX::HLE {
+#ifdef _M_X86_64
+  __attribute__((naked))
+  static void sigrestore() {
+    __asm volatile("syscall;"
+        :: "a" (0xF)
+        : "memory");
+  }
+#endif
+
   constexpr static uint32_t SS_AUTODISARM = (1U << 31);
   constexpr static uint32_t X86_MINSIGSTKSZ  = 0x2000U;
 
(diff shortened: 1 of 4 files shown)


Source/Tests/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -222,38 +231,42 @@ namespace FEX::HLE {
...
-        sigdelset(&SignalHandler.HostAction.sa_mask, i);
+        SignalHandler.HostAction.sa_mask &= ~(1ULL << (i - 1));
       }
       else if (SigIsMember(&SignalHandler.GuestAction.sa_mask, i)) {
-        sigaddset(&SignalHandler.HostAction.sa_mask, i);
+        SignalHandler.HostAction.sa_mask |= (1ULL << (i - 1));
       }
     }
 
+    LogMan::Msg::D("Installing host handler for signal: %d", Signal);
```
Sonicadvance1, line 264: Oop. Need to delete this.

Source/Tests/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -590,21 +600,19 @@ namespace FEX::HLE {
...
         }
 
         if (ThreadData.CurrentSignalMask.Val & (1ULL << i)) {
-          sigaddset(&HostSet, i + 1);
+          HostSet |= (1ULL << 1);
         }
       }
 
-      pthread_sigmask(SIG_SETMASK, &HostSet, nullptr);
+      ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &HostSet, nullptr);
```
Sonicadvance1, line 615: oop, this needs a final argument of sigsetsize and to handle 32-bit sigprocmask sigsetsize.

PR #1183, opened 2021-07-23
FlexBitSet: Use non-template type parameter for index parameters

lioncash:
These parameters are used as indexes into the tracked memory, so sizing the index variable relative to the type being stored is kind of sketchy.

e.g. If the tracked types were uint8_t for example, we'd still want the indexing parameters to be regularly sized so that we aren't implicitly truncating values all the time when passing values to a uint8_t parameter (and while all usages are currently using uint64_t as type T, we may as well address this).

While we're at it, we can make both Get() and operator[] const member functions, since they don't directly modify any underlying data, they only read it.


Sonicadvance1 (approved):
Oops. good catch.

PR #1240, opened 2021-08-31
Move BucketList into it's own file

phire:
I'm doing this to make RegisterAllocationPass.cpp slightly smaller and easier to understand


Sonicadvance1 (approved):
Thanks!
This was a pain point for me as well, but I wasn't in the area to clean it up.

PR #1243, opened 2021-08-31
RA: Add max NoteCount assert

phire:
The chance of hitting this is near zero, but still safer to have an assert.


Sonicadvance1 (approved):
Nice

PR #1244, opened 2021-09-01
Properly implement single CAS on ARMv8.0

CallumDev:
This implements unaligned CAS through the SIGBUS handler for ARMv8.0. It also rewrites the `OP_CAS` in the Interpreter to generate the pattern that can be detected by SIGBUS.

Fixes 0F_B0* tests for ARMv8.0.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
@@ -1123,20 +1123,9 @@ uint64_t DoCAS64(
   }
 }
 
-bool HandleCASAL(void *_ucontext, void *_info, uint32_t Instr) {
+static bool RunCASAL(void *_ucontext, void *_info, uint32_t Size, uint32_t DesiredReg, uint32_t ExpectedReg, uint32_t AddressReg) {
   mcontext_t* mcontext = &reinterpret_cast<ucontext_t*>(_ucontext)->uc_mcontext;
   siginfo_t* info = reinterpret_cast<siginfo_t*>(_info);
-
-  if (info->si_code != BUS_ADRALN) {
-    // This only handles alignment problems
-    return false;
-  }
-
-  uint32_t Size = 1 << (Instr >> 30);
-
-  uint32_t DesiredReg = Instr & 0b11111;
-  uint32_t ExpectedReg = (Instr >> 16) & 0b11111;
-  uint32_t AddressReg = (Instr >> 5) & 0b11111;
 
   uint64_t Addr = mcontext->regs[AddressReg];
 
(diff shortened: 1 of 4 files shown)


External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -94,6 +94,62 @@ static uint64_t AtomicFetchNeg(uint64_t *Addr) {
 
   return Expected;
 }
+
+static uint8_t AtomicCompareAndSwap8(uint8_t expected, uint8_t desired, uint8_t *addr)
```
neobrain, line 98: Looks like this could be turned into a template so we don't need to duplicate the same code three times here.

For the ARM implementations, we can use template specialization, i.e.
```cpp
template<typename T>
static T AtomicCompareAndSwap(T expected, T desired, T *addr);

template<>
uint16_t AtomicCompareAndSwap(uint16_t expected, uint16_t desired, uint16_t* addr) {
    // Impl goes here   
}
```

CallumDev: Replaced x86_64 impl with a template. ARM64 uses template specialization.

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
```diff
@@ -1559,6 +1565,53 @@ bool HandleAtomicLoad128(void *_ucontext, void *_info, uint32_t Instr) {
   return true;
 }
 
+static uint64_t HandleCAS_NoAtomics(void *_ucontext, void *_info)
+{
+  mcontext_t* mcontext = &reinterpret_cast<ucontext_t*>(_ucontext)->uc_mcontext;
+  siginfo_t* info = reinterpret_cast<siginfo_t*>(_info);
```
Sonicadvance1, line 1571: Please add a check here just like `HandleCASAL` to ensure the si_code is `BUS_ADRALN` otherwise we could get spurious checks on invalid memory accesses

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -188,6 +202,141 @@ uint64_t AtomicFetchNeg(uint64_t *Addr) {
   return Result;
 }
 
+template<typename T>
+static T AtomicCompareAndSwap(T expected, T desired, T *addr);
+
+template<>
+uint8_t AtomicCompareAndSwap(uint8_t expected, uint8_t desired, uint8_t *addr) {
```
Sonicadvance1, line 209: Technically this drops ARMv8.1+ builds to using ARMv8.0 operations, but since it is the interpreter it isn't a big deal.
I'll let this go for now and create an issue to make it more optimal in the future, when we improve the dispatch logic of the interpreter.

Sonicadvance1 (commented):
Other than that one code change request. It looks great!
Thanks!

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64.cpp
CallumDev: That would duplicate the check made in the only caller of this function `HandleAtomicLoadstoreExclusive`, is that alright?

Sonicadvance1: Oh no, you're right. I forgot that the caller also did the check. Thanks!

PR #1257, opened 2021-09-05
Fix unaligned CASPair on ARMv8.0

CallumDev:
`SIGBUS` handler can now recognise the pattern generated by the JIT for `CASPair` on ARMv8.0.

Fixes the following tests:
* `Test_Secondary/09_XX_01_7.asm`
* `Test_Secondary/09_XX_01_8.asm`
* `Test_Secondary/09_XX_01_9.asm`
* `Test_Secondary/09_XX_01_11.asm`
* `Test_Secondary/09_XX_01_12.asm`
* `Test_Secondary/09_XX_01_13.asm`

NOTE: The interpreter tests for these never failed in the first place on my Cortex-A72, not sure what code clang/GCC emit for `std::atomic` that makes this the case.


Sonicadvance1 (approved):
Very nice!
Thanks for doing this. Interpreter support that is failing for whatever reason can be wired up later.
Expectation is that everyone will be running JIT anyway.

PR #1262, opened 2021-09-09
RA validation

phire:
This is a validation pass that attempts to prove the output from the RA pass is valid.

The current design should be able to prove that the RA result is internally consistent. That no-matter what control flow path you take thought the control flow graph, the physical registers and spill slots will always contain a single possible SSA value. 

It also checks that the SSA values in the IR actually line up with the SSA value in the physical register. 

~~It's one major blind spot is that it can't prove the spill slot contains the correct SSA.~~ Fixed by modifying FIllRegister to preserve a reference to the original SSA


Sonicadvance1 (approved):
Haven't looked too closely at the logic but it seems reasonable and it doesn't seem to horribly impact CI time.

Sonicadvance1 (approved):
Nice. Might be nice to add more comments about the logic being done so future excavations can more quickly parse what is happening.

PR #1277, opened 2021-09-13
AOTIR: fix use after free

phire:
These variables are owned by other parts of the code, and AOTIR should not be freeing/deleting them.


Sonicadvance1 (approved):
Oh, I forgot about this. I noticed it like a month ago, oops :D

PR #1288, opened 2021-10-01
Scripts: Adds a new script for extracting function definitions

Sonicadvance1:
This is very useful for extracting function definitions for thunks.
Keep it in upstream to not get lost.

Sometimes it can munge a definition but it is usually fine.


Sonicadvance1:
This will need to be changed and updated once the new thunk generation is wired up.

PR #1295, opened 2021-10-01
Implement thunk library generation using libclang

### LINKED ISSUE #1278: Automate thunk library generation using libclang
**Draft** sketching a possible design for a libclang-based thunk library generator.

# Terminology

* host library: Native host library installed globally, e.g. /usr/lib/aarch64-linux-gnu/libX11.so
* guest library: Native guest library installed in the rootfs, e.g. $ROOTFS/usr/lib/x86_64-linux-gnu/libX11.so
* guest thunk library: The shim library shipped with FEX and loaded in place of the actual guest library
* host thunk library: The shim library shipped with FEX and exporting symbols linked to guest thunk functions
* guest/host thunk functions: Shim functions exported by the guest/host thunk library
* guest/host library functions: Functions exported by the guest/host library

# Syntax

[Example generator input](https://gist.github.com/neobrain/67f3dc9f14d3c8a8e88bae808242dcc2)

Thunked functions are defined by defining a struct called `fex_gen_config` and specializing it for each function that should be exported from the guest/host thunk libraries. The code emitted by the thunk generator can be customized by defining certain members to the specialized structs, as outlined below.

```cpp
template<auto>
struct fex_gen_config {
  // Global per-module (or per-namespace) settings
};

// For each exported function:
template<> struct fex_gen_config<XCreateIC> {
  // Per-function config, see below
};
```

C++ namespaces can be used to group functions together with different settings specified in the non-specialized `fex_gen_config`.

### Variadic functions

If the member alias `uniform_va_type` of type `arg_type` is defined, a variadic guest-function of the form `ret func(other_params, ...)` will be linked to a host-function `ret func(other_params, arg_type* args, size_t num_args)`. As indicated by the name "uniform", this can only be used when the implementation treats all parameters as the same type.

```cpp
template<> struct fex_gen_config<XCreateIC> {
    using uniform_va_type = unsigned long;
};
```

### Function pointer parameters ("callbacks")

Guest function pointers can't be used as callbacks for host libraries, so it's required to specify a strategy on how to deal with this:

* If `fex_gen_config` inherits from `callback_through_user_data<cb_idx, data_idx>`, the guest-provided callback (parameter index `cb_idx`) is replaced by a fixed host-function and the data pointer (parameter index `data_idx`) is replaced by a custom struct containing original data pointer and the original guest callback. This wrapped data will be heap-allocated and must be deallocated explicitly.

* If `fex_gen_config` inherits from `callback_through_user_data_on_stack<cb_idx, data_idx>`, this will have the same effect as inheriting from `callback_through_user_data`, but the internal user data struct is placed on the stack so that memory doesn't need to be deallocated explicitly. This only works if the native host library doesn't use the callback after it returns from the thunked function.

* If `fex_gen_config` inherits from `callback_stub<cb_idx>`, the guest-provided callback is replaced by a fixed host-function that will trigger a runtime error when being called.

* If `fex_gen_config` inherits from `callback_guest<cb_idx>`, the guest-provided callback is assumed to never be called on the host-side. This is ensured by replacing the corresponding parameter in the host library function with an uncallable opaque handle type that can only be passed back to the guest thunk library.

### Factory functions that return objects with vtables/function pointers

Vtables and function pointers initialized by host functions (such as when constructing objects) cannot be called by the guest. If `fex_gen_config` inherits from `fixup_vtable_in_return`, FEX will generate additional code to fixup returned function pointers with function pointers that are both host-callable and guest-callable. This is done recursively for any (possibly nested) function pointer members in the returned object.

```cpp
template<> struct fex_gen_config<XCreateImage> : fex_gen::fixup_vtable_in_return {
};
```

If `fex_gen_config` inherits from `returns_guest_pointer`, FEX will *not* generate this magic and instead trust that the host endpoint of the thunk will return a pointer that can safely be used by the guest (and will never be used by the host itself).

- [ ] TODO: If the function returns an object with vtables but this customization is not specified, should the generator fail loudly or should it assume the function did not freshly create the object?

- [ ] TODO: Should a customization point be added for the returned function pointer?

### Overriding host thunk function

Host thunk functions usually just forward their arguments to host library functions. If `fex_gen_config` inherits from `custom_host_impl`, the thunk unpacker will call `fexfn_impl_FUNC` instead of the host library function. (The latter is still available through `fex_ldr` though, hence it can be used in `fexfn_impl_FUNC` if needed.)

```cpp
template<> struct fex_gen_config<XFree> : fex_gen::custom_host_impl {
};
```

### Overriding guest thunk function

Guest thunk functions usually just pack their arguments into a struct that is passed to `fex_thunks_<lib>_<function>`. If `fex_gen_config` inherits from `custom_guest_impl`, the generator will expect the guest entrypoint to be defined manually such that functions can be customized more freely. Packing the arguments and invoking the thunk can be done explicitly by calling `fexfn_pack_<function>`.

```cpp
template<> struct fex_gen_config<XFree> : fex_gen::custom_guest_impl {
};
```

### Adding internal parameters

Host thunk functions usually have the same signature as the guest thunk function they're linked to. If a member type alias`internal_args`, its type will be appended to the host thunk function.

```cpp
template<> struct fex_gen_config<XInitThreads> {
    using internal_args = XUnlockMutex_fn_type;
};
```

- [ ] TODO: How to specify multiple parameters? Is `std::tuple` suitable or do we need a custom type list?

### Disabling thunking locally

Some functions (such as printf-likes) don't need thunking per se and may be simpler to implement without. If `fex_gen_config` inherits from `guest_only`, no thunks or (un-)packing functions will be generated. The guest-side entrypoint must be defined manually

```cpp
template<> struct fex_gen_config<XFree> : fex_gen::guest_only {
};
```

### Automatic function discovery

Instead of manually listing functions to be exported, the generator can automatically treat all functions declared in a library's header as if they had been defined with an empty `fex_gen_config`. Functions that cannot be processed without explicit annotation will trigger an error unless an explicit `fex_gen_config` is defined for them.

- [ ] TODO: What syntax to use for this? How can we appropriately constrain the set to avoid invalid exports?

### Custom host function loader
Usually, the host thunk library will load any required symbols from the native host library using dlsym. Some libraries (such as libGL) requires thunking functions that aren't exported by the library directly but exposed through a dedicated lookup function (e.g. glXGetProcAddress).

```cpp
template<auto>
struct fex_gen_config {
  // Must refer to a function defined manually
  const char* loader = "InternalGetProcAddress";
};

template<>
struct fex_gen_config<glBegin> {
};
```

### Guest-side symbol table

All host-side thunk libraries define a table called `export` listing all thunked functions. Some libraries (notably libGL) require a similar guest-side table to be created to map each function name (as `const char*`) to the corresponding packing function. This can be requested by making the unspecialized `fex_gen_config` inherit from `guest_symbol_table`. The generated symbol table will be named after the containing namespace:

```cpp
namespace internal_functions {

template<auto>
struct fex_gen_config : guest_symbol_table {
  // Will cause a guest-side symbol table called "internal_functions_symtable" to be generated
};

} // end of namespace
```

# Update history

- 17 Sep 2021: Binary function customization parameters are now implemented using inheritance from a tag type. Previously, a boolean member had to be declared.
- 17 Sep 2021: Type-based function customization parameters are now implemented using `using customization_name = type;`. Previously, a member had to be declared where its type would act as the actual parameter.
- 18 Sep 2021: `custom_host_unpack` renamed to `custom_host_impl`. Instead of replacing the `fexfn_unpack_func`, the unpacker is now autogenerated as usual but will call a user-provided `fexfn_impl_func` function instead of the host library function (`fex_ldr_func`).
- 2 Oct 2021: Added tags for callback parameters (`callback_through_user_data` and `callback_stub`)
- 5 Oct 2021: Added tags for guest-side symbol tables
- 6 Oct 2021: Added `returns_guest_pointer` tag
- 8 Oct 2021: Added `guest_only` tag for guest-only functions, which don't need thunks
- 8 Oct 2021: Added `custom_guest_impl` tag
- 30 Oct 2021: Added `callback_guest` tag to callbacks that are only called guest-side

neobrain:
# Overview

This PR adds a libclang-based thunk library generator written in C++ to replace the current python-based generator. This automates most of the boilerplate needed in the library templates and allows us to implement more complex generation patterns (e.g. for functions that construct objects with vtables) that wouldn't be feasible to implement with the python bindings for libclang.

# New generator syntax

Each thunk library now has a dedicated `_interface.cpp` file listing all functions of the library API. Some functions require annotation as fully described in #1278.

# Test architecture

The new generator is unit-tested using the Catch2 testing framework: Tests run the generator output itself through libclang, the [AST matchers API](https://clang.llvm.org/docs/LibASTMatchersReference.html) of which we use to e.g. verify functions are defined with the correct signatures.

# Migration plan

This PR replaces legacy generator templates with new-style interface files, for libraries currently built by default **only**.

Other libraries (such as `libSDL2` or `libfex_malloc`) that ship with FEX without build instructions will need to be migrated once they are enabled. Since their definition files are in the git history, this PR removes them without replacement. 

Struct verifier functionality has been lost on libvulkan. I'll need to look into whether this can be integrated into the new generator easily, possibly in a separate PR at a later point.

# TODO

- [x] Implement bare minimum functionality to make all existing thunk libraries use the new generator:
  - [x] Variadic functions
  - [x] Function pointer parameters ("callbacks")
  - [x] Custom host end points
  - [x] Custom guest end points
  - [x] Guest-side symbol tables (needed for libGL's `glXGetProcAddr`)
  - [x] Custom host function loader (needed for libGL functions loaded through `glXGetProcAddr` instead of `dlsym`)
  - [x] Library versions
- [x] Use new generator for existing thunk libraries:
  - [x] libXfixes, libXext, libXrender, libasound
  - [x] libX11
  - [x] libxshmfence
  - [x] libGL, libEGL
  - [x] libxcb, libxcb-*
  - [x] libdrm
  - [x] libvulkan
- [x] ~~Recover struct verifier functionality for libvulkan~~ (will be done in a later PR)
- [x] ~~Fully automate what's currently in the `libY_Guest/Host.cpp` files:~~ (will be done in a later PR)
  - [x] ~~Callback boilerplate~~
  - [x] ~~Export tables~~
- [x] Clean up code
  - [x] Implement the syntax changes proposed in #1278
  - [x] Fix warnings emitted during test runs
  - [x] Add CTest integration
    - [x] Use suffix-notation for test names
    - [x] Add custom target for CI integration
    - [x] Enable CI integration
  - [x] Remove old thunk generator
  - [x] Update `ThunkLibs/README.md`
  - [x] Remove `Vulkan-Docs` submodule

Fixes #1278


ThunkLibs/libX11/libX11_interface.cpp
```diff
@@ -0,0 +1,526 @@
...
+#include <X11/Xlib.h>
+#include <X11/Xutil.h>
+#include <X11/Xresource.h>
+
+#include <X11/Xlibint.h>
+#include <X11/XKBlib.h>
+
+template<auto>
+struct fex_gen_config {
+    unsigned version = 1;
```
Sonicadvance1, line 12: I believe this should actually be 6. Typo in the current version of the python scripts.

Sonicadvance1:
I still have a few files to review but I don't currently see anything egregious so far.

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,82 @@ static void DoSetupWithInstance(VkInstance instance) {
...
 }
 
-static void UNPACKFUNC(vkGetPhysicalDeviceSurfaceSupportKHR)(void *argsv){
-  struct arg_t {VkPhysicalDevice a_0;uint32_t a_1;VkSurfaceKHR a_2;VkBool32* a_3;VkResult rv;};
-  auto args = (arg_t*)argsv;
-  args->rv = LDR_PTR(vkGetPhysicalDeviceSurfaceSupportKHR)
-    (args->a_0,args->a_1,args->a_2,args->a_3);
-}
+
+// TODO: vkGetPhysicalDeviceSurfaceSupportKHR has custom host end point, but not sure why?
```
Sonicadvance1, line 122: This can be removed now. Was just leftover debugging.

Sonicadvance1 (dismissed):
Still some debug logs left over it looks like. But overall looks like it is a good improvement.

ThunkLibs/libX11/libX11_interface.cpp
neobrain: Fixed as a dedicated commit before migrating.

neobrain:
Should be good to go :)

neobrain:
Fixed all showstopping TODOs and updated the generator documentation.

Should be good to go :)

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,80 @@ static void DoSetupWithInstance(VkInstance instance) {
...
 
-static void UNPACKFUNC(vkGetDeviceProcAddr)(void *argsv){
-  struct arg_t {VkDevice a_0;const char* a_1;PFN_vkVoidFunction rv;};
-  auto args = (arg_t*)argsv;
+#define FEXFN_IMPL3(a, b, c) a##_##b##_##c
+#define FEXFN_IMPL2(a, b) FEXFN_IMPL3(fexfn_impl, a, b)
+#define FEXFN_IMPL(fn) FEXFN_IMPL2(LIBLIB_NAME, fn)
 
+static PFN_vkVoidFunction FEXFN_IMPL(vkGetDeviceProcAddr)(VkDevice a_0, const char* a_1) {
+  printf("TRACE HOST enter: %s %p %s\n", __FUNCTION__, a_0, a_1);
```
Sonicadvance1, line 75: Should get removed

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,80 @@ static void DoSetupWithInstance(VkInstance instance) {
...
+    DoSetupWithDevice(a_0);
   }
 
   // Just return the host facing function pointer
   // The guest will handle mapping if this exists
-  args->rv =
-    LDR_PTR(vkGetDeviceProcAddr)
-    (args->a_0,args->a_1);
+  auto ret = LDR_PTR(vkGetDeviceProcAddr)(a_0, a_1);
+  printf("TRACE HOST exit: %s %p %s\n", __FUNCTION__, a_0, a_1);
```
Sonicadvance1, line 83: Should get removed

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,80 @@ static void DoSetupWithInstance(VkInstance instance) {
...
+  return ret;
 }
 
-static void UNPACKFUNC(vkGetInstanceProcAddr)(void *argsv){
-  struct arg_t {VkInstance a_0;const char* a_1;PFN_vkVoidFunction rv;};
-  auto args = (arg_t*)argsv;
-
+static PFN_vkVoidFunction FEXFN_IMPL(vkGetInstanceProcAddr)(VkInstance a_0, const char* a_1) {
+  printf("TRACE HOST enter: %s %p %s\n", __FUNCTION__, a_0, a_1);
```
Sonicadvance1, line 88: Should get removed

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,80 @@ static void DoSetupWithInstance(VkInstance instance) {
...
+    DoSetupWithInstance(a_0);
   }
 
   // Just return the host facing function pointer
   // The guest will handle mapping if it exists
-  args->rv =
-    LDR_PTR(vkGetInstanceProcAddr)
-    (args->a_0,args->a_1);
+  auto ret = LDR_PTR(vkGetInstanceProcAddr)(a_0, a_1);
+  printf("TRACE HOST exit: %s %p %s\n", __FUNCTION__, a_0, a_1);
```
Sonicadvance1, line 96: Should get removed

ThunkLibs/libvulkan_device/Host.cpp
```diff
@@ -66,105 +67,80 @@ static void DoSetupWithInstance(VkInstance instance) {
...
 }
 
-static void UNPACKFUNC(vkCreateInstance)(void *argsv){
-  struct arg_t {const VkInstanceCreateInfo* a_0;const VkAllocationCallbacks* a_1;VkInstance* a_2;VkResult rv;};
-  auto args = (arg_t*)argsv;
-  args->rv =
-    LDR_PTR(vkCreateInstance)
-    (args->a_0, nullptr,args->a_2);
+static VkResult FEXFN_IMPL(vkCreateDevice)(VkPhysicalDevice a_0, const VkDeviceCreateInfo* a_1, const VkAllocationCallbacks* a_2, VkDevice* a_3){
+  printf("TRACE HOST: %s\n", __FUNCTION__);
```
Sonicadvance1, line 109: Should get removed

Sonicadvance1:
```
        vkGetInstanceProcAddr: Couldn't find Guest symbol: 'vkCreateWaylandSurfaceKHR'
Illegal instruction (core dumped)
```
When attempting to run this. I can only assume `libvulkan_device_interface.cpp` doesn't have this definition where previously this definition was autogenerated from the Vulkan-Docs repo.
This immediately raises a concern about updating the thunk being more involved than a repo update and seeing what has changed.

ThunkLibs/libvulkan_device/Host.cpp
neobrain: Good catch, cheers

neobrain:
> When attempting to run this. I can only assume libvulkan_device_interface.cpp doesn't have this definition where previously this definition was autogenerated from the Vulkan-Docs repo.
> This immediately raises a concern about updating the thunk being more involved than a repo update and seeing what has changed.

Fixed; the interface file is complete for all enabled platforms AFAICT, so this isn't a design gotcha.

The issue here was that the `VK_USE_PLATFORM_WAYLAND_KHR` define was missing from the interface file, so my function extractor didn't catch these platform-specific functions before.

Are there any other platforms we should add? Currently, we enable 
```cpp
#define VK_USE_PLATFORM_XLIB_XRANDR_EXT
#define VK_USE_PLATFORM_XLIB_KHR
#define VK_USE_PLATFORM_XCB_KHR
#define VK_USE_PLATFORM_WAYLAND_KHR
``` 
whereas the full list is
```
VK_USE_PLATFORM_ANDROID_KHR
VK_USE_PLATFORM_FUCHSIA
VK_USE_PLATFORM_IOS_MVK
VK_USE_PLATFORM_MACOS_MVK
VK_USE_PLATFORM_METAL_EXT
VK_USE_PLATFORM_VI_NN
VK_USE_PLATFORM_WAYLAND_KHR
VK_USE_PLATFORM_WIN32_KHR
VK_USE_PLATFORM_XCB_KHR
VK_USE_PLATFORM_XLIB_KHR
VK_USE_PLATFORM_DIRECTFB_EXT
VK_USE_PLATFORM_XLIB_XRANDR_EXT
VK_USE_PLATFORM_GGP
```

Other than perhaps `VK_USE_PLATFORM_ANDROID_KHR`, I don't think we care about any of the missing ones.

ThunkLibs/libvulkan_device/Guest.cpp
```diff
@@ -1,56 +1,49 @@
...
     // Initialize unordered_map from generated initializer-list
     PtrsToLookUp = {
-#define PAIR(name, ptr) { #name, (PFN_vkVoidFunction*)ptr }
-#include "function_pack_pair.inl"
+#define PAIR(name, unused) { #name, (PFN_vkVoidFunction*)name },
+FOREACH_SYMBOL(PAIR)
 #undef PAIR
     };
 
+#define PAIR(name, unused) printf("SETUP: %s\n", #name);
```
Sonicadvance1, line 33: Debug message.

Sonicadvance1 (approved):
Aside from the single debug message. Looks good.
Just needs that removed and rebased.

ThunkLibs/libvulkan_device/Guest.cpp
neobrain: Fixed, and made another pass over the changes to make sure there's no other random printfs left.

neobrain:
Fixed leftover printf and rebased on latest main.

PR #1337, opened 2021-11-02
OpcodeDispatcher: Handle BMI1 BEXTR

lioncash:
Follows up the PR handling BMI1 ANDN (#1325), this adds handling for BMI1's BEXTR instruction

After this, all that remains is to add handling for BLSI/BLSMSK/BLSR. These are similar enough that I will likely include all of them in the following PR after this one.


Sonicadvance1 (approved):
Nice. A bit sad that ARM doesn't have a comparable bitfield extract instruction that operates on a register source rather than constants.

PR #1338, opened 2021-11-02
IR: Add handling for ANDN operations 

lioncash:
In a few parts of the OpcodeDecoder there's comments about making use of BIC. Given that, we can introduce the Andn IR operation that performs the same behavior as it.

I've named it `Andn`, since I feel it's more straightforward to read at a glance, but if it's preferable to name it `Bic`, I'm not opposed to changing it.

First time adding an IR op, so if anything looks wonky, just let me know and I'll fix it.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/ALUOps.cpp
@@ -15,6 +15,11 @@ $end_info$
 #include <xbyak/xbyak.h>
 
 namespace FEXCore::CPU {
+
+#define GRS(Node) (IROp->Size <= 4 ? GetSrc<RA_32>(Node) : GetSrc<RA_64>(Node))
+#define GRD(Node) (IROp->Size <= 4 ? GetDst<RA_32>(Node) : GetDst<RA_64>(Node))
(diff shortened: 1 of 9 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/ALUOps.cpp
```diff
@@ -417,6 +422,23 @@ DEF_OP(And) {
   mov(Dst, rax);
 }
 
+DEF_OP(Andn) {
+  auto Op = IROp->C<IR::IROp_Andn>();
+  const auto& Lhs = Op->Header.Args[0];
+  const auto& Rhs = Op->Header.Args[1];
+  auto Dst = GRD(Node);
+
```
Sonicadvance1, line 430: Looks like the CI failure here is likely running in to the case where the destination is also one of the sources.

lioncash: Ah, true. Adjusted it to take that into account

Sonicadvance1 (approved):

Sonicadvance1:
Both BIC(Binary Clear) and ANDN make sense here. I don't have a preference of one name versus the other.

PR #1339, opened 2021-11-02
Core: Mark relevant Interpreter/JIT functions as [[nodiscard]]

lioncash:
Lets the compiler warn loudly when the result from any of these functions are left unused (indicating a bug).

This is primarily useful for the GetSrc/GetDst-related functions when making new IR op implementations, but the other cases are relatively useful for catching silent bugs introduced during refactors.


Sonicadvance1 (approved):
Always nice to have the compiler yell at us more.

PR #1344, opened 2021-11-04
OpcodeDispatcher: Implement handling for BLSI

lioncash:
PRing this separately from the other two, since this makes changes to the flag type in X86Tables to accommodate another flag


Sonicadvance1 (approved):
Nice!

PR #1353, opened 2021-11-07
OpcodeDispatcher: Implement BLSR/BLSMSK

lioncash:
Implements the remaining two instructions to complete BMI1 support.

Now that everything is implemented we can also signify BMI1 support in our CPUID code.


Sonicadvance1 (approved):
Woo!

Issue #1355, opened 2021-11-07
Support ADX


Sonicadvance1:
Easy enough to support as it is only two instructions.
Supported on all x86-64 major processors now.
https://en.wikipedia.org/wiki/Intel_ADX

PR #1356, opened 2021-11-10
OpcodeDecoder: Add support for ADX

lioncash:
Another CPU extension we can cross off the list.

Addresses #1355


Sonicadvance1 (approved):
Nice!

PR #1358, opened 2021-11-12
Dispatcher: Partial support for RIP adjust in signal handler

Sonicadvance1:
When the context structure is adjusted in the signal handler then this
updates the state of the CPU on sigreturn.

We check to see if the RIP was adjusted and in this case we will update
the JIT in a potentially unsafe fashion.
This works well enough with signal handlers that effectively just long
jump and not much else.

Fixes wine initial prefix setup where rundl32 expects to do a try-catch
fault capture but infinite loops without this.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
@@ -13,6 +13,11 @@
 
 namespace FEXCore::ArchHelpers::Context {
 
+enum ContextFlags : uint32_t {
+  CONTEXT_FLAG_INJIT = (1U << 0),
+  CONTEXT_FLAG_32BIT = (1U << 1),
+};
+
 struct X86ContextBackup {
   // Host State
   // RIP and RSP is stored in GPRs here
@@ -22,8 +27,12 @@ struct X86ContextBackup {
 
   // Guest state
   int Signal;
+  uint32_t Flags;
+  uint64_t OriginalRIP;
+  uint64_t FPStateLocation;
+  uint64_t UContextLocation;
+  uint64_t SigInfoLocation;
   FEXCore::Core::CPUState GuestState;
-
   static constexpr int RedZoneSize = 128;
 };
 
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -80,6 +81,77 @@ void Dispatcher::RestoreThreadState(void *ucontext) {
...
+  if (Context->UContextLocation) {
+    auto Frame = ThreadState->CurrentFrame;
+
+    if (Context->Flags &ArchHelpers::Context::ContextFlags::CONTEXT_FLAG_INJIT) {
+      // XXX: Unsupported since it needs state reconstruction
+      // If we are in the JIT then SRA might need to be restored to values from the context
+      // We can't currently support this since it might result in tearing without real state reconstruction
+    }
+
+    if (!(Context->Flags &ArchHelpers::Context::ContextFlags::CONTEXT_FLAG_32BIT)) {
```
lioncash, line 94:
```suggestion
    if (!(Context->Flags & ArchHelpers::Context::ContextFlags::CONTEXT_FLAG_32BIT)) {
```

External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
```diff
@@ -13,6 +13,11 @@
 
 namespace FEXCore::ArchHelpers::Context {
 
+enum ContextFlags {
```
lioncash, line 16: Perhaps

```suggestion
enum ContextFlags : uint32_t {
```

To give it a concrete size and allow it to be forward declared

lioncash (commented):
Just two small nits and it looks good to me

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
Sonicadvance1: Resolved.

lioncash (approved):

PR #1360, opened 2021-11-12
Linux: Emulate MAP_32BIT on mmap

Sonicadvance1:
Previous we were just using an address hint to emulate MAP_32BIT.
Seemingly this behaviour has changed on AArch64 where it now isn't
guaranteed to scan up from the hint provided if exact allocation fails.

Now we pull in the full 32-bit allocator and add support for MAP_32BIT
in it. This limits the allocations there in to the first 2GB which Linux
expects.

Necessary for Mono's trampolines to work since it requires code to be in
the first 2GB on x86-64.


Source/Tests/LinuxSyscalls/LinuxAllocator.cpp
```diff
@@ -0,0 +1,588 @@
...
+private:
+  // Set that contains 4k mapped pages
+  // This is the full 32bit memory range
+  std::bitset<0x10'0000> MappedPages;
+  std::map<uint32_t, int> PageToShm{};
+  uint64_t LastScanLocation{};
+  uint64_t LastKeyLocation{};
+  uint64_t LastKeyLocation32Bit{};
+  std::mutex AllocMutex{};
+  uint64_t FindPageRange(uint64_t Start, size_t Pages);
```
lioncash, line 76:
```suggestion
  uint64_t FindPageRange(uint64_t Start, size_t Pages) const;
```

Ditto for the `_TopDown` variant as well

lioncash (commented):
one little nit

Source/Tests/LinuxSyscalls/LinuxAllocator.cpp
Sonicadvance1: Resolved

lioncash (approved):

PR #1362, opened 2021-11-12
OpcodeDispatcher: Add helper for getting bit sizes

lioncash:
Allows expressing directly that we need the bit size instead of needing to repeat the operation in several places, given how common this is.


Sonicadvance1 (approved):
Nice. Lots of little cleanups.

neobrain:
Shouldn't these new functions return a type larger than `uint8_t`? `GetSrcSize`/`GetDstSize` may return 32, which after multiplication won't fit into 8 bits. (We might even consider returning a signed type to help ubsan catch overflows like this, though not sure how well that fits in with existing code.)

Could a test have caught this, possibly indirectly by more extensively testing some opcode for which `GetSrcSize` returns 32? I'm surprised something as elementary as the opcode dispatcher didn't barf over this.

PR #1363, opened 2021-11-12
OpcodeDecoder: Convert enums into enum classes where applicable

lioncash:
Makes them strongly typed so that invalid values can't implicitly be assigned to relevant variables

While we're at it, this adds helper utilities that make working with enum classes a little more convenient for avoiding boilerplate in regards to enum classes that can be used as flags. Now the relevant operations don't need to be defined long-form for every individual enum class.

### DIFF
--- a/External/FEXCore/include/FEXCore/Utils/EnumUtils.h
+++ b/External/FEXCore/include/FEXCore/Utils/EnumUtils.h
@@ -1,0 +1,72 @@
+#pragma once
+
+// Header for various utilities related to operating with enums
+
+#include <type_traits>
+
+namespace FEXCore {
+
+// Macro that defines all of the built in operators for conveniently using
+// enum classes as flag types without needing to define all of the basic
+// boilerplate.
+#define FEX_DECLARE_ENUM_FLAG_OPERATORS(type)                                                      \
+    [[nodiscard]] constexpr type operator|(type a, type b) noexcept {                              \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(static_cast<T>(a) | static_cast<T>(b));                           \
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator&(type a, type b) noexcept {                              \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(static_cast<T>(a) & static_cast<T>(b));                           \
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator^(type a, type b) noexcept {                              \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(static_cast<T>(a) ^ static_cast<T>(b));                           \
+    }                                                                                              \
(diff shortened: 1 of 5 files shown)


Sonicadvance1 (approved):

External/FEXCore/include/FEXCore/Utils/EnumUtils.h
```diff
@@ -0,0 +1,72 @@
...
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator&(type a, type b) noexcept {                              \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(static_cast<T>(a) & static_cast<T>(b));                           \
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator^(type a, type b) noexcept {                              \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(static_cast<T>(a) ^ static_cast<T>(b));                           \
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator<<(type a, type b) noexcept {                             \
```
neobrain, line 25: All operators in this file are useful to have, but having shift operators seems esoteric and I'd expect this to be used accidentally more often than intentionally. Are these used anywhere currently and/or do other projects use this commonly?

Without wide use I'd opt for requiring the user to do any such shifting explicitly instead.

What I could see more commonly used is an expression like `1 << ENUM_VALUE`, which wouldn't compile by the current definition (I hope? You never know with implicit conversions...). Perhaps a more useful alternative is `operator<<(underlying_int, type)` hence?

External/FEXCore/include/FEXCore/Utils/EnumUtils.h
```diff
@@ -0,0 +1,72 @@
...
+    }                                                                                              \
+    constexpr type& operator>>=(type& a, type b) noexcept {                                        \
+        a = a >> b;                                                                                \
+        return a;                                                                                  \
+    }                                                                                              \
+    [[nodiscard]] constexpr type operator~(type key) noexcept {                                    \
+        using T = std::underlying_type_t<type>;                                                    \
+        return static_cast<type>(~static_cast<T>(key));                                            \
+    }                                                                                              \
+    [[nodiscard]] constexpr bool True(type key) noexcept {                                         \
```
neobrain, line 57: What benefit do True/False have over an `explicit operator bool`?

External/FEXCore/include/FEXCore/Utils/EnumUtils.h
lioncash: These are currently unused, so we can just remove them for the time being.

External/FEXCore/include/FEXCore/Utils/EnumUtils.h
lioncash: Is it even possible to define that outside of a class?

neobrain: Good point - it's not possible to define `operator bool` (`explicit` or not) outside of a class indeed. Current approach seems fine then :)

PR #1365, opened 2021-11-12
OpcodeDispatcher: Implement MULX

lioncash:
Now with all the other instructions out of the way, we have the fun ones left to implement (PDEP and PEXT)

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -2372,6 +2372,21 @@ void OpDispatchBuilder::RORX(OpcodeArgs) {
   StoreResult(GPRClass, Op, Result, -1);
 }
 
+void OpDispatchBuilder::MULX(OpcodeArgs) {
+  // RDX is the implied source operand in the instruction
+  const auto RDXOffset = offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RDX]);
+  const auto OperandSize = GetSrcSize(Op);
+
+  OrderedNode* Src1 = LoadSource(GPRClass, Op, Op->Src[1], Op->Flags, -1);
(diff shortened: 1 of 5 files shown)


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2372,6 +2372,29 @@ void OpDispatchBuilder::RORX(OpcodeArgs) {
   StoreResult(GPRClass, Op, Result, -1);
 }
 
+void OpDispatchBuilder::MULX(OpcodeArgs) {
+  // RDX is the implied source operand in the instruction
+  const auto RDXOffset = offsetof(FEXCore::Core::CPUState, gregs[FEXCore::X86State::REG_RDX]);
+  const auto OperandSize = GetSrcSize(Op);
+
+  // Intel manual states that operands are always 32-bit
```
Sonicadvance1, line 2380: I believe this is incorrect. It should be that VEX.W is ignored. So 32-bit is always 32-bit operation.
Encoding the assembly manually for a 32-bit ASM test would prove this.

Sonicadvance1:
This reminds me that I should spend some time getting a 32-bit host runner hooked up to CI for the ASM tests.
This way we can easily test things like this.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
lioncash: Amended to remove the decode failure

Sonicadvance1: With the decode failure removed. What is the current behaviour of VEX.W in 32-bit?
Does FEX currently debug this as a 64-bit op incorrectly?

lioncash: Not anymore =)

Sonicadvance1 (approved):

PR #1366, opened 2021-11-12
OpcodeDecoder: Shorten up GPR and MM base offset retrieval

lioncash:
Previously, getting the offset of a register involved a long string of code via offsetof to get. We can wrap these behind two helper functions to make writing code that requires offsets a little more convenient


Sonicadvance1:
Do these helpers get inlined everywhere even with the one's that don't have the implementation body visible outside of OpcodeDispatcher?

lioncash:
we compile with LTO, so I'd imagine so. I can move them into the header if that's a concern though

Sonicadvance1:
Would be nice for them to be in the header to enforce intent here.

lioncash:
Done

Sonicadvance1 (approved):

Issue #581, opened 2020-12-12
Support AVX list


Sonicadvance1:
Find a list of games that use AVX unconditionally to gauge priority for implementing.
We need both good enough performance and a list of games that use AVX before thinking about implementing.
It'll have a minor perf hit to implement since all of our context switching ends up being twice the size.

Games that use AVX unconditionally (at time of writing or that we've discovered):
- Cyberpunk 2077 (Unconditional usage removed in 1.05)
- Metro Exodus (And enhanced edition)
- Sonic Frontiers
- Uncharted: Legacy of Thieves Collection
- Returnal
- Death Stranding
- Forspoken
- Star citizen
- Sea of Stars
- Starfield

https://wiki.fex-emu.com/index.php/Category:AVX_CPU_Feature

Sonicadvance1:
Death Stranding also uses AVX unconditionally.

lioncash:
AVX1 support is in place as of writing. Now all that remains is the gather instructions from AVX2 and then CPUID support for that can be put in place

Joshua-Ashton:
Helldivers 2 needs AVX

alyssarosenzweig:
🎉

PR #1369, opened 2021-11-15
OpcodeDispatcher: Amend return types of BitSize helpers

lioncash:
Amends the return size to handle a concern brought up by neobrain in #1362 

While we're at it, we can mark these helper functions as `[[nodiscard]]` to catch obvious bugs

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -4466,11 +4466,11 @@ uint8_t OpDispatchBuilder::GetSrcSize(X86Tables::DecodedOp Op) const {
   return Size;
 }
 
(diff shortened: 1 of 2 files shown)


lioncash:
I don't think we have any unit testing scaffolding set up to make API tests convenient, based off a cursory look (we don't even have anything like Catch set up). We only have stuff setup to unit-test emulated x86(-64) and IR behaviors

neobrain:
What opcodes return 32 for GetSrcSize/GetDstSize? Do we have coverage for any of them? If not, that's the part I'm suggesting to be added, since filling coverage gaps by adding tests to the existing ASM tests would be fairly easy.

lioncash:
I believe that would require AVX support, which we don't support yet (https://github.com/FEX-Emu/FEX/issues/581)

Sonicadvance1 (approved):

neobrain:
Ah, sounds good then. Thanks for the fix!

neobrain (approved):

PR #1370, opened 2021-11-15
EnumUtils: Remove shift operators from helper macro

lioncash:
These aren't strictly necessary for enum flags and through discussion in #1363, would lead to an awkward to use overload.

If these are ever needed, they can be discussed/added back at a later date.


Sonicadvance1 (approved):
Sounds good

PR #1371, opened 2021-11-15
Arm64: Adds an inline syscall optimization

Sonicadvance1:
In a syscall microbench this improves performance by ~19% on my
Snapdragon 888.
Going from ~9.6 million syscalls per second to ~11.5 million.

Macbook Pro is less effective here due to high syscall overhead due to
VM. Going form  7.2M/s to 7.6M/s, ~6% improvement

We can also inline some 32-bit syscalls but that will need some more
work which isn't done yet. Even though the op in the JIT supports it.


External/FEXCore/Source/Interface/Core/JIT/Arm64/BranchOps.cpp
```diff
@@ -370,6 +504,7 @@ void Arm64JITCore::RegisterBranchHandlers() {
   REGISTER_OP(JUMP,              Jump);
   REGISTER_OP(CONDJUMP,          CondJump);
   REGISTER_OP(SYSCALL,           Syscall);
+  REGISTER_OP(INLINESYSCALL,     InlineSyscall);
```
neobrain, line 507: Could you add a brief comment describing the difference between SYSCALL and INLINESYSCALL? It's probably obvious if you already know how SYSCALL is implemented, but if you look at them without prior knowledge it's difficult to see why there are two and how they differ.

Sonicadvance1: Added to IR.json

Sonicadvance1:
Just need to resolve an edge case issue to fix the gvisor problem. Will resolve that tomorrow.

External/FEXCore/Source/Interface/IR/IR.json
```diff
@@ -676,6 +678,33 @@
       ]
     },
 
+    "InlineSyscall": {
+      "HasSideEffects": true,
+      "Desc": ["Dispatches a guest syscall directtly to the host syscall interface",
```
neobrain, line 683: Thanks, that's helpful :)

Small rephrase with spelling/grammar fixups:

       "Desc": ["Dispatches a guest syscall directly to the host syscall interface,",
               "bypassing the SyscallHandler class used by Syscall.",
               "This has significantly less overhead than Syscall, which needs to save JIT state first.",
               "Can only be used for syscalls that match across architecture,",
               "such as gettid (matches on x86/x86-64/Arm64).",

Not sure what the "Which can be directly called" was trying to say. Maybe you can backpatch that into my rephrase?

Sonicadvance1: Was just saying that the syscall can be directly passed through without any fixups.

Sonicadvance1: Changed the description to match yours.

External/FEXCore/Source/Interface/Core/JIT/Arm64/BranchOps.cpp
```diff
@@ -235,6 +235,152 @@ DEF_OP(Syscall) {
...
+        Reg.GetCode() == x5.GetCode()) {
+
+      SpillMask |= (1U << Reg.GetCode());
+      Intersects = true;
+    }
+  }
+  // XXX: For some reason spilling only the x4, x5, and x8 registers was causing issues
+  // Come back to this once investigation reveals why it fails the gvisor ioctl test
+  // For now override to all GPRs
+  SpillMask = ~0U;
```
Sonicadvance1, line 274: This disables the SRA spilling optimization for now.
I'm not quite sure why it is causing a failure so I left a comment.
Theoretically we only need to spill:
- x8: This contains the syscall number. Always needs to spill
- x4: Only when the syscall uses 5 arguments
- x5: Only when the syscall uses 6 arguments

But the gvisor ioctl_test was having issues with it.
So instead we will spill all the SRA GPRs and refill on syscall return.
This isn't _quite_ as optimal but we still are around 11.2 million syscalls per second.

PR #1376, opened 2021-11-17
Frontend: Handle VEX RXB bits

lioncash:
These are equivalent to the REX prefix's RXB bits, except that they're in 1's complement form. These are trivial to handle and fix usages of the upper range of registers for ModRM encoded fields.

To ensure that we have coverage for existing possible usages, I've altered the BEXTR test to make use of R14 and R15.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/External/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -775,6 +775,11 @@ bool Decoder::NormalOpHeader(FEXCore::X86Tables::X86InstInfo const *Info, uint16
     const uint8_t Byte1 = ReadByte();
     DecodedHeader options{};
 
+    if ((Byte1 & 0b10000000) == 0) {
+      LOGMAN_THROW_A(CTX->Config.Is64BitMode, "VEX.R shouldn't be 0 in 32-bit mode!");
+      DecodeInst->Flags |= DecodeFlags::FLAG_REX_XGPR_R;
+    }
+
     if (Op == 0xC5) { // Two byte VEX
       pp = Byte1 & 0b11;
       options.vvvv = 15 - ((Byte1 & 0b01111000) >> 3);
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -775,6 +775,10 @@ bool Decoder::NormalOpHeader(FEXCore::X86Tables::X86InstInfo const *Info, uint16
     const uint8_t Byte1 = ReadByte();
     DecodedHeader options{};
 
+    if ((Byte1 & 0b10000000) == 0) {
+      DecodeInst->Flags |= DecodeFlags::FLAG_REX_XGPR_R;
```
Sonicadvance1, line 779: According to documentation while in 32-bit mode this must ALWAYS be 1.
Maybe add an assert for 32-bit mode && 0 for this?

External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -785,6 +789,12 @@ bool Decoder::NormalOpHeader(FEXCore::X86Tables::X86InstInfo const *Info, uint16
       map_select = Byte1 & 0b11111;
       options.vvvv = 15 - ((Byte2 & 0b01111000) >> 3);
       options.w = (Byte2 & 0b10000000) != 0;
+      if ((Byte1 & 0b01000000) == 0) {
+        DecodeInst->Flags |= DecodeFlags::FLAG_REX_XGPR_X;
```
Sonicadvance1, line 793: Same thing for XGPR_X as XGPR_R.
Must be 1.

External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -785,6 +789,12 @@ bool Decoder::NormalOpHeader(FEXCore::X86Tables::X86InstInfo const *Info, uint16
...
+      }
+      if ((Byte1 & 0b00100000) == 0) {
+        DecodeInst->Flags |= DecodeFlags::FLAG_REX_XGPR_B;
```
Sonicadvance1, line 796: According to documentation, in 32-bit mode this bit is straight up ignored.
Maybe should be ignored in that instance?

lioncash:
Addressed

Sonicadvance1 (approved):

PR #1380, opened 2021-11-18
Thunks: Fail loudly if thunking is enabled for a library that isn't installed

neobrain:
If a host library cannot be found, this now aborts program execution rather than silently continuing (and obscurely crashing once the library is actually used).

~~Furthermore, `ERROR_AND_DIE` now resets SIGTRAP/SIGABRT signal handlers to avoid games handling these signals using a custom handler and hence delaying shutdown. In Ender Lilies in particular, this avoids several lines of output being added past the actual FEX error, making it much easier to spot the latter.~~ An alternative fix for this was provided via #1383.


Sonicadvance1 (approved):

neobrain:
This behavior breaks 6 of the ASM unit tests, which expect to be able to use OP_THUNK with an invalid library name without error. I considered a couple of options to adapt for this:

1. Use a dedicated opcode for testing, which doesn't have hard-erroring behavior
2. Use more sophisticated error reporting such that we can still `ERROR_AND_DIE` in real applications while testing for the correct erroring behavior in tests
3. Add a config option to soften LoadLib() errors
4. Install an actual "not_a_real_lib-host.so" thunk library for the test harness
5. Remove those tests in favor of the upcoming thunking CI
6. Hardcode a dummy library name in the LoadLib() error handler (currently implemented in this PR)

Options 1-4 are unreasonably heavy design changes that aren't justified by the few tests that need this behavior.

Option 5 is tempting: After all, are these unit tests really useful? What we want to verify is that thunk symbols are loaded properly, but currently the tests only ensure the ocpode 0xF 0x3F is recognized by FEX at all. That's better than nothing, but it's also a property that's incredibly unlikely to regress without other things breaking *en masse*.

The simplest option was 6, hence I went with it for this PR, but ideally we'll adopt 5 once we have proper CI coverage of thunks.

Sonicadvance1:
I think Option 5 is a fine route to take. Thunk CI will come up relatively soon and just like you said, if something regresses it is unlikely that it'll come from the specialized opcode.

neobrain:
Dropped OP_THUNK unit tests and reverted back to the "always fail loudly" implementation.

Sonicadvance1 (approved):

PR #1385, opened 2021-11-19
RootFS: Keep rootfs around while in container

Sonicadvance1:
For some reason the full container path isn't actually setup right now.
Until this problem is resolved, keep the FEX rootfs configuration
around.

This lets applications still execute albiet still wrapped with the FEX
rootfs rather than the container's

### DIFF
--- a/Source/Common/RootFSSetup.cpp
+++ b/Source/Common/RootFSSetup.cpp
@@ -221,12 +221,18 @@ bool Setup(char **const envp) {
   // If it is setup to use a squashfs then we need to do something more complex
 
   FEX_CONFIG_OPT(LDPath, ROOTFS);
-  auto ContainerPrefix = FEXCore::Config::FindContainerPrefix();
-  if (!ContainerPrefix.empty()) {
-    // If we are inside of a rootfs/container then drop the rootfs path
-    // Root is already our rootfs
-    FEXCore::Config::Erase(FEXCore::Config::CONFIG_ROOTFS);
-    return true;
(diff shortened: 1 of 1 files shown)


Source/Common/RootFSSetup.cpp
```diff
@@ -221,13 +221,14 @@ bool Setup(char **const envp) {
...
   FEX_CONFIG_OPT(LDPath, ROOTFS);
-  auto ContainerPrefix = FEXCore::Config::FindContainerPrefix();
-  if (!ContainerPrefix.empty()) {
-    // If we are inside of a rootfs/container then drop the rootfs path
-    // Root is already our rootfs
-    FEXCore::Config::Erase(FEXCore::Config::CONFIG_ROOTFS);
-    return true;
-  }
+  // XXX: Disabled for now. Causing problems due to weird filesystem problems
+  //auto ContainerPrefix = FEXCore::Config::FindContainerPrefix();
```
neobrain, line 225: How about using `if (false)` instead of a block comment so it's ensured this keeps compiling properly in the future?

Is there any way to provide more detail about the issue (what goes wrong? how to reproduce?) for future reference?

Sonicadvance1: Switched over to an if statement and added more to the comment about how to reproduce.

neobrain:
Typo in the commit message: `albiet` => `albeit`

Looks fine other than that, thanks for extending the description :)

PR #1386, opened 2021-11-20
JIT: Ensures signals in compileservice JIT space is handled

Sonicadvance1:
If a SIGBUS is received in compile service code then we weren't handling
it correctly. Instead we would fail the JIT space check and hand it off
to the guest.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
@@ -3,6 +3,7 @@
 #include "Interface/Core/ArchHelpers/MContext.h"
 #include "Interface/Core/Dispatcher/Dispatcher.h"
 #include "Interface/Core/X86HelperGen.h"
+#include "Interface/Core/CompileService.h"
 
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Core/CoreState.h>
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -655,17 +656,22 @@ void Dispatcher::RemoveCodeBuffer(uint8_t* start_to_remove) {
...
 
-bool Dispatcher::IsAddressInJITCode(uint64_t Address, bool IncludeDispatcher) const {
+bool Dispatcher::IsAddressInJITCode(uint64_t Address, bool IncludeDispatcher, bool IncludeCompileService) const {
   for (auto [start, end] : CodeBuffers) {
     if (Address >= start && Address < end) {
       return true;
     }
   }
 
+  bool Res = false;
```
neobrain, line 666: How about this to clarify the return value strategy?
```
  if (IncludeDispatcher && IsAddressInDispatcher(Address)) {
      return true;
  }
  
  if (IncludeCompileService && ThreadState->CompileService &&
        ThreadState->CompileService->ISAddressInJITCode(Address)) {
     return true;
  }
  
  return false;
```

Sonicadvance1: Changed to this example.

PR #1387, opened 2021-11-20
Linux: Fixes sigprocmask with new and old set being the same address

Sonicadvance1:
If set and oldset point to the same location. We need to be careful to
store the old mask before returning it. Otherwise we will will store the
current mask over the new mask and not change anything.

### DIFF
--- a/Source/Tests/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tests/LinuxSyscalls/SignalDelegator.cpp
@@ -456,9 +456,14 @@ namespace FEX::HLE {
   }
 
   uint64_t SignalDelegator::GuestSigProcMask(int how, const uint64_t *set, uint64_t *oldset) {
-    if (!!oldset) {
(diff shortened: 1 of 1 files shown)


Source/Tests/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -456,9 +456,12 @@ namespace FEX::HLE {
   }
 
   uint64_t SignalDelegator::GuestSigProcMask(int how, const uint64_t *set, uint64_t *oldset) {
-    if (!!oldset) {
-      *oldset = ThreadData.CurrentSignalMask.Val;
-    }
+    // Ordering matters in setting plus getting
```
neobrain, line 459: These two lines are difficult to parse - any chance there's a clearer way of putting it?

Sonicadvance1: Updated comment.

PR #1393, opened 2021-11-22
Linux: Fixes MAP_32BIT error case

Sonicadvance1:
If MAP_32BIT was failing then we weren't checking error result
correctly.
This was causing us to crash.

Now add a helper function for checking syscall results since it is easy
to mess up.

Replaces the few cases where we were manually checking with the helper.

Fixes a Dota: Underlords crash

### DIFF
--- a/Source/Tests/LinuxSyscalls/Syscalls.h
+++ b/Source/Tests/LinuxSyscalls/Syscalls.h
@@ -349,6 +349,38 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
     return flags;
   }
 
+/**
+ * @brief Checks raw syscall return for error
+ *
+ * This should only be used with raw syscall usage
+ *
+ * This should not be used with glibc wrapped syscall functions
+ *   - This includes the glibc ::syscall(...) function
+ *   - This is due to glibc already wrapping the return and setting errno
+ *
+ * This function should not be used with UAPI breaking syscall results
+ * ioctl specifically will break this convention.
+ *
+ * @param Result The raw syscall return
+ *
+ * @return If the result was an error result
+ */
+
+[[maybe_unused]]
+static bool HasSyscallError(uint64_t Result) {
+  // MAX_ERRNO is part of the Linux Syscall ABI
+  // Redefined here since it doesn't exist as a visible define in the UAPI headers
+  constexpr uint64_t MAX_ERRNO = 0xFFFF'FFFF'FFFF'0001ULL;
+  // Raw syscalls are guaranteed to not return a valid result in the range of [-4095, -1]
+  // In cases where FEX needs to use raw syscalls, this helper checks for this idiom
+  return reinterpret_cast<uint64_t>(Result) >= MAX_ERRNO;
+}
+
+[[maybe_unused]]
+static bool HasSyscallError(const void* Result) {
+  return HasSyscallError(reinterpret_cast<uintptr_t>(Result));
+}
+
 }
 
 // Registers syscall for both 32bit and 64bit
--- a/Source/Tests/LinuxSyscalls/LinuxAllocator.cpp
+++ b/Source/Tests/LinuxSyscalls/LinuxAllocator.cpp
@@ -387,7 +387,7 @@ void *MemAllocator32Bit::mremap(void *old_address, size_t old_size, size_t new_s
   // First, try and allocate a region the size of the new size
   void *MappedPtr = this->mmap(nullptr, new_size, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
   std::scoped_lock<std::mutex> lk{AllocMutex};
-  if (reinterpret_cast<uintptr_t>(MappedPtr) > -4096) {
+  if (FEX::HLE::HasSyscallError(MappedPtr)) {
     // Couldn't find a region that fit our space
     return MappedPtr;
   }
--- a/Source/Tests/LinuxSyscalls/x32/Memory.cpp
+++ b/Source/Tests/LinuxSyscalls/x32/Memory.cpp
@@ -30,7 +30,7 @@ namespace FEX::HLE::x32 {
         mmap(reinterpret_cast<void*>(addr), length, prot,flags, fd, offset);
 
       auto Thread = Frame->Thread;
-      if (Result < -4096) {
+      if (!FEX::HLE::HasSyscallError(Result)) {
         if (!(flags & MAP_ANONYMOUS)) {
           auto filename = get_fdpath(fd);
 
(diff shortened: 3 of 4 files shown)


Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +349,18 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
     return flags;
   }
 
+
+template<typename T>
+static const bool HasSyscallError(T Result) {
```
lioncash, line 354:
```suggestion
static bool HasSyscallError(T Result) {
```

lioncash (commented):
LGTM other than a minor nit

Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +349,18 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
...
+  // Most raw syscalls return errors using the negative range
```
neobrain, line 355: I find this comment difficult to follow, partly because different things are thrown together that don't make it clear how specifically it relates to the rather funky-looking implementation. It's also not clear why you're explaining glibc behavior or mentioning errno.

Maybe this could be restructured:
* An actual docstring at the top, explaining:
  * What are we actually checking? (the raw syscall error or a glibc error? both?)
  * When should this function *not* be used? (ioctl, apparently?)
* In the implementation:
  * Where does the constant 4096 come from?
  * How do glibc and raw syscalls differ?

If possible, we should also avoid the implicit "unsigned > signed" comparison, since that only works if the reader knows the obscure C integer promotion rules by heart. (Worst case is we could use `0xff...ff - 0x4095`?)

Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +349,18 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
...
+  // These are guaranteed to not return a valid result in the range of [-4095, -1]
+  // glibc typically wraps these results and sets errno to -Result and returns -1 from the handler
+  // In cases where FEX needs to use raw syscalls, this helper checks for this idiom
+  // !!! WARNING !!!
+  // Some syscalls may return valid results in this range, like ioctl
+  // Don't use this in those cases
+  return reinterpret_cast<uint64_t>(Result) > -4096;
```
neobrain, line 362: Why is `reinterpret_cast` needed here over `static_cast`?

neobrain, line 362: Is there a reason to compare against `-4096` rather than checking for negativity?

Sonicadvance1: Checking for negativity won't work since something like 0xFFFF'FFFF'FFFF'0000 is a valid result.
Could check the result being `return reinterpret_cast<uint64_t>(Result) > 0xFFFF'FFFF'FFFF'0000ULL;` If you want that to be more clear.

Sonicadvance1: Because we also pass void* in through this interface with mmap routines.

neobrain: Are we? All the uses below appear to cast the result to `uint64_t` before call site.

If I missed one that doesn't, would the following work?
```
static bool HasSyscallError(uint64_t Result) {
    return Result > 0xff..0000;
}
static bool HasSyscallError(void* Result) {
    return HasSyscallError(static_cast<uintptr_t>(Result));
}
```

That would also yield a more constrained interface, where accidentally passing the wrong argument type won't make the compiler explode.

neobrain: Oh, of course. Using the unsigned hex constant sounds good.

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1: Updated

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1: Updated comments.

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1: `Source/Tests/LinuxSyscalls/LinuxAllocator.cpp:390:17` uses it as a `void*` Yes.

Sonicadvance1: Changed to unsigned hex constant.

Sonicadvance1: Changed to fit constrained used case.

Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +349,38 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
...
+ *
+ * This function should not be used with UAPI breaking syscall results
+ * ioctl specifically will break this convention.
+ *
+ * @param Result The raw syscall return
+ *
+ * @return If the result was an error result
+ */
+
+[[maybe_unused]]
```
neobrain, line 369: If `[[maybe_unused]]` is just here due to the compiler yelling about `HasSyscallError` not being used in all .cpp files that include `Syscalls.h`, I think `(static) inline bool HasSyscallError` should work without that attribute.

neobrain:
Looks good, thanks for expanding the comment!

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1: Yes, and the warning stays around with YouCompleteMe sadly. Only thing that resolved it was `[[maybe_unused]]`

Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +349,38 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
...
+  // MAX_ERRNO is part of the Linux Syscall ABI
+  // Redefined here since it doesn't exist as a visible define in the UAPI headers
+  constexpr uint64_t MAX_ERRNO = 0xFFFF'FFFF'FFFF'0001ULL;
+  // Raw syscalls are guaranteed to not return a valid result in the range of [-4095, -1]
+  // In cases where FEX needs to use raw syscalls, this helper checks for this idiom
+  return reinterpret_cast<uint64_t>(Result) >= MAX_ERRNO;
+}
+
+[[maybe_unused]]
+static bool HasSyscallError(void* Result) {
```
lioncash, line 380:
```suggestion
static bool HasSyscallError(const void* Result) {
```

Just to allow it to be used in more contexts

Sonicadvance1: Updated.

PR #1396, opened 2021-11-23
General: Migrate logging over to fmt where possible 

lioncash:
Migrates the remaining bits of the codebase over to fmt to make logs nice and uniform. Now we only need to keep around the debug portion of the printf-style logger to handle the strace syscall in LinuxSyscalls/Syscalls.cpp, everything else can be removed

Addresses https://github.com/FEX-Emu/FEX/issues/146 a little further


Sonicadvance1 (approved):

Sonicadvance1:
Let the rebase pain begin.

PR #1397, opened 2021-11-23
Syscalls: Move construction of syscall name map into GetSyscallName()

lioncash:
These maps are primarily used for debugging, so we don't need to construct and allocate the data until GetSyscallName() is called.

Eliminates a tiny bit of memory usage and a static constructor for release builds.

### DIFF
--- a/Source/Tests/LinuxSyscalls/x32/Syscalls.cpp
+++ b/Source/Tests/LinuxSyscalls/x32/Syscalls.cpp
@@ -45,18 +45,20 @@ namespace FEX::HLE::x32 {
   void RegisterTime();
   void RegisterTimer();
 
-  std::map<int, const char*> SyscallNames = {
-    #include "SyscallsNames.inl"
-  };
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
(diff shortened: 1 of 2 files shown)


Source/Tests/LinuxSyscalls/x32/Syscalls.cpp
```diff
@@ -45,17 +45,17 @@ namespace FEX::HLE::x32 {
   void RegisterTime();
   void RegisterTimer();
 
-  std::map<int, const char*> SyscallNames = {
-    #include "SyscallsNames.inl"
-  };
-
-  const char* GetSyscallName(int SyscallNumber) {
-    const char* name = "[unknown syscall]";
+  [[nodiscard]] static const char* GetSyscallName(int SyscallNumber) {
```
Sonicadvance1, line 48: Would it also make sense to to wrap these `GetSyscallName` functions with the same define check as the usage `#if  defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED` to completely remove it from release builds?

lioncash: Ah true. I've updated the commit so that `GetSyscallName()` is wrapped in the ifdef check

Sonicadvance1 (approved):

PR #1401, opened 2021-11-24
IR: Add type alias for Node IDs 

lioncash:
In quite a few places we have a raw primitive to represent an IR node's ID. This can make reading some bits of the API (or the passes) a little confusing to take in, since there's no meaningful type name for some data structure members, so sometimes scrolling to a particular usage is necessary to see what they represent

We can provide an alias that communicates this directly to the reader at the point of declaration.

This also has the nice benefit of providing a single point of definition for the type of node IDs which can allow for easier changes in the future (e.g. making Node IDs strongly-typed, etc, among other things).

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
@@ -13,7 +13,7 @@ $end_info$
 #include <cstdint>
 
 namespace FEXCore::CPU {
-#define DEF_OP(x) void InterpreterOps::Op_##x(FEXCore::IR::IROp_Header *IROp, IROpData *Data, uint32_t Node)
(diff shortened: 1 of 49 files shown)


Sonicadvance1 (approved):

neobrain:
Nice, looks great! Any opinions on going even one step further and using a strong typedef like `struct NodeID { uint32_t _; }`? That way we can't accidentally use the indescriptive type in some places.

Haven't checked how difficult that would be in this case, but usually for ID-like types you only need a few basic operators like`==` and maybe a `std::hash` specialization if used in hash maps.

lioncash:
Yeah, I'd like to turn it into a strong typedef. This was just the change to introduce the type name to all the APIs without changing everything over all at once so that subsequent changes are easier to review

PR #1404, opened 2021-11-25
jit/arm64: COND_FGT should map to gt, not hi, as it is not FGTU

### LINKED ISSUE #1127: Steamwebhelper burns a few CPU cores and doesn't render on ARM64
steamwebhelper maxes out 4+ CPU cores on ARM64 and fails to render any content.
This looks like for some reason it is just constantly spinning for work and not getting any.
This doesn't happen on x86-64. It even renders some parts of the UI correctly.

This doesn't seem to be optimization pass related. Same problem even disabling optimization passes.
From what I can tell it doesn't look like it is syscall related. I've been combing through syscalls finding weird edge cases, but nothing that has affected the behaviour.
Switching to llvmpipe doesn't resolve the issue, so it isn't freedreno/turnip related.
~~Switching to interpreter /might/ resolve the issue, but it runs so slowly that it may just not be passing jobs between threads?~~
So it looks like it might be a JIT bug? It's being a pain to nail down.

This happens with the steamwebhelper 64-bit processes. The 32-bit steam side seems to not care.
This is hard to tell if it happens in a standalone chromium process, it uses a lot of CPU time just idling.

skmp:
Fixes #1127

Finally.

I ended up taking the extremely long road on bisecting this, as I didn't validate that fcmp-eflag-bypass wasn't a culprit leading to me doing a full on bisection of steam, then proceeded to make a race condition bug in the bisection code, then went down the path of chasing that race in the guest code.

But, it's fixed. FIXED :D

Upside: We have amazing new tooling for bisections.

Will follow up with tooling PR

### DIFF
--- a/Data/AppConfig/steamwebhelper.json
+++ b/Data/AppConfig/steamwebhelper.json
@@ -1,5 +1,0 @@
-{
-  "Config": {
-    "StallProcess": "1"
-  }
-}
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):

Sonicadvance1 (approved):
Big oops!

PR #1406, opened 2021-11-26
RegisterAllocationPass: Reduce usages of NodeIDs where applicable

lioncash:
A basic tidying up of cases where NodeIDs are passed around by repeatedly calling `ID()`. This will make it a little more convenient for turning the NodeID alias into a strong type in a followup, since it minimizes locations that need to be touched (making a review a little nicer).


Sonicadvance1 (approved):
That's some dense reading.

PR #1409, opened 2021-11-26
Adds a new ReentrantMutex to use for FEXCore

Sonicadvance1:
Necessary to workaround a gdbstub hang on shutdown since regular std::mutex and std::condition_variable aren't reentrant safe.

### DIFF
--- a/External/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
+++ b/External/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
@@ -1,0 +1,89 @@
+#pragma once
+
+#include <atomic>
+#include <chrono>
+#include <climits>
+#include <cstdint>
+#include <linux/futex.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+namespace FEXCore {
+  /**
+   * @brief A condition variable that is robust against use of longjmp in signal handlers.
+   *
+   * This is opposed to common `std::condition_variable` implementations:
+   * Longjmp'ing in a signal handler while interrupting a pending `wait_for()`
+   * call can leave the condition variable in an invalid state that breaks later
+   * uses of that object and may cause hangs as a consequence.
+   */
+  class InterruptableConditionVariable final {
+    public:
+      bool Wait(struct timespec *Timeout = nullptr) {
+        while (true) {
+          uint32_t Expected = SIGNALED;
+          uint32_t Desired = UNSIGNALED;
+
+          // If the mutex was already signaled then we can early exit
+          if (Mutex.compare_exchange_strong(Expected, Desired)) {
+            return true;
+          }
+
+          constexpr int Op = FUTEX_WAIT | FUTEX_PRIVATE_FLAG;
+          // WAIT will keep sleeping on the futex word while it is `val`
+          int Result = ::syscall(SYS_futex,
+            &Mutex,
+            Op,
+            Desired, // val
+            Timeout, // Timeout/val2
+            nullptr, // Addr2
+            0); // val3
+
+          if (Timeout && Result == -1 && errno == ETIMEDOUT) {
+            return false;
+          }
+        }
+      }
+
+      template<class Rep, class Period>
+      bool WaitFor(std::chrono::duration<Rep, Period> const& time) {
+        struct timespec Timeout{};
+        auto SecondsDuration = std::chrono::duration_cast<std::chrono::seconds>(time);
+        Timeout.tv_sec = SecondsDuration.count();
+        Timeout.tv_nsec = std::chrono::duration_cast<std::chrono::nanoseconds>(time - SecondsDuration).count();
+        return Wait(&Timeout);
+      }
+
+      void NotifyOne() {
+        DoNotify(1);
+      }
+
+      void NotifyAll() {
+        // Maximum number of waiters
+        DoNotify(INT_MAX);
+      }
+
+    private:
+      std::atomic<uint32_t> Mutex{};
+      constexpr static uint32_t SIGNALED = 1;
+      constexpr static uint32_t UNSIGNALED = 0;
+
+      void DoNotify(int Waiters) {
+        uint32_t Expected = UNSIGNALED;
+        uint32_t Desired = SIGNALED;
+
+        // If the mutex was in an unsignaled state then signal
+        if (Mutex.compare_exchange_strong(Expected, Desired)) {
+          constexpr int Op = FUTEX_WAKE | FUTEX_PRIVATE_FLAG;
+
+          ::syscall(SYS_futex,
+            &Mutex,
+            Op,
+            Waiters, // val - Number of waiters to wake
+            0,       // val2
+            &Mutex,  // Addr2 - Mutex to do the operation on
+            0);    // val3
+        }
+      }
+  };
+}
--- a/unittests/APITests/CMakeLists.txt
+++ b/unittests/APITests/CMakeLists.txt
@@ -1,0 +1,25 @@
+set (TESTS
+  InterruptableConditionVariable)
+
+list(APPEND LIBS FEXCore)
+
+foreach(API_TEST ${TESTS})
+  add_executable(${API_TEST} ${API_TEST}.cpp)
+  target_link_libraries(${API_TEST} PRIVATE ${LIBS} Catch2::Catch2WithMain)
+
+  catch_discover_tests(${API_TEST}
+    TEST_SUFFIX ".${API_TEST}.APITest")
+endforeach()
+
+execute_process(COMMAND "nproc" OUTPUT_VARIABLE CORES)
+string(STRIP ${CORES} CORES)
+
+add_custom_target(
+  api_tests
+  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"
+  USES_TERMINAL
+  COMMAND "ctest" "--timeout" "302" "-j${CORES}" "-R" "\.*.APITest")
+
+foreach(API_TEST ${TESTS})
+  add_dependencies(api_tests ${API_TEST})
+endforeach()
--- a/unittests/APITests/InterruptableConditionVariable.cpp
+++ b/unittests/APITests/InterruptableConditionVariable.cpp
@@ -1,0 +1,154 @@
+#include <catch2/catch.hpp>
+#include <chrono>
+#include <csetjmp>
+#include <FEXCore/Utils/InterruptableConditionVariable.h>
+#include <thread>
+#include <signal.h>
+
+// Test that ensures the Reentrant mutex will timeout without signaling
+TEST_CASE("SimpleWait") {
+  auto Dur = std::chrono::seconds(1);
+  FEXCore::InterruptableConditionVariable Mutex{};
+
+  auto Now = std::chrono::high_resolution_clock::now();
+  bool Signaled = Mutex.WaitFor(Dur);
+  auto End = std::chrono::high_resolution_clock::now();
+
+  // We weren't signaled
+  REQUIRE(Signaled == false);
+
+  // We waited at least the full duration
+  REQUIRE((End - Now) >= Dur);
+}
+
+void WaitThread(FEXCore::InterruptableConditionVariable *Mutex, bool *Signaled) {
+  auto Dur = std::chrono::seconds(5);
+  *Signaled = Mutex->WaitFor(Dur);
+}
+
+// Test that ensure the Reentrant mutex will signal without timing out
+TEST_CASE("SignaledWait") {
+  bool Signaled{};
+  FEXCore::InterruptableConditionVariable Mutex{};
+
+  std::thread t(WaitThread, &Mutex, &Signaled);
+
+  auto Now = std::chrono::high_resolution_clock::now();
+  Mutex.NotifyAll();
+  auto End = std::chrono::high_resolution_clock::now();
+
+  t.join();
+
+  auto Dur = std::chrono::seconds(5);
+  // Expected to signal
+  REQUIRE(Signaled);
+  // Ensure we didn't timeout
+  REQUIRE((End - Now) < Dur);
+}
+
+static jmp_buf LongJump{};
+static int32_t NumberOfJumps{};
+FEXCore::InterruptableConditionVariable WaitMutex{};
+
+void SignalHandler(int Signal) {
+  ++NumberOfJumps;
+  longjmp(LongJump, 1);
+}
+
+void WaitThreadLongJump(
+  FEXCore::InterruptableConditionVariable *Mutex,
+  FEXCore::InterruptableConditionVariable *ThreadReadyMutex,
+  bool *Signaled,
+  int32_t *TID) {
+
+  // Store the TID
+  *TID = ::gettid();
+
+  // Setup a long jump signal handler
+  struct sigaction sa{};
+  sa.sa_flags = SA_RESTART | SA_NODEFER;
+  sigemptyset(&sa.sa_mask);
+  sa.sa_handler = SignalHandler;
+  sigaction(SIGUSR1, &sa, nullptr);
+
+  // long jump here
+  int Value = setjmp(LongJump);
+
+  if (Value == 0) {
+    // Only notify that we are ready once
+    ThreadReadyMutex->NotifyAll();
+  }
+
+  // Notify the loop that we are ready for signaling again
+  WaitMutex.NotifyAll();
+
+  // Time out after two seconds
+  auto Dur = std::chrono::seconds(2);
+  *Signaled = Mutex->WaitFor(Dur);
+}
+
+// Test that ensures the Reentrant mutex survives over a long jump
+// Without signaling the mutex
+TEST_CASE("SignaledWaitLongJumpNoSignal") {
+  int32_t TID{};
+  bool Signaled{};
+  FEXCore::InterruptableConditionVariable Mutex{};
+  FEXCore::InterruptableConditionVariable ThreadReadyMutex{};
+
+  NumberOfJumps = 0;
+  std::thread t(WaitThreadLongJump, &Mutex, &ThreadReadyMutex, &Signaled, &TID);
+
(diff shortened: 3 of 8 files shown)


lioncash:
Since the thunking work is using [catch](https://github.com/catchorg/Catch2), should it be used instead of gtest so that we don't have two library testing dependencies that do the same thing?

Sonicadvance1:
> Since the thunking work is using [catch](https://github.com/catchorg/Catch2), should it be used instead of gtest so that we don't have two library testing dependencies that do the same thing?

Good idea, should switch over to catch instead.

neobrain:
Could `std::recursive_mutex` be used here instead? Not fully sure what sequence of events causes the bug. If the signal handler tries to lock a mutex that was already locked in the interrupted function, naively I would expect `recursive_mutex` to at least not crash. But what are the intended semantics? The signal handler might modify state that the interrupted function expected to be unchanged upon resumption. Does ReentrantMutex do anything specific about this?

Also, where does longjmp come into play? Can't find it being used anywhere in FEX, so I'm probably missing something obvious.

Sonicadvance1:
std::recursive_mutex can't be used here since it instead causes a lock to hang forever.
and the hang manifests differently depending on what you're doing.

If thread A is doing a timed wait, long jumps away, and comes back but the condition_variable is never signaled, then the timed wait hangs forever.

additionally

If thread A is doing a wait, gets a signal and long jumps away, then comes back.
Then Thread B notify's Thread A. Thread A will now continue running but Thread B hangs inside of a lock inside of the condition_variable_any.

Additionally we don't use the longjmp interface directly but instead use the signal handlers to do a long jump. We are required to do tit this way since the glibc longjmp interface is unsafe and will mangle register state on AArch64.

Sonicadvance1:
Tests switched over to Catch2

unittests/APITests/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  target_link_libraries(${API_TEST} PRIVATE ${LIBS} Catch2::Catch2WithMain)
+
+  catch_discover_tests(${API_TEST}
+    TEST_SUFFIX ".APITest")
+endforeach()
+
+execute_process(COMMAND "nproc" OUTPUT_VARIABLE CORES)
+string(STRIP ${CORES} CORES)
+
+add_custom_target(
```
neobrain, line 17: Won't `add_custom_target` cause this to always be executed as part of the build? Is this custom target useful in the first place as opposed to just using `ctest` manually?

unittests/APITests/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+  ReentrantMutex)
+
+list(APPEND LIBS FEXCore)
+
+foreach(API_TEST ${TESTS})
+  add_executable(${API_TEST} ${API_TEST}.cpp)
+  target_link_libraries(${API_TEST} PRIVATE ${LIBS} Catch2::Catch2WithMain)
+
+  catch_discover_tests(${API_TEST}
+    TEST_SUFFIX ".APITest")
```
neobrain, line 11: Wouldn't it make more sense to make the "toplevel" category a prefix of the test name rather than a suffix? Not sure if we have a consistent scheme for existing tests, but would be great to settle for a unified scheme for Catch2 tests going forward.

unittests/APITests/ReentrantMutex.cpp
```diff
@@ -0,0 +1,154 @@
+#include <catch2/catch.hpp>
+#include <chrono>
+#include <csetjmp>
+#include <FEXCore/Utils/ReentrantMutex.h>
+#include <thread>
+#include <signal.h>
+
+// Test that ensures the Reentrant mutex will timeout without signaling
+TEST_CASE("SimpleWait", "Reentrant") {
```
neobrain, line 9: Is this way of specifying the second `TEST_CASE` parameter documented anywhere? The reference docs seem to indicate this is reserved for tags, so it should be "[Reentrant]" instead.

unittests/APITests/CMakeLists.txt
neobrain: Also, how about using `".${API_TEST}.APITest"` instead? That way, we don't need to repeat "Reentrant" for each of the `TEST_CASE`s below.

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
```diff
@@ -0,0 +1,85 @@
...
+      bool WaitFor(std::chrono::duration<Rep, Period> const& time) {
+        struct timespec Timeout{};
+        auto SecondsDuration = std::chrono::duration_cast<std::chrono::seconds>(time);
+        Timeout.tv_sec = SecondsDuration.count();
+        Timeout.tv_nsec = std::chrono::duration_cast<std::chrono::nanoseconds>(time - SecondsDuration).count();
+        return Wait(&Timeout);
+      }
+
+      void NotifyOne() {
+        // Maximum number of waiters
```
neobrain, line 53: Misplaced comment?

neobrain:
> Additionally we don't use the longjmp interface directly but instead use the signal handlers to do a long jump. We are required to do tit this way since the glibc longjmp interface is unsafe and will mangle register state on AArch64.

Hmm I've never heard the term "long jump" used to refer to anything other than a direct call to "longjmp". Are you saying FEX's signal handlers *manually* perform operations equivalent to calling longjmp? Where can I find the corresponding code to learn more about this?


> If thread A is doing a timed wait, long jumps away, and comes back but the condition_variable is never signaled, then the timed wait hangs forever.

What exactly do you mean with "come back"? I keep thinking about a return from the signal handler back to the `condition_variable::wait_for` implementation, but IIUC you're talking about a situation where the signal handler never returns to the interrupted function. So is "come back" referring to *another* call to `condition_variable::wait_for` after the signal handler's long jump as opposed to the same call?

If I got it all right, maybe we could change the class documentation to something like this?
```
A mutex that is robust against use of longjmp in signal handlers.

This is opposed to common `std::condition_variable` implementations:
Longjmp'ing in a signal handler while interrupting a pending `wait_for()`
call can leave the condition variable in an invalid state that breaks later
uses of that object and may cause hangs as a consequence.
```
(Explicitly using `longjmp` here since it's the more widely used terminology, and the fact our signal handlers don't use it directly is more of an implementation detail that can be glanced over here)

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
```diff
@@ -0,0 +1,85 @@
...
+#include <cstdint>
+#include <linux/futex.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+namespace FEXCore {
+  /**
+   * @brief A mutex class that can safely leave on signal and come back without hanging
+   */
+  class ReentrantMutex final {
```
neobrain, line 15: Is this a good name for this utility class? ReentrantMutex doesn't seem to deal with mutual exclusion, and the term seems to be overloaded already (https://en.wikipedia.org/wiki/Reentrant_mutex). My concurrency terminology isn't great either, but I think it's closer to a counting semaphore? (Or maybe RobustConditionVariable works, even though it's more of an integrated std::mutex+std::condition_variable+std::unique_lock)

unittests/APITests/CMakeLists.txt
Sonicadvance1: This follows how all the other unittests are set up.
Each one is set up with a custom target so we can do something like `ninja api_tests` to easily run a series of unit tests.
This is also not always executed as part of the default build. If we wanted that then there would be an `ALL` added to the custom_target.

unittests/APITests/CMakeLists.txt
Sonicadvance1: We have consistent scheme for the unit tests as they are all suffix based.
IR tests: ".ir"
ASM tests: ".asm"
32bit ASM tests: ".asm", but also scans by 32Bit in the filename. Should be changed.
gvisor tests: ".gvisor"
Posix tests: ".posix"
32-bit gcc tests: ".gcc-target-32"
64-bit gcc tests: ".gcc-target-64"

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
Sonicadvance1: Naming is a hard thing. I wouldn't call it similar to a counting semaphore since there isn't any counting done, and I wouldn't call it robust since that conflates it with Linux's robust futexes.

Sonicadvance1:
> > Additionally we don't use the longjmp interface directly but instead use the signal handlers to do a long jump. We are required to do tit this way since the glibc longjmp interface is unsafe and will mangle register state on AArch64.
> 
> Hmm I've never heard the term "long jump" used to refer to anything other than a direct call to "longjmp". Are you saying FEX's signal handlers _manually_ perform operations equivalent to calling longjmp? Where can I find the corresponding code to learn more about this?
> 

Yes. https://github.com/FEX-Emu/FEX/blob/main/External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp#L211
Look for SetPc.

> > If thread A is doing a timed wait, long jumps away, and comes back but the condition_variable is never signaled, then the timed wait hangs forever.
> 
> What exactly do you mean with "come back"? I keep thinking about a return from the signal handler back to the `condition_variable::wait_for` implementation, but IIUC you're talking about a situation where the signal handler never returns to the interrupted function. So is "come back" referring to _another_ call to `condition_variable::wait_for` after the signal handler's long jump as opposed to the same call?
> 

Correct.

> If I got it all right, maybe we could change the class documentation to something like this?
> 
> ```
> A mutex that is robust against use of longjmp in signal handlers.
> 
> This is opposed to common `std::condition_variable` implementations:
> Longjmp'ing in a signal handler while interrupting a pending `wait_for()`
> call can leave the condition variable in an invalid state that breaks later
> uses of that object and may cause hangs as a consequence.
> ```
> 
> (Explicitly using `longjmp` here since it's the more widely used terminology, and the fact our signal handlers don't use it directly is more of an implementation detail that can be glanced over here)

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
Sonicadvance1: Resolved.

unittests/APITests/ReentrantMutex.cpp
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
neobrain: Names need not be perfect, but the current name is misleading in that it isn't used for mutual exclusion and it conflicts with established OS-agnostic terminology (vs a Linux-specific feature that only partly overlaps, i.e. only in the descriptor).

Calling it `<Something>+ConditionVariable` seems like the way to go IMO, since it's the same interface as `std::condition_variable` modulo minor differences (like the use of an internal `std::mutex` instead of a user-provided one).

How about `InterruptableCondVar`/`InterruptableConditionVariable`? It's a lengthy name (perhaps deservedly so for a specific use case) but it aptly describes what kind of object the class is representing *and* what distinguishes it from other synchronization primitives.

unittests/APITests/CMakeLists.txt
neobrain: hmm suffix-syntax for test categories looks very alien to me, but maybe I just need to get used to it :)

Good to know it's mostly consistent already. I'll adapt the thunking PR to use the same scheme then.

unittests/APITests/CMakeLists.txt
neobrain: Ah, I misread the CMake documentation about this. The target is "always considered out of date", but I had skimmed over the part that says "nothing depends on the custom target" unless dependencies are explicitly defined.

Thanks for the clarification! I added some tasks to the thunking PR to make sure I don't forget adopting a similar approach for its tests.

unittests/APITests/CMakeLists.txt
neobrain: Did you have opinions on using `.${API_TEST}.APITest` for test names instead?

Currently, the test list doesn't contain references to the tested class:
```
    Start 37: SimpleWait.APITest
    Start 38: SignaledWait.APITest
    Start 39: SignaledWaitLongJumpNoSignal.APITest
    Start 40: SignaledWaitLongJumpSignal.APITest
1/4 Test #38: SignaledWait.APITest ...................   Passed    0.89 sec
2/4 Test #40: SignaledWaitLongJumpSignal.APITest .....   Passed    0.89 sec
3/4 Test #37: SimpleWait.APITest .....................   Passed    1.88 sec
4/4 Test #39: SignaledWaitLongJumpNoSignal.APITest ...   Passed    2.89 sec
```

With the suggested change, e.g. the first test name should become `SignaledWait.ReentrantMutex.APITest` instead, so it's more apparent what the actual entity under test is.

unittests/APITests/ReentrantMutex.cpp
```diff
@@ -0,0 +1,154 @@
...
+  REQUIRE((End - Now) >= Dur);
+}
+
+void WaitThread(FEXCore::ReentrantMutex *Mutex, bool *Signaled) {
+  auto Dur = std::chrono::seconds(5);
+  *Signaled = Mutex->WaitFor(Dur);
+}
+
+// Test that ensure the Reentrant mutex will signal without timing out
+TEST_CASE("SignaledWait", "[Reentrant]") {
```
neobrain, line 30: This is correct syntax now, but I think this tag should be dropped entirely in favor of giving the tests a subcategory name like `SignaledWait.ReentrantMutex.APITest`, which would allow for filtering out relevant tests too.

unittests/APITests/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+endforeach()
+
+execute_process(COMMAND "nproc" OUTPUT_VARIABLE CORES)
+string(STRIP ${CORES} CORES)
+
+add_custom_target(
+  api_tests
+  WORKING_DIRECTORY "${CMAKE_BINARY_DIR}"
+  USES_TERMINAL
+  COMMAND "ctest" "--timeout" "302" "-j${CORES}" "-R" "\.*.APITest")
```
neobrain, line 21: Shouldn't this depend on the test targets? I've had weird build issues locally where CTest didn't recognize the test names until I had done a full build.
```cmake
foreach(API_TEST ${TESTS})
  add_dependency(api_tests ${API_TEST})
endforeach()
```

Sonicadvance1: Added dependencies.

unittests/APITests/CMakeLists.txt
Sonicadvance1: Changed over to subcategory naming.

External/FEXCore/include/FEXCore/Utils/ReentrantMutex.h
Sonicadvance1: Renamed class to `InterruptableConditionVariable`

unittests/APITests/ReentrantMutex.cpp
Sonicadvance1: Tag dropped

External/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
```diff
@@ -0,0 +1,89 @@
...
+#include <chrono>
+#include <climits>
+#include <cstdint>
+#include <linux/futex.h>
+#include <sys/syscall.h>
+#include <unistd.h>
+
+namespace FEXCore {
+  /**
+   * @brief A mutex that is robust against use of longjmp in signal handlers.
```
neobrain, line 13:
```suggestion
   * @brief A condition variable that is robust against use of longjmp in signal handlers.
```

unittests/APITests/InterruptableConditionVariable.cpp
```diff
@@ -0,0 +1,154 @@
+#include <catch2/catch.hpp>
+#include <chrono>
+#include <csetjmp>
+#include <FEXCore/Utils/InterruptableConditionVariable.h>
+#include <thread>
+#include <signal.h>
+
+// Test that ensures the Reentrant mutex will timeout without signaling
+TEST_CASE("SimpleWait.InterruptableConditionVariable") {
```
neobrain, line 9: Sorry that I have to comment in this line yet again 😅 But why tag each `TEST_CASE` with `InterruptableConditionVariable` instead of including this subcategory in the `TEST_SUFFIX` parameter given to `catch_discover_tests`? It's already clear what entity is being tested from the filename, so this seems like a nuisance waiting for copy-paste typos to happen.

neobrain (approved):
Opened some final minor discussions, looks great otherwise!

unittests/APITests/InterruptableConditionVariable.cpp
Sonicadvance1: Because I've never used catch2 so I'm extrapolating everything from gtest knowledge.

neobrain: Ah, right. Catch does require unique test names but only within each individual test executable. Since each test compiles as its own executable, it's fine to move the subcategory to `catch_discover_tests` too.

Then again, merging all tests into a single Catch executable (in some later PR) could be nice to have, which would require manually suffixing the subcategory for uniqueness after all . Taking that into consideration I don't have a strong opinion on this topic anymore, so let's proceed either way :)

External/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
Sonicadvance1: Resolved.

unittests/APITests/InterruptableConditionVariable.cpp
Sonicadvance1: Resolved by using the TEST_SUFFIX in cmake.

PR #1417, opened 2021-12-02
CPUID: Call handler functions directly

lioncash:
Given the calls are all internal, we can store the function pointers directly and call them with the this pointer instead of indirecting through std::bind and std::function.


Sonicadvance1 (approved):
Nice cleanup

PR #1421, opened 2021-12-03
JitSymbols: Make use of fmt

lioncash:
Simplifies the format strings that get written out, making them a little nicer to visually parse

### DIFF
--- a/External/FEXCore/Source/Common/JitSymbols.cpp
+++ b/External/FEXCore/Source/Common/JitSymbols.cpp
@@ -1,15 +1,15 @@
 #include "Common/JitSymbols.h"
 
 #include <string>
-#include <sstream>
 #include <unistd.h>
+
+#include <fmt/format.h>
 
 namespace FEXCore {
   JITSymbols::JITSymbols() : fp{nullptr, std::fclose} {
-    std::stringstream PerfMap;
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):

External/FEXCore/Source/Common/JitSymbols.cpp
```diff
@@ -1,66 +1,53 @@
...
-#include <sstream>
 #include <unistd.h>
 
+#include <fmt/format.h>
+
 namespace FEXCore {
-  JITSymbols::JITSymbols() {
-    std::stringstream PerfMap;
-    PerfMap << "/tmp/perf-" << getpid() << ".map";
+  JITSymbols::JITSymbols() : fp{nullptr, std::fclose} {
```
neobrain, line 9: > fp{nullptr, std::fclose}

No offense to your code at all, but, wow, C++ handles this poorly. The `FILE*`->`std::unique_ptr<File*,...>` change adds significant overhead here (relative to what you'd expect), since it now needs to carry around a pointer to the "destructor" `std::fclose` at runtime (increasing the object size from one pointer to two).

If we hit this in more performance/memory-sensitive areas, it might make sense to add a small utility that avoids this issue: https://godbolt.org/z/Ef4Msc451

PR #1424, opened 2021-12-05
FEXCore: Reorganizes some AOT related code

Sonicadvance1:
Specifically this tries to avoid changing much behaviour and keeping the
code the same. So most of it is a direct transplant without any
modifications. This is step one of the process so I can start logically
separating the code and making sense of it.

This mostly moves the AOT IR handling to its own independent file for
separation. Cleaning up the Core.cpp file quite heavily.

Two minor behaviour changes that got mixed up with this change.

The first one is an ASAN fix.
This is the FEX_PACKED on the RegisterAllocationData class.
I didn't want to change too heavily how this serialization works but I
wanted to resolve the ASAN error. This may change in the coming work.
Problem was the padding betwene the uint32_t and the PhysicalRegister
wasn't initialized but was being read.
Since it is all uint8_t types afterwards there isn't a perf issue here.

Second fix was a crash that occurs if you're attempting to both capture
and load IR on the same run. This is a quirk where we mmap the original
IR file. Then on shutdown the IR file is getting saved.
At which point we open the IR file again, truncate it, and start
serializing all of the IR data.
The truncation makes it so our mmap of the file is no longer resident,
resulting in a crash when reading our IR cache from the mmap region.
Now open a temporary file and rename it after storing.
Resolves the crash but still doesn't really solve the issue of multiple
processes overwriting the same IR files.

### DIFF
--- a/External/FEXCore/Source/Interface/Context/Context.h
+++ b/External/FEXCore/Source/Interface/Context/Context.h
@@ -4,6 +4,7 @@
 #include "Interface/Core/CPUID.h"
 #include "Interface/Core/HostFeatures.h"
 #include "Interface/Core/X86HelperGen.h"
+#include "Interface/IR/AOTIR.h"
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Core/CoreState.h>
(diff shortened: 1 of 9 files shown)


External/FEXCore/Source/Interface/Context/Context.h
```diff
@@ -337,6 +277,26 @@ namespace FEXCore::Context {
...
+
+    void SetAOTIRLoader(std::function<int(const std::string&)> CacheReader) {
+      IRCaptureCache.SetAOTIRLoader(CacheReader);
+    }
+
+    void SetAOTIRWriter(std::function<std::unique_ptr<std::ofstream>(const std::string&)> CacheWriter) {
+      IRCaptureCache.SetAOTIRWriter(CacheWriter);
+    }
+
+    void SetAOTIRRenamer(std::function<void(const std::string&)> CacheRenamer) {
```
neobrain, line 296: Any particular reason to use `std::function` here? It has heavy allocation overhead compared to a simple function pointer argument, which would work with the capture-less lambda used currently. (Whenever we do find a need to use capturing lambdas, we can just tune the interface after all)

Probably doesn't matter in this instance in particular, but it's an antipattern I've been observing in various places.

External/FEXCore/include/FEXCore/IR/RegisterAllocationData.h
```diff
@@ -28,7 +28,8 @@ union PhysicalRegister {
 
 static_assert(sizeof(PhysicalRegister) == 1);
 
-class RegisterAllocationData {
+// This class is serialized, can't have any holes in the structure otherwise ASAN complains
```
neobrain, line 31:
```suggestion
// This class is serialized, can't have any holes in the structure otherwise ASAN complains about reading uninitialized memory
```

I can guess, but better to be explicit how they're related :)

Source/Tests/FEXLoader.cpp
```diff
@@ -667,6 +673,7 @@ int main(int argc, char **argv, char **const envp) {
 
   if (AOTIRCapture() || AOTIRGenerate()) {
     FEXCore::Context::FinalizeAOTIRCache(CTX);
+    // Rename the temporary file to atomically update the file
```
neobrain, line 676: Stray comment? Or maybe this should go above `FinalizeAOTIRCache`?

External/FEXCore/Source/Interface/Context/Context.h
Sonicadvance1: This is purely to match behaviour with the above SetAOTIRWriter and SetAOTIRLoader.
This will change as I'm hammering on this code.

External/FEXCore/include/FEXCore/IR/RegisterAllocationData.h
Sonicadvance1: Resolved.

Source/Tests/FEXLoader.cpp
Sonicadvance1: Resolved. Comment was meant to be above in the `SetAOTIRRenamer` lambda.

PR #1425, opened 2021-12-06
X86Tables: Build Unknown op definition tables at compile time

Sonicadvance1:
It doesn't make any sense anymore to have specific instruction names
set for UND versus a nullptr string anymore.

Was useful when we could use it to determine the difference between
undefined from the start versus set in the tables but with unknown
decoding. Which is an edge case.

Now instead just zero initialize the data, which means it is an unknown
type and nullptr name. Which works for use.

Improves initialization time of the InitializeInfoTables function from
423 microseconds to 37 microseconds.


External/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -45,26 +46,10 @@ static inline void GenerateTable(X86InstInfo *FinalTable, U8U8InfoStruct const *
   }
 };
 
-static inline void GenerateTable(X86InstInfo *FinalTable, U16U8InfoStruct const *LocalTable, size_t TableSize) {
+template<typename T, typename TT>
```
neobrain, line 49: The first template parameter will cause separate versions of this functions to be generated for each of the arrays it's called with (currently just three but it's an easily avoidable trap). Better to keep using the previous `X86InstInfo*` pointer instead, or `std::span` if that's available for our current targets?

neobrain:
Added two points about the new function signatures, other than that sounds like a solid improvement for a small change :)

Feels like we could build these tables fully at compile-time with a more complex refactor... would the remaining 37 useconds be worth someone's effort?

External/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
Sonicadvance1: Kept the previous raw pointer instead.

External/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -45,26 +46,10 @@ static inline void GenerateTable(X86InstInfo *FinalTable, U8U8InfoStruct const *
...
+static inline void GenerateTableWithCopy(T &FinalTable, TT const *LocalTable, size_t TableSize, T &OtherLocal) {
```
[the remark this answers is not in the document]
Sonicadvance1: Stopped using unconstrained template types.

neobrain:
I did find one minor issue with using nullptr over "UND": Log messages like the one here https://github.com/FEX-Emu/FEX/blob/main/External/FEXCore/Source/Interface/Core/Frontend.cpp#L402 will fail with an fmt-internal assert due to attempting to print a C-string from a null pointer. We'll have to adapt those messages to check for null explicitly instead.

Sonicadvance1:
Resolved the nullptr problem.

External/FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -680,12 +680,12 @@ bool Decoder::NormalOpHeader(FEXCore::X86Tables::X86InstInfo const *Info, uint16
   }
 
   if (Info->Type == FEXCore::X86Tables::TYPE_UNKNOWN) {
-    LogMan::Msg::DFmt("Unknown instruction: {} 0x{:04x} 0x{:x}", Info->Name, Op, DecodeInst->PC);
+    LogMan::Msg::DFmt("Unknown instruction: {} 0x{:04x} 0x{:x}", Info->Name ?: "UND", Op, DecodeInst->PC);
```
neobrain, line 683: Huh, when has `?:` become a thing? 🤯

neobrain: hmm this seems to be a gcc extension: https://gcc.gnu.org/onlinedocs/gcc-4.2.4/gcc/Conditionals.html#Conditionals

Sonicadvance1: Dunno. It's been around for quite a while and is used in Config.cpp as well.

External/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
```diff
@@ -582,18 +582,18 @@ void InitializeSecondaryTables(Context::OperatingMode Mode) {
...
   }
   else {
-    GenerateTable(SecondBaseOps, TwoByteOpTable_32, std::size(TwoByteOpTable_32));
+    GenerateTable(&SecondBaseOps.at(0), TwoByteOpTable_32, std::size(TwoByteOpTable_32));
   }
 
-  GenerateTableWithCopy(RepModOps, RepModOpTable, std::size(RepModOpTable), SecondBaseOps);
-  GenerateTableWithCopy(RepNEModOps, RepNEModOpTable,   std::size(RepNEModOpTable), SecondBaseOps);
-  GenerateTableWithCopy(OpSizeModOps, OpSizeModOpTable, std::size(OpSizeModOpTable), SecondBaseOps);
+  GenerateTableWithCopy(&RepModOps.at(0), RepModOpTable, std::size(RepModOpTable), &SecondBaseOps.at(0));
```
neobrain, line 594: `.data()` is a more direct way to get the data pointer than `&x.at(0)`. Feel free to leave as is though, since my followup changes will touch on this anyway.

neobrain (approved):
Feeling a bit uneasy about the use of the `?:` gcc extension, but seems to cause no issues in Qt Creator at least so it's not a big deal for me. Looks good!

PR #1426, opened 2021-12-08
Arm64: Fixes MapSelectCC FGT flag

Sonicadvance1:
As a continuation to #1404, this was mapped to the incorrect Arm64 flags

Also adds the unit tests that were missing from #1404


lioncash (approved):

skmp:
(nice catch)

PR #1440, opened 2021-12-10
OpcodeDispatcher: Handle PEXT

lioncash:
Handles one of the two remaining BMI2 opcodes. PDEP will follow after

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
@@ -493,6 +493,30 @@ DEF_OP(Extr) {
   }
 }
 
+DEF_OP(PExt) {
+  const auto Op = IROp->C<IR::IROp_PExt>();
+  const auto OpSize = IROp->Size;
+
+  if (OpSize != 4 && OpSize != 8) {
+    LOGMAN_MSG_A_FMT("Unknown PExt Size: {}\n", OpSize);
+    return;
+  }
+
+  const uint64_t Input = OpSize == 4 ? *GetSrc<uint32_t*>(Data->SSAData, Op->Args(0))
+                                     : *GetSrc<uint64_t*>(Data->SSAData, Op->Args(0));
+  uint64_t Mask = OpSize == 4 ? *GetSrc<uint32_t*>(Data->SSAData, Op->Args(1))
+                              : *GetSrc<uint64_t*>(Data->SSAData, Op->Args(1));
+
+  uint64_t Result = 0;
+  for (uint64_t Offset = 0; Mask > 0; Offset++) {
+    const uint64_t Index = std::countr_zero(Mask);
+    Mask &= Mask - 1;
+    Result |= ((Input >> Index) & 1) << Offset;
+  }
+
+  GD = Result;
+}
+
 DEF_OP(LDiv) {
   auto Op = IROp->C<IR::IROp_LDiv>();
   uint8_t OpSize = IROp->Size;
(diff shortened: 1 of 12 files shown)


External/FEXCore/Source/Interface/Core/X86Tables/VEXTables.cpp
```diff
@@ -394,7 +394,7 @@ void InitializeVEXTables() {
     {OPD(2, 0b11, 0xF3), 1, X86InstInfo{"", TYPE_VEX_GROUP_17, FLAGS_NONE, 0, nullptr}}, // VEX Group 17
 
     {OPD(2, 0b00, 0xF5), 1, X86InstInfo{"BZHI", TYPE_INST, FLAGS_MODRM | FLAGS_VEX_2ND_SRC, 0, nullptr}},
-    {OPD(2, 0b01, 0xF5), 1, X86InstInfo{"PEXT", TYPE_UNDEC, FLAGS_NONE, 0, nullptr}},
+    {OPD(2, 0b10, 0xF5), 1, X86InstInfo{"PEXT", TYPE_INST, FLAGS_MODRM | FLAGS_VEX_1ST_SRC, 0, nullptr}},
```
Sonicadvance1, line 397: Can we put a log on the line before that the AMD reference manual is incorrect with its instruction tables?
AMD instruction tables claim the instruction maps to 0b01, implementation pages correctly claim 0b10

External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -493,6 +493,28 @@ DEF_OP(Extr) {
...
+  if (OpSize != 4 && OpSize != 8) {
+    LOGMAN_MSG_A_FMT("Unknown PExt Size: {}\n", OpSize);
+    return;
+  }
+
+  const auto Input = *GetSrc<uint64_t*>(Data->SSAData, Op->Args(0));
+  auto Mask = *GetSrc<uint64_t*>(Data->SSAData, Op->Args(1));
+
+  uint64_t Result = 0;
+  for (uint64_t Offset = 0; Mask > 0; Offset++) {
```
Sonicadvance1, line 509: Does this correctly handle the case when the mask is set to a 64-bit mask, then operates on the 32-bit sized registers? Little bit pedantic but I want to make sure.
eg:
```
mov rax, 0x1234567812345678
mov rbi, 0xFF00FF00FF00FF00
pext ecx, eax, ebx
```

External/FEXCore/Source/Interface/Core/X86Tables/VEXTables.cpp
lioncash: Done

External/FEXCore/Source/Interface/Core/JIT/x86_64/ALUOps.cpp
```diff
@@ -671,6 +671,53 @@ DEF_OP(Extr) {
   }
 }
 
+DEF_OP(PExt) {
```
Sonicadvance1, line 674: If you want this to be an easy implementation, just let it use pext directly.
We already require AVX minspec on x86 host.

External/FEXCore/Source/Interface/Core/JIT/x86_64/ALUOps.cpp
```diff
@@ -671,6 +671,16 @@ DEF_OP(Extr) {
...
+  auto Op = IROp->C<IR::IROp_PExt>();
+
+  const auto Input = static_cast<Xbyak::Reg32e>(GRS(Op->Args(0).ID()));
```
Sonicadvance1, line 677: You likely want something like `const auto Input = OpSize == 4 ? GRS(Op->Args(0).ID()).cvt32() : GRS(Op->Args(0).ID())` But more clean

Sonicadvance1 (approved):

PR #1442, opened 2021-12-13
More AOT code movement

Sonicadvance1:
Only a minor fix in the ELFCodeLoader here, and FEXConfig addition of the options.
Otherwise just code movement.

### DIFF
--- a/Source/Tests/AOT/AOTGenerator.cpp
+++ b/Source/Tests/AOT/AOTGenerator.cpp
@@ -1,0 +1,161 @@
+#include <FEXCore/Core/Context.h>
+#include <FEXCore/Utils/LogManager.h>
+
+#include "ELFCodeLoader2.h"
+#include "Linux/Utils/ELFContainer.h"
+
+#include <cstddef>
+#include <set>
+#include <sys/resource.h>
+#include <sys/sysinfo.h>
+#include <thread>
+#include <queue>
+
+namespace FEX::AOT {
+void AOTGenSection(FEXCore::Context::Context *CTX, ELFCodeLoader2::LoadedSection &Section) {
+  // Make sure this section is executable and big enough
+  if (!Section.Executable || Section.Size < 16)
+    return;
+
+  std::set<uintptr_t> InitialBranchTargets;
+
+  // Load the ELF again with symbol parsing this time
+  ELFLoader::ELFContainer container{Section.Filename, "", true};
+
+  // Add symbols to the branch targets list
+  container.AddSymbols([&](ELFLoader::ELFSymbol* sym) {
+    auto Destination = sym->Address + Section.ElfBase;
+
+    if (! (Destination >= Section.Base && Destination <= (Section.Base + Section.Size)) ) {
+      return; // outside of current section, unlikely to be real code
+    }
+
+    InitialBranchTargets.insert(Destination);
+  });
+
+  LogMan::Msg::IFmt("Symbol seed: {}", InitialBranchTargets.size());
+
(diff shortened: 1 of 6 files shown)


Source/Tests/AOT/AOTGenerator.cpp
```diff
@@ -0,0 +1,161 @@
...
+  std::set<uintptr_t> InitialBranchTargets;
+
+  // Load the ELF again with symbol parsing this time
+  ELFLoader::ELFContainer container{Section.Filename, "", true};
+
+  // Add symbols to the branch targets list
+  container.AddSymbols([&](ELFLoader::ELFSymbol* sym) {
+    auto Destination = sym->Address + Section.ElfBase;
+
+    if (! (Destination >= Section.Base && Destination <= (Section.Base + Section.Size)) ) {
```
neobrain, line 29: Is `<=` correct or should it be `Destination < (Section.Base + Section.Size))`?

Source/Tests/AOT/AOTGenerator.cpp
```diff
@@ -0,0 +1,161 @@
...
+    InitialBranchTargets.insert(Destination);
+  });
+
+  LogMan::Msg::IFmt("Symbol seed: {}", InitialBranchTargets.size());
+
+  // Add unwind entries to the branch target list
+  container.AddUnwindEntries([&](uintptr_t Entry) {
+    auto Destination = Entry + Section.ElfBase;
+
+    if (! (Destination >= Section.Base && Destination <= (Section.Base + Section.Size)) ) {
```
neobrain, line 42: This is duplicated a couple of times throughout this function. Maybe move to a lambda?

Source/Tests/AOT/AOTGenerator.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  LogMan::Msg::IFmt("Symbol + Unwind seed: {}", InitialBranchTargets.size());
+
+  // Scan the executable section and try to find function entries
+  for (size_t Offset = 0; Offset < (Section.Size - 16); Offset++) {
+    uint8_t *pCode = (uint8_t *)(Section.Base + Offset);
+
+    // Possible CALL <disp32>
+    if (*pCode == 0xE8) {
+      uintptr_t Destination = (int)(pCode[1] | (pCode[2] << 8) | (pCode[3] << 16) | (pCode[4] << 24));
```
neobrain, line 57: Hmm this looks like it's intentionally sign-extending to 64-bit. An explicit `int32_t` cast does this more portably and is more explicit.

neobrain: Also, `decltype((uint8_t{} << 24))` is int, so this could trigger UB by overflowing into the sign bit. `unsigned {pCode[4]} << 24` should work.

Source/Tests/AOT/AOTGenerator.cpp
```diff
@@ -0,0 +1,161 @@
...
+      InitialBranchTargets.insert((uintptr_t)pCode);
+    }
+  }
+
+  uint64_t SectionMaxAddress = Section.Base + Section.Size;
+
+  std::set<uint64_t> Compiled;
+  std::atomic<int> counter = 0;
+
+  std::queue<uint64_t> BranchTargets;
```
neobrain, line 82: Looks like `std::vector` or `std::stack<uint64_t, std::vector>` are more lightweight alternatives that are equally suited here.

neobrain:
Reviewed a couple of things before realizing this is just moving around old code. Feel free to ignore 🙈

PR #1445, opened 2021-12-13
OpcodeDispatcher: Handle PDEP

lioncash:
Handles the last remaining BMI2 instruction, letting us signal full support for BMI2 within our emulated CPUID


Sonicadvance1 (approved):
Woo \o/

PR #1456, opened 2021-12-19
IR: Fixes sized constant mask

Sonicadvance1:
This mask was subtracting backwards, which made anything smaller than 8
byte constants not mask correctly


lioncash (approved):

skmp:
Big woops there

PR #1477, opened 2021-12-29
FEXCore: Implements recoverable INTO instruction

### LINKED ISSUE #929: INTO implementation
FEX has an INTO implementation but it was disabled when we found out that it was 32-bit x86 instruction.
Now that we are supporting 32-bit x86 we should implement a unit test for this and wire it to the dispatcher tables.

Might only be a partial unit test since we can't catch or generate synchronous signals yet. There is another issue about supporting that.

Sonicadvance1:
This instruction is only available on 32-bit x86. On overflow flag set,
this instruction will raise a SIGSEGV with overflow exception set in
siginfo_t's si_code.

This implements the everything required to raise the signal and modify
our signal results that we are giving to the guest.

This is effectively the first step in generating signal frames with
custom data in it but only enough for INTO today.

This is enough for our unit tests doing non-faulting, and also a handwritten C test that recovers from the signal.
Fixes #929

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
@@ -24,6 +24,7 @@ struct X86ContextBackup {
   uint64_t GPRs[23];
   FEXCore::x86_64::_libc_fpstate FPRState;
   uint64_t sa_mask;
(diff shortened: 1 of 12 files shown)


lioncash (approved):
LGTM, once tests are fixed

Sonicadvance1:
Oops. IR validation caught me.

PR #1488, opened 2022-01-01
Scripts: Adds a python script that can hand hold a user through FEX install

Sonicadvance1:
This is definitely a bit divisive but overall this is a win.
This is a user pattern that is emerging in a bunch of projects.
Allow an officially sourced script that lets you pipe a script directly
in to python/bash and setup the environment entirely.

This only supports Ubuntu {20.04, 21.04, 21.10, 22.04} which matches
exactly what we expose in the PPA.

Once this is in the repo, and our PPA is updated to the latest release
tag you can run this script like:
`curl --silent <Direct Github raw link> | python3`

Once the PPA is updated, the README and Wiki will be updated with Quick
Start guides to use this path.

The script steps
1) Checks if ARMv8, or fail
2) Checks if supported Ubuntu, or fail
3) Checks if PPA installed
3a) Install PPA if not, or fail
4) Check if packages are installed
4a) Install non-installed packages, or fail
5) Check if RootFS is configured/exists
5a) Run through FEXRootFSFetcher to get/setup rootfs, or fail
6) Attempt to run emulated uname -a through FEX, or fail
7) Provide some examples for how to use FEX
8) Exit with success!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -4,10 +4,25 @@ It has native support for a rootfs overlay, so you don't need to chroot, as well
 FEX presents a Linux 5.0 interface to the guest, and supports both AArch64 and x86-64 as hosts.
 FEX is very much work in progress, so expect things to change.
 
+
+## Quick start guide
+### For Ubuntu 20.04, 21.04, 21.10, 22.04
+Execute the following command in the terminal to install FEX through a PPA.
+
+`curl --silent https://raw.githubusercontent.com/FEX-Emu/FEX/main/Scripts/InstallFEX.py | python3`
+
+This command will walk you through installing FEX through a PPA, and downloading a RootFS for use with FEX.
+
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
PPA has been updated with a new version of FEX to not have the fortify problem. So it'll be safe to use document this publicly now.

Sonicadvance1:
Script needs some minor updates:
1) If PPA is installed, do an apt-get update
2) Check and update packages if they are already installed
3) Find out if a package trigger can be run to update binfmt_misc after a package update
4) fuse3 versus fuse package dependency. Which is default? Does it matter which?

Sonicadvance1:
Readme is now updated with the curl+python3 command and a link to the Wiki's quick start guide.
Which the quick start guide isn't yet populated but will be.

PR #1491, opened 2022-01-03
FEXCore: Don't use Run() inside RunUntilExit()

Sonicadvance1:
Run() is expected to be used after a pause at this point.
Make RunUntilExit() notify using the StartRunning variable entirely.

This is because if we used Run() then the initial thread hasn't yet set
up its core or signal handlers. So it'll ignore SIG63. Then its reason
for pause is still set to Resume.
Once an application tries sending SIG63, we try restoring the context
from a paused state that does exist and crash

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -389,8 +389,13 @@ namespace FEXCore::Context {
   }
 
   FEXCore::Context::ExitReason Context::RunUntilExit() {
-    if(!StartPaused)
-      Run();
+    if(!StartPaused) {
+      // We will only have one thread at this point, but just in case run notify everything
+      std::lock_guard lk(ThreadCreationMutex);
+      for (auto &Thread : Threads) {
+        Thread->StartRunning.NotifyAll();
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -389,8 +389,13 @@ namespace FEXCore::Context {
   }
 
   FEXCore::Context::ExitReason Context::RunUntilExit() {
-    if(!StartPaused)
-      Run();
+    if(!StartPaused) {
+      // We will only have one thread at this point, but just in case run notify everything
+      std::lock_guard<std::mutex> lk(ThreadCreationMutex);
```
lioncash, line 394:
```suggestion
      std::lock_guard lk(ThreadCreationMutex);
```

lioncash (commented):
Just a small nit

External/FEXCore/Source/Interface/Core/Core.cpp
Sonicadvance1: Resolved.

lioncash (approved):

PR #1513, opened 2022-01-11
FEXCore: Defer a significant number of ALU flag calculation

Sonicadvance1:
This was mainly an optimization around memory usage. ALU ops tend to
bloat the IR quite heavily, but I also noticed a 2-4% uplift in
performance of some applications. So a nice side effect.

Should let us more aggressively target reducing our IR intrusive
allocator size since this is quite reduced.

In a pedantic heavy ALU op code block this reduces the number of IR ops
from 14,756 IR ops to 2,016 prior to optimization.
After optimization both had reduced down to 50 IR ops, proving the
output IR was the same.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -78,6 +78,9 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
     LogMan::Msg::DFmt("Unhandled OSABI syscall");
   }
 
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   const uint8_t GPRSize = CTX->GetGPRSize();
   auto NewRIP = GetRelocatedPC(Op, -Op->InstSize);
   _StoreContext(GPRClass, GPRSize, offsetof(FEXCore::Core::CPUState, rip), NewRIP);
@@ -114,6 +117,9 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
 }
 
 void OpDispatchBuilder::ThunkOp(OpcodeArgs) {
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   const uint32_t RSPOffset = GPROffset(X86State::REG_RSP);
   const uint8_t GPRSize = CTX->GetGPRSize();
   uint8_t *sha256 = (uint8_t *)(Op->PC + 2);
@@ -163,6 +169,12 @@ void OpDispatchBuilder::RETOp(OpcodeArgs) {
   // ABI Optimization: Flags don't survive calls or rets
   if (CTX->Config.ABILocalFlags) {
     _InvalidateFlags(~0UL); // all flags
+    // Deferred flags are invalidated now
+    InvalidateDeferredFlags();
+  }
+  else {
+    // Calculate flags early.
+    CalculateDeferredFlags();
   }
 
   auto Constant = _Constant(GPRSize);
@@ -202,6 +214,9 @@ void OpDispatchBuilder::IRETOp(OpcodeArgs) {
     DecodeFailure = true;
     return;
   }
+
+  // Calculate flags early.
+  CalculateDeferredFlags();
 
   const uint32_t RSPOffset = GPROffset(X86State::REG_RSP);
   const uint8_t GPRSize = CTX->GetGPRSize();
@@ -376,6 +391,9 @@ void OpDispatchBuilder::SecondaryALUOp(OpcodeArgs) {
 
 template<uint32_t SrcIndex>
 void OpDispatchBuilder::ADCOp(OpcodeArgs) {
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[SrcIndex], Op->Flags, -1);
   uint8_t Size = GetDstSize(Op);
 
@@ -404,6 +422,9 @@ void OpDispatchBuilder::ADCOp(OpcodeArgs) {
 
 template<uint32_t SrcIndex>
 void OpDispatchBuilder::SBBOp(OpcodeArgs) {
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[SrcIndex], Op->Flags, -1);
   auto Size = GetDstSize(Op);
 
@@ -706,6 +727,12 @@ void OpDispatchBuilder::CALLOp(OpcodeArgs) {
   // ABI Optimization: Flags don't survive calls or rets
   if (CTX->Config.ABILocalFlags) {
     _InvalidateFlags(~0UL); // all flags
+    // Deferred flags are invalidated now
+    InvalidateDeferredFlags();
+  }
+  else {
+    // Calculate flags early.
+    CalculateDeferredFlags();
   }
 
   auto ConstantPC = GetRelocatedPC(Op);
@@ -732,6 +759,9 @@ void OpDispatchBuilder::CALLOp(OpcodeArgs) {
 void OpDispatchBuilder::CALLAbsoluteOp(OpcodeArgs) {
   const uint32_t RSPOffset = GPROffset(X86State::REG_RSP);
   const uint8_t GPRSize = CTX->GetGPRSize();
+
+  // Calculate flags early.
+  CalculateDeferredFlags();
 
   BlockSetRIP = true;
 
@@ -986,6 +1016,9 @@ OrderedNode *OpDispatchBuilder::SelectCC(uint8_t OP, OrderedNode *TrueValue, Ord
 }
 
 void OpDispatchBuilder::SETccOp(OpcodeArgs) {
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   auto ZeroConst = _Constant(0);
   auto OneConst = _Constant(1);
 
@@ -995,6 +1028,9 @@ void OpDispatchBuilder::SETccOp(OpcodeArgs) {
 }
 
 void OpDispatchBuilder::CMOVOp(OpcodeArgs) {
+  // Calculate flags early.
+  CalculateDeferredFlags();
+
   OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags, -1);
   OrderedNode *Dest = LoadSource(GPRClass, Op, Op->Dest, Op->Flags, -1);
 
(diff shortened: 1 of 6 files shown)


skmp:
mnnn :3

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -557,23 +565,409 @@ enum class SelectionFlag {
...
+    TYPE_ROLI,
+    TYPE_FCMP,
+  };
+
+  // Deferred flag generation tracking structure
+  // This structure is used to track RFlags from ALU ops for invalidation
+  // This tracks only six RFLAGS flags.
+  // Specifically: CF, PF, AF, ZF, SF, OF
+  //
+  // These six flags are heavily generated through basic ALU ops and balloon the IR if not early eliminated
```
neobrain, line 595:
```suggestion
  // These six flags are heavily generated through basic ALU ops and balloon the IR if not early eliminated.
```
Multi-line comments are easier to parse with punctuation.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -557,23 +565,409 @@ enum class SelectionFlag {
...
+  // This tracking structure only tracks single blocks and requires RFLAGS canonicalization at block ender ops
```
neobrain, line 596:
```suggestion
  // This tracking structure only tracks single blocks and requires RFLAGS canonicalization at block-ending ops
```

Would this be a good place to explain what `canonicalization` is btw? It's not clear to me if this is something specific to this optimization or some more general x86 concept.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -39,6 +39,11 @@ constexpr std::array<uint32_t, 17> FlagOffsets = {
 };
 
 void OpDispatchBuilder::SetPackedRFLAG(bool Lower8, OrderedNode *Src) {
+  // Canonicalize flags early
+  // Could be invalidate if we had masked invalidation
```
neobrain, line 43: Is `invalidate` referring to another function name here? If so, quotes or trailing `()` help explain that. Currently, the comment reads like a typo'ed sentence and it's unclear what's tried to be said.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -39,6 +39,11 @@ constexpr std::array<uint32_t, 17> FlagOffsets = {
...
+  // Need to be careful since LAHF doesn't store OF
```
neobrain, line 44: What do we need to be careful about?

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: I could drop a dictionary definition on it?
`In computer science, canonicalization (sometimes standardization or normalization) is a process for converting data that has more than one possible representation into a "standard", "normal", or canonical form.`
Or I could change the function name to something like "CalculateDeferredFlags"?
It's not really a term specific to what is going on.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -557,23 +565,409 @@ enum class SelectionFlag {
...
+  // Some flags generating ALU ops only touch part of the registers, In these cases it will do canonicalization up front
+  // This means we still need our IR passes to eliminate all redundant flags accesses but this light OpcodeDispatcher optimization
+  // doesn't take it to that level
+  //
+  // Future ideas: Use an invalidation mask to do partial canonicalization for the instructions that don't do full invalidation
```
neobrain, line 601: With "full invalidation", are you referring to instructions that only set some but not all of these six RFLAGS flags? What are we doing about these instructions currently? Do we just not apply the optimization for those, or do we backtrack to the previous instruction for non-overridden flags?

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -557,23 +565,409 @@ enum class SelectionFlag {
...
+        OrderedNode *Src2;
+
+        uint64_t Imm;
+      } TwoSrcImmediate;
+    } Sources{};
+  };
+
+  DeferredFlagData CurrentDeferredFlags{};
+
+  void CanonicalizeDeferredFlags();
```
neobrain, line 653: Could you add docstrings for CanonicalizeDeferredFlags and for InvalidateDeferredFlags to help distinguish when/why one is used over the other?

neobrain:
Does it make sense to split the ptest bugfix into a dedicated patch? If not, we should probably at least mention in the commit message :)

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Resolved. Deleted all use uses of canonicalize or canonicalization.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
Sonicadvance1: Resolved. Updated comment to reflect real function name.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
Sonicadvance1: Updated with a better comment stating that this is only a partial flag set. So we need to calculate flags early.
Also moved in to the Lower8 section since otherwise it is fine to invalidate the deferred state.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Instructions that do partial flag generation will calculate rflags up front then write to rflags directly.
Updated the comment to better reflect this.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Resolved. Added documentation strings.

Sonicadvance1:
Split the PTest fixes in to its own PR in #1515.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -557,23 +565,446 @@ enum class SelectionFlag {
...
+        OrderedNode *Src1;
+        uint64_t Imm;
+      } OneSrcImmediate;
+
+      // ADD, SUB
+      struct {
+        OrderedNode *Src1;
+        OrderedNode *Src2;
+
+        uint64_t Imm;
```
skmp, line 653: As far as I can tell, this is only used for UpdateCF and not as an imm. How about we use UpdateCF for clarity?

skmp (changes requested):
Overall, this looks good. I've noted a couple of minor things/

I'd want a config option to easily disable the optimization for bisections.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -927,7 +951,8 @@ OrderedNode *OpDispatchBuilder::SelectCC(uint8_t OP, OrderedNode *TrueValue, Ord
       case 0x5: SrcCond = _Select(FEXCore::IR::COND_NEQ, flagsOpDest, ZeroConst, TrueValue, FalseValue, flagsOpSize); break;
       //default: printf("Missed Condition %04X OP_AND\n", OP); break;
     }
-  } else if (flagsOp == SelectionFlag::FCMP) {
+  }
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1323,7 +1369,6 @@ void OpDispatchBuilder::CMPOp(OpcodeArgs) {
   if (Size < 4) {
     Result = _Bfe(Size, Size * 8, 0, ALUOp);
   }
-
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Resolved.

Sonicadvance1:
Converted to a draft for now. This cause the Half-Life 2 title screen text to vanish and I haven't diagnosed it.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -3532,16 +3645,16 @@ void OpDispatchBuilder::STOSOp(OpcodeArgs) {
     OrderedNode *TailDest = _LoadContext(GPRSize, GPROffset(X86State::REG_RDI), GPRClass);
     TailDest = _Add(TailDest, PtrDir);
     _StoreContext(GPRClass, GPRSize, GPROffset(X86State::REG_RDI), TailDest);
-
   }
   else {
+    // Calculate flags early. because end of block
+    CalculateDeferredFlags();
```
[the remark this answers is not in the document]
Sonicadvance1: Comment updated.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1865,6 +1921,8 @@ void OpDispatchBuilder::SHRDOp(OpcodeArgs) {
     Res = _Bfe(Size, 0, Res);
   }
   GenerateFlags_ShiftRight(Op, Res, Dest, Shift);
+  // Calculate flags early.
+  CalculateDeferredFlags();
```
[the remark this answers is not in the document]
Sonicadvance1: Comment updated.

Sonicadvance1:
Nevermind, looks like title text vanishing was from other WIP stuff on the same branch I was messing with. It's fine.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -160,6 +166,9 @@ void OpDispatchBuilder::RETOp(OpcodeArgs) {
   const uint32_t RSPOffset = GPROffset(X86State::REG_RSP);
   const uint8_t GPRSize = CTX->GetGPRSize();
 
+  // Calculate flags early.
```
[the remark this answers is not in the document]
Sonicadvance1: These are currently tracking at different flag granularities and without some fairly extensive refactoring it won't be easy to merge these.
It's best to keep it separate for now and come back for that once it becomes a bigger issue.

Sonicadvance1:
This would be a nice to have for the FEX-2202 tagged revision.

PR #1518, opened 2022-01-14
JIT: Fixes signed displacement wraparound on 32-bit 

### LINKED ISSUE #1517: SIGSEGV in Frontend when running wine + Multiblock
Running `wine notepad.exe` on an AArch64 host attempts to load from address `0xffffffffff470dc0` causing a SIGSEGV. Possible unintentional sign extension from `ff470dc0`.

Sonicadvance1:
This cropped up mostly with multiblock and `jmp <signed displacement>`
This also happened with non multiblock `jcc <signed displacement>`

Due to how IR relocations occur, this needs to happen fairly late but
isn't a big deal.

Fixes #1517

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
@@ -38,7 +38,13 @@ DEF_OP(Constant) {
 
 DEF_OP(EntrypointOffset) {
   auto Op = IROp->C<IR::IROp_EntrypointOffset>();
-  GD = Data->CurrentEntry + Op->Offset;
+  uint64_t Mask = ~0ULL;
+  uint8_t OpSize = IROp->Size;
+  if (OpSize == 4) {
+    Mask = 0xFFFF'FFFFULL;
+  }
+
+  GD = (Data->CurrentEntry + Op->Offset) & Mask;
 }
 
 DEF_OP(InlineConstant) {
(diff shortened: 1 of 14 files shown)


skmp (changes requested):
The x86/64 check should be moved to the frontend, as we don't want configuration state to affect the backends, so they can be used in an offline way, and IR files should contain all information to emit the code.

I'd opt in to move it as a parameter (32bitwrap or something), so what happens is clearly defined.

Sonicadvance1:
This has been moved to the opcode dispatcher. I also fixed and implemented unit tests for overflow on RIP as well as the underflow tests.

Sonicadvance1:
Side note, this PR fixes Counter-Strike: Condition Zero.

skmp:
Still thinking of this, as it doesn't handle the relocation case.

Sonicadvance1:
I don't think that should be a blocker to this change.
Even without this change relocations could still have issues.

skmp:
Carved a bit more time today to think on this a bit more, the unwrapping logic really unnerves me.

My view is that
- The opdispatch handlers should only do the 32 bit masking themselves
- `GetDynamicPC` should be extended with an "is32bit" parameter, and do the masking there too, in the offset calculations
- The IR/Backed (EntrypointOffset/InlineEntrypointOffset) should be extended to correctly do 32-bit or 64-bit operation
- While we're at it, possibly rename `GetDynamicPC` to `GetRelocatedPC` for clarity

This will simplify the semantics of the code, force us to specify 32 or 64 bit usage in GetDynamicPC everywhere (and hopefully make sure no other similar bugs exist) and also handle relocations

Sonicadvance1:
PR updated to address comments.

PR #1522, opened 2022-01-19
FEXCore: Fix build err

wannacu:


wannacu:
after sync fmt, build ok.

PR #1531, opened 2022-01-24
Linux: Fixes 32-bit getsockopt and setsockopt

Sonicadvance1:
On Set, we have four options that need to be converted.
On Get, we have two options that need to be converted.

This fixes a crash that Tomb Raider 2013 was having on launch.


skmp:
Nice catch

skmp (approved):

PR #1534, opened 2022-01-26
CMake: Adds option to support cross compile

wannacu:


CMakeLists.txt
```diff
@@ -19,6 +19,7 @@ option(ENABLE_JEMALLOC "Enables jemalloc allocator" TRUE)
 option(ENABLE_OFFLINE_TELEMETRY "Enables FEX offline telemetry" TRUE)
 option(ENABLE_COMPILE_TIME_TRACE "Enables time trace compile option" FALSE)
 option(ENABLE_LIBCXX "Enables LLVM libc++" FALSE)
+option(ENABLE_CROSS_COMPILE "Enables cross compile FEX to aarch64 on x86_64" FALSE)
```
neobrain, line 22: The checks added below seem a bit hacky, and `ENABLE_CROSS_COMPILE` is a too generic concept here imo. It could equally mean cross-compiling from `aarch64` to `x86_64`, or cross-compiling from `aarch64-linux` to `aarch64-android`.

How about using an override architecture instead? I.e. if you wanted to cross-compile, you'd set a variable `FEX_TARGET_SYSTEM_PROCESSOR` that gets used instead of `CMAKE_SYSTEM_PROCESSOR` throughout our build scripts. If we aren't cross-compiling, the new variable can just default to `CMAKE_SYSTEM_PROCESSOR` then.

wannacu: Yes, it is one hack option, use custom target architecture looks better.

Sonicadvance1:
Instead of this approach you more likely want a toolchain file that just sets the system processor appropriately
eg:
```cmake
# This is a reference AArch64 cross compile script
# Pass in to cmake when building:
# eg: cmake -DCMAKE_TOOLCHAIN_FILE=../CMakeToolchains/AArch64.cmake ..
if (NOT DEFINED ENV{SYSROOT})
  message(FATAL_ERROR "Need to have SYSROOT environment variable set")
endif()

set(CMAKE_SYSTEM_NAME Linux)
set(CMAKE_SYSTEM_PROCESSOR aarch64)
set(CMAKE_CROSSCOMPILING TRUE)

# Target triple needs to match the binutils exactly
set(TARGET_TRIPLE aarch64-linux-gnu)
set(CMAKE_C_COMPILER "clang")
set(CMAKE_CXX_COMPILER "clang++")
set(CMAKE_C_COMPILER_AR "llvm-ar")
set(CMAKE_CXX_COMPILER_AR "llvm-ar")
set(CMAKE_C_COMPILER_RANLIB "llvm-ranlib")
set(CMAKE_CXX_COMPILER_RANLIB "llvm-ranlib")
set(CMAKE_LINKER "ld.lld")

set(CMAKE_C_COMPILER_TARGET ${TARGET_TRIPLE})
set(CMAKE_CXX_COMPILER_TARGET ${TARGET_TRIPLE})

# Set the environment variable SYSROOT to the aarch64 rootfs
set(CMAKE_FIND_ROOT_PATH "$ENV{SYSROOT}")
set(CMAKE_SYSROOT "$ENV{SYSROOT}")

list(APPEND CMAKE_PREFIX_PATH "$ENV{SYSROOT}/usr/lib/${TARGET_TRIPLE}/cmake/")

set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)

set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```

wannacu:
> Instead of this approach you more likely want a toolchain file that just sets the system processor appropriately eg:

I was going to use the toolchain file before, but I didn't find a good example. I will try.

PR #1536, opened 2022-01-29
Softfloat: Stop doing special handling for FREM

### LINKED ISSUE #1374: Incorrect orbital familiars movement in The Binding of Isaac: Rebirth
Orbital familiars are supposed to spin in a circular motion around the main character. In FEX however they bounce around in a semi-circle, changing movement direction each time they have moved by 180°. This can easily be reproduced within few seconds in challenge 6 ("Solar System"), where the main character starts with a number of orbital familiars that each have this problem.

This also affects enemy groups that normally arrange in circles, such as the pink ring flies, which arrange in a semi-circle in FEX instead.

Sonicadvance1:
This isn't correct and breaks games.
This makes the FREM and REM1 implementation the same.
While not 100% correct, it is still better than before.
New issues will be created to handle the differences in the future.

Fixes #1374.
Also fixes most of the HL2 issues, just not the seam issue.

### DIFF
--- a/External/FEXCore/Source/Common/SoftFloat.h
+++ b/External/FEXCore/Source/Common/SoftFloat.h
@@ -57,22 +57,107 @@ struct X80SoftFloat {
 
   // Ops
   static X80SoftFloat FADD(X80SoftFloat const &lhs, X80SoftFloat const &rhs) {
+#ifdef DEBUG_X86_FLOAT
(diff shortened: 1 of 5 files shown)


Sonicadvance1:
#1538 and #1537 created to track the issues with frem.

skmp (approved):
Love the x87 asm fallback

PR #1541, opened 2022-01-30
OpcodeDispatcher: Implements CRC32 instruction 

Sonicadvance1:
Now that the rest of the code matches behaviour, we just need to pass
this through.

Easy enough and get Horizon Zero Dawn running.


External/FEXCore/Source/Interface/Core/Interpreter/EncryptionOps.cpp
```diff
@@ -298,6 +298,75 @@ namespace AES {
   }
 }
 
+namespace CRC32 {
+  // CRC32 per byte lookup table. Generated with bit reversed polynomial 0x1EDC6F41
+  // Generated from a fairly simple routine
+  //
+  // constexpr uint32_t Polynomial = 0x1EDC6F41;
```
neobrain, line 305: Why not use constexpr to codify this directly?

```cpp
constexpr std::array<uint32_t, 256> CRC32CTable = []() {
    std::array<uint32_t, 256> Table{};

    constexpr uint32_t Polynomial = 0x1EDC6F41;
    constexpr uint32_t PolynomialRev = __builtin_bitreverse32(Polynomial);
    for (size_t char = 0; char < 0x100; ++char) {
        uint32_t CurrentChar = char;
        for (size_t i = 0; i < 8; ++i) {
            if (CurrentChar & 1) {
                CurrentChar = (CurrentChar >> 1) ^ PolynomialRev;
            }
            else {
                CurrentChar >>= 1;
            }
        }
        Table[char] = CurrentChar;
    }
    return Table;
}();
```

This isn't exposed outside of the cpp file, so there shouldn't be significant compile-time overhead.

Sonicadvance1: Resolved.

skmp (approved):

PR #1544, opened 2022-02-01
Adds an option to disable the IR interpreter

Sonicadvance1:
By default we won't build with the interpeter to reduce user confusion.
The interpreter isn't really useful to end users so remove it.

Completely removes it from building except for the fallback operations.

This also removes the selection from FEXConfig to remove selection
confusion there.

```bash
File Stats:
FEXLoader Size with Interpreter:    3422768 bytes
FEXLoader Size without Interpreter: 3301944 bytes
Size difference:                    96.4699915%
Bytes removed:                      120824 bytes
4k pages removed:                   29.498046875 -> 30 rounded up

VM Stats (Reported from bloaty):
Memory Size with Interpreter:    6.50Mi
Memory Size without Interpreter: 6.38Mi
Size difference:                 98.1538462%
```


lioncash (approved):

Sonicadvance1:
Another minor bit of information. Previous stats I gathered was on x86-64 host.
Now for AArch64 host.
```bash
File Stats (from bloaty):
With Interpreter:   2.73Mi
Without Interpreter 2.63Mi
Size Difference:    96.2312477%

VM stats (from bloaty):
With Interpreter:    6.05Mi
Without Interpreter: 5.95Mi
Size Difference:     98.347107438%

PR #1550, opened 2022-02-04
OpcodeDispatcher: Fixes CRC32 decoding in 0F_38 table

Sonicadvance1:
This table is particularly terrible. CRC32 is the first instruction in
this table that needs either prefix `72h` OR `66h && F2h`

For 8bit CRC32, this ignores the 66h operand size override prefix.
  - But our table decoding didn't handle this
 
For 16bit/32bit/64bit CRC32 this behaviour changes depending on 66h
prefix AND REX.W
  - 66h prefix is ignored when REX.W is set, always 64bit but it falls
    down the other table path

This is an absolutely weird edge case that nobody should hit, but here
we are.


Sonicadvance1:
Facts. 66_F0 and 66_F1 unit tests can pass under valgrind. The other three Valgrind throws an unknown instruction encoding.

lioncash (approved):
Silly table

PR #1551, opened 2022-02-06
Some fixes for older environments

Sonicadvance1:
These are all fairly trivial. Should get us closer to building on older and non-typical environments.


Sonicadvance1:
This also resolves a bunch of build issues in a musl or termux environment.

lioncash (approved):

PR #1564, opened 2022-02-10
Container escape information

Sonicadvance1:
Adds some information that container software can use to escape the sandbox.
To be used by containers that know what they are doing with Multiarch + FEX-Emu.


smcv:
I'd say don't merge this until we've discussed whether it does what we need, even if reviewers are happy with it.

skmp (approved):

Sonicadvance1:
We can come back to this if it is still required but arch is now exposed with CPUID.
rootfs path can be gotten with execve of `FEXGetConfig`, if we need to escape to the host in a different way then we can look in to it then.

PR #1565, opened 2022-02-10
Enable proper IDE integration of thunk libraries

neobrain:
Previously, IDE engines couldn't parse the Host.cpp/Guest.cpp files of thunk libraries easily. This PR changes host libs to be built as part of the main project to enable parsing the former, and it adds a dummy copy of the guest libs project to enable parsing Guest.cpp.


Sonicadvance1 (approved):
Neato

PR #1572, opened 2022-02-11
Arm64Emitter: Allow non-optimizing LoadConstant

Sonicadvance1:
This is pulled from the code cache PR. Will be necessary for supporting
relocations.

Not currently being used but will be once we have code caching in place.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -28,20 +28,32 @@ Arm64Emitter::Arm64Emitter(FEXCore::Context::Context *ctx, size_t size) : vixl::
   SetCPUFeatures(Features);
 }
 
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -28,20 +28,32 @@ Arm64Emitter::Arm64Emitter(FEXCore::Context::Context *ctx, size_t size) : vixl::
...
   if (Is64Bit && ((~Constant)>> 16) == 0) {
     movn(Reg, (~Constant) & 0xFFFF);
+
+    if (NOPPad) {
+      nop(); nop(); nop();
+    }
     return;
   }
 
+  size_t NumMoves = 1;
```
lioncash, line 44:
```suggestion
  int NumMoves = 1;
```

Otherwise implicit sign conversions are being donein the loop below during the assignment

Sonicadvance1: Resolved.

lioncash (approved):

PR #1573, opened 2022-02-11
unittests: Disables Interpreter tests when its disabled

Sonicadvance1:
Would result in failures if you weren't expecting it.


skmp (approved):
I run into this the other day

PR #1579, opened 2022-02-15
Allow classifying syscalls with flags

Sonicadvance1:
In some cases we can generate more optimal code if we have more
information about a syscall which number gets const-propagated.

In particular optimizing through syscalls, not synchronizing state, and
never returning.

- Noreturn is used by a syscall that never returns, like exit.

This means that it never needs to try and synchronize state coming back

- Not synchronizing state and optimizing through syscalls

Useful for syscalls that don't read the state past arguments and only
returns a value.

Not all the syscalls are classified but enough to make a big difference.


External/FEXCore/Source/Interface/IR/IR.json
```diff
@@ -75,7 +75,13 @@
     "constexpr FEXCore::IR::BreakReason Break_Interrupt3         {2}",
     "constexpr FEXCore::IR::BreakReason Break_Halt               {3}",
     "constexpr FEXCore::IR::BreakReason Break_Overflow           {4}",
-    "constexpr FEXCore::IR::BreakReason Break_InvalidInstruction {5}"
+    "constexpr FEXCore::IR::BreakReason Break_InvalidInstruction {5}",
+
+    "constexpr uint8_t SYSCALL_FLAG_DEFAULT            {0}",
```
neobrain, line 80: APIs with untyped "flag" parameters are difficult to understand and follow, so these would be better suited as an `enum class`.

 For the composition operators, FWIW Mesa has a set of useful macros to autogenerate them if we don't already have something equivalent: https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/util/enum_operators.h

Source/Tests/LinuxSyscalls/x32/Syscalls.h
```diff
@@ -100,20 +102,42 @@ bool RegisterSyscall(int num, int32_t HostSyscallNumber, const char *name, F f){
...
+
+
+#define REGISTER_SYSCALL_IMPL_X32_FLAGS(name, flags, lambda) \
+  struct impl_##name { \
+    impl_##name() \
+    { \
+      FEX::HLE::x32::RegisterSyscall(x32::SYSCALL_x86_##name, ~0, flags, #name, lambda); \
+    } } impl_##name
+
+#define REGISTER_SYSCALL_IMPL_X32_PASS_FLAGS(name, flags, lambda) \
```
neobrain, line 131: I'm not a fan of the increasing number of these macros. What do they do in the first place compared to just calling `FEX::HLE::x32::RegisterSyscall` directly?

Can we implement this functionality using more unified syntax so we don't need 6 macros to do essentially the same thing?

neobrain: Maybe something like this could work? The stringification is still done in a macro, but everything else in regular functions. Call sites would become something like `SYSCALL_IMPL(mmap).RegisterSyscall(FEXCore::IR::SyscallFlags::OPTIMIZETHROUGH | FEXCore::IR::SyscallFlags::NOSYNCSTATEONENTRY, LAMBDA)`

```cpp
#define SYSCALL_IMPL(name) \
    struct {
        template<class F>
        bool RegisterSyscall(int num, int32_t HostSyscallNumber, F f){
            return FEX::HLE::x64::RegisterSyscall(x64::SYSCALL_x64_##name, SYSCALL_DEF(name), FEXCore::IR::SyscallFlags::DEFAULT, #name, std::move(f));
        }

        template<class F>
        bool RegisterSyscall(int num, int32_t HostSyscallNumber, SysCallFlags flags, F f){
            return FEX::HLE::x64::RegisterSyscall(x64::SYSCALL_x64_##name, SYSCALL_DEF(name), flags, #name, std::move(f));
        }
    } impl_##name
```

We can use the same name for member functions that differ only in strongly typed parameters - for others I'd add a suffix like you're doing with the macros already.

Sonicadvance1: How would that work in the face of `Templates cannot be declared inside of a local class [template_inside_local_class]`?

Source/Tests/LinuxSyscalls/Syscalls/Msg.cpp
```diff
@@ -17,22 +19,28 @@ tags: LinuxSyscalls|syscalls-shared
 
 namespace FEX::HLE {
   void RegisterMsg() {
-    REGISTER_SYSCALL_IMPL_PASS(msgget, [](FEXCore::Core::CpuStateFrame *Frame, key_t key, int msgflg) -> uint64_t {
+    using namespace FEXCore::IR;
+
+    REGISTER_SYSCALL_IMPL_PASS_FLAGS(msgget, FEXCore::IR::SyscallFlags::OPTIMIZETHROUGH | FEXCore::IR::SyscallFlags::NOSYNCSTATEONENTRY,
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/x32/Syscalls.h
neobrain: C++ was a mistake 😩

I'm sure there's a good way to do this, but let's worry about it another day.

neobrain (approved):

PR #1586, opened 2022-02-21
TestHarnessRunner: Wire up environment variable option setting

Sonicadvance1:
Wire up the environment variable option setting so asm files can set
these and it works

Allows some something like the following:
```json
"Env": {
  "FEX_MAXINST": "500"
}
```

Specifically this is for #1585 so we have a generic solution to setting the f64 option rather than a special case for that one.


Source/Tests/HarnessHelpers.h
```diff
@@ -161,6 +161,21 @@ namespace FEX::HarnessHelper {
...
+    std::vector<std::pair<char const*, char const*>> GetEnvironmentOptions() {
+      std::vector<std::pair<char const*, char const*>> Env{};
+
+      uintptr_t DataOffset = BaseConfig.OptionEnvOptionOffset;
+      for (unsigned i = 0; i < BaseConfig.OptionEnvOptionCount; ++i) {
+        // Environment variables are null terminated strings
+        char const *Key = reinterpret_cast<char const*>(RawConfigFile.data() + DataOffset);
+        char const *Value = reinterpret_cast<char const*>(RawConfigFile.data() + DataOffset + strlen(Key) + 1);
+        DataOffset += strlen(Key) + strlen(Value) + 2;
+        Env.emplace_back(std::make_pair(Key, Value));
```
lioncash, line 176:
```suggestion
        Env.emplace_back(Key, Value);
```

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -64,6 +64,53 @@ void AssertHandler(char const *Message) {
...
+#define OPT_BASE(type, group, enum, json, default) {"FEX_" #enum, FEXCore::Config::ConfigOption::CONFIG_##enum},
+#include <FEXCore/Config/ConfigValues.inl>
+}};
+
+// Claims to be a local application config layer
+class TestEnvLoader final : public FEXCore::Config::Layer {
+public:
+  explicit TestEnvLoader(std::vector<std::pair<char const*, char const*>> _Env)
+    : FEXCore::Config::Layer(FEXCore::Config::LayerType::LAYER_LOCAL_APP)
+    , Env {_Env} {
```
lioncash, line 78:
```suggestion
    , Env {std::move(_Env)} {
```

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -64,6 +64,53 @@ void AssertHandler(char const *Message) {
...
+    std::unordered_map<std::string_view, std::string_view> EnvMap;
+    for (auto &Option : Env) {
+      std::string_view Key = Option.first;
+      std::string_view Value = Option.second;
+
+#define ENVLOADER
+#include <FEXCore/Config/ConfigOptions.inl>
+
+      EnvMap[Key]=Value;
+    }
```
lioncash, line 92:
```suggestion
      EnvMap.insert_or_assign(Key, Value);
    }
```

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -64,6 +64,53 @@ void AssertHandler(char const *Message) {
...
+
+    std::function GetVar = [=](const std::string_view id)  -> std::optional<std::string_view> {
+      if (EnvMap.find(id) != EnvMap.end())
```
lioncash, line 95:
```suggestion
      const auto iter = EnvMap.find(id);
      if (iter == EnvMap.end()) {
          return std::nullopt;
      }
      
      return iter->second;
```

or something, so you don't need to hash into the map twice

lioncash (commented):
Just some minor nits, but looks fine otherwise

Source/Tests/HarnessHelpers.h
```diff
@@ -161,6 +161,21 @@ namespace FEX::HarnessHelper {
     void Init(std::string const &ConfigFilename) {
       ReadFile(ConfigFilename, &RawConfigFile);
       memcpy(&BaseConfig, RawConfigFile.data(), sizeof(ConfigStructBase));
+      GetEnvironmentOptions();
+    }
+
+    std::vector<std::pair<char const*, char const*>> GetEnvironmentOptions() {
```
neobrain, line 167:
```suggestion
    std::vector<std::pair<std::string_view, std::string_view>> GetEnvironmentOptions() {
```
Why not return `string_view` here already? Would clean up the implementation a fair bit, since you can avoid the `strlen()` calls then.

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -64,6 +64,53 @@ void AssertHandler(char const *Message) {
...
+      std::string_view Key = Option.first;
+      std::string_view Value = Option.second;
+
+#define ENVLOADER
+#include <FEXCore/Config/ConfigOptions.inl>
+
+      EnvMap[Key]=Value;
+    }
+
+    std::function GetVar = [=](const std::string_view id)  -> std::optional<std::string_view> {
```
neobrain, line 94:
```suggestion
    auto GetVar = [=](const std::string_view id)  -> std::optional<std::string_view> {
```

`std::function` incurs an unnecessary heap allocation here.

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -64,6 +64,53 @@ void AssertHandler(char const *Message) {
...
+      if (EnvMap.find(id) != EnvMap.end())
+        return EnvMap.at(id);
+
+      return std::nullopt;
+    };
+
+    std::optional<std::string_view> Value;
+    for (auto &it : EnvConfigLookup) {
+      if ((Value = GetVar(it.first)).has_value()) {
```
neobrain, line 103:
```suggestion
    if (auto Value = GetVar(it.first); Value) {
```

Source/Tests/HarnessHelpers.h
Sonicadvance1: Resolved.

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: Resolved.

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: Resolved.

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: Resolved.

Source/Tests/HarnessHelpers.h
Sonicadvance1: Resolved.

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: Resolved.

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: I definitely don't care for semicolons inside of if statements but resolved.

Source/Tests/HarnessHelpers.h
```diff
@@ -161,6 +161,21 @@ namespace FEX::HarnessHelper {
...
+      GetEnvironmentOptions();
+    }
+
+    std::vector<std::pair<std::string_view, std::string_view>> GetEnvironmentOptions() {
+      std::vector<std::pair<std::string_view, std::string_view>> Env{};
+
+      uintptr_t DataOffset = BaseConfig.OptionEnvOptionOffset;
+      for (unsigned i = 0; i < BaseConfig.OptionEnvOptionCount; ++i) {
+        // Environment variables are null terminated strings
+        std::string_view Key = reinterpret_cast<char const*>(RawConfigFile.data() + DataOffset);
```
neobrain, line 173: What's the `reinterpret_cast` for?

Sonicadvance1: You right. Resolved.

neobrain (approved):

PR #1588, opened 2022-02-24
Improve compatibility with older uapi kernel headers

Azkali:
Following up the work previously done in 2079f6b3c7be81fd10546ecd978d41a8d7d6f685.
Adding more defines for older Linux uapi headers missing some defines.

Compiles properly now against l4t linux v4.9.140.

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/Syscalls.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/Syscalls.h
@@ -28,6 +28,10 @@ namespace FHU::Syscalls {
 #define CLONE_PIDFD 0x00001000
 #endif
 
+#ifndef SYS_statx
+#define SYS_statx 291
+#endif
+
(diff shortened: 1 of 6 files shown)


FEXHeaderUtils/FEXHeaderUtils/Syscalls.h
```diff
@@ -28,6 +28,10 @@ namespace FHU::Syscalls {
 #define CLONE_PIDFD 0x00001000
 #endif
 
+#ifndef SYS_statx
+#define SYS_statx 291
```
Sonicadvance1, line 32: This is only viable on AArch64. This would break x86-64 host environments where statx syscall is 332.

Sonicadvance1 (commented):
Other than the one comment this seems fine.

Azkali:
Let me know if the latter commit is suiting what you want to have in FEX

Sonicadvance1 (approved):

Sonicadvance1:
That works.

PR #1591, opened 2022-02-26
Scripts: Updates CPU fitting script for latest CPUs

Sonicadvance1:
Clang doesn't yet understand the latest ARM CPUs so just document them
and set to the closest thing.

### DIFF
--- a/Scripts/aarch64_fit_native.py
+++ b/Scripts/aarch64_fit_native.py
@@ -16,6 +16,14 @@ BigCoreIDs = {
         tuple([0x41, 0xd0d]): "cortex-a77",
         tuple([0x41, 0xd41]): "cortex-a78",
         tuple([0x41, 0xd44]): "cortex-x1",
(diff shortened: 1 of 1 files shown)


Scripts/aarch64_fit_native.py
```diff
@@ -16,6 +16,14 @@
         tuple([0x41, 0xd0d]): "cortex-a77",
         tuple([0x41, 0xd41]): "cortex-a78",
         tuple([0x41, 0xd44]): "cortex-x1",
+        tuple([0x41, 0xd47]):
+            [ ["cortex-a78", "0.0"],
+              ["cortex-a710", "999.0"], # Doesn't exist in clang yet
```
neobrain, line 21: Could you add the Clang version you checked against for reference?

Sonicadvance1: Resolved.

neobrain (approved):

PR #1594, opened 2022-02-26
FEXCore: Implement F16C

Sonicadvance1:
I was bored while lounging so here we are.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -72,7 +72,7 @@ if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
   endif()
   set(_M_X86_64 1)
   add_definitions(-D_M_X86_64=1)
-  set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -mcx16")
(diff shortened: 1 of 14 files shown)


Sonicadvance1:
Wasn't complete, only implemented the 128-bit operations. 128-bit and 256-bit are mandatory for this extension.
Can't have F16C on a system without AVX.

Torinde:
> Can't have F16C on a system without AVX.

Oh, yes, I overlooked that. So, it can come only after (or together with) #581?

Do you plan to add also:
- SSE4a #3838
- AVX related: FMA3, FMA4 #3839
- XOP (#197) and related TBM, LWP?

Sonicadvance1:
> > Can't have F16C on a system without AVX.
> 
> Oh, yes, I overlooked that. So, it can come only after (or together with) #581?
> 
> Do you plan to add also:
> 
> * SSE4a

- SSE4a's EXTRQ and INSERTQ instructions don't map well to ASIMD. So it's low priority 

> * AVX related: FMA3, FMA4
- Yes, likely both

> * XOP ([Removes XOP frontend handling #197](https://github.com/FEX-Emu/FEX/pull/197)) and related TBM, LWP?
- XOP no, nothing uses it and even AMD deleted support from their CPUs. Which then extends to TBM and LWP.

PR #1595, opened 2022-02-26
CMake: Strip binaries with Release and MinSizeRel

Sonicadvance1:
Incase someone wants to tinker with MinSizeRel.
Reduces executable from 2.7MB to 2.1MB here.

I still recommend Release because the performance is definitely more
important than 0.6MB


neobrain:
Why strip binaries by default in the first place? Seems there's no strong reason for doing so (saving a few MB of data) vs not doing so (removing the last bit of debuggability in Release builds), but then I'd lean towards the option that doesn't need any extra code in the build system (which makes following the build scripts as a whole more difficult).

PR #1598, opened 2022-03-01
CPUID: Implements leaf 4000_0000

Sonicadvance1:
This region is reserved for hypervisor uses. Let's follow other examples
and return a hypervisor vendor id signature as another way for software
to find if it is running under FEX-Emu.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/External/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -812,6 +812,28 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_1Ah(uint32_t Leaf) {
   return Res;
 }
 
+// Hypervisor CPUID information leaf
+FEXCore::CPUID::FunctionResults CPUIDEmu::Function_4000_0000h(uint32_t Leaf) {
+  FEXCore::CPUID::FunctionResults Res{};
+  // Maximum supported hypervisor leafs
+  // We only expose the information leaf
+  //
+  // Common courtesy to follow VMWare's "Hypervisor CPUID Interface proposal"
+  // 4000_0000h - Information leaf. Advertising to the software which hypervisor this is
+  // 4000_0001h - 4000_000Fh - Hypervisor specific leafs. FEX can use these for anything
+  // 4000_0010h - 4000_00FFh - "Generic Leafs" - Try not to overwrite, other hypervisors might expect information in these
+  //
+  // CPUID documentation information:
+  // 4000_0000h - 4FFF_FFFFh - No existing or future CPU will return information in this range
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -812,6 +812,28 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_1Ah(uint32_t Leaf) {
...
+  // 4000_0010h - 4000_00FFh - "Generic Leafs" - Try not to overwrite, other hypervisors might expect information in these
+  //
+  // CPUID documentation information:
+  // 4000_0000h - 4FFF_FFFFh - No existing or future CPU will return information in this range
+  // Reserved entirely for VMs to do whatever they want.
+  Res.eax = 0x40000000;
+
+  // EBX, EDX, ECX become the hypervisor ID signature
+  constexpr static char HypervisorID[12] = "FEXIFEXIEMU";
+  memcpy(reinterpret_cast<char*>(&Res) + 4, HypervisorID, sizeof(HypervisorID));
```
lioncash, line 833: wouldn't `&Res.ebx` allow this to be done without the cast

lioncash (commented):
LGTM aside from one nit

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -812,6 +812,28 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_1Ah(uint32_t Leaf) {
...
+  // 4000_0010h - 4000_00FFh - "Generic Leafs" - Try not to overwrite, other hypervisors might expect information in these
+  //
+  // CPUID documentation information:
+  // 4000_0000h - 4FFF_FFFFh - No existing or future CPU will return information in this range
+  // Reserved entirely for VMs to do whatever they want.
+  Res.eax = 0x40000000;
+
+  // EBX, EDX, ECX become the hypervisor ID signature
+  constexpr static char HypervisorID[12] = "FEXIFEXIEMU";
+  memcpy(reinterpret_cast<char*>(&Res.ebx), HypervisorID, sizeof(HypervisorID));
```
lioncash, line 833: this should be able to be:

```suggestion
  memcpy(&Res.ebx, HypervisorID, sizeof(HypervisorID));
```

since the cast would be converted to that anyway before the call

lioncash (approved):

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Right...Rushed it a bit. Resolved.

PR #1601, opened 2022-03-04
IR: New IR JSON format

Sonicadvance1:
This greatly simplifies the IR format by using string parsing for gathering the information.
Tons of redundant information removed.
Significantly more difficult to mess up adding a new IR op.

Significantly improves the generator functions in IREmitter

Additionally, this enables MORE IR validation when assertions are enabled.
Now that we have SSA class typing in the IR definition, we can check that the incoming SSA value is the expected class.
This has already found bugs in `F80CVTToInt` and should resolve programmer mistakes while writing in the IR.

Additionally if we need more manual validation then we support a new `EmitValidation` section of IR ops that allow us to generate assert style checking right next to the IR definition.
All of this checking happens in the IREmitter function itself, so you get exact placement in code when validation fails, unlike the other IR validation passes that run after emission.

This also cuts out the IRParser's specialized IR parsing functions, since they can now logically use the generator's emitters. Nice cleanup there.

A BIG change is that the IREmitter function now generates a function declaration in the exact order of string parsing, rather than always the order of `<SSA Arguments> <User typed args> <Helper Args>`, which means that almost all of our IREmitter helper duplicate functions just get deleted.

This makes the `IR.md` documentation actually worth looking at, since the arguments will be in the direct order from the string.

This has been a long time coming and it removes such a large number of footgunning with interfacing with the IREmitter that it was totally worth it.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -435,9 +435,10 @@ void OpDispatchBuilder::MOVQOp(OpcodeArgs) {
   // This instruction is a bit special that if the destination is a register then it'll ZEXT the 64bit source to 128bit
   if (Op->Dest.IsGPR()) {
     const auto gpr = Op->Dest.Data.GPR.GPR;
-    _StoreContext(FPRClass, 8, offsetof(FEXCore::Core::CPUState, xmm[gpr - FEXCore::X86State::REG_XMM_0][0]), Src);
+
+    _StoreContext(8, FPRClass, Src, offsetof(FEXCore::Core::CPUState, xmm[gpr - FEXCore::X86State::REG_XMM_0][0]));
     auto Const = _Constant(0);
-    _StoreContext(GPRClass, 8, offsetof(FEXCore::Core::CPUState, xmm[gpr - FEXCore::X86State::REG_XMM_0][1]), Const);
+    _StoreContext(8, GPRClass, Const, offsetof(FEXCore::Core::CPUState, xmm[gpr - FEXCore::X86State::REG_XMM_0][1]));
   }
   else {
     // This is simple, just store the result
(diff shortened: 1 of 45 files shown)


External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -435,9 +435,10 @@ void OpDispatchBuilder::MOVQOp(OpcodeArgs) {
   // This instruction is a bit special that if the destination is a register then it'll ZEXT the 64bit source to 128bit
   if (Op->Dest.IsGPR()) {
     const auto gpr = Op->Dest.Data.GPR.GPR;
-    _StoreContext(FPRClass, 8, offsetof(FEXCore::Core::CPUState, xmm[gpr - FEXCore::X86State::REG_XMM_0][0]), Src);
```
Sonicadvance1, line 438: Unrelated to this PR but we should probably modify how this code behaves in the future.
We have vector ops that can move a register with zext, we should just use that zext 64-bit to 128bit instead of the nasty here.

After this PR though, I don't want to change op behaviour if possible.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -301,18 +301,29 @@ void ConstProp::LoadMemStoreMemImmediatePooling(IREmitter *IREmit, const IRListV
...
+          OffsetIndex = IR::IROp_LoadMem::Offset_Index;
+        }
+        else {
+          AddrIndex = IR::IROp_StoreMem::Addr_Index;
+          OffsetIndex = IR::IROp_StoreMem::Offset_Index;
+        }
         uint64_t Addr;
 
-        if (IREmit->IsValueConstant(IROp->Args[0], &Addr) && IROp->Args[1].IsInvalid()) {
+        if (IREmit->IsValueConstant(IROp->Args[AddrIndex], &Addr) && IROp->Args[OffsetIndex].IsInvalid()) {
```
Sonicadvance1, line 317: This line actually fixes a bug in the case of `IROp->Op == OP_STOREMEM` here.
Args[1] in LoadMem was Offset, while Args[1] in StoreMem was Value, so in the StoreMem case it was never `IsInvalid()`.

This is now resolved and StoreMem operations will now optimize with an immediate offset rather than never optimizing.

PR #1610, opened 2022-03-08
FileManager: Fix realpath failed on debian buster

wannacu:
This happend on debian buster when run realpath(i386) on arm64 host.

### DIFF
--- a/Source/Tests/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tests/LinuxSyscalls/FileManagement.cpp
@@ -387,7 +387,7 @@ uint64_t FileManager::Lstat(const char *pathname, void *buf) {
       return Result;
   }
 
-  return ::lstat(SelfPath, reinterpret_cast<struct stat*>(buf));
+  return ::lstat(pathname, reinterpret_cast<struct stat*>(buf));
 }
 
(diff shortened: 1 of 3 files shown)


Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -387,7 +387,7 @@ uint64_t FileManager::Lstat(const char *pathname, void *buf) {
       return Result;
   }
 
-  return ::lstat(SelfPath, reinterpret_cast<struct stat*>(buf));
+  return ::lstat(pathname, reinterpret_cast<struct stat*>(buf));
```
Sonicadvance1, line 390: I'll need to take a look at this.

Sonicadvance1:
Okay. I see why this works now.
lstat doesn't follow symlinks, which means realpath to `/proc/self/exe` will get information about the symlink itself rather than the final target.

In lpath case with symlinks, we want to return the symlink data, not the data that we are pointing to (which would be the target executable). Returning the target executable information causes realpath to stop its work early, thus returning that the real path is `/proc/<pid>/exe` rather than the real executable.
After lstat the program uses readlinkat to follow the symlink manually and repeats the process.

This PR is correct here.

Sonicadvance1 (approved):

PR #1611, opened 2022-03-09
WIP: Thunks: Add ABI analysis pass, struct repacking, and pointer handling

neobrain:
**STRONGLY WIP:** Uploading for design review only.

This adds an ABI analysis pass to the generator, i.e. a framework for storing information on how to marshal function argument data between guest/host before and after a thunk transition. This allows to identify compatibility problems early, e.g. when struct layouts differ across architectures. Simultaneously, it opens the door to resolve some of these problems by enabling automatic repacking of struct members.

Integrating this framework requires us to refine our existing set of library annotations: Whenever pointers are passed to (or returned from) a function, an annotation is needed to tell the generator how to proceed. This is less laborious than it sounds as annotations can be assigned to types and struct members as well now. The process will be documented in more detail in a later iteration of this PR.

Extensive tests have been added to cover the new functionality, all of which (minus 3 noncritical ones) pass. Until the existing ThunkLibs have been adapted to work, the unit tests also serve as reference for the new annotations.

TODO list (incomplete):
* [ ] Apply new annotations to existing library interface definitions
* [ ] Add annotations for pointers to null-terminated arrays/strings
* [ ] Drop unrelated patches
* [ ] Tidy up and restructure documentation
* [ ] Add union support (required for `xcb_randr_notify_data_next`)
* [ ] Review impact of `long` on ABI compatibility detection
* [ ] "Stable ABI" parameter annotation for untyped pointers to raw data (e.g. for libGL)

TODO list for followup PRs:
* [ ] Any union of trivially compatible types is itself trivially compatible (e.g. `snd_seq_timestamp_t`)
* [ ] Apply ABI detection to returned pointer types


Sonicadvance1:
Took a spin through reading this. Quite dense but I'm liking it so far.

neobrain:
The space of requirements is well-explored now, so it's safe to push a few things out of this PR to reduce scope:
* The new checks aren't applied to return types, but clearly should follow the same rules as function parameters
* ABI analysis of unions will be implemented later; for now, the TODO64 parameter-annotation or a "trust me, this type has consistent ABI" type-annotation are used instead
* Auto-detection of structs that need repacking stays in this PR, but the actual repacking may be split to a followup

neobrain:
Very pleased to say most of this has landed in main now, in a much more coherent and organized shape.

PR #1612, opened 2022-03-09
JITs: Emit identification string in the code buffers

Sonicadvance1:
At the start of each code buffer, emit a small string for letting memory
inspection know if a code region is for the JITs.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
@@ -330,6 +330,7 @@ X86JITCore::X86JITCore(FEXCore::Context::Context *ctx, FEXCore::Core::InternalTh
   }
 
   CurrentCodeBuffer = &InitialCodeBuffer;
+  EmitDetectionString();
 
   RAPass = Thread->PassManager->GetPass<IR::RegisterAllocationPass>("RA");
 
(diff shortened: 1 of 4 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -406,6 +407,13 @@ X86JITCore::~X86JITCore() {
   FreeCodeBuffer(InitialCodeBuffer);
 }
 
+void X86JITCore::EmitDetectionString() {
+  const char JITString[] = "FEXJIT::X86JITCore::";
+  for (size_t i = 0; i < sizeof(JITString); ++i) {
```
lioncash, line 412: This can probably compress to:

```suggestion
  for (char c : JITString) {
```

Sonicadvance1: Resolved.

lioncash (approved):

External/FEXCore/Source/Interface/Core/JIT/x86_64/JITClass.h
```diff
@@ -152,6 +152,8 @@ class X86JITCore final : public CPUBackend, public Xbyak::CodeGenerator {
 
   static uint64_t ExitFunctionLink(X86JITCore* code, FEXCore::Core::CpuStateFrame *Frame, uint64_t *record);
 
+  void EmitDetectionString();
```
neobrain, line 155: Do you mind adding a comment in the code as well as to why this function exists? It looks a bit out-of-place without context, and it's hard to see that it's merely a debugging aid (?) rather than something that's being relied on upon elsewhere.

PR #1613, opened 2022-03-09
GDBServer improvements

Sonicadvance1:
Gets this more usable. Still not super nice in regards to hanging with signals getting thrown around but getting significantly closer.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/GdbServer.cpp
+++ b/External/FEXCore/Source/Interface/Core/GdbServer.cpp
@@ -871,7 +871,8 @@ GdbServer::HandledPacketType GdbServer::handleV(const std::string& packet) {
     const auto F       = [](int result) { return fmt::format("F{:x}", result); };
     const auto F_error = [] { return fmt::format("F-1,{:x}", errno); };
     const auto F_data  = [](int result, const std::string& data) {
-        return fmt::format("F{:x};{}", result, data);
+      // Binary encoded data is raw appended to the end
+      return fmt::format("F{:#x};", result) + data;
     };
 
     std::optional<std::istringstream> ss;
@@ -916,6 +917,11 @@ GdbServer::HandledPacketType GdbServer::handleV(const std::string& packet) {
         if (ret < 0) {
             return {F_error(), HandledPacketType::TYPE_ACK};
         }
+
+        if (ret == 0) {
+          return {F(0), HandledPacketType::TYPE_ACK};
+        }
+
         data.resize(ret);
         return {F_data(ret, data), HandledPacketType::TYPE_ACK};
     }
--- a/External/FEXCore/Source/Interface/Core/GdbServer.cpp
+++ b/External/FEXCore/Source/Interface/Core/GdbServer.cpp
@@ -17,12 +17,14 @@ $end_info$
 #include "Interface/Context/Context.h"
 
 #include <FEXCore/Config/Config.h>
+#include <FEXCore/Core/CodeLoader.h>
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Core/CoreState.h>
 #include <FEXCore/Core/SignalDelegator.h>
 #include <FEXCore/Core/X86Enums.h>
 #include <FEXCore/Debug/InternalThreadState.h>
 #include <FEXCore/HLE/Linux/ThreadManagement.h>
+#include <FEXCore/HLE/SyscallHandler.h>
 #include <FEXCore/Utils/CompilerDefs.h>
 #include <FEXCore/Utils/NetStream.h>
 #include <FEXCore/Utils/LogManager.h>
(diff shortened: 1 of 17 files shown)


External/FEXCore/include/FEXCore/Core/SignalDelegator.h
```diff
@@ -96,14 +96,14 @@ namespace Core {
...
-      FEXCore::HostSignalDelegatorFunction Handler{};
+      std::vector<FEXCore::HostSignalDelegatorFunction> Handlers{};
       FEXCore::HostSignalDelegatorFunction FrontendHandler{};
     };
     std::array<HostSignalHandler, MAX_SIGNALS + 1> HostHandlers{};
 
   protected:
     void SetHostSignalHandler(int Signal, HostSignalDelegatorFunction Func, bool Required) {
-      HostHandlers[Signal].Handler = std::move(Func);
+      HostHandlers[Signal].Handlers.emplace_back(std::move(Func));
```
neobrain, line 106: FYI `emplace_back(std::move(...))` is a bit of an anti-pattern: It does not improve performance since there's already an overload for `push_back` that takes rvalues, but `emplace_back` produces unreadable error messages when something goes wrong. (Similarly, using `emplace_back` does not compensate for an accidentally missing `std::move`!)

See https://quuxplusone.github.io/blog/2021/03/03/push-back-emplace-back/#when-all-else-is-equal-prefer-push_back-to-emplace_back for details if you're interested :)

External/FEXCore/Source/Interface/Core/GdbServer.h
```diff
@@ -77,6 +77,7 @@ class GdbServer {
     std::string MemoryMapString{};
     std::string OSDataString{};
 
+    std::array<bool, SignalDelegator::MAX_SIGNALS + 1> PassSignals{};
```
neobrain, line 80: Could you document what this member is used for?

neobrain, line 80: What is the extra slot used for? (I.e. why not just `MAX_SIGNALS` elements?)

External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -870,8 +870,13 @@ GdbServer::HandledPacketType GdbServer::handleV(const std::string& packet) {
 
     const auto F       = [](int result) { return fmt::format("F{:x}", result); };
     const auto F_error = [] { return fmt::format("F-1,{:x}", errno); };
-    const auto F_data  = [](int result, const std::string& data) {
-        return fmt::format("F{:x};{}", result, data);
+    const auto F_data  = [](int result, const uint8_t *data, size_t size) {
+      // Binary encoded data is raw appended to the end
+      std::string String = fmt::format("F{:x};", result);
```
neobrain: Why didn't the previous code work? Wouldn't at least `fmt::format("F{:#x};", result) + data` have been equivalent?

External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -656,22 +656,10 @@ static size_t CheckMemMapping(uint64_t Address, size_t Size) {
...
-        auto str = fmt::format("Text={:x};Data={:x};Bss={:x}", Begin, Begin, Begin);
-        return {std::move(str), HandledPacketType::TYPE_ACK};
-      }
-    }
-  }
-
-  return {"Text=0;Data=0;Bss=0", HandledPacketType::TYPE_ACK};
+  auto CodeLoader = CTX->SyscallHandler->GetCodeLoader();
+  uint64_t BaseOffset = CodeLoader->GetBaseOffset();
+  auto str = fmt::format("Text={:x};Data={:x};Bss={:x}", BaseOffset, BaseOffset, BaseOffset);
```
neobrain, line 673: Wouldn't returning `BaseOffset + TextSize` be a more appropriate value for Data and Bss? (I'm assuming in our JIT code, we essentially roll up all sections into one, so there's no distinction to be made anyway. But debuggers could be confused if a text section where instantly delimited by the data section)

External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -611,6 +613,21 @@ GdbServer::HandledPacketType GdbServer::handleXfer(const std::string &packet) {
...
 
+    if (object == "auxv") {
+      auto CodeLoader = CTX->SyscallHandler->GetCodeLoader();
+      uint64_t auxv_ptr, auxv_size;
+      CodeLoader->GetAuxv(auxv_ptr, auxv_size);
+      std::string data;
+      data.resize(auxv_size * 16);
+
+      for (size_t i = 0; i < auxv_size; ++i) {
+        uint64_t *auxv = reinterpret_cast<uint64_t*>(auxv_ptr + i * 16);
```
neobrain: Looks funky. Could you add a copy explaining what data is being copied here and where the magic value 16 comes from?

External/FEXCore/include/FEXCore/Core/SignalDelegator.h
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/GdbServer.h
Sonicadvance1: Added a comment

Sonicadvance1: This is a pattern used in our signal delegation at the moment.
I created an issue to get this resolved.
This is due to signals being from [1,64] inclusive with 0 being special cased.

Happened since I wasn't expecting 0 to be special cased and now their is a maintenance burden. Created a issue so we can track this and get it resolved after this.

External/FEXCore/Source/Interface/Core/GdbServer.cpp
Sonicadvance1: You're right. Resolved.

External/FEXCore/Source/Interface/Core/GdbServer.cpp
Sonicadvance1: Changed how this is handled and also fixed an issue where 32-bit wasn't passing auxv correctly.

External/FEXCore/Source/Interface/Core/GdbServer.cpp
Sonicadvance1: No idea, we don't track that data and gdb seems fine with it so far.

PR #1614, opened 2022-03-10
JIT: Adds comment to EmitDetectionString

Sonicadvance1:


External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
```diff
@@ -184,6 +184,7 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter  {
 
   CompilerSharedData ThreadSharedData;
 
+  // This is purely a debugging aid for developers to see if they are in JIT code space
```
neobrain, line 187:
```suggestion
  // This is purely a debugging aid for developers to see if they are in JIT code space when inspecting raw memory
```

neobrain (approved):

External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
Sonicadvance1: Resolved.

PR #1622, opened 2022-03-13
Fixes #1615

### LINKED ISSUE #1615: What is the extra slot used for? (I.e. why not just `MAX_SIGNALS` elements?)
What is the extra slot used for? (I.e. why not just `MAX_SIGNALS` elements?)

_Originally posted by @neobrain in https://github.com/FEX-Emu/FEX/pull/1613#discussion_r823493879_

This is an antipattern seen through our signal handling due to signals using 1 based indexing with 0 being special cased.
Resolve this pattern and come back to it.

Sonicadvance1:
No longer special casing signal slot 0 and instead offset everything by
1.

Signals are in the range of [1,64] inclusive and we were special casing
zero and defining an extra slot everywhere.
Now we just offset the slot lookup, which makes it more similar to the
uint64_t signal mask.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/GdbServer.cpp
+++ b/External/FEXCore/Source/Interface/Core/GdbServer.cpp
@@ -80,9 +80,9 @@ GdbServer::GdbServer(FEXCore::Context::Context *ctx) : CTX(ctx) {
 
   // This is a total hack as there is currently no way to resume once hitting a segfault
   // But it's semi-useful for debugging.
-  for (uint32_t Signal = 0; Signal <= SignalDelegator::MAX_SIGNALS; ++Signal) {
+  for (uint32_t Signal = 0; Signal < SignalDelegator::MAX_SIGNALS; ++Signal) {
     ctx->SignalDelegation->RegisterHostSignalHandler(Signal, [this] (FEXCore::Core::InternalThreadState *Thread, int Signal, void *info, void *ucontext) {
-      if (PassSignals[Signal]) {
+      if (PassSignals[Signal - 1]) {
         // Pass signal to the guest
         return false;
       }
@@ -937,7 +937,7 @@ GdbServer::HandledPacketType GdbServer::handleQuery(const std::string &packet) {
     for (std::string tmp; std::getline(ss, tmp, ';'); ) {
       uint32_t Signal = std::stoi(tmp, nullptr, 16);
       if (Signal < SignalDelegator::MAX_SIGNALS) {
-        PassSignals[Signal] = true;
+        PassSignals[Signal - 1] = true;
       }
     }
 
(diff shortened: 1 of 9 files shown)


External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -80,9 +80,9 @@ GdbServer::GdbServer(FEXCore::Context::Context *ctx) : CTX(ctx) {
 
   // This is a total hack as there is currently no way to resume once hitting a segfault
   // But it's semi-useful for debugging.
-  for (uint32_t Signal = 0; Signal <= SignalDelegator::MAX_SIGNALS; ++Signal) {
+  for (uint32_t Signal = 0; Signal < SignalDelegator::MAX_SIGNALS; ++Signal) {
     ctx->SignalDelegation->RegisterHostSignalHandler(Signal, [this] (FEXCore::Core::InternalThreadState *Thread, int Signal, void *info, void *ucontext) {
-      if (PassSignals[Signal]) {
+      if (PassSignals[Signal - 1]) {
```
neobrain, line 85: The variable shadowing of `Signal` is rather confusing here (it looks like this line would index using `-1` on first iteration, but it isn't doing that since the lambda introduces a shadowed `Signal` variable that starts at 1)

I'd suggest running the loop from 1 to `MAX_SIGNALS` (inclusive) instead, so that all `Signal` variables have the same semantics.

Source/Tests/LinuxSyscalls/SignalDelegator.h
```diff
@@ -103,7 +103,7 @@ namespace FEX::HLE {
       DefaultBehaviour DefaultBehaviour {DEFAULT_TERM};
     };
 
-    std::array<SignalHandler, MAX_SIGNALS + 1> HostHandlers{};
+    std::array<SignalHandler, MAX_SIGNALS> HostHandlers{};
```
neobrain, line 106: Could you add a docstring indicating that the indexes into this array are POSIX signals offset by -1?

neobrain:
I know I've triggered this change in the first place, but now that I understand where this came from, I think the previous approach is actually better. This change adds a lot of subtle `-1`/`+1` changes that would be very easy to mess up when modifying this code at a later point. If we didn't have the existing code to fall back to, I'd probably wrap signal indexes in a strong typedef at least to make sure it's clear which interfaces use 0-based indexes and which ones start at 1 (which is easier said than done since `std::array` doesn't work with non-integer indexes of course).

The existing code isn't all that confusing if:
1. A docstring is added to PassSignals and similar arrays, explaining why there's a dummy element
2. The iteration pattern `for (int i = 1; i <= MAX_SIGNALS; ++i)` is abstracted away (making it impossible to get the start/end wrong)

One way to achieve this is to use a custom container instead of `std::array<..., MAX_SIGNALS+1>`:
```cpp
template<typename T>
class FooFoo : std::array<T, MAX_SIGNALS+1> {
    static auto indexes() const {
        // Skip the dummy element at index 0
        return std::ranges::iota_view(1, MAX_SIGNALS + 1);
    }

    auto begin() {
        return array::begin() + 1;
    }
};
```

PR #1624, opened 2022-03-14
FEXCore: Delete IR after it is used

Sonicadvance1:
For the JIT cores we don't need to keep IR around, it's only necessary
for the Interpreter. So once the AOT IR service is done dealing with the
IR, check to see if we can delete it.

This causes teeworld's title screen memory usage to go from 730MB to
566MB. 77.5% the memory usage there.

This is effectively an infinite memory leak if the codespace wasn't ever
overwritten or invalidated. So larger memory usage programs would end up
having a larger impact.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/AOTIR.cpp
+++ b/External/FEXCore/Source/Interface/IR/AOTIR.cpp
@@ -349,9 +349,17 @@ namespace FEXCore::IR {
 
       // Insert to caches if we generated IR
       if (GeneratedIR) {
-        // Add to thread local ir cache
-        Core::LocalIREntry Entry = {StartAddr, Length, decltype(Entry.IR)(IRList), decltype(Entry.RAData)(RAData), decltype(Entry.DebugData)(DebugData)};
-        Thread->LocalIRCache.insert({GuestRIP, std::move(Entry)});
+        if (Thread->CPUBackend->NeedsRetainedIRCopy()) {
+          // Add to thread local ir cache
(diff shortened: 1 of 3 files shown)


lioncash (approved):

External/FEXCore/Source/Interface/IR/AOTIR.cpp
```diff
@@ -349,9 +349,17 @@ namespace FEXCore::IR {
...
-        Core::LocalIREntry Entry = {StartAddr, Length, decltype(Entry.IR)(IRList), decltype(Entry.RAData)(RAData), decltype(Entry.DebugData)(DebugData)};
-        Thread->LocalIRCache.insert({GuestRIP, std::move(Entry)});
+        if (Thread->CPUBackend->NeedsRetainedIRCopy()) {
+          // Add to thread local ir cache
+          Core::LocalIREntry Entry = {StartAddr, Length, decltype(Entry.IR)(IRList), decltype(Entry.RAData)(RAData), decltype(Entry.DebugData)(DebugData)};
+          Thread->LocalIRCache.insert({GuestRIP, std::move(Entry)});
+        }
+        else {
+          // If the IR doesn't need to be retained then we can just delete it now
+          delete DebugData;
```
neobrain, line 359: This leaves a couple of valid-looking but dangling pointers in the caller's scope - is there no cleaner way to manage pointer ownership here? Maybe freeing the pointers needs to be delegated to the parent instead, or they should be passed to this function as `std::unique_ptr`s?

Issue #1625, opened 2022-03-14
Support IR and RAData thread pool allocation


Sonicadvance1:
Support allocating this out of a reuse pool.
This way threads that aren't actively compiling code aren't causing memory overhead.
Have them relinquish their resources to the pool after using them. If the all resources are consumed then allocate more objects in the pool.
Free pool resources in an LRU manner once they have reached some threshold.
This will reduce the memory usage overhead of FEX substantially. Currently our per thread memory allocations are quite heavy, so something like Steam consumes 1.3GB idling and half of that is due to IR/RAData staging buffers that aren't used since most threads are just sitting in pselect.

Current known areas to throw in a pool:
* OpcodeDispatcher: 8MB + 8MB
* IRCompaction: 8MB + 8MB
* Frontend: Instruction decode buffer: 6.5MB (sizeof(DecodeInst) * 0x10000)
* RAData: A few dozen megabytes per thread? Will need to check again.

Sonicadvance1:
Looks like RAData by default is ~4.2MB/thread

PR #1627, opened 2022-03-15
FEXCore: Reclaimable thread pool allocator

Sonicadvance1:
Creates a pool allocator for OpcodeDispatcher and IRCompaction that
shares memory allocations between threads in a pool and supports
reclaiming stale allocations from participating threads.

A thread will use a heuristic to keep its claimed memory allocation
around if it is allocating a lot of code. If it slows down then it will
start putting the memory allocation back in to the thread pool.

Additionally if the allocation has been "disowned" and gone to sleep
while still retaining the allocation, then another thread can inspect
 these stale allocations and reclaim it from the idling thread. Saving
further memory.

This needs some more work and cleanup but this is an interesting concept
that saves a decent amount of memory even in a basic test.

Causes teeworlds' title screen to go from 754MB to 599MB in my simple
test. 79.4% the memory usage is a good start.

First step for #1625

### DIFF
--- a/External/FEXCore/include/FEXCore/IR/IREmitter.h
+++ b/External/FEXCore/include/FEXCore/IR/IREmitter.h
@@ -19,9 +19,18 @@ friend class FEXCore::IR::Pass;
 friend class FEXCore::IR::PassManager;
 
   public:
-    IREmitter()
-      : DualListData {8 * 1024 * 1024} {
+    IREmitter(FEXCore::Utils::IntrusivePooledAllocator &ThreadAllocator)
+      : DualListData {ThreadAllocator, 8 * 1024 * 1024} {
+      ReownOrClaimBuffer();
       ResetWorkingList();
+    }
+
+    void ReownOrClaimBuffer() {
+      DualListData.ReownOrClaimBuffer();
+    }
+
+    void DelayedDisownBuffer() {
+      DualListData.DelayedDisownBuffer();
     }
 
     IRListView ViewIR() { return IRListView(&DualListData, false); }
@@ -343,7 +352,7 @@ friend class FEXCore::IR::PassManager;
     OrderedNode *CurrentWriteCursor = nullptr;
 
     // These could be combined with a little bit of work to be more efficient with memory usage. Isn't a big deal
-    DualIntrusiveAllocator DualListData;
+    DualIntrusiveAllocatorThreadPool DualListData;
 
     OrderedNode *InvalidNode;
     OrderedNode *CurrentCodeBlock{};
--- a/External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
+++ b/External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
@@ -3,6 +3,7 @@
 #include "FEXCore/IR/IR.h"
 #include <FEXCore/Utils/Allocator.h>
 #include <FEXCore/Utils/LogManager.h>
+#include <FEXCore/Utils/ThreadPoolAllocator.h>
 
 #include <cassert>
 #include <cstddef>
@@ -20,21 +21,8 @@ namespace FEXCore::IR {
  *
  * Can potentially support reallocation if we are smart and make sure to invalidate anything holding a true pointer
  */
-class DualIntrusiveAllocator final {
+class DualIntrusiveAllocator {
   public:
-    DualIntrusiveAllocator() = delete;
-    DualIntrusiveAllocator(DualIntrusiveAllocator &&) = delete;
-    DualIntrusiveAllocator(size_t Size)
-      : MemorySize {Size} {
-      Data = reinterpret_cast<uintptr_t>(FEXCore::Allocator::malloc(Size * 2));
-      List = reinterpret_cast<uintptr_t>(Data + Size);
-    }
-
-
-    ~DualIntrusiveAllocator() {
-      FEXCore::Allocator::free(reinterpret_cast<void*>(Data));
-    }
-
     [[nodiscard]] bool DataCheckSize(size_t Size) const {
       size_t NewOffset = DataCurrentOffset + Size;
       return NewOffset <= MemorySize;
@@ -81,7 +69,11 @@ class DualIntrusiveAllocator final {
       memcpy(reinterpret_cast<void*>(List), reinterpret_cast<void*>(rhs.List), ListCurrentOffset);
     }
 
-  private:
+  protected:
+    DualIntrusiveAllocator(size_t Size)
+      : MemorySize {Size} {
+    }
+
     uintptr_t Data;
     uintptr_t List;
     size_t DataCurrentOffset {0};
@@ -89,6 +81,44 @@ class DualIntrusiveAllocator final {
     size_t MemorySize;
 };
 
+class DualIntrusiveAllocatorMalloc final : public DualIntrusiveAllocator {
+  public:
+    DualIntrusiveAllocatorMalloc(size_t Size)
+      : DualIntrusiveAllocator {Size} {
+      Data = reinterpret_cast<uintptr_t>(FEXCore::Allocator::malloc(Size * 2));
+      List = reinterpret_cast<uintptr_t>(Data + Size);
+    }
+
+    ~DualIntrusiveAllocatorMalloc() {
+      FEXCore::Allocator::free(reinterpret_cast<void*>(Data));
+    }
+};
+
+class DualIntrusiveAllocatorThreadPool final : public DualIntrusiveAllocator {
+  public:
+    DualIntrusiveAllocatorThreadPool(FEXCore::Utils::IntrusivePooledAllocator &ThreadAllocator, size_t Size)
+      : DualIntrusiveAllocator {Size}
+      , PoolObject{ThreadAllocator, Size * 2} {
+      // Claim a buffer on allocation
+      PoolObject.ReownOrClaimBuffer();
+    }
+
+    ~DualIntrusiveAllocatorThreadPool() {
+      PoolObject.UnclaimBuffer();
+    }
+
+    void ReownOrClaimBuffer() {
+      Data = PoolObject.ReownOrClaimBuffer();
+      List = Data + MemorySize;
+    }
+
+    void DelayedDisownBuffer() {
+      PoolObject.DelayedDisownBuffer();
+    }
+
+  private:
+    Utils::FixedSizePooledAllocation<uintptr_t, 5000, 500> PoolObject;
+};
 
 class IRListView final {
   enum Flags {
--- a/External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
+++ b/External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
@@ -1,0 +1,502 @@
+#pragma once
+
+#include <atomic>
+#include <FEXCore/Utils/Allocator.h>
+#include <FEXCore/Utils/LogManager.h>
+
+#include <chrono>
+#include <cstddef>
+#include <list>
+#include <mutex>
+#include <sys/mman.h>
+
+namespace FEXCore::Utils {
+  /**
+   * @brief An intrusive thread pool allocator
+   *
+   * Requires coordination between the allocator and its clients to efficiently share memory allocations between threads.
+   *
+   * The `Client` in this case referring to the location in code allocating a `MemoryBuffer` from the allocator.
+   *   - The client must `Claim` a buffer to allocate it
+   *   - In claiming a buffer, the allocator is passed a `BufferOwnedFlag` that is updated by both the allocator and client.
+   *   - When the client is done with the buffer it must `Disown` or `Unclaim` the buffer.
+   *     - `Disown` the buffer when it is expected to be used again soon.
+   *       - This is relatively cheap.
+   *     - `Unclaim` when the buffer won't be used again for an extended period.
+   *       - This is expensive and requires a mutex shared between threads
+   *     - `FixedSizePooledAllocation` helper class provided to help with this.
+   *
+   * Once the client has disowned a buffer then the allocator is free to reclaim the buffer when another thread is trying to `Claim` a new buffer.
+   * The buffer getting claimed from a disowned client must have had its last use greater than the defined `DURATION` before it has a chance to get
+   * reclaimed by the Allocator.
+   *
+   * During buffer reclaiming is also when unclaimed buffers get freed. This means active threads are able to clean up idle thread's unused memory.
+   */
+  class IntrusivePooledAllocator {
+    public:
+      template<typename T>
+      struct AllocationInfo {
+        T Ptr;
+        size_t Size;
+      };
+
+      struct MemoryBuffer;
+      /**
+       * @brief Container for tracking the buffers
+       *
+       * We're using std::list explicitly because its iterators aren't invalidated when the list is adjusted.
+       * if we had list types that we can atomically erase and append elements then unclaiming could be made cheaper.
+       */
+      using ContainerType = std::list<MemoryBuffer*>;
+      /**
+       * @brief steady_clock to ensure long running applications don't hit any timeskip problems.
+       */
+      using ClockType = std::chrono::steady_clock;
+      /**
+       * @brief Atomic flag state for letting the client know if it owns the buffer
+       */
+      enum class ClientFlags : uint32_t {
+        FLAG_FREE = 0,
+        FLAG_OWNED = 1,
+        FLAG_DISOWNED = 3,
+      };
+
+      using BufferOwnedFlag = std::atomic<ClientFlags>;
+
+      struct MemoryBuffer {
+        void* Ptr;
+        size_t Size;
+        std::atomic<std::chrono::time_point<ClockType>> LastUsed;
+        BufferOwnedFlag *CurrentClientOwnedFlag{};
+      };
+      // Ensure that the atomic objects of MemoryBuffer are lock free
+      static_assert(decltype(MemoryBuffer::LastUsed){}.is_always_lock_free, "Oops, needs to be lock free");
+      static_assert(std::remove_pointer<decltype(MemoryBuffer::CurrentClientOwnedFlag)>::type{}.is_always_lock_free, "Oops, needs to be lock free");
+
+      /**
+       * @brief Lets the client easily check if they own the buffer or not
+       *
+       * @param CurrentClientFlag Client owned flag
+       *
+       * @return Is the client buffer owned at the point of checking
+       */
+      static bool IsClientBufferOwned(BufferOwnedFlag &CurrentClientFlag) {
+        return CurrentClientFlag.load() == ClientFlags::FLAG_OWNED;
+      }
+
+      /**
+       * @brief Lets the client easily check if the buffer was freed
+       *
+       * @param CurrentClientFlag Client owned flag
+       *
+       * @return Is the client buffer owned at the point of checking
+       */
+      static bool IsClientBufferFree(BufferOwnedFlag &CurrentClientFlag) {
+        return CurrentClientFlag.load() == ClientFlags::FLAG_FREE;
+      }
+
+      /**
+       * @brief Allocates and claims a buffer that is tracked from the thread pool
+       *
+       * @param Size
+       * @param CurrentClientFlag
+       *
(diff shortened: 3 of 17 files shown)


External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,160 @@
...
+
+      struct MemoryBuffer {
+        void* Ptr;
+        size_t Size;
+        std::chrono::time_point<ClockType> LastUsed;
+        std::atomic<bool> ClientClaimed;
+        std::atomic<bool> *CurrentClientOwnedFlag{};
+      };
+
+      using ContainerType = std::list<MemoryBuffer*>;
```
neobrain, line 30: Would `std::vector` work here? Looks like we're mostly just iterating over these (where `vector` provides better cache locality) and insertion/erasure is exceptional.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,160 @@
...
+      void DisownBuffer(ContainerType::iterator Buffer) {
+        // Client still owns the buffer but isn't using it
+        // Allows us to claim it back if necessary
+        (*Buffer)->LastUsed = ClockType::now();
+        (*Buffer)->ClientClaimed.store(false);
+      }
+
+      ~ThreadPoolAllocator() {
+        // Cleanup any buffers that are remaining
+        for (auto it = Buffers.begin(); it != Buffers.end(); ++it) {
```
neobrain, line 54: range-based for loops can be used here and in a couple of other places.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
```diff
@@ -20,19 +21,12 @@ namespace FEXCore::IR {
  *
  * Can potentially support reallocation if we are smart and make sure to invalidate anything holding a true pointer
  */
-class DualIntrusiveAllocator final {
+class DualIntrusiveAllocator {
```
neobrain, line 24: Should this have a virtual destructor? If not, we should use non-public inheritance instead so people can't accidentally bind child objects to `DualIntrusiveAllocator*` pointers.

External/FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -19,11 +19,37 @@ friend class FEXCore::IR::Pass;
...
+    IREmitter(FEXCore::Utils::ThreadPoolAllocator &ThreadAllocator)
+      : DualListData {ThreadAllocator, 8 * 1024 * 1024} {
       ResetWorkingList();
     }
 
+    void ClaimBuffer() {
+      DualListData.ClaimBuffer();
+    }
+    void UnclaimBuffer() {
+      auto Now = std::chrono::high_resolution_clock::now();
```
neobrain, line 31: Would a `steady_clock` be sufficient here as well? A high resolution clock doesn't appear necessary given that the reclamation interval is 5 seconds, and `steady_clock` (being a monotonic clock) will have more predictable behavior on system clock changes. (You wouldn't want reclamation to stop working for an hour because the system entered DST!)

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
```diff
@@ -20,19 +21,12 @@ namespace FEXCore::IR {
...
   public:
     DualIntrusiveAllocator() = delete;
```
neobrain, line 26: Aren't these two constructors implicitly deleted due to providing the `size_t` constructor?

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: I very explicitly chose list here so iterators aren't invalidated on insert/delete.
Could add a comment stating why I chose list.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Range based used for the ones that can. The other ones still need the iterators to be passed around.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
Sonicadvance1: Yes. Resolved.

External/FEXCore/include/FEXCore/IR/IREmitter.h
Sonicadvance1: You right. Resolved.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
Sonicadvance1: You right. Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Added a comment that I'm requiring iterators to stay valid while manipulating the container. Resolved.

Sonicadvance1:
Fixed the remaining bugs in this and added a bunch of comments about how to use the API.
Now the most major contributor to memory consumption is the register allocator. Specifically the graph nodes since they aren't ever freed.

Once this is merged those can be looked at to fix its memory consumption, since it isn't quite as straightforward as just moving the allocator over to using these.

Sonicadvance1:
Image_2022-04-24_11-30-29
Heaptrack shows a significant change but it isn't this significant because some allocations have switched from malloc to mmap. 1.6G down to 229.3MB within FEXLoader.
I don't think there is a tool that can see resident memory usage reduction other than what tools like htop give, sadly.

skmp:
I'll get on reviewing this first thing tomorrow, sorry for the delay

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+  };
+
+  /**
+   * @brief Thread pool allocator that allocates and frees objects using malloc
+   */
+  class ThreadPoolAllocatorMalloc final : public IntrusiveThreadPoolAllocator {
+    public:
+      ThreadPoolAllocatorMalloc() = default;
+
+      virtual ~ThreadPoolAllocatorMalloc() {
```
skmp, line 316: Since both dtors `ThreadPoolAllocatorMalloc`, `ThreadPoolAllocatorMMap` are the same, wouldn't it be nicer to have the dtor logic in `IntrusiveThreadPoolAllocator` and rely on the alloc/free overloads? Unless I'm missing something

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+                  LRUTime = UsedTime;
+                }
+              }
+            }
+
+            // If we found a buffer then free it
+            if (LastUsed != BuffersEnd) {
+              Free((*LastUsed)->Ptr, (*LastUsed)->Size);
+              delete *LastUsed;
+              OwnedBuffers.erase(LastUsed);
```
skmp, line 254: `it = OwnedBuffers.erase(LastUsed);` ?

Sonicadvance1: This is no longer isn't the for loop, there is no iterator to update.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+              UnsizedFit = it;
+            }
+          }
+
+          // If we didn't have an exact fit then use an unsized fit
+          if (BestFit == BuffersEnd) {
+            BestFit = UnsizedFit;
+          }
+
+          {
```
skmp, line 232:
```suggestion
          // Free owned buffers that have expired
          {
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+                  it = NonOwnedBuffers.erase(it);
+                  continue;
+                }
+              }
+            }
+
+            ++it;
+          }
+        }
+
```
skmp, line 213:
```suggestion
          // Find an owned buffer that is >= Size and Free owned buffers that have expired
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+    protected:
+      IntrusiveThreadPoolAllocator() = default;
+
+      ContainerType::iterator ClaimBufferImpl(size_t Size) {
+        auto BuffersEnd = OwnedBuffers.end();
+        ContainerType::iterator BestFit = BuffersEnd;
+        ContainerType::iterator UnsizedFit = BuffersEnd;
+
+        auto Now = ClockType::now();
```
skmp, line 188:
```suggestion
        auto Now = ClockType::now();
        // Move any expired NonOwnedBuffers to OwnedBuffers
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+      void UnclaimBufferImpl(ContainerType::iterator Buffer) {
+        (*Buffer)->CurrentClientOwnedFlag = nullptr;
+        OwnedBuffers.emplace_back(*Buffer);
+        NonOwnedBuffers.erase(Buffer);
+      }
+
+      /**
+       * @brief List of buffers that this pool allocator itself owns
+       */
+      ContainerType OwnedBuffers;
```
skmp, line 279: How about renaming to `UnclaimedBuffers`,  and `ClaimedBuffers` so that `ClaimBuffer`/`ClaimBufferImpl` & `UnclaimBuffer`/`UnclaimBufferImpl` are more intuitive to understand?

`DisownBuffer` and `ReownOrClaimBuffer` don't modify those lists, and from my understanding they track the `claimed`/`unclaimed` statuses, not `owned`/`disowned`

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+   *    - Updating two std::list containers to give the ownership back to the `Allocator`
+   *
+   *  - Claiming is very costly
+   *    - Requires owning a mutex, shared between all threads using the `Allocator`
+   *    - Scans two std::list containers to find the best fit buffer
+   *    - Or allocates another buffer when that fails
+   *    - Frees stale buffers opportunistically
+   */
+  template<typename Type, size_t PeriodMS, size_t PeriodFrequency>
+  class FixedSizeDelayedObjectReclaimer final {
```
skmp, line 414: How about we name this from `FixedSizeDelayedObjectReclaimer` to `FixedSizePooledAllocation` for clarity?

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,507 @@
...
+   *       - This is expensive and requires a mutex shared between threads
+   *     - `FixedSizeDelayedObjectReclaimer` helper class provided to help with this.
+   *
+   * Once the client has disowned a buffer then the allocator is free to reclaim the buffer when another thread is trying to `Claim` a new buffer.
+   * The buffer getting claimed from a disowned client must have had its last use greater than the defined `DURATION` before it has a chance to get
+   * reclaimed by the Allocator.
+   *
+   * During buffer reclaiming is also when unclaimed buffers get freed. This means active threads are able to clean up idle thread's unused memory.
+   */
+  class IntrusiveThreadPoolAllocator {
```
skmp, line 35: How about we rename this to  `IntrusivePooledAllocator` to make it clear it doesn't allocate from a `ThreadPool`?

skmp:
Some performance points for discussion:

Switching `IntrusiveThreadPoolAllocator::ContainerType` to a multimap structure, then ordering owned buffers by size can make finding the right buffer O(ln) from O(n), possibly a major win depending on how many buffers we have in flight.

Not sure the same idea would be applicable for unowned buffers (sorted by last used time), as that'd require a re-insertion on disown which will increase the cost there.

What is the cost of `std::chrono` and such? Do they result in syscalls? Can we guarantee somehow `rdtsc`/`cntvct_el0` for it?

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Can't
`Call to pure virtual member function 'Free' has undefined behavior; overrides of 'Free' in subclasses are not available in the destructor of 'IntrusivePooledAllocator'  ThreadPoolAllocator.h:323:7: note: 'Free' declared here [-Wcall-to-pure-virtual-from-ctor-dtor]`

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

Sonicadvance1:
`std::chrono` doesn't result in a syscall. It'll use the cycle counter registers, potentially through a vdso. quite lightweight.

I haven't noticed any performance concerns from the current `std::list` container but I've also not found something that pedantically disowns thousands of threads worth of buffers and then spins all of them up simultaneously so they are all hunting for buffers to reown.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
skmp: Doh, right, derived class is destructed at that point. How about moving them to a function so that we don't have two copies of the code around?

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
skmp: Misread the indentation. So at most one buffer gets free'd for every call to `ClaimBufferImpl`. Is this to avoid stutter?

Sonicadvance1: Yes, only one unclaimed buffer gets freed and removed from the `UnclaimedBuffers` container each time a new claim comes in. This is setup this way for longer running applications having mostly idle threads and then a handful of active threads will slowly chip away at the problem.
Also if you have an intermediate mix of active and inactive, you aren't suddenly getting a spike of deleted unclaimed buffers and new threads having absolutely no buffers available.

Basically the five second buffer of only freeing one buffer at a time is a nice sponge in both directions.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Moved duplicated freeing logic to its own function. Resolved now.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,502 @@
...
+              UnsizedFit = it;
+            }
+          }
+
+          // If we didn't have an exact fit then use an unsized fit
+          if (BestFit == BuffersEnd) {
+            BestFit = UnsizedFit;
+          }
+
+          // Free owned buffers that have expired
```
skmp, line 234:
```suggestion
          // Free up to one unclaimed buffer that has expired
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,502 @@
...
+                  continue;
+                }
+              }
+            }
+
+            ++it;
+          }
+        }
+
+        // Find an owned buffer that is >= Size and Free owned buffers that have expired
```
skmp, line 215:
```suggestion
        // Find an unclaimed buffer that is >= Size and Free up to one unclaimed buffer that has expired
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,502 @@
...
+
+      void Free(void* Ptr, size_t Size) override {
+        FEXCore::Allocator::free(Ptr);
+      }
+  };
+
+  /**
+   * @brief Thread pool allocator that allocates and frees objects that uses mmap
+   */
+  class ThreadPoolAllocatorMMap final : public IntrusivePooledAllocator {
```
skmp, line 351:
```suggestion
  class PooledAllocatorMMap final : public IntrusivePooledAllocator {
```

or 

```suggestion
  class PooledMMapAllocator final : public IntrusivePooledAllocator {
```

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -0,0 +1,502 @@
...
+       * @param Ptr buffer pointer
+       * @param Size buffer size
+       */
+      virtual void Free(void* Ptr, size_t Size) = 0;
+  };
+
+  /**
+   * @brief Thread pool allocator that allocates and frees objects using malloc
+   */
+  class ThreadPoolAllocatorMalloc final : public IntrusivePooledAllocator {
```
skmp, line 330:
```suggestion
  class PolledAllocatorMalloc final : public IntrusivePooledAllocator {
```

Or

```suggestion
  class PolledMallocAllocator final : public IntrusivePooledAllocator {
```
?

skmp (approved):
Noted a few naming nitpicks, Looks good overall

skmp:
> `std::chrono` doesn't result in a syscall. It'll use the cycle counter registers, potentially through a vdso. quite lightweight.
> 
> I haven't noticed any performance concerns from the current `std::list` container but I've also not found something that pedantically disowns thousands of threads worth of buffers and then spins all of them up simultaneously so they are all hunting for buffers to reown.

More of a note for the future then

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
neobrain: What happened to this? Looks like the destructor is gone again now.

PR #1632, opened 2022-03-21
TestHarnessRunner: Flush log on asserts

skmp:


Source/Tests/TestHarnessRunner.cpp
```diff
@@ -62,6 +62,10 @@ void MsgHandler(LogMan::DebugLevels Level, char const *Message) {
 
 void AssertHandler(char const *Message) {
   fmt::print("[ASSERT] {}\n", Message);
+
+  // make sure buffers are flushed
+  fflush(stderr);
```
Sonicadvance1, line 67: Shouldn't you just use `fflush(nullptr);` to get all streams flushed?

skmp: done

Sonicadvance1 (approved):

PR #1635, opened 2022-03-22
CPUID: Adds 4000_0001h function

Sonicadvance1:
Exposes the host architecture through this CPUID function. Only exposes
the architectures we support. Not burning 16-bits on using ELF machine
definitions here.

Uses 4 bits still for future expansion.


skmp (approved):

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -826,14 +826,33 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_4000_0000h(uint32_t Leaf) {
...
+  if (Leaf == 0) {
+    // EAX[3:0] Is the host architecture that FEX is running under
+#ifdef _M_X86_64
+    // EAX[3:0] = 1 = x86_64 host architecture
+    Res.eax |= 0b0001;
+#elif defined(_M_ARM_64)
+    // EAX[3:0] = 2 = AArch64 host architecture
+    Res.eax |= 0b0010;
+#else
+    // EAX[3:0] = 0 = Unknown architecture
```
neobrain, line 849: In the case we do add support for more architectures, wouldn't it be better to have this path `#error` out instead of silently reporting an unknown architecture? (Also see my related comment in #1640.)

PR #1640, opened 2022-03-23
WIP: Initial RISCV64 support

Sonicadvance1:
Doesn't work yet.


External/FEXCore/Source/Interface/Core/Dispatcher/RISCVDispatcher.cpp
```diff
@@ -0,0 +1,54 @@
...
+
+RISCVDispatcher::RISCVDispatcher(FEXCore::Context::Context *ctx, FEXCore::Core::InternalThreadState *Thread, uint8_t* Buffer)
+  : Dispatcher(ctx, Thread), RISCVEmitter(ctx, Buffer, MAX_DISPATCHER_CODE_SIZE) {
+
+
+  biscuit::Label LoopTop;
+  Bind(&LoopTop);
+  J(&LoopTop);
+
+  DispatchPtr = (CPUBackend::AsmDispatch)((uintptr_t)Buffer + LoopTop.GetLocation().value());
```
Sonicadvance1, line 40: @lioncash Might want to look at letting Biscuit get an absolute cursor pointer location from the Assembler.

neobrain:
Any chance we could throw out the arch-specific defines and instead consolidate them to a single `FEX_HOST_ARCH` one? That allows us to reliably check for unhandled hostarchs in `#if-#elif-#else` cascades. Since we have a fair bunch of those across the code base, it seems too easy to slip in an uncaught typo otherwise that's easily avoided with a unified define.

Sonicadvance1:
> Any chance we could throw out the arch-specific defines and instead consolidate them to a single `FEX_HOST_ARCH` one? That allows us to reliably check for unhandled hostarchs in `#if-#elif-#else` cascades. Since we have a fair bunch of those across the code base, it seems too easy to slip in an uncaught typo otherwise that's easily avoided with a unified define.

How does that change the behaviour or chance of typo?
```c++
#if defined(_M_ARM_64)
#warning Do ARM things
#elif defined(_M_X86_64)
#warning Do x86 things
#elif defined (_M_RISCV_64)
#warning Do risky things
#else 
#error Unsupported arch
#endif
```

Has the same chance to do incorrect logic in the chance of a typo as:
```c++
#if FEX_HOST_ARCH == AARCH64
#warning Do ARM things
#elif FEX_HOST_ARCH == X86
#warning Do x86 things
#elif FEX_HOST_ARCH == RISCV
#warning Do risky things
#else 
#error Unsupported arch
#endif
```

neobrain:
> How does that change the behaviour or chance of typo?

Fair point - I was probably thinking that AARCH64/X86/RISCV are easier to correctly remember (or look up) than the other defines, but as you're pointing out there's no risk in typos if an `#else` branch is included.

So the really important bit is to have that `#else` branch consistently rather than using custom defines.

External/FEXCore/Source/Utils/MemberFunctionToPointer.h
```diff
@@ -17,12 +17,13 @@ class MemberFunctionToPointerCast final {
     MemberFunctionToPointerCast(PointerToMemberType Function) {
       memcpy(&PMF, &Function, sizeof(PMF));
 
-#ifdef _M_X86_64
+#if defined(_M_X86_64) || defined(_M_RISCV_64)
+      // XXX: Is RISCV iatnium?
```
neobrain, line 21:
```suggestion
      // XXX: Is RISCV Itanium?
```

External/FEXCore/include/FEXCore/Core/CoreState.h
```diff
@@ -144,6 +144,37 @@ namespace FEXCore::Core {
...
+      uint64_t ThreadStopHandlerSpillSRA{};
+      uint64_t ThreadPauseHandlerSpillSRA{};
+      uint64_t ThreadStopHandler{};
+      uint64_t ThreadPauseHandler{};
+      uint64_t UnimplementedInstructionHandler{};
+      uint64_t OverflowExceptionHandler{};
+      uint64_t SignalReturnHandler{};
+      uint64_t L1Pointer{};
+      /**  @} */
+    } RISCV;
```
neobrain, line 177: Should this be wrapped in an `#ifdef`?

CMakeLists.txt
```diff
@@ -281,6 +286,8 @@ include_directories(External/robin-map/include/)
 add_subdirectory(External/vixl/)
 include_directories(External/vixl/src/)
 
+add_subdirectory(External/biscuit/)
```
neobrain, line 289: Do we need this if `!ENABLE_JIT_RISCV`?

skmp:
Adding my view here, I think it'd be useful to prefix those with FEX (FEX_M_ARM_64) to make it easier to lookup. Though, that's more of a general design decision.

> > Any chance we could throw out the arch-specific defines and instead consolidate them to a single `FEX_HOST_ARCH` one? That allows us to reliably check for unhandled hostarchs in `#if-#elif-#else` cascades. Since we have a fair bunch of those across the code base, it seems too easy to slip in an uncaught typo otherwise that's easily avoided with a unified define.
> 
> How does that change the behaviour or chance of typo?
> 
> ```c++
> #if defined(_M_ARM_64)
> #warning Do ARM things
> #elif defined(_M_X86_64)
> #warning Do x86 things
> #elif defined (_M_RISCV_64)
> #warning Do risky things
> #else 
> #error Unsupported arch
> #endif
> ```
> 
> Has the same chance to do incorrect logic in the chance of a typo as:
> 
> ```c++
> #if FEX_HOST_ARCH == AARCH64
> #warning Do ARM things
> #elif FEX_HOST_ARCH == X86
> #warning Do x86 things
> #elif FEX_HOST_ARCH == RISCV
> #warning Do risky things
> #else 
> #error Unsupported arch
> #endif
> ```

A question on that, are `_M_ARM_64` FEX-specific, or provided by the OS and such?

> > Any chance we could throw out the arch-specific defines and instead consolidate them to a single `FEX_HOST_ARCH` one? That allows us to reliably check for unhandled hostarchs in `#if-#elif-#else` cascades. Since we have a fair bunch of those across the code base, it seems too easy to slip in an uncaught typo otherwise that's easily avoided with a unified define.
> 
> How does that change the behaviour or chance of typo?
> 
> ```c++
> #if defined(_M_ARM_64)
> #warning Do ARM things
> #elif defined(_M_X86_64)
> #warning Do x86 things
> #elif defined (_M_RISCV_64)
> #warning Do risky things
> #else 
> #error Unsupported arch
> #endif
> ```
> 
> Has the same chance to do incorrect logic in the chance of a typo as:
> 
> ```c++
> #if FEX_HOST_ARCH == AARCH64
> #warning Do ARM things
> #elif FEX_HOST_ARCH == X86
> #warning Do x86 things
> #elif FEX_HOST_ARCH == RISCV
> #warning Do risky things
> #else 
> #error Unsupported arch
> #endif
> ```

External/FEXCore/include/FEXCore/Core/CoreState.h
skmp: We do support building non-current-host backends, though not sure if that would work with the RISCV backend.

archanox:
Is there anything the community can do to help progress this? Testing on actual hardware etc?

Sonicadvance1:
Main thing here is that I just need to continue working on this. 
SIGBUS behaviour differing between qemu and hardware is a bit annoying, but I'll just need to handle that.

External/FEXCore/Source/Utils/MemberFunctionToPointer.h
brucehoult: Also, what kind of question is that??   No, Itanium is IA64, an Intel proprietary (and failed) ISA. RISC-V is an utterly different open source ISA.

Sonicadvance1: It's not a question about the ISA being Itanium. It's a question if RISC-V follows the Itanium C++ ABI style of member function pointers.
Itanium spec around this is here: https://itanium-cxx-abi.github.io/cxx-abi/abi.html#member-function-pointers

The question is raised since Member Function pointer spec in most architectures follow exactly what Itanium does here, but AArch64 doesn't. Which is why we need to find the RISC-V spec on this. The question being for future me.

Thanks for the code review.

brucehoult: Oh I see. Maybe the comment could have been clearer :-)

I don't know of anywhere that it is explicitly stated, but it appears to be implicit that RISC-V doesn't intend to break any new ground and cause incompatibility with other software here. Certainly function pointers are just pointers, the low bit can't be set, VTBL entries can't have an odd size and so forth.

RISC-V founder Andrew Waterman references the Itanium ABI with respect to C++ classes with nontrivial copy constructors here: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/issues/7

RISC-V is explicitly documented as using the same C++ name-mangling as Itanium: https://github.com/riscv-non-isa/riscv-elf-psabi-doc/blob/master/riscv-elf.adoc#c-name-mangling

You could open an issue asking for confirmation in that repo, but the intent seems clear.

skmp:
(Converted this to a draft as it is in progress)

archanox:
RIP to this WIP

Beryesa:
:/

Nyabinary:
Aw man RIP, RISC-V support is def needed imo

vanillajonathan:
Did you completely give up on this? Have you considered picking it up again?

Sonicadvance1:
No plans for RISC-V support, there are other [projects](https://github.com/OFFTKP/felix86) which are doing this.

PR #1649, opened 2022-04-19
Termux: Work around shm symbol collision

Sonicadvance1:
Termux defines all of the shm functions as defines to libandroid-shmem
equivalents.
This messes with our function definitions that were using the same names
but weren't expecting them to be defines.

Changes our class function names to have an uppercase for the first
letter.

Changes our syscall enums to have an underscore at the start to not get
picked up by the preprocessor.


Source/Tests/LinuxSyscalls/x32/Semaphore.cpp
```diff
@@ -318,7 +318,7 @@ namespace FEX::HLE::x32 {
           case SHM_STAT_ANY:
           case IPC_STAT: {
             struct shmid64_ds buf{};
-            Result = ::syscall(SYSCALL_DEF(shmctl), shmid, cmd, &buf);
+            Result = ::syscall(SYSCALL_DEF(_shmctl), shmid, cmd, &buf);
```
skmp, line 321: Do these syscalls exist? Shouldn't they redirect to the shm functions

skmp (changes requested):
Urgh, this is a bit of a nasty thing.

I don't like the inconsistencies between some syscalls being capitalized/prefixed and others not being so. I'd rather that we do prefixing for our of handlers, or some other consistent measure around this, as it will be less confusing.

In another direction, can't we just detect the #define and manually work around it in this case? That seems like a simpler route.

Source/Tests/LinuxSyscalls/LinuxAllocator.h
```diff
@@ -12,8 +12,9 @@ namespace FEX::HLE {
     virtual void *mmap(void *addr, size_t length, int prot, int flags, int fd, off_t offset) = 0;
     virtual int munmap(void *addr, size_t length) = 0;
     virtual void *mremap(void *old_address, size_t old_size, size_t new_size, int flags, void *new_address) = 0;
-    virtual uint64_t shmat(int shmid, const void* shmaddr, int shmflg, uint32_t *ResultAddress) = 0;
-    virtual uint64_t shmdt(const void* shmaddr) = 0;
+    // Global symbol collision on undercase
```
neobrain, line 15: This comment lacks context required to understand it outside of this PR.
```suggestion
    // Named Shmat/Shmdt instead of shmat/shmdt to avoid symbol collision with XYZ
```

PR #1653, opened 2022-04-20
FEX: Stop hardcoding page size at build time

Sonicadvance1:
Query the host page size when bringing up the allocators and use that.
jemalloc still uses its hardcoded defines as expected.

Anything greater than 4k page size is still unsupported.


skmp:
Can't the host's `PAGE_SIZE` be used as a constant there? What is the intention here, to compile in a host that has, say, 4kb page size and run on another with 16kb?

I tested on asahi a few weeks back, and the elf loader had issues mapping the elfs. I think it was the elf themselves relying on 4kb-sized mappings or offsets, though I don't recall the specifics as the consensus was that we don't plan to do non-4kb page size support in the near future. Has that changed?

PR #1657, opened 2022-04-23
Linux: Fixes 32-bit mmap

### LINKED ISSUE #1630: Kega fusion fails to map memory
Looks like Kega Fusion does some allocating that breaks under FEX for some reason.
mmap return EINVAL and early exiting. Probably something simple.

Sonicadvance1:
This went unnoticed for so long since most applications are new enough
to use mmap2 instead of mmap.
This was just completely broken.

Fixes #1630

### DIFF
--- a/Source/Tests/LinuxSyscalls/x32/Memory.cpp
+++ b/Source/Tests/LinuxSyscalls/x32/Memory.cpp
@@ -26,18 +26,26 @@ static std::string get_fdpath(int fd)
 
 namespace FEX::HLE::x32 {
   void RegisterMemory() {
-    REGISTER_SYSCALL_IMPL_X32(mmap, [](FEXCore::Core::CpuStateFrame *Frame, uint32_t addr, uint32_t length, int prot, int flags, int fd, int32_t offset) -> uint64_t {
+    struct old_mmap_struct {
+      uint32_t addr;
+      uint32_t len;
+      uint32_t prot;
+      uint32_t flags;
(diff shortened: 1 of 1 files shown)


Source/Tests/LinuxSyscalls/x32/Memory.cpp
```diff
@@ -26,18 +26,26 @@ static std::string get_fdpath(int fd)
...
-    REGISTER_SYSCALL_IMPL_X32(mmap, [](FEXCore::Core::CpuStateFrame *Frame, uint32_t addr, uint32_t length, int prot, int flags, int fd, int32_t offset) -> uint64_t {
+    struct old_mmap_struct {
+      uint32_t addr;
+      uint32_t len;
+      uint32_t prot;
+      uint32_t flags;
+      uint32_t fd;
+      uint32_t offset;
+    };
+    REGISTER_SYSCALL_IMPL_X32(mmap, [](FEXCore::Core::CpuStateFrame *Frame, old_mmap_struct *const arg) -> uint64_t {
```
skmp, line 37: Shouldn't this be `const old_mmap_struct * arg` ? (const data, not const pointer?)

skmp (approved):

Source/Tests/LinuxSyscalls/x32/Memory.cpp
Sonicadvance1: Oops yes, fixed.

PR #1662, opened 2022-04-28
F64: Fix FILD and FIST for Size < 8

CallumDev:
Previously was using 64-bit registers for all loads and stores. Uses 32-bit registers where applicable, and sign-extends 16-bit numbers on load. This fixes FIST overflow and negative numbers in FILD.


Sonicadvance1 (approved):
Nice debugging!

PR #1664, opened 2022-04-28
F64: Implement FCW using host rounding mode

CallumDev:
Typical C programs truncate double to int by modifying the control world instead of using the truncating version of `fist`. Make FLDCW in F64 modify the host rounding mode.


Sonicadvance1:
You probably also want to hit the instructions that load the environment like `X87LDENV` which will load the FCW.
You should also open an issue for tracking this new problem that this introduces, which is mixing MXCSR and FCW round modes, which are independent.
It worked previously because FCW round mode was always softfloat and MXCSR was also set to host rounding mode.

Sonicadvance1 (approved):
Nice!

PR #1668, opened 2022-05-01
Segment register index optimization

Sonicadvance1:
Segment registers are indexed significantly more than they are changed.
Pay the cost of indexing during the set and store rather than the per
register index.

Should be a fairly significant performance improvement for 32-bit
applications. At least on hardware that doesn't have a data dependent
prefetcher.

Breaks Steam atm and isn't clean.

### DIFF
--- a/Source/Tests/LinuxSyscalls/x32/Thread.cpp
+++ b/Source/Tests/LinuxSyscalls/x32/Thread.cpp
@@ -68,6 +68,29 @@ namespace FEX::HLE::x32 {
     // Now we need to update the thread's GDT to handle this change
     auto GDT = &Frame->State.gdt[u_info->entry_number];
     GDT->base = u_info->base_addr;
+
+    // With the segment register optimization we need to check all of the segment registers and update.
+    const auto GetEntry = [](auto value) {
+      return value >> 3;
+    };
+    if (GetEntry(Frame->State.cs_idx) == u_info->entry_number) {
+      Frame->State.cs_cached = GDT->base;
+    }
+    if (GetEntry(Frame->State.ds_idx) == u_info->entry_number) {
+      Frame->State.ds_cached = GDT->base;
+    }
+    if (GetEntry(Frame->State.es_idx) == u_info->entry_number) {
+      Frame->State.es_cached = GDT->base;
+    }
+    if (GetEntry(Frame->State.fs_idx) == u_info->entry_number) {
+      Frame->State.fs_cached = GDT->base;
+    }
+    if (GetEntry(Frame->State.gs_idx) == u_info->entry_number) {
+      Frame->State.gs_cached = GDT->base;
+    }
+    if (GetEntry(Frame->State.ss_idx) == u_info->entry_number) {
+      Frame->State.ss_cached = GDT->base;
+    }
     return 0;
   }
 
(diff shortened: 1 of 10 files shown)


skmp:
Looks like a good idea. Naming wise, how about we keep the original segment register names, and we name the cached ones, `<segment>_cached` instead?

Sonicadvance1:
Rebased so it doesn't have merge conflicts but there is still some things wrong here.

Sonicadvance1:
Finally found out what I missed in this PR.
When a new thread is getting created I forgot to update the cached segment registers with the new TLS region.
I need to bisect a crash on upstream before testing and merging this.
Although it is ready for review now.

Sonicadvance1:
Oop, nevermind no regression. I just had the vixl simulator still enabled in my build configuration. It's all good.

Sonicadvance1:
Some minor benchmarking of this compared to main.
Steam startup time (From their startup number metric): 90s -> 83s
Grim Fandango in-game: 80-90FPS -> 90-100FPS

So it seems like it is around a 8-11% perf uplift for 32-bit applications.
Would need more benchmarking to get a real metric but seems pretty good to me.

Sonicadvance1:
Even in a benchmark that theoretically shouldn't show a major improvement, is showing an improvement with this PR.
32-bit nbench | Post-opt | Pre-opt | Improvement
-- | -- | -- | --
Numeric | 567.84 | 542.96 | 4.58%
String | 219.67 | 202.93 | 8.25%
Bitfield | 2.75E+08 | 2.64E+08 | 4.03%
FP emu | 275.06 | 264.67 | 3.93%
Fourier | 1028.4 | 974.79 | 5.50%
assignment | 10.999 | 10.471 | 5.04%
idea | 3192.7 | 3019.2 | 5.75%
huffman | 951.68 | 936.66 | 1.60%
NN | 0.83537 | 0.81323 | 2.72%
LU | 28.666 | 28.098 | 2.02%
  |   | Avg: | 4.34%

Sonicadvance1:
After some more extensive testing this actually manages to break Wine. So this again needs more work.

Sonicadvance1:
This is likely from `set_thread_area` syscall not updating the cached values, which wine uses.

Sonicadvance1:
Finally found the final bug in this!
The `UpdatePrefixFromSegment` function needed to BFE the incoming segment register value instead of shifting.
This is due to the full 32-bit GPR having garbage in the upper 16-bits which was ruining the index lookup.
Confirmed working in wine and proton.

Source/Tests/LinuxSyscalls/x32/Thread.cpp
```diff
@@ -68,6 +68,27 @@ namespace FEX::HLE::x32 {
     // Now we need to update the thread's GDT to handle this change
     auto GDT = &Frame->State.gdt[u_info->entry_number];
     GDT->base = u_info->base_addr;
+
+    // With the segment register optimization we need to check all of the segment registers and update.
+#define GETENTRY(x) (x >> 3)
```
lioncash, line 73: maybe `#undef` this before the final return, to keep the define localized or perhaps turn it into a lambda like

```cpp
const auto GetEntry = [](auto value) {
  return value >> 3;
};
```

lioncash (approved):
Just a minor nit, but overall looks good

Source/Tests/LinuxSyscalls/x32/Thread.cpp
Sonicadvance1: Good catch, I originally had it as a define since it was doing more work. Switched over to a lambda.

PR #1669, opened 2022-05-01
ScopedSignalMask: Add shared mutex support, move constructors

skmp:
## Overview
Split off from #1558, allows us to return and pass around ScopedSignalMasks

### Added signal blocking `shared_mutex` scoped locks, Implemented move semantics
- `ScopedSignalMaskWithSharedLock`, `ScopedSignalMaskWithUniqueLock`; similar to `ScopedSignalMaskWithMutex`

- Implemented move semantics so they can be passed around, similar to `std::lock_guard`, `std::shared_lock`, `std::unique_lock`

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
@@ -12,7 +12,7 @@
 
 namespace FHU {
   /**
-   * @brief A class that masks signals and locks a mutex until it goes out of scope
+   * @brief A class that masks signals and locks a mutex until it goes out of scope. It is NOT SAFE to move across threads.
    *
    * Constructor order:
    * 1) Mask signals
@@ -61,7 +61,7 @@ namespace FHU {
   };
 
   /**
-   * @brief A class that masks signals and shared locks a shared mutex until it goes out of scope
+   * @brief A class that masks signals and shared locks a shared mutex until it goes out of scope. It is NOT SAFE to move across threads.
    *
    * Constructor order:
    * 1) Mask signals
@@ -110,7 +110,7 @@ namespace FHU {
 
 
   /**
-   * @brief A class that masks signals and unique locks a shared mutex until it goes out of scope
+   * @brief A class that masks signals and unique locks a shared mutex until it goes out of scope. It is NOT SAFE to move across threads.
    *
    * Constructor order:
    * 1) Mask signals
--- a/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
@@ -5,6 +5,7 @@
 
 #include <cstdint>
 #include <mutex>
+#include <shared_mutex>
 #include <signal.h>
 #include <sys/syscall.h>
 #include <unistd.h>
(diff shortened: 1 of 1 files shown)


FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
...
 
       ~ScopedSignalMaskWithMutex() {
-        // Unlock the mutex
-        Mutex.unlock();
+        if (Mutex != nullptr) {
+          // Unlock the mutex
+          Mutex->unlock();
+
+          // Unmask back to the original signal mask
+          ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &OriginalMask, nullptr, sizeof(OriginalMask));
```
Sonicadvance1, line 55: If this was moving across a thread then the signal mask of the original thread that created the mutex will have been modified then the new thread's mask would be set to the previous thread's mask.

skmp: I'm not sure if locks can be moved across threads, and it's best to avoid for that for now, I guess.

Sonicadvance1: I'd say that we should throw a TID check in the constructor and destructor, but that would break on fork.
Should probably add some strongly worded comment on these that they should never be moved between threads.

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
    * Destructor Order:
    * 1) Unlock Mutex
    * 2) Unmask signals
+   *
+   * Masking signals around mutex locks is needed for signal-reentrant safety
    */
```
skmp, line 26:
```suggestion
   *
   * it is NOT SAFE to move across threads
   */
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
...
+   * Constructor order:
+   * 1) Mask signals
+   * 2) Lock Mutex
+   *
+   * Destructor Order:
+   * 1) Unlock Mutex
+   * 2) Unmask signals
+   *
+   * Masking signals around mutex locks is needed for signal-rentrant safety
+   */
```
skmp, line 75:
```suggestion
   *
   * it is NOT SAFE to move across threads
   */
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
...
```
skmp, line 124:
```suggestion
   *
   * it is NOT SAFE to move across threads
   */
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
...
+        }
+      }
+    private:
+      uint64_t OriginalMask{};
+      std::shared_mutex *Mutex;
+  };
+
+
+  /**
+   * @brief A class that masks signals and unique locks a shared mutex until it goes out of scope
```
skmp, line 113:
```suggestion
   * @brief A class that masks signals and unique locks a shared mutex until it goes out of scope. it is NOT SAFE to move across threads.
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
...
+          ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &OriginalMask, nullptr, sizeof(OriginalMask));
+        }
+      }
+    private:
+      uint64_t OriginalMask{};
+      std::mutex *Mutex;
+  };
+
+  /**
+   * @brief A class that masks signals and shared locks a shared mutex until it goes out of scope
```
skmp, line 64:
```suggestion
   * @brief A class that masks signals and shared locks a shared mutex until it goes out of scope. it is NOT SAFE to move across threads.
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -20,27 +21,139 @@ namespace FHU {
    * Destructor Order:
    * 1) Unlock Mutex
    * 2) Unmask signals
+   *
+   * Masking signals around mutex locks is needed for signal-reentrant safety
```
skmp, line 25:
```suggestion
   * Masking signals around mutex locks is needed for signal-reentrant safety. It is NOT SAFE to move across threads.
```

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
skmp: @Sonicadvance1 added in class description for all 3 of em

Sonicadvance1 (approved):

PR #1670, opened 2022-05-01
Core: context-wide guest code invalidations

skmp:
## Overview

Split off #1558, requires #1669.

Makes guest code invalidations context-wide by default, renames API from `Context::FlushCodeRange` to `Context::InvalidateGuestCodeRange`

### Invalidating jit code can now be done context-wide (affects: `Context`, `LookupCache`, `IR`)
- `LookupCache` has been updated to include a lock that is only ever contested on cross thread invalidations, and protects all writes and most reads from all structures, including `InternalThreadState::LocalIRCache` which is not part of LookupCache)

- Reads from `LookupCache` L1C are not protected by the mutex, and invalidations to L1C are done in a way that will likely tear without side effects. This needs to be validated & tested.

- Only invalidations of `LookupCache` may happen from another thread. All other operations are done from the owning thread.

- `FlushCodeRange()` has been changed to iterate over all threads and invalidate on them all

- Thread invalidation has been moved to `FlushThreadCodeRange()`

- In IR/Backends `OP_RemoveCodeEntry`  has been renamed to `OP_RemoveThreadCodeEntry`, as it is meant to do invalidations only on the current thread.

## Suggested follow ups
- [x] Move `LocalIRCache` to `LookupCache` (#1675)

### DIFF
--- a/External/FEXCore/Source/Interface/Context/Context.h
+++ b/External/FEXCore/Source/Interface/Context/Context.h
@@ -155,11 +155,13 @@ namespace FEXCore::Context {
     void RegisterHostSignalHandler(int Signal, HostSignalDelegatorFunction Func, bool Required);
     void RegisterFrontendHostSignalHandler(int Signal, HostSignalDelegatorFunction Func, bool Required);
 
-    static void RemoveCodeEntry(FEXCore::Core::InternalThreadState *Thread, uint64_t GuestRIP);
+    // Must be called from owning thread
+    static void RemoveThreadCodeEntry(FEXCore::Core::InternalThreadState *Thread, uint64_t GuestRIP);
 
     // Wrapper which takes CpuStateFrame instead of InternalThreadState
-    static void RemoveCodeEntryFromJit(FEXCore::Core::CpuStateFrame *Frame, uint64_t GuestRIP) {
-      RemoveCodeEntry(Frame->Thread, GuestRIP);
+    // Must be called from owning thread
+    static void RemoveThreadCodeEntryFromJit(FEXCore::Core::CpuStateFrame *Frame, uint64_t GuestRIP) {
+      RemoveThreadCodeEntry(Frame->Thread, GuestRIP);
     }
 
     // Debugger interface
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -447,6 +447,7 @@ namespace FEXCore::Context {
           : nullptr),
         decltype(Entry.DebugData)(new Core::DebugData())
       };
+      std::lock_guard<std::recursive_mutex> lk(Thread->LookupCache->WriteLock);
       Thread->LocalIRCache.insert({Addr, std::move(Entry)});
     };
 
@@ -626,6 +627,8 @@ namespace FEXCore::Context {
   }
 
   void Context::ClearCodeCache(FEXCore::Core::InternalThreadState *Thread, bool AlsoClearIRCache) {
+    std::lock_guard<std::recursive_mutex> lk(Thread->LookupCache->WriteLock);
+
     Thread->LookupCache->ClearCache();
     Thread->CPUBackend->ClearCache();
 
@@ -738,7 +741,7 @@ namespace FEXCore::Context {
           Thread->OpDispatcher->SetTrueJumpTarget(InvalidateCodeCond, CodeWasChangedBlock);
 
           Thread->OpDispatcher->SetCurrentCodeBlock(CodeWasChangedBlock);
-          Thread->OpDispatcher->_RemoveCodeEntry();
+          Thread->OpDispatcher->_RemoveThreadCodeEntry();
           Thread->OpDispatcher->_ExitFunction(Thread->OpDispatcher->_EntrypointOffset(Block.Entry + BlockInstructionsLength - GuestRIP, GPRSize));
 
           auto NextOpBlock = Thread->OpDispatcher->CreateNewCodeBlockAfter(CurrentBlock);
@@ -847,6 +850,7 @@ namespace FEXCore::Context {
     uint64_t StartAddr {};
     uint64_t Length {};
 
+    std::lock_guard<std::recursive_mutex> lk(Thread->LookupCache->WriteLock);
     // Do we already have this in the IR cache?
     auto LocalEntry = Thread->LocalIRCache.find(GuestRIP);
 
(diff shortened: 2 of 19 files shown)


External/FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -28,24 +29,51 @@ class LookupCache {
...
+    uint64_t LocalPagePointer = Pointers[PageIndex];
+    // Do we a page pointer for this address?
+    if (LocalPagePointer) {
+      // Find there pointer for the address in the blocks
+      auto BlockPointers = reinterpret_cast<LookupCacheEntry*>(LocalPagePointer);
+
+      if (BlockPointers[PageOffset].GuestCode == Address)
+      {
+        L1Entry.GuestCode = Address;
+        return L1Entry.HostCode = BlockPointers[PageOffset].HostCode;
```
neobrain, line 56: Is assignment intended here? If yes, let's be explicit about it and move it to a dedicated statement.

skmp: Yep it is intentional

skmp: Split the assignment, also removed some dead code that slipped through the slicing of the commits

External/FEXCore/Source/Interface/Context/Context.h
```diff
@@ -155,11 +155,13 @@ namespace FEXCore::Context {
     void RegisterHostSignalHandler(int Signal, HostSignalDelegatorFunction Func, bool Required);
     void RegisterFrontendHostSignalHandler(int Signal, HostSignalDelegatorFunction Func, bool Required);
 
-    static void RemoveCodeEntry(FEXCore::Core::InternalThreadState *Thread, uint64_t GuestRIP);
+    // Must be called from owning thread
```
neobrain, line 158: Owning what though? The `Context` or the `CpuStateFrame`?

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -1033,21 +1037,31 @@ namespace FEXCore::Context {
     }
   }
 
-  void FlushCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length) {
+  static void InvalidateGuestThreadCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length) {
+    std::lock_guard<std::recursive_mutex> lk(Thread->LookupCache->WriteLock);
+
+    auto lower = Thread->LookupCache->CodePages.lower_bound(Start >> 12);
+    auto upper = Thread->LookupCache->CodePages.upper_bound((Start + Length) >> 12);
```
neobrain, line 1044: This is just copied from previous code, but shouldn't it be `(Start + Length - 1) >> 12`? `upper_bound` gets the first element *greater* than its argument, so it looks like the current code potentially iterates over one page too much.

neobrain (approved):
Added a few comments, looks good otherwise!

External/FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -116,8 +151,19 @@ class LookupCache {
   constexpr static size_t L1_ENTRIES = 1 * 1024 * 1024; // Must be a power of 2
   constexpr static size_t L1_ENTRIES_MASK = L1_ENTRIES - 1;
 
+  // This needs to be taken before reads or writes to L2, L3, CodePages, Thread::LocalIRCache,
+  // and before writes to L1. Concurrent access from a thread that this LookupCache doesn't belong to 
+  // can only happen during cross thread invalidation (::Erase).
+  // All other operations must be done from the owning thread.
+  // Some care is taken so that L1 lookups can be done without locks, and even tearing happens
+  // it is unlikely to lead to a crash. This approach has not been fully vetted yet.
```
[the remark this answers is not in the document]
skmp: Handling that case is not part of the current smc/invalidation goals, so it has been added to minimize risk, and is not guaranteed to work against  pathological guest code.

I did a simple test case for this, and in practice I haven't got it to crash, yet. @Sonicadvance1 has a better understanding of the exact memory ordering and tearing characteristics of aarch64 and maybe can pitch in here.

Overall, edge cases like this one will be tackled on follow up SMC work, hopefully.

The comment is there to as a warning during bughunts, if we happen to trigger a bug around that logic.

External/FEXCore/Source/Interface/Context/Context.h
skmp: Owning the `InternalThreadState` (CpuStateFrame->Thread)

External/FEXCore/Source/Interface/Core/Core.cpp
skmp: Quite possibly. This code path wasn't heavily used before so bugs might lurk.

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -1033,21 +1037,31 @@ namespace FEXCore::Context {
...
 
-    if (Thread->CTX->Config.SMCChecks == FEXCore::Config::CONFIG_SMC_MMAN) {
-      auto lower = Thread->LookupCache->CodePages.lower_bound(Start >> 12);
-      auto upper = Thread->LookupCache->CodePages.upper_bound((Start + Length) >> 12);
+    for (auto it = lower; it != upper; it++) {
+      for (auto Address: it->second)
+      Context::RemoveThreadCodeEntry(Thread, Address);
```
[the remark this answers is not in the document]
skmp: Good catch, also added `{}`. I try to avoid single statement blocks as they can lead to bugs in general

External/FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -28,24 +29,51 @@ class LookupCache {
...
-    } else {
-      auto HostCode = BlockList.find(Address);
-
-      if (HostCode != BlockList.end()) {
-        CacheBlockMapping(Address, HostCode->second);
-        return HostCode->second;
-      } else {
-        return 0;
+    // Try L1, no lock needed
+    auto &L1Entry = reinterpret_cast<LookupCacheEntry*>(L1Pointer)[Address & L1_ENTRIES_MASK];
```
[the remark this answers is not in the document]
skmp: L1Entry might be updated after getting the lock

External/FEXCore/Source/Interface/Core/Core.cpp
skmp: Updated for -1. Good catch there.

External/FEXCore/Source/Interface/Core/LookupCache.h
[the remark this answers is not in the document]
skmp: Created #1679 for follow up

PR #1671, opened 2022-05-01
Syscalls/Linux: Refactor guest mman tracking

skmp:
## Overview
Split off #1558, requires #1670.

Centralises handling of calls to the memory manager.

### Reworked tracking of memory mappings in `HLE::SyscallsHandler` (Linux syscall abi-neutral base)
- Added `FEX::HLE::SignalHandler::Track[Mmap/Munmap/...]()` (in the base class for both 32/64 handlers)

- Now all memory syscalls, both 32 and 64 bit, call `FEX::HLE::SignalHandler::Track[Mmap/Munmap/...]()` on success of these syscalls. This provides a cleaner and uniform api, and separates syscall implementations from VMA/Flushing logic.

### DIFF
--- a/Source/Tests/LinuxSyscalls/Syscalls.h
+++ b/Source/Tests/LinuxSyscalls/Syscalls.h
@@ -81,7 +81,6 @@ uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* env
 
 class SyscallHandler : public FEXCore::HLE::SyscallHandler {
 public:
-  SyscallHandler(FEXCore::Context::Context *ctx, FEX::HLE::SignalDelegator *_SignalDelegation);
   virtual ~SyscallHandler();
 
   // In the case that the syscall doesn't hit the optimized path then we still need to go here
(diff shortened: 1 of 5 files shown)


Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -178,6 +188,8 @@ class SyscallHandler : public FEXCore::HLE::SyscallHandler {
   uint32_t HostKernelVersion{};
   uint32_t GuestKernelVersion{};
 
+  FEXCore::Context::Context *CTX;
```
neobrain, line 191: We don't repoint nor own this pointer, which would be more readily visible using a reference.

neobrain (approved):
Looks good other than one minor comment.

Much cleaner now with things in SyscallHandler instead of Context :)

Source/Tests/LinuxSyscalls/Syscalls.h
skmp: Nowhere in the code we use references for context, and pretty much for any other object that is passed around, so I'd like to stick here with the existing convention there.

Tracking allocation/free responsibilities, as well as nullable pointers would be an interesting code quality improvement though, I'd be all for it as a separate PR.

skmp: @neobrain created #1678 to continue this discussion there

PR #1677, opened 2022-05-02
Clean up and document ScopedSignalMask

neobrain:
Having three separate classes that essentially implement the same complex concept is cumbersome: It adds ambiguity to call-sites ("which of these should I use?") and wastes programmer time on looking for differences between the implementations. This isn't helped by us having to maintain the same docstring thrice, which hence hasn't seen as much love as the complexity of the concept deserves.

This change unifies the three separate classes we had before, allowing them to be documented more easily and clearly highlighting the difference of each specialization. For convenience at call-sites, the three names previously used are still available as specializations.

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
@@ -12,7 +12,15 @@
 
 namespace FHU {
   /**
-   * @brief A class that masks signals and locks a mutex until it goes out of scope. It is NOT SAFE to move across threads.
+   * @brief A drop-in replacement for std::lock_guard that masks POSIX signals while the mutex is locked
+   *
+   * Use this class to prevent reentrancy issues of C++ mutexes with certain signal handlers.
+   * Common examples of such issues are:
+   * - C++ mutexes not unlocking due to a signal handler longjmping out of a scope owning the mutex
+   * - The signal handler itself using a mutex that would be re-locked if the handler gets invoked
+   *   again before unlocking
+   *
+   * Ownership of this object may be moved, but it is NOT SAFE to move across threads.
    *
    * Constructor order:
    * 1) Mask signals
@@ -21,8 +29,6 @@ namespace FHU {
    * Destructor Order:
    * 1) Unlock Mutex
    * 2) Unmask signals
-   *
-   * Masking signals around mutex locks is needed for signal-reentrant safety
    */
   template<typename MutexType, void (MutexType::*lock_fn)(), void (MutexType::*unlock_fn)()>
   class ScopedSignalMaskWithMutexBase final {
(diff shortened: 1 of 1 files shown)


FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -12,7 +12,13 @@
 
 namespace FHU {
   /**
-   * @brief A class that masks signals and locks a mutex until it goes out of scope. It is NOT SAFE to move across threads.
+   * @brief A drop-in replacement for std::lock_guard that masks POSIX signals while the mutex is locked
+   *
+   * Use this class instead of plain C++ locks when a signal handler may use
```
skmp, line 17: The problem is not isolated in longjumps - it affects any lock that may be taken again from a signal handler, which is most locks for us

neobrain: > The problem is not related to longjumps

I thought one of the possible problems was that a signal handler may longjmp out of a scope that held a mutex, after which that mutex will not be acquirable anymore. Did I misunderstand that this happens?

> it affects any lock that may be taken again from a signal handler, which is most locks for us

Fair enough. So the signal handler interrupts a scope that held a mutex, and hence can't acquire that mutex.

> I don't think we can make a docstring that really explains all of the gotchas.

The gotcha really just boils down to "Mutexes stay locked when a POSIX signal handler is invoked, so it can't safely be locked again without precautions", doesn't it?

> We don't do this everywhere for performance reasons, and there's also a possibility of a guest deadlock if a syscall gets called while the signals are blocked

Those details are fine to omit in the docstring, I'd say. We don't need a 100% comprehensive essay about this class, but we do need to give the reader any clue whatsoever why the class exists in the first place (which was missing before this PR).

skmp: > I thought one of the possible problems was that a signal handler may longjmp out of a scope that held a mutex, after which that mutex will not be acquirable anymore. Did I misunderstand that this happen

I haven't though about the longjump case, however the most common case is when a lock gets interrupted by a signal. We have to compile code / call syscalls / do everything the guest asks us in the signal handler to to pass the signal to the guest, which makes almost all of our codebase signal re-entrant.

So, eg, without the signal masking, while we're tracking the VMA structures, a signal might happen, and the guest signal handler might do mmap, which will lead to a deadlock.

skmp:
Verified that things inline as they should, though the compiler needs to do some needless null checks :|. Oh well, can't have it all...

skmp:
Apart from the description, looks good to me. I prefered the original description, tbh.

skmp:
Did a better writeup of the issues around signals in #1682

Sonicadvance1 (approved):

FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
neobrain: I weakened the description to drop the "Use this if..." form and to only list examples instead. How is that?

Sonicadvance1 (approved):

PR #1684, opened 2022-05-06
TestHarnessRunner: Use guest mapper for test harness files

Sonicadvance1:
This will allow it to get picked up for named region handling. Thus ending up in the code caching for testing.

### DIFF
--- a/Source/Tests/HarnessHelpers.h
+++ b/Source/Tests/HarnessHelpers.h
@@ -8,6 +8,7 @@
 #include <bitset>
 #include <cassert>
 #include <cstring>
+#include <fcntl.h>
 #include <fstream>
 #include <sys/mman.h>
 #include <sys/user.h>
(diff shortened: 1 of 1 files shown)


skmp:
@Sonicadvance1 with the #1558 vma tracking and #1672 this is no longer required. I could split just the tracking features out if you want to have a quicker merge there

Sonicadvance1:
The test harness files aren't actually getting mmap'd through FD so they wouldn't get tracked as a named region I don't believe?

skmp:
All allocations done with `GuestMmap`/`GuestMunmap` will get tracked, even anonymous ones, though we'd have to change the aot notification logic there, as the callbacks only get called for fd that point files with names. 

Another option is switching to memfd for that code, that will get tracked, though there are naming issues there.

Sonicadvance1:
We could also just open the test harness with an FD and mmap it so its file gets tracked appropriately. Rather than copying it to the final location.

skmp:
> We could also just open the test harness file with an FD and mmap it so its file gets tracked appropriately. Rather than copying it to the final location.

That would be ideal

PR #1685, opened 2022-05-06
EmulatedFiles: Fixes temporary file flags

Sonicadvance1:
mode and flags were being combined incorrectly.


skmp (approved):
NICE CATCH. Tricky, that one.

PR #1686, opened 2022-05-07
ArchHelpers: Adds relocation struct defines

Sonicadvance1:
Pulled from #1548 with one of the unused relocation types removed.

Unused for now.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ObjectCache/Relocations.h
+++ b/External/FEXCore/Source/Interface/Core/ObjectCache/Relocations.h
@@ -1,0 +1,78 @@
+#pragma once
+#include <FEXCore/IR/IR.h>
+
+namespace FEXCore::CPU {
+  enum class RelocationTypes : uint8_t {
+    // 8 byte literal in memory for symbol
+    // Aligned to struct RelocNamedSymbolLiteral
+    RELOC_NAMED_SYMBOL_LITERAL,
+
+    // Fixed size named thunk move
+    // 4 instruction constant generation on AArch64
+    // 64-bit mov on x86-64
+    // Aligned to struct RelocNamedThunkMove
+    RELOC_NAMED_THUNK_MOVE,
+
+    // Fixed size guest RIP move
+    // 4 instruction constant generation on AArch64
+    // 64-bit mov on x86-64
+    // Aligned to struct RelocGuestRIPMove
+    RELOC_GUEST_RIP_MOVE,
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/Relocations.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+#include <FEXCore/IR/IR.h>
+#include <vector>
```
neobrain, line 3: This file doesn't use `vector`.

External/FEXCore/Source/Interface/Core/ArchHelpers/Relocations.h
```diff
@@ -0,0 +1,79 @@
...
+    RelocationTypeHeader Header{};
+
+    // The thunk SHA256 hash
+    IR::SHA256Sum Symbol;
+
+    // Offset in to the code section to begin the relocation
+    uint64_t Offset{};
+
+    // GPR index the constant is being moved to
+    uint8_t RegisterIndex;
```
neobrain, line 53: Moving this member in front of `Symbol` allows this struct to fit into 32 bytes of member again, doesn't it?

External/FEXCore/Source/Interface/Core/ArchHelpers/Relocations.h
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/ArchHelpers/Relocations.h
Sonicadvance1: Moved the uint8_t RegisterIndex in the two relocations to be just after the header definitions so it packs better.
Resolved.

Sonicadvance1:
> Overall looks good, though a couple of things to consider
> 
> * Ordering (as @neobrain) pointed out, we can pull in the uint8s in the front to save some space

Resolved.

> * I think the header should have a uint8_t size, so we can serialise in a more packed way (as mentioned in the union)

Once merged I will create an issue to resolve this. Only burns 48 bytes per relocation which isn't terrible. Serialization still only writes the smallest size.

> * Does this really belong to ArchHelpers? If it is serialised, shouldn't it be somewhere around the serialisation logic?

Moved to the `ObjectCache` folder that the later PRs will use

PR #1688, opened 2022-05-07
JIT relocation handling support

Sonicadvance1:
The JITs currently don't use this. This is just the handling code
itself.

One line disabled in each JIT handling Guest RIP move relocations until the JIT
Object cache is enabled.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -739,9 +739,9 @@ void *Arm64JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IR
   // X1-X3 = Temp
   // X4-r18 = RA
 
-  auto GuestEntry = GetCursorAddress<uint64_t>();
+  GuestEntry = GetCursorAddress<uint64_t>();
 
- if (CTX->GetGdbServerStatus()) {
+  if (CTX->GetGdbServerStatus()) {
     aarch64::Label RunBlock;
 
     // If we have a gdb server running then run in a less efficient mode that checks if we need to exit
@@ -831,6 +831,7 @@ void *Arm64JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IR
   if (DebugData) {
     DebugData->HostCodeSize = reinterpret_cast<uintptr_t>(CodeEnd) - reinterpret_cast<uintptr_t>(GuestEntry);
   }
+  Relocations.clear();
 
   this->IR = nullptr;
 
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
@@ -649,6 +649,7 @@ void *X86JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IRLi
   }
 
 	void *GuestEntry = getCurr<void*>();
+  CursorEntry = getSize();
   this->IR = IR;
 
   if (CTX->GetGdbServerStatus()) {
@@ -802,6 +803,8 @@ void *X86JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IRLi
   if (DebugData) {
     DebugData->HostCodeSize = reinterpret_cast<uintptr_t>(GuestExit) - reinterpret_cast<uintptr_t>(GuestEntry);
   }
+  Relocations.clear();
+
   return GuestEntry;
 }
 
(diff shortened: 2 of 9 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -648,7 +648,9 @@ void *X86JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IRLi
     ThreadState->CTX->ClearCodeCache(ThreadState, false);
   }
 
+  ClearRelocations();
```
skmp, line 651: Not specific to this commit, i think we should have a "cleanup after compile" to release this resources sooner.

Sonicadvance1: The ClearRelocations here is actually just to ensure sanitization with a `std::move` in a different part of the codebase. The vector should already be empty but std::move with std::vector is...weird. `Unless otherwise specified, such moved-from objects shall be placed in a valid but unspecified state.`

Sonicadvance1: I suppose I should probably just clear the relocation vector after move and remove this ClearRelocations.

External/FEXCore/Source/Interface/Core/JIT/x86_64/x64Relocations.cpp
```diff
@@ -0,0 +1,137 @@
...
+    mov(Reg, Constant);
+  }
+
+  Relocations.emplace_back(MoveABI);
+}
+
+bool X86JITCore::ApplyRelocations(uint64_t GuestEntry, uint64_t CodeEntry, uint64_t CursorEntry, size_t NumRelocations, const char* EntryRelocations) {
+  size_t DataIndex{};
+  for (size_t j = 0; j < NumRelocations; ++j) {
+    const FEXCore::CPU::Relocation *Reloc = reinterpret_cast<const FEXCore::CPU::Relocation *>(&EntryRelocations[DataIndex]);
```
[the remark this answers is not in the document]
Sonicadvance1: Added a `LOGMAN_THROW_A_FMT` check in both JITs doing relocations. Resolved.

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
Sonicadvance1: Added a vector clear At the end of each JIT's CompileCode function. Which will be exactly where the relocations vector gets moved.
Resolved.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -739,9 +739,9 @@ void *Arm64JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IR
   // X1-X3 = Temp
   // X4-r18 = RA
 
-  auto GuestEntry = GetCursorAddress<uint64_t>();
+  GuestEntry = GetCursorAddress<uint64_t>();
```
skmp, line 742: `GuestEntry` is a Guest or Host pointer?

skmp:
Looks good overall, and I love us having relocations!

Wouldn't it be best to put the `NamedSymbols` in some field in the Context vs tracking them as relocations?

Also, how compatible would the approach be on sharing code between threads?

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -831,6 +831,7 @@ void *Arm64JITCore::CompileCode(uint64_t Entry, [[maybe_unused]] FEXCore::IR::IR
   if (DebugData) {
     DebugData->HostCodeSize = reinterpret_cast<uintptr_t>(CodeEnd) - reinterpret_cast<uintptr_t>(GuestEntry);
   }
+  Relocations.clear();
```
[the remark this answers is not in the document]
Sonicadvance1: Nope.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
Sonicadvance1: Host pointer

Sonicadvance1:
Quite compatible. I've removed a bunch of thread-specific information but some of it is still there. The Named Symbol relocation specifically because that leads to the dispatcher which is thread specific.
With some more work we will be able to do thread sharing of data. Which is the plan once the code cache stabilizes.

skmp:
I meant the thread context pointer `CpuStateFrame`, not `FEXCore::Context::Context`

Sonicadvance1:
I followed that thread of logic. We already have a bunch of pointers in there from the initial relocation work a month ago.
Code sharing is close but not quite there.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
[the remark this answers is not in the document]
skmp: very helpful

PR #1689, opened 2022-05-08
Arm64Emitter: Optimize constants with ADRP and ADR

Sonicadvance1:
In a large number of cases we are moving pointers within a 4GB region
and some marginal pointers that are within 1MB.

This is only used in the case that MOVZ can't be used.

NOP padding still occurs after these instructions to ensure that if they
are being used with relocations it will still get padded to a full 4
instruction length.

Not all hardware fuses these and LLVM claims that Cortex beyond A72 even
doesn't, but it'll still be faster.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -42,12 +42,57 @@ void Arm64Emitter::LoadConstant(vixl::aarch64::Register Reg, uint64_t Constant, 
   }
 
   int NumMoves = 1;
-  movz(Reg, (Constant) & 0xFFFF, 0);
-  for (int i = 1; i < Segments; ++i) {
+  int RequiredMoveSegments{};
+
+  // Count the number of move segments
+  // We only want to use ADRP+ADD if we have more than 1 segment
+  for (size_t i = 0; i < Segments; ++i) {
     uint16_t Part = (Constant >> (i * 16)) & 0xFFFF;
-    if (Part) {
-      movk(Reg, Part, i * 16);
-      ++NumMoves;
+    if (Part != 0) {
+      ++RequiredMoveSegments;
+    }
+  }
+
+  // ADRP+ADD is specifically optimized in hardware
+  // Check if we can use this
+  auto PC = GetCursorAddress<uint64_t>();
+
+  // PC aligned to page
+  uint64_t AlignedPC = PC & ~0xFFFULL;
+
+  // Offset from aligned PC
+  int64_t AlignedOffset = static_cast<int64_t>(Constant) - static_cast<int64_t>(AlignedPC);
+
+  // If the aligned offset is within the 4GB window then we can use ADRP+ADD
+  // and the number of move segments more than 1
+  if (RequiredMoveSegments > 1 && vixl::IsInt32(AlignedOffset)) {
+    // If this is 4k page aligned then we only need ADRP
+    if ((AlignedOffset & 0xFFF) == 0) {
+      adrp(Reg, AlignedOffset >> 12);
+    }
+    else {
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -42,12 +42,58 @@ void Arm64Emitter::LoadConstant(vixl::aarch64::Register Reg, uint64_t Constant,
...
+
+  // PC aligned to page
+  uint64_t AlignedPC = PC & ~0xFFFULL;
+
+  // Offset from aligned PC
+  int64_t AlignedOffset = static_cast<int64_t>(Constant) - static_cast<int64_t>(AlignedPC);
+
+  // If the aligned offset is within the 4GB window then we can use ADRP+ADD
+  // and the number of move segments more than 1
+  if (RequiredMoveSegments > 1 && std::abs(AlignedOffset) < 0x7FFF'FFFFLL) {
```
skmp, line 68: Can we use vixl's `IsInt32` here?

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -42,12 +42,58 @@ void Arm64Emitter::LoadConstant(vixl::aarch64::Register Reg, uint64_t Constant,
...
+    // If this is 4k page aligned then we only need ADRP
+    if ((std::abs(AlignedOffset) & 0xFFF) == 0) {
```
skmp, line 70: abs isn't needed

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -42,12 +42,58 @@ void Arm64Emitter::LoadConstant(vixl::aarch64::Register Reg, uint64_t Constant,
...
+      adrp(Reg, AlignedOffset >> 12);
+    }
+    else {
+      // If the constant is within 1MB of PC then we can still use ADR to load in a single instruction
+      // 21-bit signed integer here
+      constexpr int64_t MAX_ADR_OFFSET = 0xF'FFFF;
+      int64_t SmallOffset = static_cast<int64_t>(Constant) - static_cast<int64_t>(PC);
+      if (std::abs(SmallOffset) <= MAX_ADR_OFFSET) {
```
skmp, line 78: Can we use vixl's `IsInt21` here?

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -42,12 +42,58 @@ void Arm64Emitter::LoadConstant(vixl::aarch64::Register Reg, uint64_t Constant,
...
+        adr(Reg, SmallOffset);
+      }
```
skmp, line 80:
```suggestion
        NumMoves = 1;
      }
```

skmp (approved):

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Resolved.

PR #1691, opened 2022-05-09
Object cache named region no-op implementation

Sonicadvance1:
Requires #1690 to be merged first.

The first 11 commits are from that PR, only the final two are new.

This sets up the named region async job interface in a no-op fashion.
The async interface itself is exercised while the async thread that is supposed to do loading and storing of these objects is a minimum no-op implementation.

Next PR will be for having the async thread actually loading and storing these object files.

### DIFF
--- a/External/FEXCore/Source/Interface/Context/Context.h
+++ b/External/FEXCore/Source/Interface/Context/Context.h
@@ -104,6 +104,7 @@ namespace FEXCore::Context {
       FEX_CONFIG_OPT(BlockJITNaming, BLOCKJITNAMING);
       FEX_CONFIG_OPT(ParanoidTSO, PARANOIDTSO);
       FEX_CONFIG_OPT(CacheObjectCodeCompilation, CACHEOBJECTCODECOMPILATION);
+      FEX_CONFIG_OPT(x87ReducedPrecision, X87REDUCEDPRECISION);
     } Config;
 
     using IntCallbackReturn =  FEX_NAKED void(*)(FEXCore::Core::InternalThreadState *Thread, volatile void *Host_RSP);
--- a/External/FEXCore/Source/Interface/Core/ObjectCache/CodeObjectSerializationConfig.h
+++ b/External/FEXCore/Source/Interface/Core/ObjectCache/CodeObjectSerializationConfig.h
@@ -1,0 +1,84 @@
+#pragma once
+
+#include <cstdint>
+
+namespace FEXCore::CodeSerialize {
+  // If any of the config options mismatch on load then the cache won't be used
+  // Any of these will result in codegen changes
+  struct CodeObjectSerializationConfig {
+		// Cookie in the header of the file, isn't part of the config hash
+    uint64_t Cookie{};
+
+    // Instructions per block configuration
+    int32_t MaxInstPerBlock{};
+
+    // Follows CPUID 4000_0001_EAX[3:0]
+    unsigned Arch : 4;
+
+    // Multiblock enabled
+    bool MultiBlock : 1;
+
+    // TSO enabled
+    bool TSOEnabled : 1;
+
+    // ABI local flag unsafe optimization
+    bool ABILocalFlags : 1;
+
+    // ABI no PF unsafe optimization
+    bool ABINoPF : 1;
+
+    // Static register allocation enabled
+    bool SRA : 1;
+
+    // Paranoid TSO mode enabled
+    bool ParanoidTSO : 1;
+
+    // Guest code execution mode (We don't support live mode switch)
+    bool Is64BitMode : 1;
+
+    // SMC checks style
+    unsigned SMCChecks : 2;
+
+    // x87 reduced precision
+    bool x87ReducedPrecision : 1;
+
+    // Padding to remove uninitialized data warning from asan
+    // Shows remaining amount of bits available for config
+    unsigned _Pad : 18;
+
+    bool operator==(CodeObjectSerializationConfig const &other) const {
+      return Cookie == other.Cookie &&
+        MaxInstPerBlock == other.MaxInstPerBlock &&
+        Arch == other.Arch &&
+        MultiBlock == other.MultiBlock &&
+        TSOEnabled == other.TSOEnabled &&
+        ABILocalFlags == other.ABILocalFlags &&
+        ABINoPF == other.ABINoPF &&
+        SRA == other.SRA &&
+        ParanoidTSO == other.ParanoidTSO &&
+        Is64BitMode == other.Is64BitMode &&
+        SMCChecks == other.SMCChecks &&
+        x87ReducedPrecision == other.x87ReducedPrecision;
+    }
+    static uint64_t GetHash(CodeObjectSerializationConfig const &other) {
+      // For < 64-bits of data just pack directly
+      // Skip the cookie
+      uint64_t Hash{};
+      Hash <<= 32; Hash |= other.MaxInstPerBlock;
+      Hash <<= 1;  Hash |= other.Arch;
+      Hash <<= 1;  Hash |= other.MultiBlock;
+      Hash <<= 1;  Hash |= other.TSOEnabled;
+      Hash <<= 1;  Hash |= other.ABILocalFlags;
+      Hash <<= 1;  Hash |= other.ABINoPF;
+      Hash <<= 1;  Hash |= other.SRA;
+      Hash <<= 1;  Hash |= other.ParanoidTSO;
+      Hash <<= 1;  Hash |= other.Is64BitMode;
+      Hash <<= 2;  Hash |= other.SMCChecks;
+      Hash <<= 1;  Hash |= other.x87ReducedPrecision;
+      return Hash;
+    }
+  };
+
+	static_assert(sizeof(CodeObjectSerializationConfig) == 16, "Size changed");
+  static_assert((sizeof(CodeObjectSerializationConfig) - sizeof(uint64_t)) == 8, "Config size exceeded 64its. Need to change how the hash is generated!");
+}
--- a/External/FEXCore/Source/Interface/Core/ObjectCache/JobHandling.cpp
+++ b/External/FEXCore/Source/Interface/Core/ObjectCache/JobHandling.cpp
@@ -13,11 +13,112 @@
 
 namespace FEXCore::CodeSerialize {
   void AsyncJobHandler::AsyncAddNamedRegionJob(uintptr_t Base, uintptr_t Size, uintptr_t Offset, const std::string &filename) {
-    // XXX: Actually add the add named region job
+    // This function adds a named region *JOB* to our named region handler
+    // This needs to be as fast as possible to keep out of the way of the JIT
+
+    auto BaseFilename = std::filesystem::path(filename).filename().string();
+
+    if (!BaseFilename.empty()) {
+      // Create a new entry that once set up will be put in to our section object map
+      auto Entry = std::make_unique<CodeRegionEntry>(
+        Base,
+        Size,
+        Offset,
+        filename,
+        NamedRegionHandler->DefaultCodeHeader(Base, Offset)
+      );
+
+      // Lock the job ref counter so we can block anything attempting to use the entry before it is loaded
+      Entry->NamedJobRefCountMutex.lock();
+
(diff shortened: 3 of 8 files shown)


External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
```diff
@@ -128,17 +235,102 @@ namespace FEXCore::CodeSerialize {
...
+      CodeObjectSerializationConfig const &GetDefaultSerializationConfig() const {
+        return DefaultSerializationConfig;
+      }
+
+    protected:
+      friend class AsyncJobHandler;
+
+      // Return a default code header based off the default serialization config
+      CodeObjectSerializationHeader DefaultCodeHeader(uint64_t Base, uint64_t Offset) const {
+        CodeObjectSerializationHeader Header{};
```
neobrain, line 270: This seems like a good place to use designated initializers. That way, it's harder to forget updating this code when adding new members in future changes, since the compiler can warn about it (e.g. gcc does this with `-Wextra`).
```suggestion
CodeObjectSerializationHeader Header{
        .Config = DefaultSerializationConfig,
        .OriginalBase = Base;
        .OriginalOffset = Offset;
        .NumCodeEntries = 0;
        .NumRelocationsTo = 0;
        .TotalRelocationsCount = 0;
};
```

External/FEXCore/Source/Interface/Core/ObjectCache/JobHandling.cpp
```diff
@@ -13,11 +13,112 @@
...
+      // Lock the job ref counter so we can block anything attempting to use the entry before it is loaded
+      Entry->NamedJobRefCountMutex.lock();
+
+      CodeRegionMapType::iterator EntryIterator;
+      {
+        std::unique_lock lk {CodeObjectCacheService->GetEntryMapMutex()};
+
+        auto &EntryMap = CodeObjectCacheService->GetEntryMap();
+
+        auto it = EntryMap.emplace(Base, std::move(Entry));
```
neobrain, line 40: The `!it.second` case can be slightly sped up by changing this to `EntryMap.emplace(Base, nullptr)`. The heap-allocating `std::make_unique<CodeRegionEntry>` only needs to be done in the `it.second == true` case then, which would explicitly overwrite the dummy entry created. (This assumes we can bear holding the lock a little bit longer in that case).

skmp:
This will have to be modified for #1558 as it replaces the NameRegion interface, though that should be fairly easy to do

External/FEXCore/Source/Interface/Core/ObjectCache/JobHandling.cpp
Sonicadvance1: I'm not sure I'm following what you mean here. We're not inserting a dummy entry. We're inserting an entry with the minimal amount of data required to kick off the async job.

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
Sonicadvance1: Resolved to use designated initializers.

External/FEXCore/Source/Interface/Context/Context.h
```diff
@@ -104,6 +104,7 @@ namespace FEXCore::Context {
       FEX_CONFIG_OPT(BlockJITNaming, BLOCKJITNAMING);
       FEX_CONFIG_OPT(ParanoidTSO, PARANOIDTSO);
       FEX_CONFIG_OPT(CacheObjectCodeCompilation, CACHEOBJECTCODECOMPILATION);
+      FEX_CONFIG_OPT(x87ReducedPrecision, X87REDUCEDPRECISION);
```
skmp, line 107: This seems unrelated?

skmp: ah used for context format. We should also add to AOTIR.

Sonicadvance1: It is necessary to have for `CodeObjectSerializationConfig` which pulls all of its configuration features from the Context Config location. Anything that effects codegen needs to be generated in that hash object.

External/FEXCore/Source/Interface/Core/ObjectCache/NamedRegionObjectHandler.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+namespace FEXCore::CodeSerialize {
+  NamedRegionObjectHandler::NamedRegionObjectHandler(FEXCore::Context::Context *ctx) {
+    DefaultSerializationConfig.Cookie = CODE_COOKIE;
+
+    // Initialize the Arch from CPUID
+    uint32_t Arch = ctx->CPUID.RunFunction(0x4000'0001, 0).eax & 0xF;
+    DefaultSerializationConfig.Arch = Arch;
+
+    DefaultSerializationConfig.MaxInstPerBlock = ctx->Config.MaxInstPerBlock;
```
skmp, line 14: `MaxInstPerBlock` and `Multiblock` also need to be included here?

skmp: Mixing them up should be fine, though we may or may not want that

Sonicadvance1: Only one missing here is Multiblock since Multiblock isn't supported with the code cache yet.

skmp: Mutliblock is actually included in this PR, though neither `MaxInstPerBlock` nor `Multiblock` should change our ABI.

`HostFeatures` makes sense.

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
```diff
@@ -128,17 +234,101 @@ namespace FEXCore::CodeSerialize {
...
+        ));
+        ++NamedWorkQueueJobs;
+      }
+
+    private:
+      // Code version. If the code emission changes then this needs to increment
+      constexpr static uint32_t CODE_VERSION = 0x0;
+
+      // Default cookie header for the file header
+      constexpr static uint64_t CODE_COOKIE = FEXCore::IR::COOKIE_VERSION("FEXC", CODE_VERSION);
```
skmp, line 310: Wouldn't "FEXO" be more consistent with IR's "FEXI" ?

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
```diff
@@ -128,17 +234,101 @@ namespace FEXCore::CodeSerialize {
...
+          Base,
+          Size,
+          std::move(Entry)
+        ));
+        ++NamedWorkQueueJobs;
+      }
+
+    private:
+      // Code version. If the code emission changes then this needs to increment
+      constexpr static uint32_t CODE_VERSION = 0x0;
```
skmp, line 307: It's gonna be tricky to keep this always updated, as minor codegen changes are common. How about using the git hash?

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
Sonicadvance1: Probably. It's just a cookie though, doesn't matter.

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
```diff
@@ -128,17 +234,101 @@ namespace FEXCore::CodeSerialize {
...
+          .TotalRelocationsCount = 0,
+        };
+      }
+
+      /**
+       * @brief Adds an asynchronous add named region work item to the object queue
+       *
+       * This adds the job that will do the loading of file resources and data tracking.
+       */
+      void AsyncAddNamedRegionWorkItem(const std::string &base, const std::string &filename, bool executable, CodeRegionMapType::iterator entry) {
```
skmp, line 284: Why do we need async versions of this?

Sonicadvance1: Because we can't block the JIT on IO.

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
Sonicadvance1: I was planning on release tag + incremental index but a git hash works just as well.

External/FEXCore/Source/Interface/Core/ObjectCache/ObjectCacheService.h
Sonicadvance1: sure JIT->MMAP->Named Region.
It's significant IO. Add it to an async job queue to stop blocking JIT execution.

skmp: I think we should design the system in a way that substantial IO doesn't need to happen on file load.

Performance wise it is probably better, and far easier to reason about, small stutters on code file mmaps vs other complications or unpredictable performance later on.

Sonicadvance1: The act of loading the index required to know what is in the file is enough IO to make the JIT slow down.
It's unfeasible to lower IO to the point that it doesn't impact JIT performance.

skmp: Can't the index be flatted down and then just mmap'd, similar to what AOTIR does?

skmp: Also WRT issue, it's probably best we keep separate "index" and "object code"/"ir code" files, as the indexes need to be rebuilt while the code files can just be appended to.

Sonicadvance1: The entire file is just mmap'd already. flattening won't work since this file is multi-process shared and these caches get incrementally appended from multiple processes.

skmp: Flattening can work just fine with multi process, why wouldn't it?

skmp: (can just use a red/black tree that uses local offsets instead of pointers)

Sonicadvance1: Well. It's not how I'm implementing it regardless.

External/FEXCore/Source/Interface/Core/ObjectCache/CodeObjectSerializationConfig.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include <cstdint>
+
+namespace FEXCore::CodeSerialize {
+  // If any of the config options mismatch on load then the cache won't be used
+  // Any of these will result in codegen changes
+  struct CodeObjectSerializationConfig {
```
skmp, line 8: Would be great to have this harmonized with the AOTIR format.

Sonicadvance1: Could probably share some of the config but not all of it. Once this config pulls in host features then there is a mismatch between what config effects IR versus which config options effects assembly.

skmp: Yeah, we could split into two parts one for "IR-abi" and one for "object-abi".

Would also be nice if we used the same postfix notations in the cache files in both.

There's a further slight complication there with #1558, `MMAN` and `MTRACK` smc modes don't change the ABI at all.

skmp: (But are different `SMCChecks` values)

Sonicadvance1: That's fine, doesn't need to be flawless.

PR #1693, opened 2022-05-09
CMake: Add option to use the mold linker

neobrain:
`mold` is a modern linker that runs faster than either `ld` or `lld`. With these changes, FEX makes use of it via CMake option `ENABLE_MOLD=ON`. (Works on clang and *recent* versions of gcc.)

(Also changed `LLD` to use the upstream spelling of `lld`, which is more consistent with `mold` also being all-lowercase.)

TODO:
* [ ] Should the `ENABLE_STATIC_PIE` path also be disabled for mold? I couldn't test this path so I don't know if it works with mold.


Sonicadvance1 (approved):
Seems fine to ignore the static pie problem for now.
It's not really supported anymore.

PR #1700, opened 2022-05-10
Standarized TODO markers: FEX_TODO, FEX_TODO_ISSUE

### LINKED ISSUE #1696: Define standarised TODO markers
We can define 
```
TODO_ISSUE(issuenumber, username, comment)
TODO_ISSUE(issuenumber, comment)
TODO(username,comment)
TODO(comment)
```
For a more structured way of handling TODOs

skmp:
Proposal for #1696
clang-tidy already has support for detailed todo comments with the format
```
//TODO(details): comment
```
which seems to be the standardised google way to handle this.

I've defined ours as `FEX_TODO` and `FEX_TODO_ISSUE` for easier grepping.

Use like
#### For TODOs
- FEX_TODO_ISSUE(github ticket number, "comment")
- FEX_TODO("comment");

#### For non-todo linking to tickets
- FEX_TICKET(github ticket number)
- FEX_TICKET(github ticket number, "comment")

`#define FEX_WARN_TODO 1`  will generate warnings. That's arguably not very useful as one can just grep for `FEX_TODO`.

Thoughts?

(Closes #1696)

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -41,6 +41,7 @@ $end_info$
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/Utils/Threads.h>
 #include <FEXHeaderUtils/Syscalls.h>
+#include <FEXHeaderUtils/TodoDefines.h>
 
 #include <algorithm>
 #include <array>
@@ -1016,7 +1017,8 @@ namespace FEXCore::Context {
     }
 
     // If it is the parent thread that died then just leave
-    // XXX: This doesn't make sense when the parent thread doesn't outlive its children
+    FEX_TODO("This doesn't make sense when the parent thread doesn't outlive its children");
+
     if (Thread->ThreadManager.parent_tid == 0) {
       CoreShuttingDown.store(true);
       Thread->ExitReason = FEXCore::Context::ExitReason::EXIT_SHUTDOWN;
--- a/FEXHeaderUtils/FEXHeaderUtils/TodoDefines.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/TodoDefines.h
@@ -5,15 +5,17 @@
 #endif
 
 #if FEX_WARN_TODO
-// Use like FEX_TODO_ISSUE(github ticket number, username, comment) or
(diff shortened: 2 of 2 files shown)


skmp:
Updated usage comments, now they are prefixed by `FEX_` like the macros name. Will merge once build passes.

neobrain:
Could you drop the username? We've done this in Citra, and it never really was all that useful compared to git-blame. Dropping the username allows for implementing these defines as non-variadic macros, which makes them easier to use consistently, and it reduces code noise.

I'm not a fan of this in general, but I don't strongly care either. If people adopt this, I'll happily go along.

neobrain:
> Git often loses author/history though, especially with large formatting or file splitting operations. I'm not certain if this is an issue in practice.

That was the reasoning in Citra too, but in practice this problem affected very few TODOs. Reformats (hopefully) don't happen daily, so often by the time someone had to trace back a TODO, the authors themselves had forgotten about the details already anyway (or moved on to other projects).

> Why not? I haven't heavily used this kind of approaches in the past so I don't have much experience with pros/cons. It does seem better than having the codebase sprinkled with non-standardized forms of todos and notes.

Adding TODOs in this way requires having a specific header included, which discourages its use.
Similarly, the macros as implemented in this PR can't be used for TODOs relating to struct members or top-level definitions. (I'm not sure if there's a way to reconcile the second point without adding another set of macros, which... ew)

skmp:
> Adding TODOs in this way requires having a specific header included
I did think of that, and I want (possibly as a follow up) to make a `FexGlobals.h` that gets automatically included in every file via `-include` in the targets we want to

> Similarly, the macros as implemented in this PR can't be used for TODOs relating to struct members or top-level definitions. (I'm not sure if there's a way to reconcile the second point without adding another set of macros, which... ew)

Oh the do/while empty statement. If we're comfortable with not requiring an `;` at the end, I'm fine with that. That was the original design locally actually :P. I can also make it generate an empty pragma to require the `;` now that I think of it.

skmp:
Updated to remove author, added `FEX_TICKET` for non todo, removed `do { } while (0)` so they can be used anywhere in the soure.

Sonicadvance1 (approved):

neobrain (approved):

PR #1702, opened 2022-05-12
Arm64: Adds support for RCPC2 extension

Sonicadvance1:
This allows us to have RCPC loadstore operations with a 9-bit signed
offset.

This gives us a small range of [-256,256) of immediate encoding range on
our TSO loadstore operations.
Updates the inline constant pass in ConstProp to support this range on
TSO IR ops if the host supports RCPC2.

Apple M1 supports this extension, didn't test with Cortex-X2/A710.

PR #1703, opened 2022-05-12
Arm64: Makes unaligned backpatching be less aggressive with DMB

Sonicadvance1:
No need to do a full inner-shareable read/write DMB on both sides of the
instruction.
We only need to do inner-shareable read OR write depending on the type
of loadstore.

Couldn't find a bench where this made a significant difference but
surely there is one somewhere.

First commit is from https://github.com/FEX-Emu/FEX/pull/1702, needs that merged first before this.


Sonicadvance1:
This might need more testing and sanity checking btw.

skmp:
Maybe a good idea to hide it behind an optimization flag as a first step then?

Sonicadvance1:
I'm going to discard this for now and come back to it later.
Thinking about this makes my head hurt and it's a small percentage if any that this improves.

PR #1705, opened 2022-05-13
32-bit FSGS instruction fixes.

Sonicadvance1:
Just a minor thing I noticed.


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5944,6 +5942,15 @@ constexpr uint16_t PF_F2 = 3;
     {OPD(FEXCore::X86Tables::TYPE_GROUP_P, PF_66, 0), 8, &OpDispatchBuilder::NOPOp},
     {OPD(FEXCore::X86Tables::TYPE_GROUP_P, PF_F2, 0), 8, &OpDispatchBuilder::NOPOp},
   };
+
+  constexpr std::tuple<uint16_t, uint8_t, FEXCore::X86Tables::OpDispatchPtr> SecondaryExtensionOpTable_64[] = {
```
skmp, line 5946: `SecondaryExtensionOpTable_64` -> `SecondaryExtensionOpTable_segment64` ?

skmp (approved):

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: The name is following the other table's names. Suffix of `_32` and `_64` depending on if those instructions are available on 32-bit or 64-bit exclusively.
We don't have 100% compliance on this and mix some of the handlers. Once we audit these tables again this table will grow to more than just the segment instructions.

skmp: kk

PR #1706, opened 2022-05-14
FEXCore: Adds refcount_shared_mutex class

Sonicadvance1:
This class is similar to std::shared_mutex except it is safe for the
same thread to increment or decrement the ref counter multiple times.

This can be passed to regular std locks.

This will be required with the code object cache service soon.

### DIFF
--- a/External/FEXCore/include/FEXCore/Utils/refcount_shared_mutex.h
+++ b/External/FEXCore/include/FEXCore/Utils/refcount_shared_mutex.h
@@ -1,0 +1,265 @@
+#pragma once
+#include <FEXCore/Utils/LogManager.h>
+
+#include <atomic>
+#include <linux/futex.h>
+#include <sys/syscall.h>
+#include <shared_mutex>
+#include <unistd.h>
+
+namespace FEXCore::Utils {
+  /**
+   * @brief This class is similar to std::shared_mutex but is safe to shared lock multiple times from the same thread.
+   *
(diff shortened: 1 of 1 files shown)


skmp:
Can you provide some more high level information on the high level implementation of this? Looks like something we want to thoroughly review.

Is this like a marriage of `std::recursive_mutex` and `std::shared_mutex` ?

Sonicadvance1:
It's quite literally `std::shared_mutex` that has defined behaviour around the same thread incrementing/decrementing the shared lock counter multiple times. Also different threads doing `unlock_shared` from a different thread having done a `lock_shared`. `std::shared_mutex` has undefined behaviour when the same thread does `lock_shared()` multiple times, which sanitizers complain about.

External/FEXCore/include/FEXCore/Utils/refcount_shared_mutex.h
```diff
@@ -0,0 +1,327 @@
...
+          }
+
+          return false;
+        };
+
+        return TryRefIncrement();
+      }
+
+      void lock_shared() {
+        auto TryRefIncrement = [this]() -> bool {
```
neobrain, line 184: This seems to be duplicated across many of the member functions. Why not move it to a private member function (generalizing it a bit if need be?)

Sonicadvance1: Changed over these lambdas to private member functions. Resolved.

PR #1718, opened 2022-05-19
X86Tables: Leave shouldn't end block

skmp:
Run into this while debugging on `skmp/guest-rt_sigreturn`. Fixes the X86Tables entry for `leave`


Sonicadvance1 (approved):
Oops. Probably a leftover from the initial table generating and I thought it was did a RET on this instruction.

PR #1720, opened 2022-05-20
FEXLogServer: Work around a libstdc++ bug

Sonicadvance1:
We can't use std::erase_if here because for some reason the developers
decided to const qualify the definition for std::map.

Works around that issue.


Source/Tools/FEXLogServer/Main.cpp
```diff
@@ -121,10 +121,24 @@ namespace GUI {
 
   static void FDClosedHandler(int FD) {
     std::unique_lock lk{PIDMappingMutex};
+    // Due to a bug in libstdc++12 headers we can't use std::erase_if here.
```
neobrain, line 124: I think this is not actually a bug. Predicates for std algorithms are typically required to not modify their arguments, but this is poorly enforced. Apparently libstdc++12 changed this by forcing the predicate to be called with a `const` argument. So the type of `Entry` is actually `const std::pair<...>`, hence you can't call `erase` on it.

Sonicadvance1: Looks like you're right. Changed the commit message and removed the `#if 0` code.

A bit spooky that this started to get enforced with an seemingly unrelated commit in libstdc++.
```
commit 5f40d34b6dd30ed076f9beb9a88e7ec264c1a55a
Author: François Dumont <fdumont@gcc.gnu.org>
Date:   Tue Nov 16 21:51:11 2021 +0100

    libstdc++: [_GLIBCXX_DEBUG] Reduce performance impact on std::erase_if

    Bypass the _GLIBCXX_DEBUG additional checks in std::__detail::__erase_node_if used
    by all implementations of std::erase_if for node based containers.

    libstdc++-v3/ChangeLog:

            * include/bits/erase_if.h (__erase_nodes_if): Add _UnsafeContainer template
            parameter. Use it to get iterators to work with.
            * include/debug/macros.h (__glibcxx_check_erase2): New.
            * include/debug/map.h (map<>::erase(_Base_const_iterator)): New.
            (map<>::erase(const_iterator)): Use latter.
            * include/debug/multimap.h (multimap<>::erase(_Base_const_iterator)): New.
            (multimap<>::erase(const_iterator)): Use latter.
            * include/debug/multiset.h (multiset<>::erase(_Base_const_iterator)): New.
            (multiset<>::erase(const_iterator)): Use latter.
            * include/debug/set.h (set<>::erase(_Base_const_iterator)): New.
            (set<>::erase(const_iterator)): Use latter.
            * include/debug/unordered_map (unordered_map<>::erase(_Base_const_iterator)): New.
            (unordered_multimap<>::erase(const_iterator)): New.
            * include/debug/unordered_set (unordered_set<>::erase(_Base_const_iterator)): New.
            (unordered_multiset<>::erase(const_iterator)): New.
            * include/experimental/map (erase_if): Adapt.
            * include/experimental/set (erase_if): Adapt.
            * include/experimental/unordered_map (erase_if): Adapt.
            * include/experimental/unordered_set (erase_if): Adapt.
            * include/std/map (erase_if): Adapt.
            * include/std/set (erase_if): Adapt.
            * include/std/unordered_map (erase_if): Adapt.
            * include/std/unordered_set (erase_if): Adapt.
```

PR #1721, opened 2022-05-20
Syscalls: Fixes clone3 stack pointer

Sonicadvance1:
clone2 stack pointer passed in points to the highest address for the
stack.

clone3 switches this around and gives us a base pointer and a size.

glibc started using clone3 for its thread cloning which finally caught
this bug. Necessary to run any application under the Ubuntu 22.04 rootfs
since that uses a new enough glibc to encounter this.


skmp:
Cute

skmp (approved):

PR #1722, opened 2022-05-21
Fix dangling curl hyphen

Hypnotron:
Fixes a regression in fa87c73b9ee60a334eace2cdc3097725cbaf5b88; curl complains "curl: option -: is unknown" when trying to fetch a RootFS without this.


Sonicadvance1 (approved):
Oops! Thanks!

Issue #1682, opened 2022-05-04
Signal safety


skmp:
Splitting from #1558 & #1677, as well as discussions with @neobrain  and @Sonicadvance1.

## The issues
(a) Signals can interrupt the JIT compiler or syscall or other FEX-related code, which are not guaranteed to be signal re-entrant safe. Any code that touches non-stack memory, or uses mutexes is possibly not signal safe. We currently block signals around some code, either using `ScopedSignalMaskWith*` guards or manually (eg, the dispatcher disabling signal handling around calls to CompileCode)

(b) Signals can interrupt the translated code in the middle of operations that would normally be atomic wrt signals. This may or may not be a problem, depending on how we have implemented x86. A good example is REP* operations. This can be an issue even without LSE elimination, as the recovered guest state might be "teared".

(c) Similar to above, signals can interrupt the translated code in places where we can't recover the guest architectural place, due to optimisations.

## Current status
Our current "signal safety strategy" for (a) is to sprinkle signal disabling code around regions that deadlock. This is very inconsistent throughput the codebase, and there are several bugs waiting to be hit. In general, this is a compromise between "likely to lockup" and "performant code". 

For (b) and (c) we currently only partially recover the guest architectural state, store it alongside the host architectural state, and hope the guest code doesn't care too much about the contents of the guest state, and that it will not modify it. We depend on returning to the interrupted host code using the stored host architectural state, in order to resume execution in the middle of any teared instructions, and eventually exit from some point with a valid guest state. This poses another limitation, that the interrupted block cannot be discarded from the code cache, so the code cache cannot be cleared. This might also have further implications around SMC and code invalidations.

## Proposed solutions
For (a) I'd like us to have clear guidelines on how to handle this, as well as a mode that might be slower but offers guaranteed stability. This needs some thought, but is not too hard.

For (b) and (c) the only viable solution I can think of is a combination of deferring the signal delivery until we have a fully recoverable guest state, and storing metadata that can help us exit from the middle of a block. (c) Can be avoided by limiting store elimination from LSE and disabling DSE. We can have a tradeoff between "defer delay" vs "runtime performance".

Related: #518, #650, #1228, #1666

skmp:
(Updated with more details about signal latency, execution overshoot, and signal queue merging)

skmp:
[...]

PR #1725, opened 2022-05-24
WIP: Guest Signals: Correctly block raised signal, implement SA_NODEFER, sigreturn sa_mask restore

skmp:
#### Overview

This is the first set of fixes from #1682

- The raised signal in the guest handler would not be blocked. As the host handler does sigreturn, we can't rely on the host handling there.
- Also implemented guest SA_NODEFER there.
- On guest signal return, the guest provided `uc_sigmask` is restored, not the one saved on signal entry, as the guest can modify this.

#### TODO
- [ ] Cleanup the code
- [ ] move more of the signal logic away from the jit dispatcher
- [ ] Make sure we don't set SA_NODEFER for the host signals, as we don't want that there
- [ ] Integrate tests from https://github.com/FEX-Emu/fex-assorted-tests-bins/tree/main/src/tests

#### Test todo
- [ ] Add test for procmask inside signal handler, with and without SA_NODEFER
- [ ] Add test for procmask after signal handler, with and without SA_NODEFER
- [ ] SA_NODEFER tests validating around host signal handling
- [ ] Verify guest_uctx sigmask is set correctly before entry to signal handler

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
@@ -23,7 +23,7 @@ struct X86ContextBackup {
   // RIP and RSP is stored in GPRs here
   uint64_t GPRs[23];
   FEXCore::x86_64::_libc_fpstate FPRState;
-  uint64_t sa_mask;
+  uint64_t sigmask;
   bool FaultToTopAndGeneratedException;
 
   // Guest state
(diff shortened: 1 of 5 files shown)


skmp:
Interestingly enough, we did have a (partially working) implementation of half the fixes there, that somehow failed to work during my tests, but works now? Peculiar.

The signal mask setting still needs to be fixed though.

skmp:

PR #1727, opened 2022-05-25
FEXGlobalIncludes.h: Automatically included via cmake

skmp:
#### Overview
Follow up from #1700, introduces `FEXHeaderUtils/FEXGlobalIncludes.h`, modifies CMake scripts to automatically include ot.

`FEXHeaderUtils/FEXGlobalIncludes.h` just includes `TodoDefines.h` for now, and we should be /very/ careful about things we forcefully include

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -478,6 +478,7 @@ endif()
 
 add_subdirectory(FEXHeaderUtils/)
 include_directories(FEXHeaderUtils/)
+add_definitions(-include FEXHeaderUtils/FEXGlobalIncludes.h)
 
 add_subdirectory(External/FEXCore)
 
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -42,7 +42,6 @@ $end_info$
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/Utils/Threads.h>
 #include <FEXHeaderUtils/Syscalls.h>
-#include <FEXHeaderUtils/TodoDefines.h>
 
 #include <algorithm>
 #include <array>
--- a/FEXHeaderUtils/FEXHeaderUtils/FEXGlobalIncludes.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/FEXGlobalIncludes.h
@@ -1,0 +1,4 @@
+// This is included via CMake for FEXCore, others
+#pragma once
+
+#include "TodoDefines.h"


lioncash:
What's the reason for not just specifying headers explicitly where they're needed? Having global includes seems quite fragile overall

skmp:
This is just to make `FEX_TODO`, `FEX_TODO_ISSUE` & `FEX_TICKET` documentation macros available to the entire codebase, without having to include them in every file.

@neobrain mentioned that the friction of having to include them will make us not use them, and I do agree on that.

In general, I don't think we'll want to include anything else globally, but just for nicety I made a new header instead of including `TodoDefines.h` directly

neobrain:
This abuse of `add_definitions` worsens developer UX, since IDE parsers can't pick up this implicit include. In local testing, `clangd` underlines `FEX_TODO` as an unknown symbol. I'd rather not have false-positive error markers when browsing source files.

Another issue is that adding C++-specific code to the headers breaks the build. This is because the header is pulled into *all* targets compiled by FEX, including C-only ones like jemalloc. You can reproduce this locally by adding `#include <cstdint>` to FEXTodo.h, which will break the jemalloc build.

I'm not a fan of having to include FEXTodo.h explicitly just to use the `FEX_TODO` macros, but adding CMake hacks to work around that is worse IMO, particularly when they have unexpected side effects like this.

skmp:
> This abuse of add_definitions
I wouldn't call it abuse or a hack as that's how CMake expects us to use this feature. CMake is notoriously slow at introducing higher level options for these widely supported features, sadly.

> worsens developer UX, since IDE parsers can't pick up this implicit include.
Sounds like an IDE issue. Open a PR for clangd? ;)

> In local testing, clangd underlines FEX_TODO as an unknown symbol. I'd rather not have false-positive error markers when browsing source files.
I still think that's far better UX than having to include the file manually. All IDEs i've used with FEX get highly confused by our source code anyway, so you can't tell if things are actually broken without a real build

> Another issue is that adding C++-specific code to the headers breaks the build. This is because the header is pulled into all targets compiled by FEX, including C-only ones like jemalloc. You can reproduce this locally by adding #include <cstdint> to FEXTodo.h, which will break the jemalloc build.
No C++ code is included, and we can guarantee that it is never included or hidden behind `__clpusplus`

skmp:
Poking this for an update. My point of view is, unless we have a better way to achieve this, this should be merged.

skmp:
> > Do you have a better proposal?
> 
> My proposal is to opt for other means of highlighting TODOs than this macro-based approach.

What other means?


> I could be convinced that this PR's impact on developer UX isn't that bad if it turns out to only affect one out of many IDE engines with nontrivial code analysis, but I don't consider the benefits of this approach big enough to investigate into this myself.

For me, no IDE I've tried works correctly for FEX, so I don't think I'll notice anything changed with this PR. Which IDE are you using?

neobrain:
> What other means?

A social convention to use the format "//TODO: comment" seems fine to me.

> For me, no IDE I've tried works correctly for FEX, so I don't think I'll notice anything changed with this PR. Which IDE are you using?

Qt Creator has no problems parsing FEX, and I would expect CLion and VSCode to work equally well. (Those are 3 out of the 5 most popular IDEs according to the Meeting C++ surveys.)

phire:
VSCode works for me, as long as you remember to point it at compileCommands.json

At least, it's not reporting invalid problems. I've absolutely had problems with it pegging CPU at 100% until I restarted, or delete the cache.

skmp:
@FEX-Emu/maintainers bump on this. 

- Do we want this in?
- Do we want some alternative implementation?
- Do we give up on structured TODOs because we can't agree on them and also rip out https://github.com/FEX-Emu/FEX/pull/1700 as it is mostly useless without this?

skmp:

Issue #267, opened 2020-07-10
Libraries we need to Thunk


Sonicadvance1:
Libraries we need for performance reasons
- libGL
- libz
- libpng
- libSSL (mbedTLS and OpenSSL?)

Libraries we need for compatibility reasons
- steam client libraries?
- libGLX?

Add more here

skmp:
(Removed myself as I'm not in thunks for now)

Sonicadvance1:
libm would be interesting

Sonicadvance1:
libopenal?

Sonicadvance1:
libfmod might be interesting as closed source to closed source thunking support.

Sonicadvance1:
libvulkan directly instead of the device drivers.
At the very least will resolve a hang in current vulkan thunking but also will mean we don't need to generate a device thunk per driver.
A lot of the work for the device drivers can be reused for the libvulkan loader.

skmp:
After some more thorough investigation on how GLX and EGL are implemented under the hood, it looks like we don't need to thunk X11 or Wayland by doing a secondary connection to the display server for GLX or EGL purposes.

While I thought that doing this on GLX is a bit of a hack, EGL will use parallel connections both for X11 and Wayland to handle `eglGetDisplay(EGL_DEFAULT_DISPLAY)` and uses `dri2_egl_display::own_device` to track this.

For X11, see https://github.com/mesa3d/mesa/blob/6f4b5fb67507b607d61ab8316dfa3bbc4bab6498/src/egl/drivers/dri2/platform_x11.c#L1327.

For wayland, see https://github.com/mesa3d/mesa/blob/6f4b5fb67507b607d61ab8316dfa3bbc4bab6498/src/egl/drivers/dri2/platform_wayland.c#L2681.

Also, detecting the platform from the native display is done in an interesting (scary?) way we can reuse, https://github.com/mesa3d/mesa/blob/3415bf0268884bf3c16d495a4e1afd30d67bc71d/src/egl/main/egldisplay.c#L129

skmp:
Chromium, and hopefully no one other, buypasses libGLX and directly assembles the GLX X11 requests and sends them over with xcb.

Chromium's `Glx::GetFBConfigs` that hand assembles the GLX request https://github.com/chromium/chromium/blob/main/ui/gfx/x/generated_protos/glx.cc#L2015

Chromium's `Connection::SendRequest` that uses xcb to send raw requests
[...]

PR #1728, opened 2022-05-25
WIP: x11 independent glx thunk

skmp:
This is very much hack and slash.

Main idea is: Use a host display/connection to do the GLX operations, so that we don't need to interface/thunk with the guest x11 side.

Very limited amount of pain as XIDs are system global. Some things need to be allocated & copied to the guest for XFree.

It should work with all mesa drivers, as well as nvidia and others, though it's only been tested with my RX580.

Will pass `glxinfo` and `glxgears`, now trying steam.


skmp:
steam complains about `Couldn't set video mode: Failed loading libGL.so.1: /lib/x86_64-linux-gnu/libGL.so.1: undefined symbol: XGetVisualInfo`

skmp:
Fixed the `XGetVisualInfo` issue, added egl support, fixed symbol exports and reuse, other misc fixes.

skmp:
Chromium & CEF buypass libGLX and send GLX server requests directly using xcb's `REQUEST_RAW` (see https://github.com/FEX-Emu/FEX/issues/267#issuecomment-1140221644)

We probably want to dig a bit further into the exact GLX/EGL architectures, as we may have to create a more conforming GLX client implementation.

I'm not sure that's the reason chrome fails to initialize GL, but it sure is spooky.

skmp:
Closing this in favor of #1734 and #1735

PR #1734, opened 2022-05-29
RFC: ThunkLibs: Ideas for changes from the FGL branch

skmp:
#### Overview

Quick and dirty changes I did

- C++17 for `gen`
- no_host_impl, no_guest_export attributes for `gen`
- changes so that symbols are by default hidden, and their aliases are public
- changes so that impls need to have mangled internal names to avoid symboltable overrides
- switches `add_guest_lib`/`add_host_lib` to use optional arg parsing
- kind-of-redefines the meaning of LIBNAME

This is meant to start a discussion, not to be a concrete impl.


skmp:
@Sonicadvance1 this one does need to be heavily re-worked, pending on coordination with @neobrain

neobrain:
As mentioned in other channels, what I need to review your ThunkLibs PRs is better separation of changes and documentation. The bare minimum needed is a rudimentary explanation of what problem each of the patches solves. Otherwise, we have no solid basis for discussion and I won't be able to gauge how the changes will impact my ongoing work.

I won't repeat another review like #1809 where I had to spend a full week until I understood what you're doing and actually got it into a working state. Since I'm continuing to drive forward the bigger-scheme thunking work, I don't have capacity to carry early-state PRs out of prototype state.

skmp:
Following up on today's discussion

Main rationale around the PR is
- Export only symbols defined on the interface files from the guest thunk, and hide all others to avoid exposing thunk lib internals to the guest.
- provide a way to control exports (no_host_impl for guest-only export, no_guest_export for internal thunk symbols)
- Separates local symbol definition from exported Symbols, so that Symbol overloads don't affect thunking internals. Ie, a library defining glGetString not affecting glGetProc("glGetString") return value. This is needed for GL and possibly other libs. In the PR setup, iirc, `IMPL(name)` can be used to refer to the internal thunk symbol (never overriden), and `name` can be used for the export symbol (may be overridden by other libs or LD_PRELOAD)

@neobrain easier to digest this way?

skmp:
> > Export only symbols defined on the interface files from the guest thunk, and hide all others to avoid exposing thunk lib internals to the guest.
> 
> Sounds good. Did exporting those functions cause any specific issues?

I had some libs highjack symbols by accident, so figured we want to replicate 1:1 the guest lib exports, and not include our own noise/


> 
> > provide a way to control exports (no_host_impl for guest-only export, no_guest_export for internal thunk symbols)
> 
> What is this needed for?

As the interface files also generate exports (as per the above feature), we need some way to avoid that (explicit guest-only export, or internal that doesn't need to be exported). We can automate this with the `_internal` or `::internal` magic most likely.


> > Separates local symbol definition from exported Symbols, so that Symbol overloads don't affect thunking internals. Ie, a library defining glGetString not affecting glGetProc("glGetString") return value. This is needed for GL and possibly other libs.
> 
> What specific problem does the current situation cause for GL thunking?

if glGetProc("glGetString") returns `&glGetString`, and the main application also defines `glGetString`, we will return the main application's `glGetString`, not the thunking library one. FTL runs into this, though with GCH FTL doesn't assert that bug anymore

> 
> Overall sounds like these changes are well-motivated. Could you reflect the different use cases you're addressing in the PR structure, so your solution to address them can be reviewed as independent patches?

neobrain:
> Yes, of course, the PR was just to propose the ideas and make sure we are aligned. Will chop them as non-rfc ones.

Great, cheers!

> As the interface files also generate exports (as per the above feature), we need some way to avoid that (explicit guest-only export, or internal that doesn't need to be exported). We can automate this with the _internal or ::internal magic most likely.

That motivates `no_guest_export`, but `no_host_impl` I'm not too sure about (if you don't have a host-side thunk, why have an entry in the interface file at all?). Maybe it'll become clear in the isolated patch.

> if glGetProc("glGetString") returns &glGetString, and the main application also defines glGetString, we will return the main application's glGetString, not the thunking library one. FTL runs into this, though with GCH FTL doesn't assert that bug anymore

Wouldn't passing the `RTLD_DEEPBIND` flag to `dlopen` achieve the same purpose in a less invasive way? (Never used that flag, but judging from documentation it seems made specifically to avoid this issue)

skmp:
> > As the interface files also generate exports (as per the above feature), we need some way to avoid that (explicit guest-only export, or internal that doesn't need to be exported). We can automate this with the _internal or ::internal magic most likely.
> 
> That motivates `no_guest_export`, but `no_host_impl` I'm not too sure about (if you don't have a host-side thunk, why have an entry in the interface file at all?). Maybe it'll become clear in the isolated patch.

Maybe you want to implement this guest side only, and/or use an `_internal` symbol to do the work.

> > if glGetProc("glGetString") returns &glGetString, and the main application also defines glGetString, we will return the main application's glGetString, not the thunking library one. FTL runs into this, though with GCH FTL doesn't assert that bug anymore
> 
> Wouldn't passing the `RTLD_DEEPBIND` flag to `dlopen` achieve the same purpose in a less invasive way? (Never used that flag, but judging from documentation it seems made specifically to avoid this issue)

No, because this happens in the relocation phase, not during `dlsym`. Think of it as, host GL library doesn't return `&glGetString`, it returns an internal, unnamed symbol.

It is the `unordred_map` that gets to have wrong values, if it is initialized with `&glGetString`.

neobrain:
> No, because this happens in the relocation phase, not during dlsym. Think of it as, host GL library doesn't return &glGetString, it returns an internal, unnamed symbol.
> 
> It is the unordred_map that gets to have wrong values, if it is initialized with &glGetString. Yes, an external lib can hijack your local symbols in the posix model.

I wasn't referring to `ldr.inl`, but I should have been more specific: Using that flag in the `dlopen` call *in Thunks.cpp* should do the job, shouldn't it? That way other libraries wouldn't interfere with local definitions.

Though, I suppose, we definitely want to set that flag *also* for the dlopen calls in `ldr.inl`, since we can run into that very same issue there IIUC.

skmp:
> > No, because this happens in the relocation phase, not during dlsym. Think of it as, host GL library doesn't return &glGetString, it returns an internal, unnamed symbol.
> > It is the unordred_map that gets to have wrong values, if it is initialized with &glGetString. Yes, an external lib can hijack your local symbols in the posix model.
> 
> I wasn't referring to `ldr.inl`, but I should have been more specific: Using that flag in the `dlopen` call _in Thunks.cpp_ should do the job, shouldn't it? That way other libraries wouldn't interfere with local definitions.
> 
> Though, I suppose, we definitely want to set that flag _also_ for the dlopen calls in `ldr.inl`, since we can run into that very same issue there IIUC.

Again, the issue is not there.

The issue is in the guest library, in our case
```
const std::unordered_map<std::string_view, uintptr_t /* guest function address */> HostPtrInvokers =
    std::invoke([]() {
#define PAIR(name, unused) Ret[#name] = reinterpret_cast<uintptr_t>(GetCallerForHostThunkFromRuntimePointer<fexthunks_libGL_hostcall_##name>(name));
        std::unordered_map<std::string_view, uintptr_t> Ret;
        FOREACH_internal_SYMBOL(PAIR)
        return Ret;
#undef PAIR
    });
```

Right now we are just using GHC, so this bug doesn't manifest. But let's assume we used there `&glGetString` instead of `(GetCallerForHostThunkFromRuntimePointer<fexthunks_libGL_hostcall_##name>(name));`, as it is to happen in the case of a custom implementation.

due to how symbols are resolved by the guest dynamic linker. `&glGetString` may not be pointing to the one defined by `libGL-guest.so`. As such, our `glGetProc...("glGetString")` would return something wrong.

neobrain:
> guest dynamic linker

Ah, I was mixing up a couple of things there, but yes of course we can't actually override the `dlopen` flags for the guest linker easily...

skmp:
> > guest dynamic linker
> 
> Ah, I was mixing up a couple of things there, but yes of course we can't actually override the `dlopen` flags for the guest linker easily...

Yeah, it is a quirky situation. Took me a while to wrap my head around it when I first saw that happening in FTL.

skmp:
@neobrain now that we have reached clarity on the exact issues, can we get an agreement on how to solve them?

skmp:

PR #1735, opened 2022-05-29
RFC: ThunkLibs/FGL: New GL/EGL/GLX thunks with parallel X11 display connections

skmp:
#### Overview
GLX, EGL and GL/GLES thunks that use a "parallel" host-side connection to the same X server the guest is connected to. As XIDs and such are globals on X11, this works. There is no requirement to thunk XLib or XCB for this to work.

A few mapping tables are kept between "host" and "guest" `Display*`s, and some structures are marshaled (XVisualInfo is re-queried from the guest connection, arrays are copied to guest allocations so that `XFree()` works on them).

Overall, this approach works for many things, including glmark2, GeeXLab, FTL and chrome.

Depends on #1734.

#### Details / A couple things to note / Follow ups

- This could use custom/raw X11 GLX packets -a la chrome- to implement some functions purely on the guest side
- This misses some gl functions that were not fully declared in my system headers
- We should switch to generate the GL/EGL functions from the xml specs
- Callback* GL functions are not handled. This is pending on further thunks work from @neobrain 
- Implementation is derived from, but not unified with the previous GL thunks
- glX*wait* and egl*wait* might be partially working
- XVisualInfo is not re-queried from Guest -> Host, only from Host -> Guest
- This could be split to separate GL/GLX/EGL libs, with some effort
- This assumes `glxGetProc` == `eglGetProc` for GL names
- This could try first to dlsym(libGL/ES/...) then `glxGetProc` to avoid excessive shim generation by MESA/GLX. Some mobile mon-mesa drivers have a limit of 255 shims or such.
- There is an edge case that can lead to application deadlock during teardown, that is also triggered by glxgears. This reveals an application bug, and I'm not sure there's anything we can do on our side to fix it. More details in: https://gitlab.freedesktop.org/xorg/lib/libxcb/-/issues/38


skmp:
After some thinking today, I realized we can merge the FGL/GL thunks, make X11 thunks an optional dependency, and automatically pick the correct thunking strategy by looking if X11 is thunked or not at runtime.

skmp:
(Closing this in favor of #1817)

PR #1737, opened 2022-06-01
unittests: Add FEXLinuxTests with a few tests

skmp:
#### Overview
FEXLinuxTests, derived from https://github.com/FEX-Emu/fex-assorted-tests-bins. I've only added a couple of signal tests, this is mostly to discuss on how we integrate with the rest of the build.

Unlike previous discussions, after experimenting a bit, I decided to include these as source and not binaries. They require `g++`/`g++-multilib` (x86-hosts) or `g++-x86-64-linux-gnu`/`g++-multilib-x86-64-linux-gnu` (non-x86 hosts) to build.

They are hidden behind a new cmake config, `BUILD_FEX_LINUX_TESTS`.

They are included via `ExternalProject_Add`, similar to `GuestThunks`. The same configuration vars are used for cross compiler selection.

Test themselves can define their compilation flags (optional, defaults to "-g O2") and a list of arguments to execute with. This is inspired from gcc-target-tests.

Format:
```
// append cxxflags: <cxxflags to append>
// append ldflags: <ldflags to append>
// libs: <libs to link with>
auto args = "call, the, test, once, with, every, comma, or space separated, value here";
```

The existing `guest_test_runner.py` is used as is with other guest tests.

The following targets are added: `fex_linux_tests` (runs emulated), `fex_linux_tests_host` (runs on host) and `fex_linux_tests_all` (runs both on x86 host, only emulated otherwise) targets.

### DIFF
--- a/unittests/FEXLinuxTests/CMakeLists.txt
+++ b/unittests/FEXLinuxTests/CMakeLists.txt
@@ -1,0 +1,104 @@
+include(ExternalProject)
+ExternalProject_Add(FEXLinuxTests
+  PREFIX FEXLinuxTests
+  SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tests"
+  BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/FEXLinuxTests"
+  CMAKE_ARGS
+    "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+    "-DX86_C_COMPILER:STRING=${X86_C_COMPILER}"
+    "-DX86_CXX_COMPILER:STRING=${X86_CXX_COMPILER}"
+  INSTALL_COMMAND ""
+  BUILD_ALWAYS ON
+)
+
+# this kind of sucks, but reglob
+file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS tests/*.cpp)
+foreach(TEST ${TESTS})
+  
+  get_filename_component(TEST_NAME ${TEST} NAME_WLE)
+
+  file(READ ${TEST} TEST_CODE)
+  
+  # Used to insert a configuration dependency to the test file
+  CONFIGURE_FILE(${TEST} ${CMAKE_BINARY_DIR}/junk.file)
+
+  set(ARGS_REGEX "auto args = \"([^\"]+)\";")
+  string(REGEX MATCH ${ARGS_REGEX} TEST_ARGS ${TEST_CODE})
+  # if cannot handle multiline variables, so we have to match the line first
+  if(${TEST_ARGS} MATCHES ${ARGS_REGEX})
+    string(REGEX REPLACE " |," ";" ARGS "${CMAKE_MATCH_1}")
+    set(VARIATIONS "")
+    foreach(ARG ${ARGS})
+      list(APPEND VARIATIONS "${TEST_NAME}-${ARG}:${ARG}")
+    endforeach()
+  else()
+    set(VARIATIONS "${TEST_NAME}:")
+  endif()
+  
+  set(ALL_BITNESS 32 64)
+  foreach(VARIATION ${VARIATIONS})
+    foreach(BITNESS ${ALL_BITNESS})
+      string(REGEX REPLACE ":" ";" VARIATION "${VARIATION}")
+      list(GET VARIATION 0 VARIATION_NAME)
+      list(GET VARIATION 1 VARIATION_ARG)
+      set(BIN_PATH "${CMAKE_CURRENT_BINARY_DIR}/FEXLinuxTests/${TEST_NAME}.${BITNESS}")
+      
+      set(TEST_CASE "${VARIATION_NAME}.${BITNESS}")
+      
+      # Add jit test case
+      add_test(NAME "${TEST_CASE}.jit.flt"
+        COMMAND "python3" "${CMAKE_SOURCE_DIR}/Scripts/guest_test_runner.py"
+        "${CMAKE_CURRENT_SOURCE_DIR}/Known_Failures"
+        "${CMAKE_CURRENT_SOURCE_DIR}/Expected_Output"
+        "${CMAKE_CURRENT_SOURCE_DIR}/Disabled_Tests"
+        "${TEST_CASE}"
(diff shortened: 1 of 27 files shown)


skmp:
@Sonicadvance1 builders are missing the `multilib` packages for this to work

neobrain:
Any thoughts of using Catch2 for the integrated sources?

For another emulation project of mine that bundles cross-compiled tests running on the emulation target, this has proven an effective way of writing more structured tests. It's easy to integrate too, since you can just `add_submodule` it to the FEXLinuxTests CMakeLists.txt.

unittests/FEXLinuxTests/CMakeLists.txt
```diff
@@ -0,0 +1,100 @@
+include(ExternalProject)
+ExternalProject_Add(FEXLinuxTests
+  PREFIX FEXLinuxTests
+  SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tests"
+  BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/FEXLinuxTests"
+  CMAKE_ARGS
+    "-DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE}"
+    "-DX86_C_COMPILER:STRING=${X86_C_COMPILER}"
```
neobrain, line 8: Instead of hardcoding the C/C++ compilers everywhere we cross-compile, how about adding CMake toolchain file to the FEX root directory instead? That allows for easier user-configuration in case they want to use a differently named cross-compiler or in case they want to use different compile flags.

skmp:
> Any thoughts of using Catch2 for the integrated sources?
> 
> For another emulation project of mine that bundles cross-compiled tests running on the emulation target, this has proven an effective way of writing more structured tests. It's easy to integrate too, since you can just `add_submodule` it to the FEXLinuxTests CMakeLists.txt.

I've never used Catch2, though I think we include it, not sure if for guest. I can take a look. Does it have any advantages?

I've also considered having a `fork()` driver for the tests, as that would cut down initialization time a lot, however that would also reduce parallelism, and would have the issue we have with gvisor with each cmake test being multiple actual tests.

neobrain:
> I've never used Catch2, though I think we include it, not sure if for guest. I can take a look. Does it have any advantages?

Same advantages as any test framework: The results/properties you're testing for are clearly highlighted between the text execution logic. In the case of errors you get well-formatted output, including the values of the actual result and the expected one.

Catch also uses a thing called Fixtures, which provide a structured way to perform test setup/teardown operations. This furthermore helps separate boilerplate from actual testing logic, and it makes writing new variants of existing tests easier.

skmp:
> > I've never used Catch2, though I think we include it, not sure if for guest. I can take a look. Does it have any advantages?
> 
> Same advantages as any test framework: The results/properties you're testing for are clearly highlighted between the text execution logic. In the case of errors you get well-formatted output, including the values of the actual result and the expected one.
> 
> Catch also uses a thing called Fixtures, which provide a structured way to perform test setup/teardown operations. This furthermore helps separate boilerplate from actual testing logic, and it makes writing new variants of existing tests easier.

I'm not sure those would be benefits for most of these tests, as setup is very likely to be specific per test, and tests for the most part will either crash, or run code that isn't supposed to run.

gvisor does this approach by (ab)using gtest, and it is quite hard to understand how things go wrong in a tests as things are hidden behind several layers of magic, and  `gdb`/assembly outputs of the test cases are over complicated by the testing framework itself.

unittests/FEXLinuxTests/CMakeLists.txt
neobrain: > Whatever change we do on this front should be synchronized with how thunks are compiled, as we don't want two different mechanisms for the same thing. I'd make a follow up investigation ticked for that though.

Yeah, agreed this should be done consistently with thunks and is fine for a followup.

> Having said that, I'm far from a CMake expert, and I'm not particularly keen on becoming one. I think we need to switch to a better built system as soon as possible.

This is worth having a deeper discussion at some point, but FTR I have strong reservations against switching away from CMake for FEX, and I don't think doing so would be a net benefit. CMake is not the ideal build system for well-established reasons, but considering some of FEX's strong requirements (using several external dependencies, mixing native builds with cross-compiled projects, ...), I consider it the least terrible option there's currently available.

A cleanup pass over FEX's current CMake files certainly has a lot of potential though.

skmp: > I consider it the least terrible option there's currently available.

That's how I see it too, right now. We are quite invested in it, swamped by other tasks, and I'm not certain other build systems will fare better :|

> A cleanup pass over FEX's current CMake files certainly has a lot of potential though.
Probably that is a better mid-term goal. Also, reviewing all the helper scripts we use for testing, etc.

skmp:
> > @neobrain for a more concrete example of how hard to understand / debug the gvisor tests are take a look at https://github.com/google/gvisor/blob/master/test/syscalls/linux/rtsignal.cc
> 
> Is this due to something specific about how these tests are written or about being unfamiliar with testing frameworks in general?
> 
> I haven't used googletest before, but for me these rtsignal tests are easier to follow than the tests added in this PR, since they follow the same general structures that most test frameworks do.
> 
> I can see the case about the debugging argument, which maybe is related to googletest's implementation. With Catch, tests haven't been any harder to debug than "non-Catch" code from personal experience. (For example: Backtraces point to the actual line being tested without any layers of magic in between, unless those layers are specifically part of the test.)

Initialization logic is non-obvious, teardown is actually not needed, a lot of test-specific wrapper functions are non obvious.

Can you tell from a quick read if the tests fork or not? ;)

skmp:
Eg, thinking of how i'd wrap the included tests, they would only increase in code size with a unit testing framework.

Sonicadvance1 (approved):

skmp:
Added SMC and did some minor changes, will document later on in the PR header.

skmp:
Fun, ubuntu's cross compilers don't seem to support -static at all

skmp:
@Sonicadvance1 Failed tests need new rootfs with GLIBC >= 2.34

skmp:
rebased on top of #1763

PR #1738, opened 2022-06-02
unittests: Workaround runner issues

Sonicadvance1:
Runners have been updated which have caused unit tests to start failing in new interesting ways.
Work around these, will create issues to do a followup on the failures.


skmp:
Spooky. Shal we make a follow up ticket so we don't forget?

You did a dist upgrade or just new packages?

skmp (approved):

PR #1745, opened 2022-06-03
mtrack: Fixes 32-bit shmat, shmdt tracking, guaranteed invalidation atomicity

skmp:
#### Overview

- Fixes 32-bit shmat, shmdt tracking when called through ipc syscall, that path missed before.

- Adds an extra lock to guarantee that invalidations happen either before or after a block has been compiled, but never during compilation. It is still possible to compile stale code if the code gets modified while being compiled, as the mprotect locks are only applied after successful compilation.

Both of these corner cases were uncovered by the FLT PR, #1737

#### Details
A shared mutex is used so that there is no contention penalty while there are no invalidations.

Locking pattern
- Around Compilation  -> lock shared
- Around Invalidation -> lock unique

### DIFF
--- a/Source/Tests/IRLoader.cpp
+++ b/Source/Tests/IRLoader.cpp
@@ -133,14 +133,15 @@ class DummySyscallHandler: public FEXCore::HLE::SyscallHandler {
     return {0, false, 0 };
   }
 
-  std::shared_mutex Mutex;
+  // These are no-ops implementations of the SyscallHandler API
+  std::shared_mutex StubMutex;
   FEXCore::HLE::AOTIRCacheEntryLookupResult LookupAOTIRCacheEntry(uint64_t GuestAddr) override {
-    return {0, 0, FHU::ScopedSignalMaskWithSharedLock {Mutex}};
+    return {0, 0, FHU::ScopedSignalMaskWithSharedLock {StubMutex}};
   }
 
-  std::shared_mutex Mutex2;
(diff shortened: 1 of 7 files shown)


Source/Tests/IRLoader.cpp
```diff
@@ -137,6 +137,11 @@ class DummySyscallHandler: public FEXCore::HLE::SyscallHandler {
   FEXCore::HLE::AOTIRCacheEntryLookupResult LookupAOTIRCacheEntry(uint64_t GuestAddr) override {
     return {0, 0, FHU::ScopedSignalMaskWithSharedLock {Mutex}};
   }
+
+  std::shared_mutex Mutex2;
```
Sonicadvance1, line 141: `Mutex2` isn't a descriptive name.

skmp: It's not used there though, it's just a stub interface for the ir loader

skmp: We probably want to get these out of the vtable tbh

Sonicadvance1: Could name it StubMutex then to imply that it is a throwaway mutex.

skmp: Can do. The way the IRLoader works is a bit wonky wrt to the rest of our architecture, though totally not willing to rework that part right now,

Sonicadvance1 (approved):

PR #1748, opened 2022-06-04
unittests: Investigate failing CI changes

### LINKED ISSUE #1741: CI: New gvisor test breaks
Due to CI upgrades.

Sonicadvance1:
One gvisor test didn't expect a file header to change layout.
Another one was testing behaviour that was removed from upstream Linux

Fixes #1741


skmp (approved):
I'll update the gvisor tests next week

Issue #1742, opened 2022-06-02
CI: Atomic interpreter test on ARMv8 break


Sonicadvance1:
Due to CI upgrades.

PR #1749, opened 2022-06-04
unittests: Reenable atomic tests on ARMv8.0

Sonicadvance1:
The latest ARMv8.0 toolchain is implementing fetch_add with a bic rather
than an and. Not sure why they started doing this but support the
remaining logical operations in our unaligned atomics handler.

Fixes the Interpreter ARMv8.0 atomic ops.

Fixes https://github.com/FEX-Emu/FEX/issues/1742


skmp (approved):
love the lambda templates

PR #1752, opened 2022-06-06
TSO: Add auto migration optimisation for applications that don't need TSO

skmp:
#### Overview
Single threaded applications that don't use shared memory don't need TSO to be enabled. This is often the case with many command line applications like `bash`, `uname`, most executions of `python/3` and many others.

For those cases, we can generate non-tso code, and switch to tso mode once they might do shared memory operations.

This introduces a new configuration, `TSOAutoMigration`, that controls this optimisation and is by default enabled.

This optimisation was inspired by qemu, that also implements the same optimisation.

#### Details
- Adds `TSOAutoMigration:bool=true` configuration
- Adds `void Context::MarkMemoryShared()` that does the migration and sets the `Context::IsMemoryShared` flag
- Adds `void MarkMemoryShared(FEXCore::Context::Context *CTX)` wrapper
- Adds `bool Context::IsTSOEnabled()` that returns true if TSO ops need to be emitted, taking into account `Context::IsMemoryShared` as well as `Config.TSOAutoMigration` and `Config.TSOEnabled`
- Modifies `OpDispatchBuilder::_StoreMemAutoTSO` and `OpDispatchBuilder::_LoadMemAutoTSO` to use `Context::IsTSOEnabled()`
- Modifies `CloneHandler` to call `MarkMemoryShared` if `CLONE_VM` is set
- Modifies `TrackMmap` to call `MarkMemoryShared` if `MAP_SHARED` is set
- Modifies `TrackShmat` to always call `MarkMemoryShared`

#### Complications
This makes IR and OBJ caching somewhat more involved, as we might have to migrate caching files during execution. This is not handled in this PR, and `TSOAutoMigration` should be turned off for those use cases.

This might also have issues with some corner cases around migrations.

If shared mode is entered while a signal return is pending, the signal return depends on the existing code cache not going away. In order to mitigate this, only the lookup and IR caches are invalidated, but not the code caches. This means non-tso code might run but a short while after signal return, even if shared mode has been entered.

Also, another edge case is when rest of the fragment right after `clone`, `mmap` `ipc` or `shmat` does some memory operation that needs to be TSO. This could be avoided if those were made (optional?) redispatches, at a small performance cost, and might be something we want to do in the future.

### DIFF
--- a/External/FEXCore/Source/Interface/Config/Config.json.in
+++ b/External/FEXCore/Source/Interface/Config/Config.json.in
@@ -260,6 +260,14 @@
           "Highly likely to break any multithreaded application if disabled."
         ]
       },
+      "TSOAutoMigration": {
+        "Type": "bool",
+        "Default": "true",
+        "Desc": [
+          "Automatically enables TSO when shared memory is used.",
+          "Should work without issues in most cases."
+        ]
+      },
       "X87ReducedPrecision": {
         "Type": "bool",
         "Default": "false",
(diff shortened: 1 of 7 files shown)


Sonicadvance1 (approved):
Very nice. Confirmed the performance boost for single threaded applications.
Also confirmed an override of the option reduces performance as one would expect.

PR #1756, opened 2022-06-07
IPR: Store copy of IRLists, Dispatcher cleanups

skmp:
#### Overview
This escalated from other cleanups I've been doing as part of the #1682 pass. Cleanups & Decouples interpreter from Context structures, separates IR precompilation from debug data, reduces lock contention, merges some code that was identical but copy-pasted, harmonizes X86 and ARM64 backends a bit.

#### Details
- Moves arm64 relocation from `Arm64Emitter` to `Arm64JIT`, which mirrors how x86 works
- Moves/Merges code buffer mgmt logic to CPUBackend
- Merges common Pointers
- Changes Interpreter dispatch to run executor fn
- Changes Interpreter to serialize IRLists to the "code buffer"
- Splits `LocalIRCache` to `PrecompiledIR` (only for IRLoader) and `DebugStore` (only if GDB attached)
- Only populates `DebugStore` if gdb is attached

#### Depends
Needs #1766 merged first


External/FEXCore/include/FEXCore/Debug/InternalThreadState.h
```diff
@@ -72,7 +73,7 @@ namespace FEXCore::Core {
   };
 
   struct InternalThreadState {
-    FEXCore::Core::CpuStateFrame* CurrentFrame = &BaseFrameState;
+    FEXCore::Core::CpuStateFrame* const CurrentFrame = &BaseFrameState;
```
skmp, line 76: Switching frames is broken, and not used. Made this const here so we don't accidentally do it. We might have to come back to this and re-merge Thread/Frame.

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterCore.cpp
```diff
@@ -26,19 +34,30 @@ namespace FEXCore::IR {
 namespace FEXCore::CPU {
 class CPUBackend;
 
-static void InterpreterExecution(FEXCore::Core::CpuStateFrame *Frame) {
-  auto Thread = Frame->Thread;
+InterpreterCore::InterpreterCore(FEXCore::Context::Context *ctx, FEXCore::Core::InternalThreadState *Thread)
+  : CPUBackend(Thread, 1024 * 1024 * 16, 1024 * 1024 * 128)
```
Sonicadvance1, line 38: The Min and Max sizes here should be using constant defines to make it clearer, like it was before.
Same for the x86 and AArch64 sides.

skmp: I'll move them out

skmp: Huh, I actually left the consts there, but didn't use them. Good catch!

Moved them to be local to JIT.cpps, changed code to use them.

skmp:
@Sonicadvance1 addressed review, is this good enough to be merged? It's blocking me in follow up work.

Sonicadvance1 (approved):

skmp:
thanks!

PR #1759, opened 2022-06-08
IR.json: Correct 'Dest' key to 'Desc'

lioncash:
In a few places the description key was accidentally written as Dest.


Sonicadvance1 (approved):
woops!

Issue #1203, opened 2021-08-13
Support thunking of vtables and function pointer members


neobrain:
Many APIs return objects with function pointer data members (e.g. vtables). These are difficult to handle, since both the host libraries (invoked through thunklibs) and the guest program expect to be able to call them. FEX's existing callback mechanism isn't suitable for this.

I've researched other strategies to handle this scenario, the most universal one being Option 3 below. The other options are included for reference, since their drawbacks highlight why option 3 is needed for the general case.

*For illustration, consider this example guest API: `Object* CreateObject()` and `struct Object { void (*funcptr)(Object*); }`.*

## Approach 1: Guest-side mirror copy

In addition to the hostlib-created object, a mirror copy is created where all function pointers are replaced with guest-function pointers. The two are stored in a wrapper struct `struct Wrapped {  Object guest_obj; Object* host_obj; }`, which can be cast to/from the `Object` pointer as seen by the guest.

This requires synchronizing the two copies at each guest<->host transition point. For guest->host, data (excluding function pointers) from `guest_obj` must be copied to `*host_obj`, and vice-versa for host->guest.

👎 Overhead due to object copying
👎 Only works for function pointers that take a `this` argument (otherwise, there's no way to look up `host_obj`)
👎 Requires lots of boilerplate
👎 Does not work for APIs where the guest is in charge of destroying the object using external functions (e.g. free() instead of XDestroyImage). More generally it doesn't work for libraries that have any dependency on the object address whatsoever.
👎 Unreliable if the host may modify Object data outside of API calls (e.g. for async code)
👎 Further workarounds required if the guest overwrites function pointers

## Approach 2: `Object*->host_vtable` dictionary

CreateObject in Host-thunklib replaces the vtable of the created object with guest function pointers but stores the old one in an `std::unordered_map` for future lookup.

This approach requires the vtable in the object to be swapped for each host<->guest transition, but compared to approach 1 this is cheaper and requires less boilerplate.

👎 Only works for function pointers that take a `this` argument (otherwise, there's no way to look up `host_vtable`)
👎 Space overhead due to the (possibly large) dictionary
👎 Small performance overhead due to dictionary lookup
👎 Requires some boilerplate
👎 Further workarounds required if the guest overwrites function pointers

## Approach 3: Making selected host addresses callable from guest

[...]

PR #1760, opened 2022-06-08
Thunks: Support returning host function pointers to the guest

neobrain:
## Overview

Currently, APIs that return function pointers (notably `glXGetProcAddress` and `vkGetDeviceProcAddr`) cannot be thunked easily as the returned pointers are not callable on the guest. Both libGL and libvulkan have limited workarounds in place that are limited to Mesa's Vulkan drivers and don't support multiple devices. These workarounds aren't transferable to other libraries, which may not have a 1:1 mapping between function pointers and symbol names, or which may use vtables. Such libraries either require new library-specific workarounds or a generic solution to the problem (the latter which is provided here).

This PR adds new infrastructure (some of which eventually went into #1770) for redirecting host-function-pointer-calls to a guest function, which allows for reliably thunking APIs that return function pointers. The new features are employed in libGL and libvulkan to allow for vendor-neutral GPU acceleration of games running in FEX, albeit issues in other libraries may block this from fully working.

This addresses one half of #1203, with the other half being support for vtables.

## Impact on thunk interface definitions

A good guiding example of the new generator features is `libvulkan`.

Libraries with functions that return function pointers must enclose potential callees in a namespace annotated with `fexgen::indirect_guest_calls`. Upon returning a host function pointer from a function, the guest-side thunk end point of that function must use `LinkAddressToFunction` to handle guest-calls to that pointer using `CallHostThunkFromRuntimePointer`.

Automation of this process will be provided in a followup PR, allowing for APIs that return function pointers *with statically known function signatures* to be thunked without manual annotations.

## Side effects of thunking libvulkan directly

`vulkaninfo` hits a code path calling `VkDebugReportCallbackCreateInfoEXT` when thunking the host Vulkan loader directly. This is currently broken as FEX has no generic support for thunking APIs that pass guest function pointers to the host (as opposed to passing host pointers to the guest). To make sure there are no functional regressions, this PR disables `VK_EXT_debug_report` and `VK_EXT_debug_utils` hence.

### Critical TODOs
* [x] Make generation of hostcall boilerplate optional
  * Done: must opt-in using the `indirect_guest_calls` namespace annotation
* [x] Consider interaction with custom guest/host entrypoints
  * Custom host entrypoints (e.g. vkAllocateMemory): No problem, since function pointers are selected by manually implemented code. (E.g. vkGetDeviceProcAddr can just check individual function names and override the native host library function pointer with its own)
  * Custom guest entrypoints (e.g. vkGetDeviceProcAddr): Needs further (simple) changes, see #1774. Current implementation does not support this either, so there is no functional regression.
* [x] Fix thunks with stubbed callbacks (e.g. vkAllocateMemory, glDebugMessageCallback)
* [x] Investigate startup delay for glxinfo in libGL's `InitFunctionPointers`
  * This was triggered by directly passing GL's 3000 API functions to the `std::unordered_map` constructor. In Release builds, GCC's optimizer turned this into code that would aggressively trigger spilling in the FEX JIT. As a workaround, the map is initialized incrementally instead.
* [x] Avoid global state in JIT.cpp (worked on by skmp in https://github.com/FEX-Emu/FEX/compare/feat/gch-trampolines)
* [ ] Disable `VK_EXT_debug_report` and `VK_EXT_debug_utils`

### Soft TODOs (may be dropped or moved to a follow-up PR)
* [x] Add support for interpreter (currently JIT-only) (see https://github.com/FEX-Emu/FEX/compare/feat/gch-trampolines)
* [x] Consider unregistering host-pointers on dlclose (see https://github.com/FEX-Emu/FEX/compare/feat/gch-trampolines)
* [x] Add x86 host support (currently ARM-only)
  * `vulkaninfo` and `vkcube` work, `glxinfo` and `glxgears` don't
* [x] Add documentation
* [ ] Deprecate custom_guest_entrypoint and generate_guest_symtable (will be done in a followup)
* [x] Investigate libvulkan issues on nvidia (resolved)


skmp:
Nice to see this in a PR!

I find it cute how the host pointers can be called directly! This is likely the only solution for thunking "vtables" that was discussed in the past.


I also like how "function pointer thunks" are handled like normal thunks, except with a function pointer parameter.

A few architectural comments on the implementation as is
- It is probably best to move the work to the IR side (intercept in Core, generate IR via a helper in the thunks, bypass the frontend, and let the arch specific backend do the codegen.
- It is probably best to use another scratch reg, like r10 or r11, as rax is used sometimes in the abi to pass the number of floating params in registers in variadic functions.

Thinking a bit further, I'm not sure doing the packer/unpacker via templates is the way to go. We already generate packers/unpackers, and it might be easier to reason, and customise, if gen just generates more of them. We can investigate in the future to switch the entire packer/unpacker generation to templates.

Some thoughts a variation on this
- We could introduce another layer, "fn pointer wrappers", that their job is to append the fn pointer, before the marshal code, and before the packer code. These would be functions that just call an internal function with an appended arg. This is similar to the PR, except it's not done in the packers directly.
- We can then generate as many of these functions as we want in the runtime, from a canonical C implementation. This would remove the need to pass the host pointer as a secret abi argument.
- We'd have to keep host:guest and guest:host maps in the FEXCore side of thunks, and we'd have to look then up every time an fn pointer needs to be marshalled in either direction

That would have the benefit of being less invasive on the JIT side, at the cost of not having callable host pointers directly.

 
Another random thought
- We can pass the hostfn ptr as a parameter in the unpacker, rather than packing it. This would work by adding another parameter to OP_THUNK in the IR, and extending the unpackers to take two args.

skmp:
I've re-implemented the JIT side of this via a generic IR implemnentation in https://github.com/FEX-Emu/FEX/commit/9ad701a3c14d79d9522aec8a635827099626dd1a (https://github.com/FEX-Emu/FEX/tree/feat/gch-trampolines). Still untested due to compile issues with thunks locally.

skmp:
Also, there is a design corner case issue with guest callable host code, as the same host pointer can only be mapped to a single guest thunk.

This can happen if the same host pointer is mapped to different thunks, either from the same library (easy to avoid) or from different libraries (not so easy to avoid).

9ad701a3c14d79d9522aec8a635827099626dd1a asserts if this case is hit.

skmp:
I've updated and rebased `feat/gch-trampolines` on top of https://github.com/FEX-Emu/FEX/pull/1760 that provides a generic infrastructure in the core to handle this

skmp:
> > Why not avoid using a map there?
> > We can pre-sort the names since the list is static and then just do a binary search there. Will be much faster init than map as it will run precisely zero code on load. It's also not like the lookup needs to be particularly fast - even a flat list will just work fine.
> 
> Other than the issue I originally pointed out, I'd expect this detail to have little overall impact. The current approach is the least deviation from the current code, and I'd rather move non-critical changes here to a followup-PR.

Sure, that's fine, it is def. not critical. Static constructors have in the past creeped up on us wrt startup beformance, that's the motivation.

External/FEXCore/Source/Interface/Context/Context.cpp
```diff
@@ -193,6 +193,10 @@ namespace FEXCore::Context {
     return CTX->UnloadAOTIRCacheEntry(Entry);
   }
 
+  uintptr_t AddCustomIREntrypoint(FEXCore::Context::Context *CTX, uintptr_t Entrypoint, std::function<void(uintptr_t Entrypoint, FEXCore::IR::IREmitter *)> Handler, const void *Creator, uintptr_t Data) {
```
neobrain, line 196: Should return `std::optional<uintptr_t>` so LoadIR can tell if custom IR had been registered here before or not.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -56,6 +62,54 @@ namespace FEXCore {
...
+         * Instructs the Core to redirect calls to functions at the given
+         * address to another function. The original callee address is passed
+         * to the target function through an implicit argument stored in r11.
+         *
+         * The primary use case of this is ensuring that host function pointers
+         * returned from thunked APIs can safely be called by the guest.
+         */
+        static void LinkAddressToGuestFunction(void* argsv) {
+            struct args_t {
+                uintptr_t original_callee;
```
skmp, line 75: I would name this something among the lines `guest_linked_address` or `guest_entrypoint_address` and `guest_function` or `guest_handler_function`. 

Keeping a tab on what is guest and what is host can get very confusing, so I like to be verbose there

ThunkLibs/Generator/gen.cpp
```diff
@@ -574,8 +597,18 @@ void GenerateThunkLibsAction::EndSourceFileAction() {
...
             }
 
+            FunctionParams args = thunk;
+            auto function_to_call = "fexldr_ptr_" + libname + "_" + function_name;
+            if (thunk.is_hostcall) {
+                // Get the host function pointer by casting the last parameter to the correct signature
+                args.param_types.pop_back();
+                function_to_call = "reinterpret_cast<" + thunk.return_type.getAsString() + "(*)(" + format_function_params(args) + ")>(args->a_" + std::to_string(args.param_types.size()) + ")";
+            } else if (thunk.custom_host_impl) {
+                function_to_call = "fexfn_impl_" + libname + "_" + function_name;
```
skmp, line 607: Shouldn't this be
`(thunk.custom_host_impl ? "fexfn_impl_" : "fexldr_ptr_") << libname << "_" << function_name`

To handle impls as well?

neobrain: This is what the original code was doing, but the `custom_host_impl` case is now handled explicitly (whereas the `!custom_host_impl` case is handled at initialization). So I don't think there should be an additional conditional.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -56,6 +62,54 @@ namespace FEXCore {
...
+                uintptr_t target_addr;     // Guest function to call when branching to replaced_addr
```
neobrain, line 76:
```suggestion
                uintptr_t target_addr;     // Guest function to call when branching to original_callee
```

ThunkLibs/libGL/libGL_Guest.cpp
```diff
@@ -8,41 +8,59 @@ desc: Handles glXGetProcAddress
...
 #include "common/Guest.h"
 
 #include "thunks.inl"
 #include "function_packs.inl"
 #include "function_packs_public.inl"
+#include "symbol_list.inl"
 
 typedef void voidFunc();
 
+// Maps Vulkan API function names to the address of a guest function which is
```
skmp, line 33: Vulkan?

ThunkLibs/libGL/libGL_Guest.cpp
```diff
@@ -8,41 +8,59 @@ desc: Handles glXGetProcAddress
...
+    }
 
-        for (int i = 0; internal_symtable[i].name; i++) {
-            if (strcmp(internal_symtable[i].name, (const char*)procname) == 0) {
-				// for debugging
-                //printf("glXGetProcAddress: looked up %s %s %p %p\n", procname, internal_symtable[i].name, internal_symtable[i].fn, &glXGetProcAddress);
-                return internal_symtable[i].fn;
-			}
-		}
+    auto TargetFuncIt = HostPtrInvokers.find(reinterpret_cast<const char*>(procname));
```
skmp, line 44: I'd rather us use a flat string list + `std::lower_bound` for binary search here, though that's something we can discuss for a follow up.

ThunkLibs/libvulkan/Host.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include <dlfcn.h>
+
+#include "ldr_ptrs.inl"
+
+static bool SetupInstance{};
+static std::mutex SetupMutex{};
+
+#define LDR_PTR(fn) fexldr_ptr_libvulkan_##fn
+
+static void DoSetupWithInstance(VkInstance instance) {
```
skmp, line 28: Is this still needed with GCH?

From what I understand we can remove special handling for all of these

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
skmp: I'd rather us rework the names in a way that the comments aren't needed

Sonicadvance1 (approved):
Once this is merged I'll be throwing a bunch of testing at it.

ThunkLibs/libvulkan/Host.cpp
neobrain: This relies on https://github.com/FEX-Emu/FEX/issues/1774.

Until the GCH framework supports functions with custom guest entrypoints, vkGetDeviceProcAddr needs `fexldr_ptr_vkGetDeviceProcAddr` to be set up before being called (which in turn needs a VkInstance to be present).

ThunkLibs/libGL/libGL_Guest.cpp
neobrain: Yeah, agreed that's an orthogonal refactor better moved to another PR.

ThunkLibs/libGL/libGL_Guest.cpp
neobrain: Woops, resolved.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
neobrain: There is no requirement for the callee address to be a host function. While it's used like that in practice, I reckon thinking about it in terms of a callee being replaced by a target address is most useful here.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
neobrain: I'm happy with the current balance between self-documenting code and descriptive docstrings. We're implementing highly unorthodox logic here, and a comment is the best tool to document the main idea here.

neobrain:
Fixed the crash on x86 caused by trying to setting up the `HostPtrInvoker` map before its constructor was called. This was resolved by setting up the map entries in its initializer instead of using a separate function for that. Unfortunately, due to #1791 this needs a dance with `std::invoke` and an initializing lambda since direct initialization causes noticeable startup delays.

Also added a patch to disable debug extensions, since these cause `vulkaninfo` to call guest-callbacks from the host (which is unsupported still). Waiting for testing feedback to see if that's enough to fix `vulkaninfo`, but should be good for merge other than that.

skmp:
> > I'd rather us use more uniform naming (hostcall vs indirect_guest_calls)
> 
> Yeah, this inconsistency is not ideal. I'd suggest postponing better naming until we tackle generic guest callbacks, since I'm having a hard time thinking of a name that won't lead to ambiguity later at the moment.
> 

Agreed there. Let's get this merged already :>

> > I'd rather not use templates, but generate the packers for hostcall thunks like we normally do
> 
> The reason for using templates here is that code generation from C++ is extremely awkward due to the involved boilerplate, which makes it hard to reason about and difficult to modify reliably. Templates use better-integrated syntax and are more easily adopted (and also don't require us to employ the even more arcane preprocessor magic for every feature added).

There are some good workarounds there, we can namespace the argsrv by fn_type so that
```details<fn_type>::argsrv```
gives us the right struct

Sonicadvance1 (approved):
Confirmed that this worked around the crashes I had before.

PR #1764, opened 2022-06-10
FEXRootFSFetcher: Update and fix xxhash file hashing

Sonicadvance1:
The final tail of the file reading was incorrect, so our hashing was
"correct" but it was using stale data from the previous block size read.

Noticed this while wiring up the CI rootfs fetching since the hashing is
a lot simpler there.

Now instead of reading a tail, just attempt to read the full block size
and use the resulting data size instead. Confirmed it matches expected
results now.

In the process we are going to need to update hyperlinks and hashes
anyway, change the hash to XXH3 so it is faster to run.

### DIFF
--- a/Scripts/InstallFEX.py
+++ b/Scripts/InstallFEX.py
@@ -306,12 +306,12 @@ def GetRootFSPath():
     return _RootFSPath
 
 def CheckRootFSInstallStatus():
-    # Matches what is available on https://rootfs.fex-emu.org/file/fex-rootfs/RootFS_links.txt
+    # Matches what is available on https://rootfs.fex-emu.org/file/fex-rootfs/RootFS_links_XXH3.txt
     UbuntuVersionToRootFS = {
         "20.04": "Ubuntu_21_04.sqsh",
         "21.04": "Ubuntu_21_04.sqsh",
         "21.10": "Ubuntu_21_10.sqsh",
-        "22.04": "Ubuntu_21_10.sqsh",
+        "22.04": "Ubuntu_22_04.sqsh",
     }
 
     return os.path.exists(GetRootFSPath() + UbuntuVersionToRootFS[GetDistro()[1]])
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -346,7 +346,7 @@ namespace WebFileFetcher {
     std::string Hash;
   };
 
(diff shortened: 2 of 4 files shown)


skmp:
Why don't we use `sha256sum` or something else for these digests?

skmp (approved):

Sonicadvance1:
> Why don't we use `sha256sum` or something else for these digests?
> 
> _edit_ It would also be nice to have the metadata information in a json, instead of a custom format

Multiple reasons really. Wanted to keep the number of external dependencies reduced, wanted to not need to parse stdout, not 100% sure if all our execution environments would have a least common denominator file hashing program, and speed.
```
ryanh@ubuntu-linux-20-04-desktop:~/.fex-emu/RootFS$ time FEXRootFSFetcher Ubuntu_22_04.sqsh
Ubuntu_22_04.sqsh has hash: 101f680c46d2ff70

real    0m0.295s
user    0m0.032s
sys     0m0.263s
ryanh@ubuntu-linux-20-04-desktop:~/.fex-emu/RootFS$ time sha1sum Ubuntu_22_04.sqsh
60f1a4f230f2147fb19f1555d41fecb7c44e7991  Ubuntu_22_04.sqsh

real    0m1.366s
user    0m1.333s
sys     0m0.033s
ryanh@ubuntu-linux-20-04-desktop:~/.fex-emu/RootFS$ time sha256sum Ubuntu_22_04.sqsh
b5326fc26ab7e4534f0981801ce439de36b4ddf6af93e346a6d708ad527f7f26  Ubuntu_22_04.sqsh

real    0m3.121s
user    0m3.083s
sys     0m0.038s
ryanh@ubuntu-linux-20-04-desktop:~/.fex-emu/RootFS$ time md5sum Ubuntu_22_04.sqsh
351392bf02037467af6bdaabf392d727  Ubuntu_22_04.sqsh

real    0m2.113s
user    0m2.056s
sys     0m0.057s
```

skmp:
> Wanted to keep the number of external dependencies reduced, wanted to not need to parse stdout, not 100% sure if all our execution environments would have a least common denominator file hashing program, and speed.

I think sha256sum ships with every distro core at this point, and not sure why speed matters - isn't that just to verify the download?

Anyway, it doesn't really affects anything, just xxh3 is not cryptographically secure so it's not a good "digest" from that sense.

Sonicadvance1:
Right, it doesn't really affect anything. I seem to recall there is some posix compliant environment that doesn't have the sum program. Don't remember now since I haven't thought about it, doesn't really matter.

PR #1767, opened 2022-06-11
Thunks: Support pressure-vessel prefixes

Sonicadvance1:
Instead of duplicating prefixes in the ThunksDB json file even more,
just do a prefix replacement when the thunks database is being parsed.

Changes the JSON overlay arrays over to @PREFIX@ and reduce the
duplication.

Adds support for the pressure-vessel prefix `/usr/lib/pressure-vessel/overrides/lib`

This gets thunks ready for running under pressure-vessel, once vulkan
thunking switches from device libraries to the vulkan loader it should
just work.


Sonicadvance1:
Needs #1766 before CI is resolved.

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -138,13 +138,45 @@ void FileManager::LoadThunkDatabase(bool Global) {
...
+            };
+
+            bool Replace = true;
+            size_t i = 0;
+            do {
+              std::string NonModifiedLibraryItem {static_cast<const char*>(json_getValue(Value))};
+              auto Replacement = ReplacePrefix(NonModifiedLibraryItem, "@PREFIX@", Prefixes[i]);
+              DBObject->second.Overlays.emplace_back(Replacement.second);
+              Replace = Replacement.first;
+            } while (Replace && ++i != Prefixes.size());
```
neobrain, line 170:
```suggestion
            std::string NonModifiedLibraryItem {static_cast<const char*>(json_getValue(Value))};
            for (auto& prefix : Prefixes[i]) {
              auto [has_prefix, modified_library_item] = ReplacePrefix(NonModifiedLibraryItem, "@PREFIX@", Prefixes[i]);
              DBObject->second.Overlays.emplace_back(std::move(modified_library_item));
              if (!has_prefix) {
                  break;
              }
            }
```

Instead of the `while (Replace)`/`if (!has_prefix)` checks, wouldn't a hard error be more appropriate? Seems like *not* starting any of the paths in the thunks ini with `@PREFIX@` would likely yield results that are unexpected otherwise.

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -138,13 +138,45 @@ void FileManager::LoadThunkDatabase(bool Global) {
...
+              "/usr/lib/pressure-vessel/overrides/lib",
+            };
+
+            auto ReplacePrefix = [](const std::string &String, const char *Prefix, const char *NewPrefix) -> std::pair<bool, std::string> {
+              auto it = String.find(Prefix);
+              if (it != String.size()) {
+                size_t SizeOfOldPrefix = strlen(Prefix);
+                std::string Replacement = String;
+                Replacement.replace(it, SizeOfOldPrefix, NewPrefix);
+                return std::make_pair(true, Replacement);
```
neobrain, line 156:
```suggestion
                return std::make_pair(true, std::move(Replacement));
```

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -138,13 +138,45 @@ void FileManager::LoadThunkDatabase(bool Global) {
...
+
+          auto AddWithReplacement = [DBObject](json_t const* Value) {
+            constexpr static std::array<const char*, 4> Prefixes = {
+              "/usr/lib",
+              "/usr/local/lib",
+              "/lib",
+              "/usr/lib/pressure-vessel/overrides/lib",
+            };
+
+            auto ReplacePrefix = [](const std::string &String, const char *Prefix, const char *NewPrefix) -> std::pair<bool, std::string> {
```
neobrain, line 150: Can avoid a heap allocation by typing `NonModifiedLibraryItem` as `const char*` and changing this parameter to `string_view` instead:
```suggestion
            auto ReplacePrefix = [](std::string_view &String, const char *Prefix, const char *NewPrefix) -> std::pair<bool, std::string> {
```

neobrain (approved):
Added a few minor comments to drop unneeded string allocations, looks good otherwise.

Source/Tests/LinuxSyscalls/FileManagement.cpp

Source/Tests/LinuxSyscalls/FileManagement.cpp

Source/Tests/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Not supporting `@PREFIX@` on the path is important for quick hacking of absolute paths for testing. Might also be necessary for replacing paths that don't match the typical `/usr` prefix.

skmp:
A minor note here, would be nice to have a more descriptive name for the prefix, eg `@PREFIX_LIB@`, so we can support multiple of those.

Source/Tests/LinuxSyscalls/FileManagement.cpp
neobrain: Sounds good. I slightly misread the code before, so this shouldn't cause any unexpected results after all.

Sonicadvance1:
> A minor note here, would be nice to have a more descriptive name for the prefix, eg `@PREFIX_LIB@`, so we can support multiple of those.

Sure thing, I don't care what the prefix name is here. Resolved.

PR #1775, opened 2022-06-14
Make Dispatcher per Context from per Thread, Simplify TestHarnessRunner

skmp:
#### Overview

Cleanup/refactor. There were very few things left per Thread, so this was fairly painless.

TestHarnessRunner used the `CPUFactory` interface and a custom core to run the tests on the host, however the context and all related classes aren't needed or used for this. It has been reworked to directly run the host tests without a FEX Context.

#### Motivation
Apart from a nice cleanup, this is yet another incremental step towards context-sharable codegen, which can work nicely in combination with object code caching.

I have a PoC of context-sharable codegen in `skmp/context-wide-buffering` though it is buggy and leaky.

#### Depends on
#1770

#### Splits
#1778


skmp:
(Rebased for #1775 split)

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -703,151 +712,160 @@ namespace FEXCore::Context {
...
-    Thread->PassManager->Run(Thread->OpDispatcher.get());
+    Thread->PassManager->Run(IREmitter);
 
     // Debug
     {
       if (Thread->CTX->Config.DumpIR() != "no") {
-        IRDumper(Thread, GuestRIP, Thread->PassManager->HasPass("RA") ? Thread->PassManager->GetPass<IR::RegisterAllocationPass>("RA")->GetAllocationData() : nullptr);
-      }
-
-      if (Thread->OpDispatcher->ShouldDump) {
```
Sonicadvance1, line 838: What replaces this `ShouldDump` usage?
This is highly useful for new instruction implementations to have instructions that are new emit their blocks of IR implementation.

Sonicadvance1 (approved):
Apart from the single comment this looks fine.

External/FEXCore/Source/Interface/Core/Core.cpp
skmp: And I /thought/ it might be so. Will address.

PR #1776, opened 2022-06-14
Arm64/EncryptionOps: Fix register specifiers in PCLMUL movs

lioncash:
Prevents an assertion from firing in vixl when they're enabled (destination needs to be a scalar)


skmp (approved):

skmp:
How did this get through our CI?

Sonicadvance1:
Looks like it got through CI because vixl assertions are only enabled in a debug build. Where we enable our assertions independently with `ENABLE_ASSERTIONS`. We could switch vixl over to this as well since it is our own cmake files there.
We don't build debug in CI because it is quite a bit slower on the CI runners.

skmp:
> We could switch vixl over to this as well since it is our own cmake files there
:+1:

PR #1777, opened 2022-06-14
CMAKE: Create directories during configuration, fixes endless generation of unittests

skmp:
The endless generation is caused because the directory timestamps are newer than some of the generated files, and the directories are a build time dependency, creating a circular dependency that always forces regeneration.
 
I'm not a cmake expert, so not sure if there are any drawbacks with this solution.


Sonicadvance1 (approved):
Weird that the timestamp issue occurs at all but sure.
This should also fix the `ninja clean` error when folders complain about not being empty.
A little jank but should be fine.

skmp:
> Weird that the timestamp issue occurs at all but sure.

- Each folder has 2 generated files
- Each generated file has the folder as the dependency
- Folder's last changed time is the last time a file was changed in it
- Thus, always the file generated first will be older than the folder, which forces a regen as the folder is a dependency

Sonicadvance1:
Cool, didn't know that changed the folder's change time. Good to know.

PR #1778, opened 2022-06-15
Context: Fix CreateThread partial initialization issue

skmp:
#### Overview
`Context::CreateThread` would add a thread to the `Threads` vector before it has been fully constructed / initialized.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -352,9 +352,6 @@ namespace FEXCore::Context {
       // Walk the threads and tell them to clear their caches
       // Useful when our block size is set to a large number and we need to step a single instruction
       for (auto &Thread : Threads) {
-        // Wait for thread to be fully constructed
-        // XXX: Look into thread partial construction issues
-        while(Thread->RunningEvents.WaitingToStart.load()) ;
         ClearCodeCache(Thread);
       }
     }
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -553,14 +550,7 @@ namespace FEXCore::Context {
   }
 
   FEXCore::Core::InternalThreadState* Context::CreateThread(FEXCore::Core::CPUState *NewThreadState, uint64_t ParentTID) {
-    FEXCore::Core::InternalThreadState *Thread{};
-
-    // Grab the new thread object
-    {
-      std::lock_guard<std::mutex> lk(ThreadCreationMutex);
-      Thread = Threads.emplace_back(new FEXCore::Core::InternalThreadState{});
-      Thread->ThreadManager.TID = ++ThreadID;
```
neobrain, line 562: Why is there no equivalent to this line in the new code?

skmp: This TID is incorrect, and seems like some leftover from when FEX implementing its own TID and PIDs. `ThreadID` isn't even an atomic.

The correct value is set by `Context::InitializeThreadTLSData` called from `Context::ExecutionThread`, which is run with the host backing thread created from `Context::InitializeThread`.

skmp: (The actual thread creation is done using a custom threading implementaiton in `FEXCore/Source/Utils/Thread.cpp `class PThread`)

Sonicadvance1 (approved):
Looks good, cleaning up old cruft.

skmp:
Obliterated `ThreadID` from the context since not used, updated docstrings, will merge if passes.

PR #1779, opened 2022-06-16
Arm64/JIT: Use IR names in opcode implementations

lioncash:
Uses the IR identifiers within the implementations instead of indexing into the header args array.

Same behavior, but provides more readable names.


Sonicadvance1 (approved):
Wow, good job. That looks like a lot of renaming hell.

skmp:
❤️  we are forever in your debt 🙇

PR #1787, opened 2022-06-18
gdb: jit integration

skmp:
#### Intro
The question no one asked: If gdb stub is too much in fex, what about fex stub in gdb?

This is a very-quick-hack-and-slash prototype to get some feedback. With a small set of features, gdb could become /much/ more welcoming to us. I'll ponder on what that means and think about talking with upstream.

Even as is, the results are very encouraging. Looks like long-term, we'd have to generate faux elfs or work with upstream for an improved interface.

##### x86-64 host
Split view debugging x86-64

##### arm64 host
Split view debugging arm64

Uses the gdb jit interface to register guest asm and guest source. Uses `x86-....-objdump -SC` to generate mixed asm / source 'source' files. Places them in `~/.fex-emu/fexsrc/fileid.fexsrc`.

Can be slow during objdump, parsing or with many symbols.

See https://github.com/FEX-Emu/fex-assorted-tests-bins/tree/main/src/gdb-jit#gdb-jit-interface-test on how to use the GDB side of things.

in gdb,
```
layout split
```

Is pretty. `ni`, `si`, `n`,`s` are your friends.

#### Limitations
- Dispatcher debug info is not registered
- Backtraces are broken. GDB depends on us for them, including for the host.
- Guest registers aren't registered as symbols
- Guest debug symbols aren't registered as symbols either

#### Depends 
- #1799

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -7,6 +7,7 @@ desc: Glues Frontend, OpDispatcher and IR Opts & Compilation, LookupCache, Dispa
 $end_info$
 */
 
+#include <cstdint>
 #include "Interface/Context/Context.h"
 #include "Interface/Core/LookupCache.h"
 #include "Interface/Core/Core.h"
@@ -33,6 +34,7 @@ $end_info$
 #include <FEXCore/Debug/InternalThreadState.h>
 #include <FEXCore/Debug/X86Tables.h>
 #include <FEXCore/HLE/SyscallHandler.h>
+#include <FEXCore/HLE/SourcecodeResolver.h>
 #include <FEXCore/HLE/Linux/ThreadManagement.h>
 #include <FEXCore/IR/IR.h>
 #include <FEXCore/IR/IREmitter.h>
@@ -50,7 +52,6 @@ $end_info$
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
-#include <cstdint>
 #include <filesystem>
 #include <functional>
 #include <fstream>
@@ -74,6 +75,7 @@ $end_info$
 #include <utility>
 #include <vector>
 #include <xxhash.h>
+
 
 namespace FEXCore::CPU {
   bool CreateCPUCore(FEXCore::Context::Context *CTX) {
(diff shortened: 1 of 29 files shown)


skmp:
This has been 
- Somewhat cleaned up
- Rebased on top of made
- Hidden behind CMAKE:ENABLE_GDB_SYMBOLS with gdb auto detect
- Updated to support aarch64
- Updated to cache index files to avoid parsing objdump output too often

skmp:
Split 6431da1387f27c3af5683f23a0508b3646ea8f96 as #1799

skmp:
@Sonicadvance1 I would really love to get some feedback from you on this.

It works on arm64 and x86-64 hosts with ubuntu 22.04, and has been quite useful for me already. I want to do one more cleanup pass - there's some comments left somewhere, and maybe split things a bit into functions - and if there are no objections to this approach to merge it?

Sonicadvance1 (approved):
Very interesting. Seems fine to me.

External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -770,6 +772,8 @@ namespace FEXCore::Context {
           DecodedInfo = &Block.DecodedInstructions[i];
           bool IsLocked = DecodedInfo->Flags & FEXCore::X86Tables::DecodeFlags::FLAG_LOCK;
 
+          Thread->OpDispatcher->_GuestOpcode(Block.Entry + BlockInstructionsLength - GuestRIP);
```
skmp, line 775: Should be made optional

External/FEXCore/Source/Interface/IR/AOTIR.cpp
```diff
@@ -281,6 +339,76 @@ namespace FEXCore::IR {
     return Result;
   }
 
+#if defined(GDB_SYMBOLS_ENABLED)
```
skmp, line 342: Should be split to its own file

Source/Tools/FEXGDBReader/FEXGDBReader.cpp
```diff
@@ -0,0 +1,82 @@
...
+#include <cstdio>
+#include <unordered_map>
+#include <mutex>
+#include <string>
+
+#include "gdb/jit-reader.h"
+
+GDB_DECLARE_GPL_COMPATIBLE_READER;
+
+// everything is stored inline as it is marshaled cross process by gdb
```
skmp, line 11: Move these to a shared header

skmp:
Did some final cleanups, will merge once it passes CI.

PR #1789, opened 2022-06-20
AVX initial groundwork

lioncash:
Provides a minimal as possible set of changes to modify various bits of FEX to make it ready to start the implementation of AVX.

This essentially just expands the core state xmm registers and ensures that nothing breaks due to that.

Subsequent PRs will make use of the extended state/implement IR handling for it

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -209,19 +209,11 @@ void Arm64Emitter::SpillStaticRegs(bool FPRs, uint32_t GPRSpillMask, uint32_t FP
     }
 
     if (FPRs) {
-      for (size_t i = 0; i < SRAFPR.size(); i+=2) {
-        auto Reg1 = SRAFPR[i];
-        auto Reg2 = SRAFPR[i+1];
+      for (size_t i = 0; i < SRAFPR.size(); i++) {
+        const auto Reg = SRAFPR[i];
 
-        if (((1U << Reg1.GetCode()) & FPRSpillMask) &&
-            ((1U << Reg2.GetCode()) & FPRSpillMask)) {
-          stp(Reg1.Q(), Reg2.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
-        }
-        else if (((1U << Reg1.GetCode()) & FPRSpillMask)) {
-          str(Reg1.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
-        }
-        else if (((1U << Reg2.GetCode()) & FPRSpillMask)) {
-          str(Reg2.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i+1][0])));
+        if (((1U << Reg.GetCode()) & FPRSpillMask) != 0) {
+          str(Reg.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
         }
       }
     }
@@ -246,19 +238,11 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
     }
 
     if (FPRs) {
-      for (size_t i = 0; i < SRAFPR.size(); i+=2) {
-        auto Reg1 = SRAFPR[i];
-        auto Reg2 = SRAFPR[i+1];
+      for (size_t i = 0; i < SRAFPR.size(); i++) {
+        const auto Reg = SRAFPR[i];
 
-        if (((1U << Reg1.GetCode()) & FPRFillMask) &&
-            ((1U << Reg2.GetCode()) & FPRFillMask)) {
-          ldp(Reg1.Q(), Reg2.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
-        }
-        else if (((1U << Reg1.GetCode()) & FPRFillMask)) {
-          ldr(Reg1.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
-        }
-        else if (((1U << Reg2.GetCode()) & FPRFillMask)) {
-          ldr(Reg2.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i+1][0])));
+        if (((1U << Reg.GetCode()) & FPRFillMask) != 0) {
+          ldr(Reg.Q(), MemOperand(STATE, offsetof(FEXCore::Core::CpuStateFrame, State.xmm[i][0])));
         }
       }
     }
--- a/External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
@@ -99,6 +99,7 @@ void Dispatcher::RestoreThreadState(FEXCore::Core::InternalThreadState *Thread, 
     SignalFrames.pop();
   }
 
+  const bool IsAVXEnabled = CTX->Config.EnableAVX;
   uintptr_t NewSP = OldSP;
   auto Context = reinterpret_cast<ArchHelpers::Context::ContextBackup*>(NewSP);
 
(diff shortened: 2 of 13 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -209,19 +209,11 @@ void Arm64Emitter::SpillStaticRegs(bool FPRs, uint32_t GPRSpillMask, uint32_t FP
     }
 
     if (FPRs) {
-      for (size_t i = 0; i < SRAFPR.size(); i+=2) {
```
Sonicadvance1, line 212: Are these Arm64Emitter changes expected to be done here?
If so, this seems like a de-optimization.

lioncash: If that isn't done, then we'll end up erroneously loading/storing the upper ymm lanes when we shouldn't be, since stp and ldp no longer cross over to the second xmm register

Sonicadvance1: Ah right

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -274,6 +289,19 @@ static uint32_t ConvertSignalToError(int Signal, siginfo_t *HostSigInfo) {
   return 0;
 }
 
+template <typename T>
+static void SetXStateInfo(T* xstate) {
```
neobrain, line 293: This function doesn't seem to be used.

lioncash: Ah good catch, I forgot to add the usages back after finishing testing

Sonicadvance1 (dismissed):

External/FEXCore/include/FEXCore/Core/CoreState.h
```diff
@@ -14,7 +14,9 @@ namespace FEXCore::Core {
     uint64_t rip; ///< Current core's RIP. May not be entirely accurate while JIT is active
     uint64_t gregs[16];
     uint64_t : 64;
-    uint64_t xmm[16][2];
+    uint64_t : 64;
+    uint64_t : 64;
+    uint64_t xmm[16][4];
     uint16_t es, cs, ss, ds;
```
skmp, line 20: We could move 
```
    uint16_t es, cs, ss, ds;
    uint64_t gs;
    uint64_t fs;
```
info the padding

skmp:
I would prefer it if the shape of the register file was adapted to the running mode - AVX or no AVX. In the vast majority of cases, we won't be running on SVE enabled hosts for the foreseeable future, so it is a benefit to not always pay the cost.

In a similar direction, in exceptions we need to only save the AVX state if AVX is enabled. The AVX state is > the MINSTACK as defined by glibc up to 2.32, and there are several reports in the wild with applications breaking in.

lioncash:
How would we detect if AVX is enabled for the running executable? Do we have anything that allows us to check for what features an application was compiled with?

Sonicadvance1:
That's impossible to tell, even with static analysis the application could JIT something.
Instead on we should determine if the FEX host environment supports running FEX.
`if (x86_host) avx = true; else if (arm_host && supports_sve2 && sve2_vreg_size >= 256bit) avx = true`
Then have an additional configuration toggle to explicitly disable AVX so we can spin up application profiles for any game that has issues with AVX stack frames being too large.

This way we check if AVX is enabled on the FEX side and change the stack frame data around depending if it is enabled or not.

Sonicadvance1:
> Where would be a good place for storing the config result in a way that it can be accessed nicely from multiple different areas (dispatcher, JIT, etc)?

HostFeatures is the likely place to put this configuration option, since we initialize that before we do anything.

External/FEXCore/include/FEXCore/Core/CoreState.h
lioncash: When I move these into the padding the tests start breaking. Are we hardcoding offsets somewhere other than DeadContextStoreElimination pass that need to be accounted for? (latest commit is what I've done so far)

Sonicadvance1:
```
UNREACHABLE in /mnt/Work/Work/work/FEXNew/External/vixl/src/aarch64/assembler-aarch64.cc, line 5876

Program received signal SIGABRT, Aborted.
__pthread_kill_implementation (threadid=549621534752, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
(gdb) bt
#0  __pthread_kill_implementation (threadid=549621534752, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x0000007ff7b6f254 in __pthread_kill_internal (signo=6, threadid=<optimised out>) at ./nptl/pthread_kill.c:78
#2  0x0000007ff7b2a67c in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x0000007ff7b17130 in __GI_abort () at ./stdlib/abort.c:79
#4  0x0000005555c61b60 in vixl::aarch64::Assembler::LoadStoreMemOperand (this=0x7ff76b7040, addr=..., access_size_in_bytes_log2=4, option=vixl::aarch64::PreferUnscaledOffset) at /mnt/Work/Work/work/FEXNew/External/vixl/src/aarch64/assembler-aarch64.cc:5876
#5  0x0000005555c205d0 in vixl::aarch64::Assembler::LoadStore (this=0x7ff76b7040, rt=..., addr=..., op=vixl::aarch64::STR_q, option=vixl::aarch64::PreferUnscaledOffset) at /mnt/Work/Work/work/FEXNew/External/vixl/src/aarch64/assembler-aarch64.cc:5886
#6  0x0000005555c21610 in vixl::aarch64::Assembler::stur (this=0x7ff76b7040, rt=..., dst=..., option=vixl::aarch64::PreferUnscaledOffset) at /mnt/Work/Work/work/FEXNew/External/vixl/src/aarch64/assembler-aarch64.cc:1338
#7  0x00000055559994d4 in FEXCore::CPU::Arm64JITCore::Op_StoreContextIndexed (this=0x7ff76b7000, IROp=0x7ff761e6e5, Node=...) at /mnt/Work/Work/work/FEXNew/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp:428
#8  0x0000005555945e1c in FEXCore::CPU::Arm64JITCore::CompileCode (this=0x7ff76b7000, Entry=65539, IR=0x7ff771e530, DebugData=0x7ff76211c0, RAData=0x7ff771e500, GDBEnabled=false) at /mnt/Work/Work/work/FEXNew/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp:758
#9  0x0000005555741768 in FEXCore::Context::Context::CompileCode (this=0x7ff769e000, Thread=0x7ff76ad000, GuestRIP=65539) at /mnt/Work/Work/work/FEXNew/External/FEXCore/Source/Interface/Core/Core.cpp:974
#10 0x0000005555741bfc in FEXCore::Context::Context::CompileBlock (this=0x7ff769e000, Frame=0x7ff76ad1b0, GuestRIP=65539) at /mnt/Work/Work/work/FEXNew/External/FEXCore/Source/Interface/Core/Core.cpp:1014
#11 0x0000005555741aec in FEXCore::Context::Context::CompileBlockJit (this=0x7ff769e000, Frame=0x7ff76ad1b0, GuestRIP=65539) at /mnt/Work/Work/work/FEXNew/External/FEXCore/Source/Interface/Core/Core.cpp:985
#12 0x0000007ff7fc1334 in ?? ()
```
Looks like an alignment issue. mm members need to be 128-bit aligned and we didn't have a static_assert for it.

Sonicadvance1: Removing padding and adjusting the RCLSE classification resolves the problems here.

lioncash: Thanks. Added a static assert as well so that this doesn't slip through again

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -209,19 +209,11 @@ void Arm64Emitter::SpillStaticRegs(bool FPRs, uint32_t GPRSpillMask, uint32_t FP
...
-        auto Reg1 = SRAFPR[i];
-        auto Reg2 = SRAFPR[i+1];
+      for (size_t i = 0; i < SRAFPR.size(); i++) {
```
skmp, line 212: We lost stp's here even with avx off? :/

PR #1794, opened 2022-06-23
Dispatchers: Use thread local emitters for backend callbacks

skmp:
Can't use the dispatcher one if > 1 threads want to use it. Quick workaround using a thread_local.


Sonicadvance1 (approved):
Nice and disgusting. I love it.

PR #1799, opened 2022-06-24
FDUtils: Fix get_fdpath

skmp:
Our `get_fdpath` was broken and would include junk data after the filename, as `readlinkat` does not null terminate,it returns length instead.


Sonicadvance1 (approved):
Oops, good catch.

PR #1800, opened 2022-06-25
RFC: ThunkLibs: Callback binding support using copyable functions

skmp:
#### Overview
The code is a bit ugly, but it works.

Copyable functions are standard functions sprinkled with some inline assembly so we can determine their size and to bake a literal pool inside them. As they are copied around, they also cannot refer to any globals as there is no relocation performed on copy. Compiler options must guarantee that no stack check guards, etc are done as well.

Callbacks need copyable functions, or something similar, so we can bind a distinct guest callback to a distinct host callback without a data argument.

As a proof of concept, libX11 callbacks are implemented here.

#### Details
- mt-safe guest -> host and host -> guest lookup tables are kept
- function storage space is allocated using mmap in 16kb chunks
- function storage space is /leaked/. We could free on lib unload.

#### Alternative implementations
Another approach to this could be to use the `r11` trick as used in GCH, and make the copyable function naked + inline asm. This would guarantee no relocations are performed, and would not repack the arguments needlessly, at the expense of depending that the compiler doesn't trample over `r11` for x86-64 and the corresponding arm64 reg we choose.

Thoughts?

### DIFF
--- a/ThunkLibs/libXext/libXext_Host.cpp
+++ b/ThunkLibs/libXext/libXext_Host.cpp
@@ -4,6 +4,8 @@ tags: thunklibs|X11
 $end_info$
 */
 
+#include "common/Host.h"
+
 #include <stdio.h>
 
 #include <X11/Xlib.h>
@@ -38,7 +40,6 @@ extern "C" {
 #include <X11/extensions/syncproto.h>
 //#include <X11/extensions/XTest.h>
 
-#include "common/Host.h"
 #include <dlfcn.h>
 
 #include "ldr_ptrs.inl"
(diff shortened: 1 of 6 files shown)


ThunkLibs/libXext/libXext_Host.cpp
```diff
@@ -4,6 +4,8 @@ tags: thunklibs|X11
 $end_info$
 */
 
+#include "common/Host.h"
```
skmp, line 7: required because some X11 include redefines `min`

skmp:
Closing this in favor of #1833

PR #1801, opened 2022-06-25
ThunkLibs: silence warnings

skmp:
Just adding a bunch of [[maybe_unused]] in unchecked returns. Should these be converted to asserts @neobrain?


Sonicadvance1 (approved):

neobrain:
No strong opinion here, but FWIW the more explicit way to unconditionally ignore return values is a `(void)` cast.
Not sure if there are any scenarios we run into currently where these return an error in practice, so not sure an assert is safe to add here.

PR #1803, opened 2022-06-25
ThunkLibs/vulkan: Work around lack of generic callback support in VK_EXT_debug_report

neobrain:
The previous approach of disabling `VK_EXT_debug_report` did not work for `vulkaninfo`, since that program calls an uninitialized `vkCreateDebugReportCallbackEXT` pointer if the extension is not present.

Instead, implement vkCreateDebugReportCallbackEXT by ignoring all callbacks and providing a stub-callback to a host function.


Sonicadvance1 (approved):
This looks good but you probably also want to catch the `VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT` passed in to `vkCreateInstance` as well.
Working backwards, vkCreateInstance -> VkInstanceCreateInfo -> pNext -> VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT -> pfnCallback

neobrain:
> This looks good but you probably also want to catch the VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT passed in to vkCreateInstance as well.
Working backwards, vkCreateInstance -> VkInstanceCreateInfo -> pNext -> VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT -> pfnCallback

Added a quick hack to override the `pNext` member of the `const` input structure in-place if a debug report structure is found in the chain. Not the nicest thing to do, but duplicating the entire pNext chain to do this cleanly would be a tad overkill.

ThunkLibs/libvulkan/Host.cpp
```diff
@@ -56,7 +55,25 @@ static VkResult FEXFN_IMPL(vkCreateShaderModule)(VkDevice a_0, const VkShaderMod
...
 static VkResult FEXFN_IMPL(vkCreateInstance)(const VkInstanceCreateInfo* a_0, const VkAllocationCallbacks* a_1, VkInstance* a_2) {
+  VkDebugReportCallbackCreateInfoEXT stub_debug_report;
+  const VkBaseInStructure* vk_struct = reinterpret_cast<const VkBaseInStructure*>(a_0);
+  for (; vk_struct->pNext; vk_struct = vk_struct->pNext) {
+    // Override guest callbacks used for VK_EXT_debug_report
+    if (*reinterpret_cast<const VkStructureType*>(vk_struct->pNext) == VK_STRUCTURE_TYPE_DEBUG_REPORT_CALLBACK_CREATE_INFO_EXT) {
+      memcpy(&stub_debug_report, a_0->pNext, sizeof(stub_debug_report));
+      stub_debug_report.pfnCallback = DummyVkDebugReportCallback;
+      // Overwrite the pNext pointer, ignoring its const-qualifier
+      memcpy(const_cast<void*>(a_0->pNext), &stub_debug_report, sizeof(a_0->pNext));
```
Sonicadvance1, line 72: whew, spooky. Easy to catch though.
I guess this will just be temporary while other work is happening to get these debug callbacks working correctly?

neobrain: Yes, we can remove this code once support for generic guest-callbacks is added, which is very soon.

Sonicadvance1: Sounds good to me

PR #1804, opened 2022-06-25
Allow building thunks on a wider range of platforms

neobrain:
* Changes our x86 compiler discovery to use an included toolchain file. This allows users to customize cross-compilation more easily by modifying this file (or using a different one) instead of finding all the various references to `X86_C(XX)_COMPILER` throughout the build system.
* Runs `clang -print-resource-dir` during CMake generation so that the generator can set up its system include directories. (libclang provides no simple way to query this programmatically)
* `libclangTooling` is not available on openSUSE, but linking against `libclang-cpp` and LLVM instead works on both Ubuntu and openSUSE


Sonicadvance1 (approved):
nice

skmp:
> This allows users to customize cross-compilation more easily by modifying this file (or using a different one) instead of finding all the various references to X86_C(XX)_COMPILER throughout the build system.

Do any distros shit toolchain files like this?

neobrain:
>  Do any distros ship* toolchain files like this?
> 
> (so we don't have to write them / users don't have to go hunting)

Not that I'm aware of. The idea is primarily to have a central place for customization if people wanted to modify the compiler settings manually (or change tools other than the compiler such as the linker or assembler, if the toolchain required doing so).

skmp:
> > Do any distros ship* toolchain files like this?
> > (so we don't have to write them / users don't have to go hunting)
> 
> Not that I'm aware of. The idea is primarily to have a central place for customization if people wanted to modify the compiler settings manually (or change tools other than the compiler such as the linker or assembler, if the toolchain required doing so).

I see.

Can we also make such a config for the main project to force clang? It's a bit tiresome to have to overide CC every time a build folder is configured.

neobrain:
> Can we also make such a config for the main project to force clang? It's a bit tiresome to have to overide CC every time a build folder is configured.

I have no particular opinion on this. I don't personally mind setting CC/CXX manually for the main project.

PR #1805, opened 2022-06-27
ThunkLibs: Fix hostcall packer generation

skmp:
#### Overview

This fixes the generation of the hostcall packers, so we don't need to wrap thunks with C++ templates.

#### Depends
#1818


neobrain:
I'm not convinced it's beneficial to rid of PackedArguments at the expense of moving more logic to the generator. The template is very lightweight, and my measurements indicated it doesn't contribute significantly to the compile time of large thunklibs like libGL.

IMO we should move towards minimizing logic in the generator. The struct-repacking branch gave me a bad feeling about how robust a code generator with lots of manual branching is, and the PackedArguments template is a good way to avoid that pitfall. I'd prefer shelving this PR for now until I've done more tinkering with the idea.

skmp:
> I'm not convinced it's beneficial to rid of PackedArguments at the expense of moving more logic to the generator. The template is very lightweight, and my measurements indicated it doesn't contribute significantly to the compile time of large thunklibs like libGL.

The code was already there, the logic it wasn't moved. I just fixed the generation so it can be used. Before it was generated, broken, and we also had the PackedArguments.

> IMO we should move towards minimizing logic in the generator. The struct-repacking branch gave me a bad feeling about how robust a code generator with lots of manual branching is, and the PackedArguments template is a good way to avoid that pitfall. I'd prefer shelving this PR for now until I've done more tinkering with the idea.

I don't necessarily agree there. The `PackedArguments` template is a non-generalized huge thing, so it is obviously not a great solution either way.

Also, we can't use templates when doing cross bitness thunking, and evne for x86-64 / arm64 not all the packing rules are the same. We'll have to generate custom structs for everything for correctness, anyway.


Personally, I prefer codegeneration over templates for these things. Templates can be used where they are a good fit, like handling the conversions and stuff.

Sonicadvance1:
"This fixes the generation of the hostcall packers" I don't understand what is being fixed here or why it is broken.

skmp:
> "This fixes the generation of the hostcall packers" I don't understand what is being fixed here or why it is broken.

"so we don't need to wrap thunks with C++ templates."

skmp:
As mentioned in a previous comment, the hostpackers were broken, and unused. I don't understand what is there to not understand here.

neobrain:
FWIW https://github.com/FEX-Emu/FEX/pull/1868 adds a feature that makes the `_hostcall` symbols unnecessary and hence removes them entirely.

Sonicadvance1:
Nothing changed on this so far.

skmp:
> Nothing changed on this so far.

I've been blocked by lack of review/alternative proposals apart from "i don't like this / I don't think it is needed". Now, more time has to be wasted on rebases as well.

skmp:

PR #1809, opened 2022-06-28
RFC: Thunks: Callback support with trampoline generation in FEX

skmp:
#### Overview

This is an alternative to #1800, where the extra callback parameters are injected via a custom abi register, similar to the GCH.

This also includes support for the generator to automatically generate the trampolines on guest -> host, and to automatically get the target from trampolines on host -> guest.

This also removes a lot of complexity, as callbacks packs are identified by an sha256 id that is exported from the host, and we don't need to export the guest unpackers to the guest.

As a side effect of this, the marshaling needs to happen on the guest side. Can be done from the host as well, with some changes in the thunks loading api.

arm64 version isn't tested, but it's a matter of implementing `HostToGuestTrampolineTemplate` in asm and defining `CUSTOM_ABI_GUESTCALL` for arm64.

#### Details
fex thunks api is extended with
- `IsLibLoaded (char * Name -> bool)`
- `HostTrampolineForGuestcall (IR::SHA256Sum *HostPacker; uintptr_t GuestUnpacker; uintptr_t GuestTarget; -> uintptr_t)`
- `GuestcallTargetForHostTrampoline (uintptr_t HostTrampoline -> uintptr_t)`

gen.cpp is extended with
- Support for automatic callback pack (for host) and callback unpack (for guest) generation
- Support for MAKE_CALLBACK on the guest to have the callback sha256s
- Support for automatic guest callback -> host trampoline (performed by default for callback params)
- Support for automatic trampoline -> guest mapping via the `fexgen::returns_from_host_trampoline` annotation

ThunkLibs have been updated to reflect these changes

#### Todo
- [ ] null pointer handling

#### Depends
#1805


skmp:
A couple of notes
- It would be nice to modify GCH interface to be orthogonal by adding a second, optional parameter
- Callback ptr should be after argsrv in HCG

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -11,47 +11,124 @@ tags: glue|thunks
...
+    "0: \n"
+    ".quad 0, 0, 0, 0 \n"
+  );
+#elif defined(_M_ARM_64)
+  asm(
+    "adr x11, 0f \n"
+    "ldr x16, [x11] \n"
+    "br x16 \n"
+    ".align 16 \n"
+    "0: \n" // CallCallback
```
neobrain, line 78:
```suggestion
    "0: \n" // TrampolineInstanceInfo
```

neobrain: Any reason for naming the label `0` and commenting it instead of just `TrampolineInstanceInfo:`?

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -156,6 +234,125 @@ namespace FEXCore {
...
+            auto found = That->GuestcallToHostTrampoline.find(gci);
+            if (found != That->GuestcallToHostTrampoline.end()) {
+              rv = found->second;
+              return;
+            }
+          }
+
+          auto HostPackerEntry = That->Thunks.find(*HostPacker);
+          if (HostPackerEntry == That->Thunks.end()) {
+            rv = 0;
```
neobrain, line 293: This should be a fatal error AFAICT.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
skmp: 0 is a special label that can exist multiple times [0..9 are all special labels]. 0f means 0forward, ie the next occurance of 0.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
skmp: Not necessarily - it is up to the called to decide what to do there. Ie, a function may return a non-trampoline host pointer registered via some thunked library, and might chose to register a GCH for it.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -156,6 +234,125 @@ namespace FEXCore {
...
+            auto CTX = Thread->CTX;
+            auto That = reinterpret_cast<ThunkHandler_impl*>(CTX->ThunkHandler.get());
+
+            {
+                std::shared_lock lk(That->ThunksMutex);
+                rv = That->Libs.contains(Name);
+            }
+        }
+
+        static void HostTrampolineForGuestcall(void* ArgsRV) {
```
neobrain, line 254: In the unlikely case of requesting a trampoline to a GuestTarget for which there already was a trampoline with different GuestUnpacker registered, this will leak the GuestcalInfo instance, right? Wouldn't it be safe to discard the old GuestcallToHostTrampoline entry in that scenario?

neobrain:
This breaks vulkaninfo and glxinfo on ARM for me. To get the branch working on ARM, I rebased on latest `main` and used `asm("mov %0, x11" : "=r" (guestcall))` to define `CUSTOM_ABI_GUESTCALL`.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
skmp: It will keep both of them.

This is be needed as the unpacker also does the marshaling, the same function can need different trampolines - eg imagine the same callback getting passed to libA, and libB, both of which need to generate trampolines. There's no guarantee that the marshaling done in libA is compatible with the one done in libB.

In another very unlikely scenario, a function is guaranteed to be callable with different function signature from the cdecl/pcs abi, as freeing the stack frame is responsibility of the caller, and an application could, in theory, call the same function entry with two different signatures, and expect it to work.

This does causes an issue as `Trampoline(Guest, Packer1) != Trampoline(Guest, Packer2)` even though `Guest == Guest`. I don't think there's much we can do there, though.

skmp:
> This breaks vulkaninfo and glxinfo on ARM for me. To get the branch working on ARM, I rebased on latest `main` and used `asm("mov %0, x11" : "=r" (guestcall))` to define `CUSTOM_ABI_GUESTCALL`.

Yeah, I haven't built on arm for this. Good to know the rest of the arm64 asm was working as intended !

skmp:
@neobrain can this be closed as #1833 is merged, or are there some features from this one that you're interested in?

neobrain:
> @neobrain can this be closed as #1833 is merged, or are there some features from this one that you're interested in?

I don't think there was anything else, though it's hard to keep track since the changes were all thrown together. Closing for now.

PR #1824, opened 2022-06-30
unittests: Adds 32-bit into test 

### LINKED ISSUE #1808: Create unit test for overflow exception checking
Brought up by #1806 that we don't have a unit test that is testing our overflow exception handling.
While I tested this locally, we don't have it in CI.
Create a test for this.

Sonicadvance1:
This is a very basic test to check for overflow exception.
`into` instruction only exists on 32-bit, so it needs to not be built as
a 64-bit executable.

Fixes #1808

### DIFF
--- a/unittests/CMakeLists.txt
+++ b/unittests/CMakeLists.txt
@@ -10,7 +10,6 @@ if (BUILD_THUNKS)
   add_subdirectory(ThunkLibs)
 endif()
 
-
 if (BUILD_FEX_LINUX_TESTS)
   add_subdirectory(FEXLinuxTests/)
 endif()
--- a/unittests/FEXLinuxTests/CMakeLists.txt
+++ b/unittests/FEXLinuxTests/CMakeLists.txt
@@ -4,73 +4,89 @@ ExternalProject_Add(FEXLinuxTests
   SOURCE_DIR "${CMAKE_CURRENT_SOURCE_DIR}/tests"
   BINARY_DIR "${CMAKE_CURRENT_BINARY_DIR}/FEXLinuxTests"
   CMAKE_ARGS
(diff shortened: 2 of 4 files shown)


skmp:
Hmm, why not introduce another small meta tag that says to compile this only for x86-32, similar to the other meta flags on the tests? Sounds simpler than having a second build folder and all.

neobrain:
> Hmm, why not introduce another small meta tag that says to compile this only for x86-32, similar to the other meta flags on the tests? Sounds simpler than having a second build folder and all.

Yeah, we could have two separate ExternalProject_Add calls on a single `FEXLinuxTests` directory, one using the existing CMake toolchain file and one using a new 32-bit toolchain file. Within `FEXLinuxTests/CMakeLists.txt`, we could then check `CMAKE_SYSTEM_PROCESSOR` for tests that need to be architecture-specific.

The benefit of this is that we could share tests between 32-bit and 64-bit. If we don't need this, it would merely clean up the source structure a bit (which is still good, but there's also some trade-off in how complex the inner CMakeLists.txt becomes then).

Sonicadvance1:
This is now much lower impact, using the original FEXLinuxTests location. Main code movement comes from putting the glob loop in to a function. Calling it once with the 32bit+64bit tests and then once for the 32-bit only tests. Passing the supported bitness to the function.

neobrain (approved):
I have a hunch we can do with an even simpler build setup, will experiment and create a follow-up if it turns out successful. This one is fine as is for now.

skmp:
My mind is a bit foggy, but I'm fairly certain the fexlinuxtests are already built as 32 and 64 bit

Sonicadvance1:
Yes, they are already built as both 32-bit and 64-bit. I need explicit support for tests that only build as 32-bit here. `into` is only available as an instruction on 32-bit, it's invalid in 64-bit mode.

PR #1826, opened 2022-07-01
FEXGDBReader: Fix install path

Sonicadvance1:


skmp:
I thought a bit about this - it needs to follow gdb's path, not ours. I guess if you install in another place, then you'll also have gdb configured that way?

Sonicadvance1:
Sadly if you hardcode the absolute path then installing won't work if you're using DESTDIR to build packages.
Will just need to be careful if you're installing to `/usr/local` instead of `/usr` with `ninja install`

skmp:
yeah, what i'm saying is, if you install fex to /usr/local and gdb look for this on /usr/gdb then it won't work

Sonicadvance1:
I don't think there is a way to solve this without breaking someone installing to `/usr/local`

skmp:
does gdb have a pgk-config that gives this path? If yes, maybe we can use that?

Sonicadvance1:
Maybe? I haven't looked.

skmp:
no, it doesn't. GDB shows a helpful error if not found, so I guess this is ok for now.

skmp (approved):

PR #1828, opened 2022-07-03
RFC: 64 and 32 bit GL thunks

skmp:
I've been down with covid the past few days, and figured i'll take a stab at this during the fever breaks. YMMV, code quality def. not the best.

#### Overview
TBD.

glxgears, glxinfo work, both for 32 and 64 bits. jemalloc might need to be compiled with DSS disabled.

Needs `i686-linux-gnu-gcc`, `x86_64-linux-gnu-gcc` and some 32 bit libs (libx11-dev:i386, libxmu-dev:i386, maybe others?)

#### Details
- Create separate x86/x86_64 toolchain files, enable them for x86 builds as well
- Fixups in the FEXCore for 32-bit thunk support (ThunkOp, GCH, Wraping)
- Fixups in thunk for 32-bit support (GCH, Wraping)
- Build guest thunks twice, once as 32 bits, and once as 64 bits

#### Depends
#1817 as only GL/GLX are thunked, not X11. I don't think X11 is feasible.


Sonicadvance1:
This mostly needs to be split out to more independent changes.
- PR for adding cross-compile cmake toolchain file
- PR for selectively choosing which libraries are currently compatible with 32-bit and 64-bit. Only compiling when supported rather than commenting out
- Splitting out whatever the memory reserving stuff is doing
- Splitting the Filemanagement stuff out
- Splitting out the other PR commits from this one while work is happening on them.

skmp:
Agreed. This is just an RFC on the concept, so I can start chopping it up. I'll proceed with that.

neobrain:
Agreed with @Sonicadvance1 's comment. Relatedly, see my comment from https://github.com/FEX-Emu/FEX/pull/1734:
> 
> As mentioned in other channels, what I need to review your ThunkLibs PRs is better separation of changes and documentation. The bare minimum needed is a rudimentary explanation of what problem each of the patches solves. Otherwise, we have no solid basis for discussion and I won't be able to gauge how the changes will impact my ongoing work.
> 
> I won't repeat another review like https://github.com/FEX-Emu/FEX/pull/1809 where I had to spend a full week until I understood what you're doing and actually got it into a working state. Since I'm continuing to drive forward the bigger-scheme thunking work, I don't have capacity to carry early-state PRs out of prototype state.

It's unclear to me how this PR relates to #1817, so I'll also quote my comment from that thread here:

> As for this PR, I don't know if we critically need to merge it right now since you haven't outlined your specific future plans about it. Currently, I see no drawback from leaving this unmerged until some specific later date so we can continue evaluating limitations of "regular" GL thunking. This will save us unnecessary refactoring and feature creep in the generator which would come back and haunt us at a later point. If the chosen date passes or if we hit a hard blocker in the regular GL thunks, we can still integrate FGL at that point.
> 
> As for how to integrate FGL, @Sonicadvance1's proposal sounds reasonable. Since this is more of a stopgap solution it makes sense to minimize code sharing between FGL and the other thunk libraries. (If we find regular-GL hits a blocker it would be preferable to have stronger integration, but again that's something we would be in a better position to decide later.) The separation must be strong enough not to overlap with any of my generator changes, and it must be possible to safely make changes to one code path without having to retest the other all the time. The last thing we need is two separate configurations where at least one is constantly broken.
> 
> Finally, I need to stress that until https://github.com/FEX-Emu/FEX/pull/1611 is merged, I'm not in a position where I can review generator or build system changes unless they are clearly communicated and well-motivated. This includes massively-copy-pasted code that would need to be updated manually if changes to the generator are made, as such generator changes are still happening frequently.

skmp:
> Agreed with @Sonicadvance1 's comment. Relatedly, see my comment from #1734:
> 
> > As mentioned in other channels, what I need to review your ThunkLibs PRs is better separation of changes and documentation. The bare minimum needed is a rudimentary explanation of what problem each of the patches solves. Otherwise, we have no solid basis for discussion and I won't be able to gauge how the changes will impact my ongoing work.
> > I won't repeat another review like #1809 where I had to spend a full week until I understood what you're doing and actually got it into a working state. Since I'm continuing to drive forward the bigger-scheme thunking work, I don't have capacity to carry early-state PRs out of prototype state.

We've been through this a 1000 times. This is not the linux kernel, it's a volatile and fast moving codebase. I can't be spending most of my time documenting and splitting to cute PRs, when we have /major/ /disruptive/ chunks of work in front of us that will change how everything works.

As far as I understand, I'll have to move forward without your review on this.

> 
> It's unclear to me how this PR relates to #1817, so I'll also quote my comment from that thread here:
> 
> > As for this PR, I don't know if we critically need to merge it right now since you haven't outlined your specific future plans about it. Currently, I see no drawback from leaving this unmerged until some specific later date so we can continue evaluating limitations of "regular" GL thunking. This will save us unnecessary refactoring and feature creep in the generator which would come back and haunt us at a later point. If the chosen date passes or if we hit a hard blocker in the regular GL thunks, we can still integrate FGL at that point.
> > As for how to integrate FGL, @Sonicadvance1's proposal sounds reasonable. Since this is more of a stopgap solution it makes sense to minimize code sharing between FGL and the other thunk libraries. (If we find regular-GL hits a blocker it would be preferable to have stronger integration, but again that's something we would be in a better position to decide later.) The separation must be strong enough not to overlap with any of my generator changes, and it must be possible to safely make changes to one code path without having to retest the other all the time. The last thing we need is two separate configurations where at least one is constantly broken.
> > Finally, I need to stress that until #1611 is merged, I'm not in a position where I can review generator or build system changes unless they are clearly communicated and well-motivated. This includes massively-copy-pasted code that would need to be updated manually if changes to the generator are made, as such generator changes are still happening frequently.

Quoting from the PR description
> https://github.com/FEX-Emu/FEX/pull/1817 as only GL/GLX are thunked, not X11. I don't think X11 is feasible.

This needs the FGL bits as X11 is not thunked in 32 bits, and implementing that is a huge chunk of work that will likely not work smoothly at all.

Again, to re-iterate the value of this RFC/PoC
- Allows us to run 32-bit executables, including steam client itself and a part of the steam library, with GL thunking
- This also enables running steam on NV platforms, such as Orin.

Interactions with other PRs
- #1891 For the base support for correctly building for the guest systems
- #1902 For cleaner 32/64 bit thunk separation (this RFC will need to be re-worked on top of that, once the issues are resolved there)

skmp:

PR #1830, opened 2022-07-05
Support EroFS

Sonicadvance1:
EroFS is a new lightweight compressed filesystem format that is designed around low power, low memory, low capacity Android devices.
Android 13 is adopting this file format as their new partition format for their read-only system partitions.
Research paper is here: https://www.usenix.org/system/files/atc19-gao.pdf

One of the biggest improvements is the reduction of read amplification in their paper.

For us there is actually a more significant win outside of read amplification. SquashFS has a serious bottleneck where multiple threads accessing the squashfuse server serialize all of their data to one thread.
In my benchmarking this means that you can get ~1.8GB/s out of zstd squashfuse due to this threading limitation.
erofsfuse has no such limitation and it actually dynamically scales the number of fuse worker threads to compensate for the amount of work it is consuming. This allows in my benchmarking to get ~10.8GB/s out of erofs.

The major win here is that we will no longer serialize all accesses to a single thread. Which is huge considering we are going to be running potentially hundreds of processes which are accessing this.

The one detriment that erofs has is that the compression used is LZ4/LZ4HC instead of ZSTD. Which actually works to its benefit since LZ4 decompression is significantly faster than LZ4.

According to zstd's own benchmarking, LZ4 got 4.5GB/s and ZSTD in its fastest mode got 2.3GB/s.


Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -402,27 +431,94 @@ namespace WebFileFetcher {
...
+          }
+          else if (strcmp(DataValue, "erofs") == 0) {
+            Target.Type = FileTargets::TYPE_EROFS;
+          }
+          else {
+            Target.Type = FileTargets::TYPE_UNKNOWN;
+          }
+        }
+      }
+      if ((Target.Type == FileTargets::TYPE_SQUASHFS && WorkingAppsTester::Has_Squashfuse) ||
```
Sonicadvance1, line 514: oop, need to change this check to `has_SquashFuse || has_unsquashfs`

skmp (approved):
Looks okay. Though, we probably want to change the use of "squashfs" to something more generic to avoid confusion in the future (source file / variable / user prompts / etc)

Source/Common/FileFormatCheck.cpp
```diff
@@ -63,4 +63,54 @@ namespace FEX::FormatCheck {
...
+
+    if (FileSize < SUPERBLOCK_SIZE) {
+      return false;
+    }
+
+    if (!File.seekg(HEADER_OFFSET, std::fstream::beg)) {
+      return false;
+    }
+
+    if (!File.read(reinterpret_cast<char*>(&Header), sizeof(EroFSHeader))) {
```
neobrain, line 110: This is overly defensive. `fstream::read` already errors out if the file is not large enough:

```suggestion
    if (!File.read(reinterpret_cast<char*>(&Header), sizeof(EroFSHeader))) {
```

Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -402,27 +431,96 @@ namespace WebFileFetcher {
...
-
-      Targets.emplace_back(Target);
+      Target.DistroName = json_getName(RootItem);
+
+      for (json_t const* DataItem = json_getChild(RootItem);
+        DataItem != nullptr;
+        DataItem = json_getSibling(DataItem)) {
+        auto DataName = json_getName(DataItem);
+
+        if (strcmp(DataName, "DistroMatch") == 0) {
```
neobrain, line 489: Minor cleanup, not critical but figured it's worth mentioning since the strcmps add a bunch of noise:

```suggestion
        auto DataName = std::string_view { json_getName(DataItem)};

        if (DataName == "DistroMatch") {
```

Source/Common/FileFormatCheck.cpp
Sonicadvance1: Resolved. seekg to HEADER_OFFSET is still required.

Source/Tools/FEXRootFSFetcher/Main.cpp
Sonicadvance1: Resolved.

Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -344,9 +365,17 @@ namespace WebFileFetcher {
 
     // This is the hash of the file
     std::string Hash;
+
+    // FileType
+    enum FileType {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

PR #1833, opened 2022-07-06
Thunks: Implement generic callback support

neobrain:
# Overview

This PR adds a generic framework for thunking APIs with function pointer arguments ("callbacks"). In other words, it adds support for calling *guest function pointers* from the host. (#1760 implemented the converse already, i.e. calling host function pointers on the guest).

The previously used manual approach of wrapping the original function pointer in the callback argument data was difficult to reason about and only supported a specific kind of API. It is hence removed in favor of the new framework.

The implementation here is based on @skmp's work in #1809. This PR supersedes that PR and #1800.

# Implementation

Before guest->host transitions, any application-provided guest function pointer is replaced by a host-callable trampoline. When called by the host, the trampoline packs its arguments into an ABI-agnostic structure (similar to the packing functions to normal functions), initiates a host->guest transition, unpacks the argument structure, and invokes the original guest function pointer.

The trampoline is just-in-time generated per-function by copying a precompiled function and patching a 32-byte data section with the target guest function pointer and information required to call it. Similarly to the tricks used in #1760, the guest unpacker queries the target function pointer address from a hidden argument written to r11 by the trampoline. Trampolines are cached per-pointer and per-signature.

# Possible regressions

APIs using the `returns_guest_pointer` annotation may now actually return a host pointer due to the implicit conversion. This affects functions like XSetErrorHandler, which return the old error handler to the caller. Applications that make use of this returned pointer and expect it to match a previously provided one could now fail.

Similarly, there are no safeguards against applications that retrieve a host function pointer and use it as a callback for other functions.

Both of these issues will be addressed by follow-up work to integrate generic callbacks with host-callable guest-functions (#1760). For the time being, these issues do not prevent glxinfo, glxgears, vulkaninfo, or vkcube from working with GL/X11 thunks enabled.

# Impact on thunk interface definitions

No changes are needed. Many APIs that previously required `fexgen::callback_stub` annotations now don't require any annotations.

### DIFF
--- a/ThunkLibs/include/common/Guest.h
+++ b/ThunkLibs/include/common/Guest.h
@@ -157,7 +157,7 @@ template<auto Thunk, typename Result, typename... Args>
 inline Result CallHostThunkFromRuntimePointer(Args... args) {
 #ifndef _M_ARM_64
     uintptr_t host_addr;
-    asm("mov %%r11, %0" : "=r" (host_addr));
+    asm volatile("mov %%r11, %0" : "=r" (host_addr));
 #else
     uintptr_t host_addr = 0;
 #endif


External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -27,19 +30,74 @@ tags: glue|thunks
...
+    ".quad 0, 0, 0, 0 \n" // TrampolineInstanceInfo
+  );
+#elif defined(_M_ARM_64)
+  asm(
+    "adr x11, 0f \n"
+    "ldr x16, [x11] \n"
+    "br x16 \n"
+    // Manually align to the next 8-byte boundary
+    // NOTE: GCC over-aligns to a full page when using .align directives on ARM (last tested on GCC 11.2)
+    "mov x0, x0 \n"
```
Sonicadvance1, line 53: This should be explicit nop.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -27,19 +30,74 @@ tags: glue|thunks
...
+        // Hash only the target address, which is generally unique.
+        // For the unlikely case of a hash collision, std::unordered_map still picks the correct bucket entry.
+        return std::hash<uintptr_t>{}(x.GuestTarget);
+      }
+    };
+
+    // Hash the least significant bits only
+    struct TruncatingSHA256Hash {
+      size_t operator()(const FEXCore::IR::SHA256Sum& SHA256Sum) const noexcept {
+        return std::hash<size_t>{}((const size_t&)SHA256Sum);
```
skmp, line 93: As `SHA256Sum`'s has random bit distribution, we don't really need to re-hash here

Sonicadvance1: uniform bit distribution is probably better for the comment to explain why not needing to hash again isn't necessary.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -119,13 +188,107 @@ namespace FEXCore {
...
+            ThunkHandler->HostTrampolineInstanceDataAvailable = allocation_step;
+            ThunkHandler->HostTrampolineInstanceDataPtr = (uint8_t *)mmap(
+                0, ThunkHandler->HostTrampolineInstanceDataAvailable,
+                PROT_READ | PROT_WRITE | PROT_EXEC,
+                MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+            LOGMAN_THROW_A_FMT(ThunkHandler->HostTrampolineInstanceDataPtr != MAP_FAILED, "Failed to mmap HostTrampolineInstanceDataPtr");
+          }
+
+          const TrampolineInstanceInfo NewTrampolineInfo {
```
skmp, line 267: Why not initialize directly in the target here, to avoid the memcpy?

ThunkLibs/include/common/Host.h
```diff
@@ -86,3 +59,20 @@ struct fex_guest_function_ptr {
...
+  uintptr_t GuestUnpacker;
+  uintptr_t GuestTarget;
+};
+
+// Helper macro for reading an internal argument passed through the `r11`
+// host register. This macro must be placed at the very beginning of
+// the function it is used in.
+#if defined(_M_X86_64)
+#define LOAD_INTERNAL_GUESTPTR_ARG(target_variable) \
+  asm("mov %%r11, %0" : "=r" (target_variable))
```
skmp, line 74: these needs to be `volatile` to prevent the compiler reordering them

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -57,54 +50,13 @@ char* fexfn_impl_libX11_XGetICValues_internal(XIC a_0, size_t count, unsigned lo
...
-int XSetErrorHandlerCB(Display* a_0, XErrorEvent* a_1) {
-    static XSetErrorHandlerCB_Args argsrv { a_0, a_1};
-    
-    call_guest(callback_unpacks->libX11_XSetErrorHandlerCB, (void*) guest_handler, &argsrv);
-    
-    return argsrv.rv;
-}
-
-XSetErrorHandlerCBFN* fexfn_impl_libX11_XSetErrorHandler_internal(XErrorHandler a_0) {
+XErrorHandler fexfn_impl_libX11_XSetErrorHandler(XErrorHandler a_0) {
```
skmp, line 55: as `a_0` is no longer a guest handler, but a host handler, we can just return directly instead of storing in `XErrorHandler guest_handler`.

skmp (approved):
Overall, looks good.

Apart from the notes I left,

Why were the trampoline -> guest lookups not included as well? At least on the FEXCore side.

Also, `TruncatingSHA256Hash` should be defined in the IR headers so that it is globally available - is there some reason to have it as a hash override?

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
neobrain: This is just being more explicit about creating a new object and overwriting any existing data in the trampoline section. Previously it could appear as if we're just patching a few specific items.

gcc and clang are predictable in optimizing away memcpy calls. There's some overhead if they can't infer that the target pointer is aligned, but considering the logic that leads up to this point I don't anticipate any practical performance difference.

skmp: tbh, I find the initialize-and-copy way more confusing that just initializing inline myself - we can do like `*Dest = TrampolineInstanceInfo { ... }` which clearly initializes and copies, and we don't have to depend on fragile compiler optimisations :)

neobrain:
Thanks for the review!

> Why were the trampoline -> guest lookups not included as well? At least on the FEXCore side.

I have a different approach in mind than the one taken in your PR, but since it's logically separate it's not included in this PR.

> Also, TruncatingSHA256Hash should be defined in the IR headers so that it is globally available - is there some reason to have it as a hash override?

I don't know how if truncation makes this a suitable general-purpose hash. We can revisit if we have more use cases for storing SHA256Sum as a map key than this particular one.

ThunkLibs/include/common/Host.h
skmp: Also, I'd rather name the define to something that includes "custom abi" to make it clear what is happening there.

skmp: (sidenote: We also need volatile on the guest side as well)

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -119,13 +188,107 @@ namespace FEXCore {
...
+            std::shared_lock lk(ThunkHandler->ThunksMutex);
+
+            auto found = ThunkHandler->GuestcallToHostTrampoline.find(gci);
+            if (found != ThunkHandler->GuestcallToHostTrampoline.end()) {
+              args->rv = found->second;
+              return;
+            }
+          }
+
+          std::lock_guard lk(ThunkHandler->ThunksMutex);
```
Sonicadvance1, line 233: This would be such a nice use case for a shared_lock with the ability to promote to unique_lock semantics.
Sadly we don't have one of these yet.

Sonicadvance1 (approved):

PR #1836, opened 2022-07-07
Thunks: Soften error condition to be non-fatal

neobrain:
Dota Underlords hit this when querying Vulkan two different symbols (vkGetPhysicalDeviceFeatures2/vkGetPhysicalDeviceFeatures2KHR) that resolve to the same function. Ignoring the error in that specific case is safe, since the linked guest functions have the same implementation.

Ideally, we'll make the target guest wrappers unique by-signature to avoid this conflict. Then we can treat the error as fatal again.


skmp:
> Ideally, we'll make the target guest wrappers unique by-signature to avoid this conflict. Then we can treat the error as fatal again.

It would be sweet if we did this for all the automatically generated ones - we might still need custom marshaling per function, even if they share the same types.

PR #1837, opened 2022-07-07
Vulkan: Handle queries for unknown functions more gracefully

neobrain:
The Vulkan driver may support functions that are not listed in FEX's `libvulkan_interface.cpp` definitions. FEX will be unable to call those functions, however it's useful to allow the application to pass anyway since often these functions are just eagerly queried and then never used.

For Release builds, we return `nullptr` in case the application has a fallback code path. For Debug builds, a function is injected that fatally errors when called.

### DIFF
--- a/ThunkLibs/libvulkan/Guest.cpp
+++ b/ThunkLibs/libvulkan/Guest.cpp
@@ -36,18 +36,40 @@ const std::unordered_map<std::string_view, uintptr_t /* guest function address *
 #undef PAIR
     });
 
+// This variable controls the behavior of vkGetDevice/InstanceProcAddr for functions we don't know the signature of:
+// - if false (default), we return a nullptr (since the application might have a fallback code path)
+// - if true, we return a stub function that fatally errors upon being called
+constexpr bool stub_unknown_functions = false;
+
+// Fatally erroring function with a thunk-like interface. This is used as a placeholder for unknown Vulkan functions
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Nice. This will fix Vulkan thunking in Aperture Desk Job.
Thanks for this!

ThunkLibs/libvulkan/Guest.cpp
```diff
@@ -36,18 +36,37 @@ const std::unordered_map<std::string_view, uintptr_t /* guest function address *
...
+    __builtin_trap();
+}
+
+static PFN_vkVoidFunction MakeGuestCallable(const char* origin, PFN_vkVoidFunction func, const char* name) {
+    auto It = HostPtrInvokers.find(name);
+    if (It == HostPtrInvokers.end()) {
+        // If we can't infer the function signature, return a fatally erroring function in Debug builds.
+        // For Release builds, return nullptr in case the application has a fallback code path.
+        fprintf(stderr, "%s: Unknown Vulkan function at address %p: %s\n", origin, func, name);
+#ifdef NDEBUG
```
skmp, line 52: I'd hide this behind a (thunk specific) #ifdef, not the generic NDEBUG, as it can affect application behaviour, and debug builds are expected to behave the same way

Sonicadvance1 (approved):

PR #1839, opened 2022-07-07
ValueDominanceValidation: Avoid stack exhaustion when aggregating predecessors

neobrain:
Noticed this while running _Forsaken Remastered_ using a Debug build of FEX.

The recursive algorithm used here previously led to deeply nested function calls, which eventually exhausted the available stack space. The simple non-recursive algorithm used now avoids this problem at the expense of small overhead.

In Forsaken Remastered, the overhead was not visibly noticeable (time spent outside of validation far outweighed time in this function).


Sonicadvance1 (approved):
Nice catch!

PR #1840, opened 2022-07-09
FEXCore: Adds assume optimizing LogManager function 

Sonicadvance1:
This comes in the form of a new define and a new function.

Assuming assert doesn't quite cover 100% of our asserting logging use
cases so we need to break this out.
If the compiler can't see through side effects at compile time for the
predicate then it'll throw a warning.

This gives us a nice behaviour. In the case that asserts are enabled, we
fall down the assert checking path. So if an assert fails to predicate
we will do an assert log like normal.
The change comes when we build release without asserts, we use the
`__builtin_assume` definition to allow the compiler to optimize around
assumptions. Since these aren't runtime asserts in release build, this
gives us some small optimizations in various locations.

Clang will very specifically do additional optimizations when you
provide it `__builtin_assume` directives that can really be worth it.

I thought of adding this as a freestanding FEXHeaderUtil function, but everywhere we would be using it would also want logging. So it doesn't make sense to duplicate functionality for assert+assume and some redundancy.

### DIFF
--- a/External/FEXCore/Source/Common/BitSet.h
+++ b/External/FEXCore/Source/Common/BitSet.h
@@ -20,12 +20,12 @@ struct BitSet final {
   ElementType *Memory;
   void Allocate(size_t Elements) {
     size_t AllocateSize = AlignUp(Elements, MinimumSizeBits) / MinimumSize;
-    LOGMAN_THROW_A_FMT((AllocateSize * MinimumSize) >= Elements, "Fail");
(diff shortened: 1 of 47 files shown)


neobrain:
Have you noticed any measurable performance / binary size differences from applying these?

Sonicadvance1:
Binary size has shifted a bit and some /very/ minor code shifting has occurred because of it. Nothing too major since we weren't throwing assume at the compiler in hot code.

PR #1841, opened 2022-07-09
Linux: Only install syscall handlers for the arch we launched with

Sonicadvance1:
We were registering syscall handlers to the temporary working vectors for
both x86 and x86_64 regardless of which bitness we launched with.

We then only installed the syscall handlers depending on bitness.
This was burning a decent amount of time and some memory on
initialization. Instead just don't register syscall handlers on the
other bitness.

Also stop wasting memory, clear the syscall registration vector after
we've registered everything.

Saves about 28KB of memory per process.

### DIFF
--- a/Source/Tests/LinuxSyscalls/x64/Syscalls.cpp
+++ b/Source/Tests/LinuxSyscalls/x64/Syscalls.cpp
@@ -13,6 +13,8 @@ $end_info$
 #include <map>
 
 namespace FEX::HLE::x64 {
+  RegisterSyscallInternalType SyscallRegisterHandler = FEX::HLE::RegisterSyscallInternalNop;
+
   void RegisterEpoll(FEX::HLE::SyscallHandler *const Handler);
   void RegisterFD();
   void RegisterInfo();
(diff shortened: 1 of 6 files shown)


Source/Tests/LinuxSyscalls/x64/Syscalls.cpp
```diff
@@ -153,6 +156,8 @@ namespace FEX::HLE::x64 {
 #endif
     }
 
+    syscalls_x64.clear();
```
neobrain, line 159: Haven't had a chance to look into this in detail, but mind you `.clear()` doesn't free the vector memory. You'll want something like `syscalls_x64 = {}` for that instead.

Sonicadvance1: ...What a fucking nightmare footgun that is.

Sonicadvance1: Resolved now though.

lioncash: Seems like only the change made it through for the syscalls_x32 vector. Was that intended?

lioncash (commented):
LGTM aside from the one review comment

Source/Tests/LinuxSyscalls/x64/Syscalls.cpp
Sonicadvance1: Oops, yep, not intentional. Fixed now.

PR #1844, opened 2022-07-12
Interpreter: Move argument names over to IR names

lioncash:
Converts the opaque args array usage over to the given IR argument names where applicable. Brings the interpreter in sync with the ARM JIT

I made these changes a while ago and I guess I forgot to actually make a PR for them (whoops!)


Sonicadvance1 (approved):

skmp (commented):
I'm loving it

PR #1845, opened 2022-07-13
FEXCore: Support synchronizing RIP on block entry through config

Sonicadvance1:
If an application is doing long jumps on exceptions then we need to
ensure that RIP is synchronized at least to block entry for some amount
of safety.

Due to our block linking which doesn't ensure that RIP is synchronized
on block entry, our exception handling wouldn't see a RIP change, thus
not going down the path that we jump to Dispatcher loop top on RIP
change.

Burn a couple of instructions on block entry to ensure that RIP
synchronized but only on config.

**Fixes Proton Experimental hangs when it is installing the VC runtimes.**

### DIFF
--- a/Data/AppConfig/VC_redist.x64.exe.json
+++ b/Data/AppConfig/VC_redist.x64.exe.json
@@ -1,0 +1,5 @@
+{
+  "Config": {
+    "x86dec_SynchronizeRIPOnAllBlocks": "1"
+  }
+}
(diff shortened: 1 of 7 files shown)


skmp (changes requested):
- It's best if this is done in the IR level
-> Just STORE_CONTEXT, RIP, EntryOffset (0)
-> Can be done in core.cpp as of the block entry

- It's best if config naming follows a nice pattern
-> we can use x86dec_<feature here>: value, thoughts?

Sonicadvance1:
>  we can use x86dec_: value, thoughts?
I don't mind naming it whatever you want. Do you just want the prefix on the config name?

Sonicadvance1:
Resolved comments.

PR #1846, opened 2022-07-13
x86_64: Migrate args over to named IR arguments

lioncash:
Now all of the existing CPU cores are consistent with each other

This also resolves a benign case where the wrong struct type was used in VRev64 (though no issues arose because layouts are the exact same).


Sonicadvance1 (approved):
Nice

PR #1851, opened 2022-07-18
Fix SOMA and sigaction definition

Sonicadvance1:
See commit messages.

Fixes SOMA, makes unity games more stable.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -500,10 +500,21 @@ namespace FEXCore::Context {
     ExecutionThreadHandler *Arg = reinterpret_cast<ExecutionThreadHandler*>(FEXCore::Allocator::malloc(sizeof(ExecutionThreadHandler)));
     Arg->This = this;
     Arg->Thread = Thread;
+    Thread->StartPaused = NeedToCheckXID;
     Thread->ExecutionThread = FEXCore::Threads::Thread::Create(ThreadHandler, Arg);
 
     // Wait for the thread to have started
     Thread->ThreadWaiting.Wait();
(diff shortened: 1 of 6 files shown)


External/FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -500,10 +500,23 @@ namespace FEXCore::Context {
...
     // Wait for the thread to have started
     Thread->ThreadWaiting.Wait();
+
+    if (NeedToCheckXID) {
+      // Only ever have one thread resetting the XID handler
+      // This will only ever fire from the first thread creation,
+      // But GLIBC overwrites our XID handler on first pthread creation.
+      //
+      // We start the first thread creation paused to ensure we don't have that
+      // thread racing us to do a XID operation.
```
neobrain, line 515: I can't make sense out of this comment. It seems to interleave what we do with why we do so, and then throws in a seemingly random fact about glibc behavior in the mix too. Any chance some restructuring would make this more clear?

neobrain: Reading some more it's starting to clear up. Maybe a structure like this would work:
1. "FEX needs a custom XID handler"
2. "glibc overwrites FEX's XID handler (always?) on first pthread creation"
3. ...

I still don't fully understand the details, but maybe that helps to come up with clearer phrasing.

Sonicadvance1: Updated the comment to hopefully clarify that we are stealing the signal handler back from glibc after first thread creation.

neobrain: Yes, that is much better. Cheers!

PR #1852, opened 2022-07-19
json_ir_generator: Remove Args() functions from IR structs

lioncash:
Since the IR argument can be accessed directly by name, this is no longer needed as a shorthand.

Also results in a decently smaller generated file


skmp:
Hmm, we still need to be able to access the args in some cases in an array (eg, syscalls)

lioncash:
That's still possible through the long-form way like:

```cpp
Op->Header.Args[i]
```

This was a shorthand I introduced a bit ago before I knew we could access args by name directly

PR #1856, opened 2022-07-22
Arm64Emitter: Re-add use of stp/ldp with hosts that don't support SVE2

### LINKED ISSUE #1847: Classify CoreState differently depending on if host supports AVX or not
Keep the XMM elements 256bit in size but classify them differently as a union or something.

```
union {
struct {
uint64_t XMM[16][4];
} SupportsAVX;
struct {
uint64_t XMM[16][2];
uint64_t _Pad[16][2];
} NoSupportsAVX;
};
```

This way we can still use LDP and STP still for non-SVE2 ARM.
Needs to set up the classification path to support both.

lioncash:
SVE2 is going to be a prerequisite for fully supporting AVX and previously hosts that don't support those requirements would be getting suboptimal codegen since the register file is sized differently and can't take advantage of STP/LDP-ing adjacent registers anymore

This re-adds the optimization by handling the case where AVX isn't supported on the host end of things.

Fixes #1847


Sonicadvance1 (approved):
Looks reasonable to me, just needs the HostRunner fixed

Sonicadvance1 (approved):

PR #1858, opened 2022-07-23
FEXRootFSFetcher: Add some options for automation without user intervention

Sonicadvance1:
Main options here are `-y` and `-a`.

Passing both will allow the user to download the exact match image, use
it as-is, and set the config to use the rootfs by default.

Additional options `--distro-name` and `--distro-option` allows one to
select a particular distro version.

eg: `FEXRootFSFetcher -y -a --distro-name=ubuntu --distro-version=22.04`
* Will download the first Ubuntu 22.04 compressed in the json list
  * Priority of erofs or squashfs depends on order in json and what the
host supports

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -1,3 +1,4 @@
+#include "OptionParser.h"
 #include "XXFileHash.h"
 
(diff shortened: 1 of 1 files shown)


skmp (approved):

QuintenQVD0:
For me the options are there but it sas it installed it in 1.9secons what does not look right

PR #1860, opened 2022-07-24
FEXConfig: Some quality of life improvements

Sonicadvance1:
Removes the "Load Default Options" menu option. This option was
confusing for new users and isn't necessary anymore.

Fixes the "Load Default" option so it actually populates the full
configuration layer in the face of partial configuration.

This is a /very/ common use case for new users that ran through
FEXRootFSFetcher, where the only configuration set is the RootFS.
The configuration would be visually confusing since the visual
representation for missing options wouldn't reflect their default
configuration state. "TSO Enabled" is an example where it would appear
disabled in the GUI, but it is default enabled.

Also fixes the issue that the default configuration window would just be
a 320x240 floating window in the center of the screen. This is due to
the window being a floating sub window in the dockspace by default, and
not docked.

Instead just remove the dockspace, it isn't serving us any purpose.
This means the child configuration window now maximizes to the window
size which is the desired behaviour from default.

Additionally only save the config file once. While the msg dialog is
open (2 seconds while it is open, or escape to make it go away
immediately) the program won't save the file again. This fixes an issue
that if you used the shortcut key to save the file, it would save the
file at the refresh rate of your screen. Which is 144hz on my setup, so
it spams my filesystem quite heavily.

### DIFF
--- a/Source/Tools/CommonGUI/IMGui.h
+++ b/Source/Tools/CommonGUI/IMGui.h
@@ -87,10 +87,10 @@ namespace FEX::GUI {
 
   static std::chrono::time_point<std::chrono::high_resolution_clock> LastUpdate{};
   constexpr auto UpdateTimeout = std::chrono::seconds(2);
-  void DrawUI(SDL_Window *window, std::function<void()> DrawFunction) {
-    bool Done{};
+  void DrawUI(SDL_Window *window, std::function<bool()> DrawFunction) {
+    bool Running {true};
     ImGuiIO& io = ImGui::GetIO();
-    while (!Done) {
+    while (Running) {
       SDL_Event event;
       auto Now = std::chrono::high_resolution_clock::now();
       auto Dur = Now - LastUpdate;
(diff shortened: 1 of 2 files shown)


neobrain (approved):
Yes please! 💯

Also, not sure if it's due to this PR or something else, but the issue where every key stroke was registered twice for me is fixed now. The screen flashes on saving are gone too.

Random other things we could improve in the future:
* `Ctrl+Q` should quit FEXConfig (currently only Alt+F4 can be used for that)
* `Open Default` automatically on startup
* Rename `Open Default` to something less ambiguous like `Open from default location`
* Warn on exit if config was modified but not saved

Sonicadvance1:
-Added `Ctrl+Q` to quit in addition to regular window close event.
- `Open Default` happens on application launch unless an argument is passed in
- Renamed `Open Default` to `Open from default location`
- Skipping `warn if unsaved` dialog since that will require inventing some new mechanism for Yes-No input in the UI.

PR #1861, opened 2022-07-25
Linux: Fixes hang in close_range 

Sonicadvance1:
It's common for Linux applications that use close_range to pass in ~0 as
the last FD. This was causing FEX to spin from [2, ~0U] in this loop.
This would take /forever/ to run.

Change over to an ordered map and use the map's range searching and
ranged erase to more quickly remove these elements.

Fixes a hang that occurs with first time Steam setup in
steam-linux-runtime heavy application `steam-runtime-identify-library-abi`.

Thanks to @koss on Discord for discovering and reporting this issue.

### DIFF
--- a/unittests/FEXLinuxTests/tests/fd/test_close_range.cpp
+++ b/unittests/FEXLinuxTests/tests/fd/test_close_range.cpp
@@ -1,0 +1,17 @@
+#include <cstdint>
+#include <unistd.h>
+
+int main() {
+  int fd_base = dup(STDOUT_FILENO);
+  for (size_t i = 0; i < 15; ++i) {
+    dup(fd_base);
+  }
+
+  // Specifically testing last as ~0U to ensure FEX doesn't hang
+  constexpr uint32_t SYS_close_range = 436;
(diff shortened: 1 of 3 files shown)


unittests/FEXLinuxTests/tests/fd/test_close_range.cpp
```diff
@@ -0,0 +1,14 @@
...
+
+int main() {
+  int fd_base = dup(STDOUT_FILENO);
+  for (size_t i = 0; i < 15; ++i) {
+    dup(fd_base);
+  }
+
+  // Specifically testing last as ~0U to ensure FEX doesn't hang
+  constexpr uint32_t SYS_close_range = 436;
+  ::syscall(SYS_close_range, fd_base, ~0U, 0);
```
neobrain, line 12: How about closing `fb_base+1` and then assert that 1) close(fd_base) succeeds since it wasn't closed by close_range 2) close (fd_base+1) fails since it's already closed?

Sonicadvance1: Sure thing. Switched over to that.

PR #1863, opened 2022-07-27
Thunks/X11: Distinguish between host and guest pointers in XFree

neobrain:
Depends on https://github.com/FEX-Emu/jemalloc/pull/3.

This function must be able to handle both guest heap pointers *and* host heap
pointers, so it only forwards to the native host library for the latter.

This is because Xlibint users allocate memory using internal macros aliasing
to libc's malloc but then they free using the function XFree. For libX11,
this is not a problem since the allocation happens in a thunked API function
(and hence on the host heap), but if a function from an unthunked library
accesses Xlibint, it will allocate on the guest heap.

One notable example where this was encountered is XF86VidModeGetAllModeLines.


ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -68,16 +73,44 @@ extern "C" {
...
+      return fexfn_pack_XFree(ptr);
+    } else {
+      free(ptr);
+      return 1;
+    }
+  }
+
+  void XFreeEventData(Display* display, XGenericEventCookie* cookie) {
+    // Has the same heap-mismatch issue as XFree, so we have to reimplement it manually
+    if (_XIsEventCookie(display, (XEvent*)cookie) && cookie->data) {
```
Sonicadvance1, line 105: Is this the only event that ever needs to be freed? No need to free cookie itself?

Sonicadvance1 (approved):

ThunkLibs/libX11/libX11_Guest.cpp
neobrain: Yeah, this reflects behavior of the proper XFreeEventData implementation. I'm guessing no other kinds use the `data` field, and the cookie itself is deallocated separately.

PR #1864, opened 2022-07-28
Syscalls: Removes staging vector usage

Sonicadvance1:
This removes about half a millisecond from syscall handler registration. But it is a bit noisy so hard to get exact numbers.

### DIFF
--- a/Source/Tests/LinuxSyscalls/Syscalls.h
+++ b/Source/Tests/LinuxSyscalls/Syscalls.h
@@ -52,27 +52,27 @@ namespace FEX::HLE {
 
 class SyscallHandler;
 class SignalDelegator;
-  void RegisterEpoll();
-  void RegisterFD(FEX::HLE::SyscallHandler *const Handler);
-  void RegisterFS(FEX::HLE::SyscallHandler *const Handler);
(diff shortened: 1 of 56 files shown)


Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -84,24 +84,6 @@ struct ExecveAtArgs {
...
-void (*)(int SyscallNumber,
-  int32_t HostSyscallNumber,
-  FEXCore::IR::SyscallFlags Flags,
-#ifdef DEBUG_STRACE
-  const std::string& TraceFormatString,
-#endif
-  void* SyscallHandler, int ArgumentCount);
-
-
-void RegisterSyscallInternalNop(int SyscallNumber,
```
neobrain, line 97: Is this still used somewhere? If not, then the definition in Syscalls.cpp should also be removed.

neobrain (approved):

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1: Oops, good catch. Yep, removed.

PR #1865, opened 2022-07-28
Cleanup syscalls headers

neobrain:
No interface changes, just throwing out unused code and shortening implementations.

### DIFF
--- a/Source/Tests/LinuxSyscalls/Syscalls.h
+++ b/Source/Tests/LinuxSyscalls/Syscalls.h
@@ -393,30 +393,6 @@ std::string CollectArgsFmtString() {
 #define ARG_TO_STR(tpy, str)
 #endif
 
-// Helper that allows us to create a variadic template lambda from a given signature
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Just to confirm, this was tested with `// #define DEBUG_STRACE` uncommented in Syscalls.h right?

neobrain:
> Just to confirm, this was tested with `// #define DEBUG_STRACE` uncommented in Syscalls.h right?

Yup. Tested running Ender Lilies, which still shows well-formatted debug messages like
```
[DEBUG] newfstatat(-100, /Build/++UE4/Sync/Engine/Binaries/ThirdParty/PhysX3/Linux/x86_64-unknown-linux-gnu/tls/x86_64, 0x7fffffffcd37, 0) = -2
```

Sonicadvance1 (approved):
TIL about `operator+` on lambdas and I love it!

skmp:
Hmm, this removes the inner impl warping structs from the registered lambdas, which means we won't get backtraces with the syscall name though. Was that intended?

PR #1869, opened 2022-07-29
vixl: Update

Sonicadvance1:
Removes three files from vixl that we aren't using. These take nearly
20 seconds apiece to compile so this just improves compile time


skmp (approved):

skmp:
Speaking of compile times, we're compiling thunks twice as well

PR #1877, opened 2022-08-01
FEXBash: Changes PS1 to hopefully help users

Sonicadvance1:
Currently FEXBash only outputs `FEXBash>` which gives weird docker
vibes. This can confuse users since they no longer see what folder they
are in.

Change this so it still has the user and path exposed like a typical PS1

eg: `FEXBash-ryanh@ryanh-TR2:/mnt/Work/Work/work/FEXNew/Build>`

### DIFF
--- a/Source/Tests/FEXBash.cpp
+++ b/Source/Tests/FEXBash.cpp
@@ -94,11 +94,11 @@ int main(int argc, char **argv, char **const envp) {
       PS1Env = envp[i];
     }
     else {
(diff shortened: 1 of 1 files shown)


skmp:
Why not simply make it `FEXBash(in rainbow color) $PS1` ?

Sonicadvance1:
Colour isn't guaranteed to be supported so I didn't want to potentially break anything.

skmp (approved):

PR #1879, opened 2022-08-03
Arm64Dispatcher: Amend memcpy in SpillSRA 

lioncash:
A missed portion when readding the stp/ldp optimization. This corrects the spilling of SRA FPRs in guest signals

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
@@ -580,9 +580,16 @@ void Arm64Dispatcher::SpillSRA(FEXCore::Core::InternalThreadState *Thread, void 
     Thread->CurrentFrame->State.gregs[i] = ArchHelpers::Context::GetArmReg(ucontext, SRA64[i].GetCode());
   }
 
-  for(int i = 0; i < SRAFPR.size(); i++) {
-    auto FPR = ArchHelpers::Context::GetArmFPR(ucontext, SRAFPR[i].GetCode());
-    memcpy(&Thread->CurrentFrame->State.xmm.avx.data[i][0], &FPR, sizeof(__uint128_t));
+  if (EmitterCTX->HostFeatures.SupportsAVX) {
+    for(int i = 0; i < SRAFPR.size(); i++) {
+      auto FPR = ArchHelpers::Context::GetArmFPR(ucontext, SRAFPR[i].GetCode());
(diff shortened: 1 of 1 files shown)


skmp (approved):

External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
```diff
@@ -572,17 +572,24 @@ size_t Arm64Dispatcher::GenerateInterpreterTrampoline(uint8_t *CodeBuffer) {
...
       // Skip this one, it's already spilled
       continue;
     }
     Thread->CurrentFrame->State.gregs[i] = ArchHelpers::Context::GetArmReg(ucontext, SRA64[i].GetCode());
   }
 
-  for(int i = 0; i < SRAFPR.size(); i++) {
-    auto FPR = ArchHelpers::Context::GetArmFPR(ucontext, SRAFPR[i].GetCode());
-    memcpy(&Thread->CurrentFrame->State.xmm.avx.data[i][0], &FPR, sizeof(__uint128_t));
+  if (EmitterCTX->HostFeatures.SupportsAVX) {
```
skmp, line 583: Shouldn't this be `GuestFeatures`?

lioncash: Nah, since the host feature option determines whether or not the AVX or SSE register file is used when accessing CPUState instances

skmp: How would that work with arm then?

lioncash: I don't understand what you mean? We check if the host is capable of emulating AVX (has at least 256-bit scalable vectors and also has SVE2) and if it's capable of emulating AVX we use the full length AVX entries in the `CPUState` instance (with the caveat being that the ldp and stp optimization for spills/fills can no longer be used. If the host cannot emulate AVX, then we fall back to using the SSE entries in the `CPUState` which does allow for the ldp/stp operations, so those platforms don't take a hit for a feature it can't support

We need to alter where the memcpy takes place, because the data offsets for each register in the `CPUState` will vary depending on whether we're emulating SSE only or if we emulate AVX

skmp: Ah, I thought it was checking if the host is supporting AVX - that's why I said it should be Guest, to check if the guest has AVX emulation enabled.

Sonicadvance1 (approved):

PR #1884, opened 2022-08-05
Thunks: Use external Vulkan-Headers 

Sonicadvance1:
This will help compiling on older distros which are shipping older
Vulkan-Headers.

We need to catch newer Vulkan features earlier than what distros ship
since it is highly common that users will update their drivers through
means that make their drivers be newer.

For example to build on Ubuntu 20.04 we will support symbols much newer
than what that verison of the distro supports.
We could wrap all uses of newer features behind `#ifdef` checks, or
include the newest version of the loader that FEX itself supports.

The submodule is much cleaner and resolves the issue of drivers
supporting newer vulkan versions than libvulkan-dev.
Again super common with Nvidia blob users, kisak-ppa users, or people
updating mesa directly to be on the bleeding edge.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -49,3 +49,7 @@
 	shallow = true
 	path = External/robin-map
 	url = https://github.com/Tessil/robin-map.git
+[submodule "External/Vulkan-Headers"]
+	shallow = true
+	path = External/Vulkan-Headers
+	url = https://github.com/KhronosGroup/Vulkan-Headers.git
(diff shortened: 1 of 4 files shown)


.gitmodules
```diff
@@ -49,3 +49,6 @@
 	shallow = true
 	path = External/robin-map
 	url = https://github.com/Tessil/robin-map.git
+[submodule "External/Vulkan-Headers"]
```
skmp, line 52: How about `shallow = true` for sane checkout times?

Sonicadvance1: Made shallow.

neobrain (approved):

PR #1887, opened 2022-08-06
FEXServer: Fix unsafe vector insert/removal

Sonicadvance1:
Depending on the operation we will do a vector insert or removal while
iterating over the vector.

Fixes a use after free that asan found when insert caused the vector to
resize.


skmp (approved):
nice catch

PR #1895, opened 2022-08-09
Thunks: Make unknown glXGetProcAddress non-fatal

Sonicadvance1:
Multiple games will query symbols that are leaked but we don't support.
It is safe to return nullptr in these cases.
Print a warning message still just in-case someone fails hard at GL.

### DIFF
--- a/ThunkLibs/libGL/libGL_Guest.cpp
+++ b/ThunkLibs/libGL/libGL_Guest.cpp
@@ -43,26 +43,30 @@ const std::unordered_map<std::string_view, uintptr_t /* guest function address *
     });
 
 extern "C" {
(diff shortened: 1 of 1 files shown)


neobrain (approved):
Would it be reasonable to add an opt-in "return stub symbol that fatally errors on call" path here for debugging, like the one we have in Vulkan?

Looks fine either way.

Sonicadvance1:
I'm not too concerned about it. I've personally generated GL function loaders that just try and grab /every/ GL function regardless of failure. I know other people did this as well.
With the followup PR where we add support for all the remaining function pointers it becomes even less of an issue since GL isn't a fast moving target.

PR #1898, opened 2022-08-09
AppConfig: Fix bug with filename

Sonicadvance1:
Turns out cmake will cut the filename at the first period, not the last.
Fixes an issue if we have application names with multiple periods in it.

### DIFF
--- a/Data/AppConfig/CMakeLists.txt
+++ b/Data/AppConfig/CMakeLists.txt
@@ -11,15 +11,15 @@ endforeach()
 # First generate then install it
 foreach(GEN_CONFIG_SRC ${GEN_CONFIG_SOURCES})
   # Get the filename only component
-  get_filename_component(CONFIG_NAME ${GEN_CONFIG_SRC} NAME_WE)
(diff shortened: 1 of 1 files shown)


Data/AppConfig/CMakeLists.txt
```diff
@@ -11,15 +11,17 @@ endforeach()
 # First generate then install it
 foreach(GEN_CONFIG_SRC ${GEN_CONFIG_SOURCES})
   # Get the filename only component
-  get_filename_component(CONFIG_NAME ${GEN_CONFIG_SRC} NAME_WE)
```
neobrain, line 14: Judging from https://cmake.org/cmake/help/latest/command/get_filename_component.html, using `NAME_WLE` here is a simpler fix that allows us to keep using `get_filename_component`. Haven't tried it out myself though :)

Sonicadvance1: You're right. That works great.

skmp (approved):

PR #1901, opened 2022-08-09
CI: Build Thunks

Sonicadvance1:
To ensure that our thunks don't become horribly broken with our changes. Build them in CI.

Requires some tinkering because the 20.04 requirement, but otherwise works.

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -125,6 +125,10 @@ target_compile_definitions(libX11-guest-deps INTERFACE -DX11_VERSION_PATCH=${X11
 generate(libXext ${CMAKE_CURRENT_SOURCE_DIR}/../libXext/libXext_interface.cpp thunks function_packs function_packs_public)
 add_guest_lib(Xext)
 
+target_compile_definitions(libXext-guest-deps INTERFACE -DX11_VERSION_MAJOR=${X11_VERSION_MAJOR})
+target_compile_definitions(libXext-guest-deps INTERFACE -DX11_VERSION_MINOR=${X11_VERSION_MINOR})
+target_compile_definitions(libXext-guest-deps INTERFACE -DX11_VERSION_PATCH=${X11_VERSION_PATCH})
+
 generate(libXrender ${CMAKE_CURRENT_SOURCE_DIR}/../libXrender/libXrender_interface.cpp thunks function_packs function_packs_public)
 add_guest_lib(Xrender)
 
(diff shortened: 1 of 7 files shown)


ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -41,8 +42,10 @@ function(generate NAME SOURCE_FILE)
       DEPENDS "${GENERATOR_EXE}"
       DEPENDS "${SOURCE_FILE}"
       COMMAND "${GENERATOR_EXE}" "${SOURCE_FILE}" "${NAME}" "-${WHAT}" "${OUTFILE}" -- -std=c++17 -DGUEST_THUNK_LIBRARY
-            # Expand include directories to space-separated list of -isystem parameters
-           "$<$<BOOL:${prop}>:;-isystem$<JOIN:${prop},;-isystem>>"
+        # Expand compile definitions to space-separated list of -D parameters
```
skmp, line 45: Why is this needed?

ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -206,7 +206,10 @@ extern "C" {
 
     MakeHostFunctionGuestCallable(ret->resource_alloc);
     MakeHostFunctionGuestCallable(ret->idlist_alloc);
+#if ((X11_VERSION_MAJOR << 24) | (X11_VERSION_MINOR << 16) | (X11_VERSION_PATCH)) >= ((1 << 24) | (7 << 16) | (0))
```
skmp, line 209: (X11_VERSION_MAJOR >= 1 && X11_VERSION_MINOR >= 16  && X11_VERSION_PATCH >= 0) ?

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: compile definitions automatically get stripped of their `-D` in cmake, so we need to add them back in to use this interface.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: Resolved.

skmp (approved):

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ function(generate NAME SOURCE_FILE)
   # Shorthand for the include directories added after calling this function.
```
neobrain, line 28: Since we now can propagate definitions, we can remove the manual `-DGUEST_THUNK_LIBRARY` switch below and in `add_guest_lib` and instead use this:

```suggestion
  target_compile_definitions(${NAME}-guest-deps INTERFACE GUEST_THUNK_LIBRARY)
  # Shorthand for the include directories added after calling this function.
```

neobrain (approved):
Glad 20.04 doesn't require any major changes here. Thanks for wiring things up with CI!

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: Good shout, resolved.

PR #1891, opened 2022-08-08
Guest Build System: Add x86_32 toolchain, formalize things  bit, cleanup FEXLinuxTests

skmp:
#### Overview
- Formalized guest 32 bit x86 as "x86_32" and guest 64 bit x86 as "x86_64" in the toolchains / build system
- Added x86_32 toolchain file
- Moved toolchain files to Toolchains/
- Main CMakeLists stores toolchain files in `TOOLCHAIN_FILE_${GuestArch}` vars
- Modified FEXLinuxTests to use x86_32 toolchain instead of adding -m32 manually
- Added support to FEXLinuxTests to only compile some architectures via `// arch: arch1, arch2`, remove the 32 bit only configuration

#### Notes
This might need some variations on the installed packages which I haven't fully identified, see #1889

PR #1902, opened 2022-08-09
Thunks: Update to use the Guest Build System

skmp:
#### Overview

- Compiles thunks using the toolchain files created for FLT in #1891
- Adds support for per-guest arch thunks in the build system, with no actual support in the thunks themselves at this point (inspired/iterated from #1828)
- Cleanups the thunks cmake files, with the future intention of sharing the cmake code between guest and host.

Builds on top of #1891.


skmp:
This has cryptic build issues with the unreviewed code introduced in #1868.

skmp:
Alright.

Converted thunkgen to a `clang++` plugin, instead of a custom tool, so that it picks up the headers correctly.

However,
- clang++ crashes in x86 and 8.0 runners. That's most likely me using it incorrectly.
- clang++ does not recognise the -fplugin-arg-XXX arguments in 8.4 (ubu20.04). This is peculiar as it does recognise the -fplugin one. I'll try with lower case.

Finally,
8.2 runner works.
Invocation times are also somehow lower now for generating, for some reason.

neobrain:
Once you're done fixing the remaining issues you mentioned, could you do a rebase pass over this to split the PR into more easily digestible patches for review?

skmp:
@neobrain yes, def. This is rebased on top of #1891 (and needs that revived and merged before it is considered)

skmp:
I cannot reproduce the armv8.4 or x64 errors in my setups. Not sure what is happening here.

skmp:
Needed packages: `sudo apt install libc6-dev-i386-cross libc6-dev-i386-cross`

Sonicadvance1:
I'm assuming the duplicate `libc6-dev-i386-cross` in that command was actually `libc6-dev-amd64-cross`.
All runners already had both packages installed.

skmp:
So uh

```
/usr/lib/llvm-14/bin/clang --target=i386-linux-gnu --sysroot=/lib/i686-linux-gnu --sysroot=/usr/lib/i686-linux-gnu  -fuse-ld=lld -L/usr/i686-linux-gnu/lib -L/usr/lib/i686-linux-gnu test.c -print-search-dirs

programs: =/usr/lib/llvm-14/bin
libraries: =/usr/lib/llvm-14/lib/clang/14.0.0
```

and

```
clang --target=i386-linux-gnu --sysroot=/lib/i686-linux-gnu --sysroot=/usr/lib/i686-linux-gnu  -fuse-ld=lld -L/usr/i686-linux-gnu/lib -L/usr/lib/i686-linux-gnu test.c -print-search-dirs

programs: =/usr/bin:/usr/lib/llvm-14/bin:/usr/bin/../lib/gcc-cross/i686-linux-gnu/11/../../../../i686-linux-gnu/bin
libraries: =/usr/lib/llvm-14/lib/clang/14.0.0:/usr/bin/../lib/gcc-cross/i686-linux-gnu/11:/usr/bin/../lib/gcc-cross/i686-linux-gnu/11/../../../../i686-linux-gnu/lib
```

and

```
skmp@mangie:~/projects/FEX/build$ which clang
/usr/bin/clang
skmp@mangie:~/projects/FEX/build$ ls -lah /usr/bin/clang
lrwxrwxrwx 1 root root 24 Απρ   1 23:51 /usr/bin/clang -> ../lib/llvm-14/bin/clang
```

skmp:
Looks like the same executable of clang behaves differently when in `/usr/bin/clang` and when in `/usr/lib/llvm-14/bin/clang`

skmp:
@Sonicadvance1 can you check that `x64` and `ARMv8.2` are using the same libclang as they are using for clang++?

Looks like some progress was made also on  ARMv8.4, now just -ldl missing from somewhere

PR #1915, opened 2022-08-12
Linux: Emulate classic getdents syscall for x64 and x32

Sonicadvance1:
Arm64 doesn't have the classic getdents syscall, only getdents64.
Old glibc versions (like 2.17) don't support getdents64

This fixes 64-bit `ls` with a centos 7 rootfs. Likely also fixes some other very
old applications.

Packing differences between 32-bit and 64-bit getdents means we need to
template this between the two types, otherwise it's quite similar.

No known applications rely on the 32-bit getdents but worked with test
applications.

### DIFF
--- a/Source/Tests/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tests/LinuxSyscalls/Syscalls.cpp
@@ -14,6 +14,8 @@ $end_info$
 #include "Tests/LinuxSyscalls/Syscalls/Thread.h"
 #include "Tests/LinuxSyscalls/x32/Syscalls.h"
 #include "Tests/LinuxSyscalls/x64/Syscalls.h"
+#include "Tests/LinuxSyscalls/x32/Types.h"
+#include "Tests/LinuxSyscalls/x64/Types.h"
 
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Core/Context.h>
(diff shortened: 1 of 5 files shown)


Sonicadvance1:
Rebased to ensure thunks are tested since that was merged. The d_type flag is quite magical with its emulation. It's not specifically after the null of the file name, but at `reclen - 1` Which due to alignment requirements it is actually a couple of bytes after null character in most instances. Very quirky.
```
--------------- nread=1160 ---------------
i-node#  file type  d_reclen  d_off   d_name
     138  directory    32 4294967296  .
       2  directory    32 1387312459340357341  ..
 1835069  directory   120 1516841174588810543  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-power-profiles-daemon.service-hIuabe
      54  regular      40 1707541011482174024  .X0-lock
  393332  directory    40 1808774146592706589  .X11-unix
    1352  regular      48 2073973872570234946  perf-2666.map
  393490  directory   112 2215090756752148303  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-systemd-timesyncd.service-8TZ6j1
  393436  directory    40 3533004142230679679  .Test-unix
     950  regular      48 3864375062407340690  perf-2661.map
 4332352  directory    56 4150988307654459019  tracker-extract-3-files.10000�ͥ
 1835081  directory   112 4232109714883896875  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-systemd-resolved.service-5LlO9W
    1784  FIFO         40 4470017937966285707  TestFifo
     252  regular      48 4642001283375626037  perf-2644.map
  393388  directory    40 5129003486921883598  .ICE-unix
  393422  directory    40 6292723327416644901  .font-unix
  393416  directory    40 6770045842560868372  .XIM-unix
      84  socket       56 7451147476822483138  1000.FEXServer.socket
 1835073  directory   112 7923020711278978384  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-systemd-logind.service-s6qmuA
 1835090  directory   104 8620148916371485190  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-bluetooth.service-DcwR1M
--------------- nread=152 ---------------
i-node#  file type  d_reclen  d_off   d_name
      59  regular      48 9026098801890872797  config-err-ralBZ6
 1835659  directory   104 9223372036854775807  systemd-private-40483cf04c9e4014b9fe2d5aa00532a9-upower.service-S1wHCM
 ```

skmp (approved):

PR #1920, opened 2022-08-15
unittests: Support skipping unit tests based on host feature support

Sonicadvance1:
For these unit tests we no longer need to put them in the disabled tests
file. Instead it will be skipped if the host doesn't support the feature
required.

Theoretically this will solve @skmp's problem of running unit tests on their old x86 hardware.


skmp (approved):
Nice!

One extra nicey would be, to automatically set the `HostFeatures` flags if the tests are in folder named "AVX", "3DNOW" or such :3

Sonicadvance1:
Considering how slowly these tests evolve, I'm not too concerned about automatically generating a hidden HostFeatures flag.

PR #1926, opened 2022-08-18
IRLoader/TestHarnessLoader: Don't build if not building tests

Sonicadvance1:
If we're not building tests then just don't build the IRLoader since it
won't get used.

### DIFF
--- a/Source/Tests/CMakeLists.txt
+++ b/Source/Tests/CMakeLists.txt
@@ -164,18 +164,20 @@ install(TARGETS FEXBash
     COMPONENT runtime
 )
 
-add_executable(TestHarnessRunner TestHarnessRunner/HostRunner.cpp TestHarnessRunner.cpp)
-target_include_directories(TestHarnessRunner
-  PRIVATE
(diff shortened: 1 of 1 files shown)


skmp:
How about using the `EXCLUDE_FROM_ALL` functionality to make the Loaders not built by default, and to bring them in the build only if some other dep (eg, a unit test) needs them?

Source/Tests/CMakeLists.txt
```diff
@@ -164,18 +164,20 @@ install(TARGETS FEXBash
...
+      ${CMAKE_BINARY_DIR}/generated
+  )
+  target_link_libraries(TestHarnessRunner
+    PRIVATE
+      ${LIBS}
+      LinuxEmulation
+      ${PTHREAD_LIB}
+  )
 
 # add_executable(UnitTestGenerator UnitTestGenerator.cpp)
```
neobrain, line 182: This now has misleading indentation.

neobrain (approved):

PR #1927, opened 2022-08-18
FDUtils: Don't make unknown get_fdpath fatal

Sonicadvance1:
Encountered this while running wine things.
Non-fatal so don't explode

### DIFF
--- a/Source/Tests/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tests/LinuxSyscalls/FileManagement.cpp
@@ -575,7 +575,7 @@ uint64_t FileManager::Readlinkat(int dirfd, const char *pathname, char *buf, siz
         dirfd != AT_FDCWD) {
     // Passed in a dirfd that isn't magic FDCWD
     // We need to get the path from the fd now
-    Path = FEX::get_fdpath(dirfd);
+    Path = FEX::get_fdpath(dirfd).value_or("");
 
     if (pathname) {
       if (!Path.empty()) {
--- a/Source/Common/FDUtils.h
+++ b/Source/Common/FDUtils.h
@@ -5,12 +5,13 @@
 #include <fcntl.h>
 #include <filesystem>
 #include <linux/limits.h>
+#include <optional>
 #include <unistd.h>
 
 namespace FEX {
 [[maybe_unused]]
 static
-std::string get_fdpath(int fd) {
+std::optional<std::string> get_fdpath(int fd) {
   char SymlinkPath[PATH_MAX];
   std::filesystem::path Path = std::filesystem::path("/proc/self/fd") / std::to_string(fd);
   int Result = readlinkat(AT_FDCWD, Path.c_str(), SymlinkPath, sizeof(SymlinkPath));
@@ -18,8 +19,8 @@ std::string get_fdpath(int fd) {
     return std::string(SymlinkPath, Result);
   }
 
-  LOGMAN_MSG_A_FMT("Couldn't get symlink from /proc/self/fd/{}", fd);
-  return {};
+  // Not fatal if an FD doesn't point to a file
+  return std::nullopt;
 }
 
 }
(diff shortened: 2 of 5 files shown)


skmp:
Several places in the codebase depend on being able to always get a string here. How about we convert this to `std::optional<std::string>` so we catch things correctly?

Sonicadvance1:
Made it `std::optional<std::string>` and made the few places that use it ensure they are safely handling the case no value returns.

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -575,10 +575,14 @@ uint64_t FileManager::Readlinkat(int dirfd, const char *pathname, char *buf, siz
         dirfd != AT_FDCWD) {
     // Passed in a dirfd that isn't magic FDCWD
     // We need to get the path from the fd now
-    Path = FEX::get_fdpath(dirfd);
+    auto FDPath = FEX::get_fdpath(dirfd);
+
+    if (FDPath.has_value()) {
+      Path = FDPath.value();
+    }
```
neobrain, line 582: FYI a nice way to simplify this is:
```suggestion
    auto Path = FEX::get_fdpath(dirfd).value_or("");
```

Not terribly important, just seemed worth mentioning :)

Sonicadvance1: Guess that depends if we want the rest of the baggage with it, doesn't matter to me either way.
```diff
diff --git a/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp b/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
index 1c053ac8f..a5916e43a 100644
--- a/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
+++ b/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
@@ -714,14 +714,10 @@ namespace FEX::EmulatedFile {
         dirfs != AT_FDCWD) {
       // Passed in a dirfd that isn't magic FDCWD
       // We need to get the path from the fd now
-      auto FDPath = FEX::get_fdpath(dirfs);
-
-      if (FDPath.has_value()) {
-        Path = FDPath.value();
-      }
+      Path = FEX::get_fdpath(dirfs).value_or("");

       if (pathname) {
-        if (FDPath.has_value()) {
+        if (!Path.empty()) {
           // If the path returned empty then we don't need a separator
           Path += "/";
         }
diff --git a/Source/Tests/LinuxSyscalls/FileManagement.cpp b/Source/Tests/LinuxSyscalls/FileManagement.cpp
index ce1bd7675..9b101fd23 100644
--- a/Source/Tests/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tests/LinuxSyscalls/FileManagement.cpp
@@ -575,14 +575,10 @@ uint64_t FileManager::Readlinkat(int dirfd, const char *pathname, char *buf, siz
         dirfd != AT_FDCWD) {
     // Passed in a dirfd that isn't magic FDCWD
     // We need to get the path from the fd now
-    auto FDPath = FEX::get_fdpath(dirfd);
-
-    if (FDPath.has_value()) {
-      Path = FDPath.value();
-    }
+    Path = FEX::get_fdpath(dirfd).value_or("");

     if (pathname) {
-      if (FDPath.has_value()) {
+      if (!Path.emtpy()) {
         // If the path returned empty then we don't need a separator
         Path += "/";
       }
       ```

neobrain: Oh, I hadn't noticed we were actually using the `optional` after. Replacing the typesystem-level `has_value()` check with a value-level `empty()` isn't ideal, but I think I prefer it over managing the same state using two separate variables.

Sonicadvance1: `I prefer it` Which one do you prefer? Original in PR or proposed diff in my previous comment? Ambiguous.

neobrain: Apologies. I prefer the `value_or` approach, as it makes for shorter code overall and eliminates the need to have a dedicated variable for the `optional`.

Sonicadvance1: blammo. Swapped over.

neobrain (approved):

skmp:
:+1:

PR #1928, opened 2022-08-18
Thunks/X11: Adds missing XLibint functions

Sonicadvance1:
Some of these were required to get thunking to work with Proton and
DXVK.

### DIFF
--- a/ThunkLibs/libX11/libX11_interface.cpp
+++ b/ThunkLibs/libX11/libX11_interface.cpp
@@ -636,3 +636,53 @@ template<> struct fex_gen_config<_XInitImageFuncPtrs> : fexgen::custom_guest_ent
 template<> struct fex_gen_config<XVaCreateNestedList> {
     using uniform_va_type = void*;
 };
+
+// Xlibint
+template<> struct fex_gen_config<_XError> {};
+template<> struct fex_gen_config<_XIOError> {};
+template<> struct fex_gen_config<_XGetHostname> {};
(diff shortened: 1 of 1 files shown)


ThunkLibs/libX11/libX11_interface.cpp
```diff
@@ -636,3 +636,48 @@ template<> struct fex_gen_config<_XInitImageFuncPtrs> : fexgen::custom_guest_ent
...
+template<> struct fex_gen_config<_XUnknownCopyEventCookie> {};
+template<> struct fex_gen_config<_XWireToEvent> {};
+template<> struct fex_gen_config<_XDefaultWireError> {};
+template<> struct fex_gen_config<_XPollfdCacheInit> {};
+template<> struct fex_gen_config<_XPollfdCacheAdd> {};
+template<> struct fex_gen_config<_XPollfdCacheDel> {};
+template<> struct fex_gen_config<_XAllocID> {};
+template<> struct fex_gen_config<_XAllocIDs> {};
+template<> struct fex_gen_config<_XFreeExtData> {};
+// XXX: Needs investigation in to this API
```
neobrain, line 661: What kind of investigation is needed that wasn't needed for the other functions?

neobrain (approved):

ThunkLibs/libX11/libX11_interface.cpp
Sonicadvance1: Oops, that was just a note to myself. Just needed to check that it matched the other `XESet*` functions and they do.

PR #1929, opened 2022-08-18
Thunks/X11: Support Variadic stack packing

Sonicadvance1:
Found an issue with wine + DXVK + thunks where these were passing in
more than 7 arguments and crashing.

Create some assembly to support any size of variadic stack packing.
Only implemented for AArch64 for now.

### DIFF
--- a/ThunkLibs/libX11/libX11_Host.cpp
+++ b/ThunkLibs/libX11/libX11_Host.cpp
@@ -25,101 +25,299 @@ $end_info$
 #include <utility>
 
 #include "ldr_ptrs.inl"
+#ifdef _M_ARM_64
+// This Variadic asm only works for one signature
+// ({uint32_t,uint64_t} a_0, size_t count, uint64_t *list)
+//
+// Variadic ABI for AArch64 (flat uint64_t):
+// Arguments 0-7 is in registers
+// 8+ stored on to stack
+//
+// The X11 functions we are calling need an additional nullptr passed in.
+// nullptr will be at the end of the list of generated stack items when called through this.
+// We will always generate a variadic frame of `count` objects + 1 for nullptr.
+//
+// Incoming:
+// x0 = XIM
+// x1 = count
+// x2 = array of 64-bit values
+// x3 = Function to call
+//
+// Outgoing:
+// x0: Ptr
+
+__attribute__((naked))
+void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void *Func) {
+  asm volatile(R"(
+    # Move our function to x8, which will be unused
+    mov x8, x3
+
+    # Move our list to x9, which will be unused
+    mov x9, x2
+
+    # >6 means use stack callback
+    cmp x1, 6
+    b.gt .stack%=
+
+    # Setup a jump table
+    adr x10, .zero%=
+    adr x11, .jump_table%=
+    ldrb w11, [x11, x1]
+    add x10, x10, x11, lsl 2
+    br x10
+
+    .zero%=:
+      mov x1, 0
+      br x8
+
+    .one%=:
+      ldr x1, [x9, 0]
+      mov x2, #0
+      br x8
+
+    .two%=:
+      ldp x1, x2, [x9, 0]
+      mov x3, 0
+      br x8
+
+    .three%=:
+      ldp x1, x2, [x9, 0]
+      ldr x3, [x9, 16]
+      mov x4, 0
+      br x8
+
+    .four%=:
+      ldp x1, x2, [x9, 0]
+      ldp x3, x4, [x9, 16]
+      mov x5, 0
+      br x8
+
+    .five%=:
+      ldp x1, x2, [x9, 0]
+      ldp x3, x4, [x9, 16]
+      ldr x5, [x9, 32]
+      mov x6, 0
+      br x8
+
+    .six%=:
+      ldp x1, x2, [x9, 0]
+      ldp x3, x4, [x9, 16]
+      ldp x5, x6, [x9, 32]
+      mov x7, 0
+      br x8
+
+    .stack%=:
+      # Store LR and x28
+      stp x28, x30, [sp, -16]!
+
+      # x8 = <arg ptr>
+      # x0 = <arg im>
+      # x1 = <count>
+      # x9 = <list ptr>
+
+      # Stack objects
+      # Count >= 7
+      # Subtract 6 count objects
+      # Leaves us at least 1 (nullptr)
+      sub x1, x1, 6
+
+      # Round up to the nearest pair
+      and x10, x1, 1
+      add x10, x10, x1
+
+      # Multiply by eight to get the size of stack we need to create
+      lsl x10, x10, 3
+
+      # Allocate stack space
+      sub sp, sp, x10
+
+      # Store how much data we added to the stack in our callee saved register we stole
+      mov x28, x10
+
+      # Subtract one member due to nullptr ender
+      sub x10, x1, 1
+
+      # x11 - stack offset
+      mov x11, sp
+
+      # x12 - load offset
+      add x12, x9, (7 * 8)
+
+      cmp x10, 1
+      b.eq .single%=
+      b.lt .no_single%=
+
+      .load_pair%=:
+      ldp x1, x2, [x12], 16
+      stp x1, x2, [x11], 16
+      sub x10, x10, 8
+      cmp x10, 1
+      b.gt .load_pair%=
+      b.lt .no_single%=
+
+      # One variable at most
+      .single%=:
+      ldr x1, [x12]
+      stp x1, xzr, [x11]
+      b .top_reg_args%=
+
+      .no_single%=:
+      # Need to store nullptr
+      str xzr, [x11]
+
+      .top_reg_args%=:
+      ldp x1, x2, [x9, 0]
+      ldp x3, x4, [x9, 16]
+      ldp x5, x6, [x9, 32]
+      ldr x7, [x9, 48]
+
+      # Stack is setup going in to this
+      blr x8
+
+      # Move stack back
+      add sp, sp, x28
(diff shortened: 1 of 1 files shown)


ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -25,103 +25,301 @@ desc: Handles callbacks and varargs
...
+
+    # >6 means use stack callback
+    cmp x2, 6
+    b.gt .stack
+
+    # Setup a jump table
+    adr x10, .zero
+    adr x11, .jump_table
+    ldrb w11, [x11, x2]
+    add x10, x10, x11, lsl 2
```
neobrain, line 64:
```suggestion
    adr x11, .jump_table
    ldrb w11, [x11, x2]
    mov x10, x11, lsl 2
```

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -25,103 +25,301 @@ desc: Handles callbacks and varargs
 #include <utility>
 
 #include "ldr_ptrs.inl"
+#ifdef _M_ARM_64
+__attribute__((naked))
+void* InternalVariadic(void* Ptr, uint64_t a_0, size_t count, void** list) {
+  // Arguments 0-7 is in registers
+  // 8+ stored on to stack
+  // 0 need to be on the end of the list
```
neobrain, line 33: Does this refer to the input list or to the output arguments? Wording is unclear; I thought it's the input list at first, but looks like you mean the outputs.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -25,103 +25,301 @@ desc: Handles callbacks and varargs
 #include <utility>
 
 #include "ldr_ptrs.inl"
+#ifdef _M_ARM_64
```
neobrain, line 28: Are you planning to implement a proper x86 code path?

Maybe we can just implement a generic (if less performant) fallback template for non-ARM architectures, with the same logic as before but more cases.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -25,103 +25,301 @@ desc: Handles callbacks and varargs
...
+  // Incoming:
+  // x0 = Function
+  // x1 = XIM
+  // x2 = count
+  // x3 = stack array of 64-bit values
+  //
+  // Outgoing:
+  // x0: Ptr
+  __asm volatile(R"(
+    # Store LR and frame pointer if it exists
```
neobrain, line 44: Does this produce significantly better results than a switch case where only the `count > 6` case is implemented as inline-assembly?

neobrain (approved):
Looks good, though if possible I'd prefer using plain C code for the "simple" bits of the helper function.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -25,103 +25,301 @@ desc: Handles callbacks and varargs
 #include <utility>
 
 #include "ldr_ptrs.inl"
+#ifdef _M_ARM_64
+__attribute__((naked))
+void* InternalVariadic(void* Ptr, uint64_t a_0, size_t count, void** list) {
+  // Arguments 0-7 is in registers
```
skmp, line 31: There's no guarantee that they are stored in integer registers, fp arguments are stored in fp registers, in a way you can't parse them without knowing the argument order and count

skmp: printf("a", 1, 1.0, 2, 2.0) ends up putting
x0 = "a"
x1 = 1
x2 = 2
fp0=1.0
fp1 = 2.0

Sonicadvance1: This isn't meant to be completely generic to all variadic packing packing. I'll rename it to make that more clear that it is just for flat uint64_t

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: Interestingly, depending on the branch, the assembly is more or less efficient depending.

C code is generally more efficient for the 1-6 case but only because it does a tail call optimization, adding that to the assembly is easy.
The variadic call ends up poisoning the rest of the function requiring the stack to be adjusted for the entire thing.

I'll shuffle things around so this gets inlined for the full asm to remove the jump, which will make the inline asm more optimal overall.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: I'm not sure what you're suggesting here.
We use `adr x10, .zero` to load the base offset address of the jump targets.
Then we index the jump table with `ldrb w11, [x11, x2]`, only needing a byte here since the offsets are so small.
We then add that offset from the jump table to the base of the jump targets, multiplying by two since we compacted with `add <destination reg>, <jump target base>, <jump target offset>, lsl 2`

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: Updated comment to be more clear.
```
// nullptr needs to be at the end of the list of generated stack items for this API.
// We will always generate a variadic frame of `count` objects + 1 for nullptr.
```

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: I don't have the heart at the moment to write an x86 specific code path at the moment for this. For now it hits the C path and I added the 8 case variadic version which would be required.
Main purpose for the ARM side is that it now covers the variadic path regardless of number of arguments passed in rather than any sort of performance optimization. Slightly elevated performance is a by-product but not the reason its ASM.

Sonicadvance1:
> Looks good, though if possible I'd prefer using plain C code for the "simple" bits of the helper function.

It ends up being nicer on the codegen weirdly enough when it is all in inline asm.
Main thing is I removed the trampoline so it all gets inlined now which makes it quite a bit nicer.

neobrain:
> It ends up being nicer on the codegen weirdly enough when it is all in inline asm.
> Main thing is I removed the trampoline so it all gets inlined now which makes it quite a bit nicer.

Is it a significant difference? Adding the `naked` attribute to all these functions and including a stateful inline header seems like a level of voodoo that needs some justification when the same could be implemented with an isolated helper function. Considering we already have to do the manual guest-side packing, I'm not sure one extra direct function call makes a big dent.

Thanks for trying out different approaches here though!

ThunkLibs/libX11/libX11_Host.cpp
neobrain: Ah, bummer this doesn't work out in practice.

ThunkLibs/libX11/libX11_Host.cpp
neobrain: Sounds fair. If the final version of this PR moves back to using a InternalVariadic helper function, would be nice to adapt the non-arm bits to also use this function (implemented in C for `count<8` on those platforms), so the host endpoints don't need to be specialized per-arch.

ThunkLibs/libX11/libX11_Host.cpp
neobrain: The initialization of x10 (`adr x10, .zero`) and the subsequent `add x10, x10, x11, lsl 2` can be combined into `mov x10, x11, lsl 2`.

In terms of equivalent C code, instead of doing `x10 = 0; x10 = x10 + x11 << 2`, we'd just be doing `x10 = x11 << 2`.

Sonicadvance1: `.zero` is a relative address label called `.zero` in the following asm. Not constant zero.

Sonicadvance1: so `adr x10, .zero` is effectively going to set x10 to `PC + <N number of instructions>`

neobrain: Derp, you're right of course. Appreciate your patience on this one 🙈

ThunkLibs/libX11/libX11_Host.cpp
neobrain: Unfortunately I find this even more confusing, now that this is just a blob of inline-assembly without the context that was previously provided by `void* InternalVariadic(void* Ptr, uint64_t a_0, size_t count, void** list)`.

I'll try to put this into more concrete terms: "This API" isn't unclear about whether it refers to the called function or the blob provided by VariadicImplementation.inl (which isn't really an "interface" at this point). This in turn allows for reading "generated stack items" in two different ways (the array generated on the guest-side or the stack contents generated here?). At the end of the comment I'm left thinking "Did you just describe a requirement for using this blob or an implementation detail?".

Hope that helps explain where my confusion is coming from. I can see us using this feature more widely in other libraries in the future, so having clear phrasing would be useful to have.

Sonicadvance1: Updated the comment more to explain that the functions we are calling are expecting nullptr to be at the end of the stack frame, not the stack frame coming in to the function.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: Moved back to a function call for the default switch statement case instead. We can take the minor perf hit here for now.

neobrain (approved):
The cases `.zero` through `.six` in the new helper function seem unused now, though arguably they'll be useful to have around later. Looks fine either way!

ThunkLibs/libX11/libX11_Host.cpp
neobrain: Much better with the updated comment and with the old context around again. Thanks!

Sonicadvance1:
Indeed, useful to keep around for having the asm handle all the cases, and if we every want to go all-in for some game really abusing these.

PR #1885, opened 2022-08-05
Allocator: Simplify StealMemory, make it less chatty with kernel space

skmp:
#### Overview

This parses `/proc/self/maps` with a very tiny FSM, and allocates all unmapped regions.

#### Performance
This makes `/bin/true` run in < 10-15 msecs from 38-43 with #1842

#### Follow ups
- #1914

PR #1932, opened 2022-08-20
64BitAllocator: Fixes a significant state tracking perf problem

Sonicadvance1:
Due to how the 64-bit allocator previously worked, it was never subjected to memory
regions larger than 64GB to be tracked. With the change in
PR #1885, this has changed to have regions that will hit sizes larger
than 170TB on some platforms.

Better yet, even with smaller regions it still had a performance issue,
it just wasn't as visible.

First problem: We used MemSet instead of MemClear for the live page
clearing. This caused pages to be claimed as "always in use".

This would cause us to always scan the entire region on allocation, find
that it didn't work and allocate a fresh region on every slab
allocation.
jemalloc saving us here since it allocates slabs from the OS fairly
aggressively.

Second problem: We used MemSet (now changed to MemClear) to "clear" the
state tracking for pages.

This causes ~600MB of memory to be used purely for state tracking.
This was physically backed since we were writing to every bit of
tracking for handling 256TB of VA.
This had a fault dance with the kernel for every new page being hit
here.
Instead of clearing the the bits with a memset, clear it with madvise so
it doesn't consume physical pages at all.

This means we use significantly less physical memory for 32-bit
applications.

With this change, pressure-vessel startup time goes from 24 seconds down
to 17 seconds. 70% of the original startup time.
But really the main savings here comes from the memory reduction that
PR #1885 ballooned, but has been an unseen problem before.

Before that PR we were burning 2MB of physical memory per region for no
reason.
After that PR we were burning up to 600MB of physical memory per region
for no reason. Changing a bit depending on how large the region ended up
being.

This now ends up being 2 pages starting out and grows as more pages are
are used. A significant improvement.

### DIFF
--- a/External/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
+++ b/External/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
@@ -70,7 +70,10 @@ namespace Alloc::OSAllocator {
         ReservedVMARegion *SlabInfo;
         uint64_t FreeSpace{};
         uint32_t LastPageAllocation{};
-        FEXCore::FlexBitSet<uint64_t> UsedPages;
+
+        // Align UsedPages so it pads to the next page.
+        // Necessary to take advantage of madvise zero page pooling.
+        alignas(4096) FEXCore::FlexBitSet<uint64_t> UsedPages;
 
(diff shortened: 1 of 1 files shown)


skmp (approved):

skmp:
nice

PR #1941, opened 2022-08-23
Thunks: Adds VDSO thunk library 

Sonicadvance1:
VDSO is heavily abused by Proton games to the point it is showing up as
CPU time.
Implement a guest-facing only thunk library using the hardcoded VDSO
interface in Thunks.

If available this will always be loaded on application load and set the
auxv value to support it.

This requires a bit of special treatment as our first user of linker
scripts since the format of the ELF must be careful crafted to not break
applications trying to parse it.

This library exposes a handful of symbols:
- clock_gettime
- clock_getres
- gettimeofday
- time
- getcpu
- All previous with `__vdso_` prefix
- LINUX_2.6

All of these symbols get routed directly to the host architecture VDSO
interface if they exist.
AArch64 doesn't have getcpu or time VDSO.

In a microbench, VDSO improved bench times substantially
x86-64 host: 3.612s -> 1.369s - 2.63x speed
AArch64 host: 3.821s -> 2.284s - 1.67x speed
  - AArch64 isn't as improved due to missing VDSO symbols

This is also our first /always/ enabled thunk as long as the file exists

### DIFF
--- a/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
+++ b/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
@@ -17,10 +17,12 @@ $end_info$
 #include <cstdint>
 #include <dlfcn.h>
 
+#include <elf.h>
 #include <Interface/Context/Context.h>
 #include "FEXCore/Core/X86Enums.h"
 #include <malloc.h>
 #include <mutex>
+#include <sys/auxv.h>
 #include <unordered_map>
 #include <memory>
 #include <shared_mutex>
@@ -67,6 +69,107 @@ extern char __start_HostToGuestTrampolineTemplate[];
 extern char __stop_HostToGuestTrampolineTemplate[];
 
 namespace FEXCore {
+  namespace VDSO {
+    using TimeType = decltype(::time)*;
+    using GetTimeOfDayType = decltype(::gettimeofday)*;
+    using ClockGetTimeType = decltype(::clock_gettime)*;
+    using ClockGetResType = decltype(::clock_getres)*;
+    using GetCPUType = decltype(::getcpu)*;
+
+    TimeType TimePtr = ::time;
+    GetTimeOfDayType GetTimeOfDayPtr = ::gettimeofday;
+    ClockGetTimeType ClockGetTimePtr = ::clock_gettime;
+    ClockGetResType ClockGetResPtr = ::clock_getres;
+    GetCPUType GetCPUPtr = ::getcpu;
+
+    static void time(void* ArgsRV) {
+      struct ArgsRV_t {
+        time_t *a_0;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = TimePtr(args->a_0);
+    }
+
+    static void gettimeofday(void* ArgsRV) {
+      struct ArgsRV_t {
+        struct timeval *tv;
+        struct timezone *tz;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = GetTimeOfDayPtr(args->tv, args->tz);
+    }
+
+    static void clock_gettime(void* ArgsRV) {
+      struct ArgsRV_t {
+        clockid_t clk_id;
+        struct timespec *tp;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = ClockGetTimePtr(args->clk_id, args->tp);
+    }
+
+    static void clock_getres(void* ArgsRV) {
+      struct ArgsRV_t {
+        clockid_t clk_id;
+        struct timespec *tp;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = ClockGetResPtr(args->clk_id, args->tp);
+    }
+
+    static void getcpu(void* ArgsRV) {
+      struct ArgsRV_t {
+        uint32_t *cpu;
+        uint32_t *node;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = GetCPUPtr(args->cpu, args->node);
+    }
+
+    void LoadVDSO() {
+      void *vdso = dlopen("linux-vdso.so.1", RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
+      if (!vdso) {
+        vdso = dlopen("linux-gate.so.1", RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
+      }
+
+      if (!vdso) {
+        // Couldn't load VDSO
+        return;
+      }
+
+      auto SymbolPtr = dlsym(vdso, "__vdso_time");
+      if (SymbolPtr) {
+        TimePtr = reinterpret_cast<TimeType>(SymbolPtr);
+      }
+
+      SymbolPtr = dlsym(vdso, "__vdso_gettimeofday");
+      if (SymbolPtr) {
+        GetTimeOfDayPtr = reinterpret_cast<GetTimeOfDayType>(SymbolPtr);
+      }
+
+      SymbolPtr = dlsym(vdso, "__vdso_clock_gettime");
+      if (SymbolPtr) {
+        ClockGetTimePtr = reinterpret_cast<ClockGetTimeType>(SymbolPtr);
+      }
+
+      SymbolPtr = dlsym(vdso, "__vdso_clock_getres");
+      if (SymbolPtr) {
+        ClockGetResPtr = reinterpret_cast<ClockGetResType>(SymbolPtr);
+      }
+
+      SymbolPtr = dlsym(vdso, "__vdso_getcpu");
+      if (SymbolPtr) {
+        GetCPUPtr = reinterpret_cast<GetCPUType>(SymbolPtr);
+      }
+      dlclose(vdso);
+    }
+  }
+
     struct ExportEntry { uint8_t *sha256; ThunkedFunction* Fn; };
 
     struct TrampolineInstanceInfo {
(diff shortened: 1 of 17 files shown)


ThunkLibs/libVDSO/libVDSO_Guest.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "function_packs.inl"
+
+extern "C" {
+time_t __vdso_time(time_t *tloc) __attribute__((alias("fexfn_pack_time")));
+int __vdso_gettimeofday(struct timeval *tv, struct timezone *tz) __attribute__((alias("fexfn_pack_gettimeofday")));
+int __vdso_clock_gettime(clockid_t, struct timespec *) __attribute__((alias("fexfn_pack_clock_gettime")));
+int __vdso_clock_getres(clockid_t, struct timespec *) __attribute__((alias("fexfn_pack_clock_getres")));
+int __vdso_getcpu(uint32_t *, uint32_t *) __attribute__((alias("fexfn_pack_getcpu")));
+
+__attribute__((alias("fexfn_pack_time"))) auto time(time_t * a_0) -> time_t;
```
neobrain, line 34: Why can't these use the generated `function_packs_public.inl` instead?

Sonicadvance1: You're right, resolved.

Source/Tests/ELFCodeLoader2.h
```diff
@@ -403,6 +406,28 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
       Entrypoint = MainElfEntrypoint;
     }
 
+    if (Is64BitMode()) {
+      // Load VDSO if we can
+      FEX_CONFIG_OPT(ThunkGuestLibs, THUNKGUESTLIBS);
```
skmp, line 411: This should be pass in, not read through the config directly

skmp: In fact, I think it would be best to pass in (VDSOFD/VDSOSize) directly, or load the VDSO before elf loading and pass the base pointer (though that might lead to address conflicts I guess?)

I don't think ELFLoader and thunks need to be coupled

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -67,6 +69,107 @@ extern char __start_HostToGuestTrampolineTemplate[];
 extern char __stop_HostToGuestTrampolineTemplate[];
 
 namespace FEXCore {
+  namespace VDSO {
```
skmp, line 72: This should be moved to some VDSO file, and should live in the linux frontend, as it is specific to that

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -139,7 +242,33 @@ namespace FEXCore {
                 // sha256(fex:allocate_host_trampoline_for_guest_function)
                 { 0x9b, 0xb2, 0xf4, 0xb4, 0x83, 0x7d, 0x28, 0x93, 0x40, 0xcb, 0xf4, 0x7a, 0x0b, 0x47, 0x85, 0x87, 0xf9, 0xbc, 0xb5, 0x27, 0xca, 0xa6, 0x93, 0xa5, 0xc0, 0x73, 0x27, 0x24, 0xae, 0xc8, 0xb8, 0x5a },
                 &AllocateHostTrampolineForGuestFunction
-            }
+            },
+            // VDSO
+            {
+                // sha256(libVDSO:time)
```
neobrain, line 248: Same question here - aren't those provided by `thunks.inl`?

(Apologies if this seems picky; the reason I'm asking is that I plan on unifying the generated files in the future, making inclusion of them non-optional)

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -75,6 +79,10 @@ function(add_guest_lib NAME)
   ## Make signed overflow well defined 2's complement overflow
   target_compile_options(${NAME}-guest PRIVATE -fwrapv)
 
+  if (EXISTS "${SOURCE_LDS_FILE_ABS}")
```
neobrain, line 82:
```suggestion
  # Add linker script if set
  if (EXISTS "${SOURCE_LDS_FILE_ABS}")
```

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -173,3 +181,14 @@ generate(libdrm ${CMAKE_CURRENT_SOURCE_DIR}/../libdrm/libdrm_interface.cpp thunk
...
 target_include_directories(libdrm-guest-deps INTERFACE /usr/include/libdrm/)
 add_guest_lib(drm)
+
+generate(libVDSO ${CMAKE_CURRENT_SOURCE_DIR}/../libVDSO/libVDSO_interface.cpp thunks function_packs function_packs_public)
+add_guest_lib(VDSO)
+# Can't use a stack protector because otherwise cross-compiling fails
+# Not necessary anyway because it only trampolines
+target_compile_options(VDSO-guest PRIVATE "-fno-stack-protector")
+target_link_options(VDSO-guest PRIVATE "-T" "${CMAKE_CURRENT_SOURCE_DIR}/../libVDSO/libVDSO_Guest.lds" "-nostdlib" "-Wl,-soname,linux-vdso.so.1"
+  "-Wl,--no-undefined" "-Wl,-z,max-page-size=4096" "-Wl,--hash-style=both")
```
neobrain, line 191:
```suggestion
  "LINKER:--no-undefined" "LINKER:-z,max-page-size=4096" "LINKER:--hash-style=both")
```

Not sure this matters much since we only ever use ld/lld here, but in principle this is portable across linkers and we use it like that in HostLibs already.

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -173,3 +181,14 @@ generate(libdrm ${CMAKE_CURRENT_SOURCE_DIR}/../libdrm/libdrm_interface.cpp thunk
...
+
+# Need to override SONAME in linker options
```
neobrain, line 193: Could you elaborate on why, or is that something that's obvious when you're more familiar with VDSO?

unittests/FEXLinuxTests/tests/vdso/vdso_test.cpp
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  it = VDSO.GetVDSOSymbol("__vdso_getcpu");
+  if (it) {
+    getcpu_vdso = reinterpret_cast<getcpu_type>(it);
+  }
+}
+
+
+int main() {
```
neobrain, line 122: I think this would benefit from being a Catch2 test, where we can test results more precisely than "did it crash or not?".

Not sure what effort is needed to do that with the current FexLinuxTests build setup though, so this is more of a wishlist item :)

skmp: We need #1902 merged before we can easily import catch2, I think, so canoe works in the guest context

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
skmp: I think the issue there is that the vdso is implicitly loaded, so the onload entry points aren't run. We could solve that though.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -67,6 +69,107 @@ extern char __start_HostToGuestTrampolineTemplate[];
...
+        uint32_t *cpu;
+        uint32_t *node;
+        uint64_t rv;
+      } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
+
+      args->rv = GetCPUPtr(args->cpu, args->node);
+    }
+
+    void LoadVDSO() {
+      void *vdso = dlopen("linux-vdso.so.1", RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
```
neobrain, line 135: Hmm this is also strikingly similar to the contents of `ldr_ptrs.inl`, though the dlopen logic is slightly different.

I'm starting to understand why things are done this way: The `dlopen` logic here doesn't match `ldr_ptrs.inl`, hence the autogenerated unpacking functions don't work for us, hence the autogenerated export table doesn't work for us.

If I got this right, changing things so that Thunks.cpp can pass in a custom DL handle to a thunklib's InitFN would resolve this elegantly, wouldn't it? LoadVDSO would then do almost the same things as LoadLib, and all autogenerated entities could be used.

If this sounds good, I can help with the needed interface changes if you want.

neobrain (commented):
Interesting use case with the guest-facing only thunk library. Would be great if we can find a way to make it more in line with the rest of the thunk libraries, but if I've missed something and it's not possible the changes look good overall still.

Using Catch2 for tests is still on my personal wishlist, but I don't want to bother people with it until we make more widespread use of it.

ThunkLibs/libVDSO/libVDSO_Guest.lds
```diff
@@ -0,0 +1,52 @@
...
+  text PT_LOAD FLAGS(PF_R | PF_X) FILEHDR PHDRS;
+  dynamic PT_DYNAMIC FLAGS(PF_R);
+  note PT_NOTE FLAGS(PF_R);
+}
+
+VERSION {
+  LINUX_2.6 {
+  global:
+    __vdso_time;
+    time;
```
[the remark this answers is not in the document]
neobrain: > I don't speak linker script, but does it need the exports for the non-underscore-prefixed symbols time, gettimeofday, ...?

Disregard this; I wrote the comment before I noticed all of these symbols are exported by the native libraries themselves.

ThunkLibs/GuestLibs/CMakeLists.txt
skmp: Likely the soname also needs to be overridden for all guest thunk libraries to match what the guest libraries would have.

Myself I don't exactly know how soname (DT_SONAME entry) affects things, apart that it is a "canonical name for the library"

unittests/FEXLinuxTests/tests/vdso/vdso_test.cpp
skmp: Possibly, but the compiler would be setup as a chimera with mixed guest/host configurations, so likely we'll run into compilations.

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: Every thunk guest library now sets its SONAME directly with a comment in the cmake file as to why.

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: Every thunk guest library now sets its SONAME directly with a comment in the cmake file as to why.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -67,6 +69,107 @@ extern char __start_HostToGuestTrampolineTemplate[];
...
+      if (!vdso) {
+        vdso = dlopen("linux-gate.so.1", RTLD_LAZY | RTLD_LOCAL | RTLD_NOLOAD);
+      }
+
+      if (!vdso) {
+        // Couldn't load VDSO
```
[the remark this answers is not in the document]
Sonicadvance1: Adding an informational message, since even if we can't load the host VDSO (uncommon but can happen on some ARM configurations) then we still want to pass the guest down an optimal libc implementation path still.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
Sonicadvance1: This `LoadVDSO` function has now been moved to the frontend as `LoadHostVDSO`.
I'm not sure if this would end up being worth making it in to the generic thunking interface. This is a one-off since there aren't any other thunks that are loaded in to userspace without filesystem accesses like VDSOs.

More ideally we would hardcode the guest side thunks directly in to the executable here, since binfmt_misc setups would then always have 2 pages loaded (one for 64-bit, one for 32-bit once that is in place). This would resolve the issue of VDSO thunks failing to be found inside a rootfs for example and not require an additional filesystem load. More matching what Linux on the host is doing.

But I didn't find a good way to have an `xxd` dependency, so a filesystem load and dependency it is.

unittests/FEXLinuxTests/tests/vdso/vdso_test.cpp
Sonicadvance1: If we are going to implement catch2 in the FEXLinuxTests then it should be an indepedent task from this PR. No need to complicate this PR with changing test behaviours here.

neobrain: Sounds fair, though similar line of reasoning was already applied when FLT was originally introduced. Now is a good time to revisit this though, so I went and created #1972 for tracking this.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
neobrain: > This is a one-off since there aren't any other thunks that are loaded in to userspace without filesystem accesses like VDSOs.

The problem with "one-offs" in thunks is that they increase friction when making changes to the common thunk generator (particularly when autogenerated parts interact with handwritten code that mimics other autogenerated bits). There's a natural trade-off since some APIs (like vDSO, clearly) have special requirements, but my general guideline is to avoid hand-copying boilerplate that would otherwise be autogenerated.

In this case specifically, the manually implemented unpacking functions could come back to bite us once we implement efficient (i.e. pack-less) Guest<->Host transitions.

I'm fine to wave this PR through since it looks fine otherwise and there's no specific indicator this will cause problems in the future, particularly since it's just 5 actual functions anyway.

Sonicadvance1: Reasonable there. I wonder if in the future we could shift dependencies around to pull the generated unpacking code in to the frontend similar to this.

neobrain (approved):
Fingers crossed this won't intervene with any future thunk generator changes, but looks fine now otherwise. Appreciate the level of detail in the SONAME documentation, and the vDSO-specific bits are neatly isolated as well now.

Sonicadvance1:
Thanks! Worst case we can temporarily disable while I handle any fallout from thunks changes in the future.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
neobrain: > I wonder if in the future we could shift dependencies around to pull the generated unpacking code in to the frontend similar to this

Would this be just to avoid adding vDSO-specific interfaces to Thunks.cpp, or is there a deeper reason the unpacking code has to be compiled into the frontend rather than having an internal "libvdso-host.so" that we dlopen from the frontend?

Sonicadvance1: I guess some things here is that the loading is a quite a bit different than regular thunks.
- Guest library is only copied in to location, it can't have a static initializer to load the host side.
  - Need special handling for host load regardless.
- Host side loading doesn't load an equivalent library so it needs special casing
- There's an expectation that we will be missing symbols on the host side VDSO anyway, so we need special handling around pushing these to the libc implementation anyway. But we still want to try and fetch these from VDSO anyway since we may fix some of the missing symbols in the future. 
  - time & getcpu are likely candidates to get implementations if abused.
- Also another thunk implementation where missing host symbols aren't fatal because of the special handling.
  - We still need to provide the guest with an implementation regardless. Some applications treat these like a hard-dependency. (golang had a hard dependency for a while if I remember correctly?)

PR #1942, opened 2022-08-23
VectorOps: Extend VectorZero

lioncash:
Initial opening commit to begin bringing forward some ops to support SVE. Particularly the super trivial operations that allow us to get things up and running.

More to follow after this PR, since all of the necessary helper functions and other minor things will be in place.


Sonicadvance1 (approved):
Neat!

PR #1944, opened 2022-08-24
VectorOps: Extend VectorImm

lioncash:
Extends VectorImm to be capable of using SVE to handle 256-bit length vectors.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -40,15 +40,37 @@ DEF_OP(VectorImm) {
   auto Op = IROp->C<IR::IROp_VectorImm>();
 
   const uint8_t OpSize = IROp->Size;
-  const uint8_t Elements = OpSize / Op->Header.ElementSize;
+  const uint8_t ElementSize = Op->Header.ElementSize;
+  const uint8_t Elements = OpSize / ElementSize;
 
-  if (Op->Header.ElementSize == 8) {
-    // movi with 64bit element size doesn't do what we want here
-    LoadConstant(TMP1.X(), Op->Immediate);
-    dup(GetDst(Node).V2D(), TMP1.X());
-  }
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/VectorOps.cpp
```diff
@@ -61,9 +62,13 @@ DEF_OP(VectorImm) {
 
   mov(TMP1, Element);
   vmovq(Dst, TMP1);
-  if (OpSize == 16) {
-    // Duplicate to the upper 64bits if we are 128bits
-    movddup(Dst, Dst);
+
+  if (OpSize >= 16) {
+    LOGMAN_THROW_AA_FMT(OpSize == 16 || OpSize == 32,
```
skmp, line 67: Love this assert

skmp (approved):
looks good

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -40,15 +40,37 @@ DEF_OP(VectorImm) {
...
+  const uint8_t Elements = OpSize / ElementSize;
 
-  if (Op->Header.ElementSize == 8) {
-    // movi with 64bit element size doesn't do what we want here
-    LoadConstant(TMP1.X(), Op->Immediate);
-    dup(GetDst(Node).V2D(), TMP1.X());
-  }
-  else {
-    movi(GetDst(Node).VCast(OpSize * 8, Elements), Op->Immediate);
+  if (CanUseSVE) {
```
skmp, line 46: I think `CanUseSVE` should be called more like "Support256Vectors"

lioncash: Agreed. `HostSupportsSVE` is way more descriptive. Will migrate that over after the current IR PRs get merged

Sonicadvance1 (approved):

PR #1945, opened 2022-08-24
VectorOps: Extend VMov

lioncash:
Extends VectorImm to be capable of using SVE to handle 256-bit length vectors.


skmp (commented):
Hmm, does this also cover the Interpreter?

Seeing how these are split as single PRs, I'm a bit wary of merging them without them having accompanying unit tests. 

Is it possible to also implement some frontend feature that consumes them, and a unit test that exercises them?

Sonicadvance1 (approved):

PR #1946, opened 2022-08-24
x86_64/JIT: Resolve lingering fmt deprecation warning

lioncash:
Just a log that was missed during the previous fmt deprecation cleanup.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
@@ -25,6 +25,7 @@ $end_info$
 #include <FEXCore/IR/IntrusiveIRList.h>
 #include <FEXCore/IR/RegisterAllocationData.h>
 #include <FEXCore/Utils/Allocator.h>
+#include <FEXCore/Utils/EnumUtils.h>
 #include <FEXCore/Utils/LogManager.h>
 
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -293,7 +294,8 @@ void X86JITCore::Op_Unhandled(IR::IROp_Header *IROp, IR::NodeID Node) {
       case FABI_UNKNOWN:
       default:
 #if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
-        LOGMAN_MSG_A_FMT("Unhandled IR Fallback ABI: {} {}", FEXCore::IR::GetName(IROp->Op), Info.ABI);
+        LOGMAN_MSG_A_FMT("Unhandled IR Fallback ABI: {} {}",
```
skmp, line 297: Shouldn't this be `ERROR_AND_DIE`?

skmp (approved):

External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
lioncash: All the other backends seem to use LOGMAN_MSG_A_FMT for similar scenarios, so I just left it as is

Sonicadvance1 (approved):

PR #1947, opened 2022-08-24
unittests/ThunkLibs: Fix warning about "dangerous" use of tmpnam

neobrain:
tmpnam is considered insecure since it's vulnerable to TOCTOU issues (as generating the filename and creating the directory are done separately). This is not an issue for these tests, but replacing tmpnam is not any more complicated than silencing the warning.


skmp (approved):
nice!

PR #1948, opened 2022-08-24
VectorOps: Extend VAnd/VBic/VOr/VXor

lioncash:
Fairly trivial extending that allows SVE to be used to handle 256-bit vectors


skmp (commented):
Same comment as with #1945 (https://github.com/FEX-Emu/FEX/pull/1945#pullrequestreview-1084230472)

Sonicadvance1 (approved):

PR #1949, opened 2022-08-25
InterpreterOps: Extend SSAData size to accomodate 256-bit operations

lioncash:
Extends how the SSAData block is allocated so that 256-bit operations can be implemented safely.

One other point to bring up is the preferred path forward to supporting those operations in the interpreter. We use `__uint128_t` fairly extensively, but there is no `__uint256_t` equivalent to translate over to.  Currently I was thinking of using an existing solution like [this project](https://github.com/calccrypto/uint256_t) that provides a uint256_t type, but wanted to hear other ideas if anyone had them.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
@@ -347,15 +347,17 @@ void InterpreterOps::InterpretIR(FEXCore::Core::CpuStateFrame *Frame, FEXCore::I
   constexpr size_t ListEntrySizeInBytes = sizeof(InterpVector256);
   const size_t SSADataSize = ListSize * ListEntrySizeInBytes;
 
-  InterpreterOps::IROpData OpData{};
-  OpData.State = Frame->Thread;
-  OpData.SSAData = alloca(SSADataSize);
-  OpData.CurrentEntry = Frame->State.rip;
-  OpData.CurrentIR = CurrentIR;
-  OpData.StackEntry = StackEntry;
-  OpData.BlockIterator = CurrentIR->GetBlocks().begin();
+  InterpreterOps::IROpData OpData{
+    .State = Frame->Thread,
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
I'm fine with this approach, or just defining a struct type that is 256-bit wide so we can still index the same way. Doesn't really matter to me, everything still needs to get type cast regardless.
```cpp
struct Vector256Bit {
  __uint128_t Lower;
  __uint128_t Upper;
};
```

Sonicadvance1:
Let me know if you want to keep it the current way and merge as-is, or if you want to switch it to some struct/union.

lioncash:
Oh sorry! I meant to get back to this. I'll convert it over to a struct when I wake up so the indexing is simpler

External/FEXCore/Source/Interface/Core/Interpreter/InterpreterOps.cpp
```diff
@@ -336,23 +336,28 @@ void InterpreterOps::Op_NoOp(FEXCore::IR::IROp_Header *IROp, IROpData *Data, IR:
...
-  OpData.State = Frame->Thread;
-  OpData.SSAData = alloca(ListSize * 16);
-  OpData.CurrentEntry = Frame->State.rip;
-  OpData.CurrentIR = CurrentIR;
-  OpData.StackEntry = StackEntry;
-  OpData.BlockIterator = CurrentIR->GetBlocks().begin();
-
-  // Clear them all to zero. Required for Zero-extend semantics
-  memset(OpData.SSAData, 0, ListSize * 16);
+  constexpr size_t ListEntrySizeInBytes = 32;
```
Sonicadvance1, line 346: Probably want this 32 to be `sizeof(InterpVector256)` now

lioncash: Done!

Sonicadvance1 (approved):

PR #1971, opened 2022-09-02
Syscalls: Use underscored shm syscall names 

Sonicadvance1:
Removes us needing to carry a patch downstream in the termux package
repo.


neobrain (approved):
Could do with a one-line comment somewhere mentioning why an underscore is used for those 4 syscalls only (if we don't already have that somewhere else), but looks fine!

Sonicadvance1:
Added a comment about why we are prefixing underscore on those syscall names in the generating script.

PR #1973, opened 2022-09-05
Thunks: Consolidate all generated code to one file per library per platform

neobrain:
*NOTE: This change is best viewed in a git client that supports highlighting moved files [see gitconfig's diff.colormoved]*

Previously, the thunk generator created 8 different files. In practice, all thunked libraries used those files in the same way, so splitting the output across multiple files provides no benefit. (It could be argued that things are more structured like this, but since there's no clear naming scheme, looking up specific definitions still often required to dig through multiple of these files.)

Instead of splitting generator across multiple files, the generator output is now consolidated into 2 files in total (one host-side header and one guest-side header).

As a nice side-effect, this makes some of the iteration patterns ("For each API function") more straightforward, making it more predictable what code will be generated from looking at `gen.cpp`. I'm hoping this will particularly pay off for struct-repacking (#1611), since that PR adds a few more generated symbols into the existing mix.

### DIFF
--- a/ThunkLibs/Generator/gen.cpp
+++ b/ThunkLibs/Generator/gen.cpp
@@ -527,23 +527,6 @@ void GenerateThunkLibsAction::EndSourceFileAction() {
             file << ") -> " << data.return_type.getAsString() << ";\n";
         }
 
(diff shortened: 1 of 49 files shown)


Sonicadvance1 (approved):
Very nice cleanup. Makes it a lot easier when grepping the generated files for some information.

PR #1979, opened 2022-09-13
FEXConfig: Ensure APP_CONFIG_NAME isn't stored in json

Sonicadvance1:
Also in FEXLoader make sure to use `EraseSet` for these runtime options.

Fixes a bug where the config was being set to nothing, breaking the ThunksDB configuration option.

Makes it so the following json works again.
```
~/.fex-emu/AppConfig$ cat vulkaninfo.json
{
  "ThunksDB": {
    "GL": 1,
    "Vulkan": 1
  }
}
```

### DIFF
--- a/Source/Tests/FEXLoader.cpp
+++ b/Source/Tests/FEXLoader.cpp
@@ -328,9 +328,9 @@ int main(int argc, char **argv, char **const envp) {
     return -ENOEXEC;
   }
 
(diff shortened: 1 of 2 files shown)


lioncash:
Seems like a spurious test failure, but LGTM

lioncash (approved):

Issue #1980, opened 2022-09-13
FEXLinuxTests: SIGFPE not raised in ./synchronous-signal-block.64


neobrain:
`./synchronous-signal-block.64 sfpe` and `./synchronous-signal-block.64 afpe` should print `Got 8 at address: ...` since they're triggering a SIGFPE. However, when running these tests in FEXInterpreter, that doesn't seem to be the case.

I have not ran those tests on an x86 host yet, so it's unclear if the bug is in FEX or in the test.

Sonicadvance1:
For one thing, the test is abusing the fact that integer divide by zero on x86 results in SIGFPE.
Secondly, AArch64 divide by zero doesn't raise an exception.
Thirdly, the only ARM hardware that supports float exceptions is Apple hardware, Which we check for in HostFeatures.cpp with `SupportsFloatExceptions`, but we don't currently enable or wire up at all.

neobrain:
Thanks, I forgot about those architecture constraints.

This is actually mentioned in the `Known_Failures` lists already, so I'm closing this issue in favor of those.

PR #1981, opened 2022-09-13
FEXLinuxTests: Migrate to Catch2

### LINKED ISSUE #1972: Consider using Catch2 in FEXLinuxTests
Current FEXLinuxTests are limited to binary test results ("does it run without crashes?") and manual inspection of test logs. Catch2 allows for structured testing, where several properties can be tested independendly, each of which is highlighted explicitly with a REQUIRES macro. Catch2's automatic test discovery could also help clean up the build system logic.

> If we are going to implement catch2 in the FEXLinuxTests then it should be an indepedent task from this PR. No need to complicate this PR with changing test behaviours here.

_Originally posted by @Sonicadvance1 in https://github.com/FEX-Emu/FEX/pull/1941#discussion_r962034993_

neobrain:
## Overview

Catch2 is a test framework that enables structured testing using assertion macros like `CHECK` that accumulate results and print out a nice summary at the end. FEXLinuxTests was previously limited to binary test results ("does it run without crashes?") and manual inspection of test logs. With the new test framework, several properties can be tested independently, using a language that clearly highlights what exactly is being tested. This also makes test automation easier, as test success is not defined by observed logging output anymore.

We'll also be able to drop some ugly build system hacks we employed for testing before. For example, many FEXLinuxTests define a `auto args = firstrun secondrun` variable in their sources that controls how often the test is invoked. This is not necessary with a test framework in place, since we can just use independent checks or test sections instead.

Catch2 also provides more sophisticated features like fixtures and sections, which allow for unifying boilerplate setup code. These are not used currently in FLT, other than one minor use of the SECTION macro in one test. 

Fixes #1972.

## Migration plan

All existing FEXLinuxTests will be migrated as part of this PR. Legacy "bring your own main()" tests should not be added to this subtree after merge.

## Interaction with `guest_test_runner`

`guest_test_runner` has functionality for `Known_Failures` lists that operate on test executables (or rather specific invocations thereof). In addition, Catch2 has a built-in feature to mark individual subtests (of which there may be multiple in a single test executable) as expected to fail. Unfortunately neither of the features is clearly superior to the other, since our testing setup supports individual subtest flagging per-host-arch and per-running-environment:

|   | Catch2 | `guest_test_runner` |
|--|------------|---------|
|Known failures per-test rather than per-file|✅|❌|
|Known failures per host-arch (ARM vs x86)|❓[^2]|✅|
|Known failures per running environment (FEX vs native)|❓[^1]|✅|

[^1]: Might be possible to implement using runtime-detection of FEX coupled with conditionally setting the `[!shouldfail]` test case tag 
[^2]: Might be possible to implement using runtime-detection of FEX coupled with conditionally setting the `[!shouldfail]` test case tag. Needs FEX to expose information about the host CPU architecture.

Without runtime FEX detection, Catch2's `[!shouldfail]` tags are not useful for FEXLinuxTests, since the tests should by design always successfully run outside of FEX. `[!mayfail]` annotations are not a viable fallback since they don't catch "accidental" fixes.

This means with the move to Catch2 we lose the ability to mark individual subtests as KnownFailure, since the way this was enabled previously was to run the same test executable several times with different parameters (based on annotations in the C++ source). With Catch2, all subtests are ran in a single invocation of the test executable.

For the current set of FEXLinuxTests, failing tests are all localized to individual test executables already. If we need subtest-based KnownFailure lists in the future, we may need to add better integration of Catch2 into `guest_test_runner`. Custom Catch2 tags may come in handy here.

Another venue of experimentation may be to use runtime-detection of FEX as indicated in the footnotes of the table above.

## Implementation details / challenges

Converting tests over was mostly straightforward. I haven't aimed for an "idiomatic" transition to minimize risk of test regressions, so it was mostly a matter of turning `main()` into a `TEST_CASE` and turning returned status codes into `CHECK`s.

The signal tests proved to be most challenging, since they often used `exit(0)`/`exit(1)` to indicate success/failure from within the signal handler. This circumvents the framework runtime though, so instead the signal handler now `longjmp`s back into the test logic to resume after triggering the signal.

A similar issue presents itself in the `invalid_*` tests. For these, simply skipping past the signal-triggering instruction by incrementing the program counter in the signal handler is sufficient.

There was some non-functional placeholder code in smc-1-dynamic.cpp, for which I made a best-effort guess on how to migrate it to Catch2 without making it any harder to get that placeholder code working.

smc-shared-2.cpp uses `fork()`, which forks the entire Catch2 runtime. That's not ideal, but doesn't appear to cause any issues either. Might be worth `exit()`ing the child process and checking its status code from the parent instead.

## TODO
* [x] ~~Handle test discovery using Catch2~~ (this can't be done since the tests are cross-compiled in an ExternalProject, but `catch_discover_tests` requires a visible CMake target)
* [x] Remove legacy `auto args = ` syntax
* [x] Port 32-bit tests
* [x] Integrate into test failure lists used by `guest_test_runner`
  * [x] `synchronous-signal-block` test is failing due to #1980
  * [x] `sigtest_samask`, `sigtest_samask`
  * [x] Review alternative approaches


unittests/FEXLinuxTests/tests/signal/invalid_int1.cpp
```diff
@@ -9,7 +13,7 @@ __attribute__((naked, nocf_check))
 static void InvalidINT() {
   __asm volatile(R"(
   .byte 0xF1; # int1
-  ret; # Just incase it gets past the int
```
neobrain, line 12: This line should be removed from the other `invalid_` tests too, since this is now the expected path of execution.

unittests/FEXLinuxTests/tests/signal/synchronous-signal-block.failing.cpp
```diff
@@ -0,0 +1,76 @@
...
+#include <errno.h>
+#include <malloc.h>
+#include <optional>
+#include <setjmp.h>
+#include <signal.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <sys/mman.h>
+
+#define handle_error(msg)                                                                                                                  \
```
neobrain, line 13: This is not used anymore (same for the non-failing version of this file).

neobrain:
`signals/synchronous-signal-block.cpp` have proven to be difficult to port over well. These specifically test for process exits when triggering behavior specific to Linux's POSIX implementation. The new tests now `fork()` into a child to test the exiting signal. Furthermore, for asynchronous signals we now verify that the signal handling indeed is delayed until the signal is unmasked (previously, the test would succeed even if we eagerly handled the signal ignoring the signal mask).

Will need some iteration to get the `Known_Fails` list right, but should be fine on all platforms after.

neobrain:
Ready for review.

I added an important note about the interaction between Catch2 and `guest_test_runner` in the description. By moving to Catch2, we're losing the ability to specify `KnownFailures` per subtest (instead only entire test executables can be flagged). I don't think this is critical for moving FEXLinuxTests to Catch2 now, since none of these tests rely on subtest-flagging.

For the future, we should be able to restore subtest flagging by conditionally using Catch2's `[!shouldfail]` based on CPUID checks to detect FEX at runtime. Then the only missing feature would be subtest flagging based on the host CPU architecture (ARM vs x86), for which we would likely need to extend FEX itself.

Sonicadvance1:
> Then the only missing feature would be subtest flagging based on the host CPU architecture (ARM vs x86), for which we would likely need to extend FEX itself.

The CPUID implementation in FEX also exposes the true host architecture so it is at least visible to the guest.

Sonicadvance1 (approved):
Very good work. Lots of fiddly edge cases in our tests.

PR #1984, opened 2022-09-14
Thunks: Adds functional thunk testing to CI

Sonicadvance1:
This is the bare minimum, it only tests glxinfo and vulkaninfo with and
without thunks. Nothing more special than that. Already found the .1 bug
with libvulkan host library loading.


Sonicadvance1:
Interesting failure in CI here. I've seen this before but I thought it went away. Will need to look at why vulkaninfo crashes after doing stuff.

Sonicadvance1:
The broadcom vulkan driver is breaking if it is installed on non-broadcom hardware. trying to find a way to cleanly resolve. This is where this problem comes from.

unittests/ThunkFunctionalTests/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+
+function(AddThunksTest Bin ThunksFile)
+  string(LENGTH "${ThunksFile}" ThunkFileLength)
+
+  set (ARGS
+    "-t" "${CMAKE_INSTALL_PREFIX}/lib/fex-emu/HostThunks"
+    "-j" "${CMAKE_INSTALL_PREFIX}/share/fex-emu/GuestThunks"
+    "-o" "stderr" "--no-silent" "-c" "irjit" "-n" "500"
+    )
+  if (ThunkFileLength EQUAL 0)
```
neobrain, line 11: CMake is a really good scripting language (/s), so this works:
```suggestion
  if (NOT ThunkFile)
```

unittests/CMakeLists.txt
```diff
@@ -6,10 +6,15 @@ add_subdirectory(POSIX/)
...
 if (BUILD_THUNKS)
   add_subdirectory(ThunkLibs)
 endif()
 
 if (BUILD_FEX_LINUX_TESTS)
   add_subdirectory(FEXLinuxTests/)
 endif()
+
+if (BUILD_THUNKS)
+  add_subdirectory(ThunkFunctionalTests)
```
neobrain, line 19: Why not move this into the same scope as ThunkLibs for grouping?

unittests/ThunkFunctionalTests/CMakeLists.txt
Sonicadvance1: Oh right. Resolved.

unittests/CMakeLists.txt
Sonicadvance1: Resolved.

Sonicadvance1:
Had to update the CI rootfs to get the thunk tests running. Now shipping a rootfs to the CI machines that includes mesa from https://gitlab.freedesktop.org/sonicadvance1/mesa/-/commits/fex_ci

This also unveiled some weaknesses in the script that CI runs for updating the rootfs, so that needs to live here as well. It now deletes the squashfs image on error to be more robust to problems, and also deletes the extracted folder to ensure permissions errors don't occur on update.

neobrain (approved):
Worth noting `vulkaninfo` fails to exit successfully (both with and without thunks) in Parallels VMs on M1, but we can merge this since the CI machines are fine.

PR #1986, opened 2022-09-14
FEXLinuxTests: Use the build system instead of setting up compile flags via source-code annotations

neobrain:
Previously, a RegEx was ran over the test source to extract compile flags, which means there was a circular C++ Source <-> CMake dependency. Compile flags are just specified explicitly in CMakeLists.txt now. Combined with #1981 (which removes the `args` annotation), this eliminates the circular dependency.

The intent of these annotations was presumably to make it easier to adjust compile flags on a per-test basis, but doing this in the build system is actually much cleaner.

### DIFF
--- a/unittests/FEXLinuxTests/tests-32/CMakeLists.txt
+++ b/unittests/FEXLinuxTests/tests-32/CMakeLists.txt
@@ -12,45 +12,7 @@ file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS *.cpp)
 
 foreach(TEST ${TESTS})
   get_filename_component(TEST_NAME ${TEST} NAME_WLE)
-
-  # Used to insert a configuration dependency to the test file
-  CONFIGURE_FILE(${TEST} ${CMAKE_BINARY_DIR}/junk.file)
-
-  file(READ ${TEST} TEST_CODE)
-  set(FLAGS_REGEX "//[ ]*append cxxflags: ([^\n]+)")
-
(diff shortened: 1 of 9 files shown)


Sonicadvance1 (commented):
Nice cleanup but it also looks like the build flags were lost so 32-bit tests are actually being compiled as 64-bit.

unittests/FEXLinuxTests/tests-32/CMakeLists.txt
```diff
@@ -12,45 +12,5 @@ file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS *.cpp)
...
-  else()
-    set(LIBS "")
-  endif()
-
-  set(BIN_NAME_32 "${TEST_NAME}.32")
-
-  add_executable(${BIN_NAME_32} ${TEST})
-  set_target_properties(${BIN_NAME_32} PROPERTIES COMPILE_FLAGS "${APPEND_CXX_FLAGS} -m32 -g -O2 " LINK_FLAGS "${APPEND_LD_FLAGS} -m32")
-  target_link_libraries(${BIN_NAME_32} ${LIBS})
+  add_executable(${TEST_NAME}.32 ${TEST})
```
neobrain, line 15: I accidentally dropped the `-m32` flag here, which is why CI is failing.

neobrain:
> Nice cleanup but it also looks like the build flags were lost so 32-bit tests are actually being compiled as 64-bit.

Oops, missed the notification about this before writing my own comment 😅

Fixed in the latest version, so CI is fine now.

Sonicadvance1 (approved):

PR #1989, opened 2022-09-14
CI: Adds support for flakes

Sonicadvance1:
If a test is marked as a flake then it will be tried five times before giving up.

Works around the problem of needing to babysit CI once a PR is pushed. As long as we have all the flake tests marked.


Sonicadvance1:
`conformance-interfaces-sigtimedwait-2-1.test.int.posix` Another flake that needs to be added.

Sonicadvance1:
Added the additional flakes.

lioncash (approved):

PR #1990, opened 2022-09-15
cmake: Adds uninstall target

### LINKED ISSUE #1592: No uninstall with ninja
Wondering is there plans for a way to uninstall FEX from you system

Sonicadvance1:
Following guidance from cmake's FAQ:
https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake

Due to some of the special handling that we do with installs, we need to do additional uninstall handling that the install manifest doesn't cover.

Specifically we need to add additional uninstall targets for:
- FEXInterpreter
- binfmt_misc
- guest_thunks (Doing its own uninstall target, so passthrough)

While it isn't generally advised to install and uninstall through source systems, this is something that users want to do all the time. This has been asked for a couple of times now.

Fixes #1592

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -10,6 +10,17 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 
   set(TARGET_TYPE SHARED)
   set(GENERATE_GUEST_INSTALL_TARGETS TRUE)
+
+  # uninstall target
+  if(NOT TARGET uninstall)
+    configure_file(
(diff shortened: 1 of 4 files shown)


ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -10,6 +10,17 @@ if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
 
   set(TARGET_TYPE SHARED)
   set(GENERATE_GUEST_INSTALL_TARGETS TRUE)
+
+# uninstall target
```
neobrain, line 14:
```suggestion
  # uninstall target
```

neobrain (approved):

PR #1998, opened 2022-09-17
StructVerifier: Fixes CI failure

Sonicadvance1:
The x86 runner had unattended-upgrades accidentally still enabled. It upgraded a bunch of development packages which broke CI.

Fix the struct verifier so it works with the new packages. Sadly python3-clang doesn't support all the new CursorKind types so we need to self-define some of them for now.

Once this tool gets converted over to C++ it will be a non-issue.

### DIFF
--- a/Scripts/StructPackVerifier.py
+++ b/Scripts/StructPackVerifier.py
@@ -10,6 +10,18 @@ import logging
 logger = logging.getLogger()
 logger.setLevel(logging.WARNING)
 
+# These defines are temporarily defined since python3-clang doesn't yet support these.
+# Once this tool gets switched over to C++ then this won't be an issue.
+
+# Expression that references a C++20 concept.
+CursorKind.CONCEPTSPECIALIZATIONEXPR = CursorKind(153),
+
+# C++2a std::bit_cast expression.
+CursorKind.BUILTINBITCASTEXPR = CursorKind(280)
+
+# a concept declaration.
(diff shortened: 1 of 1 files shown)


Scripts/StructPackVerifier.py
```diff
@@ -10,6 +10,18 @@
 logger = logging.getLogger()
 logger.setLevel(logging.WARNING)
 
+# These defines are temporarily defined since python3-clang doesn't yet support these.
```
neobrain, line 13: Why do we need these defines despite not using them?

Sonicadvance1: We need to define these globally ourselves since python3-clang will pull them in to their global namespace. Otherwise we get an assert when we query the CXCursor kind through its `kind` function.
Python3-clang is missing a couple hundred CursorKind values sadly.

phire:
So the issue is that libclang was updated, but not the corresponding python package?

Sonicadvance1:
The primary issue is that libstdc++ was updated. Python3-clang even on the newest 15.0 revision doesn't have the new defines necessary to parse the new libstdc++ headers.

phire:
Ohhhh. That makes sense. And the new headers are c++20.

Though really awkward that the python defines are that far out of sync; Guess this is a workable workaround until then.

phire (approved):

Sonicadvance1:
Hopefully once we hit the issue again we'll have switched over to C++ for this tool already

PR #1999, opened 2022-09-18
CMake: Add toolchain file for 32-bit cross-compiler

neobrain:
A dedicated toolchain for 32-bit is required to make sure dependencies of 32-bit targets are compiled using the correct compiler/settings. For instance, the FEXLinuxTests in #1981 fail to compile for 32-bit if `-m32` is only specified for the test targets themselves but not their Catch2 dependency.


phire (approved):
I think I'll end up moving the x86 toolchains to clang/lld later.

But this looks good for now.

Sonicadvance1 (approved):
Also seems fine here, just needs #1998 merged and this rebased.

PR #2000, opened 2022-09-19
CI: Fixes struct verifier on Ubuntu 20.04

Sonicadvance1:
Older clang fails to pull in these include paths when cross compiling. Add them manually.


lioncash (approved):

neobrain (approved):
Does changing the targets to `--target=i686-linux-gnu` et al make it work without having to specify the include directories manually?

If not, looks good to me.

Sonicadvance1:
> Does changing the targets to `--target=i686-linux-gnu` et al make it work without having to specify the include directories manually?
> 
> If not, looks good to me.

nope

PR #2001, opened 2022-09-19
Arm64: Centralize location for register defines

lioncash:
Gets rid of a few repeated definitions and allows the emitter itself to make use of these defines without causing a circular dependency on the JIT.


Sonicadvance1 (approved):
Nice cleanup

PR #2002, opened 2022-09-19
JITs: Expand max spill slot size to 32 bytes

lioncash:
This will be necessary to handle spilling 256-bit vectors. Also centralizes the constant into one spot so that future modifications to slot sizes (if any), will only need to change the constant value.


Sonicadvance1 (approved):
Very nice.

PR #2005, opened 2022-09-22
Arm64: Fixes SVE VectorImm

Sonicadvance1:
SVE DUP instruction does sign extension on the incoming immediate, while ASIMD MOVI does zero extension.

If the immediate doesn't fit then move in to a GPR first and then DUP from GPR.


lioncash (approved):
Nice catch!

PR #2010, opened 2022-09-24
Thunks: Implement the Thunk IR op for 32-bit mode

Sonicadvance1:
Use the fastcall ABI for 32-bit x86 to make our lives easier. Fastcall ABI puts the first two 32-bit arguments in ECX and EDX respectively.

Compilers are nice today and allow us to do cross-abi function calls like this.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -126,10 +126,20 @@ void OpDispatchBuilder::ThunkOp(OpcodeArgs) {
   const uint8_t GPRSize = CTX->GetGPRSize();
   uint8_t *sha256 = (uint8_t *)(Op->PC + 2);
 
-  _Thunk(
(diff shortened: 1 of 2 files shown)


phire:
Is it worth thinking about `__vectorcall` instead? 

I haven't really researched the issue.

Sonicadvance1:
> Is it worth thinking about `__vectorcall` instead?
> 
> I haven't really researched the issue.

Might be worth it in the future once we start doing argument packing optimizations. Currently the only argument we pass is a single pointer to the struct we build on the stack. So it won't change behaviour currently at least.

Sonicadvance1:
Looks like GCC doesn't support `__vectorcall`, while in clang we could use `[[clang::vectorcall]]`.
If we want to go down a more optimal path with GCC then we can use `__attribute__((fastcall, sseregparm))`, which only gives us three float arguments in SSE registers rather than the five of vector call.
Alternatively if we go down the path of using clang for cross-compiling we can just use vectorcall then

PR #2012, opened 2022-09-25
32-bit VDSO support

### LINKED ISSUE #690: 32bit VDSO
Sonicadvance1 implementing

Sonicadvance1:
As promised, here is the other half of the hand-written host-side implementation of VDSO.
VDSO for 32-bit and 64-bit are the only thunks that need this handwritten path.

This has ~~#2008~~, ~~#2009~~, and ~~#2010~~ merged in to which needs to be merged before this PR is merged.
Fixes #690

Some special things here
1) Now defaulting to hidden visibility
2) 32-bit thunks compiled with features to make PLT overhead more bearable
3) Only fexthunks_exports_<library> and other required symbols should be publicly exposed.
4) 32-bit VDSO needs clock_gettime64 and vsyscall on top of the other symbols
5) 32-bit thunks use fastcall ABI to reduce overhead

The PLT and hidden things are required to reduce overhead of thunks on 32-bit:
Without:
Before

With hidden plus PLT compile flags:
After

As we can see, it removes a function call and some EAX usage.

### DIFF
--- a/Source/Tests/VDSO_Emulation.cpp
+++ b/Source/Tests/VDSO_Emulation.cpp
@@ -1,5 +1,6 @@
 #include "VDSO_Emulation.h"
 #include "FEXCore/IR/IR.h"
+#include "Tests/LinuxSyscalls/x32/Types.h"
 
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Utils/MathUtils.h>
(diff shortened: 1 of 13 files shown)


Source/Tests/VDSO_Emulation.cpp
```diff
@@ -74,7 +75,94 @@ namespace FEX::VDSO {
...
+      if (args->tv) {
+        *args->tv = tv64;
+      }
+    }
+
+    static void clock_gettime(void* ArgsRV) {
+      struct ArgsRV_t {
+        clockid_t clk_id;
+        HLE::x32::compat_ptr<HLE::x32::timespec32> tp;
+        uint64_t rv;
```
Sonicadvance1, line 116: Needs to change to int

Source/Tests/VDSO_Emulation.cpp
```diff
@@ -74,7 +75,94 @@ namespace FEX::VDSO {
...
+      if (args->tp) {
+        *args->tp = tp64;
+      }
+    }
+
+    static void clock_gettime64(void* ArgsRV) {
+      struct ArgsRV_t {
+        clockid_t clk_id;
+        HLE::x32::compat_ptr<struct timespec> tp;
+        uint64_t rv;
```
Sonicadvance1, line 131: Needs to change to int

Sonicadvance1:
Thunk tests in CI failing because I forgot to make the aliases not be hidden. Will fix in a bit. (Woo CI doing its job!)
Build failure on ARM devices because of some missing 32-bit build dependency I think, will track and fix whatever it is at the same time.

Source/Tests/VDSO_Emulation.cpp
```diff
@@ -141,12 +236,27 @@ namespace FEX::VDSO {
...
+    FEX_CONFIG_OPT(ThunkGuestLibs32, THUNKGUESTLIBS32);
+
+    std::filesystem::path ThunkGuestPath{};
+    if (Is64Bit) {
+      ThunkGuestPath = std::filesystem::path(ThunkGuestLibs()) / "libVDSO-guest.so";
+    }
+    else {
+      ThunkGuestPath = std::filesystem::path(ThunkGuestLibs32()) / "libVDSO-guest.so";
+
+      VDSODefinitions[0].ThunkFunction = &FEX::VDSO::x32::time;
```
neobrain, line 251: I was surprised to see 32-bit specific code here that seemingly had no equivalent code in the 64-bit path, but turns out you're just overriding the entries that are set up at global scope. Could either use a comment for clarification here, or initialize the global table with `nullptr`s and opt for consistent initialization in both paths instead.

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -87,12 +97,22 @@ function(add_guest_lib NAME SONAME)
 
   ## Make signed overflow well defined 2's complement overflow
   target_compile_options(${NAME}-guest PRIVATE -fwrapv)
+  set (BITNESS_FLAGS "")
```
neobrain, line 100: Nit: Newline here for separation from the `-fwrapv` setup.

```suggestion

  set (BITNESS_FLAGS "")
```

ThunkLibs/include/common/Guest.h
```diff
@@ -91,7 +99,12 @@ template<auto Thunk, typename Result, typename... Args>
 inline Result CallHostFunction(Args... args) {
 #ifndef _M_ARM_64
     uintptr_t host_addr;
+#if __SIZEOF_POINTER__ == 8
     asm volatile("mov %%r11, %0" : "=r" (host_addr));
+#else
+    // XXX: This function is currently unsupported for 32-bit and needs some more work.
+    assert(false && "This is currently unsupported");
```
neobrain, line 106: Can we ifdef out all unsupported functions instead?

ThunkLibs/include/common/Host.h
```diff
@@ -72,6 +72,7 @@ struct fex_guest_function_ptr {
 
 #define EXPORTS(name) \
   extern "C" { \
+    __attribute__((visibility("default"))) \
```
neobrain, line 75: Why is this needed? Aren't those functions only accessed through the exports table?

neobrain (approved):
Looks good!

ThunkLibs/include/common/Guest.h
Sonicadvance1: We could comment out the uses. It's just libvulkan and libGL which are already not being compiled for 32-bit.
I don't mind whichever path we take here. This was mostly to make sure we don't forget about it later.

Source/Tests/VDSO_Emulation.cpp
Sonicadvance1: I did both

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: Oops, this was actually a remaining variable that was unused. Removed.

ThunkLibs/include/common/Host.h
Sonicadvance1: Oops, that was when I was overreaching with default hidden visibility. Removed.

ThunkLibs/include/common/Guest.h
Sonicadvance1: Actually, just removed this commit since it is unnecessary, since the libraries using it aren't compiled and will result in a compile error when enabled which is more useful.

PR #2015, opened 2022-09-26
ELFCodeLoader: Map primary ELF more like the kernel 

Sonicadvance1:
The kernel maps the primary ELF with a hint to some place *near* the
middle of the virtual address space. While the interpreter stays at the
top of the address space.

This also adds ASLR to the ELF loading, with a define for debugging and
testing purposes.

Requires both https://github.com/FEX-Emu/FEX/pull/2013 and ~~https://github.com/FEX-Emu/FEX/pull/2014~~ merged first.

### DIFF
--- a/External/FEXCore/Source/Utils/Allocator.cpp
+++ b/External/FEXCore/Source/Utils/Allocator.cpp
@@ -45,6 +45,8 @@ namespace FEXCore::Allocator {
   FREE_Hook free {::free};
 #endif
 
+  uint64_t HostVASize{};
+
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Looks like ASLR on the 32-bit side is managing to break something.
Will do more testing and see what that is.

Sonicadvance1:
Easy fix, I just mixed up the order of operations on the ASLR bias.

lioncash (approved):

PR #2025, opened 2022-09-27
IR: Handle 256-bit VFAddP

lioncash:
Extends VFAddP to be able to handle 256-bit vectors.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -775,17 +775,77 @@ DEF_OP(VFAdd) {
 }
 
 DEF_OP(VFAddP) {
-  auto Op = IROp->C<IR::IROp_VFAddP>();
-  switch (Op->Header.ElementSize) {
-    case 4: {
-      faddp(GetDst(Node).V4S(), GetSrc(Op->VectorLower.ID()).V4S(), GetSrc(Op->VectorUpper.ID()).V4S());
-    break;
+  const auto Op = IROp->C<IR::IROp_VFAddP>();
+  const auto OpSize = IROp->Size;
+
+  const auto ElementSize = Op->Header.ElementSize;
+
+  const auto Dst = GetDst(Node);
+  const auto VectorLower = GetSrc(Op->VectorLower.ID());
+  const auto VectorUpper = GetSrc(Op->VectorUpper.ID());
+
+  const bool Is256Bit = OpSize == 32;
+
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -775,17 +775,54 @@ DEF_OP(VFAdd) {
...
+    const auto Pred = OpSize == 32 ? PRED_TMP_32B.Merging()
+                                   : PRED_TMP_16B.Merging();
+
+    switch (ElementSize) {
+      case 2: {
+        faddp(Dst.Z().VnH(), Pred, VectorLower.Z().VnH(), VectorUpper.Z().VnH());
+        break;
+      }
+      case 4: {
+        faddp(Dst.Z().VnS(), Pred, VectorLower.Z().VnS(), VectorUpper.Z().VnS());
```
Sonicadvance1, line 797: Looks like we have an issue here where behaviour changes between ASIMD and SVE faddp.
Destination must equal vn since there is only enough encoding space for two registers, causing us to destructively write to VectorLower and also won't actually write to the Dst here.
Need a temporary and moves here for this.

Sonicadvance1:
Looks like vixl simulator might be incorrect?
```
Core1: XMM[0][0]: : 0x4240000040400000 != 0x4140000040400000 (Expected)
Core1: XMM[0][1]: : 0x4340000041400000 != 0x4340000042400000 (Expected)
```

As far as I can tell the results are interleaved incorrectly but I haven't looked too closely at the instruction implementation details.

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -775,17 +775,78 @@ DEF_OP(VFAdd) {
...
+  const auto Op = IROp->C<IR::IROp_VFAddP>();
+  const auto OpSize = IROp->Size;
+
+  const auto ElementSize = Op->Header.ElementSize;
+
+  const auto Dst = GetDst(Node);
+  const auto VectorLower = GetSrc(Op->VectorLower.ID());
+  const auto VectorUpper = GetSrc(Op->VectorUpper.ID());
+
+  if (HostSupportsSVE) {
```
Sonicadvance1, line 787: Maybe this path should only be used for 32byte OpSize instructions, considering how disgusting it is?

lioncash: Yeah, good call. Will do that

Sonicadvance1 (approved):

PR #2044, opened 2022-09-29
IR: Handle 256-bit VSMax/VUMax

lioncash:
Extends VSMax and VUMax to handle 256-bit vectors


Sonicadvance1 (approved):
It's funny that SVE supports 64bit integer min/max and they decided not to port it back to ASIMD.

PR #2056, opened 2022-10-10
Thunks/Host: Adds bool operator to fex_guest_function_ptr

Sonicadvance1:
Lets us check if nullptr was passed in

### DIFF
--- a/ThunkLibs/include/common/Host.h
+++ b/ThunkLibs/include/common/Host.h
@@ -61,13 +61,17 @@ typedef void fex_call_callback_t(uintptr_t callback, void *arg0, void* arg1);
  */
 struct fex_guest_function_ptr {
 private:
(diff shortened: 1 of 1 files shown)


ThunkLibs/include/common/Host.h
```diff
@@ -68,6 +68,10 @@ struct fex_guest_function_ptr {
 
     template<typename Ret, typename... Args>
     fex_guest_function_ptr(Ret (*ptr)(Args...)) : value(reinterpret_cast<void*>(ptr)) {}
+
+    inline operator bool() const {
+      return value != nullptr;
```
lioncash, line 73: With this operator being introduced, the `[[maybe_unused]]` tag can be removed from `value`

Sonicadvance1: You right, removed it.

PR #2059, opened 2022-10-10
Thunks: Adds libOpenCL

Sonicadvance1:
This takes a bit of faff due to all of the different style of callbacks that it does. These have different lifetimes depending on the object which means that a generator for these can't really exist.

Depends on the previous PRs to be merged before use.

Needs functional testing which I'll open a PR after this.

### DIFF
--- a/ThunkLibs/libOpenCL/libOpenCL_Common.h
+++ b/ThunkLibs/libOpenCL/libOpenCL_Common.h
@@ -1,0 +1,189 @@
+#pragma once
+
+#include <cstring>
+#include <memory>
+#include <new>
+
+#define CL_VERSION_3_0
+#define CL_USE_DEPRECATED_OPENCL_1_0_APIS
+#define CL_USE_DEPRECATED_OPENCL_1_1_APIS
+#define CL_USE_DEPRECATED_OPENCL_1_2_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_0_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_1_APIS
+#define CL_USE_DEPRECATED_OPENCL_2_2_APIS
+#define CL_TARGET_OPENCL_VERSION 300
+#include <CL/cl.h>
+
+namespace FEX::CL {
+  enum class AsyncCallbackType {
+    // Tied to context
+    TYPE_CLCREATECONTEXT,
+    TYPE_CLCREATECONTEXTFROMTYPE,
+    TYPE_CLSETCONTEXTDESTRUCTORCALLBACK,
+    TYPE_CLSETMEMOBJECTDESTRUCTORCALLBACK,
+    TYPE_CLSETPROGRAMRELEASECALLBACK,
+    TYPE_CLSETEVENTCALLBACK,
+    // Tied to cl_progam
+    TYPE_CLBUILDPROGRAM,
+    TYPE_CLCOMPILEPROGRAM,
+    TYPE_CLLINKPROGRAM,
+    // Tied to command_queue
+    TYPE_CLENQUEUENATIVEKERNEL,
+    TYPE_CLENQUEUESVMFREE,
+  };
+
+  union CallbackData {
+    struct _Header {
+      AsyncCallbackType Type;
+#ifdef GUEST_THUNK_LIBRARY
+      uintptr_t cb;
+#else
+      fex_guest_function_ptr cb;
+#endif
+    };
+  };
+
+  struct Type_SingleUserData {
+    CallbackData::_Header Header;
+
+    // User provided data that we need to keep around
+    void *user_data;
+
+    template <AsyncCallbackType Type>
+    static Type_SingleUserData Create(decltype(CallbackData::_Header::cb) a_0, void * a_1) {
+      static_assert(Type == AsyncCallbackType::TYPE_CLCREATECONTEXT ||
+        Type == AsyncCallbackType::TYPE_CLCREATECONTEXTFROMTYPE ||
+        Type == AsyncCallbackType::TYPE_CLSETCONTEXTDESTRUCTORCALLBACK ||
+        Type == AsyncCallbackType::TYPE_CLSETMEMOBJECTDESTRUCTORCALLBACK ||
+        Type == AsyncCallbackType::TYPE_CLSETPROGRAMRELEASECALLBACK ||
+        Type == AsyncCallbackType::TYPE_CLSETEVENTCALLBACK ||
+        Type == AsyncCallbackType::TYPE_CLBUILDPROGRAM ||
+        Type == AsyncCallbackType::TYPE_CLCOMPILEPROGRAM ||
+        Type == AsyncCallbackType::TYPE_CLLINKPROGRAM ||
+        Type == AsyncCallbackType::TYPE_CLENQUEUESVMFREE
+        ,
+        "Mismatch type for handling clProgram");
+
+      FEX::CL::Type_SingleUserData Data {
+        .Header {
+          .Type = Type,
+          .cb = a_0,
+        },
+        .user_data = a_1,
+      };
+
+      return Data;
+    };
+  };
+
+  struct Type_SingleUserDataCopy {
+    CallbackData::_Header Header;
+
+    // User provided data that we need to keep around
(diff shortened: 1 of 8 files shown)


Sonicadvance1:
Had to update the PR to bring in OpenCL-Headers version v2022.09.30 so older build environments can build this. Kind of amazing that OpenCL is still a moving target.

ThunkLibs/libOpenCL/libOpenCL_Host.cpp
```diff
@@ -0,0 +1,595 @@
...
+
+#include "common/CrossArchEvent.h"
+#include "common/Host.h"
+#include <vector>
+
+#include "libOpenCL_Common.h"
+
+#include "thunkgen_host_libOpenCL.inl"
+
+// clCreateContext - pfn_notify - Is live until context is destroyed (with refcount)
```
neobrain, line 23: This is missing a heading comment explaining why the thunk library needs to worry about lifetimes in the first place.

ThunkLibs/libOpenCL/libOpenCL_Common.h
```diff
@@ -0,0 +1,229 @@
...
+  };
+
+  struct Type_SingleUserDataCopy {
+    CallbackData::_Header Header;
+
+    // User provided data that we need to keep around
+    void *user_data;
+    size_t size;
+    ~Type_SingleUserDataCopy() {
+      free(user_data);
```
neobrain, line 84: Any particular reason for using malloc/free over new/delete?

ThunkLibs/libOpenCL/libOpenCL_Common.h
```diff
@@ -0,0 +1,229 @@
...
+          .Type = Type,
+          .cb = a_0,
+        },
+        .user_data = Memory,
+        .size = size,
+      });
+    };
+  };
+
+  union GuestCallbackData {
```
neobrain, line 105: Couldn't this be solved more cleanly using `struct GuestCallbackData : CallbackData::_Header`?

ThunkLibs/libOpenCL/libOpenCL_Common.h
```diff
@@ -0,0 +1,229 @@
...
+    CallbackData::_Header Header;
+
+    struct Type_clCreateContext {
```
neobrain, line 108: Couldn't this be solved more cleanly using struct `Type_clCreateContext : CallbackData::_Header`?

ThunkLibs/libOpenCL/libOpenCL_Common.h
```diff
@@ -0,0 +1,229 @@
...
+        Type_clSetContextDestructorCallback Data;
+        Data.Header.Type = OriginalData->Header.Type;
+        Data.Header.cb = OriginalData->Header.cb;
+        Data.a_0 = a_0;
+        Data.a_1 = OriginalData->user_data;
+        return Data;
+      }
+    } clSetContextDestructorCallback;
+
+    struct Type_clProgram {
```
neobrain, line 143: This seems to duplicate a lot of functionality provided by `PackedArguments`. Couldn't most of the `Type_` definitions be replaced with a template struct?

ThunkLibs/libOpenCL/libOpenCL_Common.h
Sonicadvance1: This is untyped data, the OpenCL function `clEnqueueNativeKernel` only gives us a `void* args` and `size_t cb_args`. Which by spec the driver needs to make a copy of.
Since this is untyped, we could change this over to `operator new` and `operator delete`, which feels more disgusting than just malloc+free.

ThunkLibs/libOpenCL/libOpenCL_Common.h
```diff
@@ -0,0 +1,229 @@
...
+
+      return Data;
+    };
+  };
+
+  struct Type_SingleUserDataCopy {
+    CallbackData::_Header Header;
+
+    // User provided data that we need to keep around
+    void *user_data;
```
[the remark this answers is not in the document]
Sonicadvance1: Mostly because all of these pointers from the OpenCL API is always void*, it's opaque to the API, so why change its type.

ThunkLibs/libOpenCL/libOpenCL_Common.h
Sonicadvance1: Since this is data getting pushed between Guest and Host, keeping it POD is a requirement. Making it a derived class would break that.

ThunkLibs/libOpenCL/libOpenCL_Common.h
Sonicadvance1: Maybe? Sounds like it would make the interface more complex to read. Only three have the same arrangement of arguments where a_0 changes between `cl_program`, `cl_mem`, and `cl_context`. The other four all have different arrangements of arguments and of course POD is still a requirement.

ThunkLibs/libOpenCL/libOpenCL_Host.cpp
Sonicadvance1: Added a comment block about this.

ThunkLibs/libOpenCL/libOpenCL_Common.h
neobrain: Why is new/delete more disgusting than malloc/free? Using the latter feels rather foreign in a C++ codebase.

Sonicadvance1: I switched it over already, but operator new and operator delete for opaque pointer is...weird.

neobrain: Ah, my bad for not seeing you updated the code already.

Why is it using `operator new` rather than `new`? Is that related to our allocator tricks?

Should `~Type_SingleUserDataCopy` be using `delete[]` instead of `delete`? Not sure if `operator delete(user_data);` is equivalent to the former.

Sonicadvance1:
```
void *Get(size_t size) {
    return new void[size];
}
```
```
<source>:4:16: error: allocation of incomplete type 'void'
    return new void[size];
```

As far as I'm aware you can't do opaque new and delete.

neobrain: Welp, TIL. I've always just done `new char[size]` in these cases, but your approach is actually more correct.

Does the OpenCL spec say anything about data alignment, or is it alright that we potentially downgrade alignment here?

Sonicadvance1: Nothing in spec talks about alignment of this data. The data gets shuttled off to the GPU/OpenCL Device so likely ends up in a UBO and then the driver needs to deal with whatever BS the user passes in.

ThunkLibs/libOpenCL/libOpenCL_Common.h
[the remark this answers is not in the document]
neobrain: I think I misunderstood `User provided` when first glancing over this. `std::any` definitely would be overkill here.

However, I think using `std::unique_ptr` allows for omitting the explicit destructor and ensures we don't accidentally create copies of this struct (which would trigger an invalid `delete` currently). Not 100% sure how it interacts with that `operator delete()`, however.

ThunkLibs/libOpenCL/libOpenCL_Common.h
neobrain: Would something like this work? https://godbolt.org/z/dTKEooMdo

Generally I'd like to avoid playing manual code generator in thunking code, since it makes structural changes harder to pull off in the future.

Sonicadvance1: You're right, that does work. Changed the code over to use this.

ThunkLibs/libOpenCL/libOpenCL_Common.h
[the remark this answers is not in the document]
Sonicadvance1: unique_ptr doesn't allow you to create a type of `unique_ptr<void>` to allow generic allocations.

PR #2062, opened 2022-10-11
SMC: Fix possible deadlock

wannacu:


Sonicadvance1 (approved):
Thanks for the fixes!

PR #2069, opened 2022-10-13
Flags: Refine _Bfe's shift

wannacu:
It will be helpful to translate lsl Eb, Ib and ror Eb, Ib correctly


Sonicadvance1:
I'm curious if you noticed this as a bug in some application or just by inspecting source.

Sonicadvance1 (approved):

PR #2072, opened 2022-10-13
unittests: Amend mm register usage in H0F38/66_04.asm test

lioncash:
This should be using xmm2 rather than mm2.


Sonicadvance1 (approved):
Nice catch!

PR #2073, opened 2022-10-13
IR: Handle 256-bit VSXTL/VSXTL2

lioncash:
Extends VSXTL and VSXTL2 to handle 256-bit vectors.


Sonicadvance1 (approved):
I love the comment describing the issues for the SVE implementation, looks great!

PR #2075, opened 2022-10-14
FEXCore: Adds support for a timeline profiler interface

Sonicadvance1:
This creates a generic interface that FEXCore can use for timeline profiling. This allows us to create a generic interface which the backend details are hidden so we can support multiple timeline profile APIs.

The only API supported right now is ftrace/gpuvis. Which is extremely lightweight of an interface with minimal overhead.

We must be careful here since in most cases will will have dozens of FEX instances running at any given time. So a timeline profiler like Microprofiler can have major issues since that only ever expects a single process at a time.

Not enabled by default but just needs the `ENABLE_FEXCORE_PROFILER` cmake option set to enable.

Gives us pretty pictures like this:
Image_2022-10-13_18-16-24

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,10 +29,23 @@ option(ENABLE_INTERPRETER "Enables FEX's Interpreter" FALSE)
 option(ENABLE_CCACHE "Enables ccache for compile caching" TRUE)
 option(ENABLE_TERMUX_BUILD "Forces building for Termux on a non-Termux build machine" FALSE)
 option(ENABLE_VIXL_SIMULATOR "Forces the FEX JIT to use the VIXL simulator" FALSE)
+option(ENABLE_FEXCORE_PROFILER "Enables use of the FEXCore timeline profiling capabilities" FALSE)
+set (FEXCORE_PROFILER_BACKEND "gpuvis" CACHE STRING "Set which backend you want to use for the FEXCore profiler")
 
 set (X86_32_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_32.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting i686")
 set (X86_64_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_64.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
 set (DATA_DIRECTORY "${CMAKE_INSTALL_PREFIX}/share/fex-emu" CACHE PATH "global data directory")
+
+if (ENABLE_FEXCORE_PROFILER)
+  add_definitions(-DENABLE_FEXCORE_PROFILER=1)
+  string(TOUPPER "${FEXCORE_PROFILER_BACKEND}" FEXCORE_PROFILER_BACKEND)
+
+  if (FEXCORE_PROFILER_BACKEND STREQUAL "GPUVIS")
+    add_definitions(-DFEXCORE_PROFILER_BACKEND=1)
+  else()
+    message(FATAL_ERROR "Unknown FEXCore profiler backend ${FEXCORE_PROFILER_BACKEND}")
+  endif()
+endif()
 
 # uninstall target
 if(NOT TARGET uninstall)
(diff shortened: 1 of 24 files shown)


CMakeLists.txt
```diff
@@ -29,11 +29,24 @@ option(ENABLE_INTERPRETER "Enables FEX's Interpreter" FALSE)
...
 
 set (X86_32_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_32.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting i686")
 set (X86_64_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_64.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
 set (DATA_DIRECTORY "${CMAKE_INSTALL_PREFIX}/share/fex-emu" CACHE PATH "global data directory")
 
+if (ENABLE_FEXCORE_PROFILER)
+  add_definitions(-DENABLE_FEXCORE_PROFILER=1)
+  string(TOUPPER "${FEXCORE_PROFILER_BACKEND}" FEXCORE_PROFILER_BACKEND)
+
+  if (FEXCORE_PROFILER_BACKEND MATCHES "GPUVIS")
```
neobrain, line 43:
```suggestion
  if (FEXCORE_PROFILER_BACKEND STREQUAL "GPUVIS")
```

CMakeLists.txt
```diff
@@ -29,11 +29,24 @@ option(ENABLE_INTERPRETER "Enables FEX's Interpreter" FALSE)
...
+option(ENABLE_FEXCORE_PROFILER "Enables use of the FEXCore timeline profiling capabilities" FALSE)
+set (FEXCORE_PROFILER_BACKEND "gpuvis" CACHE STRING "Set which backend you want to use for the FEXCore profiler")
 
 set (X86_32_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_32.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting i686")
 set (X86_64_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_64.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
 set (DATA_DIRECTORY "${CMAKE_INSTALL_PREFIX}/share/fex-emu" CACHE PATH "global data directory")
 
+if (ENABLE_FEXCORE_PROFILER)
+  add_definitions(-DENABLE_FEXCORE_PROFILER=1)
+  string(TOUPPER "${FEXCORE_PROFILER_BACKEND}" FEXCORE_PROFILER_BACKEND)
```
neobrain, line 41: This seems unnecessary, since the code below will already error out when given an unknown backend.

Sonicadvance1: This just is a developer convenience feature to ensure if you set the option as `gpuvis` or `GPUVis` then it still understands the option. We do the same thing for `CMAKE_BUILD_TYPE` to make people's lives easier.

neobrain: If you insist. I'd rather we not add code like this though, particularly since CMake tends to be surprising with code that tries to be clever. For `CMAKE_BUILD_TYPE` it makes more sense since getting it wrong doesn't spawn an error, whereas here it does.

CMakeLists.txt
```diff
@@ -29,11 +29,24 @@ option(ENABLE_INTERPRETER "Enables FEX's Interpreter" FALSE)
...
+
+  if (FEXCORE_PROFILER_BACKEND STREQUAL "GPUVIS")
+    add_definitions(-DFEXCORE_PROFILER_BACKEND=1)
```
neobrain, line 44: This should be fine until we get other profiler backends. Once others are added, would be better to make this a per-file property on Profiler.cpp to avoid huge recompiles when changing the backend.

neobrain (approved):
Looks useful! I might try to hook up tracy support with this at some point.

neobrain:
Any chance you could move time measurement from `FEXCORE_PROFILE_SCOPED` to the backend? Tracy uses its own tracking, which would make ours redundant.

Other than that, a quick proof-of-concept works fine:
fex_tracy

Sonicadvance1:
I can move the time measurement to the bacend.

Sonicadvance1:
Moved time measurement to the backend

PR #2076, opened 2022-10-14
Thunks: Update Vulkan thunk to v1.3.231 

Sonicadvance1:
Only missing a few function definitions, resorted to match order of
definitions in the headers so future changes don't mix up as much

How definitions were extracted:
```
./Scripts/DefinitionExtract.py test.h -IExternals/Vulkan-Headers/include/
```

```
$ cat test.h
#define VK_USE_PLATFORM_XLIB_XRANDR_EXT
#define VK_USE_PLATFORM_XLIB_KHR
#define VK_USE_PLATFORM_XCB_KHR
#define VK_USE_PLATFORM_WAYLAND_KHR
#include <vulkan/vulkan.h>
```

Then non-Vulkan definitions were removed and placed in to `libvulkan_interface.cpp` sorted by which header included what definition


neobrain:
Looks like you autogenerated the new interface definitions using a different shell command than I used (which was simply a sorted grep over the vulkan xml IIRC), hence the diff is overly large due to reordering definitions. Ideally we'd split the main change into two patches, one for the reordering and one for the new definitions.

Mainly though, could you mention how you generated the definitions in the PR description for future reference?

Sonicadvance1:
Updated PR message to explain how definitions were updated.

neobrain (approved):
Gotta be careful about not adding too many points that require manual fixups when updating the Vulkan interface files. Seems fine for now, though.

PR #2078, opened 2022-10-15
Allocator: Expand stack space when stealing virtual address space

Sonicadvance1:
If we take all of the stack space then the auto expanding stack doesn't work and we get stuck with a small stack that breaks thunks.

### DIFF
--- a/External/FEXCore/Source/Utils/Allocator.cpp
+++ b/External/FEXCore/Source/Utils/Allocator.cpp
@@ -145,8 +145,10 @@ namespace FEXCore::Allocator {
   #define STEAL_LOG(...) // fprintf(stderr, __VA_ARGS__)
 
   std::vector<MemoryRegion> StealMemoryRegion(uintptr_t Begin, uintptr_t End) {
+    void * const StackLocation = alloca(0);
+    const uintptr_t StackLocation_u64 = reinterpret_cast<uintptr_t>(StackLocation);
     std::vector<MemoryRegion> Regions;
-    
+
     int MapsFD = open("/proc/self/maps", O_RDONLY);
     LogMan::Throw::AFmt(MapsFD != -1, "Failed to open /proc/self/maps");
 
@@ -155,6 +157,8 @@ namespace FEXCore::Allocator {
     uintptr_t RegionBegin = 0;
     uintptr_t RegionEnd = 0;
 
+    uintptr_t PreviousMapEnd = 0;
+
     char Buffer[2048];
     const char *Cursor;
     ssize_t Remaining = 0;
@@ -162,7 +166,7 @@ namespace FEXCore::Allocator {
     for(;;) {
 
       if (Remaining == 0) {
-        do { 
+        do {
           Remaining = read(MapsFD, Buffer, sizeof(Buffer));
         } while ( Remaining == -1 && errno == EAGAIN);
 
@@ -172,8 +176,8 @@ namespace FEXCore::Allocator {
       if (Remaining == 0 && State == ParseBegin) {
         STEAL_LOG("[%d] EndOfFile; RegionBegin: %016lX RegionEnd: %016lX\n", __LINE__, RegionBegin, RegionEnd);
 
-        auto MapBegin = std::max(RegionEnd, Begin);
-        auto MapEnd = End;
+        const auto MapBegin = std::max(RegionEnd, Begin);
+        const auto MapEnd = End;
 
         STEAL_LOG("     MapBegin: %016lX MapEnd: %016lX\n", MapBegin, MapEnd);
 
@@ -209,9 +213,12 @@ namespace FEXCore::Allocator {
         if (c == '-') {
           STEAL_LOG("[%d] ParseBegin; RegionBegin: %016lX RegionEnd: %016lX\n", __LINE__, RegionBegin, RegionEnd);
 
-          auto MapBegin = std::max(RegionEnd, Begin);
-          auto MapEnd = std::min(RegionBegin, End);
-          
+          const auto MapBegin = std::max(RegionEnd, Begin);
+          const auto MapEnd = std::min(RegionBegin, End);
+
+          // Store the location we are going to map.
+          PreviousMapEnd = MapEnd;
+
           STEAL_LOG("     MapBegin: %016lX MapEnd: %016lX\n", MapBegin, MapEnd);
 
           if (MapEnd > MapBegin) {
@@ -225,6 +232,7 @@ namespace FEXCore::Allocator {
 
             Regions.push_back({(void*)MapBegin, MapSize});
           }
+
           RegionBegin = 0;
           RegionEnd = 0;
           State = ParseEnd;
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -145,6 +145,8 @@ namespace FEXCore::Allocator {
   #define STEAL_LOG(...) // fprintf(stderr, __VA_ARGS__)
 
   std::vector<MemoryRegion> StealMemoryRegion(uintptr_t Begin, uintptr_t End) {
+    void *StackLocation = alloca(0);
+    uintptr_t StackLocation_u64 = reinterpret_cast<uintptr_t>(StackLocation);
```
neobrain, line 149:
```suggestion
    void * const StackLocation = alloca(0);
    const uintptr_t StackLocation_u64 = reinterpret_cast<uintptr_t>(StackLocation);
```

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -145,6 +145,8 @@ namespace FEXCore::Allocator {
   #define STEAL_LOG(...) // fprintf(stderr, __VA_ARGS__)
 
   std::vector<MemoryRegion> StealMemoryRegion(uintptr_t Begin, uintptr_t End) {
+    void *StackLocation = alloca(0);
```
neobrain, line 148: Is it safe to call this at the top of the function, rather than after all other variables are declared? If any of the variables are spilled, the stack location would change by the time we use `StackLocation`.

neobrain, line 148: Does this query the position of the last byte already allocated on stack or the position of the *next* byte that would be allocated for a non-zero size argument? Comment might be useful here. Not sure we can safely rely on the behavior of `alloca(0)` to begin with though, so just using `alloca(1)` or taking the reference to a local variable might be better.

Sonicadvance1: This is safe since we are only using the stack location to find which VMA is holding our stack when parsing through `/proc/self/maps`.

Once we find the VMA that contains the stack range we know that we have the right one and allocate the prior one so our stack can keep growing.

The way this code works is a bit hard to parse since I didn't write it. I'll likely rewrite this in the future to make it less of a nightmare.

Sonicadvance1: `alloca(0)` returns the current stack pointer, so the last byte already allocated on the stack.
We use this pattern in a few other places of code since we need to pull the stack location for various reasons.
By spec this function will return an address even if it results in a stack overflow, since it doesn't do any checking of what is passed in

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -240,6 +248,22 @@ namespace FEXCore::Allocator {
           STEAL_LOG("[%d] ParseEnd; RegionBegin: %016lX RegionEnd: %016lX\n", __LINE__, RegionBegin, RegionEnd);
 
           State = ScanEnd;
+
+          // If the previous map's ending and the region we just parsed overlaps the stack then we need to save the stackmapping.
```
neobrain, line 252:
```suggestion
          // If the previous map's ending and the region we just parsed overlap the stack then we need to save the stack mapping.
```

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -240,6 +248,22 @@ namespace FEXCore::Allocator {
...
+          // Otherwise we will only have ~144KB stack size which crashes quickly.
```
neobrain, line 253: Where does the 144kB value actually come from?

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -211,7 +215,7 @@ namespace FEXCore::Allocator {
 
           auto MapBegin = std::max(RegionEnd, Begin);
           auto MapEnd = std::min(RegionBegin, End);
-          
+
```
neobrain, line 218: This seems better placed here:
```suggestion
          // Store the previous map location
          PreviousMapEnd = MapEnd;
```

Alternatively, MapEnd should be made `const` to make it easier to see you're just storing the unmodified value later on.

External/FEXCore/Source/Utils/Allocator.cpp
neobrain: > alloca(0) returns the current stack pointer, so the last byte already allocated on the stack.

As a followup, is it the last *byte* or the last *value*?

>  By spec this function will return an address even if it results in a stack overflow, since it doesn't do any checking of what is passed in

Is the behavior guaranteed for the argument 0 though? `malloc(0)` for instance will return nullptr on some implementations, which would silently break here if a future libc update adopted similar behavior for alloca.

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -240,6 +248,22 @@ namespace FEXCore::Allocator {
           STEAL_LOG("[%d] ParseEnd; RegionBegin: %016lX RegionEnd: %016lX\n", __LINE__, RegionBegin, RegionEnd);
 
           State = ScanEnd;
+
+          // If the previous map's ending and the region we just parsed overlaps the stack then we need to save the stackmapping.
+          // Otherwise we will only have ~144KB stack size which crashes quickly.
+          if (PreviousMapEnd <= StackLocation_u64 && RegionEnd > StackLocation_u64) {
```
neobrain, line 254: How did you come up with this check? Specifically, why does `PreviousMapEnd <= StackLocation_u64` check that the previous map overlaps with the stack? (My understanding being that the stack ends at `StackLocation_u64` and begins at an address higher than that)

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: From me inspecting a crash from thunks where that was the amount of stack space used up to that point of crash.
Can change the comment to some non-constant value like "We will only have an itty-bitty stack size..."

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: From the spec wording. probably.
If alloca starts returning zero on us then all of our guest threads immediately break/crash which will be a stronger indication that something is wrong than us failing to allocate more stack space. Since they do the same thing.

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: Moved and made const.

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: Changed comment to not explicitly talk about a stack size, which will be variable.

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: Updated.

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: This is absolutely bonkers how the logic in this function works.
First step is that it parses the region below the stack space, mmap'ing it with no privileges. Saving the `PreviousMapEnd` variable to the start of the mapping address.

It then parses the stack mapping in the `State == ParseEnd` step.
This is where the `if (PreviousMapEnd <= StackLocation_u64 && RegionEnd > StackLocation_u64)` comparison comes in.
First check ensuring that the previous mapping end address is less than the current stack pointer. Ensuring we haven't progressed beyond the stack comparison point.

Second check ensuring that the /current/ parsed region is higher than the current stack pointer, ensuring that this full region is the stack region.
As long as everything is allocated in order like expected then this holds true that it is the stack.

As an additional step of safety, when we pull the last allocation off the vector of allocations, we double check that `Ptr+Size` is the start of the mapping, ensuring the regions are contiguous with each other.

Absolute nightmare, but we'll rewrite this code to be easier to parse.

Sonicadvance1:
This one still needs a final review since otherwise thunks don't work in most Proton titles.

lioncash (approved):

Issue #2021, opened 2022-09-27
32-bit wine-mono hangs on call to Math.Sin


CallumDev:
Very specific one,

Application will hang when running a .NET Framework 32-bit application under wine (using wine-mono) if that application calls Math.Sin/Math.Cos. Other functions work.

Precompiled .exe binary here:

[CallumSinCos.zip](https://github.com/FEX-Emu/FEX/files/9653755/CallumSinCos.zip)

C# source:

```cs
using System;
class Program
{
	static void Main(string[] args)
	{
		Console.WriteLine("Sin {0}", Math.Sin(Math.PI));
		Console.WriteLine("Cos {0}", Math.Cos(Math.PI));
	}
}
```

PR #2083, opened 2022-10-16
X87: Claim incoming float was in the range for trancendental ops

Sonicadvance1:
We don't detect the range of the long F80, so we need to set that the source was in range to fix sin/cos/tan calculations.

If we don't set this flag to zero then glibc will do some additional operations that causes the value to be incorrect.

Fixes the output of the test application in #2021, probably fixes some camera orientation problems in games as well.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
@@ -782,6 +782,12 @@ void OpDispatchBuilder::X87UnaryOp(OpcodeArgs) {
   // Overwrite the op
   result.first->Header.Op = IROp;
 
+  if constexpr (IROp == IR::OP_F80SIN ||
+                IROp == IR::OP_F80COS) {
+    // TODO: ACCURACY: should check source is in range –2^63 to +2^63
+    SetRFLAG<FEXCore::X86State::X87FLAG_C2_LOC>(_Constant(0));
+  }
+
   // Write to ST[TOP]
   _StoreContextIndexed(result, top, 16, MMBaseOffset(), 16, FPRClass);
 }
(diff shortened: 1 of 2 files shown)


neobrain:
Is it safe to assume the source was in range, or is this a shortcut taken as range-checking would be too expensive?

(FTR, sadly neither this PR nor 2082 fix the camera issue in Xonotic or the mouse-aim issue in Dandara)

Sonicadvance1:
It's usually safe to assume it is in range but also range checking here is very expensive.
According to documentation, `The source value must be in the range –2^63 to +2^63 radians.`

phire:
I mean... sure. 

But at some point we might want a fully accurate codepath. Can the comments make it clear this is an accuracy hack? Or maybe it would be better to start putting things like this in a "fast-default" codepath, even if we leave the fallback path as aborting unimplemented

phire (approved):

phire (dismissed):
Just need some way of finding this later if we ever wish to improve accuracy.

Sonicadvance1:
> Just need some way of finding this later if we ever wish to improve accuracy.

Should I just update the comments that it is incorrect and that we should add range checking in the future for a "accurate" mode?
Adding runtime range detection isn't really feasible for perf right now.

Sonicadvance1:
I'm fine with this paradigm, will change the comments in a bit.

Sonicadvance1:
Update the comments to `// TODO: ACCURACY: should check source is in range –2^63 to +2^63`

Sonicadvance1:
This needs a final review since otherwise camera in Source engine games is buggy.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -782,6 +782,12 @@ void OpDispatchBuilder::X87UnaryOp(OpcodeArgs) {
   // Overwrite the op
   result.first->Header.Op = IROp;
 
+  if constexpr (IROp == IR::OP_F80SIN ||
```
lioncash, line 785: Perhaps line the ops up like so:

```cpp
   if constexpr (IROp == IR::OP_F80SIN ||
                 IROp == IR::OP_F80COS) {
```

which makes it a little nicer to vertically scan down the ops being handled

lioncash (approved):
Just one nit. It's completely subjective though, so feel free to ignore it.

I think the code itself is totally fine

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
Sonicadvance1: Aligned.

PR #2084, opened 2022-10-17
ELFCodeLoader: Implement four more auxv values

### LINKED ISSUE #788: AUXV values are wrong
- AT_PHDR doesn't point to the header (crashes ldconfig), it also points to the interpreter when not static elf
- AT_PHNUM doesn't match host value (maybe it is set from the interpreter instead of the program ?)
- AT_ENTRY points to the interpreter entrypoint
- *IDs are fixed to 1000, not the current user information
- AT_HWCAP is set to 0 (bfebfbff on my laptop for 64 bit process)
- AT_HWCAP2 doesn't exist in guest
- AT_SYSINFO shouldn't exist for x86-64 binaries

Sonicadvance1:
Implements AT_PLATFORM: Ends up being `i686` or `x86_64` depending on ELF arch

Implements AT_HWCAP and AT_HWCAP2
AT_HWCAP is just CPUID function 01h EDX result
AT_HWCAP2 only has two defined bits in it, which we don't support either.

Implements AT_RANDOM
Previously we were just sticking hardcoded values in to this. Now we pass along the host's AT_RANDOM, or we generate our own if that doesn't exist

Fixes #788

### DIFF
--- a/Source/Tests/ELFCodeLoader2.h
+++ b/Source/Tests/ELFCodeLoader2.h
@@ -14,9 +14,9 @@
 #include <cstring>
 #include <filesystem>
 #include <fstream>
+#include <list>
 #include <random>
 #include <string>
-#include <vector>
 
 #include <FEXCore/Core/CodeLoader.h>
 #include <FEXCore/Core/CoreState.h>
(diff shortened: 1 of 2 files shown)


Source/Tests/ELFCodeLoader2.h
```diff
@@ -734,19 +756,40 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
...
+
   constexpr static uint64_t BRK_SIZE = 8 * 1024 * 1024;
   constexpr static uint64_t STACK_SIZE = 8 * 1024 * 1024;
 
   std::vector<std::string> Args;
   std::vector<std::string> EnvironmentVariables;
   std::vector<char const*> LoaderArgs;
 
-  std::vector<auxv_t> AuxVariables;
+  std::list<auxv_t> AuxVariables;
```
neobrain, line 778: Why change this to `std::list`?

neobrain (approved):

Source/Tests/ELFCodeLoader2.h
Sonicadvance1: When storing references from std::vector emplace_back, when it resizes the internal container the reference is invalidated. Making it impossible to cleanly modify AT_RANDOM and AT_PLATFORM auxv_t values.
So switched it to a list so they don't get invalidated.

neobrain: Ah, that makes sense 👍

PR #2087, opened 2022-10-18
External: Update vixl submodule

lioncash:
Brings in a fix for the narrowing and widening operations erroneously clearing the upper portion of the vector register when using lower narrowing instruction variants in the case where the source register and destination register are the same. I've also made a PR upstream to fix this as well (https://github.com/Linaro/vixl/pull/49)

Since this fixes the issue, we can also re-enable the tests that make use of narrowing and widening on the simulator.

This will also let me extend the narrowing and widening IR ops reliably, since I'll now have test cases


Sonicadvance1 (approved):
Nice!

PR #2088, opened 2022-10-18
IR: Handle 256-bit VSMull/VSMull2

lioncash:
Extends VSMull and VSMull2 to handle 256-bit vectors


Sonicadvance1 (approved):
Oop, nice catch.

PR #2089, opened 2022-10-19
Implements DAA, DAS, AAA, AAS, AAM and AAD instruction

wannacu:


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5439,11 +5655,17 @@ void InstallOpcodeHandlers(Context::OperatingMode Mode) {
 
     {0x20, 6, &OpDispatchBuilder::ALUOp},
 
+    {0x27, 1, &OpDispatchBuilder::DAAOp},
```
Sonicadvance1, line 5658: These instructions are only valid in 32-bit mode, so these should be moved down to the array below called `BaseOpTable_32`

Sonicadvance1:
Did you notice an application or game that was using these instructions or just noticed through inspection?
I had completely forgot about implementing these instructions.

Sonicadvance1:
Wow! Very cool!

Sonicadvance1 (approved):
A smidge dirty but considering the nature of new contributions, pretty good.

PR #2090, opened 2022-10-19
Arm64/VectorOps: Simplify SVE VSXTL/VSXTL2/VUXTL/VUXTL2 implementations

lioncash:
Turns out I completely missed the existence of SUNPKLO/SUNPKHI and UUNPKLO/UUNPKHI which do exactly what we need.


Sonicadvance1 (approved):
haha oops, ohwell, easier code is nice to understand.

PR #2091, opened 2022-10-20
IR: Handle 256-bit VSQXTN/VSQXTN2

lioncash:
Extends VSQXTN and VSQXTN2 to handle 256-bit vectors.


Sonicadvance1 (approved):
Interesting behaviour. I do love how all of these SVE ops change behaviour so the ops can be split in smaller pipelines more easily.

PR #2099, opened 2022-10-23
FEXServer: Be robust against invalid packets.

Sonicadvance1:
Chrome seems to like sending us invalid packets of data sometimes. With an invalid packet type just skip parsing the data entirely.

Fixes an infinite loop in Vampire Survivors.

### DIFF
--- a/Source/Tools/FEXServer/ProcessPipe.cpp
+++ b/Source/Tools/FEXServer/ProcessPipe.cpp
@@ -431,6 +431,9 @@ namespace ProcessPipe {
           // Invalid
         case FEXServerClient::PacketType::TYPE_ERROR:
         default:
+          // Something sent us an invalid packet. To ensure we don't spin infinitely, consume all the data.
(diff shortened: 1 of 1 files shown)


neobrain:
Aren't those packets generated from within FEX itself? Seems like we should investigate where those invalid packets come from instead of working around that, unless there's some mechanism by which emulated applications can modify our packets.
Additinally, not crashing/infinite-looping is good, but I would assume silently dropping these packets could lead to less obviously hidden problems further down the line too.

Sonicadvance1:
Added a log message to FEXServer to let us know when an invalid packet is received.
These packets are /typically/ generated from within FEX itself.
Nothing stops and application from scanning the open FDs and assuming that anything that is a socket is the application's own socket. Which is what is happening here since Vampire Survivors is an embedded chrome instance. It sends our FEXServer some packet of data, NOT from FEX, which we need to ignore.

lioncash (approved):

PR #2100, opened 2022-10-25
ThunksDB: Fixes Thunks loaded boolean pointer check

Sonicadvance1:
Need to dereference the boolean to ensure we only load the thunksDB files once.


neobrain (approved):
Good catch - not the cleanest way of carrying lambda-specific state here, but good bug fix.

PR #2101, opened 2022-10-25
Thunks: Fixes missing thunk librarie so versions

Sonicadvance1:
Some libraries were missing these version defines, which was causing dlopen to fail.

This was causing thunks to break in pressure-vessel.


neobrain (approved):
Looks good. Maybe the version annotation should be made non-optional entirely?

Sonicadvance1:
I agree, making it non-optional would make sense. Potentially with a `-1` to explicitly say we don't want a versioned library. I can think of at least one library that isn't versioned that we may want in the future.

PR #2102, opened 2022-10-25
Thunks: Adds a new cross-arch job queue system

Sonicadvance1:
This will work on both x86-64 and x86.
This gives a thread-safe job queue system.

First step is to make xcb use this new job queue, next step is to move OpenCL thunking over to it to make their async jobs safe.

Fixes the issue that multiple jobs in flight could trample state of a current job being operated on.

Uses a ring-buffer of 32 elements to ensure we have a bit of breathing room for jobs.

Also fixed uint64_t for the job pointer so we pass the same data between guest and host even with different pointer sizes.

### DIFF
--- a/ThunkLibs/include/common/CrossArchEvent.h
+++ b/ThunkLibs/include/common/CrossArchEvent.h
@@ -1,49 +1,405 @@
 #pragma once
 
+#include <cassert>
 #include <linux/futex.h>
 #include <sys/syscall.h>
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
This needs a bunch more comments on the new objects, threw it out super quickly while I was doing things.

Sonicadvance1:
All commented up!

PR #2103, opened 2022-10-25
Thunks/Guest: Enable SSE2 on thunks and set fpmath to sse

Sonicadvance1:
Clang thunks already have these default enabled, but let's also enable this on the GCC side.

sse2 will enable most things we care about, which matches ASIMD quite closely.
fpmath=sse removes some x87 usage for 32-bit thunks specifically.

Should effectively be a non-functional-change

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -103,6 +103,11 @@ function(add_guest_lib NAME SONAME)
 
   ## Make signed overflow well defined 2's complement overflow
   target_compile_options(${NAME}-guest PRIVATE -fwrapv)
+  if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
+    ## Compile for SSE2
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Looks like we have a bug where guest libraries are getting compiled for the host.

neobrain:
> sse2 will enable most things we care about, which matches ASIMD quite closely.

Are there other important changes, besides the avoided use of x87?

> Looks like we have a bug where guest libraries are getting compiled for the host.

Not a bug: This is done so that guest thunk source files are written to `compile_commands.json`, which enables IDEs to parse guest thunks properly.

There may be better ways to achieve this while avoiding compilation in regular builds. Maybe explicitly passing `-fsyntax-only` when building on the host?

Sonicadvance1:
Updated this to only enable the options using the same checks as prior in the cmake file to see if we are cross-compiling.

neobrain (approved):
Not sure I prefer the `CMAKE_CURRENT_SOURCE_DIR` check over two `CMAKE_SYSTEM_PROCESSOR` ones; I'm guessing both are equally confusing for first-time readers.

Looks fine 👍

PR #2106, opened 2022-10-25
IR: Handle 256-bit VDupElement

lioncash:
Extends VDupElement to handle 256-bit vectors.


Sonicadvance1 (approved):
Wow, I didn't know x86-64 element broadcast was so derpy for AVX.

PR #2108, opened 2022-10-25
Thunks: Add support for disabling thunks in config

Sonicadvance1:
Previously the config options could only have ever enabled thunks rather than disable them.

Now sort the code so it can enable thunks, then following configs can redisable them.  Allowing testing with global thunks enabled and disabling problematic applications.

Also sorts the "ThunkConfigFile" config as lower priority than the application configs. I wasn't thinking about ordering that hard for these five configuration paths, but application configs should be higher priority in this case.

### DIFF
--- a/Source/Tests/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tests/LinuxSyscalls/FileManagement.cpp
@@ -239,60 +239,25 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
     // Now load this property
     for (json_t const* Item = json_getChild(ThunksDB); Item != nullptr; Item = json_getSibling(Item)) {
       const char *LibraryName = json_getName(Item);
-      int64_t LibraryEnabled = json_getInteger(Item);
-      if (LibraryEnabled != 0) {
-        // If the library is enabled then find it in the DB
-        // Enable the overlay and all the dependencies in one go
-        auto DBObject = ThunkDB.find(LibraryName);
-        if (DBObject != ThunkDB.end() &&
-            DBObject->second.Enabled == false) {
-
-          auto ThunkPath = ThunkGuestPath / DBObject->second.LibraryName;
-          if (std::filesystem::exists(ThunkPath)) {
-            for (auto Overlay : DBObject->second.Overlays) {
-              // Direct full path in guest RootFS to our overlay file
-              ThunkOverlays.emplace(Overlay, ThunkPath);
-            }
-          }
-          DBObject->second.Enabled = true;
-          // Now walk the dependencies and set them up as well
-          // Make sure to enable each one as we go to remove circular dependencies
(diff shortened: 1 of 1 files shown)


Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -313,6 +278,44 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
     }
   }
 
+  // Now that we loaded the thunks object, walk through and ensure dependencies are enabled as well.
+  for (auto const &DBObject : ThunkDB) {
+    if (DBObject.second.Enabled) {
```
neobrain, line 283: To keep indentation levels low:
```suggestion
    if (!DBObject.second.Enabled) {
        continue;
    }
```

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -313,6 +278,44 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
...
+      auto ThunkPath = ThunkGuestPath / DBObject.second.LibraryName;
+      if (std::filesystem::exists(ThunkPath)) {
+        for (auto Overlay : DBObject.second.Overlays) {
```
neobrain, line 286: This avoids unnecessary copies of the contained `std::string`s:
```suggestion
        for (const auto& Overlay : DBObject.second.Overlays) {
```

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -313,6 +278,44 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
...
+      // Now walk the dependencies and set them up as well
+      // Make sure to enable each one as we go to remove circular dependencies
+      std::function<void(const std::unordered_set<std::string> &Depends)> InsertDependencies
+        = [this, &ThunkGuestPath, &InsertDependencies](const std::unordered_set<std::string> &Depends) -> void {
+        for (auto const &Depend : Depends) {
+          auto DBDepend = ThunkDB.find(Depend);
+          if (DBDepend != ThunkDB.end() &&
+              DBDepend->second.Enabled == false) {
+
+            auto ThunkPath = ThunkGuestPath / DBDepend->second.LibraryName;
```
neobrain, line 301: This block is now duplicated outside the loop. Can we instead call `InsertDependencies({DBObject.first})` at the top level? That should make the loop body a tad more organized, and the extra lookup should be cheap.

neobrain (approved):
Added some minor comments, looks fine otherwise!

Source/Tests/LinuxSyscalls/FileManagement.cpp

Source/Tests/LinuxSyscalls/FileManagement.cpp

Source/Tests/LinuxSyscalls/FileManagement.cpp

PR #2111, opened 2022-10-26
IR: Handle 256-bit Vector_FtoF

lioncash:
Extends Vector_FtoF to handle 256-bit vectors.


Sonicadvance1 (approved):
Fancy how the SVE versions differ.

PR #2113, opened 2022-10-26
IR: Check for invalid conversion masks in Float_FromGPR_S

lioncash:
Previously this would silently ignore unhandled masks.


Sonicadvance1 (approved):
Nice catch

PR #2114, opened 2022-10-26
Arm64/BranchOps: Remove unused std::vector

lioncash:
Removes a heap allocation for inline syscalls.


Sonicadvance1 (approved):
hah, oops, leftover code.

PR #2115, opened 2022-10-26
Thunks/libX11: Fix recursive initialize

Sonicadvance1:
Fixes a crash that occurs due to `_XInitDisplayLock` due to the display lock function being initialized to our own handler.

steamwebhelper was hitting this.

### DIFF
--- a/ThunkLibs/libX11/libX11_Host.cpp
+++ b/ThunkLibs/libX11/libX11_Host.cpp
@@ -5,6 +5,7 @@ desc: Handles callbacks and varargs
 $end_info$
 */
 
+#include <atomic>
 #include <cstdlib>
 #include <stdio.h>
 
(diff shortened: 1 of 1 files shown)


neobrain:
Sounds like something fishy is going on.

Maybe XInitThreads is just called twice? Looks like that would lead to the scenario this patch is working around:
```cpp
// Expansion of std::exchange from first call
ACTUAL_XInitDisplayLock_fn = *_XInitDisplayLock_fn;
*_XInitDisplayLock_fn = _XInitDisplayLock;
// Expansion of std::exchange from second call
ACTUAL_XInitDisplayLock_fn = *_XInitDisplayLock_fn; // = _XInitDisplayLock
*_XInitDisplayLock_fn = _XInitDisplayLock;
```

If that's the culprit, I'd go for fixing the issue in XInitThreadsInternal instead.

Sonicadvance1:
> Sounds like something fishy is going on.
> 
> Maybe XInitThreads is just called twice? Looks like that would lead to the scenario this patch is working around:
> 
> ```c++
> // Expansion of std::exchange from first call
> ACTUAL_XInitDisplayLock_fn = *_XInitDisplayLock_fn; // = user-provided
> *_XInitDisplayLock_fn = _XInitDisplayLock;
> // Expansion of std::exchange from second call
> ACTUAL_XInitDisplayLock_fn = *_XInitDisplayLock_fn; // = _XInitDisplayLock
> *_XInitDisplayLock_fn = _XInitDisplayLock;
> ```
> 
> If that's the culprit, I'd go for fixing the issue in XInitThreadsInternal instead.

You're right, XInitThreads getting called multiple times should be a no-op past the first call.
Switched over to a atomic CAS in that function to resolve it.

neobrain (approved):

PR #2117, opened 2022-10-26
IR: Handle 256-bit VInsGPR

lioncash:
Extends VInsGPR to handle 256-bit vectors.

This should finish off most of the conversion ops


Sonicadvance1 (approved):
Yeesh. If only there was a way to set an predicate value to true quickly, could have used CPY to duplicate a GPR to only the predicated element we selected by index.
Doubt it is any nicer to do a store of zero to stack, plus manual modify, load in to predicate and CPY.

PR #2118, opened 2022-10-26
Arm64/VectorOps: Make use of MOVPRFX where applicable

lioncash:
Allows hardware to pack the move and following destructive operation together into one constructive operation if possible.

e.g.

```
movprfx VTMP1.D, VectorLower.D
addp VTMP1.B, Pred, VTMP1.B, VectorUpper.B
```

is allowed to be merged as if it executed constructively like:

```
addp VTMP1.B, Pred, VectorLower.B, VectorUpper.B
```

if the hardware supports it. If it doesn't, then the instructions will behave like a regular move and destructive addp operation separately.


Sonicadvance1 (approved):
Very fancy!

PR #2119, opened 2022-10-27
IR: Handle 256-bit VTBL1

lioncash:
Extends VTBL1 to handle 256-bit vectors.


Sonicadvance1 (approved):
Huh, TIL about the vpshufb AVX limitation. Very interesting.

PR #2122, opened 2022-10-29
OpcodeDispatcher: Fixes ROR imm OF calculation

Sonicadvance1:
Turns out this was calculating OF incorrectly, breaking Denuvo early in its execution.

Changes the ROL imm OF calculation code as well to be more consistent and not keep src1 alive longer than it needs to be.

Also adds two new unit tests to ensure this stays correct.

Denuvo Anti-tamper still crashes, but it gets farther at least.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
@@ -938,6 +938,7 @@ void OpDispatchBuilder::CalculcateFlags_RotateRight(uint8_t SrcSize, OrderedNode
     auto OldOF = GetRFLAG(FEXCore::X86State::RFLAG_OF_LOC);
 
     // OF is set to the XOR of the new CF bit and the most significant bit of the result
+    // OF is architecturally only defined for 1-bit rotate, which is why this only happens when the shift is one.
     auto NewOF = _Xor(_Bfe(1, OpSize - 2, Res), NewCF);
 
     // If shift == 0, don't update flags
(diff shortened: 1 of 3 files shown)


neobrain:
Should OF in `CalculcateFlags_RotateRight` also be changed to be the XOR of the two MSBs?

For reference, this is what the Intel manuals say:
```
The OF flag is defined only for the 1-bit rotates; it is undefined in all other cases (except RCL and RCR instructions
only: a zero-bit rotate does nothing, that is affects no flags). For left rotates, the OF flag is set to the exclusive OR
of the CF bit (after the rotate) and the most-significant bit of the result. For right rotates, the OF flag is set to the
exclusive OR of the two most-significant bits of the result.
```

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -1003,17 +1004,21 @@ void OpDispatchBuilder::CalculcateFlags_RotateLeftImmediate(uint8_t SrcSize, Ord
...
+    SetRFLAG<FEXCore::X86State::RFLAG_CF_LOC>(NewCF);
   }
 
   // OF
   {
     if (Shift == 1) {
-      // OF is the top two MSBs XOR'd together
-      SetRFLAG<FEXCore::X86State::RFLAG_OF_LOC>(_Xor(_Bfe(1, OpSize - 1, Src1), _Bfe(1, OpSize - 2, Src1)));
+      // OF is the LSB and MSB XOR'd together
+      // OF is set to the XOR of the new CF bit and the most significant bit of the result
```
neobrain, line 1018: Might be worth adding a note this is only well-defined for 1-bit shifts, since otherwise it's unclear why those specific bits are used.

neobrain (approved):

Sonicadvance1:
> Should OF in `CalculcateFlags_RotateRight` also be changed to be the XOR of the two MSBs?
> 
> For reference, this is what the Intel manuals say:
> 
> ```
> The OF flag is defined only for the 1-bit rotates; it is undefined in all other cases (except RCL and RCR instructions
> only: a zero-bit rotate does nothing, that is affects no flags). For left rotates, the OF flag is set to the exclusive OR
> of the CF bit (after the rotate) and the most-significant bit of the result. For right rotates, the OF flag is set to the
> exclusive OR of the two most-significant bits of the result.
> ```

RorateRight already does this, which is why I managed to catch this as a difference in behaviour.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
Sonicadvance1: Updated the comment to say that OF is only architecturally defined for a 1-bit shift.

neobrain:
> RorateRight already does this, which is why I managed to catch this as a difference in behaviour.

Ah yes, correct. I was mixing up the `NewCF` definition while reading over the comment.

PR #2128, opened 2022-11-01
Thunks: Add support for asynchronous host-side callback invocation

neobrain:
## Overview

This PR resolves a limitation of FEX's thunking of APIs with function pointer parameters ("callbacks"). Previously, callbacks could only be used by the native host library if they were called *synchronously*, i.e. before returning to the guest application. *Asynchronous* calls would quickly break if the callback used nontrivial functions or accessed TLS data.

Previously, we worked around this limitation by implementing a minimal remote-procedure-call (RPC) system. This system marshaled callbacks and their arguments across architecture boundaries so that they could safely be called on a dedicated guest-thread. This heavyweight approach had to be implemented from scratch for each thunked library that needed asynchronous callbacks (see libxcb and #2059).

With this change, *asynchronous* callback invocations are instead supported by temporarily stealing the x86 context from a dummy thread provided by the guest thunk library. In effect, this executes callbacks *as if* they ran in sequence on that dummy thread, without any of the overhead of a full-blown cross-arch remote-procedure-call system.

## Impact and Scalability

Since this feature carries some overhead, this change is opt-in: Function pointers that may be called asynchronously must explicitly be made async-callable in the host-thunk library (see `MakeHostTrampolineForGuestFunctionAsyncCallable`). If need be, this could be automated with a thunkgen annotation in the future.

The dummy guest threads are currently identified by hardcoded per-library IDs, since only a small number of libraries require asynchronous callbacks (libxcb, Vulkan, OpenCL). We may want to establish a more sophisticated system for this if the number of libraries grows.

## TODO

* [ ] Make TLS registration more lightweight (we can just copy over worker thread state)
* [ ] Test libxcb for regressions
* [ ] Test on x86
* [ ] Clean up worker thread code in libxcb
* [ ] Cleanup: Rename (Un)RegisterAsyncWorkerThread, assign proper thunk IDs, finish docstrings


ThunkLibs/libxcb/libxcb_Guest.cpp
```diff
@@ -22,71 +22,35 @@ tags: thunklibs|xcb
...
+fexthunks_fex_register_async_worker_thread(&args);
 
-void FEX_Helper_GiveEvents() {
-  struct {
-    CrossArchEvent *WaitForWork;
-    CrossArchEvent *WorkDone;
-    CBWork *Work;
-  } args;
+std::unique_lock lock(m);
+var.wait(lock, [this]() -> bool { return done; });
```
Sonicadvance1, line 41: How does this handle the case that spurious wakeup on mutex will corrupt the stack of the running thread when we have stolen its state?

neobrain: Fair point. Maybe this should be moved into a thunk implemented in Thunks.cpp, which would do the waiting purely outside the VM and which wouldn't return until shutdown?

Sonicadvance1: That will...probably work

neobrain: Done, still passes my artificial test program.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -67,13 +66,22 @@ extern char __start_HostToGuestTrampolineTemplate[];
 extern char __stop_HostToGuestTrampolineTemplate[];
 
 namespace FEXCore {
+    struct AsyncCallbackState {
+      FEXCore::Core::InternalThreadState *Thread { nullptr };
+
+      std::mutex m;
```
neobrain, line 72: These need better names.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -151,15 +169,89 @@ namespace FEXCore {
...
+          } args = *reinterpret_cast<args_t*>(argsv);
+
+          auto ThunkHandler = reinterpret_cast<ThunkHandler_impl*>(Thread->CTX->ThunkHandler.get());
+
+          AsyncCallbackState *WorkerState = nullptr;
+          {
+            // Create and initialize new entry
+            std::unique_lock lock(ThunkHandler->AsyncWorkerThreadsMutex);
+            WorkerState = &ThunkHandler->AsyncWorkerThreads[args.id];
+            WorkerState->Thread = Thread;
```
Sonicadvance1, line 198: I think this should probably modify the thread state coming in to this to modify RSP to offset by the redzone.

neobrain:
Closing this: One and a half years later we still haven't found a real-world use case that actually needs this. If we do find one, we know which PR to revive.

PR #2131, opened 2022-11-02
Arm64/MemoryOps: Merge if statement into switch in ParanoidLoadMemTSO

lioncash:
There's nothing preventing the `OpSize == 1` case from being merged into the switch, so we can do that to make things a little more consistent.


Sonicadvance1 (approved):
Ah yes, these used to be wrapped by nop instructions so this makes sense now.

PR #2133, opened 2022-11-02
IR: Handle 256-bit VInsElement

lioncash:
Extends VInsElem to handle 256-bit vectors.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -3776,37 +3776,125 @@ DEF_OP(VSShrS) {
 }
 
 DEF_OP(VInsElement) {
-  auto Op = IROp->C<IR::IROp_VInsElement>();
+  const auto Op = IROp->C<IR::IROp_VInsElement>();
+  const auto OpSize = IROp->Size;
+  const auto Is256Bit = OpSize == Core::CPUState::XMM_AVX_REG_SIZE;
 
-  auto reg = GetSrc(Op->DestVector.ID());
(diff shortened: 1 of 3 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -3776,37 +3778,204 @@ DEF_OP(VSShrS) {
...
-    case 4: {
-      mov(reg.V4S(), Op->DestIdx, GetSrc(Op->SrcVector.ID()).V4S(), Op->SrcIdx);
-    break;
+
+    // Load our mask.
+    aarch64::Label DataLocation;
+    adr(TMP1, &DataLocation);
+    ld1b(VTMP3.Z().VnB(), PRED_TMP_32B.Zeroing(), SVEMemOperand(TMP1));
+
+    // Now finally combine the value from the source temporary we broadcasted into.
```
Sonicadvance1, line 3935: I'm curious if Predicated `CPY` and Predicated `SEL` could be used to make this a bit nicer or not?
Would require generating temporary predicate literals instead of a 256-bit mask.

Sonicadvance1: Looks like EXT also is very nice to allow you to barrel rotate any element in to the scalar position for predicated CPY to pick up?

lioncash: Yeah, converting it over to setting up a predicate register greatly simplifies everything

Sonicadvance1 (approved):
Nice Nice Nice Nice Nice.

PR #2146, opened 2022-11-13
ELFCodeLoader: Ensure we set AT_SYSINFO for 32-bit

Sonicadvance1:
AT_SYSINFO points to the vsyscall location that the kernel provides. This AUXV value isn't used on x86-64.
If we have VDSO installed then we can use the one provided from there, otherwise we need to provide a code page.

It seems like some behaviour changed with glibc provided in Ubuntu 22.10 that it now requires AT_SYSINFO.

Fixes wine 7.0 execution inside the Ubuntu 22.10 rootfs.

### DIFF
--- a/Source/Tests/VDSO_Emulation.cpp
+++ b/Source/Tests/VDSO_Emulation.cpp
@@ -8,6 +8,7 @@
 #include <FEXHeaderUtils/Syscalls.h>
 
 #include <dlfcn.h>
+#include <elf.h>
 #include <fcntl.h>
 #include <filesystem>
 #include <sys/mman.h>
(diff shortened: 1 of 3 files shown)


Source/Tests/VDSO_Emulation.cpp
```diff
@@ -291,6 +292,21 @@ namespace FEX::VDSO {
     return VDSOBase;
   }
 
+  uint64_t GetVSyscallEntry(void* VDSOBase) {
```
lioncash, line 295: seems like this could take a `const void*`

lioncash (commented):
LGTM aside from one nit

Source/Tests/VDSO_Emulation.cpp
Sonicadvance1: Resolved.

lioncash (approved):

PR #2149, opened 2022-11-13
ELFCodeLoader: Calculate AT_MINSIGSTKSZ

Sonicadvance1:
This is the last remaining auxv value that we were missing.

This requires a little bit of setup to match what we are doing in FEXCore's Dispatcher.

This only tracks how much space is required by the kernel to store its required state.

### DIFF
--- a/Source/Tests/ELFCodeLoader2.h
+++ b/Source/Tests/ELFCodeLoader2.h
@@ -21,6 +21,8 @@
 
 #include <FEXCore/Core/CodeLoader.h>
 #include <FEXCore/Core/CoreState.h>
+#include <FEXCore/Utils/MathUtils.h>
+#include <FEXCore/Core/UContext.h>
 #include <FEXCore/Core/X86Enums.h>
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXHeaderUtils/Syscalls.h>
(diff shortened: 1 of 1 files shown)


Source/Tests/ELFCodeLoader2.h
```diff
@@ -806,6 +809,57 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
...
+      }
+
+      Result += sizeof(siginfo_t);
+      Result = FEXCore::AlignUp(Result, alignof(siginfo_t));
+    }
+    else {
+      Result += sizeof(FEXCore::x86::ucontext_t);
+      Result = FEXCore::AlignUp(Result, alignof(FEXCore::x86::ucontext_t));
+      if (SupportsAVX) {
+        Result += sizeof(FEXCore::x86_64::xstate);
```
lioncash, line 850: Shouldn't this be `sizeof(FEXCore::x86::xstate)` and the non-avx case be using `sizeof(FEXCore::x86::_libc_fpstate)` instead of the x86_64 versions?

Sonicadvance1: Yes. Copy and paste fail, thanks for the catch.

lioncash (approved):

PR #2151, opened 2022-11-14
IR: Removes the only uses of VSLI and VSRI

Sonicadvance1:
We use these two IR ops for PSRLQ and PSLLQ respectively, Which is actually implemented in a quite inefficient way.

Instead switch the ops over to using VExtr which maps directly to one instruction and can emulate both.


Sonicadvance1:
This likely affects @lioncash who is working on AVX stuff right now.

lioncash (approved):
Nice! This helps out a bunch. Thanks!

PR #2154, opened 2022-11-15
OpcodeDispatcher: Handle VMOVAPS

lioncash:
Starts the AVX instruction implementation work. Also pls let me know if the decoder changes are A-OK. There's so much to keep track of mentally, I'm not sure if I missed anything.


Sonicadvance1:
I think the frontend changes are fine here. It would have been how I would have wrote it.
In the future we will need ASM tests that prepend VEX.L on instructions that don't support it to ensure testing, but seems fine for now.
Just need CI to go green.

lioncash:
Looking at the x86 failure. Though the ARM builder failures have me stumped, considering the AVX bits shouldn't even be running on them aside from the simulator builders

Sonicadvance1:
I think the VEX table needs to be moved to only be filled out when the host supports AVX, because we are hitting a host side SIGILL rather than a guest side SIGILL. That should resolve the ARM runners.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5724,7 +5724,7 @@ void OpDispatchBuilder::InstallHostSpecificOpcodeHandlers() {
   Initialized = true;
 }
 
-void InstallOpcodeHandlers(Context::OperatingMode Mode) {
+void InstallOpcodeHandlers(Context::OperatingMode Mode, Context::Context *CTX) {
```
Sonicadvance1, line 5727: You don't want to do this in `InstallOpcodeHandlers`, you'll want to do this in `InstallHostSpecificOpcodeHandlers` just above this function.

lioncash: Oh, duh, missed that.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -38,6 +38,18 @@ void OpDispatchBuilder::MOVAPSOp(OpcodeArgs) {
   StoreResult(FPRClass, Op, Src, -1);
 }
 
+void OpDispatchBuilder::VMOVAPSOp(OpcodeArgs) {
+  OrderedNode *Src = LoadSource(FPRClass, Op, Op->Src[0], Op->Flags, -1);
+  const auto Is128BitDest = GetDstSize(Op) == Core::CPUState::XMM_SSE_REG_SIZE;
+
+  if (Op->Dest.IsGPR() && Is128BitDest) {
+    // Perform 32-bit store to clear the upper lane.
```
Sonicadvance1, line 46: 32-bit or 32-byte?

lioncash: oh whoops, definitely meant byte there. Was probably thinking about 256-bit and 32-byte at the same time and both won out in their own way.

Sonicadvance1 (approved):
Nice nice nice nice nice.

PR #2155, opened 2022-11-16
x86_64/VectorOps: Separate 128-bit/256-bit paths

lioncash:
Ensures we have more convenient upper-lane clearing behavior with AVX 128-bit operations


Sonicadvance1 (approved):
Powerful code movement.

PR #2156, opened 2022-11-17
Systemd fixes

Sonicadvance1:
Systemd has....problems that we need to work around.

### DIFF
--- a/Source/Tools/FEXServer/ProcessPipe.cpp
+++ b/Source/Tools/FEXServer/ProcessPipe.cpp
@@ -83,7 +83,7 @@ namespace ProcessPipe {
     if (!std::filesystem::exists(ServerFolder, ec)) {
       // Doesn't exist, create the the folder as a user convenience
       if (!std::filesystem::create_directories(ServerFolder, ec)) {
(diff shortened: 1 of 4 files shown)


Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -189,7 +189,7 @@ namespace ProcessPipe {
     // Create the initial unix socket
     ServerSocketFD = socket(AF_UNIX, SOCK_STREAM | SOCK_CLOEXEC, 0);
     if (ServerSocketFD == -1) {
-      LogMan::Msg::AFmt("Couldn't create AF_UNIX socket: %d %s\n", errno, strerror(errno));
+      LogMan::Msg::EFmt("Couldn't create AF_UNIX socket: %d %s\n", errno, strerror(errno));
```
lioncash, line 192: Seems like this and the format message below is using printf specifiers instead of fmt specifiers

Sonicadvance1: oops, these weren't ever converted.

Sonicadvance1: Resolved.

lioncash (approved):

PR #2157, opened 2022-11-17
FEXServer: More Systemd fixes

Sonicadvance1:
Two changes here.

- Make the mount path follow server temp folder requirements.
  - Will be mounted in `/tmp/` or `$XDG_RUNTIME_DIR/` now
- Switch the FEXServer socket to an "abstract" AF_UNIX socket.
  - If the socket is in `/tmp/` then systemd will put the service in a private `/tmp` folder that only exists for the service.
  - If the socket is in `$XDG_RUNTIME_DIR` then pressure-vessel can't chroot anymore since they make their own runtime directory.
  - If it is in `$HOME/.fex-emu/` then it breaks usage where the filesystem is a mount that doesn't support AF_UNIX like sshfs.

The only reasonable thing to do is to switch over to `abstract` sockets which will work in all cases.
Tested with pressure-vessel and systemd and now it works in all situations.

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -100,21 +100,29 @@ namespace FEXServerClient {
     return GetServerLockFolder() + "RootFS.lock";
   }
 
-  std::string GetServerSocketFile() {
-    FEX_CONFIG_OPT(ServerSocketPath, SERVERSOCKETPATH);
-    if (ServerSocketPath().empty()) {
-      auto TmpDir = std::filesystem::temp_directory_path().string();
-
-      auto XDGRuntimeEnv = getenv("XDG_RUNTIME_DIR");
-      if (XDGRuntimeEnv) {
-        // If the XDG runtime directory works then use that instead.
-        TmpDir = XDGRuntimeEnv;
-      }
-
-      return fmt::format("{}/{}.FEXServer.socket", TmpDir, ::geteuid());
+  std::string GetServerTempFolder() {
+    // We need a server temporary folder that has the following requirements.
+    // - Can't be `/tmp/`
+    //   - systemd services use `PrivateTmp` feature to gives services their own tmp.
+    // - Can't be `$HOME/.fex-emu/`
+    //   - Can be mounted with a filesystem (sshfs) which can't handle mount points inside it.
+    auto Folder = std::filesystem::temp_directory_path().string();
+    auto XDGRuntimeEnv = getenv("XDG_RUNTIME_DIR");
+    if (XDGRuntimeEnv) {
+      // If the XDG runtime directory works then use that instead.
+      Folder = XDGRuntimeEnv;
     }
 
-    return ServerSocketPath;
+    if (FEXCore::Config::FindContainer() == "pressure-vessel") {
+      // In pressure-vessel the server location changes.
+      Folder = "/run/host/" + Folder;
+    }
+
+    return Folder;
+  }
+
+  std::string GetServerSocketName() {
+    return fmt::format("{}.FEXServer.Socket", ::geteuid());
   }
 
   int GetServerFD() {
@@ -122,7 +130,7 @@ namespace FEXServerClient {
   }
 
   int ConnectToServer() {
-    auto ServerSocketFile = GetServerSocketFile();
+    auto ServerSocketName = GetServerSocketName();
 
     // Create the initial unix socket
     int SocketFD = socket(AF_UNIX, SOCK_STREAM, 0);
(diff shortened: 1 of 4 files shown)


Source/Common/FEXServerClient.cpp
```diff
@@ -100,41 +100,56 @@ namespace FEXServerClient {
...
-      auto TmpDir = std::filesystem::temp_directory_path().string();
-
-      auto XDGRuntimeEnv = getenv("XDG_RUNTIME_DIR");
-      if (XDGRuntimeEnv) {
-        // If the XDG runtime directory works then use that instead.
-        TmpDir = XDGRuntimeEnv;
-      }
+  std::string GetServerTempFolder() {
+    // We need a server temporary folder that has the following requirements.
+    // - Can't be `/tmp/`
```
neobrain, line 105: If it can't be that, why do we proceed to call `temp_directory_path`? IIUC the comment line only applies to FEXServer running as a systemd service, but that's not evident from the comment.

Source/Common/FEXServerClient.cpp
```diff
@@ -100,41 +100,56 @@ namespace FEXServerClient {
...
+    auto Folder = std::filesystem::temp_directory_path().string();
+    auto XDGRuntimeEnv = getenv("XDG_RUNTIME_DIR");
+    if (XDGRuntimeEnv) {
+      // If the XDG runtime directory works then use that instead.
+      Folder = XDGRuntimeEnv;
+    }
 
-      return fmt::format("{}/{}.FEXServer.socket", TmpDir, ::geteuid());
+    if (FEXCore::Config::FindContainer() == "pressure-vessel") {
+      // In pressure-vessel the server location changes.
```
neobrain, line 117: Why? It's not apparent where this change comes from for someone without knowledge how pressure-vessel works. Is it an expectation of pressure-vessel? Or does it change the runtime environment somehow to make the other paths inaccessible?

Source/Common/FEXServerClient.cpp
```diff
@@ -100,41 +100,56 @@ namespace FEXServerClient {
...
 
+    // AF_UNIX has a special feature for named socket paths.
+    // If the name of the socket begins with `\0` then it is an "abstract" socket address.
+    // The entirety of the name is used as a path to a socket that doesn't have any filesystem backing.
     struct sockaddr_un addr{};
     addr.sun_family = AF_UNIX;
-    strncpy(addr.sun_path, ServerSocketFile.data(), std::min(ServerSocketFile.size(), sizeof(addr.sun_path)));
+    // + 1 for null character initializer.
+    size_t SizeOfSocketString = std::min(ServerSocketName.size(), sizeof(addr.sun_path) - 2);
+    strncpy(addr.sun_path + 1, ServerSocketName.data(), SizeOfSocketString);
```
neobrain, line 148: Would suggest an explicit `addr.sun_path[0]=0` before this line, and then remove the `+1 for null character initializer` comment (I found that comment more confusing since I thought you were just talking about terminators twice)

Source/Common/FEXServerClient.cpp
```diff
@@ -100,41 +100,56 @@ namespace FEXServerClient {
...
     }
 
+    // AF_UNIX has a special feature for named socket paths.
+    // If the name of the socket begins with `\0` then it is an "abstract" socket address.
+    // The entirety of the name is used as a path to a socket that doesn't have any filesystem backing.
     struct sockaddr_un addr{};
     addr.sun_family = AF_UNIX;
-    strncpy(addr.sun_path, ServerSocketFile.data(), std::min(ServerSocketFile.size(), sizeof(addr.sun_path)));
+    // + 1 for null character initializer.
+    size_t SizeOfSocketString = std::min(ServerSocketName.size(), sizeof(addr.sun_path) - 2);
```
neobrain, line 147: Logic in these three lines would be more accessible if we include the null-terminator in the size:
```cpp
size_t SizeOfSocketString = std::min(ServerSocketName.size() + 1, sizeof(addr.sun_path) - 1)
```
(Saves a couple of mental constant folding steps, such as to figure out where the `-2` came from)

Source/Common/FEXServerClient.cpp
Sonicadvance1: Cleared up the comment a bit and changed around the ordering of the code to more correctly reflect the fact that `/tmp/` is only used as a fallback path.

Source/Common/FEXServerClient.cpp
Sonicadvance1: Cleared up the comment a bit here.
Yes, pressure-vessel if changing the runtime environment to make the paths inaccessible.
Pressure-vessel is a fancy chroot environment that does a whole bunch of directory rebinding which means we need to tip-toe around it a bit to be correct with our implementations.
`$XDG_RUNTIME_DIR` specifically disappears from our original view once it does a pivot_root.

Source/Common/FEXServerClient.cpp
Sonicadvance1: Did so.

Source/Common/FEXServerClient.cpp
Sonicadvance1: You're right. I did it in this bit of an obtuse way because I wasn't sure if C++ strings are guaranteed null-terminated.
Looked it up and with latest C++ specs it is guaranteed. A lot cleaner now.

neobrain (approved):

PR #2160, opened 2022-11-21
Arm64/VectorOps: Simplify VMov IR op on SVE 

lioncash:
Initially I put this in very conservatively to make sure we always clear out the upper lanes, but since Adv. SIMD operations have zero-extending behavior when storing results, we can just use a few operations as is without needing to unnecessarily do the same work twice.


Sonicadvance1 (approved):
Makes sense

PR #2167, opened 2022-11-22
Arm64: Optimize Break IR op codegen

Sonicadvance1:
This isn't really a performance issue, more just something that is gross looking at while looking at unit tests.

Break /usually/ isn't abused heavily by games (Except Denuvo) so not really a perf concern regardless.

### DIFF
--- a/External/FEXCore/include/FEXCore/Core/CoreState.h
+++ b/External/FEXCore/include/FEXCore/Core/CoreState.h
@@ -218,12 +218,13 @@ namespace FEXCore::Core {
 
     uint32_t SignalHandlerRefCounter{};
 
-    struct SynchronousFaultDataStruct {
+    struct alignas(8) SynchronousFaultDataStruct {
       bool FaultToTopAndGeneratedException{};
       uint8_t Signal;
-      uint32_t TrapNo;
-      uint32_t err_code;
-      uint32_t si_code;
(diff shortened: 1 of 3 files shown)


External/FEXCore/include/FEXCore/Core/CoreState.h
```diff
@@ -237,6 +238,9 @@ namespace FEXCore::Core {
   static_assert(offsetof(CpuStateFrame, Pointers) + sizeof(CpuStateFrame::Pointers) <= 32760, "JITPointers maximum pointer needs to be less than architecture maximum 32768");
 
   static_assert(std::is_standard_layout<CpuStateFrame>::value, "This needs to be standard layout");
+  static_assert(sizeof(CpuStateFrame::SynchronousFaultData) == 8, "This needs to by 8 bytes");
+  static_assert(std::alignment_of_v<CpuStateFrame::SynchronousFaultDataStruct> == 8, "This needs to by 8 bytes");
```
neobrain, line 242: typos

neobrain (approved):

External/FEXCore/include/FEXCore/Core/CoreState.h
Sonicadvance1: Resolved.

PR #2170, opened 2022-11-22
OpcodeDispatcher: Moves all GPR and XMM accesses to direct register accesses

Sonicadvance1:
This is a bit of a large commit since it is an all or nothing sort of
change.

Instead of doing LoadContext and StoreContext for GPRs and FPRs, any
that are statically allocated (GPR and XMM) should use
{Load,Store}Register directly.

This is now enforced that {Load,Store}Context{,Indexed} will assert if
trying to access these ranges.

This is the first step towards accelerating our JIT compile times in
that it removes the need for the SRA pass to convert all the IR over.

This ensures that from the Dispatcher directly we are handling SRA
accesses as "registers", ensuring that future changes don't need to run
in to this problem.

Even though it wasn't a target of this change, in a simple test
application this removes ~12% of the total JIT compile time.

Interpreter and x86 JIT will treat the {Load,Store}Register as regularly register allocated operations, while the ARM JIT will handle it as true SRA, which makes it easy to support.


Sonicadvance1:
Found out why the ARMv8.0 runner was crashing. Our LoadRegister IR op uses half-float fmov unconditionally, which isn't supported on that hardware.
Ideally we'll move {Load,Store}Register to only ever operating at 128-bit/256-bit sizes in the future, but we aren't there yet.

Sonicadvance1:
Did a wackload of game testing today. So even though this looks like a big scary change, I haven't picked up a compatibility regression in my suite of games.

lioncash (approved):
Did a run-down through everything and this looks very nice. Really dig that we don't need to do long offsetof indexing anymore in most cases

PR #2178, opened 2022-11-27
Arm64: Const on unmodified argument

Sonicadvance1:
Just noticed this when tinkering around the JIT.
These arguments can safely be const.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/AtomicOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/AtomicOps.cpp
@@ -10,7 +10,7 @@ $end_info$
 namespace FEXCore::CPU {
 using namespace vixl;
 using namespace vixl::aarch64;
-#define DEF_OP(x) void Arm64JITCore::Op_##x(IR::IROp_Header *IROp, IR::NodeID Node)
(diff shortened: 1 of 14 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/AtomicOps.cpp
```diff
@@ -10,7 +10,7 @@ tags: backend|arm64
 namespace FEXCore::CPU {
 using namespace vixl;
 using namespace vixl::aarch64;
-#define DEF_OP(x) void Arm64JITCore::Op_##x(IR::IROp_Header *IROp, IR::NodeID Node)
+#define DEF_OP(x) void Arm64JITCore::Op_##x(IR:: IROp_Header const *IROp, IR::NodeID Node)
```
lioncash, line 13: In a few of these, it seems like a space was added after the `IR::` qualifier

Sonicadvance1: Weird, not sure how that got there, will fix.

Sonicadvance1: Resolved.

lioncash (approved):

PR #2180, opened 2022-11-27
FEXLoader: Disables some AOT shutdown overhead when not enabled

Sonicadvance1:
When AOT wasn't enabled it was still doing some accesses to the filesystem on shutdown.

Slightly improves shutdown time.

### DIFF
--- a/Source/Tests/FEXLoader.cpp
+++ b/Source/Tests/FEXLoader.cpp
@@ -428,37 +428,38 @@ int main(int argc, char **argv, char **const envp) {
     });
   }
 
-  if (AOTIRLoad() || AOTIRCapture() || AOTIRGenerate()) {
+  const bool AOTEnabled = AOTIRLoad() || AOTIRCapture() || AOTIRGenerate();
+  if (AOTEnabled) {
     LogMan::Msg::IFmt("Warning: AOTIR is experimental, and might lead to crashes. "
                       "Capture doesn't work with programs that fork.");
+
+    FEXCore::Context::SetAOTIRLoader(CTX, [](const std::string &fileid) -> int {
+      auto filepath = std::filesystem::path(FEXCore::Config::GetDataDirectory()) / "aotir" / (fileid + ".aotir");
+
+      return open(filepath.c_str(), O_RDONLY);
+    });
+
+    FEXCore::Context::SetAOTIRWriter(CTX, [](const std::string& fileid) -> std::unique_ptr<std::ofstream> {
+      auto filepath = std::filesystem::path(FEXCore::Config::GetDataDirectory()) / "aotir" / (fileid + ".aotir.tmp");
+      auto AOTWrite = std::make_unique<std::ofstream>(filepath, std::ios::out | std::ios::binary);
+      if (*AOTWrite) {
+        std::filesystem::resize_file(filepath, 0);
+        AOTWrite->seekp(0);
+        LogMan::Msg::IFmt("AOTIR: Storing {}", fileid);
+      } else {
+        LogMan::Msg::IFmt("AOTIR: Failed to store {}", fileid);
+      }
+      return AOTWrite;
+    });
+
+    FEXCore::Context::SetAOTIRRenamer(CTX, [](const std::string& fileid) -> void {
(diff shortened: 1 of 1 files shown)


Source/Tests/FEXLoader.cpp
```diff
@@ -428,37 +428,41 @@ int main(int argc, char **argv, char **const envp) {
...
 
-    // Rename the temporary file to atomically update the file
-    std::filesystem::rename(TmpFilepath, NewFilepath);
-  });
+      // Rename the temporary file to atomically update the file
+      std::filesystem::rename(TmpFilepath, NewFilepath);
+    });
+  }
 
   if (AOTIRGenerate()) {
```
neobrain, line 467: Side note: This will call AOTIRGenerate a second time if `!AOTIRLoad() && !AOTIRCapture()`. Is this intended or should we drop one of the calls?

Source/Tests/FEXLoader.cpp
```diff
@@ -428,37 +428,41 @@ int main(int argc, char **argv, char **const envp) {
...
-    } else {
-      LogMan::Msg::IFmt("AOTIR: Failed to store {}", fileid);
-    }
-    return AOTWrite;
-  });
 
-  FEXCore::Context::SetAOTIRRenamer(CTX, [](const std::string& fileid) -> void {
-    auto TmpFilepath = std::filesystem::path(FEXCore::Config::GetDataDirectory()) / "aotir" / (fileid + ".aotir.tmp");
-    auto NewFilepath = std::filesystem::path(FEXCore::Config::GetDataDirectory()) / "aotir" / (fileid + ".aotir");
+  if (AOTEnabled) {
```
neobrain, line 438: Can be merged into the previous if-block.

Source/Tests/FEXLoader.cpp
```diff
@@ -468,21 +472,23 @@ int main(int argc, char **argv, char **const envp) {
...
-    FEXCore::Context::WriteFilesWithCode(CTX, [](const std::string& fileid, const std::string& filename) {
-      auto filepath = std::filesystem::path(FEXCore::Config::GetDataDirectory()) / "aotir" / (fileid + ".path");
-      int fd = open(filepath.c_str(), O_CREAT | O_EXCL | O_WRONLY, 0644);
-      if (fd != -1) {
-        write(fd, filename.c_str(), filename.size());
-        close(fd);
-      }
-    });
-  }
+  if (AOTEnabled) {
```
neobrain, line 475: Can likely be merged into the previous block. That's because the current structure is:
```cpp
if (AOTEnabled) { ... }
if (AOTIRGenerate()) { ... }
if (AOTEnabled) { ... }
```
but I think that is equivalent to:
```cpp
if (AOTEnabled) {
  ...
  if (AOTIRGenerate()) { ... }
  ...
}
```
since if AOTEnabled is false, we'd just be calling AOTIRGenerate a second time without having modified any AOTIR configuration.

neobrain (approved):

Source/Tests/FEXLoader.cpp
Sonicadvance1: This is intended because the config option `AOTIRGenerate` has different behaviour while relying on the previous `AOTIRLoader` `AOTIRWriter` and `AOTIRRenamer` need to be set up.

If the `AOTIRGenerate` config option is enabled then it statically analyzes the ELF passed in and AOT compiles it, skipping execution entirely.

Source/Tests/FEXLoader.cpp
Sonicadvance1: This would be ugly to merge because of the if statement above it.

This tail if statement needs to be run after `RunUntilExit` as run or `AOTGenSection` has run to write the generated IR to disk.

Source/Tests/FEXLoader.cpp
Sonicadvance1: You're right, merged.

Source/Tests/FEXLoader.cpp
neobrain: Ah, true - I missed the "else" branch in the middle block.

PR #2181, opened 2022-11-27
EmulatedFiles: Defer cpuinfo file initialization to first access

Sonicadvance1:
This improves startup time by a couple of milliseconds.

Most applications don't query cpuinfo so deferring improves most application's startup times.


Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
```diff
@@ -622,6 +622,16 @@ namespace FEX::EmulatedFile {
...
+      // Check if deferred cpuinfo initialization has occured.
+      if (!cpu_info_initialized) {
+        // Only allow a single thread to initialize the cpu_info.
+        // Just in-case multiple threads try to initialize at once.
+        std::unique_lock lk{cpu_info_intialization_lock};
+        if (!cpu_info_initialized) {
+          cpu_info = GenerateCPUInfo(ctx, ThreadsConfig());
+          cpu_info_initialized = true;
+        }
+      }
```
neobrain, line 634: We can save on some state here by using `call_once` instead:

```suggestion
      // Only allow a single thread to initialize the cpu_info.
      // Just in-case multiple threads try to initialize at once.
      std::once_flag cpu_info_initialized; // in EmulatedFiles.h
      std::call_once(cpu_info_initialized, []() { cpu_info = GenerateCPUInfo(ctx, ThreadsConfig()); }
```

(Fwiw this should not [have any effect on performance](https://sourceware.org/git/?p=glibc.git;a=blob;f=nptl/pthread_once.c;h=beaeb79ef8afe76af809d01e45851300a391f627;hb=HEAD#l135))

Sonicadvance1: You're right, I completely forgot that `std::call_once` was a thing.
Resolved.

lioncash (approved):

PR #2193, opened 2022-12-05
IoctlEmu: Support radeon

Sonicadvance1:


Sonicadvance1:
Tested by user and seems to work.

PR #2194, opened 2022-12-05
FEXServerClient: Disable confusing connection log

Sonicadvance1:
On first FEXInterpreter execution, it is expected that `ConnectToServer` will fail with `ECONNREFUSED` because FEXServer won't be running.

Skip printing this first messaage to stderr if configured. If it is some other error message then ensure it is still printed.

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -147,7 +147,7 @@ namespace FEXServerClient {
     return ServerFD;
   }
 
-  int ConnectToServer() {
+  int ConnectToServer(ConnectionOption ConnectionOption) {
     auto ServerSocketName = GetServerSocketName();
 
     // Create the initial unix socket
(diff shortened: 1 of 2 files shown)


Source/Common/FEXServerClient.cpp
```diff
@@ -196,7 +198,7 @@ namespace FEXServerClient {
   }
 
   int ConnectToAndStartServer(char *InterpreterPath) {
-    int ServerFD = ConnectToServer();
+    int ServerFD = ConnectToServer(false);
```
lioncash, line 201: hmm, maybe have a wrapper function like:

```cpp
// Insert appropriate name here
int ConnectToServerButBeCoolAboutIt() {
    return ConnectToServer(false);
}
```

or something, or maybe like:

```cpp
enum class ConnectOption {
    Default,
    NoPrintConnectionError,
};

int ConnectToServer(ConnectionOption option) {
  // ...
}
```

since `ConnectToServer(false)` and ConnectToServer(true)` seem a little opaque in terms of what the bool is doing

Sonicadvance1: I like the enum approach. Switched over to that.

lioncash (approved):

PR #2195, opened 2022-12-05
FEXLoader: Make `IsInterpreterInstalled` check less horrible.

Sonicadvance1:
`std::filesystem::exists` is particularly gnarly in how it checks to see if the file exists.
It allocates memory, it creates lists, it splits things, then eventually checking the status

Remove all this overhead to help out minorly for applications that execve a lot.

Before:
Image_2022-12-05_02-57-28

After:
Image_2022-12-05_02-57-33


lioncash:
Wow, that *is* pretty gross

PR #2198, opened 2022-12-05
OpcodeDispatcher: Handle VADDPD/VADDPS/VPADDB/VPADDW/VPADDD/VPADDQ

lioncash:
Gets all of the basic addition instructions out of the way


Sonicadvance1 (approved):
nice nice nice.

PR #2200, opened 2022-12-05
OpcodeDispatcher: Simplify VANDN

lioncash:
We can just use VBic here to simplify everything.


Sonicadvance1 (approved):
Should be able to use VBic with the ANDNOp above this as well.

Sonicadvance1 (approved):

PR #2207, opened 2022-12-07
OpcodeDispatcher: Handle VADDSD/VADDSS

lioncash:
First AVX scalar ops and rounds out the remaining basic add operations


Sonicadvance1 (approved):
nice nice nice.

PR #2208, opened 2022-12-07
OpcodeDispatcher: Explicitly zero upper lanes

lioncash:
Makes our intent to zero-extend the upper lanes explicit and lets us remove a special case in the LoadSource implementation.

This also makes things a little nicer since we're not hardcoding 32 byte stores.


Sonicadvance1 (commented):
Thanks for picking this up.
One thing that I see from this is that if the host doesn't support 256-bit wide vector registers then we will now be generating extraneous moves, where the special case handler in StoreResult wouldn't have.

We could probably use a little helper implementation in the OpDispatcher that does the `VMov(16, Src)` if the host supports AVX, or just returns the value if it doesn't.

Either checking `CTX->HostFeatures.SupportsAVX` on each call or using a function member pointer to special case one or the other.

lioncash:
How would extraneous moves be occurring? Our AVX impl requires 256-bit host vector registers and that's where all the explicit clearing/moves mainly take place (unless I'm being silly and missing some other case)

Sonicadvance1 (approved):

PR #2215, opened 2022-12-08
OpcodeDispatcher: Handle VBROADCASTSD/VBROADCASTSD/VBROADCASTF128

lioncash:


unittests/ASM/VEX/vbroadcastsd.asm
```diff
@@ -0,0 +1,40 @@
...
+%endif
+
+lea rdx, [rel .data]
+
+vmovaps ymm0, [rdx + 32 * 0]
+vmovaps ymm1, [rdx + 32 * 1]
+
+; Register broadcasting
+vbroadcastsd ymm2, xmm0
+vbroadcastsd ymm3, xmm1
```
Sonicadvance1, line 22: Was this and the following `ymm5` usage line meant to be xmm usage instead to ensure zero extend of 128-bit op like the vbroadcastss test?

lioncash: Nah. `VBROADCASTSD` doesn't have a variant with an xmm destination.

Sonicadvance1: Oh fun

Sonicadvance1 (approved):

PR #2216, opened 2022-12-09
Initial 32-bit host thunk feature support

Sonicadvance1:
Not wired up to anything yet, but is preparation work.

To test, one can modify the `ThunkLibs/GuestLibs/CMakeLists.txt` file as follows

```diff
diff --git a/ThunkLibs/GuestLibs/CMakeLists.txt b/ThunkLibs/GuestLibs/CMakeLists.txt
index 1620a6134..5506d890f 100644
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -152,6 +152,10 @@ function(add_guest_lib NAME SONAME)
   endif()
 endfunction()

+# 32-bit and 64-bit supported
+generate(libGL ${CMAKE_CURRENT_SOURCE_DIR}/../libGL/libGL_interface.cpp)
+add_guest_lib(GL "libGL.so.1")
+
 # These thunks only support 64-bit
 if (BITNESS EQUAL 64)
   #add_guest_lib(fex_malloc_loader)
@@ -166,9 +170,6 @@ if (BITNESS EQUAL 64)
   generate(libEGL ${CMAKE_CURRENT_SOURCE_DIR}/../libEGL/libEGL_interface.cpp)
   add_guest_lib(EGL "libEGL.so.1")

-  generate(libGL ${CMAKE_CURRENT_SOURCE_DIR}/../libGL/libGL_interface.cpp)
-  add_guest_lib(GL "libGL.so.1")
-
   # libGL must pull in libX11.so, so generate a placeholder libX11.so to link against
   add_library(X11 SHARED ../libX11/libX11_NativeGuest.cpp)
   target_link_libraries(GL-guest PRIVATE X11)
```

Did a minor test and it at least allowed some function calls correctly.


ThunkLibs/HostLibs/CMakeLists.txt
```diff
@@ -15,27 +15,27 @@ endif()
 # - custom command: Main build step that runs the thunk generator on the given interface definition
 # - libxyz-interface: Target for IDE integration (making sure libxyz-interface.cpp shows up as a source file in the project tree)
 # - libxyz-deps: Interface target to read include directories from which are passed to libclang when parsing the interface definition
-function(generate NAME SOURCE_FILE)
+function(generate NAME SOURCE_FILE GUEST_BITNESS)
```
neobrain, line 18: We don't need to compile interface definitions per-bitness, do we?

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -225,7 +225,9 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
 
   bool LoadedThunkDatabase{};
   auto ThunkConfigFile = ThunkConfig();
-  auto ThunkGuestPath = std::filesystem::path(ThunkGuestLibs());
+  auto ThunkGuestPath = Is64BitMode() ?
+    std::filesystem::path(ThunkGuestLibs()) :
+    std::filesystem::path(ThunkGuestLibs32());
```
neobrain, line 230:
```suggestion
  auto ThunkGuestPath = std::filesystem::path { Is64BitMode() ? ThunkGuestLibs() : ThunkGuestLibs32() };
```

ThunkLibs/HostLibs/CMakeLists.txt
Sonicadvance1: I kept it in there just in-case we do need it for something. Not sure if it will matter long-term.

ThunkLibs/include/common/Guest.h
```diff
@@ -121,6 +122,14 @@ inline Result CallHostFunction(Args... args) {
   // This magic incantation also works in that instance so this is about the best we can do without adding a new attribute to clang for modifying the
   // ABI.
   asm volatile("" : "=r" (host_addr));
+#else
+  // mm0 register is chosen as the register that contains the host_addr.
+  // This register is chosen as it doesn't conflict with vectorcall.
+  // This register also overlaps the x87 st(0) register which is used as float value return.
+  // st(0) is required to be empty coming in to the function because of this.
+  register uintptr_t host_addr asm ("mm0");
```
neobrain, line 130: Has this been tested, or should I hack up some unit tests to verify this produces working assembly?

ThunkLibs/include/common/Guest.h
```diff
@@ -121,6 +122,14 @@ inline Result CallHostFunction(Args... args) {
   // This magic incantation also works in that instance so this is about the best we can do without adding a new attribute to clang for modifying the
   // ABI.
   asm volatile("" : "=r" (host_addr));
+#else
+  // mm0 register is chosen as the register that contains the host_addr.
+  // This register is chosen as it doesn't conflict with vectorcall.
+  // This register also overlaps the x87 st(0) register which is used as float value return.
```
neobrain, line 128: What does this mean in practice? Do function pointers returning a `float` not work?

neobrain (commented):
Tested against my toy thunk library, works great. Thanks for taking up the work on this :)

Some reservations about function pointer support, but looks good otherwise.

ThunkLibs/include/common/Guest.h
Sonicadvance1: This has been tested to get through at least one function call correctly, and inspected the output assembly to ensure something wasn't breaking it.
glXGetVisual or something was working.

ThunkLibs/include/common/Guest.h
Sonicadvance1: In practice this works because we only care about mm0 coming in to the function, on return mm0/st(0) will contain the resulting float/double, overwriting whatever previously was there, which we don't care about at that point.

Source/Tests/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Good idea.

ThunkLibs/include/common/Guest.h
neobrain: How does the sentence after this (`st(0) is required to be empty coming in to the function because of this`) factor in here? Is that a precondition we can guarantee or do we rely on the guest application to be well-behaving in this regard?

Sonicadvance1: If the application is badly behaving here then we will overwrite what is in st(0) by the thunk instruction storing in to mm0, clobbering the data. Which for 32-bit applications you're sort of expecting libraries like GL/Vulkan to stomp on any x87 data anyway. Nothing will survive if it lived in the stack.

Source/Tests/LinuxSyscalls/FileManagement.cpp

ThunkLibs/include/common/Guest.h
neobrain: Alright, makes sense. May I suggest this rephrase?
```
// Use mm0 to pass in host_addr (chosen to avoid conflicts with vectorcall).
// Note this register overlaps the x87 st(0) register (used to return float values),
// so applications that expect this register to be preserved could run into problems.
```
Contains the same information AFAICT but makes it a bit clearer why these things are mentioned in the first place.

neobrain (approved):

ThunkLibs/include/common/Guest.h
Sonicadvance1: Updated comment.

PR #2218, opened 2022-12-09
OpCodeDispatcher: Optimize a case of GOT calculation

Sonicadvance1:
32-bit GOT calculation needs to do a call+pop to do get the EIP on 32-bit. LEA doesn't work because it there is no EIP relative ops like on x86-64.

This causes a terrible block split on every GOT calculation without the optimization in place.

Now the block can continue through this weird GOT calculation.

This will be worthwhile for our 32-bit thunks where for some reason the GOT calculation can't be removed. The GOT is calculated even though it isn't used.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Frontend.h
+++ b/External/FEXCore/Source/Interface/Core/Frontend.h
@@ -58,6 +58,7 @@ private:
   bool DecodeInstruction(uint64_t PC);
 
   void BranchTargetInMultiblockRange();
+  bool BranchTargetCanContinue(bool FinalInstruction) const;
 
   uint8_t ReadByte();
   uint8_t PeekByte(uint8_t Offset) const;
--- a/unittests/32Bit_ASM/FEX_bugs/GOT_calculation.asm
+++ b/unittests/32Bit_ASM/FEX_bugs/GOT_calculation.asm
@@ -1,0 +1,20 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x10011"
+  },
+  "Mode": "32BIT"
+}
+%endif
+
+mov esp, 0xe0000010
+
+; This is a common pattern in 32-bit PIE code.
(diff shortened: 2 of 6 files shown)


External/FEXCore/Source/Interface/Core/Frontend.h
```diff
@@ -58,6 +58,7 @@ class Decoder final {
   bool DecodeInstruction(uint64_t PC);
 
   void BranchTargetInMultiblockRange();
+  bool BranchTargetCanContinue(bool FinalInstruction);
```
lioncash, line 61:
```suggestion
  bool BranchTargetCanContinue(bool FinalInstruction) const;
```

This seems like it can be const, though disregard if it turns out to not be the case.

lioncash (commented):
Seems good. Also one nit

External/FEXCore/Source/Interface/Core/Frontend.h
Sonicadvance1: You're right, resolved.

unittests/32Bit_ASM/FEX_bugs/GOT_calculation.asm
```diff
@@ -0,0 +1,19 @@
...
+  },
+  "Mode": "32BIT"
+}
+%endif
+
+mov esp, 0xe0000010
+
+; This is a common pattern in 32-bit PIE code.
+; It uses a call + pop to calculate the GOT
+; Don't ask me why the compiler doesn't just use lea.
```
JosJuice, line 14: Isn't this for the reason mentioned in the PR description?

Sonicadvance1: oop, old comment while I was investigating. Indeed.

Sonicadvance1: Comment updated.

lioncash (approved):

PR #2237, opened 2022-12-13
OpcodeDispatcher: Handle VHADDP{D, S}

lioncash:


External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -990,6 +990,30 @@ void OpDispatchBuilder::VANDNOp(OpcodeArgs) {
...
+template <size_t ElementSize>
+void OpDispatchBuilder::VHADDPOp(OpcodeArgs) {
+  const auto SrcSize = GetSrcSize(Op);
+  const auto Is256Bit = SrcSize == Core::CPUState::XMM_AVX_REG_SIZE;
+
+  OrderedNode *Src1 = LoadSource(FPRClass, Op, Op->Src[0], Op->Flags, -1);
+  OrderedNode *Src2 = LoadSource(FPRClass, Op, Op->Src[1], Op->Flags, -1);
+
+  OrderedNode *Res = _VFAddP(SrcSize, ElementSize, Src1, Src2);
+  OrderedNode *Dest = _VMov(SrcSize, Res);
```
Sonicadvance1, line 1002: For the 256-bit case is this VMov necessary instead of using `Res` directly in the first SSA argument in the first `VInsElement` below?
Then only do VMov on the 128-bit case for the zero extension?

lioncash: Oh, good catch. Will fix that

Sonicadvance1 (approved):

PR #2238, opened 2022-12-14
OpcodeDispatcher: Handle VPERMQ/VPERMPD

lioncash:
First two AVX+-only ops


Sonicadvance1 (approved):
ooo, spicy, AVX2 instructions.
Also, didn't know nasm supported binary literals, cute.

PR #2239, opened 2022-12-14
Frontend: Handle 256-bit destination sizes directly

lioncash:
Previously the only time we'd promote to a 256-bit size is if the VEX.L bit was set in the 128-bit path.

This allows setting the 256-bit size directly in the tables if we need to.


Sonicadvance1 (approved):
`If VPERMPQ is encoded with VEX.L= 0 or EVEX.128, an attempt to execute the instruction will cause an #UD exception.`
Very cute. Makes sense to extend support to 256-bit explicit sized ops on destination.
I see we already supported it on sources.

PR #2243, opened 2022-12-14
IR: Remove HasDest and NumArgs from IR ops.

Sonicadvance1:
Shaves 8-bits off of each IR op.
Reduces what is necessary to generate an IR op by no longer needing to set NumArgs and HasDest bits.

No need to cart around this data when it is an IR op constant for each op, especially since most optimization passes don't need the data anyway.


Sonicadvance1:
This is infinite looping in the RA pass, will need to see why behaviour changed.

Sonicadvance1:
Supersceded by #2319 and #2320

PR #2244, opened 2022-12-14
ARM64: Moves RA functions to header

Sonicadvance1:
These are just some basic address calculations and a load, we want these to be inlined as much as possible.


External/FEXCore/Source/Interface/Core/JIT/Arm64/JITClass.h
```diff
@@ -89,25 +89,89 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter  {
...
   [[nodiscard]] aarch64::Register GetReg(IR::NodeID Node) const;
 
+  template<uint8_t RAType>
+  [[nodiscard]] std::pair<aarch64::Register, aarch64::Register> GetSrcPair(IR::NodeID Node) const;
+
+  [[nodiscard]] FEXCore::IR::RegisterClassType GetRegClass(IR::NodeID Node) const;
+
   template<>
-  [[nodiscard]] aarch64::Register GetReg<RA_32>(IR::NodeID Node) const;
+  [[nodiscard]] aarch64::Register GetReg<Arm64JITCore::RA_32>(IR::NodeID Node) const {
```
lioncash, line 98: Seems like this version can call the RA_64 version and call `.W()` on the result. Not sure if that would be desirable though

Sonicadvance1: Could definitely do that. We are going to be paying the vixl cost of this `.W()` ending up in the stack regardless.

lioncash: yeah, was more just considering unifying the underlying behavior into one spot. Though either way is fine, considering how small it is

Sonicadvance1: Resolved.

lioncash (approved):

PR #2245, opened 2022-12-15
OpcodeDispatcher: Move template impl to regular function where applicable

lioncash:
Reduces the amount of code size generated by the specializations.

Only targets ones that are heavily templated like the generic op helper functions.


Sonicadvance1 (approved):
Nice nice nice nice.

PR #2247, opened 2022-12-15
OpcodeDispatcher: Narrow memory access with scalar rounding operations

lioncash:
These should only be accessing a 32-bit or 64-bit portion of memory depending on single or double precision variants are used. Previously we'd be doing a full 128-bit load.


Sonicadvance1 (approved):
Woops, good catch.

PR #2248, opened 2022-12-16
X86Tables: Restrict CVTDQ2PD and CVTTSD2SI to 64-bit memory accesses

lioncash:
Previously these were doing 128-bit accesses, when they only need to do 64-bit ones


Sonicadvance1 (approved):
Nice. It would probably behoove us at some point to create sub-register sized unit tests that push the memory src/dst up against the edge of a page boundary so ensure we aren't overstepping our access sizes.

PR #2257, opened 2022-12-16
OpcodeDispatcher: Handle immediate variants of VPSLL{D, Q, W}

lioncash:


Sonicadvance1 (approved):
Looks like github's action runner failed to send our runners the job again. Will need a quick force push to run it.

PR #2261, opened 2022-12-16
LookupCache: Use a PMR map for our Blocklinks with monotonic allocator

Sonicadvance1:
We generate a /lot/ of block links which causes our shutdown time to take a while for this map. For short running applications this shutdown time can take a statistically significant amount of time, and this has been haunting us for a while.

Using a monotonic buffer resource, we can cut the dellocation time down to effectively zero on cache clear and shutdown.
For short running applications this basically means we clear their shutdown time, for long running applications this means a cache clear is less painful.

I've measured up to 2ms , but more often this takes ~0.5ms before optimization. Now it is consistently ~0.25ms on a short running application.
Times will vary /greatly/ depending on how full it has filled this map.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/LookupCache.cpp
+++ b/External/FEXCore/Source/Interface/Core/LookupCache.cpp
@@ -18,6 +18,8 @@ LookupCache::LookupCache(FEXCore::Context::Context *CTX)
   : ctx {CTX} {
 
   TotalCacheSize = ctx->Config.VirtualMemSize / 4096 * 8 + CODE_SIZE + L1_SIZE;
+  // Setup our PMR map.
+  BlockLinks = BlockLinks_pma.new_object<BlockLinksMapType>();
 
   // Block cache ends up looking like this
   // PageMemoryMap[VirtualMemoryRegion >> 12]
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -239,7 +240,11 @@ class LookupCache {
   };
 
 
-  std::map<BlockLinkTag, std::function<void()>> BlockLinks;
+  std::pmr::monotonic_buffer_resource BlockLinks_mbr;
```
neobrain, line 243: Should have a small comment here to explain why the map itself is allocated through an allocator.

External/FEXCore/Source/Interface/Core/LookupCache.cpp
```diff
@@ -17,6 +17,11 @@ namespace FEXCore {
 LookupCache::LookupCache(FEXCore::Context::Context *CTX)
   : ctx {CTX} {
 
+  // Setup our PMR map.
+  // BlockLinks = BlockLinks_pma.new_object<BlockLinksMapType>(/*BlockLinks_pma*/); // Doesn't work with argument?
```
neobrain, line 21: `new_object` (or rather `polymorphic_allocator::construct`) is perhaps too smart for its own good: It automatically appends the allocator to the given argument list if the constructed type itself is allocator-aware! Hence in this commented-out line, the allocator argument ends up being passed twice to the constructors, which is why you were seeing a compile-error.

https://godbolt.org/z/5aPThb578

https://en.cppreference.com/w/cpp/memory/polymorphic_allocator/construct

> If the object is of type that itself uses allocators [...] passes this->resource() down to the constructed object.

neobrain (approved):
Nice use case for the monotonic buffer resource!

External/FEXCore/Source/Interface/Core/LookupCache.cpp
Sonicadvance1: Wow, I hate it. But at least I can understand it.
Shifted over to `new_object` now that this is understood.

External/FEXCore/Source/Interface/Core/LookupCache.h
Sonicadvance1: Added a comment.

PR #2273, opened 2022-12-20
OpcodeDispatcher: Handle 128-bit AVX AES instructions

lioncash:
Base AVX is only required to handle 128-bit operations. The VAES feature bit controls whether or not 256-bit operations are available.


Sonicadvance1 (approved):
Fancy

PR #2276, opened 2022-12-21
OpcodeDispatcher: Handle VPSLLDQ

lioncash:
x86 is a coward for not lane-crossing with this btw


Sonicadvance1 (approved):
Such coward, much amazing.

PR #2278, opened 2022-12-21
OpcodeDispatcher: Remove unnecessary usage of VMov in VPSLLDQOp

lioncash:
The extract operation essentially does this for us.


Sonicadvance1 (approved):
Nice

PR #2279, opened 2022-12-21
OpcodeDispatcher: Handle VPSRLDQ

lioncash:
Finishes off the last remaining op within the group table


Sonicadvance1 (approved):
\o/ Woo Groups down.

PR #2285, opened 2022-12-22
OpcodeDispatcher: Handle VPHMINPOSUW

lioncash:
Fairly simple. Same behavior as the SSE one, but we clear the upper lane


Sonicadvance1 (approved):
Love these instructions that the AVX implementers decided we don't need it to be 256-bit.

PR #2290, opened 2022-12-24
Create a new ARM64 Emitter and move JIT over to it.

Sonicadvance1:
While profiling the runtime code, vixl has been showing up as a reasonable amount of time in the JIT code due to how bloated the code emission is. Primarily it has a ton of conveniences in both the MacroAssembler and Assembler which makes the programmer's life easier. Most of this ends up in runtime code rather than compile time code which means it gets quite branchy and a just needs a lot of code to be run.

So let's switch over to our own custom ARM Emitter that is tailored towards FEX's performance needs. This emitter is written around the idea that we know the operation sizes at runtime and can use ternary operations to do the minor instruction modifications. This removes a lot of vixl's developer conveniences for the sake of SPEED. To the point that for some of the more complex IR ops and dispatcher we are now generating 128-bit loadstore operations when the code is being emitted. Compared to vixl almost always branching out and doing a single instruction at a time.

Some /very/ preliminary testing shows that this is removing ~27% CPU time from the code emission step of the ARM64 JIT. This is a better improvement than the original suspected 25% time consumed from vixl. Showing that vixl was poisoning cache, BTB, etc due to its large code footprint. Still need to dive in for some more detailed testing about how much of an improvement this is.

Remaining work:
ARMEmitter:
- [X] Implement remaining SVE operations - **Remainder to be done post-merge**
- [X] Implement remaining vector operations - **Remainder to be done post-merge**
- [X] Implement remaining scalar operations - **Remainder to be done post-merge**
- [X] Implement remaining misc operations - **Remainder to be done post-merge**
- [X] Fix ARMv8.0 atomics.
- [X] Clean up
- [X] Documentation
- [X] Basic level of testing

Testing:
- [X] Need to ensure all FEX tests pass.
- [X] Need to ensure Steam still runs
- [X] Need to test a bunch of games
- [X] Figure out why Steam is crashing early in startup.

This is going to be a nightmare to review I know.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Emitter.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Emitter.h
@@ -1,0 +1,721 @@
+#pragma once
+
+#include "Interface/Core/ArchHelpers/CodeEmitter/Buffer.h"
+#include "Interface/Core/ArchHelpers/CodeEmitter/Registers.h"
+
+#include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/Utils/EnumUtils.h>
+#include <FEXCore/Utils/LogManager.h>
+
+#include <aarch64/assembler-aarch64.h>
+
+#include <cstdint>
+#include <utility>
+#include <type_traits>
+#include <vector>
+
+/*
+ * Welcome to FEX-Emu's custom AArch64 emitter.
+ * This was written specifically to avoid the performance cost of the vixl emitter.
+ *
+ * There are some specific design constraints in this design to target a couple features:
+ *   - High performance
+ *   - Low CPU cache performance hit
+ *   - Significantly reduced code footprint
+ *   - Low number of branches
+ *
+ * These requirements are mostly achieved by removing a bunch of developer conveniences
+ * that vixl provides. The developer needs to take a lot of care to not shoot themselves in the foot.
+ *
+ * Misc design decisions:
+ * - Registers are encoded as basic uint32_t enums.
+ *   - Converting between different registers is zero-cost.
+ *   - Passing around as arguments are as cheap as registers
+ *     - Contrast to vixl where every register requires living on the stack.
+ *   - Registers can get encoded in to instructions with a simple `BFM` instruction.
+ *
+ * - Instructions are very simply emitted, allowing direct inlining most of the time.
+ *   - These are simple enough that multiple back-to-back instructions get optimized to 128-bit load-store operations.
+ *     - Contrast to vixl where pretty much no instruction emitter gets inlined.
+ *
+ * - Instruction emitters are /mostly/ unsized. Most instructions take a size argument first, which gets encoded
+ *   directly in to the instruction.
+ *   - Contrast to vixl where the register arguments are how the instructions determine operating size.
+ *   - Size argument allows FEX to use `CSEL` to select a size at runtime, instead of branching.
+ *   - Some instructions are explicitly sized based on register type. Read comments in the respective `inl` files to
+ *     see why.
+ *     Some scalar/vector operations are an example of this.
+ *
+ * - Almost zero helper functions.
+ *   - Primary exception to this rule is load-store operations. These will use a helper to make
+ *     it easier to select the correct load-store instruction. Mostly because these are a nightmare selecting
+ *     the right instruction.
+ */
+namespace FEXCore::ARMEmitter {
+  /*
+   * This `Size` enum is used for most ALU operations.
+   * These follow the AArch64 encoding style in most cases.
+   */
+  enum class Size : uint32_t {
+    i32Bit = 0,
+    i64Bit,
+  };
+
+  // This allows us to get the `Size` enum in bits.
+  template<Size size>
+  constexpr size_t RegSizeInBits() {
+    constexpr size_t RegSize[] = {
+      32, 64, 128,
+    };
+    return RegSize[FEXCore::ToUnderlying(size)];
+  }
+
+  [[maybe_unused]]
+  static inline size_t RegSizeInBits(Size size) {
+    constexpr size_t RegSize[] = {
+      32, 64, 128,
+    };
+    return RegSize[FEXCore::ToUnderlying(size)];
+  }
+
+  /* This `SubRegSize` enum is used for most ASIMD operations.
+   * These follow the AArch64 encoding style in most cases.
+   */
+  enum class SubRegSize : uint32_t {
+    i8Bit   = 0b00,
+    i16Bit  = 0b01,
+    i32Bit  = 0b10,
+    i64Bit  = 0b11,
+    i128Bit = 0b100,
+  };
+
+  // This allows us to get the `SubRegSize` in bits.
+  template<SubRegSize size>
+  constexpr size_t SubRegSizeInBits() {
+    return (1 << FEXCore::ToUnderlying(size)) * 8;
+  }
+
+  [[maybe_unused]]
+  static inline size_t SubRegSizeInBits(SubRegSize size) {
+    return (1 << FEXCore::ToUnderlying(size)) * 8;
+  }
+
+  /* This `ScalarRegSize` enum is used for most scalar float
+   * operations.
+   *
+   * This is specifically duplicated from `SubRegSize` to have strongly
+   * typed functions.
+   *
+   * `ScalarRegSize` specifically doesn't have `i128Bit` because scalar operations
+   * can't operate at 128-bit.
+   */
+  enum class ScalarRegSize : uint32_t {
+    i8Bit   = 0b00,
+    i16Bit  = 0b01,
+    i32Bit  = 0b10,
+    i64Bit  = 0b11,
+  };
+
+  // This allows us to get the `ScalarRegSize` in bits.
+  template<ScalarRegSize size>
+  constexpr size_t ScalarRegSizeInBits() {
+    return (1 << FEXCore::ToUnderlying(size)) * 8;
+  }
+
+  [[maybe_unused]]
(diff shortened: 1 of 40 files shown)


Sonicadvance1:
Found the issue with Steam, I was using shifted-register ALU encoding for spill slots. Which treats RSP as the zero register.
Oops, switched it over to extended-register and now everything is fine.

Sonicadvance1:
This is now getting ready for review.
I'll be doing a bunch of testing, benching, profiling, and documentation writing tomorrow.

Won't be getting 100% of instructions implemented in the emitter immediately, but the remaining SVE ops will be relatively high priority.
We have all of the operations that FEX currently uses, and I'll be spending a few hours each week implementing the rest of the operations.
But we can get this merged before then since the JIT won't be changing super quickly to require new ops quickly.

Sonicadvance1:
Profiling complete. This has better results than expected.
Measuring JIT backend emission time only this cuts emission time by 31.3% on Cortex-X1. On Cortex-A55 it cuts emission time by 39.7%.
Total JIT time as measured by full `CompileBlock` time program execution. 7.1% reduction on Cortex-X1, 7.9% reduction on Cortex-A55.

Raw numbers:
JIT time only, X1. 5,789,700 nanoseconds -> 3,976,805.4 nanoseconds -> 68.687%
JIT time only, A55. 35,493,985 nanoseconds -> 21,370,773 nanoseconds -> 60.209%

Full Compile time, X1. 302,443,555 nanoseconds -> 280,850,928 nanoseconds -> 92.86%
Full Compile time, A55. 1,080,218,555 nanoseconds -> 994,860,986 nanoseconds -> 92.098%

Sonicadvance1:
Tested a wackload of games and Steam and I don't notice anything being more broken than previously. High probability that nothing is broken in combination with all the unit tests.

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Emitter.h
```diff
@@ -0,0 +1,721 @@
...
+#include "Interface/Core/ArchHelpers/CodeEmitter/BranchOps.inl"
+#include "Interface/Core/ArchHelpers/CodeEmitter/LoadstoreOps.inl"
+#include "Interface/Core/ArchHelpers/CodeEmitter/SystemOps.inl"
+#include "Interface/Core/ArchHelpers/CodeEmitter/ScalarOps.inl"
+#include "Interface/Core/ArchHelpers/CodeEmitter/ASIMDOps.inl"
+#include "Interface/Core/ArchHelpers/CodeEmitter/SVEOps.inl"
+
+  private:
+    template<typename T>
+    uint32_t Encode_ra(T Reg) {
```
lioncash, line 663: seems like these helpers can either be static or const qualified

lioncash (commented):
Based off a cursory review, things look mostly fine

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Emitter.h
Sonicadvance1: You're right, I can switch those over tomorrow morning.

External/FEXCore/unittests/Emitter/System_Tests.cpp
```diff
@@ -0,0 +1,147 @@
+#include "TestDisassembler.h"
+
+#include <catch2/catch.hpp>
+#include <fcntl.h>
+
+static TestDisassembler Emit;
```
neobrain, line 6: Instead of having a static global for this, you can turn it into a "Fixture". In this case, that just requires replacing `TEST_CASE(...)` with `TEST_CASE_METHOD(TestDisassembler, ...)`. You can then access TestDisassembler's member functions directly in the test logic, i.e. you can drop the global TestDisassembler and `TEST_SINGLE` simplifies to:
```cpp
#define TEST_SINGLE(emit, expected) \
{ \
  emit; \
  const auto Str = DisassembleSingle(); \
  CHECK(Str == expected); \
}
```

External/FEXCore/unittests/Emitter/System_Tests.cpp
```diff
@@ -0,0 +1,147 @@
...
+using namespace FEXCore::ARMEmitter;
+
+#define TEST_SINGLE(emit, expected) \
+{ \
+  Emit.emit; \
+  const auto Str = Emit.DisassembleSingle(); \
+  CHECK(Str == expected); \
```
neobrain, line 13: On failure, this currently outputs something like:
```
~/FEX/External/FEXCore/unittests/Emitter/ALU_Tests.cpp:771: FAILURE:
  CHECK( Str == "neg x30, x29" )
with expansion:
  "neg x30, x29" == "neg x30, x29"
```

By changing this to `CHECK((emit, DisassembleSingle()) == expected)`, we'd make this a bit more descriptive at the expense of verbosity:
```
~/FEX/External/FEXCore/unittests/Emitter/ALU_Tests.cpp:128: FAILURE:
  CHECK( (sub(Size::i32Bit, Reg::r29, Reg::r28, 0, true), DisassembleSingle()) == "sub w29, w28, #0x0 (0)" )
with expansion:
  "sub w29, w28, #0x0 (0)"
  ==
  "sub w29, w28, #0x0 (0)"
```

Not sure it's a better trade-off, but maybe worth considering :)

External/FEXCore/unittests/Emitter/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+if (ENABLE_VIXL_DISASSEMBLER)
```
neobrain, line 1: Since this check is hidden in a third-level CMakeLists.txt, maybe add a warning about emitter tests being disabled if tests are enabled but vixl is not? When testing this PR, it took me a couple of minutes to figure out why the emitter tests weren't building for me.

External/FEXCore/unittests/Emitter/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  option(CATCH_BUILD_STATIC_LIBRARY "" ON)
```
neobrain, line 2: This is already done in the main CMakeLists.txt and shouldn't be needed anywhere else (with the exception of FEXLinuxTests, which is cross-compiled as an external project).

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Emitter.h
Sonicadvance1: Resolved.

External/FEXCore/unittests/Emitter/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS *.cpp)
+
+  set (LIBS fmt::fmt vixl Catch2::Catch2WithMain FEXCore_Base)
+  foreach(TEST ${TESTS})
+    get_filename_component(TEST_NAME ${TEST} NAME_WLE)
+    add_executable(Emitter_${TEST_NAME} ${TEST})
+    target_link_libraries(Emitter_${TEST_NAME} PRIVATE ${LIBS})
+    target_include_directories(Emitter_${TEST_NAME} PUBLIC "${CMAKE_CURRENT_SOURCE_DIR}/../../Source/")
+    set_target_properties(Emitter_${TEST_NAME} PROPERTIES RUNTIME_OUTPUT_DIRECTORY "${CMAKE_BINARY_DIR}/EmitterTests")
+    add_test(NAME Emitter_${TEST_NAME} COMMAND "${CMAKE_BINARY_DIR}/EmitterTests/Emitter_${TEST_NAME}")
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

External/FEXCore/unittests/Emitter/System_Tests.cpp
Sonicadvance1: Resolved.

External/FEXCore/unittests/Emitter/System_Tests.cpp
Sonicadvance1: Changed it over to that. Resolved.

External/FEXCore/unittests/Emitter/CMakeLists.txt
Sonicadvance1: Resolved.

External/FEXCore/unittests/Emitter/CMakeLists.txt
Sonicadvance1: Resolved.

neobrain:
I just realized enabling `ENABLE_VIXL_DISASSEMBLER` also automatically enables printing of disassembly. That's not great behavior, since it makes the resulting FEX binary useless for real-world testing (and presumably also adds a ton of runtime overhead for all other tests?). Can we split this feature off from the use of vixl for testing?

Maybe just make vixl a hard requirement for testing, and make `ENABLE_VIXL_DISASSEMBLER` control whether or not to print disassembly at runtime?

Sonicadvance1:
> I just realized enabling `ENABLE_VIXL_DISASSEMBLER` also automatically enables printing of disassembly during normal operation of FEX. That's not great behavior, since it makes the resulting FEX binary useless for real-world testing (and presumably also adds a ton of runtime overhead for all other tests?). Can we split this feature off from the use of vixl for testing?
> 
> Maybe just make vixl a hard requirement for testing, and make `ENABLE_VIXL_DISASSEMBLER` control whether or not to print disassembly at runtime?

We could /probably/ do that, but would need another vixl submodule update to enable the disassembler/decoder when tests are enabled. Currently we don't enable this code unless the disassembler or simulator is enabled.

Maybe worth thinking about resolving these issues once it gets wired up to CI?

lioncash (approved):

PR #2294, opened 2022-12-28
Thunks: Implement support for 32-bit Vulkan thunking

neobrain:
Simplified version of #1611, redesigned from the ground up around requirements for 32-bit Vulkan specifically. For the time being, this allows us to omit the full-blown analysis pass to check for ABI-compatibility between guest/host architectures.

## Overview

This PR adds assisted automated repacking of structs with differing data layout between guest and host. Repacking is fully automatic for "simple" structures (i.e. no pointers, and all members are *either* built-in types with same size on guest and host *or* they are simple structs themselves).

For pointer members/parameters, repacking must be assisted by one of these generator annotations:
* Type annotation `opaque_type`: Used for "handle"-like types that don't have any publicly visible member data. (This will skip any repacking.)
* Type-member annotation `custom_repack`: Allows for a user-provided function to be called after size-extending the pointer to host-layout. All other architecture differences are assumed to be taken care of by that function.
* Type-member/Function annotation `param_is_sized_array`: Used to specify that a function parameter is a dynamically-sized array, the size of which is specified by another function parameter. Repacking logic (if needed) is recursively applied to each element in the array. (NOT IMPLEMENTED)

## Impact

All currently thunked libraries will likely need annotations for pointer-types throughout their respective APIs. The full scope of this is TBD.

## TODO

(incomplete)
* [ ] Implement `param_is_sized_array` annotation
* [ ] Add full compatibility with the Vulkan API


ThunkLibs/libfex_thunk_test/Host.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+
+static std::unordered_map<StructType, const TestBaseStruct*(*)(const TestBaseStruct*)> next_handlers {
+    converters<StructType::Struct1, TestStruct1>,
+    converters<StructType::Struct2, TestStruct2>,
+};
+
+// Normally, we would implement fex_custom_repack individually for each customized struct.
+// In this case, they all need the same repacking, so we just implement it once and alias all fex_custom_repack instances
+extern "C" const_void_ptr default_fex_custom_repack(const const_void_ptr& source) {
```
Sonicadvance1, line 65: I guess these shouldn't be references since checking for nullptr on a reference is...icky?

Sonicadvance1: Oop wait, it's a reference to a pointer type, nevermind.

neobrain:
Superseded by #3487.

PR #2295, opened 2022-12-29
OpcodeDispatcher: Handle VCVTSS2SI/VCVTTSS2SI/VCVTSD2SI/VCVTTSD2SI

lioncash:
Nice thing about this is the existing implementation for SSE Just Works™ here.


Sonicadvance1 (approved):
Nice

PR #2306, opened 2023-01-01
OpcodeDispatcher: Handle immediate variants of VPERMILPD/VPERMILPS

lioncash:


Sonicadvance1 (approved):
What a weirdly behaving instruction.

PR #2307, opened 2023-01-03
ThunksDB: Fix misspelt guest library names

neobrain:
Found by turning overly defensive code into a hard error.

Wasn't sure whether `ERROR_AND_DIE` was the right thing to use here. Do we have a better mechanism for critical failures?

### DIFF
--- a/Data/ThunksDB.json
+++ b/Data/ThunksDB.json
@@ -54,7 +54,7 @@
       ]
     },
     "xcb-dri2": {
-      "Library": "libxcb_dri2-guest.so",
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
`LogMan::Msg::AFmt` could be a choice if you want the assert to go away on release builds.
But if you want to ensure a fatal error stops the program, `ERROR_AND_DIE_FMT` is the best we have.

neobrain:
Alright, will keep the `ERROR_AND_DIE_FMT` then.

PR #2308, opened 2023-01-03
ThunkDB: Clean up database loading

neobrain:
A number of various cleanups eliminating the bits that confused me while trying to understand this code the first time.

* Reduces floating state carried around in lambdas (and the overall number of lambdas used)
* Replaces thunk DB template handling (such as `%ARCH_PREFIX%`): Instead of iteratively replacing each "prefix" one after the other, all of them are replaced in one go now
* Decouples processing steps that were previously intertwined (such as DB template filling and the actual insertion into the `Overlays` list)

As a side effect, this avoids a few heap allocations from `std::string` and `std::function`.

### DIFF
--- a/Source/Tests/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tests/LinuxSyscalls/FileManagement.cpp
@@ -300,10 +300,15 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
       decltype(FileManager::ThunkOverlays)& ThunkOverlays;
       decltype(ThunkDB)& ThunkDB;
       const std::filesystem::path& ThunkGuestPath;
+      bool Is64BitMode;
 
       void SetupOverlay(const ThunkDBObject& DBDepend) {
           auto ThunkPath = ThunkGuestPath / DBDepend.LibraryName;
           if (!std::filesystem::exists(ThunkPath)) {
+            if (!Is64BitMode) {
+              // Guest libraries not existing is expected since not all libraries are thunked on 32-bit
+              return;
+            }
             ERROR_AND_DIE_FMT("Requested thunking via guest library \"{}\" that does not exist", ThunkPath.string());
           }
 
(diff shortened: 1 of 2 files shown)


neobrain:
Not sure whether 0f51e03191c5c0ea9f051918125c1e1cab17b3b4 or 0f51e03191c5c0ea9f051918125c1e1cab17b3b4 is nicer to fill the thunk path templates. Apart from that this change set is complete.

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -293,52 +268,71 @@ FileManager::FileManager(FEXCore::Context::Context *ctx)
...
+      decltype(FileManager::ThunkOverlays)& ThunkOverlays;
+      decltype(ThunkDB)& ThunkDB;
+      const std::filesystem::path& ThunkGuestPath;
 
-          auto ThunkPath = ThunkGuestPath / DBDepend->second.LibraryName;
+      void SetupOverlay(const ThunkDBObject& DBDepend) {
+          auto ThunkPath = ThunkGuestPath / DBDepend.LibraryName;
           if (!std::filesystem::exists(ThunkPath)) {
-              ERROR_AND_DIE_FMT("Requested thunking via guest library \"{}\" that does not exist", ThunkPath.string());
+            ERROR_AND_DIE_FMT("Requested thunking via guest library \"{}\" that does not exist", ThunkPath.string());
```
Sonicadvance1, line 307: Oop, just tested with this and this error needs to be non-fatal.
A common use case is to enable thunks globally while testing, ignoring if any 32-bit games accidentally are run.
So changing this over to a `LogMan::Msg::IFmt` is going to be necessary.

neobrain: Should be resolved now by `return`ing on 32-bit instead of printing an error.

Sonicadvance1 (approved):

PR #2313, opened 2023-01-08
Arm64: Fixes large offset spill slots

Sonicadvance1:
Found an application today (hashtree tests) that causes us to spill a large amount of values on to the stack.

We were encoding larger offsets than what unsigned offset load and store can handle.

If the offset is too large for the loadstore, use a temporary to put the offset in to first.


neobrain (approved):
Yikes, good catch!

Any chance we could add range-checking debug asserts to immediate-taking instruction helpers?

Sonicadvance1:
> Any chance we could add range-checking debug asserts to immediate-taking instruction helpers?

We actually already have debug range checking asserts on all immediate taking instructions. This is how I found this.
https://github.com/FEX-Emu/FEX/blob/main/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/LoadstoreOps.inl#L4304

PR #2314, opened 2023-01-08
F64: Fix integer immediates for add,mul,div,sub

CallumDev:
FADDF64 & co. had incorrect logic for loading integer arguments.
Copies the correct logic over from FILDF64.


Sonicadvance1 (approved):
Very nice catch!

PR #2318, opened 2023-01-08
JitSymbols: Fixes a crash that can occur

Sonicadvance1:
When a process is in the process of forking and getting ready for execve, it is common practice to do a `close_range` or close loop to close all file descriptors before the execve.

This is a security/sanitization feature to ensure that FDs aren't leaked to the child process. While it is more reliable to have these FDs opened with O_CLOEXEC, people get it wrong all the time so this feature has been put in place. Both python and glibc wrappers for launching applications do this.

The problem with this for FEX is that we were using a FILE handle for emitting JIT symbols to the perf file. When the underlying FD is ripped out from under the FILE handle, it throws an assert that we can't recover from.

Switching to a raw FD and checking to ensure the FD is still open on writes means that we can safely stop JIT symbol logging when a process is closing FDs under us.

Fixes a crash in Steam early startup where a python script is run for checking if packages are installed.

### DIFF
--- a/External/FEXCore/Source/Common/JitSymbols.cpp
+++ b/External/FEXCore/Source/Common/JitSymbols.cpp
@@ -1,64 +1,86 @@
 #include "Common/JitSymbols.h"
 
+#include <fcntl.h>
 #include <string>
 #include <unistd.h>
 
 #include <fmt/format.h>
 
 namespace FEXCore {
-  JITSymbols::JITSymbols() : fp{nullptr, std::fclose} {
+  JITSymbols::JITSymbols() {
   }
 
-  JITSymbols::~JITSymbols() = default;
-
-  void JITSymbols::InitFile() {
-    const auto PerfMap = fmt::format("/tmp/perf-{}.map", getpid());
-
-    fp.reset(fopen(PerfMap.c_str(), "wb"));
-    if (fp) {
-      // Disable buffering on this file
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Common/JitSymbols.cpp
```diff
@@ -1,64 +1,91 @@
...
-      // Disable buffering on this file
-      setvbuf(fp.get(), nullptr, _IONBF, 0);
-    }
+    fd = open(PerfMap.c_str(), O_CREAT | O_EXCL | O_WRONLY | O_DIRECT, 0644);
   }
 
   void JITSymbols::Register(const void *HostAddr, uint64_t GuestAddr, uint32_t CodeSize) {
-    if (!fp) return;
+    if (fd == -1) return;
+    char TmpBuffer[512];
```
neobrain, line 28: Symbol names (particularly C++ ones) can easily become larger than this, can't they? Is this path critical enough to warrant not using `std::string` instead?

neobrain (approved):

External/FEXCore/Source/Common/JitSymbols.cpp
Sonicadvance1: Not sure of the performance implications of these yet. Switched over to std::string usage.

PR #2320, opened 2023-01-09
IR: Removes NumArgs member from IR ops

Sonicadvance1:
Split off from #2243 to remove each member individually.

Shaves 8-bits off of each IR op.
No need to cart around this data when it is constant for each operation. Especially since most optimization passes don't need the data anyway.

Needed to add a new `GetRAArgs` to get the number of SSA arguments that get RA versus `GetArgs` which returns all SSA arguments the IR operation owns. This is what was causing #2243 to fail CI since it needs to know the difference in some places.


lioncash:
Just needs a rebase, since 2319 was merged

lioncash (approved):

PR #2323, opened 2023-01-09
ConstProp: Pool inline constants

Sonicadvance1:
In large blocks we can be generating a ton of inline constants. But in most cases these end up being 0, 1, or (1 << N).
Add these to a map and reuse if possible. Makes some IR blocks significantly smaller for later optimization passes.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -29,6 +29,7 @@ $end_info$
 #include <string.h>
 #include <tuple>
 #include <unordered_map>
+#include <tsl/robin_map.h>
 #include <utility>
 
 namespace FEXCore::IR {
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -198,6 +199,16 @@ class ConstProp final : public FEXCore::IR::Pass {
 
   std::unordered_map<uint64_t, OrderedNode*> ConstPool;
   std::map<OrderedNode*, uint64_t> AddressgenConsts;
+
+  // Pool inline constant generation. These are typically very small and pool efficiently.
+  tsl::robin_map<uint64_t, OrderedNode*> InlineConstantGen;
+  OrderedNode *CreateInlineConstant(IREmitter *IREmit, uint64_t Constant) {
+    if (InlineConstantGen[Constant]) {
```
lioncash, line 206: Maybe do something like:

```cpp
const auto iter = InlineConstantGen.find(Constant);
if (iter != InlineConstantGen.end()) {
  return iter->second;
}
```

So there's only one access into the map in this case

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -198,6 +199,16 @@ class ConstProp final : public FEXCore::IR::Pass {
...
+      return InlineConstantGen[Constant];
+    }
+    auto InlineConstant = InlineConstantGen[Constant] = IREmit->_InlineConstant(Constant);
```
lioncash, line 209: I think robin_map has the same API as the std for this, so maybe this could be:

```cpp
auto res = InlineConstantGen.insert_or_assign(Constant, IREmit->_InlineConstant(Constant));
return res.first->second;
```

or something like that

lioncash (approved):

PR #2176, opened 2022-11-26
Syscalls: Adds support for rt_sigreturn and sigreturn 

Sonicadvance1:
By utilizing the stack unwinding using the guest stack space, we can now
implement these directly.

With all of this work in place, libunwind now successfully unwinds a
stack frame from a signal!

It needed to find the appropriate signal frame to know that it is in a signal stack to unwind accordingly.

Before:
```
*** signal 11
Register dump:

<... Register dump snipped ...>

Backtrace:
/mnt/Work/Projects/work/glibc-build/debug/libSegFault.so(+0x37b4)[0x7f245c9537b4]
[0x7f245c996000]
```

After:
```
*** signal 11
Register dump:

<... Register dump snipped ...>

Backtrace:
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/invalid_int3(+0x1130)[0x5589d3060130]
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/invalid_int3(+0x1146)[0x5589d3060146]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x7fffe25d7d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x7fffe25d7e40]
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/invalid_int3(+0x1065)[0x5589d3060065]
```

This will be a draft for now because this is a fairly spooky change. I don't want this merged without some excessive testing.

PR #2326, opened 2023-01-11
MContext: Insert a stack cookie with assertions enabled

Sonicadvance1:
Pulled from #2176.
Ensures that when we are handling signals we are actually restoring a stack state that is what we expect..

While this could randomly intersect with other stack data, it is highly unlikely and will still capture incorrect stack frames otherwise.

Keeps it out of release build to ensure we aren't sticking random data in the stack when it wouldn't have even been checked.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
@@ -18,8 +18,19 @@ enum ContextFlags : uint32_t {
   CONTEXT_FLAG_32BIT = (1U << 1),
 };
 
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+constexpr uint64_t STACK_COOKIE_MAGIC = 0x4142434445464748ULL;
+#endif
+
 struct X86ContextBackup {
   // Host State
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
```diff
@@ -18,8 +18,15 @@ enum ContextFlags : uint32_t {
...
 };
 
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+constexpr uint64_t STACK_COOKIE_MAGIC = 0x4142434445464748ULL;
+#endif
+
 struct X86ContextBackup {
   // Host State
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+  uint64_t StackCookie;
```
neobrain, line 28: Could this get a comment explaining how this is useful during signal handling?

neobrain (approved):

External/FEXCore/Source/Interface/Core/ArchHelpers/MContext.h
Sonicadvance1: Added a comment.

PR #2327, opened 2023-01-11
Dispatcher: Fixes x86-64 SA_SIGINFO generation 

Sonicadvance1:
Pulled from https://github.com/FEX-Emu/FEX/pull/2176.

On x86-64 the SA_SIGINFO sa_flag is actually a no-op. It is always used
even if not set.

Ensure that we setup siginfo_t regardless of flag being set.

On 32-bit x86 this still needs to be adhered to.

Little side bits that don't change anything
- EFLAGS is passed in signfo correctly.
- User provided restorer usage locations is documented but not
  implemented.


Sonicadvance1:
This is blocking some additional work that I want to do. Would be nice to get a review on this @neobrain.

unittests/FEXLinuxTests/tests/signal/sigtest_siginfo.64.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+TEST_CASE("siginfo") {
+  // On x86-64, the signal handler receives siginfo even if SA_SIGINFO isn't set.
+  // This flag is effectively a no-op, not changing behaviour.
+  struct sigaction act{};
+  act.sa_sigaction = Handler;
+  act.sa_flags = 0;
+  sigaction(SIGSEGV, &act, nullptr);
+
+  CauseInt();
```
neobrain, line 40: How about adding a variant here with SA_SIGINFO set to check that the flag does have an effect on 32-bit?

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -427,226 +427,262 @@ bool Dispatcher::HandleGuestSignal(FEXCore::Core::InternalThreadState *Thread, i
...
-      Frame->State.gregs[X86State::REG_RDX] = UContextLocation;
+    // Set up the new SP for stack handling
+    // The host is required to provide us a restorer.
+    // If the guest didn't provide a restorer then the application should fail with a SIGSEGV.
+    // TODO: Emulate SIGSEGV when the guest doesn't provide a restorer.
+    // TODO: Support the guest's restorer. Required for libunwind.
+    NewGuestSP -= 8;
+    *(uint64_t*)NewGuestSP = SignalReturn;
+    // TODO: Once we support the guest's restorer.
+    // *(uint64_t*)NewGuestSP = (uint64_t)GuestAction->restorer;
```
neobrain, line 565: Since there are a couple of disabled lines like this, I'd recommend declaring a `const bool incomplete_guest_restorer_support` variable at the top, so we make sure this code at least compiles and remains somewhat consistent.

neobrain (commented):
Tests need some rework, looks good otherwise!

unittests/FEXLinuxTests/tests/signal/sigtest_siginfo.64.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+static void Handler(int Signal, siginfo_t *siginfo, void *context) {
+  ucontext_t* _context = (ucontext_t*)context;
+
+  if (_context->uc_mcontext.gregs[REG_RIP] == EXPECTED_RIP &&
+      Signal == EXPECTED_SIGNAL) {
+    _context->uc_mcontext.gregs[REG_RIP] += 2;
+  }
+  else {
+    exit(-1);
```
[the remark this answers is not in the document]
Sonicadvance1: Switched over to using the helper.

unittests/FEXLinuxTests/tests/CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ if(BITNESS EQUAL 64)
   list(REMOVE_ITEM TESTS ${TESTS_32_ONLY})
 endif()
 
+if(BITNESS EQUAL 32)
+  file(GLOB_RECURSE TESTS_64_ONLY CONFIGURE_DEPENDS *.64.cpp)
+  list(REMOVE_ITEM TESTS ${TESTS_64_ONLY})
```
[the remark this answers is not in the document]
Sonicadvance1: Not dead code, this is removing the 64-bit tests from list `TESTS_64_ONLY` from list `TESTS`.

External/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: Used const.

unittests/FEXLinuxTests/tests/signal/sigtest_siginfo.64.cpp
Sonicadvance1: That's the next step after this PR. We don't handle some of the 32-bit subtleties around this yet.
32-bit has four combinations that we need to support.
- sigaction + SA_SIGINFO
- sigaction - SA_SIGINFO
- rt_sigaction + SA_SIGINFO
- rt_sigaction - SA_SIGINFO

I wanted this PR to stay about 64-bit so I can solve the 32-bit side independently.

unittests/FEXLinuxTests/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ ExternalProject_Add(FEXLinuxTests_32
 file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS tests/*.cpp)
 file(GLOB_RECURSE TESTS_32_ONLY CONFIGURE_DEPENDS tests/*.32.cpp)
 
-set(TESTS_32 ${TESTS})
+set(TESTS_32 ${TESTS_32_ONLY})
```
neobrain, line 30: Could you make it so that the 32-bit-exclusive tests are treated symmetrically to the 64-bit-exclusive tests instead?

Conceptually there's no difference between the two AFAICT, so it's unclear now why there needs to be 32-bit specific logic here.

I don't fully understand what's going on the new version, but I'm guessing instead we'd want to have dedicated `TESTS_64_ONLY`/`TESTS_64` sets.

unittests/FEXLinuxTests/tests/CMakeLists.txt
```diff
@@ -14,6 +14,11 @@ if(BITNESS EQUAL 64)
   list(REMOVE_ITEM TESTS ${TESTS_32_ONLY})
 endif()
 
+if(BITNESS EQUAL 32)
```
neobrain, line 17: This can be an else/elseif instead.

unittests/FEXLinuxTests/tests/CMakeLists.txt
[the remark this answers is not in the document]
neobrain: Ah, yes, my bad.

unittests/FEXLinuxTests/CMakeLists.txt
Sonicadvance1: Resolved. These are now treating 32-bit tests and 64-bit tests the same.
So these kind of fall in to three categories now, Tests that are supported on both 32-bit and 64-bit, tests that are supported on 64-bit only, and tests that are supported on 32-bit only.
In switching this over, it also uncovered a bug that we weren't running all the 32-bit tests, resulting in some test failures that cropped up due to a prologue being prepended on some 32-bit naked functions. Had to resolve that as well.

neobrain: Thanks, and great to have that issue with the 32-bit tests uncovered and resolved!

neobrain (approved):

PR #2333, opened 2023-01-13
ELFCodeLoader: Don't use std::random_device for RNG

### LINKED ISSUE #2095: std::random_device can fail!
```
terminate called after throwing an instance of 'std::runtime_error'
  what():  random_device::random_device(const std::string&): device not available
```

We need to support a fallback path in this case, I didn't realize this was a possibility.

Sonicadvance1:
Fixes #2095

std::random_device can fail to find a random source and throw an assert during initialization.

The constructor allows you to provide a token string to select an explicit random source, but this is c++ library specific and will still assert if the source isn't found. Also specific tokens are very much target and library specific, so it is unsafe to use.

In the libstdc++ case, the default implementation will try to open `/dev/urandom` which might not be available on all targets.

Instead of relying on this C++ object, use the `getrandom` syscall directly to generate our RNG used in the ELFCodeLoader. Resolves any sort of asserting case, blocks on RNG generation, and is guaranteed to be available since this syscall has been available since a very old kernel version.

### DIFF
--- a/Source/Tests/ELFCodeLoader2.h
+++ b/Source/Tests/ELFCodeLoader2.h
@@ -34,6 +34,7 @@
 #include <sys/auxv.h>
 #include <sys/mman.h>
 #include <sys/personality.h>
+#include <sys/random.h>
 
 #define PAGE_START(x) ((x) & ~(uintptr_t)(4095))
 #define PAGE_OFFSET(x) ((x) & 4095)
@@ -174,6 +175,17 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
     }
 
     return LoadBase;
+  }
+
+  static bool GetRandom(void *Data, size_t DataSize) {
+    ssize_t Result{};
+    do {
+      // This is guaranteed to not be interrupted by a signal,
+      // since fewer than 256 bytes of RNG data are requested
+      Result = getrandom(Data, DataSize, 0);
+    } while (Result != -1 && Result != DataSize);
+
+    return Result != -1;
   }
 
   public:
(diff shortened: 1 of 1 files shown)


Source/Tests/ELFCodeLoader2.h
```diff
@@ -472,9 +473,18 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
...
         constexpr uint64_t ASLR_BITS_64 = 28;
         constexpr uint64_t ASLR_BITS_32 = 8;
-        std::random_device rd;
-        std::uniform_int_distribution<uint64_t> d(0);
-        uint64_t ASLR_Offset = d(rd);
+        uint64_t ASLR_Offset{};
+        ssize_t Result{};
+        do {
+          // Use the getrandom syscall instead of std::random_device because std can assert.
+          // <256 bytes of RNG requested, guaranteed to not be interrupted by signal.
```
neobrain, line 480: Are these two unrelated facts, or does one imply the other? In the latter case I'd suggest:
```suggestion
          // This is guaranteed to not be interrupted by a signal,
          // since fewer than 256 bytes of RNG data are requested
```

Source/Tests/ELFCodeLoader2.h
```diff
@@ -720,11 +730,18 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
...
     else {
       // Nothing provided from the kernel, generate our own random values.
-      std::random_device rd;
-      std::uniform_int_distribution<uint64_t> d(0);
-
-      RandomLoc[0] = d(rd);
-      RandomLoc[1] = d(rd);
+      ssize_t Result{};
+      do {
+        // Use the getrandom syscall instead of std::random_device because std can assert.
```
neobrain, line 735: Moving this to a file-local helper template function allows us to avoid duplicating this docstring.

neobrain (approved):

Source/Tests/ELFCodeLoader2.h
Sonicadvance1: Switched to local helper.

Source/Tests/ELFCodeLoader2.h
Sonicadvance1: Modified comment.

Issue #2136, opened 2022-11-03
execveat syscall incorrect with `AT_EMPTY_PATH` flag


Sonicadvance1:
We always pass in `/proc/self/exe` which means we end up ignoring this flag.
This causes us to potentially execute a process incorrectly since we are supposed to execute the FD rather whatever is in the arguments.
This is a fairly tricky edge case and not many people (that I'm aware of) are abusing it right now.
Will need a test-case with executing an FD that is a she-bang file, regular ELF, and with and without binfmt_misc.

PR #2334, opened 2023-01-13
FEXLoader: Adds support for execveat with AT_EMPTY_PATH 

Sonicadvance1:
Fixes https://github.com/FEX-Emu/FEX/issues/2136

This is a fairly tricky edge case to support with FEX.
If execveat is used with AT_EMPTY_PATH then the application can pass an
FD to execve instead of a filename. This includes FDs that have been
deleted from the disk so the child process can't open it by filename
anymore.

To work around this limitation, we need to pass the FD to the new FEX
process and open it directly, similar to how binfmt_misc works with FDs.
The FD will get passed through environment variables, which the new
process will check for and then remove the variable from the
environment.

Lots of prickly edge cases to support here.

Without binfmt_misc:
- Passes the FD to FEXLoader directly.
  - Requires duplicating the FD if it has O_CLOEXEC on the FD.

With binfmt_misc:
- Shebang file, pass directly to FEXLoader, just like without binfmt.
- x86 ELF Files, rely on the kernel's binfmt_misc support here.
- Unsupported ELF files, let kernel handle it through binfmt_misc

Argument handling:
- The application can pass in no arguments.
  - Means our application configurations were failing to find a config
  - Also various checks in the frontend were failing.
  - If opened through an FD, find the symlink for that FD for the
    application configuration instead.

Side note:
Fixed a performance issue in execve where when we were checking for file
format support. Either ELF or Shebang files, we were reading the /whole/
file upfront. We only need to read a header worth of ELF files, and only
257 bytes if it is potentially a shebang file. Should dramatically
reduce some application's execve times.

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
@@ -1,0 +1,28 @@
+#pragma once
+#include <fcntl.h>
+#include <string>
+#include <sys/stat.h>
+#include <span>
+#include <unistd.h>
+
+namespace FHU::Symlinks {
+// Checks to see if a filepath is a symlink.
+inline bool IsSymlink(const std::string &Filename) {
+  struct stat Buffer{};
+  int Result = lstat(Filename.c_str(), &Buffer);
+  return Result == 0 && S_ISLNK(Buffer.st_mode);
+}
+
+// Resolves a symlink path.
+// Doesn't handle recursive symlinks.
+// Doesn't append null terminator character.
+// Returns a string_view of the resolved path, or an empty view on error.
+inline std::string_view ResolveSymlink(const std::string &Filename, std::span<char> ResultBuffer) {
+  ssize_t Result = readlink(Filename.c_str(), ResultBuffer.data(), ResultBuffer.size());
+  if (Result == -1) {
+    return {};
+  }
+
+  return std::string_view(ResultBuffer.data(), Result);
+}
+}
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -2,11 +2,13 @@
 #include "Common/Config.h"
 
 #include <FEXCore/Config/Config.h>
+#include <FEXHeaderUtils/SymlinkChecks.h>
 
 #include <cstring>
 #include <filesystem>
 #include <fstream>
 #include <map>
+#include <linux/limits.h>
 #include <list>
 #include <unordered_map>
 #include <utility>
@@ -40,12 +42,68 @@ namespace FEX::Config {
     }
   }
 
-  std::pair<std::string, std::string> LoadConfig(
+  std::string RecoverGuestProgramFilename(std::string Program, bool ExecFDInterp, const std::string_view ProgramFDFromEnv) {
+    // If executed with a FEX FD then the Program argument might be empty.
+    // In this case we need to scan the FD node to recover the application binary that exists on disk.
+    // Only do this if the Program argument is empty, since we would prefer the application's expectation
+    // of application name.
+    if (!ProgramFDFromEnv.empty() && Program.empty()) {
+      // Get the `dev` node of the execveat fd string.
+      Program = "/dev/fd/";
+      Program += ProgramFDFromEnv;
+    }
+
+    // If FEX was invoked through an FD path (either binfmt_misc or execveat) then we need to check the
+    // Program to see if it is a symlink to find the real path.
+    //
+    // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
+    //   - Full application path with execve (See Side Note)
+    //   - FD path with execveat and FD doesn't have an existing file on the disk
+    //
+    // ProgramFDFromEnv: Arg[0] is Application provided data or `/dev/fd/<FD>` from above fix-up.
+    //   - execveat was either passed no arguments (argv=NULL) or the first argument is an empty string (argv[0]="").
+    //   - FD path with execveat and FD doesn't have an existing file on the disk
+    //
+    // Side Note:
+    //  The `execve` syscall doesn't take an FD but binfmt_misc will give FEX an FD to execute still.
+    //  Arg[0] will always contain a pathname to an application in this case and it won't be a symlink.
+    //  Kernel resolves the symlink before passing the application to the interpreter.
+    //
+    // Examples:
+    //  - Regular execve. Application must exist on disk.
+    //    execve binfmt_misc args layout:   `FEXInterpreter <Full path> <user provided argv[0]> <user provided argv[n]>...`
+    //  - Regular execveat with FD. FD is backed by application on disk.
+    //    execveat binfmt_misc args layout: `FEXInterpreter <Full path> <user provided argv[0]> <user provided argv[n]>...`
+    //  - Regular execveat with FD. FD points to file on disk that has been deleted.
+    //    execveat binfmt_misc args layout: `FEXInterpreter /dev/fd/<FD> <user provided argv[0]> <user provided argv[n]>...`
+    if (ExecFDInterp || !ProgramFDFromEnv.empty()) {
+      // Only in the case that FEX is executing an FD will the program argument potentially be a symlink.
+      // This symlink will be in the style of `/dev/fd/<FD>`.
+      //
+      // If the argument /is/ a symlink then resolve its path to get the original application name.
+      if (FHU::Symlinks::IsSymlink(Program)) {
+        char Filename[PATH_MAX];
+        auto SymlinkPath = FHU::Symlinks::ResolveSymlink(Program, Filename);
+        if (SymlinkPath.starts_with('/')) {
+          // This file was executed through an FD.
+          // Remove the ` (deleted)` text if the file was deleted after the fact.
+          // Otherwise just get the symlink without the deleted text.
+          return std::string{SymlinkPath.substr(0, SymlinkPath.rfind(" (deleted)"))};
+        }
+      }
+    }
+
+    return Program;
+  }
+
+  ApplicationNames LoadConfig(
     bool NoFEXArguments,
     bool LoadProgramConfig,
     int argc,
     char **argv,
-    char **const envp) {
+    char **const envp,
+    bool ExecFDInterp,
+    const std::string_view ProgramFDFromEnv) {
     FEXCore::Config::Initialize();
     FEXCore::Config::AddLayer(FEXCore::Config::CreateGlobalMainLayer());
     FEXCore::Config::AddLayer(FEXCore::Config::CreateMainLayer());
@@ -68,7 +126,8 @@ namespace FEX::Config {
         return {};
       }
 
-      std::string Program = Args[0];
+      Args[0] = RecoverGuestProgramFilename(std::move(Args[0]), ExecFDInterp, ProgramFDFromEnv);
+      std::string& Program = Args[0];
 
       bool Wine = false;
       std::filesystem::path ProgramName;
@@ -121,7 +180,7 @@ namespace FEX::Config {
         FEXCore::Config::AddLayer(FEXCore::Config::CreateAppLayer(SteamAppName, FEXCore::Config::LayerType::LAYER_LOCAL_STEAM_APP));
       }
 
-      return std::make_pair(Program, ProgramName);
+      return ApplicationNames{std::move(Program), std::move(ProgramName)};
     }
     return {};
   }
--- a/Source/Common/Config.h
+++ b/Source/Common/Config.h
@@ -19,11 +19,33 @@ namespace FEX::Config {
 
   void SaveLayerToJSON(const std::string& Filename, FEXCore::Config::Layer *const Layer);
 
-  std::pair<std::string, std::string> LoadConfig(
+  struct ApplicationNames {
+    // This is the full path to the program (if it exists).
+    std::string ProgramPath;
+    // This is the program executable name (if it exists).
+    std::string ProgramName;
+  };
+
+  /**
+   * @brief Loads the FEX and application configurations for the application that is getting ready to run.
+   *
+   * @param NoFEXArguments Do we want to parse FEXLoader arguments, Or is this FEXInterpreter?
+   * @param LoadProgramConfig Do we want to load application specific configurations?
+   * @param argc The `argc` passed to main(...)
+   * @param argv The `argv` passed to main(...)
+   * @param envp The `envp` passed to main(...)
+   * @param ExecFDInterp If FEX was executed with binfmt_misc FD argument
+   * @param ProgramFDFromEnv The execveat FD argument passed through FEX
+   *
+   * @return The application name and path structure
+   */
+  ApplicationNames LoadConfig(
     bool NoFEXArguments,
     bool LoadProgramConfig,
     int argc,
     char **argv,
-    char **const envp
+    char **const envp,
+    bool ExecFDInterp,
+    const std::string_view ProgramFDFromEnv
   );
 }
--- a/Source/Linux/Utils/ELFContainer.cpp
+++ b/Source/Linux/Utils/ELFContainer.cpp
@@ -12,65 +12,83 @@ $end_info$
 #include <algorithm>
 #include <cstring>
 #include <elf.h>
+#include <fcntl.h>
 #include <filesystem>
 #include <fstream>
 #include <memory>
 #include <system_error>
+#include <sys/stat.h>
+#include <unistd.h>
 #include <vector>
 
 namespace ELFLoader {
-  ELFContainer::ELFType ELFContainer::GetELFType(std::string const &Filename) {
-  std::fstream ELFFile(Filename, std::fstream::in | std::fstream::binary);
 
-  if (!ELFFile.is_open()) {
-    return ELFType::TYPE_NONE;
-  }
-
-  ELFFile.seekg(0, ELFFile.end);
-  size_t FileSize = ELFFile.tellg();
-  ELFFile.seekg(0, ELFFile.beg);
-
-  size_t ELFHeaderSize = std::max(sizeof(Elf32_Ehdr), sizeof(Elf64_Ehdr));
-  if (FileSize < ELFHeaderSize) {
-    return ELFType::TYPE_NONE;
-  }
-
-  FileSize = ELFHeaderSize;
-
-  std::vector<char> RawFile(FileSize);
-  ELFFile.read(RawFile.data(), FileSize);
-  ELFFile.close();
-
-  uint8_t *Ident = reinterpret_cast<uint8_t*>(&RawFile.at(0));
-
-  if (Ident[EI_MAG0] != ELFMAG0 ||
-      Ident[EI_MAG1] != ELFMAG1 ||
-      Ident[EI_MAG2] != ELFMAG2 ||
-      Ident[EI_MAG3] != ELFMAG3) {
-    return ELFType::TYPE_NONE;
-  }
-
-  union {
-    Elf32_Ehdr _32;
-    Elf64_Ehdr _64;
-  } Header;
-
-  if (Ident[EI_CLASS] == ELFCLASS32) {
-    memcpy(&Header, reinterpret_cast<Elf32_Ehdr *>(&RawFile.at(0)),
-      sizeof(Elf32_Ehdr));
-    if (Header._32.e_machine == EM_386) {
-      return ELFType::TYPE_X86_32;
+  static ELFContainer::ELFType CheckELFType(uint8_t* Data) {
+    if (Data[EI_MAG0] != ELFMAG0 ||
+        Data[EI_MAG1] != ELFMAG1 ||
+        Data[EI_MAG2] != ELFMAG2 ||
+        Data[EI_MAG3] != ELFMAG3) {
+      return ELFContainer::ELFType::TYPE_NONE;
     }
-  }
-  else if (Ident[EI_CLASS] == ELFCLASS64) {
-    memcpy(&Header, reinterpret_cast<Elf64_Ehdr *>(&RawFile.at(0)),
-      sizeof(Elf64_Ehdr));
-    if (Header._64.e_machine == EM_X86_64) {
-      return ELFType::TYPE_X86_64;
+
+    if (Data[EI_CLASS] == ELFCLASS32) {
+      Elf32_Ehdr *Header = reinterpret_cast<Elf32_Ehdr *>(Data);
+      if (Header->e_machine == EM_386) {
+        return ELFContainer::ELFType::TYPE_X86_32;
+      }
     }
+    else if (Data[EI_CLASS] == ELFCLASS64) {
+      Elf64_Ehdr *Header = reinterpret_cast<Elf64_Ehdr *>(Data);
+      if (Header->e_machine == EM_X86_64) {
+        return ELFContainer::ELFType::TYPE_X86_64;
+      }
+    }
+
+    return ELFContainer::ELFType::TYPE_OTHER_ELF;
   }
 
-  return ELFType::TYPE_OTHER_ELF;
+ELFContainer::ELFType ELFContainer::GetELFType(std::string const &Filename) {
+  // Open the Filename to determine if it is a shebang file.
+  int FD = open(Filename.c_str(), O_RDONLY | O_CLOEXEC);
+  if (FD == -1) {
+    return ELFType::TYPE_NONE;
+  }
+
+  auto ELFType = GetELFType(FD);
+  close(FD);
+  return ELFType;
+}
+
+ELFContainer::ELFType ELFContainer::GetELFType(int FD) {
+  // We don't know the state of the FD coming in since this might be a guest tracked FD.
+  // Need to be extra careful here not to adjust file offsets and status flags.
+  //
+  // We can't use dup since that makes the FD have the same underlying state backing both FDs.
+
+  // We need to first determine the file size through fstat.
+  struct stat buf{};
+  if (fstat(FD, &buf) == -1) {
+    // Couldn't get size.
+    return ELFType::TYPE_NONE;
+  }
+
+  constexpr size_t ELFHeaderSize = std::max(sizeof(Elf32_Ehdr), sizeof(Elf64_Ehdr));
+  if (buf.st_size < ELFHeaderSize) {
+    // Is not a valid ELF.
+    return ELFType::TYPE_NONE;
+  }
+
+  std::array<char, ELFHeaderSize> RawFile;
+
+  // Read the header so we can tell if it is a supported ELF file.
+  // Can't adjust file offset, so use pread.
+  if (pread(FD, &RawFile.at(0), RawFile.size(), 0) != RawFile.size()) {
+    // Couldn't read
+    LogMan::Msg::EFmt("Couldn't read potential ELF FD");
+    return ELFType::TYPE_NONE;
+  }
+
+  return CheckELFType(reinterpret_cast<uint8_t*>(&RawFile.at(0)));
 }
 
 ELFContainer::ELFContainer(std::string const &Filename, std::string const &RootFS, bool CustomInterpreter) {
--- a/Source/Tests/ELFCodeLoader2.h
+++ b/Source/Tests/ELFCodeLoader2.h
@@ -27,6 +27,7 @@
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXHeaderUtils/Syscalls.h>
 #include <FEXHeaderUtils/TypeDefines.h>
+#include <FEXHeaderUtils/SymlinkChecks.h>
 
 #include <elf.h>
 #include <fcntl.h>
@@ -191,10 +192,10 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
       // Do some special handling if the RootFS's linker is a symlink
       // Ubuntu's rootFS by default provides an absolute location symlink to the linker
       // Resolve this around back to the rootfs
-      auto SymlinkSize = FEX::HLE::GetSymlink(RootFSLink, Filename, PATH_MAX - 1);
-      if (SymlinkSize > 0 && Filename[0] == '/') {
+      auto SymlinkPath = FHU::Symlinks::ResolveSymlink(RootFSLink, Filename);
+      if (SymlinkPath.starts_with('/')) {
         RootFSLink = RootFS;
-        RootFSLink += std::string_view(Filename, SymlinkSize);
+        RootFSLink += SymlinkPath;
       }
       else {
         break;
@@ -215,11 +216,23 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
 
   std::vector<LoadedSection> Sections;
 
-  ELFCodeLoader2(std::string const &Filename, std::string const &RootFS, [[maybe_unused]] std::vector<std::string> const &args, std::vector<std::string> const &ParsedArgs, char **const envp = nullptr, FEXCore::Config::Value<std::string> *AdditionalEnvp = nullptr) :
+  ELFCodeLoader2(std::string const &Filename, const std::string_view FEXFDString, std::string const &RootFS, [[maybe_unused]] std::vector<std::string> const &args, std::vector<std::string> const &ParsedArgs, char **const envp = nullptr, FEXCore::Config::Value<std::string> *AdditionalEnvp = nullptr) :
     Args {args} {
 
     bool LoadedWithFD = false;
     int FD = getauxval(AT_EXECFD);
+
+    if (!FEXFDString.empty()) {
+      // If we passed the execve FD to us then use that.
+      const char *StartPtr = FEXFDString.data();
+      char *EndPtr{};
+      FD = ::strtol(StartPtr, &EndPtr, 10);
+      if (EndPtr == StartPtr) {
+        LogMan::Msg::AFmt("FEXInterpreter passed invalid FD to exececute: {}", FEXFDString);
+        return;
+      }
+      unsetenv("FEX_EXECVEFD");
+    }
 
     // If we are provided an EXECFD then attempt to execute that first
     // This happens in the case of binfmt_misc usage
--- a/Source/Tests/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tests/LinuxSyscalls/Syscalls.cpp
@@ -42,6 +42,7 @@ $end_info$
 #include <memory>
 #include <regex>
 #include <sched.h>
+#include <span>
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
@@ -166,33 +167,17 @@ uint64_t GetDentsEmulation<false>(int, FEX::HLE::x64::linux_dirent*, uint32_t);
 template
 uint64_t GetDentsEmulation<true>(int, FEX::HLE::x32::linux_dirent_32*, uint32_t);
 
-static bool IsSupportedByInterpreter(std::string const &Filename) {
-  // If it is a supported ELF then we can
-  if (ELFLoader::ELFContainer::IsSupportedELF(Filename.c_str())) {
-    return true;
+static bool IsShebangFile(std::span<char> Data) {
+  // File isn't large enough to even contain a shebang.
+  if (Data.size() <= 2) {
+    return false;
   }
 
-  // If it is a shebang then we also can
-  std::fstream File;
-  size_t FileSize{0};
-  File.open(Filename, std::fstream::in | std::fstream::binary);
-
-  if (!File.is_open())
-    return false;
-
-  File.seekg(0, File.end);
-  FileSize = File.tellg();
-  File.seekg(0, File.beg);
-
-  // Is the file large enough for shebang
-  if (FileSize <= 2)
-    return false;
-
-  // Handle shebang files
-  if (File.get() == '#' &&
-      File.get() == '!') {
-    std::string InterpreterLine;
-    std::getline(File, InterpreterLine);
+  // Handle shebang files.
+  if (Data[0] == '#' &&
+      Data[1] == '!') {
+    std::string_view InterpreterView { Data.data(), Data.size() };
+    std::string InterpreterLine { Data.data(), InterpreterView.find('\n') };
     std::vector<std::string> ShebangArguments{};
 
     // Shebang line can have a single argument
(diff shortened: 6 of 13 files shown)


Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -166,33 +166,19 @@ uint64_t GetDentsEmulation<false>(int, FEX::HLE::x64::linux_dirent*, uint32_t);
...
+  if (Data[0] == '#' &&
+      Data[1] == '!') {
+    std::istringstream FileData{std::string(std::string_view(Data, Size))};
 
-  // Handle shebang files
-  if (File.get() == '#' &&
-      File.get() == '!') {
     std::string InterpreterLine;
-    std::getline(File, InterpreterLine);
+    std::getline(FileData, InterpreterLine);
```
neobrain, line 181:
```suggestion
    std::string InterpreterLine { Data, std::find(Data, Data + Size, '\n') };
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +212,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
   return false;
 }
 
-uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* envp, ExecveAtArgs *Args) {
+static bool IsShebangFilename(std::string const &Filename) {
+  // If it is a shebang then we also can
```
neobrain, line 216: incomplete comment

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +212,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
-      strcmp(pathname, "/proc/thread-self/exe") == 0 ||
-      strcmp(pathname, PidSelfPath) == 0) {
-    // If pointing to self then redirect to the application
-    // JRE and shapez.io does this
-    Filename = FEX::HLE::_SyscallHandler->Filename();
+    if (strcmp(pathname, "/proc/self/exe") == 0 ||
+        strcmp(pathname, "/proc/thread-self/exe") == 0 ||
+        strcmp(pathname, PidSelfPath) == 0) {
+      // If pointing to self then redirect to the application
+      // JRE and shapez.io does this
```
neobrain, line 321:
```suggestion
      // If pointing to self then redirect to the application
      // (JRE and shapez.io do this)
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +212,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+    Result = ::syscall(SYS_execveat, Args.dirfd, Filename.c_str(), argv, envp, Args.flags);
     SYSCALL_ERRNO();
   }
 
   // We don't have an interpreter installed or we are executing a non-ELF executable
   // We now need to munge the arguments
   std::vector<const char *> ExecveArgs{};
+  std::vector<const char *> EnvpArgs{};
+  char *const *EnvpPtr = envp;
+  const char NullString[] = "\0";
```
neobrain, line 364: Is it important to use `"\0"` over simply `""` here?

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Fixed comment.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Nope, changed.

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+  FileSize = File.tellg();
+  File.seekg(0, File.beg);
+
+  // Is the file large enough for shebang
+  if (FileSize <= 2)
+    return false;
+
+  // The maximum length of the shebang line is `#!` + 255 chars
+  std::vector<char> Header(std::min(257ul, FileSize));
+  File.read(&Header.at(0), Header.size());
```
neobrain, line 233:
```suggestion
  std::array<char, 257> Header;
  auto ChunkSize = std::min(257ul, FileSize);
  File.read(Header.data(), ChunkSize);
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+    // Couldn't get size.
+    return false;
+  }
+
+  // Is the file large enough for shebang
+  if (buf.st_size <= 2)
+    return false;
+
+  // The maximum length of the shebang line is `#!` + 255 chars
+  std::vector<char> Header(std::min(257l, buf.st_size));
```
neobrain, line 256: => std::array

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
   return false;
 }
 
-uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* envp, ExecveAtArgs *Args) {
+static bool IsShebangFilename(std::string const &Filename) {
```
neobrain, line 214: If we have to implement this twice anyway, why not use POSIX `open` and make `IsShebangFilename` call `IsShebangFD`?

Source/Tests/LinuxSyscalls/Syscalls.cpp
neobrain, line 233: Seeking twice to perform the pathological size check is probably (?) more expensive than always reading and checking later, so:
```cpp
  std::array<char, 257> Header;
  auto ChunkSize = File.read(Header.data(), Header.size());
```
(where the size check now happens in `IsShebangFile`)

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
   std::error_code ec;
   std::string RootFS = FEX::HLE::_SyscallHandler->RootFSPath();
-
-  // Check the rootfs if it is available first
-  if (pathname[0] == '/') {
-    auto Path = FEX::HLE::_SyscallHandler->FM.GetEmulatedPath(pathname, true);
-    if (!Path.empty() && std::filesystem::exists(Path, ec)) {
-      Filename = Path;
+  ELFLoader::ELFContainer::ELFType Type{};
+  bool IsFDExec{};
```
neobrain, line 275:
```suggestion
  const bool IsFDExec = (Args.flags & AT_EMPTY_PATH) && strlen(pathname) == 0;
```
(Keeps state a bit more local)

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
-  int pid = getpid();
+    char PidSelfPath[50];
+    snprintf(PidSelfPath, 50, "/proc/%i/exe", pid);
 
-  char PidSelfPath[50];
-  snprintf(PidSelfPath, 50, "/proc/%i/exe", pid);
+    if (strcmp(pathname, "/proc/self/exe") == 0 ||
+        strcmp(pathname, "/proc/thread-self/exe") == 0 ||
+        strcmp(pathname, PidSelfPath) == 0) {
+      // If pointing to self then redirect to the application
```
neobrain, line 319: What exactly is the scenario described here? Is it "FEX runs `java`, and `java` then calls `execve` on a symlink to its own executable to start the guest Java application"?

At that point, what is `/proc/self/exe` actually pointing to - FEX or `java`? Comment needs to be a bit clearer, I think.

Source/Tests/LinuxSyscalls/Syscalls.h
```diff
@@ -80,9 +81,15 @@ uint64_t UnimplementedSyscallSafe(FEXCore::Core::CpuStateFrame *Frame, uint64_t
 struct ExecveAtArgs {
   int dirfd;
   int flags;
+  static ExecveAtArgs Empty() {
```
neobrain, line 84:
```suggestion
  static constexpr ExecveAtArgs Empty = { .dirfd = AT_FDCWD, .flags = 0 };
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,156 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+  if (pread(FD, &Header.at(0), Header.size(), 0) != Header.size()) {
+    // Couldn't read
+    LogMan::Msg::EFmt("Couldn't read potential shebang FD");
+    return false;
+  }
+
+  return IsShebangFile(&Header.at(0), Header.size());
+}
+
+uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* envp, ExecveAtArgs &Args) {
```
neobrain, line 269: Do we actually need to take `Args` by non-const-reference or can we just pass by value here instead?

neobrain (commented):
Logic here seems very intricate. I've done a first pass trying to wrap my head around it, adding a few comments with minor performance improvements and suggestions to clarify some documentation.

Source/Linux/Utils/ELFContainer.cpp
```diff
@@ -37,40 +64,43 @@ namespace ELFLoader {
...
 
-  if (Ident[EI_MAG0] != ELFMAG0 ||
-      Ident[EI_MAG1] != ELFMAG1 ||
-      Ident[EI_MAG2] != ELFMAG2 ||
-      Ident[EI_MAG3] != ELFMAG3) {
+ELFContainer::ELFType ELFContainer::GetELFType(int FD) {
+  // We don't know the state of the FD coming in since this might be a guest tracked FD.
+  // Need to be extra careful here not to adjust file offsets and status flags.
+  //
+  // Can't use dup since that makes the FD have the same file description backing both FDs.
```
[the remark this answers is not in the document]
Sonicadvance1: This one is a detail of how Linux implement file handles.
Your typical `FD` is officially a `File Descriptor`.
`FD` is an integer that points to a structure in the kernel called a `File Description`.
The `File Descriptor` is distinct from the `File Description` in that you can have multiple Descriptors pointing to a single Description.

The `File Description` holds all of the state of the file, including locks, read/write position, and other various data.

An example in this case is that you can use `dup` to duplicate the `File Descriptor`, but it still points to the same `File Description`. So if you seek in one FD, the duplicated FD has its position updated.

ex
```cpp
int FD1 = open(...); // Open an FD
lseek(FD1, 512, SEEK_SET); // Offset FD1 to 512
int FD2 = dup(FD1); // Duplicate the file descriptor
lseek(FD2, 0, SEEK_SET); // Offset FD2 to 0

off_t Off1 = lseek(FD1, 0, SEEK_CUR);
off_t Off2 = lseek(FD2, 0, SEEK_CUR);
assert(Off1 == Off2, "This will match");
```

`FD` will always be referring to the integer. `File Description` always referring to the backend data structure.
Which this is all just a detail about why we can't use dup.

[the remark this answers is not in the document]
neobrain: Is this `File Description` a widely known term? It's the first time I'm hearing about it, and not sure it's a suitable term to use here if it's an obscure term only kernel hackers know about.

Would `Can't use dup since that makes the FD have the same underlying state backing both FDs.` convey the same information?

Source/Tests/LinuxSyscalls/Syscalls.h
Sonicadvance1:
```
/mnt/Work/Work/work/FEXNew/Source/Tests/LinuxSyscalls/Syscalls.h:84:33: error: constexpr variable cannot have non-literal type 'const FEX::HLE::ExecveAtArgs'
  static constexpr ExecveAtArgs Empty = { .dirfd = AT_FDCWD, .flags = 0 };
                                ^
/mnt/Work/Work/work/FEXNew/Source/Tests/LinuxSyscalls/Syscalls.h:84:33: note: incomplete type 'const FEX::HLE::ExecveAtArgs' is not a literal type
/mnt/Work/Work/work/FEXNew/Source/Tests/LinuxSyscalls/Syscalls.h:81:8: note: definition of 'FEX::HLE::ExecveAtArgs' is not complete until the closing '}'
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -166,33 +166,18 @@ uint64_t GetDentsEmulation<false>(int, FEX::HLE::x64::linux_dirent*, uint32_t);
...
-  FileSize = File.tellg();
-  File.seekg(0, File.beg);
-
-  // Is the file large enough for shebang
-  if (FileSize <= 2)
-    return false;
+  // Handle shebang files.
+  if (Data[0] == '#' &&
+      Data[1] == '!') {
+    std::istringstream FileData{std::string(std::string_view(Data, Size))};
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <fcntl.h>
+#include <string>
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace FHU::Symlinks {
+[[maybe_unused]]
+static bool IsSymlink(const std::string &Filename) {
+  // Checks to see if a filepath is a symlink.
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+#include <fcntl.h>
+#include <string>
+#include <sys/stat.h>
+#include <unistd.h>
+
+namespace FHU::Symlinks {
+[[maybe_unused]]
+static bool IsSymlink(const std::string &Filename) {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Common/Config.cpp
```diff
@@ -45,7 +47,9 @@ namespace FEX::Config {
     bool LoadProgramConfig,
     int argc,
     char **argv,
-    char **const envp) {
+    char **const envp,
+    bool ExecFDInterp,
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
```diff
@@ -0,0 +1,20 @@
...
+  // Checks to see if a filepath is a symlink.
+  struct stat Buffer{};
+  int Result = lstat(Filename.c_str(), &Buffer);
+  return Result == 0 && S_ISLNK(Buffer.st_mode);
+}
+
+[[maybe_unused]]
+static ssize_t GetSymlink(const std::string &Filename, char *ResultBuffer, size_t ResultBufferSize) {
```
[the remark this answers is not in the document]
Sonicadvance1: Named switched over to `ResolveSymlink` with documentation.
Switched over to span.
Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Updated comment.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Passed by value now.

Source/Linux/Utils/ELFContainer.cpp
```diff
@@ -37,40 +64,43 @@ namespace ELFLoader {
 
   FileSize = ELFHeaderSize;
 
-  std::vector<char> RawFile(FileSize);
-  ELFFile.read(RawFile.data(), FileSize);
+  std::vector<char> RawFile(ELFHeaderSize);
+  ELFFile.read(RawFile.data(), ELFHeaderSize);
```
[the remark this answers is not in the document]
Sonicadvance1: Actually this wasn't an issue, FileSize was getting set to ELFHeaderSize above this.
Resolved by changing how this code worked anyway.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,38 @@ namespace FEX::Config {
 
       std::string Program = Args[0];
 
+      if (ExecFD && Program.empty()) {
+        // If executed with a FEX FD then we could be missing the program argument.
```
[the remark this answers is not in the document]
Sonicadvance1: We are resolving the symlink in the `/dev/fd/<FD>` node. Which will point us to a potentially deleted file.
Updated the comments here to be clarified more.

Source/Common/Config.cpp
```diff
@@ -45,7 +47,9 @@ namespace FEX::Config {
     bool LoadProgramConfig,
     int argc,
     char **argv,
-    char **const envp) {
+    char **const envp,
+    bool ExecFDInterp,
+    bool ExecFD) {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,38 @@ namespace FEX::Config {
...
+        char *FDString = getenv("FEX_EXECVEFD");
+
+        if (FDString) {
+          Program = "/dev/fd/";
+          Program += std::string(FDString);
+        }
+      }
+
+      if (ExecFDInterp || ExecFD) {
+        // If the application was executed with binfmt_misc and an FD then the first argument might be in the form of
```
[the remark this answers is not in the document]
Sonicadvance1: Updated the comments to be more clear here.
`Program` is non-empty when `ExecFD==true` when the user provides an argument in `execveat`'s argv structure that isn't empty.
Comments updated to clarify around this.

FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
```diff
@@ -0,0 +1,20 @@
...
+  return readlink(Filename.c_str(), ResultBuffer, ResultBufferSize);
```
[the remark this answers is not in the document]
Sonicadvance1: This does not handle recursive symlinks.
Documented to specifically state that it doesn't.
Couple of places we need recursive symlinks, we need to do rootfs insertion each step, so doing a spin on `IsSymlink` + `ResolveSymlink` wouldn't work elsewhere anyway.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,38 @@ namespace FEX::Config {
...
+          if (SymlinkSize > 0 && Filename[0] == '/') {
+            // Null terminate string.
+            Filename[SymlinkSize] = 0;
+
+            // Check if the FD was deleted.
+            if (access(Filename, F_OK) != 0) {
+              // This file was executed through an FD and was deleted.
+              // Remove the ` (deleted)` text
+              std::string_view ProgramView = std::string_view(Filename, SymlinkSize);
+              Program = Args[0] = std::string_view(Filename, ProgramView.rfind(" (deleted)"));
```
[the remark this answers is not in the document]
Sonicadvance1: Changed the logic a bit here

Source/Common/Config.cpp
```diff
@@ -70,6 +74,38 @@ namespace FEX::Config {
...
+          char Filename[PATH_MAX];
+          auto SymlinkSize = FHU::Symlinks::GetSymlink(Program, Filename, PATH_MAX - 1);
+          if (SymlinkSize > 0 && Filename[0] == '/') {
+            // Null terminate string.
+            Filename[SymlinkSize] = 0;
+
+            // Check if the FD was deleted.
+            if (access(Filename, F_OK) != 0) {
+              // This file was executed through an FD and was deleted.
+              // Remove the ` (deleted)` text
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Linux/Utils/ELFContainer.cpp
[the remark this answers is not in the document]
Sonicadvance1: Changed comment.

Source/Tests/ELFCodeLoader2.h
```diff
@@ -220,6 +220,13 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
 
     bool LoadedWithFD = false;
     int FD = getauxval(AT_EXECFD);
+    char *FDString = getenv("FEX_EXECVEFD");
+
+    if (FDString) {
+      // If we passed the execve FD to us then use that.
+      FD = std::stoi(std::string(FDString));
```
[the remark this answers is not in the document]
Sonicadvance1: Changed how we get FD here. 0 is technically a safe FD if an application was being mean. Not that the surrounding code might not handle it correctly.
strtol lets us get the integer and check if it couldn't parse without doing a try-catch.

Source/Linux/Utils/ELFContainer.cpp
```diff
@@ -37,40 +64,43 @@ namespace ELFLoader {
...
+    return ELFType::TYPE_NONE;
   }
-  else if (Ident[EI_CLASS] == ELFCLASS64) {
-    memcpy(&Header, reinterpret_cast<Elf64_Ehdr *>(&RawFile.at(0)),
-      sizeof(Elf64_Ehdr));
-    if (Header._64.e_machine == EM_X86_64) {
-      return ELFType::TYPE_X86_64;
-    }
+
+  std::vector<char> RawFile(ELFHeaderSize);
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -166,33 +166,18 @@ uint64_t GetDentsEmulation<false>(int, FEX::HLE::x64::linux_dirent*, uint32_t);
...
-    return true;
-  }
-
-  // If it is a shebang then we also can
-  std::fstream File;
-  size_t FileSize{0};
-  File.open(Filename, std::fstream::in | std::fstream::binary);
-
-  if (!File.is_open())
+static bool IsShebangFile(char *Data, size_t Size) {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Good call. Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Some of the logic with this comment and others conflicted. So you'll need to check that again with the changes.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Some of the logic with this comment and others conflicted. So you'll need to check that again with the changes.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Some of the logic with this comment and others conflicted. So you'll need to check that again with the changes.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,65 @@ namespace FEX::Config {
...
+      }
+
+      // If FEX was invoked through an FD path (either binfmt_misc or execveat) then we need to check the
+      // Program to see if it is a symlink to find the real path.
+      //
+      // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
+      //   - Full application path with execve (See Side Note)
+      //   - FD path with execveat and FD doesn't have an existing filesystem on the disk
+      //
+      // ProgramFDFromEnv: Arg[0] is set to Application provided data or `/dev/fd/<FD>` from above fix-up.
```
neobrain, line 98: This is useful information, but it mixes "things that applications do" with "how they do it" and (maybe?) "how FEX handles" it. Currently it's difficult to tell which of these you're describing at which point.

If you're describing what values certain values take, I'd use "the value of X will be" over "X is set", since the latter is ambiguous about who sets it (FEX before this line? FEX after this line? The application? ...).

If you're describing values that we actively set after the comment, active phrasing ("We set"/"FEX sets") helps clarify this over passive phrasing ("is set").

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,133 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+    if (!IsShebangFD(Args.dirfd) && Type == ELFLoader::ELFContainer::ELFType::TYPE_NONE) {
+      // If our interpeter doesn't support this file format AND ELF format is NONE then ENOEXEC
+      // binfmt_misc could end up handling this case but we can't know that without parsing binfmt_misc ourselves
+      // Return -ENOEXEC until proven otherwise
+      return -ENOEXEC;
+    }
+  }
+  else
+  {
+    // Check the rootfs if it is available first
```
neobrain, line 267:
```suggestion
    // For absolute paths, check the rootfs first (if available)
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,133 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
-      strcmp(pathname, PidSelfPath) == 0) {
-    // If pointing to self then redirect to the application
-    // JRE and shapez.io does this
-    Filename = FEX::HLE::_SyscallHandler->Filename();
+    if (!IsShebangFilename(Filename) && Type == ELFLoader::ELFContainer::ELFType::TYPE_NONE) {
+      // If our interpeter doesn't support this file format AND ELF format is NONE then ENOEXEC
+      // binfmt_misc could end up handling this case but we can't know that without parsing binfmt_misc ourselves
+      // Return -ENOEXEC until proven otherwise
+      return -ENOEXEC;
+    }
```
neobrain, line 308: This is repeated twice now. Instead, we can write the result of IsShebangFilename to a boolean in each branch and use a single `!is_shebang && Type == ...` check after the branch.

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -321,26 +350,60 @@ uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* env
...
+      while (*OldEnvp) {
+        EnvpArgs.emplace_back(*OldEnvp);
+        ++OldEnvp;
+      }
+    }
+
+    int Flags = fcntl(Args.dirfd, F_GETFD);
+    if (Flags & FD_CLOEXEC) {
+      // We need to duplicate the FD if FD_CLOEXEC is set.
+      // FEX needs the FD to live past execve when binfmt_misc isn't used.
```
neobrain, line 384:
```suggestion
      // FEX needs the FD to live past execve when binfmt_misc isn't used,
      // so duplicate the FD if FD_CLOEXEC is set.
```

An even better wording might be `Duplicate the FD if FD_CLOEXEC is set, so that FEX can...`

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -321,26 +350,60 @@ uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* env
...
+
+    // Remove AT_EMPTY_PATH now. Can't exist in this path.
+    Args.flags &= ~AT_EMPTY_PATH;
+
+    // Create the environment variable to pass the FD to our FEX.
+    // Needs to stick around until execveat completes.
+    FDExecEnv = "FEX_EXECVEFD=" + std::to_string(Args.dirfd);
+
+    // Insert the FD for FEX to track.
+    EnvpArgs.emplace_back(FDExecEnv.data());
```
neobrain, line 396: Where is this extra argument actually used?

neobrain (commented):
Second pass done: I mostly understand what's happening now. The added documentation really helped!
We're getting there :)

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -226,89 +211,133 @@ static bool IsSupportedByInterpreter(std::string const &Filename) {
...
+  // Can't use dup since that makes the FD have the same file description backing both FDs.
+
+  // The maximum length of the shebang line is `#!` + 255 chars
+  std::array<char, 257> Header;
+  const auto ChunkSize = 257l;
+  const auto ReadSize = pread(FD, &Header.at(0), ChunkSize, 0);
+
+  if (ReadSize <= 2) {
+    return false;
+  }
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Common/Config.h
```diff
@@ -19,11 +19,33 @@ namespace FEX::Config {
...
+    std::string ProgramPath;
+    // This is the program executable name (if it exists).
+    std::string ProgramName;
+  };
+
+  /**
+   * @brief Loads the FEX and application configurations for the application that is getting ready to run.
+   *
+   * @param NoFEXArguments Do we want to parse FEXLoader arguments, Or is this FEXInterpreter?
+   * @param LoadProgramConfig Do we want to load application specific applications?
```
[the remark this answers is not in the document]
Sonicadvance1: Indeed, replaced with `application specific configurations`

Source/Common/Config.cpp
```diff
@@ -70,6 +74,65 @@ namespace FEX::Config {
...
+      // In this case we need to scan the FD node to recover the application binary that exists on disk.
+      // Only do this if the Program argument is empty, since we would prefer the application's expectation
+      // of application name.
+      if (ProgramFDFromEnv && Program.empty()) {
+        // Get the execvefd string, then link to the `dev` node of that FD.
+        char *FDString = getenv("FEX_EXECVEFD");
+
+        if (FDString) {
+          Program = "/dev/fd/";
+          Program += std::string(FDString);
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/SymlinkChecks.h
```diff
@@ -0,0 +1,23 @@
...
+  struct stat Buffer{};
+  int Result = lstat(Filename.c_str(), &Buffer);
+  return Result == 0 && S_ISLNK(Buffer.st_mode);
+}
+
+// Resolves a symlink path.
+// Doesn't handle recursive symlinks.
+// Doesn't append null terminator character.
+// Returns the length of the filepath.
+inline ssize_t ResolveSymlink(const std::string &Filename, std::span<char> ResultBuffer) {
```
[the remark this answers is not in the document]
Sonicadvance1: Switched over to a string_view

Source/Common/Config.cpp
```diff
@@ -121,7 +184,7 @@ namespace FEX::Config {
         FEXCore::Config::AddLayer(FEXCore::Config::CreateAppLayer(SteamAppName, FEXCore::Config::LayerType::LAYER_LOCAL_STEAM_APP));
       }
 
-      return std::make_pair(Program, ProgramName);
+      return ApplicationNames{Program, ProgramName};
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Tests/ELFCodeLoader2.h
```diff
@@ -186,12 +187,12 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
 
     std::string RootFSLink = RootFS + File;
 
-    char Filename[PATH_MAX];
+    char Filename[PATH_MAX + 1];
```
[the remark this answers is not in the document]
Sonicadvance1: You're right. Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -321,26 +350,60 @@ uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* env
...
+    }
+
+    int Flags = fcntl(Args.dirfd, F_GETFD);
+    if (Flags & FD_CLOEXEC) {
+      // We need to duplicate the FD if FD_CLOEXEC is set.
+      // FEX needs the FD to live past execve when binfmt_misc isn't used.
+      Args.dirfd = dup(Args.dirfd);
+    }
+
+    // Remove AT_EMPTY_PATH now. Can't exist in this path.
```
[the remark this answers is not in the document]
Sonicadvance1: Wording changed.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,65 @@ namespace FEX::Config {
...
+          Program += std::string(FDString);
+        }
+      }
+
+      // If FEX was invoked through an FD path (either binfmt_misc or execveat) then we need to check the
+      // Program to see if it is a symlink to find the real path.
+      //
+      // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
+      //   - Full application path with execve (See Side Note)
+      //   - FD path with execveat and FD doesn't have an existing filesystem on the disk
```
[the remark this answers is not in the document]
Sonicadvance1: Typo, should have been `file` not `filesystem`

Source/Common/Config.cpp
```diff
@@ -70,6 +74,65 @@ namespace FEX::Config {
 
       std::string Program = Args[0];
 
+      // If executed with a FEX FD then the Program argument might be empty.
```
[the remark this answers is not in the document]
Sonicadvance1: moved to a helper.

Source/Common/Config.cpp
```diff
@@ -70,6 +74,65 @@ namespace FEX::Config {
...
+      //
+      // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
+      //   - Full application path with execve (See Side Note)
+      //   - FD path with execveat and FD doesn't have an existing filesystem on the disk
+      //
+      // ProgramFDFromEnv: Arg[0] is set to Application provided data or `/dev/fd/<FD>` from above fix-up.
+      //   - execveat was passed either zero arguments or empty string in first argument.
+      //     - Can be done in two ways:
+      //       - Give `argv` in `execveat` a nullptr
+      //       - Pass `argv[0]` in `execveat` a zero length string.
```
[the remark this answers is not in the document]
Sonicadvance1: Changed.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Duplication removed.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: Comment updated.

Source/Tests/LinuxSyscalls/Syscalls.cpp
Sonicadvance1: It's used in the next FEXInterpreter instance that is getting executed.

neobrain: Where can I see this in the code? Do we actively have to extract it again from envp on the other end, or does this happen indirectly?

Sonicadvance1: https://github.com/FEX-Emu/FEX/pull/2334/files#diff-cd1944a762a8c051124ab5532528cddff1933f5c0d2e3a45785c36751868b244R224
We extract it from envp here.

Source/Common/Config.cpp
```diff
@@ -70,6 +132,8 @@ namespace FEX::Config {
 
       std::string Program = Args[0];
 
+      Program = Args[0] = RecoverGuestProgramFilename(Args[0], ExecFDInterp, ProgramFDFromEnv);
```
neobrain, line 135: * Moving `Args[0]` into the function enables zero-copy behavior for the common path
* `Program` does not have a different value than `Args[0]` on function exit (I think), so it can be a reference
```suggestion
      Args[0] = RecoverGuestProgramFilename(std::move(Args[0]), ExecFDInterp, ProgramFDFromEnv);
      std::string& Program = Args[0];
```

Source/Tests/ELFCodeLoader2.h
```diff
@@ -220,6 +221,18 @@ class ELFCodeLoader2 final : public FEXCore::CodeLoader {
...
+
+    if (FDString) {
+      // If we passed the execve FD to us then use that.
+      char *EndPtr{};
+      FD = ::strtol(FDString, &EndPtr, 10);
+      if (EndPtr == FDString) {
+        LogMan::Msg::AFmt("FEXInterpreter passed invalid FD to exececute: {}", FDString);
+        return;
+      }
+      unsetenv("FEX_EXECVEFD");
```
Sonicadvance1, line 234: We pass envp to the guest application below.
We need to remove `FEX_EXECVEFD` getting passed to the guest down below on line 294.
Either doing string compares or using `environ` instead of `envp`

Source/Tests/LinuxSyscalls/Syscalls.cpp
neobrain: Ah, yes - I was somehow looking for a manual `envp` readout as opposed to `getenv`. Makes sense!

Source/Common/Config.cpp
Sonicadvance1: Ah, you're right. For some reason I was thinking `Argv[0]` was a reference and was getting modified while reading this and used later when FEXLoader's main pulls the arguments.

Only the following wine executable name searching will see the difference in change beyond what we return. Will fix.

Source/Common/Config.h
```diff
@@ -19,11 +19,33 @@ namespace FEX::Config {
...
+   */
+  ApplicationNames LoadConfig(
     bool NoFEXArguments,
     bool LoadProgramConfig,
     int argc,
     char **argv,
-    char **const envp
+    char **const envp,
+    bool ExecFDInterp,
+    bool ProgramFDFromEnv
```
neobrain, line 49: Since the we're only passing `!!getenv(...)` (and false) in here anyway, how about passing in the full value as a `string_view` or an `optional` instead?

That reduces the amount of hidden state accessed in LoadConfig, avoids env-querying twice, and removes one of the weakly typed boolean toggles in the parameter list.

neobrain (approved):
Looks good!

Source/Common/Config.h
Sonicadvance1: Switched over to a string_view.

Source/Common/Config.cpp
Sonicadvance1: Resolved.

Source/Tests/ELFCodeLoader2.h
Sonicadvance1: I was mistaken here, documentation is a but unclear but unsetenv modifies envp in place. Resolved.

PR #2338, opened 2023-01-14
CPUID: Optimize initialization

Sonicadvance1:
Map lookup was quite expensive, switched over to three small vectors
that are constexpr instead.

Some file querying and parsing was fairly slow as well. Optimized to
make that CPU time to go away.

This improves initialization time of CPUIDEmu by 33%

### DIFF
--- a/External/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/External/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -88,7 +88,6 @@ static uint32_t CalculateNumberOfCPUs() {
 //       when AVX implementations are further along.
 constexpr uint32_t SUPPORTS_AVX = 0;
 
-// #define CPUID_AMD
 #ifdef CPUID_AMD
 constexpr uint32_t FAMILY_IDENTIFIER =
   0 |          // Stepping
@@ -122,25 +121,24 @@ void CPUIDEmu::SetupHostHybridFlag() {
   uint64_t MIDR{};
   for (size_t i = 0; i < CPUs; ++i) {
     std::error_code ec{};
-    std::string MIDRPath = "/sys/devices/system/cpu/cpu" + std::to_string(i) + "/regs/identification/midr_el1";
-    if (std::filesystem::exists(MIDRPath, ec)) {
-      std::vector<char> Data{};
-      // Needs to be a fixed size since depending on kernel it will try to read a full page of data and fail
-      // Only read 18 bytes for a 64bit value prefixed with 0x
-      if (FEXCore::FileLoading::LoadFile(Data, MIDRPath, 18)) {
-        uint64_t NewMIDR{};
-        std::string_view MIDRView(&Data.at(0), 18);
-        if (FEXCore::StrConv::Conv(MIDRView, &NewMIDR)) {
-          if (MIDR != 0 && MIDR != NewMIDR) {
-            // CPU mismatch, claim hybrid
-            Hybrid = true;
-          }
+    std::string MIDRPath = fmt::format("/sys/devices/system/cpu/cpu{}/regs/identification/midr_el1", i);
 
-          // Truncate to 32-bits, top 32-bits are all reserved in MIDR
-          PerCPUData[i].ProductName = ProductNames::ARM_UNKNOWN;
-          PerCPUData[i].MIDR = NewMIDR;
-          MIDR = NewMIDR;
+    std::array<char, 18> Data;
+    // Needs to be a fixed size since depending on kernel it will try to read a full page of data and fail
+    // Only read 18 bytes for a 64bit value prefixed with 0x
+    if (FEXCore::FileLoading::LoadFileToBuffer(MIDRPath, Data) == sizeof(Data)) {
+      uint64_t NewMIDR{};
+      std::string_view MIDRView(Data.data(), sizeof(Data));
+      if (FEXCore::StrConv::Conv(MIDRView, &NewMIDR)) {
+        if (MIDR != 0 && MIDR != NewMIDR) {
+          // CPU mismatch, claim hybrid
+          Hybrid = true;
         }
+
+        // Truncate to 32-bits, top 32-bits are all reserved in MIDR
+        PerCPUData[i].ProductName = ProductNames::ARM_UNKNOWN;
+        PerCPUData[i].MIDR = NewMIDR;
+        MIDR = NewMIDR;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -77,18 +77,30 @@ static uint32_t GetCPUID() {
...
+  // Gives us a maximum of 999 CPU cores
+  constexpr size_t MAX_SIZE = sizeof(MaxPath);
+  static_assert(MAX_SIZE == 32, "Fixed max");
+#undef CPU_PATH
+
+  for (CPUs = 1; CPUs <= 999; ++CPUs) {
+    fmt::format_to_n(&MaxPath[CPU_PATH_SIZE], 3, "{}", CPUs);
+    if (access(MaxPath, F_OK) != 0) {
+      return CPUs - 1;
+    }
```
neobrain, line 93: Instead of using this much more verbose logic, wouldn't this be a good place to use a static local variable?

```suggestion
  static size_t CPUs = std::invoke([]() {
      size_t CPUs = 1;
      while(std::filesystem::exists("/sys/devices/system/cpu/cpu" + std::to_string(CPUs))) {
        CPUs++;
      }
      return CPUs;
    });
```

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -122,25 +134,24 @@ void CPUIDEmu::SetupHostHybridFlag() {
...
-          // Truncate to 32-bits, top 32-bits are all reserved in MIDR
-          PerCPUData[i].ProductName = ProductNames::ARM_UNKNOWN;
-          PerCPUData[i].MIDR = NewMIDR;
-          MIDR = NewMIDR;
+    std::string MIDRPath = fmt::format("/sys/devices/system/cpu/cpu{}/regs/identification/midr_el1", i);
+
+    char Data[18];
+    // Needs to be a fixed size since depending on kernel it will try to read a full page of data and fail
+    // Only read 18 bytes for a 64bit value prefixed with 0x
+    if (FEXCore::FileLoading::LoadFileToBuffer(MIDRPath, Data, 18)) {
```
neobrain, line 142: DRY here and for MIDRView:
```suggestion
    if (FEXCore::FileLoading::LoadFileToBuffer(MIDRPath, Data, sizeof(Data))) {
```

External/FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -47,5 +49,16 @@ bool LoadFile(std::vector<char> &Data, const std::string &Filepath, size_t Fixed
   return true;
 }
 
+bool LoadFileToBuffer(const std::string &Filepath, char *Buffer, size_t FixedSize) {
```
neobrain, line 52: Can we use `std::span`, yet? Would be nicer than passing pointer+size pairs around.

External/FEXCore/Source/Utils/FileLoading.h
```diff
@@ -14,5 +14,16 @@ namespace FEXCore::FileLoading {
    * @return true on file loaded, false on failure
    */
   bool LoadFile(std::vector<char> &Data, const std::string &Filepath, size_t FixedSize = 0);
+
+  /**
+   * @brief Loads a filepath in to a buffer of data with a fixed size
```
neobrain, line 19: I like the general idea, but not being able to handle files smaller than FixedSize makes this a bad abstraction IMO.

External/FEXCore/Source/Interface/Core/CPUID.h
```diff
@@ -13,6 +13,8 @@ namespace Context {
   struct Context;
 }
 
+
+// #define CPUID_AMD
```
neobrain, line 17: Leftover debug comment?

neobrain (commented):
map->array change looks good!

The CalculateNumberOfCPUs optimization would probably better be done with static variables, and then we can keep that one more readable with `std::fs`.

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Just dropped this change fragment for now.

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Utils/FileLoading.cpp
Sonicadvance1: Switched over to using span.

External/FEXCore/Source/Utils/FileLoading.h
Sonicadvance1: It now returns size read instead of bool.

External/FEXCore/Source/Interface/Core/CPUID.h
Sonicadvance1: Added a comment to this line stating that it is for debugging.
Allows switching between which CPU family FEX represents itself as for testing.

External/FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -47,5 +50,16 @@ bool LoadFile(std::vector<char> &Data, const std::string &Filepath, size_t Fixed
   return true;
 }
 
+ssize_t LoadFileToBuffer(const std::string &Filepath, std::span<char> Buffer) {
+  int FD = open(Filepath.c_str(), O_RDONLY);
+
+  if (FD == -1) {
+    return false;
```
neobrain, line 57:
```suggestion
    return -1;
```

External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -122,25 +121,24 @@ void CPUIDEmu::SetupHostHybridFlag() {
...
-          // Truncate to 32-bits, top 32-bits are all reserved in MIDR
-          PerCPUData[i].ProductName = ProductNames::ARM_UNKNOWN;
-          PerCPUData[i].MIDR = NewMIDR;
-          MIDR = NewMIDR;
+    std::string MIDRPath = fmt::format("/sys/devices/system/cpu/cpu{}/regs/identification/midr_el1", i);
+
+    char Data[18];
+    // Needs to be a fixed size since depending on kernel it will try to read a full page of data and fail
+    // Only read 18 bytes for a 64bit value prefixed with 0x
+    if (FEXCore::FileLoading::LoadFileToBuffer(MIDRPath, std::span{Data, sizeof(Data)}) == sizeof(Data)) {
```
neobrain, line 129:
```suggestion
    std::array<char, 18> Data;
    // Needs to be a fixed size since depending on kernel it will try to read a full page of data and fail
    // Only read 18 bytes for a 64bit value prefixed with 0x
    if (FEXCore::FileLoading::LoadFileToBuffer(MIDRPath, Data) == sizeof(Data)) {
```

(Actually might even work without changing the type of `Data` to `std::array` :) )

neobrain (approved):

External/FEXCore/Source/Utils/FileLoading.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Resolved.

PR #2340, opened 2023-01-14
Config: Remove usage of fs::exists

Sonicadvance1:
We know this is super slow. Remove usages to get a free perf boost.

### DIFF
--- a/External/FEXCore/Source/Interface/Config/Config.cpp
+++ b/External/FEXCore/Source/Interface/Config/Config.cpp
@@ -5,12 +5,12 @@
 
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Utils/LogManager.h>
+#include <FEXHeaderUtils/Filesystem.h>
 
 #include <array>
 #include <assert.h>
 #include <cstdlib>
 #include <filesystem>
-#include <fstream>
 #include <functional>
 #include <map>
 #include <memory>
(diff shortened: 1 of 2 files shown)


neobrain:
Is there a measurable impact to this for Config loading? If yes, what's the chance we could achieve a better improvement by re-opening the configuration less often?

Sonicadvance1:
Since this is all path processing, this had a very small performance improvement. It might help more if we cache application configurations once loaded so we don't duplicate the app profile load twice.
But since application profiles are only available for a small number of application relative to how many we load, it would be less felt.

neobrain:
Can you put numbers on this?

I find `access(ConfigFile.c_str(), F_OK) != 0` significantly harder to make sense of than the self-describing `!std::filesystem::exists(ConfigFile, ec)`. If there is a compelling reason to move away from `std::filesystem` this can be a valid trade-off, but if this becomes a more widely used pattern (which it looks it now is) it may make more sense to provide a drop-in filesystem abstraction instead.

Sonicadvance1:
I already deleted my benching of this, was a couple of percentage. I would need to rerun if you want documented numbers.
We could add a FHU helper for an `FS::Exists` helper that uses access like this.

Sonicadvance1:
Added a new FHU header which has an `Exists` helper and switched over to using it.

neobrain (commented):
(meant to attach these comments to my last comment, woops)

PR #2343, opened 2023-01-16
FEXLoader: Build FEXInterpreter and FEXLoader independently

Sonicadvance1:
This is causing some heartburn with the hardlink.

- Removes some termux cmake list hacking.
- Removes the need to do post-install packaging fixups when hardlinks get dropped.
- Removes a custom uninstall target that was necessary before.

Might want to hide whitespace when reviewing this, it looks like a lot of changes but it is fairly minimal.

### DIFF
--- a/Source/Tests/CMakeLists.txt
+++ b/Source/Tests/CMakeLists.txt
@@ -7,78 +7,47 @@ if (TERMUX_BUILD)
   list(APPEND LIBS android-shmem)
 endif()
 
-add_executable(FEXLoader
-  FEXLoader.cpp
-  VDSO_Emulation.cpp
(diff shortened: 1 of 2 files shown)


neobrain (approved):

neobrain:
Looks fine, but...

> This is causing some heartburn with the hardlink.

What does this mean?

Sonicadvance1:
> Looks fine, but...
> 
> > This is causing some heartburn with the hardlink.
> 
> What does this mean?

FEXInterpreter was setup to be a hardlink to FEXLoader. Which was causing pain in some situations. Thus heartburn.

PR #2345, opened 2023-01-18
Support user supplied signal restorer. 

Sonicadvance1:
This depends on #2327 and #2344 being commited first

With this in-place, we now support libSegFault's backtrace capabilities.

eg 64-bit:
```
Backtrace:
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test(+0x1130)[0x56454f0e5130]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x7fffd9c29d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x7fffd9c29e40]
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test(+0x1065)[0x56454f0e5065]
```

eg 32-bit:
```
Backtrace:
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test_32(+0x1190)[0xaab7b190]
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test_32(+0x11bc)[0xaab7b1bc]
/lib/i386-linux-gnu/libc.so.6(+0x21519)[0xf7821519]
/lib/i386-linux-gnu/libc.so.6(__libc_start_main+0x93)[0xf78215f3]
/mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test_32(+0x108b)[0xaab7b08b]
```

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -72,6 +72,11 @@ jobs:
       # Execute the build.  You can specify a specific target with "--target <NAME>"
       run: cmake --build . --config $BUILD_TYPE
 
+    - name: Install
+      working-directory: ${{runner.workspace}}/build
+      shell: bash
+      run: cmake --build . --config $BUILD_TYPE --target install
+
     - name: ASM Tests
       working-directory: ${{runner.workspace}}/build
       shell: bash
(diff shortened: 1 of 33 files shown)


Sonicadvance1:
Rebased but I want to double check if sigreturn needs to be in VDSO and make a better fallback for if the host doesn't have the 32-bit VDSO installed. Since Arch repos are currently stuck without thunks.

Sonicadvance1:
Rebased and implemented a fallback path for if the 32-bit VDSO thunk library isn't installed. Will ensure that Arch users without thunks can still run 32-bit software without it crashing on signal returns.
LibSegFault happily uses the helper gen code for its backtrace.

lioncash (approved):

PR #2351, opened 2023-01-22
ARMEmitter: Support helper for long address generation 

Sonicadvance1:
The current separated adr and adrp handlers are difficult to use if you
don't know if the resulting address is going to be within 1MB or 4GB.

Adds a `LongAddressGen` helper that will generate the various pieces of
code that will need to be emitted.

Backward labels:
 - Can generate three different code segments depending on distance to
   label
   - adr if label is within 1MB
   - adrp if label is 4K page aligned and within 4GB
   - adrp+add if label is within 4GB

Forward labels:
- Can generate three different code segments depending on distance to
  label
  - nop+adr if label is within 1MB
  - nop+adrp if label is 4K page aligned and within 4GB
  - adrp+add if label is within 4GB

There is still the limitation that this can't generate addresses to
labels that are >4GB away. Which is fine.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ALUOps.inl
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ALUOps.inl
@@ -10,6 +10,16 @@
  * FEX-Emu ALU operations usually have a 32-bit or 64-bit operating size encoded in the IR operation,
  * This allows FEX to use a single helper function which decodes to both handlers.
  */
+private:
+  static bool IsADRRange(int64_t Imm) {
+    return Imm >= -1048576 && Imm <= 1048575;
+  }
+  static bool IsADRPRange(int64_t Imm) {
+    return Imm >= -4294967296 && Imm <= 4294963200;
+  }
+  static bool IsADRPAligned(int64_t Imm) {
+    return (Imm & 0xFFF) == 0;
+  }
 public:
   // PC relative
   void adr(FEXCore::ARMEmitter::Register rd, uint32_t Imm) {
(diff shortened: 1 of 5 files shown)


External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ALUOps.inl
```diff
@@ -10,6 +10,16 @@
  * FEX-Emu ALU operations usually have a 32-bit or 64-bit operating size encoded in the IR operation,
  * This allows FEX to use a single helper function which decodes to both handlers.
  */
+private:
+  bool IsADRRange(int64_t Imm) {
```
lioncash, line 14: seems like these can be const qualified or made static

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Buffer.h
```diff
@@ -87,6 +87,11 @@ namespace FEXCore::ARMEmitter {
         return Size;
       }
 
+      template<typename T>
+      size_t GetCursorOffsetFromAddress(T* Address) {
```
lioncash, line 91:
```suggestion
      size_t GetCursorOffsetFromAddress(const T* Address) const {
```

would make this a little more flexible for use

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ALUOps.inl
Sonicadvance1: You're right, made static.

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Buffer.h
Sonicadvance1: Added const-ness

lioncash (approved):

PR #2358, opened 2023-01-30
Dispathcer: Fixes crash in 32-bit non-rt context generation

Sonicadvance1:
This line was moved lower and this was accidentally leftover while rebasing.
Fixes a crash in 32-bit applications that aren't using SA_SIGINFO.

Unnecessary if #2344 gets merged first.


Sonicadvance1:
No longer necessary since #2344 is merged.

PR #2360, opened 2023-02-01
Arm64/VectorOps: Clamp shift amount to esize-1 for VSShr

lioncash:
Was lying in bed and had a strange feeling I forgot about something, and well...

We need to treat values larger than 31 as if they were 31 bit shifts in order to handle sign-extending behavior properly.

Makes the behavior consistent with the x86 JIT. Also adjusts the unit-test to catch this in the future


Sonicadvance1 (approved):
nice catch!

PR #2361, opened 2023-02-01
Thunks: Fixes host symbol overrides

Sonicadvance1:
1) The host library needs to be loaded in the global namespace.

2) We need to use `RTLD_DEFAULT` instead of querying the object
   directly.

We need to load the host library in the global namespace so the symbols
end up in the global symbol table. This follows how all these symbols
/usually/ get loaded. Either by linking directly to the library or how
loaders will end up loading these.

We need to use RTLD_DEFAULT to follow symbol overriding rules that tend
to occur. For example, MangoHUD will LD_PRELOAD a library that provides
GLX and EGL symbols. Which FEX's thunk libraries need to pick up this
override.
If we are querying the host library directly then we fail to pickup
these overrides, thus breaking MangoHUD and other overlays.

### DIFF
--- a/ThunkLibs/Generator/gen.cpp
+++ b/ThunkLibs/Generator/gen.cpp
@@ -442,7 +442,7 @@ void GenerateThunkLibsAction::ParseInterface(clang::ASTContext& context) {
                 }
 
                 thunked_api.push_back(ThunkedAPIFunction { (const FunctionParams&)data, data.function_name, data.return_type,
-                                                            namespace_info.host_loader.empty() ? "dlsym" : namespace_info.host_loader,
+                                                            namespace_info.host_loader.empty() ? "dlsym_default" : namespace_info.host_loader,
                                                             data.is_variadic || annotations.custom_guest_entrypoint,
                                                             data.is_variadic,
                                                             std::nullopt });
@@ -761,7 +761,11 @@ void GenerateThunkLibsAction::EmitOutput() {
           version_suffix = '.' + std::to_string(*lib_version);
         }
         const std::string library_filename = libfilename + ".so" + version_suffix;
-        file << "  fexldr_ptr_" << libname << "_so = dlopen(\"" << library_filename << "\", RTLD_LOCAL | RTLD_LAZY);\n";
+
+        // Load the host library in the global symbol namespace.
+        // This follows how these libraries get loaded in a non-emulated environment,
+        // Either by directly linking to the library or a loader (In OpenGL or Vulkan) putting everything in the global namespace.
+        file << "  fexldr_ptr_" << libname << "_so = dlopen(\"" << library_filename << "\", RTLD_GLOBAL | RTLD_LAZY);\n";
 
         file << "  if (!fexldr_ptr_" << libname << "_so) { return false; }\n\n";
         for (auto& import : thunked_api) {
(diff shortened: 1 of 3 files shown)


ThunkLibs/Generator/gen.cpp
```diff
@@ -761,12 +761,29 @@ void GenerateThunkLibsAction::EmitOutput() {
...
-        file << "  fexldr_ptr_" << libname << "_so = dlopen(\"" << library_filename << "\", RTLD_LOCAL | RTLD_LAZY);\n";
+
+        // Load the host library in the global symbol namespace.
+        // This follows how these libraries get loaded in a non-emulated environment,
+        // Either by directly linking to the library or a loader (In OpenGL or Vulkan) putting everything in the global namespace.
+        file << "  fexldr_ptr_" << libname << "_so = dlopen(\"" << library_filename << "\", RTLD_GLOBAL | RTLD_LAZY);\n";
 
         file << "  if (!fexldr_ptr_" << libname << "_so) { return false; }\n\n";
         for (auto& import : thunked_api) {
+          if (import.host_loader == "dlsym") {
```
neobrain, line 772: Could this be done by adding a utility function to `Host.h` instead and changing the `host_loader` default to that function?

This avoids specializing code generator logic for specific symbols and hence reduces overall generator complexity.

Sonicadvance1: Changed over to a utility function.

unittests/ThunkLibs/generator.cpp
```diff
@@ -332,7 +332,10 @@ SourceWithAST Fixture::run_thunkgen_host(std::string_view prelude, std::string_v
         "};\n"
         "template<typename F>\n"
         "void FinalizeHostTrampolineForGuestFunction(F*);\n"
-        "struct ExportEntry { uint8_t* sha256; void(*fn)(void *); };\n";
+        "struct ExportEntry { uint8_t* sha256; void(*fn)(void *); };\n"
+        "inline void *dlsym_default(void *__restrict handle, const char *__restrict symbol) {\n"
```
neobrain, line 336: A declaration without definition should suffice here, since we don't involve this function body in AST testing.
```suggestion
        "void *dlsym_default(void* handle, const char* symbol);\n"
```

neobrain (approved):
Added two minor comments, looks good otherwise!

ThunkLibs/include/common/Host.h
```diff
@@ -192,3 +193,14 @@ void FinalizeHostTrampolineForGuestFunction(F* PreallocatedTrampolineForGuestFun
...
 }
+
+// In the case of the thunk host_loader being the default, FEX need to use dlsym with RTLD_DEFAULT.
+// If FEX queried the symbol object directly then it wouldn't follow symbol overriding rules.
+//
+// Common usecase is LD_PRELOAD with a library that defines some symbols.
+// And then programs and libraries will pick up the preloaded symbols.
+// ex: MangoHud overrides GLX and EGL symbols.
+inline
+void *dlsym_default(void *__restrict handle, const char *__restrict symbol) {
```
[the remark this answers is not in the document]
neobrain: I don't see this mattering here, and I'd prefer to keep utility code simple and readable.

If `__restrict` were to make a difference anywhere, we'd get it anyway since `dlsym_default` is trivially inlined and doesn't override `dlsym`'s `__restrict` qualifiers.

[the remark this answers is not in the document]
Sonicadvance1: Sure thing, removed

unittests/ThunkLibs/generator.cpp
Sonicadvance1: Removed the body.

PR #2363, opened 2023-02-02
Config: Fix relative execve applications.

Sonicadvance1:
I made the assumption from some bad historical knowledge that the kernel will canonicalize relative filenames and symlinks for applications that execute through execve.

This turns out to not be true. In fact it passes pathname untouched to the interpreter. So we need to do an additional fix up on relative paths to ensure glibc doesn't break.

Fixes a major bug that breaks a bunch of games.

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -53,11 +53,18 @@ namespace FEX::Config {
       Program += ProgramFDFromEnv;
     }
 
+    // If we were provided a relative path then we need to canonicalize it to become absolute.
+    // If the program name isn't resolved to an absolute path then glibc breaks inside it's `_dl_get_origin` function.
+    // This is because we rewrite `/proc/self/exe` to the absolute program path calculated in here.
+    if (!Program.starts_with('/')) {
+      Program = std::filesystem::canonical(std::move(Program)).string();
+    }
+
     // If FEX was invoked through an FD path (either binfmt_misc or execveat) then we need to check the
     // Program to see if it is a symlink to find the real path.
     //
-    // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
-    //   - Full application path with execve (See Side Note)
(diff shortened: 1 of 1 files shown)


Source/Common/Config.cpp
```diff
@@ -53,11 +53,18 @@ namespace FEX::Config {
...
+    if (!Program.starts_with('/')) {
+      Program = std::filesystem::canonical(Program).string();
+    }
+
     // If FEX was invoked through an FD path (either binfmt_misc or execveat) then we need to check the
     // Program to see if it is a symlink to find the real path.
     //
-    // binfmt_misc: Arg[0] is actually the full application path or `/dev/fd/<FD>` path
-    //   - Full application path with execve (See Side Note)
+    // binfmt_misc: Arg[0] is actually the execve `pathname` or `/dev/fd/<FD>` path
```
neobrain, line 66: It's not really clear what `pathname` refers to.

neobrain (approved):

Source/Common/Config.cpp
Sonicadvance1: Would `execve pathname argument` work?

neobrain: Ah yes, that would be fine.

Sonicadvance1: Switched it to that

Source/Common/Config.cpp
```diff
@@ -53,11 +53,18 @@ namespace FEX::Config {
       Program += ProgramFDFromEnv;
     }
 
+    // If we were provided a relative path then we need to canonicalize it to become absolute.
+    // If the program name isn't resolved to an absolute path then glibc breaks inside it's `_dl_get_origin` function.
+    // This is because we rewrite `/proc/self/exe` to the absolute program path calculated in here.
+    if (!Program.starts_with('/')) {
+      Program = std::filesystem::canonical(Program).string();
```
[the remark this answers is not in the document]
Sonicadvance1: used move

PR #2371, opened 2023-02-04
ARMEmitter: Handle SVE predicated mul/div and finish off integer reduction category

lioncash:
Finishes off MUL/DIV (predicated) and the remaining integer reduction operations.


Sonicadvance1 (approved):
Oops on 64-bit missing. Nice catch.

PR #2375, opened 2023-02-04
ELFCodeLoader: Adds an option to inject libSegFault

Sonicadvance1:
When used in conjuction with #2345 this is a useful way to enable libSegFault in applications using application profiles.

Very useful for applications and games that use launcher scripts that set LD_PRELOAD to nothing prior to launch.

A user was wanting this.


lioncash (approved):

neobrain:
Excellent, thanks for adding this :)

PR #2378, opened 2023-02-06
ARMEmitter: Fixes some warnings that cropped up.

Sonicadvance1:


lioncash:
Oh, what warning pops up with these, so I can better avoid them if I add anything in the future?

Sonicadvance1:
The warnings that cropped up on all of these was `__builtin_assume` expression had side effects which were being discarded. Will be visible on non-assertion builds.

lioncash:
Oh huh, reducing it down to a minimal repro, it creates warnings even when the respective conversion functions are constexpr, but doesn't create any warnings when marked `[[gnu::const, gnu::pure]]`, odd

lioncash (approved):

Sonicadvance1:
Weird. I had also poked at it and didn't find a solution. Not sure why the gnu options change behaviour.

PR #2379, opened 2023-02-06
Syscalls: Renamed fstatat64 to fstatat_64

Sonicadvance1:
Similar to our other syscall conflicts, musl/Alpine Linux has a global define that is conflicting with our name here


Joshua-Ashton (approved):
This fixes the issue.

LGTM!

lioncash (approved):

PR #2380, opened 2023-02-06
Fix SDL2 directfb includes under Alpine Linux

Joshua-Ashton:


Sonicadvance1 (approved):
Thanks for the fix!

Source/Tools/FEXConfig/CMakeLists.txt
```diff
@@ -12,6 +12,9 @@ add_executable(${NAME} ${SRCS})
 target_include_directories(${NAME} PRIVATE ${CMAKE_SOURCE_DIR}/Source/)
 target_include_directories(${NAME} PRIVATE ${CMAKE_SOURCE_DIR}/External/imgui/examples/)
 
+# Fix for SDL2 includes under Alpine Linux.
+target_include_directories(${NAME} PRIVATE /usr/include/directfb/)
```
neobrain, line 16: Did you by any chance check which path of the `TARGET SDL2::SDL2` branch Alpine goes through? Would be suspicious if there was a CMake target on Alpine that doesn't properly pull in required include paths.

Newbytee: Someone tried to fix it, but it wasn't merged: https://github.com/libsdl-org/SDL/pull/6084

Now that SDL 3 has been branched and dropped DirectFB support, it likely won't ever be fixed.

PR #2382, opened 2023-02-07
ARMEmitter: Centralize instruction handling for a few categories

lioncash:
Deduplicates quite a bit of code and centralizes the op handling in one place, as opposed to each individual instruction implementation


Sonicadvance1 (approved):
Nice nice nice.

PR #2387, opened 2023-02-07
Arm64/VectorOps: Use movprfx with VBSL

lioncash:
We can use movprfx here to allow compressing the move and bsl operation together on cpus that can handle it.

Noticed after #2386 was merged


Sonicadvance1 (approved):
Nice

PR #2389, opened 2023-02-07
FEXCore: Removes C wrapper interface

Sonicadvance1:
This has been a long time coming. The C interface has been a thorn in our side for no reason for a long time.

The purpose of this step is to remove the C interface without changing behaviour as much as possible. This means that with this commit there are still some bad practices but the remaining issues will be solved with followup PRs.

Primarily, we still have a `DestroyContext(CTX)` static function which calls the Context implementation's `DestroyContext` and does a raw C++ delete.

Follow up PR will remove that, but I didn't want to touch it yet since it'll require checking to ensure the unique_ptr changes play nice with our allocator hooking. Which this is already a huge PR without trying to change behaviour.


External/FEXCore/include/FEXCore/Core/Context.h
```diff
@@ -73,220 +73,215 @@ namespace FEXCore::Context {
...
+      /**
+       * @brief [[theadsafe]] Checks if the Context is either done working or paused(in the case of single stepping)
+       *
+       * Use this when the context is async running to determine if it is done
+       *
+       * @param CTX the context that we created
+       *
+       * @return true if the core is done or paused
+       */
+      FEX_DEFAULT_VISIBILITY virtual bool IsDone() = 0;
```
lioncash, line 191: Should this be const qualified?

Sonicadvance1: Yes. It wasn't previously const qualified. Switched it over to being const qualified.

lioncash (approved):

PR #2390, opened 2023-02-08
OpcodeDispatcher: Handle VEXTRACTF128/VEXTRACTI128

lioncash:


Sonicadvance1 (dismissed):

Sonicadvance1:
Looks like this test needs to be removed from the fail list? Previously it failed because the instruction was missing.
```
[DEBUG] Invalid or Unknown instruction: VEXTRACTF128 0x0919 0xaab9b0f4
[DEBUG] Couldn't Decode something at 0xaab9b0f4, Started at 0xaab9b0ad
```

Sonicadvance1 (approved):

PR #2395, opened 2023-02-09
Hardcode SDL2 includes only when no CMake target is not available

neobrain:
The CMake target should have all dependency information already included.


Sonicadvance1:
Paging @Joshua-Ashton to see if this still works for them.

neobrain:
Tested in an Alpine podman container. Sadly SDL2's CMake setup just doesn't propagate CFLAGS from dependencies, so this patch won't work :(

PR #2400, opened 2023-02-12
Dispatcher: Support reconstructing RIP from block entry

Sonicadvance1:
This allows us to not update RIP on block entry, but still allow
reconstructing the RIP up until that point.

While still not full RIP reconstruction, this lets us update the signal
context's RIP just like the `x86dec_SynchronizeRIPOnAllBlocks` without
eating the cost of writing to RIP on block entry.

This was tested on a few Proton games that were using the VC setup files to ensure it still worked as expected.

- [X] Retest this with Sonic Mania specifically.

### DIFF
--- a/External/FEXCore/include/FEXCore/Core/CPUBackend.h
+++ b/External/FEXCore/include/FEXCore/Core/CPUBackend.h
@@ -85,6 +85,8 @@ namespace CPU {
     // Any data that is explicitly tied to the JIT code and needs to be cached with it
     // should end up in this data structure.
     struct JITCodeTail {
+      // The total size of the codeblock from [BlockBegin, BlockBegin+Size).
+      size_t Size;
       // RIP that the block's entry comes from.
       uint64_t RIP;
     };
--- a/Data/AppConfig/VC_redist.x64.exe.json
+++ b/Data/AppConfig/VC_redist.x64.exe.json
@@ -1,5 +1,0 @@
-{
-  "Config": {
-    "x86dec_SynchronizeRIPOnAllBlocks": "1"
-  }
-}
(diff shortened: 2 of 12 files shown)


External/FEXCore/include/FEXCore/Core/CPUBackend.h
```diff
@@ -85,6 +85,17 @@ namespace CPU {
     // Any data that is explicitly tied to the JIT code and needs to be cached with it
     // should end up in this data structure.
     struct JITCodeTail {
+      /**
```
Sonicadvance1, line 88: oops, this comment needs to be updated.

Sonicadvance1:
Looks like 32-bit vc_redist doesn't actually fully work with this path. Likely because it needs more synchronization points than block entry.
Right now we have synchronization points on:
1) Block entry (Also multiblock with the INI setting)
2) Syscall JIT exit
3) Callback return
4) Fault generating instruction (int3, ud2, hlt, int1, into, int imm)
5) SIGILL instruction

So this introduces a chance of going back in time if we hit any of the additional RIP setters beyond block entry. We need to have a bit more of a robust RIP reconstruction at each of these points so we don't go backward in time.

I'll take a peak at implementing the remaining few sync points for reconstructing, got some ideas for it.

Sonicadvance1:
Looks like this is a change in behaviour between Wine 6.0.3 and 7.0.
This implementation works fine in 6.0.3, but 7.0 hits an infinite loop.

Sonicadvance1:
Found the few issues that I had in the code and resolved them.
Visual C runtimes now install correctly without application profiles. Probably fixes some other random edge case things.

PR #2404, opened 2023-02-13
IR: Add VDupFromGPR

lioncash:
Allows broadcasting constants into vectors from GPRs.

Resolves the only remaining TODOs within our vector ops.


External/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp
```diff
@@ -108,6 +108,40 @@ DEF_OP(VCastFromGPR) {
...
+                      "Unexpected {} element size: {}", __func__, ElementSize);
+
+  const auto SubEmitSize =
+    ElementSize == 8 ? ARMEmitter::SubRegSize::i64Bit :
+    ElementSize == 4 ? ARMEmitter::SubRegSize::i32Bit :
+    ElementSize == 2 ? ARMEmitter::SubRegSize::i16Bit :
+    ElementSize == 1 ? ARMEmitter::SubRegSize::i8Bit : ARMEmitter::SubRegSize::i8Bit;
+
+  if (HostSupportsSVE && Is256Bit) {
+    if (SubEmitSize == ARMEmitter::SubRegSize::i64Bit) {
```
Sonicadvance1, line 131: I don't believe you need to check SubEmitSize here for GPR register size. `dup` emitter op should take a regular `Register` type and not care about GPR register type.

lioncash: Oh, true. Amended it

External/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp
```diff
@@ -108,6 +108,36 @@ DEF_OP(VCastFromGPR) {
...
+    ElementSize == 8 ? ARMEmitter::SubRegSize::i64Bit :
+    ElementSize == 4 ? ARMEmitter::SubRegSize::i32Bit :
+    ElementSize == 2 ? ARMEmitter::SubRegSize::i16Bit :
+    ElementSize == 1 ? ARMEmitter::SubRegSize::i8Bit : ARMEmitter::SubRegSize::i8Bit;
+
+  if (HostSupportsSVE && Is256Bit) {
+    dup(SubEmitSize, Dst.Z(), Src);
+  } else {
+    if (SubEmitSize == ARMEmitter::SubRegSize::i64Bit) {
+      dup(SubEmitSize, Dst.Q(), Src.X());
```
Sonicadvance1, line 134: Same thing here for the non-SVE version.

Sonicadvance1 (approved):
You can see how I designed the emitter to remove as many branches as possible :)

PR #2414, opened 2023-02-15
Arm64/VectorOps: Use SVE only with 256-bit op sizes

lioncash:
Keeps all of the IR ops consistent with each other in terms of behavior in relation to op sizes. Also removes some redundant scalar checks that weren't really necessary.


Sonicadvance1 (approved):
Very nice cleanup

PR #2416, opened 2023-02-15
VectorOps: Remove unnecessary mov in VUShrNI2/VSQXTN2/VSQXTUN2

lioncash:
We can move the initial move down by SPLICE, which not only lets us turn it into a MOVPRFX, but also we can safely move into the final destination register directly, since we can be sure there's no potential dependencies at this point


Sonicadvance1 (approved):
Nice

PR #2423, opened 2023-02-20
OpcodeDispatcher: Handle VPERMD/VPERMPS

lioncash:
Handles the 32-bit non-in-lane permutation instructions.

To make this not ugly to implement, we add two other IR operations (`VTrn` and `VTrn2`), which act like, well, TRN1 and TRN2. We also extend `VShlI` on the x86 side to handle 1-byte element sizes.

When reading the documentation text for `VPERMD` op, you may want to open it directly, since GitHub will probably mangle the vector diagrams in the text.


Sonicadvance1 (approved):
Very nice comment documentation!

PR #2424, opened 2023-02-21
OpcodeDispatcher: Handle register variants of VPERMILPD/VPERMILPS

lioncash:


Sonicadvance1 (approved):
Such a wacky instruction

PR #2425, opened 2023-02-21
VEXTables: Remove VPERMIL2PD and VPERMIL2PS entries

lioncash:
These are actually XOP instructions. That, despite being so, are encoded using a VEX prefix.

e.g. from the AMD manual:

image

image


Sonicadvance1 (approved):
Rude!

PR #2426, opened 2023-02-21
FileManagement: Optimize GetEmulatedFDPath with an FD!

Sonicadvance1:
Performance stats up front:
This improves pressure-vessel startup time on my test device by 10.1% Improving the startup time from 9.71425 seconds to 8.7421 seconds.

Most filesystem based syscalls support a file descriptor version with an *at suffix. This allows us to do these syscalls with pathnames that are relative to the directory FD that is passed to the syscall.

This is pretty much exactly what we want when we are searching for files inside of our rootfs. The only quirk ends up being that we are getting passed absolute paths. This ends up being very simple to workaround by stripping off the front '/' character. Doing this is just offsetting the pointer passed to the syscall by one byte.

This does require having two temporary buffers of size PATH_MAX passed to the handler since just like in the other implementation, we need to keep the previous result around. The difference being now that we aren't doing a bunch of std::string temporary manipulation and now we are returning one of the passed in buffers back depending on the result.


Source/Tests/LinuxSyscalls/FileManagement.h
```diff
@@ -78,6 +80,7 @@ class FileManager final {
   void UpdatePID(uint32_t PID) { CurrentPID = PID; }
 
   std::string GetEmulatedPath(const char *pathname, bool FollowSymlink = false);
+  std::optional<std::pair<int, const char*>> GetEmulatedFDPath(const char *pathname, bool FollowSymlink, std::array<char[PATH_MAX], 2> &TmpFilename);
```
lioncash, line 83: perhaps the

```cpp
std::array<char[PATH_MAX], 2>
```

should be extracted into a `using` alias, so it doesn't need to written out long-form in several cases and keeps the underlying type definition in one place

Looks good otherwise

Sonicadvance1: Switched over to a `using` alias

lioncash (approved):

PR #2429, opened 2023-02-22
OpcodeDispatcher: Handle alignment for MOVAPS a little better

lioncash:
IIRC: 

0x10/0x11 in the two byte op table corresponds to MOVUPS
0x28/0x29 in the two byte op table corresponds to MOVAPS

so we were treating both MOVAPS encodings as MOVUPS, unless I'm reading the table wrong.


Sonicadvance1 (approved):
Ah yea. We aren't doing optimizations when we know there are explicit alignment requirements so this slipped through.
Probably explains when I did some simple tests that enforced alignment I was getting unexpected alignment faults.

Sonicadvance1 (approved):
Makes sense, bit of quirkiness with the MOVLHPS things :)

PR #2438, opened 2023-02-23
OpcodeDispatcher: Restrict partial XMM stores to FPRs in StoreResult_WithOpSize

lioncash:
As far as I know, nothing actually uses this path. Partially resolves the TODO of dealing with partial writes.


Sonicadvance1 (approved):
I guess we did manage to fix all users of this. Nice!

PR #2440, opened 2023-02-25
FEXServer: Add backtrace support to Coredump service

Sonicadvance1:
Relies on #2432 to be merged first

ELFMapping needs to be cleaned up and consolidated with the rest of our
ELF loading.

This gives us a backtrace of the guest in the cooperative backtrace service.

eg:
```bash
        PID: 111969
        TID: 111969
        UID: 1000 (ryanh)
        GID: 1000 (ryanh)
     Signal: 11 (Segmentation fault)
  Host Arch: x86_64
 Guest Arch: x86_64
  Timestamp: Sat 2023-02-25 01:52:56 PST
CommandLine: /mnt/Work/Work/work/FEXNew/Build/Bin/FEXInterpreter /mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test
Host Program counter: 0x7fffdaac38fd
Thread's Dispatcher: [0x7ffff7fb9000, 0x7ffff7fba000) -> 0x1000 bytes
Thread's JIT regions: 1
  JIT regions 0: [0x7fffdaa00000, 0x7fffdba00000) -> 0x1000000 bytes *
@@ Crash was inside FEX's JIT. Highly likely a guest side crash.

                Found module /mnt/Work/Work/work/FEXNew/Build/Bin/FEXInterpreter
                Found module /mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/sigsegv_test
                Found module /usr/lib/x86_64-linux-gnu/libc.so.6
                Found module /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
                Found module /usr/lib/x86_64-linux-gnu/libc.so.6
                Found module /usr/lib/x86_64-linux-gnu/libm.so.6
                Found module /usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.30
                Found module /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
                Found module /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
                Found module /usr/lib/x86_64-linux-gnu/libm.so.6
                Found module /usr/lib/x86_64-linux-gnu/libgcc_s.so.1
                Found module /usr/share/fex-emu/GuestThunks/libVDSO-guest.so
                Found module /usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
Register data
RIP: 000055a0b7d7b130
RAX: 000055a0b7d7b140 RBX: 0000000000000000 RCX: 000055a0b7d7ddf0 RDX 00007ffff6ffec36
RSI: 00007ffff6ffec26 RDI: 0000000000000000 RSP: 00007ffff6ffeb08 RBP 0000000000000001
R8:  00007fffd9c1b030 R9:  00007ffff7e4e040 R10: 00007ffff7e48908 R11 00007ffff7e63680
R12: 00007ffff6ffec26 R13: 000055a0b7d7b140 R14: 000055a0b7d7ddf0 R15 00007ffff7e82040

                Stack trace of thread 111969:
                #0  0x000055a0b7d7b130 <_Z10StoreValuePc>               sp=(0x00007ffff6ffeb08)
```


Source/Tools/FEXServer/Unwind/ELFMapping.cpp
```diff
@@ -426,4 +389,12 @@ namespace ELFMapping {
     return nullptr;
   }
 
+  ELFMemMapping::~ELFMemMapping() {
+    if (BaseRemapped) {
+      // If the FD was mapped then make sure to unmap it.
+      munmap(BaseRemapped, TotalSize);
```
neobrain, line 395: I think adding a destructor still leaves the implicit copy constructor around, so we should explicitly delete it to avoid accidental double-unmaps here.

Source/Tools/FEXServer/Unwind/FileMapping.h
```diff
@@ -1,20 +1,19 @@
...
-
 namespace FileMapping {
   struct MemMapping;
 
   struct FileMapping {
     uint64_t Begin, End;
     std::string Path;
     std::vector<MemMapping*> MemMappings;
-    ELFMapping::ELFMemMapping *ELFMapping{};
+    std::unique_ptr<ELFMapping::ELFMemMapping> ELFMapping;
```
neobrain, line 16: Looks like `std::optional` might be the more fitting abstraction here now, since we don't need heap allocations or ownership transfer.

Source/Tools/FEXServer/FDCountWatch.h
```diff
@@ -0,0 +1,7 @@
+#pragma once
+#include <sys/types.h>
+
+namespace FDCountWatch {
```
neobrain, line 4: Should probably have a docstring what exactly is being watched here and how and why.

Source/Tools/FEXServer/FDCountWatch.cpp
```diff
@@ -0,0 +1,76 @@
+#include <atomic>
+#include <cstddef>
+#include <cstdint>
+#include <cstring>
+#include <filesystem>
+#include <mutex>
+#include <sys/resource.h>
+
+namespace FDCountWatch {
+  // FD count watching
```
neobrain, line 10: Redundant comment:
```suggestion
```

Source/Tools/FEXServer/FDCountWatch.cpp
```diff
@@ -0,0 +1,76 @@
...
+  constexpr size_t static MAX_FD_DISTANCE = 32;
+  rlimit MaxFDs{};
+  std::atomic<ssize_t> NumFilesOpened{};
+  std::mutex FDLimitMutex;
```
neobrain, line 14: These should probably be at least static.

Source/Tools/FEXServer/FDCountWatch.cpp
```diff
@@ -0,0 +1,76 @@
...
+      // No need to raise the limit.
+      return;
+    }
+
+    if (FDLimitMutex.try_lock()) {
+      if (MaxFDs.rlim_cur == MaxFDs.rlim_max) {
+        fprintf(stderr, "[FEXMountDaemon] Our open FD limit is already set to max and we are wanting to increase it\n");
+        fprintf(stderr, "[FEXMountDaemon] FEXMountDaemon will now no longer be able to track new instances of FEX\n");
+        fprintf(stderr, "[FEXMountDaemon] Current limit is %zd(hard %zd) FDs and we are at %zd\n", MaxFDs.rlim_cur, MaxFDs.rlim_max, GetNumFilesOpen());
+        fprintf(stderr, "[FEXMountDaemon] Ask your administrator to raise your kernel's hard limit on open FDs\n");
```
neobrain, line 46: To be more user friendly, this should also mention what impact it will have on "normal" FEX usage. If the average user is unlikely to care, we might even turn most of the existing message into a source comment.

Source/Tools/FEXServer/FDCountWatch.cpp
```diff
@@ -0,0 +1,76 @@
...
+        return;
+      }
+
+      rlimit NewLimit = MaxFDs;
+
+      // Just multiply by two
+      NewLimit.rlim_cur <<= 1;
```
neobrain, line 53:
```suggestion
      NewLimit.rlim_cur *= 2;
```

Source/Tools/FEXServer/FDCountWatch.cpp
```diff
@@ -0,0 +1,76 @@
...
+      return;
+    }
+
+    if (FDLimitMutex.try_lock()) {
+      if (MaxFDs.rlim_cur == MaxFDs.rlim_max) {
+        fprintf(stderr, "[FEXMountDaemon] Our open FD limit is already set to max and we are wanting to increase it\n");
+        fprintf(stderr, "[FEXMountDaemon] FEXMountDaemon will now no longer be able to track new instances of FEX\n");
+        fprintf(stderr, "[FEXMountDaemon] Current limit is %zd(hard %zd) FDs and we are at %zd\n", MaxFDs.rlim_cur, MaxFDs.rlim_max, GetNumFilesOpen());
+        fprintf(stderr, "[FEXMountDaemon] Ask your administrator to raise your kernel's hard limit on open FDs\n");
+        return;
```
neobrain, line 47: Looks like this will leave `FDLimitMutex` locked permanently.

Source/Tools/FEXServer/FDCountWatch.h
```diff
@@ -0,0 +1,7 @@
...
+  void GetMaxFDs();
+  void IncrementFDCountAndCheckLimits(ssize_t Num);
```
neobrain, line 6: Should have a comment mentioning `Num` may be negative, since the name suggests it shouldn't.

Actually might be better to just duplicate this interface, adding an explicit `DecrementFDCount` that negates its argument and forwards to a common internal-only function. This would get rid of the strangely looking `Increment...(-thing.size() - 1)` calls entirely.

Source/Tools/FEXServer/CoreDumpService.cpp
```diff
@@ -321,6 +322,19 @@ namespace CoreDumpService {
   }
 
   void CoreDumpClass::HandleSocketData() {
+    auto PeekMem = [&](uint64_t Addr, uint8_t Size) -> uint64_t {
+      return PeekMemory(ServerSocket, Addr, Size);
+    };
+    auto GetFD = [&](std::string const *Filename) -> int {
```
neobrain, line 328:
```suggestion
    auto GetFD = [this](std::string const *Filename) -> int {
```

Source/Tools/FEXServer/Unwind/Unwind_x86.cpp
```diff
@@ -0,0 +1,132 @@
...
+#include <list>
+#include <unistd.h>
+#include <signal.h>
+#include <string>
+#include <sys/stat.h>
+#include <unistd.h>
+#include <vector>
+
+namespace Unwind::x86 {
+#define panic(args...) \
```
neobrain, line 29: `panic` suggests fatal error that causes FEX to exit immediately. Something like `print_error` would be more fitting.

Source/Tools/FEXServer/Unwind/Unwind_x86_64.cpp
```diff
@@ -0,0 +1,155 @@
...
+        .access_reg             = UnwinderFuncs::access_reg,
+        .access_fpreg           = UnwinderFuncs::access_fpreg,
+        .resume                 = UnwinderFuncs::resume,
+        .get_proc_name          = UnwinderFuncs::get_proc_name,
+      };
+  };
+
+#define ACCESSOR_IMPL
+#define UNWINDER_TYPE Unwinder_x86_64
+#include "Unwind/Unwind_accessors.inl"
```
neobrain, line 148: Is this really necessary? Wouldn't `virtual` member functions in `Unwinder` allow for using a single set of accessors?

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
```diff
@@ -0,0 +1,429 @@
...
+    Mapping->SectionHeaders.resize(Mapping->Header._64.e_shnum);
+    Mapping->ProgramHeaders.resize(Mapping->Header._64.e_phnum);
+
+    const Elf64_Shdr *RawShdrs =
+      reinterpret_cast<const Elf64_Shdr *>(&RawBase[Mapping->Header._64.e_shoff]);
+    const Elf64_Phdr *RawPhdrs =
+      reinterpret_cast<const Elf64_Phdr *>(&RawBase[Mapping->Header._64.e_phoff]);
+
+    if (Mapping->Header._64.e_shoff > Mapping->TotalSize) {
+      __builtin_unreachable();
```
neobrain, line 56: Looks like this could be reached if the input is erroneous/corrupted. `assert` or another panic-ing function would be more fitting.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
```diff
@@ -0,0 +1,429 @@
...
+    // Calculate symbols
+    Elf32_Shdr const *SymTabHeader{nullptr};
+    Elf32_Shdr const *StringTableHeader{nullptr};
+    char const *StrTab{nullptr};
+
+    Elf32_Shdr const *DynSymTabHeader{nullptr};
+    Elf32_Shdr const *DynStringTableHeader{nullptr};
+    char const *DynStrTab{nullptr};
+
+    for (uint32_t i = 0; i < Mapping->SectionHeaders.size(); ++i) {
```
neobrain, line 221: This can be a range-for loop instead.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
```diff
@@ -0,0 +1,429 @@
...
+    if (Bits != 0) {
+      if (Bits == 64) {
+        ParseELFSymbols64(Mapping);
+      }
+      if (Bits == 32) {
+        ParseELFSymbols32(Mapping);
+      }
+
+      return;
+    }
```
neobrain, line 348:
```suggestion
if (Bits == 64) {
...
} else if (Bits == 32) {
...
}

File: Source/Tools/FEXServer/Unwind/ELFMapping.cpp [hunk 16]
```diff
@@ -0,0 +1,429 @@
...
+    auto Sym = Mapping->SymbolMapByAddress.upper_bound(Addr);
+    if (Sym != Mapping->SymbolMapByAddress.begin()) {
+      --Sym;
+    }
+
+    if (Sym == Mapping->SymbolMapByAddress.end()) {
+      return nullptr;
+    }
+
+    // <= necessary for single byte sized instructions, like thunks.
```
neobrain, line 385: Is this right? In that case, `Addr` would still equal `Sym->second->Address` and hence be strictly smaller than `Sym->second->Address + Sym->second->Size`, wouldn't it?

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
Sonicadvance1: Removed the copy constructor.

Source/Tools/FEXServer/Unwind/FileMapping.h
Sonicadvance1: Switched to optional.

Source/Tools/FEXServer/FDCountWatch.cpp
Sonicadvance1: Removed.

Source/Tools/FEXServer/FDCountWatch.h
Sonicadvance1: Added a comment block about why we need to watch FDs.

Source/Tools/FEXServer/FDCountWatch.cpp
Sonicadvance1: Switched to static.

Source/Tools/FEXServer/FDCountWatch.cpp
Sonicadvance1: Added an additional line  to alert the user that additional instances of FEX may hang.

Source/Tools/FEXServer/FDCountWatch.cpp
Sonicadvance1: Changed.

Source/Tools/FEXServer/FDCountWatch.cpp
Sonicadvance1: Fixed.

Source/Tools/FEXServer/CoreDumpService.cpp
Sonicadvance1: Resolved.

Source/Tools/FEXServer/CMakeLists.txt
```diff
@@ -1,19 +1,59 @@
...
+  ${CMAKE_SOURCE_DIR}/External/libunwind/include/
+  ${CMAKE_SOURCE_DIR}/External/libunwind/include/tdep-x86_64/
+  ${CMAKE_BINARY_DIR}/unwind_x86_64/include/
+  )
+
+target_link_libraries(fex_unwind_x86_64
+  unwind-x86_64
+  FEXCore
+  fmt)
+target_compile_options(fex_unwind_x86_64 PRIVATE -fcommon)
```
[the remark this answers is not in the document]
Sonicadvance1: Thought it was required but it seems not, removed.

Source/Tools/FEXServer/Unwind/Unwind_x86.cpp
Sonicadvance1: Removed the panics and just printed the errors instead, which is the better outcome. Since the rest of the code will safely early exit or continue.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
Sonicadvance1: Returned early safely so we don't bring down the FEXServer on ELF parsing.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
Sonicadvance1: Resolved.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
Sonicadvance1: Resolved.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
Sonicadvance1: Weirdly yes otherwise Thunk entrypoints won't show up in the backtrace.

Source/Tools/FEXServer/Unwind/Unwind_x86_64.cpp
Sonicadvance1: We can't use a single set of accessors because type definitions change sizes between x86 and x86-64.
Easiest thing to see is that `unw_word_t` is 4 or 8 bytes depending on target remote archirtecture.

neobrain: We only compile a single set of these accessors on a given architecture anyway, so the type differences don't really matter. (If they did in the future, we could abstract over them easily since `unw_word_t` still always fits into 8 bytes.)

The reason I propose `virtal` member functions is so we can call e.g. `BacktraceHeader` from an `Unwinder*` pointer while `Unwinder_x86` is not defined yet. Specifically, this is what `Unwinder` would look like:
```cpp
class Unwinder {
    virtual void SetReg(unw_regnum_t regnum, unw_word_t Value);

    virtual void BacktraceHeader() = 0;

    void Backtrace() {
        // ...
    }
};
```

The functions in `Unwind_accessors.inl` can then be defined purely in terms of `Unwinder` functions. We wouldn't need X-macros and the `.inl` file would just be a normal header.

Source/Tools/FEXServer/Unwind/ELFMapping.cpp
neobrain: Is there a coherent explanation we can find for this? Sounds like this is masking a bug elsewhere.

PR #2442, opened 2023-02-27
Dispatcher: Fixes crash with misalign stack returning from signal

Sonicadvance1:
When we were taking a signal that had a misaligned stack, we would store the host stack at a weird offset.

After that point when we were trying to sigreturn we wouldn't know the alignment of the stack coming back and we would try loading the host stack from the wrong offset. Easy fix is to just align the host stack location.

Fixes Ender Lilies, which was consistently crashing from a SIGCHLD due to having a misaligned stack.

Side-change: Move the cookie check to the start of the restore. Doesn't make sense to check the cookie after restoring state since it could be quite wrong.


lioncash (approved):
ooh, nice catch!

PR #2444, opened 2023-02-27
ARMEmitter: Handle SVE Integer Compare - Scalars group

lioncash:


Sonicadvance1 (approved):

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/SVEOps.inl
```diff
@@ -1419,11 +1419,76 @@ public:
 
   // SVE Integer Compare - Scalars
   // SVE integer compare scalar count and limit
-  // XXX:
+  template <typename T>
+  requires IsXOrWRegister<T>
+  void whilege(SubRegSize size, PRegister pd, T rn, T rm) {
```
neobrain, line 1424: FWIW these could be written somewhat more concisely now:

```suggestion
template<IsXOrWRegister T>
void whilege(SubRegSize size, PRegister pd, T rn, T rm) {
```

We also have some instances where we do an `std::is_same` check, which could just be a regular function instead of a template (unless these must be templates for some other reason).

lioncash: Oh cool, I didn't know that

> We also have some instances where we do an std::is_same check, which could just be a regular function instead of a template (unless these must be templates for some other reason).

Yup. I brought that up in the discussion call yesterday. I just haven't gotten around to fixing those yet.

PR #2446, opened 2023-02-28
ARMEmitter: Simplify advanced SIMD copy

lioncash:
Same behavior, but collapses some if statements.


Sonicadvance1 (approved):
Nice. I love the if statement collapsing.

PR #2448, opened 2023-02-28
EmulatedFiles: Optimize openat handler

### LINKED ISSUE #2443: EmulatedFiles: High CPU usage in canonicalize
https://github.com/FEX-Emu/FEX/blob/f2aa0026b58c9ec5ee56ecbf66b4f00fe01e16de/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp#L744

I was running AntiChamber and I saw >30% CPU time spent on this canonicalize routine during loading (which took hecking long).
We need to optimize this use case. Dropping to `realpath` is likely a good first step. Will need to run some benchmarks to see how much of an improvement that alone is and if it is worth going further.

Sonicadvance1:
Fixes #2443
I found out with some profiling that this we were spending a decent amount of time with the `openat` syscall in heavily utilized situations. While not super common in active gameplay situations, it matters significantly in loading screens that this is fairly optimal.

The bulk of the time is spent in the emulated files handler to ensure that whatever path we are given, we can capture file paths that we need to emulate. The largest contributor being the std::filesystem::canonical function call.

A couple of optimizations in place here.
1) Do a quick hashmap check right at the start to see if we exactly fit
2) Change from `std::fs::canonical` to `realpath`
3) Switch `GetEmulatedFDPath` to not use optional so it stops building
   on the stack

I'm still not super happy with the performance of `realpath` and also not happy that we still need to use `lexically_normal` in one code path. But short of writing a super hand-optimized `realpath` that fits our constraints, I don't think we can do better.

Micro benchmark needs to test four different situations due to this optimization.
1) Non-EmuFD path
2) Non-EmuFD path with dirfs
3) EmuFD path
4) EmuFD path with dirfs

And the performance improvement for each situation respectively
1) 12% performance improvement
   - 213413 openat syscalls/s -> 238999 syscalls/s
2) 17% performance improvement
   - 202085 openat syscalls/s -> 237309 syscalls/s
3) 17% performance improvement (/proc/cpuinfo)
   - 56616 openat syscalls/s -> 66231 syscalls/s
   - Includes overhead of generating temp FD and close syscall
4) 5% performance improvement (/proc/cpuinfo)
   - 51080 openat syscalls/s -> 53956 syscalls/s
   - Includes overhead of generating temp FD and close syscall

And for sake of comparison to the non-emulated system; My test system can hit around 1-1.1 million openat syscalls per second in the same microbench.

Nice little performance uplift.

### DIFF
--- a/Source/Common/FDUtils.h
+++ b/Source/Common/FDUtils.h
@@ -23,4 +23,10 @@ std::optional<std::string> get_fdpath(int fd) {
   return std::nullopt;
 }
 
+inline
+int get_fdpath(int fd, char *SymlinkPath) {
+  auto Path = fmt::format("/proc/self/fd/{}", fd);
+  return readlinkat(AT_FDCWD, Path.c_str(), SymlinkPath, PATH_MAX);
+}
+
 }
--- a/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
+++ b/Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
@@ -711,48 +711,68 @@ namespace FEX::EmulatedFile {
   }
 
   int32_t EmulatedFDManager::OpenAt(int dirfs, const char *pathname, int flags, uint32_t mode) {
-    std::string Path{};
-    if (((pathname && pathname[0] != '/') || // If pathname exists then it must not be absolute
-        !pathname) &&
-        dirfs != AT_FDCWD) {
-      // Passed in a dirfd that isn't magic FDCWD
-      // We need to get the path from the fd now
-      Path = FEX::get_fdpath(dirfs).value_or("");
+    char Tmp[PATH_MAX];
+    const char *Path{};
 
-      if (pathname) {
-        if (!Path.empty()) {
-          // If the path returned empty then we don't need a separator
-          Path += "/";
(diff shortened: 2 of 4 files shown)


Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.h
```diff
@@ -30,7 +30,7 @@ namespace FEX::EmulatedFile {
       std::once_flag cpu_info_initialized{};
       std::string cpu_info{};
       using FDReadStringFunc = std::function<int32_t(FEXCore::Context::Context *ctx, int32_t fd, const char *pathname, int32_t flags, mode_t mode)>;
-      std::unordered_map<std::string, FDReadStringFunc> FDReadCreators;
+      std::unordered_map<std::string_view, FDReadStringFunc> FDReadCreators;
```
neobrain, line 33: There is one dynamic key (`procAuxv`) used with this, so a `string_view` creates a dangling reference on that one.

Sonicadvance1: Oop, you're right. dropped back to `std::string`

lioncash (approved):

PR #2451, opened 2023-03-01
ARMEmitter: Simplify size handling Advanced SIMD 3 different group

lioncash:
A large amount of size handling in this category is just decrementing the size by 1, so we can tidy up a bunch of conditionals by just doing that instead.


Sonicadvance1 (approved):
Nice nice nice

PR #2454, opened 2023-03-01
ARMEmitter: Remove most implicit conversion operators for vector register types

lioncash:
These can create conversion ambiguity compiler errors and in the worst case, cause the wrong thing to happen. So we can get rid of all the implicit conversion operators, except for the the one to `VRegister`, since it's supposed to be the catch-all type.

For the other cases, we already have explicit conversion functions that can be used instead to specify behavior.

This also catches a few cases where the wrong conversion functions were being called, but happened to work, since they'd implicitly be converted to the right type.


Sonicadvance1 (approved):
One of those things that seemed like a good idea during initial prototyping and became significantly less great a thousand instructions in.

PR #2455, opened 2023-03-02
ARMEmitter: Remove predicate uint32_t conversion operators

lioncash:
Adds defaulted comparison operators to the interface, which allows comparisons to be done without directly accessing the index. With these in place, the implicit conversion to uint32_t operators can be removed.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Registers.h
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Registers.h
@@ -1,5 +1,8 @@
 #pragma once
+
 #include <FEXCore/Utils/EnumUtils.h>
+
+#include <compare>
 #include <cstdint>
 
 namespace FEXCore::ARMEmitter {
@@ -14,6 +17,8 @@ namespace FEXCore::ARMEmitter {
       Register() = delete;
       constexpr explicit Register(uint32_t Idx)
         : Index {Idx} {}
+
+      friend constexpr auto operator<=>(const Register&, const Register&) = default;
 
       uint32_t Idx() const {
         return Index;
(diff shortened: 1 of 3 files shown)


Sonicadvance1 (approved):

External/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/Registers.h
```diff
@@ -15,6 +18,8 @@ namespace FEXCore::ARMEmitter {
       constexpr explicit Register(uint32_t Idx)
         : Index {Idx} {}
 
+      friend constexpr auto operator<=>(const Register&, const Register&) = default;
```
neobrain, line 21: What use does `friend` have in this case?

lioncash: In the event we add comparisons with other register types (for whatever reason), then it stays consistent, since

```cpp
friend bool operator==(OtherRegisterType, RegisterType) { /* ... */ }
```
and
```cpp
friend bool operator==(RegisterType, OtherRegisterType) { /* ... */ }
```

would need to be handled. There's also nothing necessitating the comparison be bound to the class, so friend it was.

neobrain: > In the event we add comparisons with other register types (for whatever reason), then it stays consistent

Fair.

> There's also nothing necessitating the comparison be bound to the class, so friend it was.

Does `friend` have a difference as used here versus not using `friend`? This sentence suggests it does, but I don't see it.

lioncash: what I meant was, that coupled with my first point, I didn't really see any necessitating reason to tie the function to the class as a member function as opposed to a free function

neobrain: Ah, right. I guess I'm mostly used to seeing such operators defined out-of-class rather than in-class, but thanks for clearing it up!

lioncash: Oh, no prob!

PR #2460, opened 2023-03-02
OpcodeDispatcher: Optimize REP STOS to MemSet operation

Sonicadvance1:
x86's REP STOS instruction is a memset (with element size!) with the
ability to choose a direction of execution.
Additionally it has a feature where if it faults part-way through the
copy, an application can catch the fault and continue afterwards to know
how many bytes got copied.

RCX is the counter which decrements for each element, and RDI is the
memory pointer. On fault these will reflect the last location that was
attempted to be written. FEX doesn't support this behaviour which makes
our lives easier.

Without supporting that feature, this turns in to a directional memset
by element size. Let's remove all the multiple blocks and just emit a
single IR operation to improve performance of the JIT.
Our generated code here was terrible, the IR was terrible, multiblock is
always slow with RA. Just a general overall improvement.

With profiling pressure-vessel this change deletes the hottest block that
appeared in the trace. This instruction is very commonly used for
memsetting a region to zero so it should be quite fast.

We can also optimize REP MOVS in the future with a Memcpy IR operation
in a similar fashion.

Additionally in the future these can be optimized to use ARM's new MOPS
instructions since the most common case is memset by byte. Which is when
we should expose the "Fast REP STOS" CPUID bit. Both setp/setm/sete and
cpyfp/cpyfm/cpyfe match `REP STOS` and `REP MOVS` respectively.

This shaves a decent amount of time off of pressure-vessel initialization time, but I'll need to spend time doing some microbenching to see the exact impact. Also testing real games to see how much they improve.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
@@ -288,6 +288,111 @@ DEF_OP(StoreMem) {
   }
 }
 
+DEF_OP(MemSet) {
+  const auto Op = IROp->C<IR::IROp_MemSet>();
+  const int32_t Size = Op->Size;
+
+  char *MemData = *GetSrc<char **>(Data->SSAData, Op->Addr);
+  const auto Value = *GetSrc<uint64_t*>(Data->SSAData, Op->Value);
+  const auto Length = *GetSrc<uint64_t*>(Data->SSAData, Op->Length);
+  const auto Direction = *GetSrc<uint8_t*>(Data->SSAData, Op->Direction);
+
+  auto MemSetElements = [](auto* Memory, uint64_t Value, size_t Length) {
+    for (size_t i = 0; i < Length; ++i) {
+      Memory[i] = Value;
+    }
+  };
+
+  auto MemSetElementsInverse = [](auto* Memory, uint64_t Value, size_t Length) {
+    for (size_t i = 0; i < Length; ++i) {
+      Memory[-i] = Value;
+    }
+  };
+
+  if (Direction == 0) { // Forward
+    if (Op->IsAtomic) {
+      switch (Size) {
+        case 1:
+          MemSetElements(reinterpret_cast<std::atomic<uint8_t>*>(MemData), Value, Length);
+          break;
+        case 2:
+          MemSetElements(reinterpret_cast<std::atomic<uint16_t>*>(MemData), Value, Length);
+          break;
+        case 4:
+          MemSetElements(reinterpret_cast<std::atomic<uint32_t>*>(MemData), Value, Length);
+          break;
+        case 8:
+          MemSetElements(reinterpret_cast<std::atomic<uint64_t>*>(MemData), Value, Length);
+          break;
+        default:
+          LOGMAN_MSG_A_FMT("Unhandled {} size: {}", __func__, Size);
+          break;
+      }
+    }
+    else {
+      switch (Size) {
+        case 1:
+          MemSetElements(reinterpret_cast<uint8_t*>(MemData), Value, Length);
+          break;
+        case 2:
+          MemSetElements(reinterpret_cast<uint16_t*>(MemData), Value, Length);
+          break;
+        case 4:
+          MemSetElements(reinterpret_cast<uint32_t*>(MemData), Value, Length);
+          break;
+        case 8:
+          MemSetElements(reinterpret_cast<uint64_t*>(MemData), Value, Length);
+          break;
+        default:
+          LOGMAN_MSG_A_FMT("Unhandled {} size: {}", __func__, Size);
+          break;
+      }
+    }
+    GD = reinterpret_cast<uint64_t>(MemData + (Length * Size));
+  }
+  else { // Backward
+    if (Op->IsAtomic) {
+      switch (Size) {
+        case 1:
+          MemSetElementsInverse(reinterpret_cast<std::atomic<uint8_t>*>(MemData), Value, Length);
+          break;
+        case 2:
+          MemSetElementsInverse(reinterpret_cast<std::atomic<uint16_t>*>(MemData), Value, Length);
+          break;
+        case 4:
+          MemSetElementsInverse(reinterpret_cast<std::atomic<uint32_t>*>(MemData), Value, Length);
+          break;
+        case 8:
+          MemSetElementsInverse(reinterpret_cast<std::atomic<uint64_t>*>(MemData), Value, Length);
+          break;
+        default:
+          LOGMAN_MSG_A_FMT("Unhandled {} size: {}", __func__, Size);
+          break;
+      }
+    }
+    else {
+      switch (Size) {
+        case 1:
+          MemSetElementsInverse(reinterpret_cast<uint8_t*>(MemData), Value, Length);
+          break;
+        case 2:
+          MemSetElementsInverse(reinterpret_cast<uint16_t*>(MemData), Value, Length);
+          break;
+        case 4:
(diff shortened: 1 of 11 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1340,6 +1340,165 @@ DEF_OP(StoreMemTSO) {
   }
 }
 
+DEF_OP(MemSet) {
```
lioncash, line 1343: Just so it isn't lost or anything, maybe leave just an info or general todo comment here that mentions that MOPS instructions could be used here whenever its feasible to do so

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1340,6 +1340,165 @@ DEF_OP(StoreMemTSO) {
...
+      case 8:
+        str<ARMEmitter::IndexType::POST>(Value.X(), TMP2, Size);
+        break;
+      default:
+        LOGMAN_MSG_A_FMT("Unhandled {} size: {}", __func__, Size);
+        break;
+    }
+  };
+
+  auto MemStoreTSO = [&](auto OpSize, int32_t Size) {
```
neobrain, line 1395: If possible, a `this` capture restricts what state can leak into this lambda:
```suggestion
  auto MemStoreTSO = [this](auto OpSize, int32_t Size) {
```

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1340,6 +1340,165 @@ DEF_OP(StoreMemTSO) {
...
+  }
+  else {
+    const auto Prefix = GetReg(Op->Prefix.ID());
+    add(TMP2, Prefix.X(), MemReg.X());
+  }
+
+  // Backward or forwards implementation depends on flag
+  cbnz(ARMEmitter::Size::i64Bit, Direction, &BackwardImpl);
+
+  auto MemStore = [&](auto OpSize, int32_t Size) {
```
neobrain, line 1375: Using `auto` for lambda parameters turns them into templates, which aren't needed here (but might blow up compiler errors during future work). Just using `int`/`unsigned` would be better here.

External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
```diff
@@ -288,6 +288,111 @@ DEF_OP(StoreMem) {
...
+DEF_OP(MemSet) {
+  const auto Op = IROp->C<IR::IROp_MemSet>();
+  const int32_t Size = Op->Size;
+
+  char *MemData = *GetSrc<char **>(Data->SSAData, Op->Addr);
+  const auto Value = *GetSrc<uint64_t*>(Data->SSAData, Op->Value);
+  const auto Length = *GetSrc<uint64_t*>(Data->SSAData, Op->Length);
+  const auto Direction = *GetSrc<uint8_t*>(Data->SSAData, Op->Direction);
+
+  auto MemSetElements = []<typename T>(T* Memory, uint64_t Value, size_t Length) {
```
neobrain, line 300: No need for explicit `typename T` in these lambdas. `auto* Memory` also works to ensure the argument is a pointer.

External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
```diff
@@ -288,6 +288,111 @@ DEF_OP(StoreMem) {
...
+          break;
+        default:
+          LOGMAN_MSG_A_FMT("Unhandled {} size: {}", __func__, Size);
+          break;
+      }
+    }
+    else {
+      switch (Size) {
+        case 1:
+          MemSetElements(reinterpret_cast<uint8_t*>(MemData), Value, Length);
```
neobrain, line 335:
```suggestion
          std::ranges::fill(std::span { reinterpret_cast<uint8_t*>(MemData), Length }, Value);
```

neobrain: (For MemSetElementsInverse, `std::ranges::fill(std::span { reinterpret_cast<uint8_t*>(MemData), Length } | ranges::views::reverse, Value);` should do the job)

neobrain: Not sure this is an option for the `atomic` variants, since `fill` may not guarantee any order of writes. FWIW I'm not strongly attached to this suggestion, since the `MemSetElements` inline lambda isn't too horrid to begin with.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: Added a comment at the head of the implementation.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -4855,29 +4802,34 @@ OrderedNode *OpDispatchBuilder::AppendSegmentOffset(OrderedNode *Value, uint32_t
     // With the segment register optimization we store the GDT bases directly in the segment register to remove indexed loads
     switch (Prefix) {
       case FEXCore::X86Tables::DecodeFlags::FLAG_ES_PREFIX:
-        Segment = _LoadContext(GPRSize, GPRClass, offsetof(FEXCore::Core::CPUState, es_cached));
+        return _LoadContext(GPRSize, GPRClass, offsetof(FEXCore::Core::CPUState, es_cached));
         break;
```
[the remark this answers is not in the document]
Sonicadvance1: breaks removed.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1340,6 +1340,165 @@ DEF_OP(StoreMemTSO) {
...
+  {
+    MemSetLoop(Size, Size);
+    b(&Done);
+  }
+
+  // Backward implementation
+  {
+    Bind(&BackwardImpl);
+    MemSetLoop(Size, -Size);
+  }
```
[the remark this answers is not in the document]
Sonicadvance1: Switched over even if it is a bit harder to read now.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: restricted.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: removed auto.

External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
Sonicadvance1: Nice. Removed lambda template.

External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
Sonicadvance1: I prefer the readability that each stage is the same with typing changes, rather than some doing something different for "no reason".

lioncash (approved):

PR #2479, opened 2023-03-08
OpcodeDispatcher: Handle VTESTPD/VTESTPS

lioncash:


Sonicadvance1 (approved):
Nice! Slightly confusing to read when AMD's Volume 4 documentation is incorrect but otherwise follows behaviour as expected.

PR #2494, opened 2023-03-11
Arm64: Reclaim SRA registers on 32-bit

Sonicadvance1:
Causes Portal to go from 85FPS to 120FPS on Lenovo X13s.

When running in 32-bit mode we were wasting 8 GPRs and 8 FPRs by still allocating the top 8 registers of each even though 32-bit can't use them.

Reallocate them to be register allocated registers when running 32-bit applications which reduces spills and lowers the cost of spilling/filling SRA registers.

Quite a significant speed boost for a little bit of work.

### DIFF
--- a/Source/Tests/IRLoader.cpp
+++ b/Source/Tests/IRLoader.cpp
@@ -174,6 +174,10 @@ int main(int argc, char **argv, char **const envp)
   FEXCore::Config::AddLayer(std::make_unique<FEX::ArgLoader::ArgLoader>(argc, argv));
   FEXCore::Config::AddLayer(FEXCore::Config::CreateEnvironmentLayer(envp));
   FEXCore::Config::Load();
+  // Ensure the IRLoader runs in 64-bit mode.
+  // This is to ensure that static register allocation in the JIT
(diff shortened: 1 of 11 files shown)


Source/Tests/IRLoader.cpp
```diff
@@ -174,6 +174,8 @@ int main(int argc, char **argv, char **const envp)
   FEXCore::Config::AddLayer(std::make_unique<FEX::ArgLoader::ArgLoader>(argc, argv));
   FEXCore::Config::AddLayer(FEXCore::Config::CreateEnvironmentLayer(envp));
   FEXCore::Config::Load();
+  // Ensure the IRLoader runs in 64-bit mode.
```
lioncash, line 177: Maybe expand this, just so it has the reason why we need it to be 64-bit

lioncash (commented):
One minor nit, but looks good otherwise

lioncash (approved):

Source/Tests/IRLoader.cpp
Sonicadvance1: Updated comment that it needs to run in 64-bit mode so the IRLoader has correct SRA for top 8 GPRs and XMM registers.

PR #2498, opened 2023-03-14
FEXCore: Adds fexctl container alias objects

Sonicadvance1:
Completely unused right now, this will very quickly spread throughout the entire codebase.

Currently only containers with FEX's global allocator usage. Any STL class that is allocation free and has been vetted to fit within the guidelines of not allocating memory will get an alias inside of `fextl`.

This means that any std:: namespace usage going forward will need to be scrutinized heavily. As aliasing it will mean it is vetted, otherwise it needs to be checked with FEX CI (once in place) to ensure it doesn't allocate memory behind our back.

Aliases everything we allow from std:: reduces mental burden from needing to check if it is okay to use or not. Additionally it will allow us to quickly pivot to the EASTL if we find as some point during this transition that this will in-fact not work.

Once this is committed, we are going to be moving very quickly to piecemeal convert /everything/ in the codebase over to fextl:: namespace from std::.

I am sorry for the hellscape that will follow.

### DIFF
--- a/External/FEXCore/include/FEXCore/fextl/allocator.h
+++ b/External/FEXCore/include/FEXCore/fextl/allocator.h
@@ -1,0 +1,27 @@
+#pragma once
+#include <FEXCore/Utils/Allocator.h>
+
+namespace fextl {
+  /**
+   * @brief C++ allocator class interface in to FEXCore::Allocator for memory allocations.
+   */
+  template <typename T>
+  class FEXAlloc : public std::allocator<T> {
+    public:
+      using value_type = T;
+      using propagate_on_container_move_assignment = std::true_type;
+
+      FEXAlloc() noexcept {}
+      template <class U> FEXAlloc(FEXAlloc<U> const&) noexcept {}
+
+      inline value_type *allocate(std::size_t n) {
+        return reinterpret_cast<value_type*>(::FEXCore::Allocator::aligned_alloc(std::alignment_of_v<value_type>, n * sizeof(value_type)));
+      }
+
+      inline void deallocate(value_type* p, size_t) noexcept {
+        ::FEXCore::Allocator::free(p);
+      }
+
+      inline bool operator==(const FEXAlloc&) const { return true; }
+  };
+}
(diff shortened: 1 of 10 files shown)


External/FEXCore/include/FEXCore/fextl/vector.h
```diff
@@ -0,0 +1,9 @@
+#pragma once
+#include <fextl/allocator.h>
+
+#include <vector>
+
+namespace fextl {
+  template<class T>
```
neobrain, line 7: For the sake of consistency, we might want to keep the allocator template parameter for the custom containers as well, since there's a couple of places we use PMR allocators. Using the same interface for those allows us to `grep` for `std::vector` to ensure we don't use it.

neobrain (approved):
Looks good, though we should try this out in practice before merging.

External/FEXCore/include/FEXCore/fextl/allocator.h
```diff
@@ -0,0 +1,24 @@
+#pragma once
+#include <FEXCore/Utils/Allocator.h>
+
+namespace fextl {
+  template <typename T>
+  class FEXAlloc : public std::allocator<T> {
```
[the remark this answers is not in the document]
Sonicadvance1: Added a docstring

External/FEXCore/include/FEXCore/fextl/vector.h
Sonicadvance1: Passed through the allocator as a default templated argument. std::pmr interfaces will convert to `fextl::pmr` at some point as well, but they are less of an immediate concern since the allocators those interfaces use are all very specific.

Sonicadvance1:
> Looks good, though we should try this out in practice before merging.

Gave it a try already locally by replacing a couple of `std::vector` and `std::string` and ensured that those allocations were going through the jemalloc allocators instead of my local faulting glibc allocator override path. Tested in both FEXCore and frontend code to work.
It was almost entirely sed replace `s/std::/fextl::/g` but requires tiptoeing around code that was still using the std variants.

I'll be opening PRs as quickly as possible to handle converting freestanding std:: usage as much as possible and then not stop until it is all converted.

PR #2522, opened 2023-03-14
FEXLoader: Move ELFCodeLoader2 to remove 2

Sonicadvance1:
We aren't going back to the old code at this point.


Sonicadvance1:
Make sure to look at the commits independently to see the delete of the old code and the move without significant changes.

neobrain (approved):

lioncash (approved):

PR #2544, opened 2023-03-15
fextl: Bulk merge

Sonicadvance1:
Takes all the previous PRs and merges them in to one mega PR. All the other ones can be closed once this is merged.

### DIFF
--- a/External/FEXCore/Source/Utils/FileLoading.cpp
+++ b/External/FEXCore/Source/Utils/FileLoading.cpp
@@ -1,3 +1,5 @@
+#include <FEXCore/fextl/vector.h>
+
 #include <string>
 #include <vector>
 #include <fcntl.h>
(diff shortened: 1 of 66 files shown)


lioncash (approved):
Looks good to me

External/FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -7,6 +9,7 @@
 #include <unistd.h>
 
 namespace FEXCore::FileLoading {
+// TODO: Delete once all uses of std::vector LoadFile is removed.
```
neobrain, line 12: Are there still uses of this remaining?

Generally, `[[deprecated]]` would be a good use here instead of a loose comment.

neobrain (approved):

External/FEXCore/Source/Utils/FileLoading.cpp
Sonicadvance1: Yes, there are still uses of this. It's getting deleted in a follow-up PR soon after this one.

PR #2545, opened 2023-03-15
Convert most things to fextl

Sonicadvance1:
Based off of #2544, so the first 39 commits are from that.

Primary thing remaining is std::string isn't converted which is going to spaghetti out to everything in one huge commit.


Source/Common/ArgumentLoader.cpp
```diff
@@ -37,8 +38,20 @@ namespace FEX::ArgLoader {
     using uint32 = uint32_t;
 #define AFTER_PARSE
 #include <FEXCore/Config/ConfigOptions.inl>
-    RemainingArgs = Parser.args();
-    ProgramArguments = Parser.parsed_args();
+    // TODO: Convert cpp-optparse over to fextl::vector
+    auto ParserArgs = Parser.args();
+    auto ParsedArgs = Parser.parsed_args();
+
+    RemainingArgs.reserve(ParserArgs.size());
```
neobrain, line 45:
```cpp
RemainingArgs.insert(RemainingArgs.begin(), ParserArgs.begin(), ParserArgs.end());
ProgramArguments.insert(...);
```

neobrain (approved):

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Resolved.

PR #2547, opened 2023-03-16
Convert most std::string over to fextl

Sonicadvance1:


External/FEXCore/Source/Common/Paths.cpp
```diff
@@ -9,8 +10,8 @@
 #include <unistd.h>
 
 namespace FEXCore::Paths {
-  std::unique_ptr<std::string> CachePath;
-  std::unique_ptr<std::string> EntryCache;
+  std::unique_ptr<fextl::string> CachePath;
```
lioncash, line 13: A little unrelated, but wouldn't these unique_ptr's be able to be `std::optional` instances instead? Seems like it would get rid of the need to heap allocate the strings themselves, and would just require handling the empty cases in `GetCachePath` and `GetEntryCachePath`

Sonicadvance1: Converted over in the followup PR to ensure this doesn't change behaviour here.
There's a bunch of surrounding changes that makes changing behaviour beyond names a pain.

Sonicadvance1:
Additional information, there is some c_str() and std::to_string breaking up some operations in this PR. This is temporary as the next couple PRs are working to clean it up. It's just hard to do it all in one go with these systems linking to each other.

lioncash (approved):

Source/Tests/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
```diff
@@ -673,7 +672,9 @@ namespace FEX::EmulatedFile {
     FDReadCreators["/sys/devices/system/cpu/online"] = NumCPUCores;
     FDReadCreators["/sys/devices/system/cpu/present"] = NumCPUCores;
 
-    string procAuxv = string("/proc/") + std::to_string(getpid()) + string("/auxv");
+    fextl::string procAuxv = "/proc/";
```
neobrain, line 675: With `fextl::format`, we don't need to go through `to_string` here.

neobrain (commented):
Looks promising overall, found some minor issues and added some suggestions to make this more coherent when merged on its own.

External/FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -352,7 +352,7 @@ namespace JSON {
       // Only return if it exists
       std::error_code ec{};
       if (std::filesystem::exists(Path, ec)) {
-        return Path;
+        return Path.c_str();
```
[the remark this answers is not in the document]
Sonicadvance1: This and the next comment about application names gets resolved in the next PR, so I'm hesitant to change again since merge conflicts are a pain with these.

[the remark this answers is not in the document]
neobrain: This pattern recurs a *lot* among the 126 matches for `c_str()` in this PR. Since we've opened *several* PRs very recently to remove performance overhead from unnecessary string/path copies, having a controlled way of adding new ones seems desirable.

I think an easy way to achieve this with reasonable effort would be patch up this PR with one "add fextl::string_from_path" commit and rebase the next PR on a partial revert of that same patch. That way you don't need to do resolve any merge conflicts but review becomes easier.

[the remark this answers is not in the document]
Sonicadvance1: Converted the c_str usage over to helpers that will be removed in the next PR.

External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
```diff
@@ -543,7 +543,8 @@ void Arm64Dispatcher::EmitDispatcher() {
   ClearICache(reinterpret_cast<void*>(DispatchPtr), End - reinterpret_cast<uint64_t>(DispatchPtr));
 
   if (CTX->Config.BlockJITNaming()) {
-    std::string Name = "Dispatch_" + std::to_string(FHU::Syscalls::gettid());
+    fextl::string Name = "Dispatch_";
+    Name += FHU::Syscalls::gettid();
```
neobrain, line 547: Was it intended to drop the `std::to_string` here? Won't this truncate the thread id to a `char` that is appended to the string?

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
```diff
@@ -407,7 +408,8 @@ X86Dispatcher::X86Dispatcher(FEXCore::Context::ContextImpl *ctx, const Dispatche
   End = Start + getSize();
 
   if (CTX->Config.BlockJITNaming()) {
-    std::string Name = "Dispatch_" + std::to_string(FHU::Syscalls::gettid());
+    fextl::string Name = "Dispatch_";
+    Name += FHU::Syscalls::gettid();
```
neobrain, line 412: Same comment about dropping `to_string`.

External/FEXCore/Source/Interface/Core/GdbServer.cpp
```diff
@@ -984,7 +985,7 @@ GdbServer::HandledPacketType GdbServer::ThreadAction(char action, uint32_t tid)
     case 's': {
       CTX->Step();
       SendPacketPair({"OK", HandledPacketType::TYPE_ACK});
-      auto str = fmt::format("T05thread:{:02x};", getpid());
+      fextl::string str = fmt::format("T05thread:{:02x};", getpid()).c_str();
```
neobrain, line 988: Not sure I pointed this out before, but should be `fextl::fmt::format` (feel free to skip this one if that's done in the followup PR though).

neobrain (approved):
Found one more potential bug that needs looking into, otherwise looks good!

External/FEXCore/Source/Interface/Core/Dispatcher/Arm64Dispatcher.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
Sonicadvance1: Resolved.

PR #2549, opened 2023-03-18
Move FEX away from the remaining glibc allocations that we can

Sonicadvance1:
~~This is based on #2547, first 28 commits are from that and needs to be merged first.~~

This moves FEX almost entirely over to jemalloc without glibc hooks.

There are a handful of locations that still use glibc due to using some interfaces that use it without allocator overrides.
For these cases we have a new scoped type that will termporarily disable glibc faulting if enabled, or do nothing in the case that faulting isn't enabled.

```cpp
FEXCore::Allocator::YesIKnowImNotSupposedToUseTheGlibcAllocator glibc;
```
Named that way to ensure it is easily greppable and also shaming the code locations that use it.

Next PRs will be adding CI for this, adding documentation about known APIs that allocate memory and hopefully removing a handful of locations that still use the fault check override.

```bash
$ ag "FEXCore::Allocator::YesIKnowImNotSupposedToUseTheGlibcAllocator glibc" | wc -l
25
```


External/FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -719,10 +722,10 @@ namespace JSON {
 #define ENVLOADER
 #include <FEXCore/Config/ConfigOptions.inl>
 
-      EnvMap[Key]=Value;
+      EnvMap[Key] = Value;
     }
 
-    std::function GetVar = [=](const std::string_view id)  -> std::optional<std::string_view> {
+    std::function GetVar = [](EnvMapType &EnvMap, const std::string_view id)  -> std::optional<std::string_view> {
```
neobrain, line 728: This doesn't need to be a `std::function` anymore:
```suggestion
    auto GetVar = [](EnvMapType &EnvMap, const std::string_view id)  -> std::optional<std::string_view> {
```

External/FEXCore/Source/Interface/Context/Context.h
```diff
@@ -71,6 +72,14 @@ namespace FEXCore::Context {
 
   class ContextImpl final : public FEXCore::Context::Context {
     public:
+      void *operator new(size_t size) {
```
neobrain, line 75: Why do we need to add these here?

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
```diff
@@ -418,13 +418,11 @@ X86Dispatcher::X86Dispatcher(FEXCore::Context::ContextImpl *ctx, const Dispatche
 
 }
 
-// Used by GenerateGDBPauseCheck, GenerateInterpreterTrampoline
-static thread_local Xbyak::CodeGenerator emit(1, &emit); // actual emit target set with setNewBuffer
```
neobrain, line 420: Can we not have thread_locals anymore, or is this merely not needed here?

External/FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -243,12 +243,12 @@ class LookupCache {
   // walking each block member and destructing objects.
   //
   // This makes `BlockLinks` look like a raw pointer that could memory leak, but since it is backed by the MBR, it won't.
-  std::pmr::monotonic_buffer_resource BlockLinks_mbr;
+  fextl::unique_ptr<std::pmr::monotonic_buffer_resource> BlockLinks_mbr;
```
neobrain, line 246: `unique_ptr` isn't really needed here - we can move initialization of `BlockLinks_mbr` into the constructor's initialization list instead (or initialize it right here in the header, but seems like a leaky implementation detail).

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,139 @@
...
+#include <fmt/format.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <fstream>
+#include <sstream>
+#include <unistd.h>
+
+extern "C" {
+#ifdef GLIBC_ALLOCATOR_FAULT
```
neobrain, line 14: Where is the actual fault triggered?

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,139 @@
...
+  // glibc allocator faulting. If anything uses a glibc allocator symbol then cause a fault.
+  // Ensures that FEX only ever allocates memory through its allocator routines.
+  // If thunks are enabled then this should crash immediately. Thunks should still go through glibc allocator.
```
neobrain, line 17: The phrasing is a bit convoluted here. Generally, it's unclear what you mean if something "should" do something (is this an expectation to the user or to the implementation? If it's the latter, "should" indicates a degree of uncertainty... why is there uncertainty and what should the user do about it?)

For this comment specifically:
* What does thunks being enabled have to do with crashing here?
* I'm not sure what information to extract from "Thunks should still go through glibc allocator". My best guess actually contradicts the preceding sentence.

Something like this might be better:
* "The majority of FEX-internal code should avoid using the glibc allocator. To ensure glibc allocations don't accidentally slip in, we override these glibc functions with faulting variants."
* "A notable exception is thunks[, which should still ...]. Other minor exceptions throughout FEX use the YesIKnowImNotSupposedToUseTheGlibcAllocator helper to temporarily disable faulting."

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,139 @@
...
+  int posix_memalign(void **, size_t, size_t) GLIBC_ALIAS_FUNCTION(fault_posix_memalign);
+
+  extern size_t __malloc_usable_size(void*);
+  size_t malloc_usable_size(void*) GLIBC_ALIAS_FUNCTION(fault_malloc_usable_size);
+
+  // Reuse __libc_memalign
+  void *aligned_alloc(size_t, size_t) GLIBC_ALIAS_FUNCTION(fault_aligned_alloc);
+}
+
+namespace FEXCore::Allocator {
```
neobrain, line 48: The rest of this file needs *some* documentation.

External/FEXCore/Source/Interface/Core/Dispatcher/X86Dispatcher.cpp
Sonicadvance1: Merely not needed.

External/FEXCore/include/FEXCore/fextl/memory_resource.h
```diff
@@ -0,0 +1,77 @@
...
+     * The difference is that class doesn't have ownership of the backing memory and
+     * it also doesn't have any growth factor.
+     *
+     * If the amount of memory allocated is overrun then this will overwrite memory unless assertions are enabled.
+     *
+     * Ensure that you know how much memory you're going to use before using this class.
+     */
+    class fixed_size_monotonic_buffer_resource final : public std::pmr::memory_resource {
+    public:
+        fixed_size_monotonic_buffer_resource(void* Base, size_t Size)
```
neobrain, line 40: `Size` should have a `[[maybe_unused]]` attribute.

Source/Tests/LinuxSyscalls/FileManagement.cpp
```diff
@@ -662,7 +655,14 @@ uint64_t FileManager::Readlinkat(int dirfd, const char *pathname, char *buf, siz
...
     // We need to get the path from the fd now
-    Path = FEX::get_fdpath(dirfd).value_or("");
+    char Tmp[PATH_MAX];
+    auto PathLength = FEX::get_fdpath(dirfd, Tmp);
+    if (PathLength != -1) {
+      Path = fextl::string(Tmp, PathLength);
+    }
+    else {
+      Path = "";
+    }
```
neobrain, line 665:
```suggestion
    char Tmp[PATH_MAX] = "";
    auto PathLength = FEX::get_fdpath(dirfd, Tmp);
    if (PathLength != -1) {
      Path = fextl::string(Tmp, PathLength);
    }
```

Source/Tests/LinuxSyscalls/Syscalls.cpp
```diff
@@ -268,8 +268,8 @@ uint64_t ExecveHandler(const char *pathname, char* const* argv, char* const* env
       Filename = pathname;
     }
 
-    bool exists = std::filesystem::exists(Filename, ec);
-    if (ec || !exists) {
+    bool exists = FHU::Filesystem::Exists(Filename.c_str());
```
neobrain, line 271: Could we add a dedicated overload for `Exists(fextl::string)`? Seems to come up a couple of times throughout the PR.

Source/Tests/TestHarnessRunner.cpp
```diff
@@ -115,31 +115,33 @@ class TestEnvLoader final : public FEXCore::Config::Layer {
...
-  FEXCore::Config::AddLayer(std::make_unique<FEX::ArgLoader::ArgLoader>(argc, argv));
+  FEXCore::Config::AddLayer(fextl::make_unique<FEX::ArgLoader::ArgLoader>(argc, argv));
   FEXCore::Config::AddLayer(FEXCore::Config::CreateEnvironmentLayer(envp));
   FEXCore::Config::Load();
 
   auto Args = FEX::ArgLoader::Get();
 
   if (Args.size() < 2) {
     LogMan::Msg::EFmt("Not enough arguments");
+    FEXCore::Allocator::ClearFaultEvaluate();
```
neobrain, line 130: Why don't we use a scope guard for this? Are there exit paths where we *don't* want to call this function?

unittests/APITests/LexicallyNormal.cpp
```diff
@@ -0,0 +1,30 @@
+#include <catch2/catch.hpp>
+#include <filesystem>
+#include <FEXHeaderUtils/Filesystem.h>
+
+#define TestPath(Path) \
+  REQUIRE(std::string_view(FHU::Filesystem::LexicallyNormal(Path)) == std::string_view(std::filesystem::path(Path).lexically_normal().string()));
```
neobrain, line 6: Catch2 has generators for this:

```cpp
TEST_CASE("LexicallyNormal") {
    auto path = GENERATE("", "/", ...);
    REQUIRE(std::string_view(FHU::Filesystem::LexicallyNormal(path)) == std::string_view(std::filesystem::path(path).lexically_normal().string()));
}
```

External/FEXCore/Source/Interface/Config/Config.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Context/Context.h
Sonicadvance1: Removed until I can validate again that it is necessary.

External/FEXCore/include/FEXCore/fextl/memory_resource.h
Sonicadvance1: Resolved.

Source/Tests/LinuxSyscalls/FileManagement.cpp

Source/Tests/TestHarnessRunner.cpp
Sonicadvance1: Changed over to a new scope guard. Resolved.

unittests/APITests/LexicallyNormal.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
Sonicadvance1: Added a comment that it is down below.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
Sonicadvance1: Added more comments.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+  inline bool Exists(const fextl::string &Path) {
+    return access(Path.c_str(), F_OK) == 0;
+  }
+
+  /**
+   * @brief Creates a directory at the provided path.
+   *
+   * @param Path The path to create a directory at.
+   *
+   * @return True if the directory was created or already exists.
```
neobrain, line 35: This is not consistent with `std::create_directory`, which returns false if the directory already existed. (This is different from `std::create_directories`, so CreateDirectories doesn't need changes in behavior but needs to revisit how it uses CreateDirectory internally)

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+    if (LastSeparator == fextl::string::npos) {
+      // No separator. Likely relative `.`, `..`, `<Application Name>`, or empty string.
+      return Path;
+    }
+
+    return Path.substr(0, LastSeparator);
+  }
+
+  inline bool IsRelative(const fextl::string &Path) {
+    return Path.starts_with('/');
```
neobrain, line 105: Shouldn't this be `!Path.starts_with('/')`? Currently, even the path "." would be reported as "not relative".

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+  }
+
+  enum class CopyOptions {
+    NONE,
+    SKIP_EXISTING,
+    OVERWRITE_EXISTING,
+  };
+
+  inline bool CopyFile(const fextl::string &From, const fextl::string &To, CopyOptions Options = CopyOptions::NONE) {
```
neobrain, line 114: docstring needed (just referring to `std::filesystem::copy_file` is fine though)

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -98,6 +98,14 @@ namespace {
     fextl::vector<SpillStackUnit> SpillStack;
     fextl::unordered_map<IR::NodeID, fextl::unordered_set<IR::NodeID>> BlockPredecessors;
     fextl::unordered_map<IR::NodeID, fextl::unordered_set<IR::NodeID>> VisitedNodePredecessors;
+
+    void *operator new(size_t size) {
```
neobrain, line 102: Leftover from previous version?

External/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
```diff
@@ -30,6 +31,14 @@ namespace Alloc::OSAllocator {
 
   class OSAllocator_64Bit final : public Alloc::HostAllocator {
     public:
+      void *operator new(size_t size) {
```
neobrain, line 34: More leftovers

External/FEXCore/include/FEXCore/Debug/InternalThreadState.h
```diff
@@ -62,6 +63,14 @@ namespace FEXCore::Core {
     fextl::vector<DebugDataSubblock> Subblocks;
     fextl::vector<DebugDataGuestOpcode> GuestOpcodes;
     fextl::vector<FEXCore::CPU::Relocation> *Relocations;
+
+    void *operator new(size_t size) {
```
neobrain, line 67: More leftovers (?)

External/FEXCore/include/FEXCore/Debug/InternalThreadState.h
```diff
@@ -118,6 +127,14 @@ namespace FEXCore::Core {
     std::shared_mutex ObjectCacheRefCounter{};
     bool DestroyedByParent{false};  // Should the parent destroy this thread, or it destory itself
 
+    void *operator new(size_t size) {
```
neobrain, line 130: Leftovers

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include <fmt/format.h>
+
+#include <cstddef>
+#include <cstdint>
+#include <fstream>
+#include <sstream>
+#include <unistd.h>
+
+extern "C" {
+#ifdef GLIBC_ALLOCATOR_FAULT
```
neobrain, line 14: Preprocessor guard not needed since this file is only compiled if this is true. (Could turn the `!defined` case into a hard error maybe)

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+    }
+
+    return Path.substr(LastSeparator + 1);
+  }
+
+  inline fextl::string ParentPath(const fextl::string &Path) {
+    auto LastSeparator = Path.rfind('/');
+    if (LastSeparator == fextl::string::npos) {
+      // No separator. Likely relative `.`, `..`, `<Application Name>`, or empty string.
+      return Path;
```
neobrain, line 98: E.g. `ParentPath(".")` should return `""`.

Maybe this function could use some unit tests similar to LexicallyNormal?

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+    if (PathSize == 0) {
+      return {};
+    }
+
+    // Count the number of separators up front
+    const auto SeparatorCount = std::count(Path.begin(), Path.end(), '/');
+
+    // Needs to be a list so iterators aren't invalidated while erasing.
+    // Maximum number of list regions will be the counted separators plus 2.
+    // Multiplied by two since it allocates * 2 the size (32-bytes per element)
```
neobrain, line 185: What does this line mean?

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+      const bool IsEllipses = Part == "." || Part == "..";
+
+      const bool NeedsSeparator =
+        !IsFinal || // Needs a separator if this isn't the final part.
+        (IsFinal && !IsEllipses); // Needs a separator if the final part doesn't end in `.` or `..`
+
+      if (NeedsSeparator) {
+        OutputPath += '/';
+      }
+    }
```
neobrain, line 287: Looks like this could be simplified a lot by using `fextl::format("{}", fmt::join(Parts, '/'))` combined with a special-case for `IsEllipses`.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+    // Needs to be a list so iterators aren't invalidated while erasing.
+    // Maximum number of list regions will be the counted separators plus 2.
+    // Multiplied by two since it allocates * 2 the size (32-bytes per element)
+    // Use a small stack allocator to be more optimal.
+    // Needs to be a list so iterators aren't invalidated while erasing.
+    size_t DataSize = sizeof(std::string_view) * (SeparatorCount + 2) * 2;
+    void *Data = alloca(DataSize);
+    fextl::pmr::fixed_size_monotonic_buffer_resource mbr(Data, DataSize);
+    std::pmr::polymorphic_allocator pa {&mbr};
+    std::pmr::list<std::string_view> Parts{pa};
```
neobrain, line 192: Is per-element size of the `next` and `previous` pointers in the list considered in the DataSize above?

neobrain, line 192: Instead of doing this dance to allocate a list on the stack, couldn't a `fextl::vector` be used with indexes instead?

neobrain (commented):
Shaping up nicely! Adding a final set of comments.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,159 @@
...
+  // Reuse __libc_memalign
+  void *aligned_alloc(size_t, size_t) GLIBC_ALIAS_FUNCTION(fault_aligned_alloc);
+}
+
+namespace FEXCore::Allocator {
+  // Enable or disable allocation faulting globally.
+  static bool GlobalEvaluate{};
+
+  // Enable or disable allocation faulting per-thread.
+  thread_local uint64_t SkipEvalForThread{};
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+  // Evaluate if a glibc hooked allocation should fault.
+  void EvaluateReturnAddress(void* Return) {
+    if (!GlobalEvaluate) {
+      // Fault evaluation disabled globally.
+      return;
+    }
+
+    if (SkipEvalForThread) {
+      // Fault evaluation disabled for this thread.
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+  enum class CopyOptions {
+    NONE,
+    SKIP_EXISTING,
+    OVERWRITE_EXISTING,
+  };
+
+  inline bool CopyFile(const fextl::string &From, const fextl::string &To, CopyOptions Options = CopyOptions::NONE) {
+    bool DestTested = false;
+    bool DestExists = false;
+    if (Options == CopyOptions::SKIP_EXISTING && Exists(To.c_str())) {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+      DestTested = true;
+      DestExists = Exists(To.c_str());
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+      if (DestExists) {
+        // If the destination file exists already and the skip existing flag is set then
+        // return true without error.
+        return true;
+      }
+    }
+
+    if (!DestTested) {
+      DestTested = true;
+      DestExists = Exists(To.c_str());
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

unittests/APITests/LexicallyNormal.cpp
```diff
@@ -0,0 +1,31 @@
+#include <catch2/catch.hpp>
+#include <filesystem>
+#include <FEXHeaderUtils/Filesystem.h>
+
+#define TestPath(Path) \
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: That it should be. Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Resolved by returning an enum result about what happened rather than boolean.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +22,270 @@ namespace FHU::Filesystem {
...
+    auto LastSeparator = Path.rfind('/');
+    if (LastSeparator == fextl::string::npos) {
+      // No separator. Likely relative `.`, `..`, `<Application Name>`, or empty string.
+      return Path;
+    }
+
+    return Path.substr(LastSeparator + 1);
+  }
+
+  inline fextl::string ParentPath(const fextl::string &Path) {
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Resolved. Added unit tests.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: We're walking iterators and deleting iterators as we're walking the list. Which doesn't mesh what vector gives us.

Sonicadvance1: Yes, that's what the comment above that you were confused about. Rewrote that part of the comment to explain more.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Rewrote that bit since it is confusing. Is just making sure to allocate enough stack space to handle the list element size.

unittests/APITests/LexicallyNormal.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+TEST_CASE("LexicallyNormal") {
+  auto Path = GENERATE("",
```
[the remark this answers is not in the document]
Sonicadvance1: Added edge cases to the unit tests.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: That wouldn't really be feasible with how the logic works.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Also this is specifically done on the stack for performance reasons. This function is absolutely hammered by pressure-vessel and using this custom pmr allocator. Microbench showing that a regular list without pmr is 2.6x slower

Sonicadvance1:
Fixed the random crash I was having with this, I accidentally changed some logic in the thread creation logic. Left a comment in the source that I can get back to since we're currently leaking some stack objects.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
neobrain: What speaks against this?
```cpp
auto OutputPath = fextl::format("{}", fmt::join(Parts, '/'))
if (!Parts.empty && Parts.back() != "." && Parts.back() != "..") {
  // Add a separator unless the final element is ellipses.
  OutputPath += '/';
}
```

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
neobrain: > We're walking iterators and deleting iterators as we're walking the list. Which doesn't mesh what vector gives us.

Does "doesn't mesh well" mean it literally won't work or you anticipate performance to be worse? We're not *storing* more than one iterator at a time though, so I think it should be fine as long as we use indexes rather than iterators.

> Also this is specifically done on the stack for performance reasons. This function is absolutely hammered by pressure-vessel and using this custom pmr allocator is worth it. Microbench showing that a regular list without pmr is 2.6x slower

Sounds fine. If you have spare time I'm curious how vector would perform with the same stack-allocation trick. There'd be a bunch of small memory reallocations, but we'd save on pointer traversals for the majority of iteration steps that don't insert/erase. (and we also wouldn't have std::list's internal representation leak into our implementation through the size estimate.)

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
neobrain: Now all the information needed to understand is there, but it's very disconnected. It's also unclear which of the comments refers to which part of the code.

Suggested rephrase:
```
Use std::list to store path elements to avoid iterator invalidation on insert/erase.
The list is allocated on stack to be more optimal. The size is determined by the
maximum number of list objects (separator count plus 2) multiplied by the list
element size (32-bytes per element: the string_view itself and the prev/next pointers).
```

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
```diff
@@ -170,6 +170,14 @@ class IRListView final {
     }
   }
 
+  void *operator new(size_t size) {
```
neobrain, line 173: Did you end up finding a need to keep these?

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: I don't see how it would work with the idea of iterator deletion without shuffling the entire vector each time an erase occurs.

If the issue of leaking std::list implementation details is the concern then wiring up a doubly linked list that is guaranteed to be 2x pointers + object size by providing a static element size query would resolve this?

neobrain: > I don't see how it would work with the idea of iterator deletion without shuffling the entire vector each time an erase occurs.

You would have to relocate all memory past the erased element, yes. (std::vector::erase does this for you)

This is fairly lightweight since we're only relocating `string_views`, and more importantly it's relatively rare (unless there are more `.`/`..` nodes in the path than actual directories). So I'd still expect vector to perform better on average than list here, since the latter will do pointer lookups crossing the L1 cacheline size every second iteration.

Sonicadvance1: I really don't want to implement a new container paradigm for a task that is already a perfect fit for a doubly linked list with a fixed maximum number of nodes. If you're itching to implement new container paradigms for microbenching then go for it.
The farthest I would go is implementing a `fixed_list` type to avoid leaking `std::list` implementation details.

External/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
Sonicadvance1: Required due to raw new usage. Added a comment.

External/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
Sonicadvance1: Removed, looks like leftover.

External/FEXCore/include/FEXCore/Debug/InternalThreadState.h
Sonicadvance1: Required due to raw new usage. Added a comment.

External/FEXCore/include/FEXCore/Debug/InternalThreadState.h
Sonicadvance1: Required due to raw new usage. Added a comment.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Switched over to it. For me it's harder to reason about but oh well.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
Sonicadvance1: Updated comment to rephrased.

unittests/APITests/Filesystem.cpp
```diff
@@ -0,0 +1,79 @@
...
+    "./../",
+    "./../",
+    "./.././.././.",
+    "./.././.././..",
+
+    "./foo1/../",
+    // std::fs here keeps the `/` after `foo2/`
+    // FEX algorithm doesn't keep behaviour here.
+    // "foo2/./bar/..",   // std::fs -> "foo2/"
+    // "foo2/.///bar/..", // std::fs -> "foo3/"
```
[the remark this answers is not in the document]
Sonicadvance1: Added test for failing parameters.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
neobrain: Ah yes, I see now. Since these overloaded operators all function the same, how do you feel about defining a helper struct with them that uses inherit from?

For example:
```cpp
struct FexAllocOperators {
    void *operator new(size_t size) {
        return FEXCore::Allocator::malloc(size);
    }

  void operator delete(void *ptr) {
    return FEXCore::Allocator::free(ptr);
  }
};

struct IRListView : FexAllocOperators {
    // ...
};
```

neobrain (approved):
Thanks for the review followups! Now let's get this merged 🥳

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -14,4 +23,290 @@ namespace FHU::Filesystem {
...
+    // The list is allocated on stack to be more optimal. The size is determined by the
+    // maximum number of list objects (separator count plus 2) multiplied by the list
+    // element size (32-bytes per element: the string_view itself and the prev/next pointers).
+    size_t DataSize = (sizeof(std::string_view) + sizeof(void*) * 2) * (SeparatorCount + 2);
+    void *Data = alloca(DataSize);
+    fextl::pmr::fixed_size_monotonic_buffer_resource mbr(Data, DataSize);
+    std::pmr::polymorphic_allocator pa {&mbr};
+    std::pmr::list<std::string_view> Parts{pa};
+    // Calculate the expected string size while parsing to reduce allocations.
+    size_t ExpectedStringSize{};
```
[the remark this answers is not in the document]
Sonicadvance1: Removed bookkeeping.

External/FEXCore/include/FEXCore/IR/IntrusiveIRList.h
Sonicadvance1: Switched over to inheritance.

PR #2552, opened 2023-03-20
Docs: Adds programming concerns documentation

Sonicadvance1:
Explaining FEX's memory concerns.

### DIFF
--- a/docs/ProgrammingConcerns.md
+++ b/docs/ProgrammingConcerns.md
@@ -1,0 +1,185 @@
+# Memory allocation routines
+## What is the problem?
+FEX-Emu needs to allocate memory differently than regular applications. This problem happens because FEX runs both 32-bit and 64-bit guest
+applications in the same address space as FEX itself. When running 32-bit applications, FEX reserves up all memory above 4GB in order to correctly
+emulate the 32-bit address space. We then use that reserved space for FEX allocations, so we don't interrupt application's own allocations.
+
+### Why not just replace the system allocator?
+We could control the placement of FEX's internal allocations by overriding the system
+allocator. However, 32-bit thunks (and their corresponding native host libraries) still
+need to allocate memory in the lower 4 GB of memory so that they produce
+guest-accessible pointers. Since overriding the system allocator is a global operation,
+selectively overriding it like this is not possible.
+
+Since we found no way to resolve this conflict, we had to resort to the alternative of avoiding use of the system allocator for FEX's internal
+allocations entirely (where possible).
+
+## Sub-projects and applications that need to follow this.
+- FEXCore
+- FEXInterpreter/FEXLoader
+
+## Sub-projects that explicitly cannot follow this
+- Thunks
+
+## APIs which allocate memory that FEX needs to avoid
+Most C++ APIs allow you to replace their allocators, but some don't and we need to avoid those APIs. If FEX uses them then the 32-bit application
+running might run out of memory. This isn't an all encompassing list and we will add to it as our CI captures more problems.
+
+### `get_nprocs_conf`
+Use `FEXCore::CPUInfo::CalculateNumberOfCPUs` instead.
+
+### `getcwd`
+Don't use getcwd with a nullptr buffer. It will allocate memory behind our back and return a pointer that needs a free.
+
+### `strerror`
+This allocates and frees memory based on locale! Even with C local it'll attempt to free(0).
+FEX-Emu should avoid using this function and instead just return the number.
+If necessary FEX will provide its own routine for getting this string back.
+
+### `std::make_unique`
+Use `fextl::make_unique` instead.
+
+### `std::unique_ptr`
+Use `fextl::unique_ptr` instead.
+
+### `std::filesystem`
+This namespace is /highly/ likely to allocate memory behind our back.
+FEX should avoid using this API as much as possible.
+
+#### std::filesystem::path
+#### std::filesystem::path::is_relative
+Use `FHU::Filesystem::IsRelative` instead.
+
+#### std::filesystem::absolute
+Always allocates memory.
+Use `realpath` instead.
+
+#### std::filesystem::exists
+Creates a std::filesystem::path when passing in to it.
+Use `FHU::Filesystem::Exists` instead.
+
+#### std::filesystem::canonical
+Always allocates memory.
+Use `realpath` instead.
+
+#### std::filesystem::path::lexically_normal
+Use `FHU::Filesystem::LexicallyNormal` instead.
+
+#### std::filesystem::create_directory
+#### std::filesystem::create_directories
+Creates a std::filesystem::path when passing in to it.
+Use `FHU::Filesystem::CreateDirectory` and `FHU::Filesystem::CreateDirectories` instead.
+
+#### std::filesystem::path::parent_path
+Use `FHU::Filesystem::ParentPath` instead.
+
+#### std::filesystem::path::filename
+Use `FHU::Filesystem::GetFilename` instead.
+
+#### std::filesystem::copy_file
+Use `FHU::Filesystem::CopyFile` instead.
+
+### `std::fstream`
+This API always allocates memory and should be avoided.
+Use a combination of open and fextl::string APIs instead of fstream.
+
+### `std::fwrite`
+This API allocates a buffer in the background for buffering output.
+Use raw `write` instead.
+
+### `std::string`
+Use `fextl::string` instead.
+
+#### std::to_string
+Use `fextl::fmt::format` instead.
+
+### `std::stol`
+### `std::stoul`
+### `std::stoll`
+### `std::stoull`
+These all consume a `std::string` as their first argument. Use the equivalent functions that don't use `std::string`
+- `std::strtol`
+- `std::strtoul`
+- `std::strtoll`
+- `std::strtoull`
+
+### `fmt::`
+### `fmt::format`
+Use `fextl::fmt::` instead
+
+### APIs that FEX doesn't have a replacement for
+Don't use any of these APIs in FEXLoader/FEXInterpreter. Shoutout to
+[this](https://stackoverflow.com/questions/43056338/standard-library-facilities-which-allocate-but-dont-use-an-allocator) StackOverflow post for this
+huge list.
+
+#### `std::any`
+#### `std::function` and lambdas
+One must take additional considerations when using these to ensure that they don't allocate memory. These don't have any way to replace which
+allocator is used for these objects. Additionally there is no way up-front to know if these will allocate memory or if the compiler will use
+small-function optimizations to avoid allocations. The only real way to check this is to enable the glibc faulting compile option.
+- Lambdas without anything in the capture list will never allocate memory.
+- One pointer in the capture list is likely to hit small-function optimizations and not allocate memory.
+   - This isn't guaranteed.
+- Passing the `std::function` as an argument is unlikely to optimize away their memory allocations.
+
+#### `std::valarray`
+#### `std::filebuf`
+#### `std::inplace_merge`
+#### `<stdexcept>`
+#### `std::boyer_moore_searcher`
+#### `std::filesystem::path`
+#### `std::filesystem::directory_iterator`
+#### `std::regex`
+#### `std::thread`
+#### `std::async`
+#### `std::packaged_task`
+#### `std::promise`
+#### `<iostream>`
+#### Remember this is not an all-encompassing list! We may find APIs that still allocate memory and need to be avoided!
+
+### Regular memory allocation routines.
+Don't use these directly as they will the glibc allocator.
+
+#### mmap
+Use `FEXCore::Allocator::mmap`
+
+#### munmap
+Use `FEXCore::Allocator::munmap`
+
+#### malloc
+Use `FEXCore::Allocator::malloc`
+
+#### calloc
+Use `FEXCore::Allocator::calloc`
+
+#### memalign 
+Use `FEXCore::Allocator::memalign`
+
+#### valloc
+Use `FEXCore::Allocator::valloc`
+
+#### posix_memalign
+Use `FEXCore::Allocator::posix_memalign`
+
+#### realloc
+Use `FEXCore::Allocator::realloc`
+
+#### free
+Use `FEXCore::Allocator::free`
+
+#### aligned_alloc
+Use `FEXCore::Allocator::aligned_alloc`
+
+#### __libc_malloc
+#### __libc_calloc
+#### __libc_memalign
+#### __libc_valloc
+#### __posix_memalign
+#### __malloc_usable_size
+!! DO NOT USE !!
+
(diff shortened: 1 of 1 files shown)


docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+Use `fextl::fmt::format` instead.
+
+### APIs that FEX doesn't have a replacement for
+Don't use any of these APIs in FEXLoader/FEXInterpreter. Shoutout to
+[this](https://stackoverflow.com/questions/43056338/standard-library-facilities-which-allocate-but-dont-use-an-allocator) StackOverflow post for this
+huge list.
+
+#### `std::any`
+#### `std::function`
+Ensure the std::function doesn't allocate memory.
```
neobrain, line 90: What actionable advice is this intended to give? Should clarify if we implicitly rely on a small-function optimization here, and (if so) what size this commonly limits us to with the standard libraries we compile against.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
+# Please read this before contributing to FEX-Emu
```
neobrain, line 1: This is the wrong place for that kind of note. The section header and the paragraph don't really add anything, and instead a CONTRIBUTING.md could be created to point developers to important documents like this one.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+FEX-Emu is mostly a regular C++20 application but we have some interesting problems that most applications don't need to contend with.
+There are concerns with dealing with signals and memory allocation that we must be careful with handling.
+
+# Memory allocation routines
+## What is the problem?
+FEX-Emu needs to allocate memory differently than regular applications. This problem happens because FEX runs both 32-bit and 64-bit guest
```
neobrain, line 7: This is backwards: You're not starting with the problem itself but with FEX's strategy of addressing it.

A more approachable intro would be "Games expect the full 32-bit address space to be available to them. Since FEX shares a single address space with its emulated game, FEX must ensure to place its internal allocations outside of that address space whenever possible. To this end...".

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+applications in the same address space as FEX itself. When running 32-bit applications, FEX reserves up all memory above 4GB in order to correctly
+emulate the 32-bit address space. We then use that reserved space for FEX allocations, so we don't interrupt application's own allocations.
+
+### Why not just just replace the system allocator?
+JEMalloc and similar allocators allow your application to replace the system allocator. This is something that FEX used to do, but this has since been
```
neobrain, line 12: Nit: The project name is spelt `jemalloc`.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+removed because 32-bit thunks need to allocate memory in the lower 4GB.
```
neobrain, line 13: This raises the question why thunks need to do this, and why overriding the system allocator would interfere with it. The previous section should probably have an extra paragraph explaining the thunking issue first ("One notable exception are thunks: Allocations made in native host libraries must always be valid 32-bit pointers, ..."), since bringing thunks up here is a bit random.

With that extra paragraph, we could just say "With this approach [of replacing the system allocator], we couldn't find an easy way to customize allocation behavior for the requirements of 32-bit thunking" here.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+Use `FHU::Filesystem::ParentPath` instead.
+
+#### std::filesystem::path::filename
+Use `FHU::Filesystem::GetFilename` instead.
+
+#### std::filesystem::copy_file
+Use `FHU::Filesystem::CopyFile` instead.
+
+### `std::fstream`
+This API always allocates memory and should be avoided.
```
neobrain, line 74: But we still use `fstream` in plenty of places. Do we need to replace them in the future, or should this guideline be changed to something like "make sure the `const char*` constructor is used"?

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+**Don't use**
+
+#### __malloc_usable_size
+**Don't use**
+
+## FEX thunk considerations
+Now that we understand the problems of FEX memory allocations. How do these problems affect FEX's thunks?
+
+### Don't use FEXCore APIs that allocate memory through FEX's allocator
+It's as simple as that. Thunks need to use the glibc allocator so their memory allocations end up using the host glibc allocator.
```
neobrain, line 161: FEXCore APIs are not accessible from within thunks in the first place (unless explicitly wired up), so this is a bit misleading. Describing why it's undesirable for thunks to access those APIs and which specific allocations unconditionally *mustn't* be allocated through them (i.e. pointers that are guest-facing because they are returned from functions or passed to a callback) seems more instructive here.
The following paragraph would become redundant then too, since it's clear we have more leeway with allocations that aren't guest-facing.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+
+If there becomes a point in which FEX's thunks need to allocate memory outside of the 32-bit virtual address space then FEX can expose APIs for this.
+Currently this isn't supported but it can be as necessary.
+
+## How does FEX ensure that this paradigm doesn't break?
+FEX has the cmake option `ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT` to hook in to glibc's allocator and fault if anything is allocating through it.
+This can't be used with thunks for thunk testing as those actually use the glibc allocator.
+With this option enabled, FEX's CI will run as many tests as possible (None with thunks) to ensure that FEX isn't allocating memory through glibc.
```
neobrain, line 169: `as many tests as possible (None with thunks)` seems confusing here. Is "no thunks" the only restriction? If yes, this is already implied by the previous sentence. If no, the actual restrictions should be elaborated. If no and the restrictions aren't all that important, I think it's fine to say "CI will run FEX's test suite with extra verification to ensure that FEX isn't allocating memory through glibc".

neobrain: Thinking further about it, a better approach is probably to rewire this and the previous sentence into something like:
`CI will run FEX's test suite with extra verification to ensure FEX makes no allocations are made through glibc. Thunking must be disabled for this run, since thunks are by design the only place where glibc allocation still happen.`

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,176 @@
+# Memory allocation routines
```
neobrain, line 1: Could you add a section to the document explaining which subprojects the guidelines apply to? E.g. external applications like FEXRootFSFetcher are obviously not bound to this, but for unit tests it's not so clear.

Some of the mentions of pecularities around thunking could be moved to such a section as well.

.github/workflows/glibc_fault.yml
```diff
@@ -0,0 +1,182 @@
...
+      shell: bash
+      working-directory: ${{runner.workspace}}/build
+      run: mv ${{runner.workspace}}/build/Testing/Temporary/LastTest.log ${{runner.workspace}}/build/Testing/Temporary/LastTest_GCC32.log || true
+
+    - name: APITest tests
+      working-directory: ${{runner.workspace}}/build
+      shell: bash
+      run: cmake --build . --config $BUILD_TYPE --target api_tests
+
+    - name: APITest Test Results move
```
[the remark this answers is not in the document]
Sonicadvance1: glibc faulting is opt-in rather than opt-out so these tests will run fine.

Sonicadvance1:
Dropped the CI workflow file from this PR so documentation and CI can be split up sanely. Letting this merge earlier than the CI workflow.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,170 @@
...
+
+### `std::fstream`
+This API always allocates memory and should be avoided.
+Use a combination of open and fextl::string APIs instead of fstream.
+
+### `std::string`
+Use `fextl::string` instead.
+
+#### std::to_string
+Use `fextl::fmt::format` instead.
```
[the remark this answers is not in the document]
Sonicadvance1: Added fmt::format to this

docs/ProgrammingConcerns.md
Sonicadvance1: Changed the comment around this stating how it is actually impossible to determine if memory is getting allocated or not.

docs/ProgrammingConcerns.md
Sonicadvance1: Removed this.

docs/ProgrammingConcerns.md
Sonicadvance1: Fixed.

docs/ProgrammingConcerns.md
Sonicadvance1: It's not the `std::fs::path` constructors that allocate memory. It allocates memory internally that we can't control.

docs/ProgrammingConcerns.md
Sonicadvance1: Modified

docs/ProgrammingConcerns.md
Sonicadvance1: Modified

docs/ProgrammingConcerns.md
Sonicadvance1: Modified

docs/ProgrammingConcerns.md
Sonicadvance1: Added a section

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,194 @@
...
+Use `FEXCore::Allocator::realloc`
+
+#### free
+Use `FEXCore::Allocator::free`
+
+#### aligned_alloc
+Use `FEXCore::Allocator::aligned_alloc`
+
+#### __libc_malloc
+**Don't use**
```
neobrain, line 173: This is formatted poorly, since the section nesting depth is so deep that the header gets printed in the same style as the `Don't use` text.

Sonicadvance1: Changed.

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,184 @@
+# Memory allocation routines
+## What is the problem?
+FEX-Emu needs to allocate memory differently than regular applications. This problem happens because FEX runs both 32-bit and 64-bit guest
+applications in the same address space as FEX itself. When running 32-bit applications, FEX reserves up all memory above 4GB in order to correctly
+emulate the 32-bit address space. We then use that reserved space for FEX allocations, so we don't interrupt application's own allocations.
+
+### Why not just just replace the system allocator?
```
neobrain, line 7:
```suggestion
### Why not just replace the system allocator?
```

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,184 @@
...
+jemalloc and similar allocators allow your application to replace the system allocator. This is something that FEX used to do, but this has since been
+removed because 32-bit thunks need to allocate memory in the lower 4GB.
+Overriding the system allocator is a global operation, including any libraries get that opened at runtime. This would cause thunks to use FEX's
+allocators. This would interfere with 32-bit thunks in such a way that their memory allocations would never end up in 32-bit virtual address space.
+Allocations made in native host libraries must always be valid 32-bit pointers for these 32-bit thunks to operate.
+
+With this approach, we couldn't find an easy way to customize allocation behaviour for the requirements of 32-bit thunking.
```
neobrain, line 14: This is a bit misleading now since we do use jemalloc to override the system allocator again (probably best to just drop the historical note).

I think I found a good concise wording that also streamlines this a little bit in general to first mention the requirements of 32-bit thunks and then the subsequent problem with overriding the allocator:

```suggestion
We could control the placement of FEX's internal allocations by overriding the system
allocator. However, 32-bit thunks (and their corresponding native host libraries) still
need to allocate memory in the lower 4 GB of memory so that they produce
guest-accessible pointers. Since overriding the system allocator is a global operation,
selectively overriding it like this is not possible.

Since we found no way to resolve this conflict, we had to resort to the alternative of avoiding use of the system allocator for FEX's internal allocations entirely (where possible).
```

Does that sound reasonable?

docs/ProgrammingConcerns.md
```diff
@@ -0,0 +1,184 @@
...
+Use `fextl::fmt::` instead
+
+### APIs that FEX doesn't have a replacement for
+Don't use any of these APIs in FEXLoader/FEXInterpreter. Shoutout to
+[this](https://stackoverflow.com/questions/43056338/standard-library-facilities-which-allocate-but-dont-use-an-allocator) StackOverflow post for this
+huge list.
+
+#### `std::any`
+#### `std::function` and lambdas
+One must take additional considerations when using these to ensure that they don't allocate memory. These don't have any way to replace which
```
neobrain, line 116: Lambdas never heap-allocate memory as far as I'm aware, even with large capture lists. So the use of `these` is a bit misleading.

What's dangerous (and worth highlighting hence) is passing an innocent-looking lambda as a `std::function` parameter, though.

neobrain (approved):
Added some final suggestions. Looks like we ended up with a nicely exhaustive list!

docs/ProgrammingConcerns.md
Sonicadvance1: Lambdas can allocate memory. I had to change this lambda because it was allocating memory: https://github.com/FEX-Emu/FEX/blob/main/External/FEXCore/Source/Interface/Config/Config.cpp#L725
Was allocating memory to store `EnvMapType EnvMap` from the capture list.

docs/ProgrammingConcerns.md
Sonicadvance1: Resolved.

docs/ProgrammingConcerns.md
Sonicadvance1: Sounds reasonable to me. Switched over to it.

docs/ProgrammingConcerns.md
neobrain: Are you sure this was actually from the lambda itself and not e.g. from accidentally copying the EnvMapType into a non-fextl container when capturing? `[&EnvMap]() {...}` really is just syntactic sugar for creating a stack-allocated class with the capture list as members, so I don't see how this possible could lead heap-allocate.

neobrain: Checking https://github.com/FEX-Emu/FEX/commit/1f9458a3c3c61de68b886e1e089db2c9f5a3835c, it looks like the lambda was assigned to an `std::function` before. The lambda itself was still stack-allocated, but `std::function` would copy the large lambda to heap-allocated storage.

Sonicadvance1: ah, that would make sense.

PR #2554, opened 2023-03-21
HostFeatures: Mark DCZID related utilities as [[maybe_unused]]

lioncash:
When building with the simulator enabled, these aren't used and cause some unused funtion/variable warnings.

We can silence these with [[maybe_unused]], since these are used in non-simulator builds.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/HostFeatures.cpp
+++ b/External/FEXCore/Source/Interface/Core/HostFeatures.cpp
@@ -17,12 +17,12 @@ namespace FEXCore {
 // Data Zero Prohibited flag
 // 0b0 = ZVA/GVA/GZVA permitted
 // 0b1 = ZVA/GVA/GZVA prohibited
-constexpr uint32_t DCZID_DZP_MASK = 0b1'0000;
+[[maybe_unused]] constexpr uint32_t DCZID_DZP_MASK = 0b1'0000;
 // Log2 of the blocksize in 32-bit words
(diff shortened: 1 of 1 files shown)


External/FEXCore/Source/Interface/Core/HostFeatures.cpp
```diff
@@ -17,12 +17,12 @@ namespace FEXCore {
...
 // 0b1 = ZVA/GVA/GZVA prohibited
-constexpr uint32_t DCZID_DZP_MASK = 0b1'0000;
+[[maybe_unused]] constexpr uint32_t DCZID_DZP_MASK = 0b1'0000;
 // Log2 of the blocksize in 32-bit words
-constexpr uint32_t DCZID_BS_MASK = 0b0'1111;
+[[maybe_unused]] constexpr uint32_t DCZID_BS_MASK = 0b0'1111;
 
 #ifdef _M_ARM_64
-static uint32_t GetDCZID() {
+[[maybe_unused]] static uint32_t GetDCZID() {
```
neobrain, line 25: For future reference: `inline` will also prevent the warning for functions and isn't as overly verbose as `[[maybe_unused]] static`. Not sure if that also works for `constexpr` variables though.

lioncash: Indeed, though I prefer being explicit when certain functions/variables may go unused, even if it's just a tad longer to be straightforward that way.

PR #2555, opened 2023-03-21
ARMEmitter: Handle SVE Element Count category

lioncash:


Sonicadvance1 (approved):
Weird that ARM decided to name these instruction mnemonics with a suffix stating the element size rather than just using the element size on the register argument like all the other instructions.

PR #2570, opened 2023-03-27
Arm64/VectorOps: Remove a few unnecessary EORs from comparisons in SVE path

lioncash:
We can make use of the zeroing variant of movprfx to accomplish the same thing


Sonicadvance1 (approved):
Nice catch

PR #2571, opened 2023-03-28
ARMEmitter: Fix treating 32-bit elements as 64-bit with ld1w

lioncash:
These conditionals were accidentally inverted and were treating 32-bit elements as 64-bit ones, when this is unintended.

Also add missing tests to ensure this doesn't slip through in the future.


Sonicadvance1 (approved):
Oops. Nice catch!

PR #2572, opened 2023-03-28
OpcodeDispatcher: Handle load variants of VMASKMOVPD/VMASKMOVPS

lioncash:
Introduces VLoadVectorMasked IR op to make this significantly less gross to implement (and collapse it to only two instructions).

After this, there's just the store variants to do and AVX 1 support is mostly in place


Sonicadvance1 (approved):
nice nice nice.

lioncash:
Yeah, I was going to do those all in one PR once the equivalent store IR op is in place, since the only real large thing in that PR would just be the unit-tests

Sonicadvance1:
Makes sense

PR #2574, opened 2023-03-29
Add support for thunking Wayland

neobrain:
This PR implements basic support for Wayland thunking on 64-bit guests. Two applications have been tested: [hello-wayland](https://github.com/emersion/hello-wayland) and vkcube.

Wayland is relatively simple to thunk since most of its API is exposed through an internal so-called *wire protocol*, whereby function signatures are encoded as ASCII strings and arguments are serialized to a stack array.

One obstacle for FEX is support for callbacks that also use this wire protocol. Our main job here is to decode the function signatures so that our host trampolines are aware of the arguments they're receiving. This is done by a handy little `ArgType` template which maps an ASCII character into a C type according to the wire protocol. Applying this to the entire signature string allows for constructing entire function signatures.

~~One open issue memory management for `wl_proxy_add_listener`: FEX must allocate a array of replacement function pointers, which must live until the proxy listener is unregistered (or the `wl_proxy` itself is destroyed). It's not clear how to best proceed here, so we'll just leak 160 bytes of memory each time a listener is registered for now.~~ Fixed in latest version: Since only one set of callbacks may be registered per proxy, the logic required to manage this memory is simpler than I expected at first.

### DIFF
--- a/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
+++ b/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
@@ -269,12 +269,12 @@ namespace FEXCore {
 
             auto Args = reinterpret_cast<LoadlibArgs*>(ArgsV);
 
-            auto Name = Args->Name;
+            std::string_view Name = Args->Name;
 
             auto SOName = (CTX->Config.Is64BitMode() ?
                            CTX->Config.ThunkHostLibsPath() :
                            CTX->Config.ThunkHostLibsPath32())
-                          + "/" + (const char*)Name + "-host.so";
+                          + "/" + Name.data() + "-host.so";
 
             LogMan::Msg::DFmt("LoadLib: {} -> {}", Name, SOName);
 
@@ -283,7 +283,10 @@ namespace FEXCore {
                 ERROR_AND_DIE_FMT("LoadLib: Failed to dlopen thunk library {}: {}", SOName, dlerror());
             }
 
-            const auto InitSym = fextl::string("fexthunks_exports_") + Name;
+            // Library names often include dashes, which may not be used in C++ identifiers.
+            // They are replaced with underscores hence.
+            auto InitSym = "fexthunks_exports_" + fextl::string { Name };
+            std::replace(InitSym.begin(), InitSym.end(), '-', '_');
 
             ExportEntry* (*InitFN)();
             (void*&)InitFN = dlsym(Handle, InitSym.c_str());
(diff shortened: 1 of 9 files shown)


ThunkLibs/libwayland-client/Guest.cpp
```diff
@@ -0,0 +1,235 @@
...
+}
+
+#define WL_CLOSURE_MAX_ARGS 20
+
+extern "C" int wl_proxy_add_listener(struct wl_proxy *proxy,
+      void (**callback)(void), void *data) {
+  auto interface = ((wl_proxy_private*)proxy)->interface;
+
+  // NOTE: This pointer must remain valid past the return of this function.
+  // TODO: What's a good way to manage the lifetime of this?
```
Sonicadvance1, line 67: Could probably create a map of proxies that have listeners attached and destroy the allocation on `wl_proxy_destroy`?

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -283,7 +283,14 @@ namespace FEXCore {
                 ERROR_AND_DIE_FMT("LoadLib: Failed to dlopen thunk library {}: {}", SOName, dlerror());
             }
 
-            const auto InitSym = fextl::string("fexthunks_exports_") + Name;
+            // Library names often include dashes, which may not be used in C++ identifiers.
+            // They are replaced with underscores hence.
+            auto InitSym = "fexthunks_exports_" + fextl::string { Name };
+            for (auto& c : InitSym) {
```
Sonicadvance1, line 289: Could use `fextl::string::replace` here I think?

neobrain: Maybe I'm overlooking a more useful overload, but `string::replace` seems to only replace characters *by position* rather than *by value*.

Sonicadvance1: Oh sorry, I was thinking `std::replace`

neobrain: Ohh, yes that would work. I thought it only replaced the first occurrence, but turns out it replaces all. Thanks!

Sonicadvance1 (approved):
Seems reasonable enough to me once the nits and CI passes.

ThunkLibs/libwayland-client/Guest.cpp
neobrain: Yes, was a bit worried about potential other ways to destroy proxies and about multiple listeners being set up. But proxy wrappers don't get their own set of listener callbacks, and `wl_proxy_add_listener` can't be called more than once per proxy. So fixing this was fairly simple after all :)

PR #2582, opened 2023-04-01
Thunks: Enable ccache if available

Sonicadvance1:
More vrooms.

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -8,6 +8,12 @@ if (ENABLE_CLANG_THUNKS)
   add_link_options(${LD_OVERRIDE})
 endif()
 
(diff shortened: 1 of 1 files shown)


ThunkLibs/HostLibs/CMakeLists.txt
```diff
@@ -10,6 +10,12 @@ if (ENABLE_CLANG_THUNKS)
   add_link_options(${LD_OVERRIDE})
 endif()
 
+find_program(CCACHE_PROGRAM ccache)
```
neobrain, line 13: Does HostLibs need this? Host thunks are compiled using the host toolchain and hence follow the setup from the main `CMakeLists.txt`, which already sets up ccache AFAICT.

Sonicadvance1: You're right, removed.

neobrain (approved):

PR #2583, opened 2023-04-03
Thunks: Make xcb's callback more robust.

Sonicadvance1:
Fixes a crash in xcb thunks since more stuff on FEX side has moved over to jemalloc.

Destructors don't actually get called when a shared library is removed. It's some weirdo quirk that we can't work around. Instead refcount Display connections being created and disconnected. Creating the thread on the first display creation, and tearing down on final display teardown.

Must be merged before #2564 otherwise that PR will break thunks.

### DIFF
--- a/ThunkLibs/libxcb/libxcb_Guest.cpp
+++ b/ThunkLibs/libxcb/libxcb_Guest.cpp
@@ -30,6 +30,8 @@ $end_info$
 
 static std::thread CBThread{};
 static std::atomic<bool> CBDone{false};
+static std::mutex CBThreadMutex;
+static uint32_t ScreenRefCount;
 
 static CrossArchEvent WaitForWork{};
 static CrossArchEvent WorkDone{};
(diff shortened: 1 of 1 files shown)


ThunkLibs/libxcb/libxcb_Guest.cpp
```diff
@@ -117,17 +129,33 @@ extern "C" {
 
   xcb_connection_t * xcb_connect(const char * a_0,int * a_1){
     auto ret = fexfn_pack_xcb_connect(a_0, a_1);
+
+    if (xcb_get_file_descriptor(ret) != -1) {
```
neobrain, line 133: XCB documentation suggests `xcb_connection_has_error` should be used for error checking.

ThunkLibs/libxcb/libxcb_Guest.cpp
```diff
@@ -30,6 +30,8 @@ tags: thunklibs|xcb
 
 static std::thread CBThread{};
 static std::atomic<bool> CBDone{false};
+static std::mutex CBThreadMutex;
+static std::atomic<uint32_t> ScreenRefCount;
```
neobrain, line 34: This doesn't need to be atomic since it's guarded by a mutex.

neobrain (approved):
The general observation that destructors don't get called on library unload is concerning... can we hook into `dlclose` to trigger a manual cleanup for other libraries?

The XCB change itself looks fine.

ThunkLibs/libxcb/libxcb_Guest.cpp
Sonicadvance1: We could use that on connect, but we can't use that on disconnect. I opted to use the same valid display connection testing for all three instances to be consistent. I could switch the two connect ones over if you want.

Sonicadvance1:
> can we hook into dlclose to trigger a manual cleanup for other libraries?

Once we have redirected the FD for the thunk library to be loaded, we lose control of this. I haven't found a good workaround.

ThunkLibs/libxcb/libxcb_Guest.cpp
neobrain: Fair enough. `xcb_get_file_descriptor` is fine then.

neobrain:
Does compiling with `--fno-gnu-unique` cause destructors to be invoked on `dlclose` by any chance?

neobrain:
Ah, bummer. Thanks for testing though!

PR #2585, opened 2023-04-05
Allocator: Remove pointer indirection overhead

Sonicadvance1:
Every time we are calling a function in `FEXCore::Allocator::` this is a
pointer indirection. Which means on x86 it is always a `call [rdi]` and
on AArch64 it is a `ldr x17, [x0]; blr x17;`.

Instead of doing this, use inline functions in the header that call the
correct allocation function directly. This function gets inlined and is
no longer an indirect call.

When compiling with jemalloc, we forward declare the jemalloc function
definitions so we don't have to pull in the entire jemalloc interface in
to the public header definitions.

### DIFF
--- a/External/FEXCore/Source/Utils/AllocatorOverride.cpp
+++ b/External/FEXCore/Source/Utils/AllocatorOverride.cpp
@@ -53,15 +53,15 @@ namespace FEXCore::Allocator {
   static thread_local uint64_t SkipEvalForThread{};
 
   // Internal memory allocation hooks to allow non-faulting allocations through.
-  CALLOC_Hook calloc_ptr = __libc_calloc;
-  FREE_Hook free_ptr = __libc_free;
-  MALLOC_Hook malloc_ptr = __libc_malloc;
-  MEMALIGN_Hook memalign_ptr = __libc_memalign;
-  REALLOC_Hook realloc_ptr = __libc_realloc;
-  VALLOC_Hook valloc_ptr = __libc_valloc;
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Updated this to remove the other commits that were unnecessary. This is now ready to merge.

External/FEXCore/Source/Utils/AllocatorOverride.cpp
```diff
@@ -53,6 +53,16 @@ namespace FEXCore::Allocator {
...
+  using MEMALIGN_Hook = void *(*)(size_t, size_t);
+  using VALLOC_Hook = void *(*)(size_t);
+  using POSIX_MEMALIGN_Hook = int (*)(void**, size_t, size_t);
+  using REALLOC_Hook = void *(*)(void*, size_t);
+  using FREE_Hook = void(*)(void*);
+  using MALLOC_USABLE_SIZE_Hook = size_t (*)(void*);
+  using ALIGNED_ALLOC_Hook = void *(*)(size_t, size_t);
+
   // Internal memory allocation hooks to allow non-faulting allocations through.
   CALLOC_Hook calloc_ptr = __libc_calloc;
```
neobrain, line 67: Can we just use `auto` to declare these? Then we don't need most of the extra typedefs above.

neobrain (approved):

External/FEXCore/Source/Utils/AllocatorOverride.cpp
Sonicadvance1: You're right. Switched over to auto for these.

PR #2586, opened 2023-04-05
Dispatcher: Fixes restoring of AVX state

AndreRH:
Found this copy&paste issue while doing the Hangover integration.
Though only checked that compilation doesn't introduce new Warnings.


Sonicadvance1 (approved):
Good catch, definitely hecked this up. Especially with us accidentally leaking some AVX state around even though it isn't fully wired up.

PR #2587, opened 2023-04-06
Allocator: Disable glibc sbrk allocations

Sonicadvance1:
This is done by consuming a single page at the end of the current sbrk memory region. Then consuming any remaining bytes that could have potentially ended up in it.

This ensures that glibc won't be able to return 64-bit pointers to 32-bit thunks once the remaining work is in place.

This can be merged independently of the other pull requests.

### DIFF
--- a/External/FEXCore/Source/Utils/Allocator.cpp
+++ b/External/FEXCore/Source/Utils/Allocator.cpp
@@ -2,6 +2,7 @@
 #include <FEXCore/Utils/Allocator.h>
 #include <FEXCore/Utils/CompilerDefs.h>
 #include <FEXCore/Utils/LogManager.h>
+#include <FEXCore/fextl/fmt.h>
 #include <FEXCore/fextl/memory_resource.h>
 #include <FEXHeaderUtils/Syscalls.h>
 #include <FEXHeaderUtils/TypeDefines.h>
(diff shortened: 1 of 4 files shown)


External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -91,6 +92,44 @@ namespace FEXCore::Allocator {
     return Result;
   }
 
+  void *DisableSBRKAllocations() {
```
neobrain, line 95: Should document somewhere *why* this is useful, and what effect it will have on glibc. IIUC we're effectively making brk/sbrk always fail, so why doesn't this break anything? Does glibc have a fallback path if sbrk doesn't work?

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -91,6 +92,44 @@ namespace FEXCore::Allocator {
...
+
+    // Now allocate the next page after the sbrk address to ensure it can't grow.
+    // In most cases at the start of `main` this will already be page aligned, which means subsequent `sbrk`
+    // calls won't allocate any memory through that.
+    void* AlignedBRK = reinterpret_cast<void*>(FEXCore::AlignUp(reinterpret_cast<uint64_t>(StartingSBRK), FHU::FEX_PAGE_SIZE));
+    void *AfterBRK = mmap(AlignedBRK, FHU::FEX_PAGE_SIZE, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED_NOREPLACE | MAP_NORESERVE, -1, 0);
+    if (AfterBRK == INVALID_PTR) {
+      // Couldn't allocate the page after the aligned brk? This should never happen.
+      // FEXCore::LogMan isn't configured yet so we just need to print the message.
+      fextl::fmt::print("Couldn't allocate page after SBRK. Expect 32-bit thunks to break.\n");
```
neobrain, line 112: Seems like `ERROR_AND_DIE` is more suitable here, since it's a scenario we don't anticipate will happen. If it did, we'd want to notice quickly and see if we can do something about it.

External/FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -91,6 +92,44 @@ namespace FEXCore::Allocator {
...
+    void *StartingSBRK = sbrk(0);
+    if (StartingSBRK == INVALID_PTR) {
+      // If sbrk is already returning invalid pointers then nothing to do here.
+      return INVALID_PTR;
+    }
+
+    // Now allocate the next page after the sbrk address to ensure it can't grow.
+    // In most cases at the start of `main` this will already be page aligned, which means subsequent `sbrk`
+    // calls won't allocate any memory through that.
+    void* AlignedBRK = reinterpret_cast<void*>(FEXCore::AlignUp(reinterpret_cast<uint64_t>(StartingSBRK), FHU::FEX_PAGE_SIZE));
```
neobrain, line 107:
```suggestion
    void* AlignedBRK = reinterpret_cast<void*>(FEXCore::AlignUp(reinterpret_cast<uintptr_t>(StartingSBRK), FHU::FEX_PAGE_SIZE));
```

neobrain (approved):
Funky what kind of arcane APIs still exist... Change looks good though!

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Utils/Allocator.cpp
Sonicadvance1: Can't use an `ERROR_AND_DIE` here since it is before LogManager is setup, but I did add a `FEX_TRAP_EXECUTION` just like `ERROR_AND_DIE` to behave the same but with output to stderr.

PR #2588, opened 2023-04-07
External: Update jemalloc

Sonicadvance1:
First 39 commits are from https://github.com/FEX-Emu/FEX/pull/2564 and that needs to be merged first.
Also needs #2580 to be merged first.

- Disables sbrk allocation.
- Disable TLS `initial-exec` model
   - Necessary for dlopen of libFEXCore

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -48,7 +48,7 @@
 [submodule "External/robin-map"]
 	shallow = true
 	path = External/robin-map
-	url = https://github.com/Tessil/robin-map.git
+	url = https://github.com/FEX-Emu/robin-map.git
 [submodule "External/Vulkan-Headers"]
 	shallow = true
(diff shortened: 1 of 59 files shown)


Sonicadvance1:
Looks like this needs XFree to change in libX11_Guest since we no longer have glibc overrides.
It's trying to determine if it is a host allocation or not and choosing host XFree or `free` depending. Now that glibc overriding is disabled the `IsHostHeapAllocation` check always returns false which always sends it down the `free` path. Which results in guest glibc trying to free the pointer and crashing.

We should now just pass XFree over to the host XFree since that will be using the expected glibc allocator (Host versus guest).

Sonicadvance1:
Closing this as I need to change how the jemalloc behaviour works.

PR #2590, opened 2023-04-10
Update drm headers to v6.2

Sonicadvance1:
Nothing here that changes anything really. Just need to handle `__user` define and a new Aligned Attribute for clang.
i915 still has some changes that we might not be picking up, but that is a known issue and considering i915 doesn't compile on non-x86 platforms I'm inclined to not care.

No uapi changes otherwise.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/x32/Ioctl/drm.h
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/x32/Ioctl/drm.h
@@ -7,6 +7,9 @@
 
 #include <cstdint>
 extern "C" {
+// drm headers use a `__user` define that has an address_space attribute. This allows their tooling to see unsafe user-space accesses.
+// Define this to nothing so we don't need to modify those headers.
+#define __user
 #include "fex-drm/drm.h"
 #include "fex-drm/drm_mode.h"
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/x32/Ioctl/drm.h
```diff
@@ -7,6 +7,7 @@
 
 #include <cstdint>
 extern "C" {
+#define __user
```
neobrain, line 10: It's mentioned that this is needed, but what is it actually used for and where is it normally defined?

Sonicadvance1: It's used in the drm headers themselves. I previously modified them all to remove the attribute but that modification has been the biggest change for every drm header uprev so I've stopped so we can use them unmodified from upstream.

neobrain (approved):

Source/Tools/FEXLoader/LinuxSyscalls/x32/Ioctl/drm.h
Sonicadvance1: Added a comment.

PR #2591, opened 2023-04-11
Add in jemalloc glibc hooking again 

### LINKED ISSUE #2565: can't dlopen: libFEXCore.so: cannot allocate memory in static TLS block
Hi,
I'd like to use libFEXCore in Hangover ( https://github.com/AndreRH/hangover/ ), but a major blocker is that I can't dlopen it.
The message "libFEXCore.so: cannot allocate memory in static TLS block" seems to indicate that the library want's to use more TLS memory than statically available, but even if I (for testing purposes) remove all thread_locals from libFEXCore, I get the same message. Any idea?

Note that using LD_PRELOAD works for a test application, but won't be viable for Hangover/Wine

Sonicadvance1:
We still need to hook glibc for thunks to work with
`IsHostHeapAllocation`.
So now we link in two jemalloc allocators in different namespaces.

As usual we have multiple heap allocators that we need to be careful about.

1. jemalloc with `je_` namespace.
  - This is FEX's regular heap allocator and what gets used for all the
    fextl objects.
  - This allocator is the one that the FEX mmap/munmap hooks hook in to
     - This mmap hooking gives this allocator the full 48-bit VA even in
       32-bit space.

2. jemalloc with `glibc_je_` namespace.
  - This is the allocator that overrides the glibc allocator routines
  - This is the allocator that native libraries (including the host side of thunks) will use.
  - This is what `IsHostHeapAllocation` will check for.

3. Guest glibc allocator
  - We don't touch this one. But it is distinct from the host side
    allocators.
  - The guest side of thunks will use this heap allocator.

4. Host glibc allocator
  - Allocator 2 replaces this one unless explicitly disabled.
  - Always expected to override the allocator, so this configuration
    isn't expected.

Already tested this with Dota Underlords to ensure this works with
thunks.

Fixes #2565

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -53,3 +53,6 @@
 	shallow = true
 	path = External/Vulkan-Headers
 	url = https://github.com/KhronosGroup/Vulkan-Headers.git
+[submodule "External/jemalloc_glibc"]
+	path = External/jemalloc_glibc
+	url = https://github.com/FEX-Emu/jemalloc.git
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,6 +22,7 @@ option(ENABLE_VISUAL_DEBUGGER "Enables the visual debugger for compiling" FALSE)
 option(ENABLE_STRICT_WERROR "Enables stricter -Werror for CI" FALSE)
 option(ENABLE_WERROR "Enables -Werror" FALSE)
 option(ENABLE_JEMALLOC "Enables jemalloc allocator" TRUE)
+option(ENABLE_JEMALLOC_GLIBC_ALLOC "Enables jemalloc glibc allocator" TRUE)
 option(ENABLE_OFFLINE_TELEMETRY "Enables FEX offline telemetry" TRUE)
 option(ENABLE_COMPILE_TIME_TRACE "Enables time trace compile option" FALSE)
 option(ENABLE_LIBCXX "Enables LLVM libc++" FALSE)
@@ -55,6 +56,10 @@ if (ENABLE_FEXCORE_PROFILER)
   else()
     message(FATAL_ERROR "Unknown FEXCore profiler backend ${FEXCORE_PROFILER_BACKEND}")
   endif()
+endif()
+
+if (ENABLE_JEMALLOC_GLIBC_ALLOC AND ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT)
+  message(FATAL_ERROR "Can't have both glibc fault allocator and jemalloc glibc allocator enabled at the same time")
 endif()
 
 if (ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT)
@@ -190,7 +195,22 @@ if (ENABLE_TSAN)
   link_libraries(-fno-omit-frame-pointer -fsanitize=thread)
 endif()
 
+if (ENABLE_JEMALLOC_GLIBC_ALLOC)
+  # The glibc jemalloc subproject which hooks the glibc allocator.
+  # Required for thunks to work.
+  # All host native libraries will use this allocator, while *most* other FEX internal allocations will use the other jemalloc allocator.
+  add_definitions(-DENABLE_JEMALLOC_GLIBC=1)
+  add_subdirectory(External/jemalloc_glibc/)
+else()
+  message (STATUS
+    " jemalloc glibc allocator disabled!\n"
+    " This is not a recommended configuration!\n"
+    " This will very explicitly break thunk execution!\n"
+    " Use at your own risk!")
+endif()
+
 if (ENABLE_JEMALLOC)
+  # The jemalloc subproject that all FEXCore fextl objects allocate through.
   add_definitions(-DENABLE_JEMALLOC=1)
   add_subdirectory(External/jemalloc/)
   include_directories(External/jemalloc/pregen/include/)
--- a/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
+++ b/External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
@@ -29,8 +29,14 @@ $end_info$
 #include <stdint.h>
 #include <utility>
 
-#ifdef ENABLE_JEMALLOC
-#include "jemalloc/jemalloc.h"
+#ifdef ENABLE_JEMALLOC_GLIBC
+extern "C" {
+// jemalloc defines nothrow on its internal C function signatures.
+#define JEMALLOC_NOTHROW __attribute__((nothrow))
+// Forward declare jemalloc functions because we can't include the headers from the glibc jemalloc project.
+// This is because we can't simultaneously set up include paths for both of our internal jemalloc modules.
+FEX_DEFAULT_VISIBILITY JEMALLOC_NOTHROW extern int glibc_je_is_known_allocation(void *ptr);
+}
 #endif
 
 struct LoadlibArgs {
@@ -251,13 +257,13 @@ namespace FEXCore {
          * and host heap pointers.
          */
         static void IsHostHeapAllocation(void* ArgsRV) {
-#ifdef ENABLE_JEMALLOC
+#ifdef ENABLE_JEMALLOC_GLIBC
             struct ArgsRV_t {
                 void* ptr;
                 bool rv;
             } *args = reinterpret_cast<ArgsRV_t*>(ArgsRV);
 
-            args->rv = je_is_known_allocation(args->ptr);
+            args->rv = glibc_je_is_known_allocation(args->ptr);
 #else
             // Thunks usage without jemalloc isn't supported
             ERROR_AND_DIE_FMT("Unsupported: Thunks querying for host heap allocation information");
--- a/docs/jemalloc_usage.md
+++ b/docs/jemalloc_usage.md
@@ -1,0 +1,35 @@
+# JEMalloc usage
+FEX-Emu uses two different jemalloc heap allocators at once, each for different purposes:
+- jemalloc: The primary heap allocator (to keep FEX's internal allocations out of the 32-bit address space used by guest applications)
+- jemalloc_glibc: The second heap allocator (to add allocation introspection features used by thunks)
+
+## jemalloc - primary heap allocator
+This allocator overrides `mmap` and `munmap` by forwarding them to FEXCore's internal VMA region allocator.
+
+All of FEXCore's `fextl::` namespaced objects allocate memory with this method.
+
+### FEXCore internal VMA region allocator
+When running a 32-bit guest application, the VMA region allocator allocates from memory *above* the 4GB of virtual address space reserved for the
+32-bit application.
+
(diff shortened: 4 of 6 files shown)


neobrain:
Do we have existing documentation on our use of jemalloc that should be updated for this?

AFAIK this PR description is the only documentation we have so far, but it took me too many read-throughs to understand what the descriptions for each of the 4 heaps are trying to say. Future-us and new developers would benefit from having more a accessible overview.

For example, this bit in heap 3 was really confusing to me:
> This is the allocator that overrides the glibc allocator routines
> This is the allocator that thunks will use.

A key missing point here is that this allocator will *also* be used by native libraries, which is the reason this PR works in the first place. ("Thunk" is a bit of an overloaded term, but we really shouldn't use it to cover the thunked libraries themselves.)

.gitmodules
```diff
@@ -53,3 +53,6 @@
 	shallow = true
 	path = External/Vulkan-Headers
 	url = https://github.com/KhronosGroup/Vulkan-Headers.git
+[submodule "External/jemalloc_glibc"]
```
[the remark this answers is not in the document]
Sonicadvance1: Each submodule is using a different branch with different preconfigured headers.
`jemalloc` using branch `v5.3.0`
`jemalloc_glibc` using branch `v5.3.0_glibc`

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -29,8 +29,13 @@ tags: glue|thunks
 #include <stdint.h>
 #include <utility>
 
-#ifdef ENABLE_JEMALLOC
-#include "jemalloc/jemalloc.h"
+#ifdef ENABLE_JEMALLOC_GLIBC
+extern "C" {
+// jemalloc defines nothrow on its internal C function signatures.
+#define JEMALLOC_NOTHROW __attribute__((nothrow))
+// Forward declare jemalloc functions because we can't include the headers from the glibc jemalloc project.
```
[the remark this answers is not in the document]
Sonicadvance1: This comes down to cmake not being able to do per-file different header include directories.
Two independent jemalloc prefixes have the same include path `include/jemalloc/` and `pregen/include/jemalloc/`, we can't rename the paths otherwise the headers will fail to find some includes.
Since we are using two independent jemalloc implementations these include paths would intersect and means depending on search order we will get the `je_` prefixed function definitions or the `glibc_je_` prefixed function definitions depending.

Sonicadvance1:
> Do we have existing documentation on our use of jemalloc that should be updated for this?

No we don't.

> 
> AFAIK this PR description is the only documentation we have so far, but it took me too many read-throughs to understand what the descriptions for each of the 4 heaps are trying to say. Future-us and new developers would benefit from having more a accessible overview.
> 
> For example, this bit in heap 3 was really confusing to me:
> 
> > This is the allocator that overrides the glibc allocator routines
> > This is the allocator that thunks will use.
> 
> A key missing point here is that this allocator will _also_ be used by native libraries, which is the reason this PR works in the first place. ("Thunk" is a bit of an overloaded term, but we really shouldn't use it to cover the thunked libraries themselves.)

CMakeLists.txt
```diff
@@ -190,6 +195,17 @@ if (ENABLE_TSAN)
   link_libraries(-fno-omit-frame-pointer -fsanitize=thread)
 endif()
 
+if (ENABLE_JEMALLOC_GLIBC_ALLOC)
```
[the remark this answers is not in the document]
Sonicadvance1: Added a comment

Sonicadvance1:
Added a document with a brief overview of jemalloc usage.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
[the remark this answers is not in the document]
neobrain:
```suggestion
// Forward declare jemalloc functions because we can't include the headers from the glibc jemalloc project.
// This is because we can't simultaneously set up include paths for both of our internal jemalloc modules.
```

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
+# JEMalloc usage
+FEX-Emu uses two different jemalloc heap allocators at once.
+- jemalloc: The primary heap allocator
+- jemalloc_glibc: The second heap allocator
+
+## jemalloc - primary heap allocator
+This heap allocator is the primary allocator that FEX-Emu uses. All of FEXCore's `fextl::` namespaced objects will allocate memory through this
+allocator. This allocator has its usage of mmap and munmap overridden to use FEXCore's internal VMA region allocator.
```
neobrain, line 8: Phrasing is a bit roundabout here. I'd describe what this actually does first, then where it's used (since the section header already mentions it's the primary allocator):

```suggestion
This allocator overrides `mmap` and `munmap` by forwarding them to FEXCore's internal VMA region allocator.

All of FEXCore's `fextl::` namespaced objects allocate memory with this method.
```

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
...
+
+### FEXCore internal VMA region allocator
+- When running a 64-bit guest application, this VMA region allocator is disabled and passes through to mmap and munmap in the host kernel.
+- This only gets interesting when running a 32-bit guest application.
+
+When running a 32-bit guest application, the VMA region allocator will allocate all memory above 4GB virtual address space (Up to 256TB on 48-bit VA
+systems!). It then provides a mmap and munmap interface to allow allocating memory out of this reserved virtual address space. This ensures that all
+of FEX's allocations stay out of the lower 32-bit 4GB VA space, otherwise games would quickly run out of virtual address space.
```
neobrain, line 16: This is backwards: Instead of starting with the case where literally nothing happens, start with the interesting bit:

```suggestion
When running a 32-bit guest application, the VMA region allocator allocates from memory *above* the 4GB of virtual address space reserved for the 32-bit application.

This ensures that all of FEX's allocations stay out of the lower 32-bit 4GB VA space, since games would quickly run out of virtual address space otherwise.

When running a 64-bit guest application, this VMA region allocator is disabled and passes through to mmap and munmap in the host kernel.
```

> (Up to 256TB on 48-bit VA systems!)

Since we're describing fairly intricate logic here, omitting details like this makes it easier to digest the important pieces.

> It provides a mmap and munmap interface to allow allocating memory out of this reserved virtual address space.

This seems redundant. It's an allocator, so of course it provides those interfaces.

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
...
+
+## jemalloc_glibc - secondary heap allocator
+This heap allocator replaces glibc's allocator using weak symbol overriding. This explicitly does **not** use the FEXCore VMA region allocator. This
```
neobrain, line 19: > This explicitly does **not** use the FEXCore VMA region allocator

This is backwards again: It's starting to explain something by explaining what it's not.

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
...
+ensures that all allocations that go through this heap allocator will use the kernel mmap and munmap interface. All shared libraries will use this
+allocator including the host-side of thunks.
```
neobrain, line 21: > This explicitly does not use the FEXCore VMA region allocator

This is backwards again: It's starting to explain something by explaining what it's not.

> This ensures that all allocations that go through this heap allocator will use the kernel mmap and munmap interface.

This describes the technical implementation but not what the allocator intends to achieve.

```suggestion
This heap allocator replaces the host glibc's allocator using weak symbol overriding. It adds introspection features used for thunking, but has no functional differences otherwise: All memory is allocated in the 4GB of 32-bit address space. The FEXCore VMA region allocator is explicitly **not** involved hence.

All native shared libraries use this allocator including the host-side of thunks.

Internally, all allocations that go through this heap allocator use the kernel mmap and munmap interface.
```

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
...
+
+### Thunks
+When thunks allocate memory, some pointers need to be tracked from which heap allocator they come from when getting freed. In this case some pointers
+might be coming from the guest heap-allocator, and some might be coming from the host heap-allocator.
+
+jemalloc provides a `is_known_allocation` function that allows us to query if the pointer comes from the host heap-allocator. Which FEXCore can use to
+determine which heap should free the pointer. glibc doesn't provide this same interface which is why we need to use jemalloc for this.
```
neobrain, line 28:
```suggestion
Thunks may allocate memory either through the guest-side (on the guest glibc heap) or the host-side (on the `jemalloc_glibc` heap). To properly free this memory, FEX must be able to determine which heap allocator it belongs to.

`glibc` provides no public interface to do this, but FEX's `jemalloc` fork does. The `is_known_allocation` function is used by FEXCore to query whether a given pointer originated from the `jemalloc_glibc` allocator. This enables FEXCore to determine the appropriate heap for freeing the pointer.
```

(`verify`->`query`, since the former suggests the condition is expected to hold true)

docs/jemalloc_usage.md
```diff
@@ -0,0 +1,29 @@
+# JEMalloc usage
+FEX-Emu uses two different jemalloc heap allocators at once.
+- jemalloc: The primary heap allocator
+- jemalloc_glibc: The second heap allocator
```
neobrain, line 4:
```suggestion
FEX-Emu uses two different jemalloc heap allocators at once, each for different purposes:
- jemalloc: The primary heap allocator (to keep FEX's internal allocations out of the 32-bit address space used by guest applications)
- jemalloc_glibc: The second heap allocator (to add allocation introspection features used by thunks)
```

neobrain (approved):
Thanks for adding the document! Did a cleanup pass over it and added some clarifications. Change itself is fine as-is.

Definitely a relief that we don't need another solution for X11 thunks after all.

External/FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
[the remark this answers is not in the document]
Sonicadvance1: Updated comment

docs/jemalloc_usage.md
Sonicadvance1: Changed phrasing to yours.

docs/jemalloc_usage.md
Sonicadvance1: Changed phrasing to yours.

docs/jemalloc_usage.md
Sonicadvance1: Changed phrasing to yours.

docs/jemalloc_usage.md
Sonicadvance1: Changed phrasing to yours.

docs/jemalloc_usage.md
Sonicadvance1: Changed phrasing to yours.

PR #2592, opened 2023-04-11
fextl::fmt: Remove fwrite usage

Sonicadvance1:
fwrite allocates some backing memory for buffering outputs which FEX can't track.

Switch to using `fileno` to get the fd from the FILE and write directly. This will need to be changed for llvm-mingw support but that will come after this.

This will be added to the documentation that we can't use fwrite.


neobrain (approved):
What a rabbit hole 😬

Issue #2594, opened 2023-04-11
FEX might leak stacks from threads using exit/cancel


Sonicadvance1:
Noticed this when looking at fault checking. Need to ensure we're not leaking stacks.

Sonicadvance1:
Looks like FEX is leaking something around 16-40KB per thread but it looks like heaptrack doesn't see it (mmap related?)
And valgrind is just broken now? Throws an error even though debug packages are already installed.
```bash
$ /usr/local/bin/valgrind ./Bin/FEXInterpreter /mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/leaked_stacks_test
==331992== Memcheck, a memory error detector
==331992== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==331992== Using Valgrind-3.20.0 and LibVEX; rerun with -h for copyright info
==331992== Command: ./Bin/FEXInterpreter /mnt/Work/Work/Unicorn_Tests/ELFLoader/Tests/leaked_stacks_test
==331992==
==331992== brk segment overflow in thread #1: can't grow to 0x483c000
==331992== (see section Limitations in user manual)
==331992== NOTE: further instances of this message will not be shown

valgrind:  Fatal error at startup: a function redirection
valgrind:  which is mandatory for this platform-tool combination
valgrind:  cannot be set up.  Details of the redirection are:
valgrind:
valgrind:  A must-be-redirected function
valgrind:  whose name matches the pattern:      strlen
valgrind:  in an object with soname matching:   ld-linux-x86-64.so.2
valgrind:  was not found whilst processing
valgrind:  symbols from the object with soname: ld-linux-x86-64.so.2
valgrind:
valgrind:  Possible fixes: (1, short term): install glibc's debuginfo
valgrind:  package on this machine.  (2, longer term): ask the packagers
valgrind:  for your Linux distribution to please in future ship a non-
valgrind:  stripped ld.so (or whatever the dynamic linker .so is called)
valgrind:  that exports the above-named function using the standard
valgrind:  calling conventions for this platform.  The package you need
valgrind:  to install for fix (1) is called
valgrind:
valgrind:    On Debian, Ubuntu:                 libc6-dbg
valgrind:    On SuSE, openSuSE, Fedora, RHEL:   glibc-debuginfo
valgrind:
valgrind:  Note that if you are debugging a 32 bit process on a
valgrind:  64 bit system, you will need a corresponding 32 bit debuginfo
[...]

PR #2595, opened 2023-04-11
Threads: Disable glibc allocator fault testing with `exit`

Sonicadvance1:
FEX might not ever make it to its cleanup routines when exit is used. So make sure to disable the fault checking if exit is used.

Opened #2594 to track this investigation.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
@@ -292,6 +292,11 @@ namespace FEX::HLE {
     REGISTER_SYSCALL_IMPL_FLAGS(exit, SyscallFlags::OPTIMIZETHROUGH | SyscallFlags::NOSYNCSTATEONENTRY | SyscallFlags::NORETURN,
       [](FEXCore::Core::CpuStateFrame *Frame, int status) -> uint64_t {
       auto Thread = Frame->Thread;
+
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -292,6 +292,11 @@ namespace FEX::HLE {
     REGISTER_SYSCALL_IMPL_FLAGS(exit, SyscallFlags::OPTIMIZETHROUGH | SyscallFlags::NOSYNCSTATEONENTRY | SyscallFlags::NORETURN,
       [](FEXCore::Core::CpuStateFrame *Frame, int status) -> uint64_t {
       auto Thread = Frame->Thread;
+
+      // TLS/DTV teardown is something FEX can't control. Disable glibc checking when we leave a pthread.
+      // Since this thread is hard stopping, we can't track this in FEX's thread handling.
```
neobrain, line 297: > we can't track this

What is this referring to?

neobrain (approved):
Bit confused about the code comment, looks fine otherwise.

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
Sonicadvance1: Changed the comment that it is about the TLS/DTV teardown.

PR #2596, opened 2023-04-11
StringConv: Convert to conversion functions that don't use std::string

Sonicadvance1:
`std::stoul` and `std::stroull` take a std::string which was converting the string_view to a std::string first. Causing glibc fault testing to catch this since not much uses this.

These will be added to the documentation.


neobrain (approved):
The implementations of these give me a bit of a headache, since we have no proper error handling code and just always return success currently. Should ideally move to something like `std::from_chars` since strtoull doesn't allow distinguish an error case from a successful parse of `0`.

Anyway, this change itself is fine as-is.

PR #2607, opened 2023-04-16
llvm-mingw: Fix SoftFloat compiling

Sonicadvance1:


External/FEXCore/Source/Common/SoftFloat.h
```diff
@@ -562,4 +568,7 @@ struct X80SoftFloat {
   static constexpr uint32_t ExponentBias = 16383;
 };
 
+#ifndef _WIN32
+// Padding on this extends to 16-bytes rather than 10-bytes on WIN32.
```
lioncash, line 572: Maybe add a static assert for that size on the windows side too

Sonicadvance1: Good shout, added a WIN32 static assert

lioncash (approved):

PR #2608, opened 2023-04-16
CPUID: Fix std::min type cast

Sonicadvance1:


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -969,14 +969,14 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0004h(uint32_t Leaf) {
 FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0002h(uint32_t Leaf, uint32_t CPU) {
   FEXCore::CPUID::FunctionResults Res{};
   memset(&Res, ' ', sizeof(FEXCore::CPUID::FunctionResults));
-  memcpy(&Res, &ProcessorBrand[0], std::min(16L, DESCRIBE_STR_SIZE));
+  memcpy(&Res, &ProcessorBrand[0], std::min(static_cast<ssize_t>(16L), DESCRIBE_STR_SIZE));
```
lioncash, line 972: Seems like these can just use the type like

```cpp
ssize_t{16}
```

instead of using static_cast

Sonicadvance1: You're right. Converted.

lioncash (approved):

PR #2632, opened 2023-04-17
OpcodeDispatcher: Handle PCMPESTRI/VPCMPESTRI

lioncash:
Handles the rather behavior-abundant PCMPESTRI instruction variants.

We only implement this instruction in this PR, since the fallback itself on top of the tests are quite large. This allows the base implementation to get reviews, making it much nicer to review following instructions that will add handling for implicit lengths and mask handling

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -449,6 +449,46 @@ void Arm64JITCore::Op_Unhandled(IR::IROp_Header const *IROp, IR::NodeID Node) {
         ins(ARMEmitter::SubRegSize::i16Bit, Dst, 4, ARMEmitter::Reg::r1);
       }
       break;
+      case FABI_I32_I64_I64_I128_I128_I16: {
+        SpillStaticRegs();
+        PushDynamicRegsAndLR(TMP1);
+
+        const auto Op = IROp->C<IR::IROp_VPCMPESTRX>();
+        const auto Is64Bit = Op->GPRSize == 8;
+
+        const auto Src1 = GetVReg(Op->LHS.ID());
+        const auto Src2 = GetVReg(Op->RHS.ID());
+        const auto SrcRAX = GetReg(Op->RAX.ID());
+        const auto SrcRDX = GetReg(Op->RDX.ID());
+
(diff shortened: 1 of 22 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -449,6 +449,50 @@ void Arm64JITCore::Op_Unhandled(IR::IROp_Header const *IROp, IR::NodeID Node) {
...
+        const auto SrcRAX = GetReg(Op->RAX.ID());
+        const auto SrcRDX = GetReg(Op->RDX.ID());
+
+        // We can be cheeky and encode the size at bit 8 to save a parameter
+        const auto Control = Op->Control | (uint16_t(Is64Bit) << 8);
+
+        // Handle last three parameters first to avoid stepping on SRA registers
+        // (R4, R5, and R6 can be allocated to SrcRAX and SrcRDX, so we don't want)
+        // to obliterate any data prematurely.
+        mov(TMP1, SrcRAX.X());
```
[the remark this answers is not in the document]
lioncash: Should be all good now

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -4552,4 +4552,80 @@ void OpDispatchBuilder::VPERMILRegOp<4>(OpcodeArgs);
...
+
+  // Set all of the necessary flags.
+  // We use the top 16-bits of the result to store the flags
+  // in the form:
+  //
+  // Bit:  19   18   17   16
+  //      [OF | CF | SF | ZF]
+  //
+  const auto GetFlagBit = [this, OneConst, ZeroConst, IntermediateResult](int BitIndex) {
+    return _Select(IR::COND_NEQ,
```
Sonicadvance1, line 4609: Should this entire operation just be a bfe?

Sonicadvance1 (approved):
Aside from the single nit, seems good! Fantastic documentation, very thorough! This is one whale of an instruction!

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
lioncash: Nice catch! Amended

PR #2635, opened 2023-04-18
OpcodeDispatcher: Handle VCVTSD2SS/VCVTSS2SD

lioncash:
Gets the scalar variants out of the way, and should allow games listed in #2634 to progress further


Sonicadvance1 (approved):
Nice nice nice

PR #2643, opened 2023-04-24
OpcodeDispatcher: Handle VPMASKMOVD/VPMASKMOVQ

lioncash:
We can reuse the same helper we have for handling VMASKMOVPD and VMASKMOVPS, though we need to move some handling around to account for the fact that VPMASKMOVD and VPMASKMOVQ 'hijack' the REX.W bit to signify the element size of the operation.

Now all that remains for AVX/AVX2 is the remaining three SSE4.2 string instructions and the AVX2 gather instructions


Sonicadvance1 (approved):
\o/

PR #2646, opened 2023-04-26
SignalDelegator: Calculate siginfo layout

Sonicadvance1:
There are a handful of siginfo layout types. To determine which layout to use we need to check a combination of si_code and signal number because each one in isolation doesn't explain the layout type.

Once the layout is calculated then calculate the siginfo using that layout type. Adds a couple of different layout types to our guest siginfo_t to handle the previously missing types.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
@@ -90,6 +90,62 @@ namespace FEX::HLE {
    * @name Signal frame setup
    * @{ */
 
+  // Total number of layouts that siginfo supports.
+  enum class SigInfoLayout {
+    LAYOUT_KILL,
+    LAYOUT_TIMER,
+    LAYOUT_POLL,
+    LAYOUT_FAULT,
+    LAYOUT_FAULT_RIP,
+    LAYOUT_CHLD,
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -90,6 +90,63 @@ namespace FEX::HLE {
...
+  // Calculate the siginfo layout based on Signal and si_code.
+  static SigInfoLayout CalculateSigInfoLayout(int Signal, int si_code) {
+    SigInfoLayout Layout = SigInfoLayout::LAYOUT_KILL;
+    if (si_code > SI_USER && si_code < SI_KERNEL) {
+      // For signals that are not considered RT.
+      if (Signal == SIGSEGV ||
+          Signal == SIGBUS ||
+          Signal == SIGTRAP) {
+        // Regular FAULT layout.
+        Layout = SigInfoLayout::LAYOUT_FAULT;
```
lioncash, line 114: Wouldn't these also be able to be returns like in the else case below?

Sonicadvance1: oop, you're right. Fixed.

lioncash (approved):

PR #2647, opened 2023-04-30
SignalDelegator: Make sure to save and restore `InSyscallInfo`

### LINKED ISSUE #2560: VC redistributables still crash loop
For some reason these still crash loop depending on the exact version. Nailing down why might be a bit difficult.

Sonicadvance1:
Fixes #2560

This was a forgotten member of the context that needs to be saved and restored when jumping around the signal state.
When FEX was receiving signals back to back, there was a chance that the signals would ride the edge of having set `InSyscallInfo` in the JIT, which meant the SRA state would get saved once, then another signal would occur with the previous SRA data, saving SRA again. Then when unwinding the frames it would corrupt the SRA registers. This would result in trying to load SRA state that was no longer valid, looking like a crash in the JIT that was hard to see what happens.

Should also make Mono games a little less crash happy.

Cleans up CPUState memcpy as well, since it was a little weird looking.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/ArchHelpers/MContext.h
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/ArchHelpers/MContext.h
@@ -38,6 +38,7 @@ struct X86ContextBackup {
   uint64_t GPRs[23];
   FEXCore::x86_64::_libc_fpstate FPRState;
   uint64_t sa_mask;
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Need to stability test this before pushing it in. Don't want to break stuff

Sonicadvance1:
Stress tested this and it looks fine.

lioncash (approved):

PR #2649, opened 2023-05-02
OpcodeDispatcher: Handle PCMPISTRI/VPCMPISTRI

lioncash:
Handles the implicit variant of the string-compare-returning-index instructions. Now all that remains is the masked variants.


Sonicadvance1 (approved):
Wow those unit tests!

PR #2653, opened 2023-05-03
Win32 memory allocation fixes

julliard:
This lets me run at least a few instructions of Win32 code under Wine.


Sonicadvance1 (approved):
Oops! Thanks for the fixes

PR #2654, opened 2023-05-04
FEXCore: Moves SIGBUS handler to FEXCore/Utils

Sonicadvance1:
This can be done in an OS agnostic fashion. FEXCore knows the details of
its JIT and should be done in FEXCore itself.

The frontend is only necessary to inform FEXCore where the fault occured
and provide the array of GPRs for accessing and modifying the signal
state.

This is necessary for supporting both Linux and Wine signal contexts
with their unaligned access handlers.


Sonicadvance1:
Converting to a draft temporarily.
Going to move this back to FEXCore as a utility instead of ToolsCommon. Since FEXCore should be what understands how to handle unaligned accesses, while the frontend just informs where they happened.

Sonicadvance1:
Moved the unaligned handler back to FEXCore while the frontend contains on the OS specific information still. Nice clean separation now.

lioncash (approved):

PR #2662, opened 2023-05-05
CPUID: Disable RDTSCP under wine

Sonicadvance1:
We don't have a sane way to query cpu index under wine. We could technically still use the syscall since we know that we are still executing under Linux, but that seems a bit terrible.

Disable for now until something can be worked out. Not like it is used heavily anyway.


External/FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -872,6 +872,12 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0000h(uint32_t Leaf) {
 
 // Extended processor and feature bits
 FEXCore::CPUID::FunctionResults CPUIDEmu::Function_8000_0001h(uint32_t Leaf) {
+
+#ifndef _WIN32
```
lioncash, line 876: Perhaps add a comment above this with the rationale

Sonicadvance1: Added a comment

lioncash (approved):

PR #2665, opened 2023-05-09
FEXCore: Adds support for hardware x86-TSO prctl

Sonicadvance1:
From https://github.com/AsahiLinux/linux/commits/bits/220-tso

This fails gracefully in the case the upstream kernel doesn't support this feature, so can go in early.

This feature allows FEX to use hardware's TSO emulation capability to reduce emulation overhead from our atomic/lrcpc implementation. In the case that the TSO emulation feature is enabled in FEX, we will check if the hardware supports this feature and then enable it.

If the hardware feature is supported it will then use regular memory accesses with the expectation that these are x86-TSO in strength.

The only hardware that anyone cares about that supports this is Apple's M class SoCs. Theoretically NVIDIA Denver/Carmel supports sequentially consistent, which isn't quite the same thing. I haven't cared to check if multithreaded SC has as strong of guarantees. But also since Carmel/Denver hardware is fairly rare, it's hard to care about for our use case.

### DIFF
--- a/External/FEXCore/Source/Interface/Context/Context.h
+++ b/External/FEXCore/Source/Interface/Context/Context.h
@@ -372,10 +372,27 @@ namespace FEXCore::Context {
     FEXCore::Utils::PooledAllocatorVirtual OpDispatcherAllocator;
     FEXCore::Utils::PooledAllocatorVirtual FrontendAllocator;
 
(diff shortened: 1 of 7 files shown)


lioncash (approved):
LGTM

PR #2666, opened 2023-05-09
Wine TestHarnessRunner support

Sonicadvance1:
This gets the test harness runner working with wine. 

Some things of note.
- This requires AArch64 built wine
- This requires a very new llvm-mingw to build FEXCore and TestHarnessRunner
- Interpreter is untested and probably has some issues
- x87 soft float implementation has even more inaccurate emulation since transcendental operations use 64-bit double instead of 128-bit long double.
   - The ASM tests that fail are because of this precision loss
   - `99% tests passed, 27 tests failed out of 5166` otherwise.
- This isn't wired up to CI yet because there is no builder that has AArch64 wine installed
   - This will be forthcoming. 
- Syscalls aren't wired up for wine at all because that's still an open question for now.


Source/Tools/FEXLoader/CMakeLists.txt
```diff
@@ -1,148 +1,156 @@
...
+        PRIVATE
+          "LINKER:--gc-sections"
+          "LINKER:--strip-all"
+          "LINKER:--as-needed"
       )
+    endif()
 
-      add_dependencies(uninstall uninstall_binfmt_misc_32)
-      add_dependencies(uninstall uninstall_binfmt_misc_64)
+    if (NOT MINGW_BUILD)
```
neobrain, line 43: This is already checked in the outer scope

Source/Tools/FEXLoader/TestHarnessRunner.cpp
```diff
@@ -168,7 +255,15 @@ int main(int argc, char **argv, char **const envp) {
 
   CTX->InitializeContext();
 
+#ifndef _WIN32
   auto SignalDelegation = FEX::HLE::CreateSignalDelegator(CTX.get());
+#else
+  // Enable exit on HLT while Wine's longjump is broken.
+  //
+  // Once they fix longjump, we can remove this.
```
neobrain, line 263: Is there any reference to an upstream bug or similar to this? As is, the "Once they fix longjmp, we can remove this" comment isn't very useful since there's no way to tell if it's fixed unless you already knew about it.

Source/Tools/FEXLoader/TestHarnessRunner.cpp
```diff
@@ -183,32 +278,29 @@ int main(int argc, char **argv, char **const envp) {
     (!HostFeatures.SupportsCLZERO && Loader.RequiresCLZERO()) ||
     (!HostFeatures.SupportsBMI1 && Loader.RequiresBMI1()) ||
     (!HostFeatures.SupportsBMI2 && Loader.RequiresBMI2()) ||
-    (!HostFeatures.SupportsCLWB && Loader.RequiresCLWB());
+    (!HostFeatures.SupportsCLWB && Loader.RequiresCLWB()) ||
+#ifdef _WIN32
+    (Loader.RequiresLinux()) ||
+#endif
+    false;
```
neobrain, line 285:
```suggestion
    (!HostFeatures.SupportsCLWB && Loader.RequiresCLWB());
#ifdef _WIN32
    TestUnsupported |= Loader.RequiresLinux();
#endif
```

Source/Tools/FEXLoader/CMakeLists.txt
Sonicadvance1: Resolved.

Source/Tools/FEXLoader/TestHarnessRunner.cpp
Sonicadvance1: Resolved.

Source/Tools/FEXLoader/TestHarnessRunner.cpp
Sonicadvance1: Nothing currently tracked in upstream wine about this issue.

lioncash (approved):

PR #2674, opened 2023-05-13
unittests: Adds step to remove stale SHM regions.

Sonicadvance1:
Some of the unit tests we run will leak shm regions. Presumably this is because they never called `shmctl(IPC_RMID)` so the ID is leaked forever.

This can be seen by querying `/proc/sysvipc/shm` to see a list of old shm regions that eventually hit the maximum capacity of 4096 shm ids.

Once CI is done running, run a utility application that all it does is check for SHM IDs that have zero attachments (thus unused), it was created by the UID of the runner, and it is older than ten minutes. At which point it will erase it.

This will fix spurious failures in our CI caused by running out of SHM IDs, previously I had a cron job setup to restart the CI runners every hour or so which caused its own spurious failure problems.

FINALLY this bug was triaged which has been annoying us for...years?

### DIFF
--- a/unittests/Utilities/DeleteOldSHMRegions.cpp
+++ b/unittests/Utilities/DeleteOldSHMRegions.cpp
@@ -1,0 +1,42 @@
+#include <fstream>
+#include <stdio.h>
+#include <sys/shm.h>
+#include <unistd.h>
+
+static bool ctime_is_old(uint64_t ctime) {
+  time_t curtime;
+  time(&curtime);
+
+  // If it is older than ten minutes.
+  return (curtime - ctime) > 600;
+}
+
+int main() {
+  // key      shmid perms                  size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime                   rss                  swap
+  //	 0     360448   777                  4096 165187 165187      0  1002  1002  1002  1002 1679659857 1679659857 1679659857                  4096                     0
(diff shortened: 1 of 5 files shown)


unittests/Utilities/DeleteOldSHMRegions.cpp
```diff
@@ -0,0 +1,43 @@
...
+
+  // If it is older than ten minutes.
+  return (curtime - ctime) > 600;
+}
+
+int main() {
+  // key      shmid perms                  size  cpid  lpid nattch   uid   gid  cuid  cgid      atime      dtime      ctime                   rss                  swap
+  //	 0     360448   777                  4096 165187 165187      0  1002  1002  1002  1002 1679659857 1679659857 1679659857                  4096                     0
+  //	 0      32769   777                  4096 153814 153814      0  1002  1002  1002  1002 1676490841 1676490841 1676490841                     0                  4096
+  std::fstream fs;
```
lioncash, line 18: I think this can be shortened to:

```suggestion
  std::ifstream fs{"/proc/sysvipc/shm", std::fstream::binary};
```

lioncash (commented):
One tiny nit, but woo to getting rid of this annoyance!

unittests/Utilities/DeleteOldSHMRegions.cpp
Sonicadvance1: Resolved.

lioncash (approved):

PR #2678, opened 2023-05-17
OpcodeDispatcher: Handle PCMPESTRM/VPCMPESTRM

lioncash:
...and with that all of the SSE4.2 string instructions are implemented now


Sonicadvance1 (approved):
\o/

PR #2680, opened 2023-05-19
Syscalls: Optimize getdents{64,}

Sonicadvance1:
I originally wrote this emulation prior to me fully understanding how the syscall works. So there are two optimizations here.

1) No need to consume the incoming buffer at all.
   - Originally I thought the incoming dirent structures were used to calculate offset.
   - This is not the case, the FD's file position is used instead.
   - This means we can remove the incoming buffer consuming overhead entirely. 2) No need to allocate a temporary buffer at all.
   - With getdents and getdents64 we are guaranteed to be dealing with structures that are the same size or smaller than the host structure.
   - This lets us encode the real host dirents in to the provided buffer.
   - After the `getdents64` host syscall, we then iterate forward through the list, modifying as we go.
   - Need to make sure to shift the elements of the structure in order.
   - Need to make sure to use memmove on the `d_name` member since the movement region can overlap.

These two optimizations significantly reduce the amount of time spent in getdents, which has a noticeable impact on load times.

Side-tangent: I noticed a fun quirk of how NFS operates with getdents. If the FSCache hasn't populated the metadata for that folder, then it will early return with "some" data, not fully maxing out the buffer. The kernel will start prefetching metadata assuming directory iterating is happening. The next `getdents` happens and it should return a larger number of elements.

Very neat.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
@@ -61,56 +61,21 @@ namespace FEX::HLE {
 class SignalDelegator;
 SyscallHandler *_SyscallHandler{};
 
-
 template<bool IncrementOffset, typename T>
 uint64_t GetDentsEmulation(int fd, T *dirp, uint32_t count) {
-  fextl::vector<uint8_t> TmpVector(count);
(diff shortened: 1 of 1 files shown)


lioncash (approved):
Love it when optimizations consist entirely of removals

PR #2686, opened 2023-05-22
FEXCore: Implements support for xgetbv

Sonicadvance1:
This returns the `XFEATURE_ENABLED_MASK` register which reports what features are enabled on the CPU.
This behaves similarly to CPUID where it uses an index register in ecx.

This is a prerequisite to enabling XSAVE/XRSTOR and AVX since applications will expect this to exist.

xsetbv is a privileged instruction and doesn't need to be implemented.


Sonicadvance1:
Implementing this I actually found a pretty nasty bug in `SpillStaticRegs` where there were two conflicting assumptions made. Probably had some weird edge case bugs that are now fixed.

lioncash (approved):
Nice catch on the spill bug

PR #2687, opened 2023-05-22
Telemetry: Save on signal terminate

Sonicadvance1:
When a signal handler is not installed and is a terminal failure, make sure to save telemetry before faulting.

We know when an application is going down in this case so we can make sure to have the telemetry data saved.

Adds a telemetry signal mask data point as well to know which signal took it down.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
@@ -1440,6 +1440,15 @@ namespace FEX::HLE {
     else if (Handler.OldAction.handler == SIG_DFL &&
       (Handler.DefaultBehaviour == DEFAULT_COREDUMP ||
        Handler.DefaultBehaviour == DEFAULT_TERM)) {
+
+      // In the case of signals that cause coredump or terminate, save telemetry early.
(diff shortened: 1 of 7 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -2020,7 +2030,7 @@ namespace FEX::HLE {
     return Result == -1 ? -errno : Result;
   }
 
-  fextl::unique_ptr<FEX::HLE::SignalDelegator> CreateSignalDelegator(FEXCore::Context::Context *CTX) {
-    return fextl::make_unique<FEX::HLE::SignalDelegator>(CTX);
+  fextl::unique_ptr<FEX::HLE::SignalDelegator> CreateSignalDelegator(FEXCore::Context::Context *CTX, fextl::string const *ApplicationName) {
```
neobrain, line 2033: `Program.ProgramPath` doesn't change after SignalDelegator is created, is it? We should just pass the string in by copy instead then, to avoid this quasi-global and the lifetime dependencies it implicitly builds.

Sonicadvance1: Converted over to a copy.

PR #2689, opened 2023-05-27
CPUID: Only enable BMI1 and BMI2 if AVX is supported

Sonicadvance1:
These two extensions rely on AVX being supported to be used. Primarily because they are VEX encoded.

GTA5 is using these flags to determine if it should enable its AVX support.


alyssarosenzweig:
R-b

alyssarosenzweig (approved):

lioncash (approved):

PR #2690, opened 2023-05-27
Linux: Make vfork act more similar to how it should.

Sonicadvance1:
Noticed this while debugging Proton Experimental hanging and thought this could be related. Didn't solve that issue but this should be merged anyway.

vfork doesn't fork the host's process space in to the child process. Saving Copy-On-Write overhead problems. It also puts the parent process to sleep until the fork terminates or executes.

This is a major issue under FEX where we can't emulate vfork correctly because we need to do other work before this process terminates or executes a new process. We have been treating `vfork` as a `fork` this entire time.

This can likely cause problems for applications that actually use vfork to wait for a process to complete. So let's actually emulate that feature by using a pipe with poll to determine when that FD gets removed.

FEX can't use waitpid to wait for this process to terminate since we would affect the guest also wanted to use a waitpid.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
@@ -26,6 +26,7 @@ $end_info$
 #include <stdint.h>
 #include <sched.h>
 #include <sys/personality.h>
+#include <sys/poll.h>
 #include <sys/prctl.h>
 #include <sys/resource.h>
 #include <sys/syscall.h>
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -187,11 +188,18 @@ namespace FEX::HLE {
...
+
+    const bool IsVFork = flags & CLONE_VFORK;
     pid_t Result{};
-    if (flags & CLONE_VFORK) {
-      // XXX: We don't currently support a vfork as it causes problems.
-      // Currently behaves like a fork, which isn't correct. Need to find where the problem is
+    int VForkFDs[2];
+    if (IsVFork) {
+      // Use pipes as a mechanism for knowing when the child process is exiting.
+      // FEX can't use `waitpid` for this since the child process may want to use it.
```
neobrain, line 197: Can't we use `WNOWAIT` to handle this?

Sonicadvance1: Sadly no.
```
wait4(488421, 0xffff94ece7a4, WNOWAIT, NULL) = -1 EINVAL (Invalid argument)
```
(wait4 implements waitpid on AArch64).
No way to work around this as far as I can tell.

Sonicadvance1: Oh, actually. `waitid(P_PID, pid, &Info, WNOWAIT | WEXITED);` Gets close. Except it doesn't return until exit.
We need to block until process exit OR execve. This would wait until the execve application exits which isn't the same semantics.

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -248,6 +256,12 @@ namespace FEX::HLE {
         Thread->ThreadManager.clear_child_tid = child_tid;
       }
 
+      if (IsVFork) {
+        // Close the read end of the pipe.
+        // Keep the write end open so the parent can poll it.
+        close(VForkFDs[0]);
```
neobrain, line 262: Can we move the initial `close`s of both the child and the parent to the `fork`? It's a bit randomly placed here, and moving it closer to the fork makes it easier to quickly see which of the FDs is still valid in each branch of execution.

Sonicadvance1: Moved.

neobrain (approved):
Bummer that `waitpid` doesn't cover the execve case. The FD approach looks like a reasonable fallback!

PR #2691, opened 2023-05-27
ScopedSignalMask: Clean up API and use std::unique_lock/shared_lock

neobrain:
This replaces the poorly designed `ScopedSignalMaskWith{Mutex,SharedLock,UniqueLock}` abstractions (and the similar ones added in #2739). Instead, three separate scope guards are used (one to mask POSIX signals, one `std::unique_lock`/`std::shared_lock` to acquire the mutex, and one to manage the signal deferral refcount) and combined with helper functions specific to each use case.

This new design has better separation of concerns and is easier to adapt for other use cases. For example, #2493 had to add two full copies of `ScopedSignalMask*` with small variations, but now it can re-use the same code.

Further, this PR changes the WIN32 implementation of `ForkableUniqueMutex`/`ForkableSharedMutex` to just extend `std::mutex`/`std::shared_mutex`. This makes the interface more powerful than the POSIX-implementation, but it's more of a placeholder anyway.


FEXHeaderUtils/FEXHeaderUtils/ScopedSignalMask.h
```diff
@@ -1,109 +1,61 @@
...
+   * - The signal handler itself using a mutex that would be re-locked if the handler gets invoked
+   *   again before unlocking
+   *
+   * Ownership of the returned object may be moved, but it is NOT SAFE to move across threads.
+   */
+  template<template<typename> class LockType = std::unique_lock, typename MutexType>
+  [[nodiscard]] static auto MaskSignalsAndLockMutex(MutexType& mutex, uint64_t Mask = ~0ULL) {
+#ifndef _WIN32
+      // Signals are masked first, and then the lock is acquired
+      return std::pair { ScopedSignalMasker { Mask }, LockType<MutexType> { mutex } };
```
neobrain, line 55: As found during development of #2739, `std::pair` has unspecified order of destruction and hence is not safe to use here.

neobrain:
Updated for the definitions recently added via `DeferredSignalMutex.h` and fixed the `std::pair` bug. Now saving more than 200 lines of implementation!

Sonicadvance1:
Looks like ScopedSignalMasker is still double unmasking.
```cpp
#include <FEXCore/Utils/MathUtils.h>
#include <FEXCore/Utils/SignalScopeGuards.h>
#include <catch2/catch.hpp>

TEST_CASE("Mutex") {
  FEXCore::ForkableUniqueMutex mutex{};
  {
    auto lk = FEXCore::GuardSignalDeferringSectionWithFallback(mutex, nullptr);
    printf("In Lock\n");
  }
  printf("After lock\n");
}
```

```diff
diff --git a/External/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h b/External/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
index 90b5c4a38..065841af3 100644
--- a/External/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
+++ b/External/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
@@ -115,6 +115,7 @@ namespace FEXCore {
     public:
       explicit DeferredSignalRefCountGuard(FEXCore::Core::InternalThreadState *Thread) : Thread(Thread) {
         // Needs to be atomic so that operations can't end up getting reordered around this.
+        printf("%s Increment\n", __func__);
         Thread->CurrentFrame->State.DeferredSignalRefCount.Increment(1);
       }

@@ -126,6 +127,7 @@ namespace FEXCore {
       }

       ~DeferredSignalRefCountGuard() {
+        printf("%s Decrement\n", __func__);
         if (Thread) {
 #ifdef _M_X86_64
           // Needs to be atomic so that operations can't end up getting reordered around this.
@@ -153,6 +155,7 @@ namespace FEXCore {
     public:
       explicit ScopedSignalMasker(uint64_t Mask) : OriginalMask(0) {
         // Mask all signals, storing the original incoming mask
+        printf("%s Mask\n", __func__);
         ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &Mask, &*OriginalMask, sizeof(*OriginalMask));
       }

@@ -163,6 +166,7 @@ namespace FEXCore {

       ~ScopedSignalMasker() {
         if (OriginalMask) {
+          printf("%s Unmask\n", __func__);
           ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &OriginalMask, nullptr, sizeof(*OriginalMask));
         }
       }
```

```bash
write(1, "ScopedSignalMasker Mask\n", 24ScopedSignalMasker Mask
) = 24
rt_sigprocmask(SIG_SETMASK, ~[], [], 8) = 0
write(1, "~ScopedSignalMasker Unmask\n", 27~ScopedSignalMasker Unmask
) = 27
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
write(1, "In Lock\n", 8In Lock
)                = 8
write(1, "~ScopedSignalMasker Unmask\n", 27~ScopedSignalMasker Unmask
) = 27
rt_sigprocmask(SIG_SETMASK, [], NULL, 8) = 0
write(1, "After lock\n", 11After lock
)            = 11
```

neobrain:
> Looks like ScopedSignalMasker is still double unmasking.

Ah, thanks. DeferredSignalRefCountGuard was fine but ScopedSignalMasker indeed didn't properly reset the `optional` member. Should be fixed now!

Sonicadvance1:
Is there any way to make the code generation around this optional usage not be terrible?
Image_2023-07-25_06-13-25

neobrain:
> Is there any way to make the code generation around this optional usage not be terrible?

Is that debug or optimized assembly?

My understanding is that ScopedPotentialDeferredSignalWithForkableMutex is just a temporary workaround used to allow some external code to pass in nullptr `Thread`s, so I didn't see reason to worry about a few extra instructions.

We could use an `std::variant` instead, which would avoid some of the branches. Try replacing the declarations with `std::variant<DeferredSignalRefCountGuard, ScopedSignalMasker>`? (The `.emplace` calls my need to be replaced with `.emplace<0>(Thread)` and `.emplace<1>(Mask)` then.)

Sonicadvance1:
> > Is there any way to make the code generation around this optional usage not be terrible?
> 
> Is that debug or optimized assembly?
> 
> My understanding is that ScopedPotentialDeferredSignalWithForkableMutex is just a temporary workaround used to allow some external code to pass in nullptr `Thread`s, so I didn't see reason to worry about a few extra instructions.
> 
> We could use an `std::variant` instead, which would avoid some of the branches. Try replacing the declarations with `std::variant<DeferredSignalRefCountGuard, ScopedSignalMasker>`? (The `.emplace` calls my need to be replaced with `.emplace<0>(Thread)` and `.emplace<1>(Mask)` then.)

This is a `RelWithDebInfo` build on both.

neobrain:
Switching `GuardSignalDeferringSectionWithFallback` to `std::variant` got rid of most of the branches. In your `Thread=nullptr` example, this actually produces better code than previously. However in practice `Thread` is never a constant, so it's more useful to inspect this code instead:

```cpp
#include <mutex>
#include <FEXCore/Utils/SignalScopeGuards.h>

extern "C" void RunTheTestCase_MaskSignalsAndLockMutex(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  auto lk = FEXCore::MaskSignalsAndLockMutex(mutex, 0xdeadbeef);
}

extern "C" void RunTheTestCaseLegacy_MaskSignalsAndLockMutex(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  FEXCore::ScopedSignalMaskWithMutexBase<FEXCore::ForkableUniqueMutex, &FEXCore::ForkableUniqueMutex::lock, &FEXCore::ForkableUniqueMutex::unlock> lk(mutex, 0xdeadbeef);
}

extern "C" void RunTheTestCase_X_GuardSignalDeferringSection(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  auto lk = FEXCore::GuardSignalDeferringSection(mutex, Thread);
}

extern "C" void RunTheTestCaseLegacy_X_GuardSignalDeferringSection(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  FEXCore::ScopedDeferredSignalWithMutexBase<FEXCore::ForkableUniqueMutex, &FEXCore::ForkableUniqueMutex::lock, &FEXCore::ForkableUniqueMutex::unlock> lk(mutex, Thread);
}

extern "C" void RunTheTestCase_GuardSignalDeferringSectionWithFallback(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  auto lk = FEXCore::GuardSignalDeferringSectionWithFallback(mutex, Thread);
}

extern "C" void RunTheTestCaseLegacy_GuardSignalDeferringSectionWithFallback(FEXCore::ForkableUniqueMutex& mutex, FEXCore::Core::InternalThreadState *Thread) {
  FEXCore::ScopedPotentialDeferredSignalWithForkableMutex lk(mutex, Thread);
}
```

Comparing each of the three helper functions:
MaskSignalsAndLockMutex: 3 (old) ↔ 5 blocks (new)
GuardSignalDeferringSection: 21 instructions both (equivalent asm)
GuardSignalDeferringSectionWithFallback: 7 (old) ↔ 8 (new) blocks

GuardSignalDeferringSectionWithFallback was an outlier previously and is much improved now. The remaining overhead seem within reason and shouldn't fall into weigh overall.

Sonicadvance1 (approved):
Thanks for fixing the codegen.

PR #2700, opened 2023-06-04
IR: Fixes major performance regression

Sonicadvance1:
When #2170 was merged I didn't see any perceived performance regression but this turned out to be untrue. When register accesses were switched from {Load,Store}Context to {Load,Store}Register the DeadStoreElimination and RCLSE passes weren't updated to understand the IR ops.

Since the SRA pass used to run after these passes, they didn't need to be aware of these IR operations. Since we moved the usage of the IR ops up in to the OpcodeDispatcher these now flooded the IR passes by replacing a significant portion of the operations.

The fix to this is to treat LoadRegister as LoadContext and StoreRegister as StoreContext in both of these IR passes. So a significant portion of the change is just shuffling the code to operate on both IR operations.

A minor change to RCLSE where {Load,Store}ContextIndexed is now guaranteed to /never/ touch GPRs and XMM registers, so classification can see through these IR operations now. Adjusted to not clear GPR and XMM classifications when encountering those operations.

The platform that is most affected by this change was x86-64 since it doesn't support SRA so this would turn in to hundreds of thousands of new memory accesses that weren't getting deleted. Mostly relying on store-load forwarding to save that hardware.

Stats:
This was taken from running Hollow Knight to the title screen, which is where I noticed the performance regression.

* RCLSE pass
   * Pre-optimization: Removed 0 redundant stores
   * Post-optimization: Removed 1725887 redundant stores
   * Untested: redundant loads also get removed here. Harder to see stats for removed loads.

* Dead store elimination pass
   * Pre-optimization: Removed 0 redundant stores
   * Post-optimization: Removed 116 redundant stores
   * Turns out RCLSE removes most of these already.

This doesn't fully restore the lost performance on x86-64 platforms, but it is a significant uplift. More PRs need to be done to see where the remaining perf lost went.

So known regression since FEX-2212 was released effectively.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/DeadContextStoreElimination.cpp
@@ -366,7 +366,7 @@ namespace {
       ContextClassificationInfo->Lookup.size(), sizeof(FEXCore::Core::CPUState));
   }
 
-  static void ResetClassificationAccesses(ContextInfo *ContextClassificationInfo, bool SupportsAVX) {
+  static void ResetClassificationAccesses(ContextInfo *ContextClassificationInfo, bool SupportsAVX, bool SkipGPRsAndXMM = false) {
     auto ContextClassification = &ContextClassificationInfo->ClassificationInfo;
 
     auto SetAccess = [&](size_t Offset, auto Access) {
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
This won't make it for the 2306 release. Need to figure out why enabling the optimization makes Proton/Sonic Mania no longer boot. Probably have a bug elsewhere that interacts with this.

MastaG:
Just out of curiosity, were you able to fix the issue with Sonic Mania on Proton when running this PR?

Sonicadvance1:
Haven't gotten back to it yet so far.

PR #2701, opened 2023-06-07
OpcodeDispatcher: Optimize EFLAG unpacking

Sonicadvance1:
Noticed this was slightly unoptimal. Resulting in a 18% code reduction in the case of of a simple four instruction test ASM case.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -1544,7 +1544,7 @@ void OpDispatchBuilder::SAHFOp(OpcodeArgs) {
   OrderedNode *Src = LoadGPRRegister(X86State::REG_RAX, 1, 8);
 
   // Clear bits that aren't supposed to be set
-  Src = _And(Src, _Constant(~0b101000));
+  Src = _Andn(Src, _Constant(0b101000));
 
   // Set the bit that is always set here
   Src = _Or(Src, _Constant(0b10));
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
@@ -52,10 +52,9 @@ void OpDispatchBuilder::SetPackedRFLAG(bool Lower8, OrderedNode *Src) {
     InvalidateDeferredFlags();
   }
 
-  auto OneConst = _Constant(1);
   for (size_t i = 0; i < NumFlags; ++i) {
     const auto FlagOffset = FlagOffsets[i];
-    auto Tmp = _And(_Lshr(Src, _Constant(FlagOffset)), OneConst);
+    auto Tmp = _Bfe(4, 1, FlagOffset, Src);
     SetRFLAG(Tmp, FlagOffset);
   }
 }


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1544,7 +1544,7 @@ void OpDispatchBuilder::SAHFOp(OpcodeArgs) {
   OrderedNode *Src = LoadGPRRegister(X86State::REG_RAX, 1, 8);
 
   // Clear bits that aren't supposed to be set
-  Src = _And(Src, _Constant(~0b101000));
+  Src = _Andn(Src, _Constant(0b101000));
```
alyssarosenzweig, line 1547: Is there a good reason why the optimizer doesn't do this one? Would it help in cases where the guest code does masking itself?

Obviously this makes sense in addition, for compile-time reasons, but for the code reduction it smells funny

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -52,10 +52,9 @@ void OpDispatchBuilder::SetPackedRFLAG(bool Lower8, OrderedNode *Src) {
     InvalidateDeferredFlags();
   }
 
-  auto OneConst = _Constant(1);
   for (size_t i = 0; i < NumFlags; ++i) {
     const auto FlagOffset = FlagOffsets[i];
-    auto Tmp = _And(_Lshr(Src, _Constant(FlagOffset)), OneConst);
+    auto Tmp = _Bfe(4, 1, FlagOffset, Src);
```
alyssarosenzweig, line 57: This one looks good, assuming the guest's compiler can be trusted to use BEXTR or something.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: This line actually doesn't do any code reduction by itself, the real optimization is in the `SetPackedRFLAG` code. This one is more about signaling intent that we want to clear some bits. ConstProp pass should theoretically in the future see if the immediate fits in the encoding space and does a `mov + bic` or `and <reg>, #<negated constant>`. This isn't the case today sadly.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
Sonicadvance1: This IR goes through our backend, so usually 1:1 relates to instructions (since the IR optimization are light).
So this causes a `lsr + and + ubfx + strb` per flag to convert in to a `ubfx + strb`. Cutting the number of uops in half and the number of data dependencies in to far less.

alyssarosenzweig: right, makes sense :+1:

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
alyssarosenzweig: Ah, I see. If it's just intent then, makes sense to me :+1: Thanks for explaining

alyssarosenzweig (approved):

lioncash (approved):

PR #2703, opened 2023-06-07
OpcodeDispatcher: Optimize ADC/ADD OF flag calculation

Sonicadvance1:
`eor <reg>, <reg>, #-1` can't be encoded as an instruction. Instead use
mvn which does the same thing.

Removes a single instruction from each OF calculation for ADC and ADD.

Also no reason to use a switch statement for the source size, just use
_Bfe and calculate the offset based on operation size.

SBB caught in the crossfire to ensure it also isn't using a switch
statement.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
@@ -303,28 +303,11 @@ void OpDispatchBuilder::CalculcateFlags_ADC(uint8_t SrcSize, OrderedNode *Res, O
   // OF
   // Signed
   {
-    auto NegOne = _Constant(~0ULL);
-    auto XorOp1 = _Xor(_Xor(Src1, Src2), NegOne);
+    auto XorOp1 = _Not(_Xor(Src1, Src2));
     auto XorOp2 = _Xor(Res, Src1);
     OrderedNode *AndOp1 = _And(XorOp1, XorOp2);
 
-    switch (Size) {
-    case 8:
-      AndOp1 = _Bfe(1, 7, AndOp1);
-    break;
-    case 16:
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/MoveOps.cpp
```diff
@@ -73,11 +73,29 @@ DEF_OP(CreateElementPair) {
...
+  switch (OpSize) {
+  case 4:
+    mov(GetDst<RA_32>(Node), GetSrc<RA_32>(Op->Src.ID()));
+    not_(GetDst<RA_32>(Node));
+  break;
+  case 8:
+    mov(GetDst<RA_64>(Node), GetSrc<RA_64>(Op->Src.ID()));
+    not_(GetDst<RA_64>(Node));
+  break;
+  default: LOGMAN_MSG_A_FMT("Unknown UMul size: {}", OpSize);
```
lioncash, line 89:
```suggestion
  default: LOGMAN_MSG_A_FMT("Unknown MVN size: {}", OpSize);
```

Sonicadvance1: Resolved.

alyssarosenzweig:
Why `_MVN` instead of `_Not`? Surely `_Not(_Xor(a, b))` is clearer than `_MVN(_Xor(a, b))`?

Sonicadvance1:
> Why `_MVN` instead of `_Not`? Surely `_Not(_Xor(a, b))` is clearer than `_MVN(_Xor(a, b))`?

You're right actually. We already had a Not IR op even. Although we usually try to design our IR to match AArch64 semantics. And since AArch64 doesn't have a `Not` I thought we had missed it and created MVN.

Dropping first commit and switching it over to Not instead.

PR #2713, opened 2023-06-12
VectorFallbacks: Fix PCMPSTR fallback ZF/SF flag setting 

lioncash:
So, uh, this was a little silly to track down. Having the upper limit as unsigned was a mistake, since this would cause negative valid lengths to convert into an unsigned value (woo sign conversions, gotta love 'em!) within the first two flag comparison cases. A -1 valid length can occur if one of the strings starts with a null character in a vector's first element. (It will be zero and we then subtract it to make the length zero-based).

Fixes this edge-case up and expands a test to check for this in the future


Sonicadvance1 (approved):
Nice!

PR #2717, opened 2023-06-13
OpcodeDispatcher: Handle XSAVE/XRSTOR

lioncash:
Knocks out the final two instructions needed for AVX 1 support


Sonicadvance1 (approved):
Whoa! Loads of stuff happening!
Should this now allow CPUID function 01h to return XSAVE and OSXSAVE bits to be enabled when AVX is supported?

lioncash:
Yup it should. Added a commit for that

PR #2719, opened 2023-06-14
OpcodeDispatcher: Ensure MXCSR is saved/restored with FXSAVE/FXRSTOR

lioncash:
Previously, the bits that we support in the MXCSR weren't being saved, which means that some opcode patterns may fail to restore the rounding mode properly.

e.g. FXSAVE, followed by FNINIT, followed by FXRSTOR wouldn't restore the rounding mode properly

This fixes that.


Sonicadvance1 (approved):
Nice catch!

PR #2728, opened 2023-06-22
FEXServer: Adds support for coredump file generation. 

Sonicadvance1:
- Works for both 32-bit and 64-bit guests.
- Only active if `FEX_COOPERATIVECOREDUMP=1` is set on the FEXInterpreter side
- Each time a coredump is generated, FEXServer will do some garbage collection of old coredumps.
   - By default if there is more than 512MB of coredumps or files that are older than 24hours old then FEXServer begins erasing files from oldest to newest.
   - All coredumps older than the age limit get erased
   - Coredumps will be erased oldest to newest until under the size threshold. 
- Some missing features
   - Float state might not be getting saved entirely
   - Only crashing thread's state is saved.
      - Need some investigation in to if deferred Signals now let FEX safely pause all running threads in the case of JIT crash
      - Theoretically possible but this interface hasn't been tested in a long time.

Needs some more testing with more robust applications, currently I have
only tested it with a simple SIGSEGV application for 32-bit and 64-bit.

ConfigOptions:
- `FEX_COREDUMP_FOLDER`:
   - By default FEXServer will save coredumps in to `$HOME/.fex-emu/coredump`
   - This allows overriding this folder
- `FEX_COREDUMP_MAXIMUM_THRESHOLD`:
   - By default FEXServer allows 512MB of coredump space usage.
   - This allows an override for the maximum size
- `FEX_COREDUMP_MAXIMUM_AGE`:
   - By default FEXServer will erase coredumps older than 24 hours
   - This allows overriding this time
- `FEX_COREDUMP_COMPRESS_ZSTD`:
   - By default FEXServer will try to compress the coredump using zstd
   - If zstd isn't found then it falls back to uncompressed.
   - This allows explicitly disabling compression.
   - Saves a significant amount of file space on coredumps.    

Needs #2432 merged before this can be merged.


Source/Tools/FEXServer/CoreFileWriter/CoreFileWriter.cpp
```diff
@@ -0,0 +1,865 @@
...
+      int Ret = waitpid(pid, &WaitStatus, 0);
+
+      if (Ret != -1 && WIFEXITED(WaitStatus) && WEXITSTATUS(WaitStatus) == 0) {
+        LogMan::Msg::IFmt("Wrote Coredump to {}", CompressedCoredump);
+        return true;
+      }
+
+    }
+    return false;
+  }
```
neobrain, line 784: Isn't all of this equivalent to `if (0 == system("zstd -q ...")) { ... }`?

Source/Tools/FEXServer/CoreFileWriter/CoreFileWriter.cpp
```diff
@@ -0,0 +1,865 @@
...
+
+  void CoreFileWriter::CleanupOldCoredumps() {
```
neobrain, line 786: Do we really need this? `/tmp` seems to be a more suitable location for these, and it gets automatically cleaned up by the system so we wouldn't need to bring our own system for that.

PR #2734, opened 2023-06-28
Emitter: Adds support for CSSC 

Sonicadvance1:
Not used currently but will be used in the future.

Needed to rebase vixl on upstream in the process so unittests understand cssc correctly. Which also removed a decent amount of patches from our downstream fork.

### DIFF
--- a/External/FEXCore/unittests/Emitter/ALU_Tests.cpp
+++ b/External/FEXCore/unittests/Emitter/ALU_Tests.cpp
@@ -782,23 +782,17 @@ TEST_CASE_METHOD(TestDisassembler, "Emitter: ALU: AddSub - shifted register") {
     TEST_SINGLE(add(Size::i64Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 63), "add x30, x29, x28, lsl #63");
     TEST_SINGLE(add(Size::i32Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 31), "add w30, w29, w28, lsl #31");
 
-    TEST_SINGLE(add(Size::i32Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 32), "unallocated (Unallocated)");
-
     // LSR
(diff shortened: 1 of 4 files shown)


dougallj:
Just in case it's unintentional, it looks like this is missing the 8-bit immediate forms of min/max.

External/FEXCore/unittests/Emitter/ALU_Tests.cpp
```diff
@@ -782,23 +782,26 @@ TEST_CASE_METHOD(TestDisassembler, "Emitter: ALU: AddSub - shifted register") {
     TEST_SINGLE(add(Size::i64Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 63), "add x30, x29, x28, lsl #63");
     TEST_SINGLE(add(Size::i32Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 31), "add w30, w29, w28, lsl #31");
 
-    TEST_SINGLE(add(Size::i32Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 32), "unallocated (Unallocated)");
+    // Vixl decodes this as a "valid" instruction, should be "unallocated (Unallocated)".
+    TEST_SINGLE(add(Size::i32Bit, Reg::r30, Reg::r29, Reg::r28, ShiftType::LSL, 32), "add w30, w29, w28, lsl #32");
```
dougallj, line 803: I'd consider removing these instead... I agree it's a Vixl bug, but I'm dubious of the premise of the original test ("if we put rubbish in, we want specific rubbish out"). If it's going to generate a bad instruction, I'd add an assert in the generating function instead?

Sonicadvance1: You're right. I added an assert to the instruction emitters and removed the tests.

Sonicadvance1:
> Just in case it's unintentional, it looks like this is missing the 8-bit immediate forms of min/max.

Yep, that was unintentional. Added the immediate encodings and unit tests.

Should be ready again.

dougallj (approved):

lioncash (approved):

PR #2737, opened 2023-07-01
Some small fixes for android building

bylaws:
Fixes a few issues I encountered when trying to build for android (non-termux, via NDK).  A successful build does currently require either using a beta NDK or pulling in a local libc++ through CMake though due to FEX's reliance on std::pmr.


Sonicadvance1 (approved):
Nice! Thanks for the patches.

PR #2739, opened 2023-07-02
Linux: Fixes hangs due to mutexes locked while fork happens.

### LINKED ISSUE #2664: AOTIR futexes causing hangs
Steam has a really high chance of hanging due to AOTIR futexes. Theoretically because these are being held while a fork is called?
Really frustrating since this interface is currently broken and unused anyway, so it getting in the way is a pain.
Needs to be fixed because it is likely the cause of hanging in other obscure situations.

Sonicadvance1:
When a fork occurs FEX needs to be incredibly careful as any thread
(that isn't forking) that holds a lock will vanish when the fork occurs.

At this point if the newly forked process tries to use these mutexes
then the process hangs indefinitely.

The three major mutexes that need to be held during a fork:
- Code Invalidation mutex
  - This is the highest priority and causes us to hang frequently.
  - This is highly likely to occur when one thread is loading shared
    libraries and another thread is forking.
     - Happens frequently with Wine and steam.
- VMA tracking mutex
  - This one happens when one thread is allocating memory while a fork
    occurs.
  - This closely relates to the code invalidation mutex, just happens at
    the syscall layer instead of the FEXCore layer.
  - Happens as frequently as the code invalidation mutex.
- Allocation mutex
  - This mutex is used for FEX's 64-bit Allocator, this happens when FEX
    is allocating memory on one thread and a fork occurs.
  - Fairly infrequent because jemalloc doesn't allocate VMA regions that
    often.

While this likely doesn't hit all of the FEX mutexes, this hits the ones
that are burning fires and are happening frequently.

Fixes #2664

### DIFF
--- a/External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
+++ b/External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
@@ -11,6 +11,91 @@
 #include <unistd.h>
 
 namespace FEXCore {
+#ifndef _WIN32
+  // Replacement for std::mutexes to deal with unlocking issues in the face of Linux fork() semantics.
+  //
+  // A fork() only clones the parent's calling thread. Other threads are silently dropped, which permanently leaves any mutexes owned by them locked.
+  // To address this issue, ForkableUniqueMutex and ForkableSharedMutex provide a way to forcefully remove any dangling locks and reset the mutexes to their default state.
+  class ForkableUniqueMutex final {
+    public:
+      ForkableUniqueMutex()
+        : Mutex (PTHREAD_MUTEX_INITIALIZER) {
+      }
+
+      // Move-only type
+      ForkableUniqueMutex(const ForkableUniqueMutex&) = delete;
+      ForkableUniqueMutex& operator=(const ForkableUniqueMutex&) = delete;
+      ForkableUniqueMutex(ForkableUniqueMutex &&rhs) = default;
+      ForkableUniqueMutex& operator=(ForkableUniqueMutex &&) = default;
+
+      void lock() {
+        [[maybe_unused]] const auto Result = pthread_mutex_lock(&Mutex);
+        LOGMAN_THROW_A_FMT(Result == 0, "{} failed to lock with {}", __func__, Result);
+      }
+      void unlock() {
+        [[maybe_unused]] const auto Result = pthread_mutex_unlock(&Mutex);
+        LOGMAN_THROW_A_FMT(Result == 0, "{} failed to unlock with {}", __func__, Result);
+      }
+      // Initialize the internal pthread object to its default initializer state.
+      // Should only ever be used in the child process when a Linux fork() has occured.
+      void StealAndDropActiveLocks() {
+        Mutex = PTHREAD_MUTEX_INITIALIZER;
+      }
+    private:
+      pthread_mutex_t Mutex;
+  };
+
+  class ForkableSharedMutex final {
+    public:
+      ForkableSharedMutex()
+        : Mutex (PTHREAD_RWLOCK_INITIALIZER) {
+      }
+
+      // Move-only type
+      ForkableSharedMutex(const ForkableSharedMutex&) = delete;
+      ForkableSharedMutex& operator=(const ForkableSharedMutex&) = delete;
+      ForkableSharedMutex(ForkableSharedMutex &&rhs) = default;
+      ForkableSharedMutex& operator=(ForkableSharedMutex &&) = default;
+
+      void lock() {
+        [[maybe_unused]] const auto Result = pthread_rwlock_wrlock(&Mutex);
+        LOGMAN_THROW_A_FMT(Result == 0, "{} failed to lock with {}", __func__, Result);
+      }
+      void unlock() {
+        [[maybe_unused]] const auto Result = pthread_rwlock_unlock(&Mutex);
+        LOGMAN_THROW_A_FMT(Result == 0, "{} failed to unlock with {}", __func__, Result);
+      }
+      void lock_shared() {
+        [[maybe_unused]] const auto Result = pthread_rwlock_rdlock(&Mutex);
+        LOGMAN_THROW_A_FMT(Result == 0, "{} failed to lock with {}", __func__, Result);
+      }
+
+      void unlock_shared() {
+        unlock();
+      }
+
+      bool try_lock() {
+        const auto Result = pthread_rwlock_trywrlock(&Mutex);
+        return Result == 0;
+      }
+
+      bool try_lock_shared() {
+        const auto Result = pthread_rwlock_tryrdlock(&Mutex);
+        return Result == 0;
+      }
+      // Initialize the internal pthread object to its default initializer state.
+      // Should only ever be used in the child process when a Linux fork() has occured.
+      void StealAndDropActiveLocks() {
+        Mutex = PTHREAD_RWLOCK_INITIALIZER;
+      }
+    private:
+      pthread_rwlock_t Mutex;
+  };
+#else
+  // Windows doesn't support forking, so these can be standard mutexes.
+  using ForkableUniqueMutex = std::mutex;
+  using ForkableSharedMutex = std::shared_mutex;
+#endif
 
   template<typename MutexType, void (MutexType::*lock_fn)(), void (MutexType::*unlock_fn)()>
   class ScopedDeferredSignalWithMutexBase final {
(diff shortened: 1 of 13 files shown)


External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
```diff
@@ -11,6 +11,82 @@
...
+
+  class ForkableSharedMutex final {
+    public:
+      ForkableSharedMutex()
+        : Mutex (PTHREAD_RWLOCK_INITIALIZER) {
+      }
+
+      // Move-only type
+      ForkableSharedMutex(const ForkableSharedMutex&) = delete;
+      ForkableSharedMutex& operator=(ForkableSharedMutex&) = delete;
```
lioncash, line 49:
```suggestion
      ForkableSharedMutex& operator=(const ForkableSharedMutex&) = delete;
```

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
```diff
@@ -11,6 +11,82 @@
...
+      ForkableSharedMutex(ForkableSharedMutex &&rhs) = default;
```
lioncash, line 50: Should also have the `ForkableSharedMuted& operator=(ForkableSharedMutex &&) = default;` specified too

ditto for ForkableUniqueMutex

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
```diff
@@ -11,6 +11,84 @@
...
+        const auto Result = pthread_rwlock_trywrlock(&Mutex);
+        return Result == 0;
+      }
+
+      bool try_lock_shared() {
+        const auto Result = pthread_rwlock_tryrdlock(&Mutex);
+        return Result == 0;
+      }
+
+      void DefaultInitialize() {
```
neobrain, line 81: This is not a useful name. What does "default initialize" mean in terms of the mutex functionality?

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
```diff
@@ -11,6 +11,84 @@
...
+        Mutex = PTHREAD_RWLOCK_INITIALIZER;
+      }
+    private:
+      pthread_rwlock_t Mutex;
+  };
+#else
+  // Windows doesn't support forking, these can use c++ types.
```
neobrain, line 88:
```suggestion
  // Windows doesn't support forking, so these can be standard mutexes
```

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
```diff
@@ -11,6 +11,84 @@
 #include <unistd.h>
 
 namespace FEXCore {
+#ifndef _WIN32
+  class ForkableUniqueMutex final {
```
neobrain, line 15: Needs a docstring explaining why `std::*mutex`es can't be forked reliably.

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Added a docstring.

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
```diff
@@ -808,17 +830,16 @@ uint64_t UnimplementedSyscallSafe(FEXCore::Core::CpuStateFrame *Frame, uint64_t
...
 }
 
-void SyscallHandler::UnlockAfterFork() {
-  // Add other mutexes here
-
-  // XXX shared_mutex has issues with locking and forks
-  // VMATracking.Mutex.unlock();
+void SyscallHandler::UnlockAfterFork(bool Child) {
+  if (Child) {
+    VMATracking.Mutex.DefaultInitialize();
```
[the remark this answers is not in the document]
Sonicadvance1: Due to fork semantics, the child also technically ends up owning the mutex but with a different PID with the internal TID tracking of the pthread mutex.

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Renamed to `StealAndDropActiveLocks`

External/FEXCore/include/FEXCore/Utils/DeferredSignalMutex.h
Sonicadvance1: Changed the comment over to the quoted one.

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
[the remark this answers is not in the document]
neobrain: I phrased the question poorly and actually meant to ask about the *active thread* of the child process (i.e. the only child process thread that may still own a "valid" lock).

Sounds like your reply about internal TID tracking still applies (so we can't safely call `std::mutex::unlock`?), but won't the thread later try to `unlock` the thread regardless (in the function that originally acquired the mutex)?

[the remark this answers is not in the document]
Sonicadvance1: The child process of fork can't safely call std::mutex::unlock, nor can it call the pthread equivalents. Once it gets initialized back to its default state with `StealAndDropActiveLocks` it is then safe to lock and unlock as it wants.

The parent thread of the fork process is still allows to lock and unlock as it pleases, and we lock all critical mutexes going in to a fork so that the rest of the code doing state tracking doesn't get in to a weird state.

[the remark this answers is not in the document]
neobrain: > The child process of fork can't safely call std::mutex::unlock, nor can it call the pthread equivalents. Once it gets initialized back to its default state with StealAndDropActiveLocks it is then safe to lock and unlock as it wants.

It *is* safe to lock+unlock, but is it safe to *just* unlock after the fork? From the looks of it the new code would fire an assertion in that case (as would `std::mutex` previously).

The scenario I'm worried about is where a fork happens from a thread that FEX holds an `std::unique_lock` in. Sure, `StealAndDropActiveLocks` will reset the underlying mutex, but the `unique_lock` destructor will still attempt to `unlock()` that mutex after its reset.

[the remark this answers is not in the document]
Sonicadvance1: > is it safe to just unlock after the fork? 
Depends on if it is the parent thread or the child process
Parent thread is fine, child process needs to `StealAndDropActiveLocks`.

> but the unique_lock destructor
These are all locked without using `std::unique_lock` and instead locking the mutex objects directly, there is no destructor being called here.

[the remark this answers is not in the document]
neobrain: > Depends on if it is the parent thread or the child process

I'm not concerned about the parent thread, all of this is referring to the child process.

>  Parent thread is fine, child process needs to StealAndDropActiveLocks.

All this is referring to StealAndDropActiveLocks already having been called after fork().

> These are all locked without using std::unique_lock and instead locking the mutex objects directly, there is no destructor being called here.

Instead of a `unique_lock` we'll have a manual `lock` + `unlock` pair in the acquiring the code then. Since we're talking about the active thread, we'll end up with a sequence of events like this:
1. lock in some FEX function F
2. fork (maybe due to a signal handler?) and then return to FEX function F
3. unlock at the end of FEX function F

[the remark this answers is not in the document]
neobrain: Actually, I think I'm hitting this point:
> While this likely doesn't hit all of the FEX mutexes, this hits the ones that are burning fires and are happening frequently.

Since we're only dealing with 3 specific mutexes, the issue pointed out above is easily avoided with manual care. More complicated scenarios (where we `fork` with an unknown set of mutexes being held in the forking thread) just don't happen, right?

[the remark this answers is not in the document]
Sonicadvance1: There should be no case in which we have locked a mutex and then forked (On the calling thread that is forking at least).
Fork being a syscall means it always comes from the the JIT where that is guaranteed to not hold any host side mutexes.
Any case where we would get interrupted while holding a critical mutex either disables the signal mask entirely, or ends up in the deferred signal path, which gets deferred to a point where a critical mutex isn't getting held.

lioncash (approved):
LGTM unless I accidentally skipped over ongoing discussion

neobrain (approved):
Fine from my end as well - had to think a bit about the last discussion I opened, but should be all safe indeed!

PR #2747, opened 2023-07-08
Thunks: Fixes thunks in non-multiarch

Sonicadvance1:
Removes the @PREFIX_ARCH@ replacement  string in the thunks path.

The library prefix paths now get generated upfront and everything gets replaced to handle the differences between multiarch distros.

Fixes Thunks on Arch and Fedora.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/FileManagement.cpp
@@ -15,6 +15,7 @@ $end_info$
 #include "LinuxSyscalls/x64/Syscalls.h"
 
 #include <FEXCore/Utils/LogManager.h>
+#include <FEXCore/Utils/FileLoading.h>
 #include <FEXCore/fextl/fmt.h>
 #include <FEXCore/fextl/list.h>
 #include <FEXCore/fextl/string.h>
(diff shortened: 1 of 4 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/FileManagement.cpp
```diff
@@ -59,48 +60,69 @@ namespace JSON {
...
+        "i386-linux-gnu", // 32-bit
+        "x86_64-linux-gnu", // 64-bit
+      };
 
-  close(fd);
-  return ReadSize == FileSize;
-}
+      // Now just generate the combination of the two.
+      // We only need to generate 32-bit or 64-bit depending on the operating mode.
+      const auto ArchPrefix = MultiArchPrefixes[Is64BitMode()];
```
neobrain, line 97: Why not just `const auto ArchPrefix = Is64BitMode() ? "x86_64..." : "i386..."` ?

neobrain (approved):
The two instances of `MultiArchPrefixes` are unnecessarily verbose, but looks good otherwise!

Source/Tools/FEXLoader/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Dropped the array indexing and just used a ternary.

PR #2750, opened 2023-07-09
OpcodeDispatcher: Fixes bug with pcmpestri

Sonicadvance1:
When this instruction returns the index in to the ecx register, this is defined as a 32-bit result. This means it actually gets zero-extended to the full 64-bit GPR size on 64-bit processes.
Previously FEX was doing a 32-bit insert which leaves garbage data in the upper 32-bits of the RCX register.

Adds a unit test to ensure the result is zero extended. Fixes running Java games under FEX now that SSE4.2 is exposed.


lioncash:
Oh! Nice catch!

lioncash (approved):

PR #2760, opened 2023-07-11
Arm64: Switch to using half barriers

Sonicadvance1:
Inspired from: https://github.com/dotnet/runtime/issues/8072

Currently FEX is /very/ heavy handed with our backpatching where we wrap every backpatched loadstore with `dmb ish`.

This can be relaxed slightly according to the linked issue.

For TSO load instructions the instruction sequence changes to:
  ldr <args>;
  dmb ld; <-- Slightly less strict dmb

For TSO store instructions the instruction sequence changes to:
  dmb ish; <-- Still the all encompassing dmb
  str <args>;

For backpatching loadstores this does the same thing where only one side needs the nop and it uses the same instruction sequence when backpatched.

The minor change is that on load backpatching, we are no longer backing up a single instruction, instead just re-executing the instruction we patched directly.

Took a long time to come back to this (Last looked in August 2020). Previously when I was implementing this idea it didn't work, but that was because our CompareExchange operation was broken back then. With the CAS now, it should just work.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -1231,8 +1231,6 @@ DEF_OP(LoadMemTSO) {
       ldapurb(Dst, MemReg, Offset);
     }
     else {
-      // Aligned
-      nop();
       switch (OpSize) {
         case 2:
           ldapurh(Dst, MemReg, Offset);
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Oops, forgot to change the DMB on backpatch from `dmb ish` to `dmb ishld` on load backpatches.
Fixed now.

Sonicadvance1:
Half-Barrier provides some very good performance improvements depending on the benchmark. Up to 193% the performance.
Image_2023-07-25_05-48-45

lioncash (approved):

neobrain:
Would it be worthwhile to use half-barriers for [memcpy](https://github.com/FEX-Emu/FEX/blob/d01db8f2932235325f40cd4dc01acdc6c5099b2a/FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp#L1989-L2047) as well?

PR #2763, opened 2023-07-12
Thunks/X11: Fixes variadic packing and callbacks.

Sonicadvance1:
Two bugs here that caused thunking X11 thunking in Wine/Proton to not work.

The easier of the two. The various variadic functions that we thunk actually take key:value pairs where the first is a string pointer, and the value can be various things.

We need to handle these as true key:value pairs rather than finding the first nullptr and dropping the remainder.

Additionally, there are 12 keys that specify a callback that FEX needs to catch and convert to host callable. Wine is the first application that I have seen that actually uses this. If these callbacks aren't wired up then it it can miss events.

The harder of the two problems is the `libX11_Variadic_u64` function was subtly incorrect. Nothing had previously truly exercised this and my test program didn't notice anything wrong while writing it.

The first incorrect thing was that it was subtracting the nullptr ender variable before the stack size calculation, causing the value to overwrite the stack if the number of remaining elements was event.

Secondly the assembly that was storing two elements per step was decrementing the counter by 8 instead of two. Didn't pick this up before since I believe the code was only hitting the non-pair path before.

This gets Proton thunking working under FEX now.

### DIFF
--- a/ThunkLibs/libX11/libX11_Guest.cpp
+++ b/ThunkLibs/libX11/libX11_Guest.cpp
@@ -20,134 +20,214 @@ extern "C" {
 #undef max
 }
 
+#include <algorithm>
+#include <array>
+#include <cassert>
 #include <cstdint>
 #include <stdio.h>
 #include <cstring>
 #include <map>
+#include <list>
 #include <string>
+#include <unistd.h>
 
 #include "common/Guest.h"
 #include <stdarg.h>
 
 #include "thunkgen_guest_libX11.inl"
+#include "X11Common.h"
 
 // Custom implementations //
 
 #include <vector>
 
+namespace {
+  // The various X11 variadic functions take a flattened sequence of key:value pairs as arguments.
+  // The key element is a pointer to a string.
+  // The value element is a pointer to the data of that key type.
+  //
+  // Some keys describe function callbacks for various events that the server interface can call.
+  // FEX needs to walk these keys and ensure any callback function has a trampoline so the native
+  // X11 library can call it.
+  template<typename CallbackType>
+  static std::list<CallbackType> ConvertCallbackArguments(std::vector<void*> &IncomingArguments) {
+    assert(IncomingArguments.size() % 2 == 0 && "Incoming arguments needs to be in pairs");
+
+    std::list<CallbackType> Callbacks;
+    // Walk the arguments and convert any callbacks.
+    const size_t ArgumentPairs = IncomingArguments.size() / 2;
+    for (size_t i = 0; i < ArgumentPairs; ++i) {
+      const char *Key = static_cast<const char*>(IncomingArguments[i * 2]);
+      void** Data = &IncomingArguments[i * 2 + 1];
+      if (!*Data) {
+        continue;
+      }
+
+      // Check if the key is a callback and needs to be modified.
+      auto KeyIt = std::find(X11::CallbackKeys.begin(), X11::CallbackKeys.end(), Key);
+      if (KeyIt == X11::CallbackKeys.end()) {
+        continue;
+      }
+
+      // Key matches a callback, we need to wrap this.
+      CallbackType *IncomingCallback = reinterpret_cast<CallbackType*>(*Data);
+      CallbackType *ConvertedCallback = &Callbacks.emplace_back(CallbackType {
+        // Client data stays the same.
+        .client_data = IncomingCallback->client_data,
+
+        // Callback needs a trampoline.
+        .callback = AllocateHostTrampolineForGuestFunction(IncomingCallback->callback),
+      });
+
+      // Add this converted back in.
+      *Data = ConvertedCallback;
+    }
+
+    return Callbacks;
+  }
+}
+
 extern "C" {
     char* XGetICValues(XIC ic, ...) {
-        fprintf(stderr, "XGetICValues\n");
-        va_list ap;
-        std::vector<unsigned long> args;
-        va_start(ap, ic);
-        for (;;) {
-            auto arg = va_arg(ap, unsigned long);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%016lX\n", arg);
-        }
+      fprintf(stderr, "XGetICValues\n");
+      va_list ap;
+      std::vector<void*> args;
+      va_start(ap, ic);
+      for (;;) {
+        auto arg = va_arg(ap, void*);
+        if (arg == 0)
+          break;
+        args.push_back(arg);
+        fprintf(stderr, "%p\n", arg);
+      }
 
-        va_end(ap);
-        auto rv = fexfn_pack_XGetICValues_internal(ic, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+      va_end(ap);
+
+      auto rv = fexfn_pack_XGetICValues_internal(ic, args.size(), &args[0]);
+      fprintf(stderr, "RV: %p\n", rv);
+      return rv;
     }
 
     char* XSetICValues(XIC ic, ...) {
-        fprintf(stderr, "XSetICValues\n");
-        va_list ap;
-        std::vector<unsigned long> args;
-        va_start(ap, ic);
-        for (;;) {
-            auto arg = va_arg(ap, unsigned long);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%016lX\n", arg);
-        }
+      fprintf(stderr, "XSetICValues\n");
+      va_list ap;
+      std::vector<void*> IncomingArguments;
+      va_start(ap, ic);
+      for (;;) {
+        auto Key = va_arg(ap, void*);
+        if (Key == 0)
+          break;
 
-        va_end(ap);
-        auto rv = fexfn_pack_XSetICValues_internal(ic, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+        auto Value = va_arg(ap, void*);
+        IncomingArguments.emplace_back(Key);
+        IncomingArguments.emplace_back(Value);
+      }
+
+      va_end(ap);
+
+      // Callback memory needs to live beyond internal function call.
+      std::list<XICCallback> Callbacks = ConvertCallbackArguments<XICCallback>(IncomingArguments);
+
+      auto rv = fexfn_pack_XSetICValues_internal(ic, IncomingArguments.size(), &IncomingArguments[0]);
+      fprintf(stderr, "RV: %p\n", rv);
+      return rv;
     }
 
     char* XGetIMValues(XIM ic, ...) {
-        fprintf(stderr, "XGetIMValues\n");
-        va_list ap;
-        std::vector<void*> args;
-        va_start(ap, ic);
-        for (;;) {
-            auto arg = va_arg(ap, void*);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%p\n", arg);
-        }
+      fprintf(stderr, "XGetIMValues\n");
+      va_list ap;
+      std::vector<void*> args;
+      va_start(ap, ic);
+      for (;;) {
+        auto arg = va_arg(ap, void*);
+        if (arg == 0)
+          break;
+        args.push_back(arg);
+        fprintf(stderr, "%p\n", arg);
+      }
 
-        va_end(ap);
-        auto rv = fexfn_pack_XGetIMValues_internal(ic, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+      va_end(ap);
+      auto rv = fexfn_pack_XGetIMValues_internal(ic, args.size(), &args[0]);
+      fprintf(stderr, "RV: %p\n", rv);
+      return rv;
     }
 
     char* XSetIMValues(XIM ic, ...) {
-        fprintf(stderr, "XSetIMValues\n");
-        va_list ap;
-        std::vector<void*> args;
-        va_start(ap, ic);
-        for (;;) {
-            auto arg = va_arg(ap, void*);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%p\n", arg);
-        }
+      fprintf(stderr, "XSetIMValues\n");
+      va_list ap;
+      std::vector<void*> IncomingArguments;
+      va_start(ap, ic);
+      for (;;) {
+        auto Key = va_arg(ap, void*);
+        if (Key == 0)
+          break;
 
-        va_end(ap);
-        auto rv = fexfn_pack_XSetIMValues_internal(ic, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+        auto Value = va_arg(ap, void*);
+        IncomingArguments.emplace_back(Key);
+        IncomingArguments.emplace_back(Value);
+        fprintf(stderr, "%s\n", (char*)Key);
+      }
+
+      va_end(ap);
+
+      // Callback memory needs to live beyond internal function call.
+      std::list<XIMCallback> Callbacks = ConvertCallbackArguments<XIMCallback>(IncomingArguments);
+
+      // Send a count (not including nullptr);
+      auto rv = fexfn_pack_XSetIMValues_internal(ic, IncomingArguments.size(), &IncomingArguments[0]);
+      fprintf(stderr, "RV: %p\n", rv);
+      return rv;
     }
 
     _XIC* XCreateIC(XIM im, ...) {
-        fprintf(stderr, "XCreateIC\n");
-        va_list ap;
-        std::vector<unsigned long> args;
-        va_start(ap, im);
-        for (;;) {
-            auto arg = va_arg(ap, unsigned long);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%016lX\n", arg);
-        }
+      fprintf(stderr, "XCreateIC\n");
+      va_list ap;
+      std::vector<void*> IncomingArguments;
 
-        va_end(ap);
-        auto rv = fexfn_pack_XCreateIC_internal(im, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+      va_start(ap, im);
+      for (;;) {
+        auto Key = va_arg(ap, void*);
+        if (Key == 0)
+          break;
+
+        auto Value = va_arg(ap, void*);
+        IncomingArguments.emplace_back(Key);
+        IncomingArguments.emplace_back(Value);
+      }
+
+      va_end(ap);
+
+      // Callback memory needs to live beyond internal function call.
+      std::list<XICCallback> Callbacks = ConvertCallbackArguments<XICCallback>(IncomingArguments);
+
+      auto rv = fexfn_pack_XCreateIC_internal(im, IncomingArguments.size(), &IncomingArguments[0]);
+      return rv;
     }
 
     XVaNestedList XVaCreateNestedList(int unused_arg, ...) {
-        fprintf(stderr, "XVaCreateNestedList\n");
-        va_list ap;
-        std::vector<void*> args;
-        va_start(ap, unused_arg);
-        for (;;) {
-            auto arg = va_arg(ap, void*);
-            if (arg == 0)
-                break;
-            args.push_back(arg);
-            fprintf(stderr, "%p\n", arg);
-        }
+      fprintf(stderr, "XVaCreateNestedList\n");
+      va_list ap;
+      std::vector<void*> IncomingArguments;
 
-        va_end(ap);
-        auto rv = fexfn_pack_XVaCreateNestedList_internal(unused_arg, args.size(), &args[0]);
-        fprintf(stderr, "RV: %p\n", rv);
-        return rv;
+      va_start(ap, unused_arg);
+      for (;;) {
+        auto Key = va_arg(ap, void*);
+        if (Key == 0)
+          break;
+
+        auto Value = va_arg(ap, void*);
+        IncomingArguments.emplace_back(Key);
+        IncomingArguments.emplace_back(Value);
+      }
+
+      va_end(ap);
+
+      // Callback memory needs to live beyond internal function call.
+      std::list<XICCallback> Callbacks = ConvertCallbackArguments<XICCallback>(IncomingArguments);
+
+      auto rv = fexfn_pack_XVaCreateNestedList_internal(unused_arg, IncomingArguments.size(), &IncomingArguments[0]);
+      fprintf(stderr, "RV: %p\n", rv);
+      return rv;
     }
 
     static void LockMutexFunction(LockInfoPtr) {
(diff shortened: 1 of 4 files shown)


ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -35,119 +38,198 @@ extern "C" {
...
+    XNPreeditStateNotifyCallback,
+    XNStatusStartCallback,
+    XNStatusDoneCallback,
+    XNStatusDrawCallback,
+    XNR6PreeditCallback,
+    XNStringConversionCallback,
+  }};
+
+  template<typename CallbackType>
+  static void ConvertArguments(std::vector<void*> &IncomingArguments, std::list<CallbackType> *Callbacks) {
```
neobrain, line 58: This can construct the `std::list<CallbackType>` internally and return it to the caller to avoid unnecessary out-parameters.

neobrain, line 58: `IncomingArguments` should be const

neobrain, line 58: Should have a docstring explaining that (apparently?) all variadic X11 functions have in common that they take a flattened sequence of X/Y pairs as arguments.

ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -35,119 +38,198 @@ extern "C" {
...
+    // Walk the arguments and convert any callbacks.
```
neobrain, line 59: Should have an assert that `IncomingArguments.size()%2==0`

ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -35,119 +38,198 @@ extern "C" {
...
+    const size_t ArgumentPairs = IncomingArguments.size() / 2;
+    for (size_t i = 0; i < ArgumentPairs; ++i) {
+      const char *Key = static_cast<const char*>(IncomingArguments[i * 2]);
+      void** Data = &IncomingArguments[i * 2 + 1];
+
+      // Check if the key is a callback and needs to be modified.
+      for (auto CallbackKey : CallbackKeys) {
+        if (CallbackKey == Key && *Data) {
```
neobrain, line 67: This can be flattened a lot:

```suggestion
      void** Data = &IncomingArguments[i * 2 + 1];
      if (!*Data) {
      	continue;
      }

      // Check if the key is a callback and needs to be modified.
      auto KeyIt = std::find(CallbackKeys.begin(), CallbackKeys.end(), Key);
      if (KeyIt == CallbackKeys.end()) {
      	continue;
      }
      
      // ...
```

ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -35,119 +38,198 @@ extern "C" {
...
+          // Key matches a callback, we need to wrap this.
+          CallbackType *IncomingCallback = reinterpret_cast<CallbackType*>(*Data);
+          CallbackType *ConvertedCallback = &Callbacks->emplace_back(CallbackType {
+            // Client data stays the same.
+            .client_data = IncomingCallback->client_data,
+
+            // Callback needs a trampoline.
+            .callback = AllocateHostTrampolineForGuestFunction(IncomingCallback->callback),
```
neobrain, line 75: Why don't we modify this callback in-place? We don't need the original value of `IncomingCallback->callback` after setting up the trampoline.

neobrain, line 75: Where is the corresponding `FinalizeHostTrampolineForGuestFunction` call on the host-side?

Sonicadvance1: We can't modify the callback in place because the data is coming from the guest program.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: We're modifying in-place, so it being const wouldn't make sense.

ThunkLibs/libX11/libX11_Guest.cpp
neobrain: Ah yes, I see now. It's modifying the inner pointee data.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: Returning callback list now

Sonicadvance1: Added a comment about this.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: Assert added.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: Flattened.

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: You're right, missed it.
Added now.

Sonicadvance1:
As an additional note, this also fixes some Unity native Linux game thunking.
Tested Unpacking, which previously didn't work with thunking on their Linux client, and it works after this PR.

Sonicadvance1:
Additional side-note. This fixes thunking on XWayland as well it looks like.

alyssarosenzweig:
Does `ConvertCallbackArguments` get called in a hot path? If so, it looks like a good candidate for a [perfect hash table](https://en.wikipedia.org/wiki/Perfect_hash_function). If not, the `std::find` is probably ok if not ideal

alyssarosenzweig:
The assembly is over my head. The C looks good, other than the possible performance concern from all the string compares.

Sonicadvance1:
> Does `ConvertCallbackArguments` get called in a hot path? If so, it looks like a good candidate for a [perfect hash table](https://en.wikipedia.org/wiki/Perfect_hash_function). If not, the `std::find` is probably ok if not ideal

This code only gets called during application initialization, or massive state changes. So I figured linear scan was fine here.

alyssarosenzweig:
> > Does `ConvertCallbackArguments` get called in a hot path? If so, it looks like a good candidate for a [perfect hash table](https://en.wikipedia.org/wiki/Perfect_hash_function). If not, the `std::find` is probably ok if not ideal
> 
> This code only gets called during application initialization, or massive state changes. So I figured linear scan was fine here.

:+1:

alyssarosenzweig (approved):

neobrain:
Seeing how much overhead there is with temporary `std::vector` allocation, variadic argument list parsing, and potentially `std::list` allocation, is the inline assembly in `libX11_Variadic_u64` still warranted? I do remember spending a while to wade through the assembly during review of #1929...

Sonicadvance1:
> Seeing how much overhead there is with temporary `std::vector` allocation, variadic argument list parsing, and potentially `std::list` allocation, is the inline assembly in `libX11_Variadic_u64` still warranted? I do remember spending a while to wade through the assembly during review of #1929...

The inline assembly is required for correctness not performance. There's no other way to manually pack these variadic functions that don't have an upper limit on arguments. Wine only uses up to 14 arguments, but there's no reason that they couldn't use more in the future.

neobrain:
> The inline assembly is required for correctness not performance.

Ah, right. I got confused by some notes in #1929 focusing on codegen...

> There's no other way to manually pack these variadic functions that don't have an upper limit on arguments. Wine only uses up to 14 arguments, but there's no reason that they couldn't use more in the future.

As much as I want to agree with having a general solution, FTR locking us into an upper bound for the argument count using trivially understandable C code would have been correct from the beginning and would require less review time to validate. I guess it's fine as long as it's *really* correct now, but let's revisit this approach if we find *more* bugs later.

ThunkLibs/libX11/libX11_Guest.cpp
```diff
@@ -20,134 +20,216 @@ extern "C" {
...
+namespace {
+  // The various X11 variadic functions take a flattened sequence of key:value pairs as arguments.
+  // The key element is a pointer to a string.
+  // The value element is a pointer to the data of that key type.
+  //
+  // Some keys describe function callbacks for various events that the server interface can call.
+  // FEX needs to walk these keys and ensure any callback function has a trampoline so the native
+  // X11 library can call it.
+  template<typename CallbackType>
+  static std::list<CallbackType> ConvertCallbackArguments(std::vector<void*> &IncomingArguments) {
```
neobrain, line 53: Much of the logic in this function can be cleaned up by changing the type of this to `std::vector<std::pair<const char*, void*>>`

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,29 +166,36 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
       mov x11, sp
 
-      # x12 - load offset
+      # x12 - load offset - From array source.
+      # Skipping the first seven arguments.
       add x12, x9, (7 * 8)
 
-      cmp x10, 1
+      # x1 currently contains the number of elements including nullptr.
+      # We want the number of arguments to store, not including the nullptr.
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,29 +166,36 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
+      sub x10, x10, 2
       cmp x10, 1
       b.gt .load_pair%=
       b.lt .no_single%=
 
-      # One variable at most
       .single%=:
+      # One variable at most
+      # Load the argument from the source.
+      # Then store that and nullptr ender.
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -124,8 +127,30 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
 
       # Stack objects
       # Count >= 7
+      # The number of these arguments that end up in stack space is slightly different than expected.
+      # Argument layout:
+      #   r0 = <arg im>
```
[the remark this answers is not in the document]
Sonicadvance1: Renamed to `a_0` to match the passed in argument name.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -124,8 +127,30 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
+      #   Stack[0] = <Args or nullptr>
+      #   Stack[...] = <Last element always nullptr>
+      # The number of arguments on the stack will always be (Count - 7) + 1
+
+      # Common:
+      # 14 objects:
+      #   - 7 register arguments
+      #   - 7 arguments on stack
+      #   - nullptr on stack
+      #   = 8 objects total on stack.
```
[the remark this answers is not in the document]
Sonicadvance1: Removed the comment to not be misleading. Already have a description of the stack layout.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,29 +166,36 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
+      # x1 currently contains the number of elements including nullptr.
+      # We want the number of arguments to store, not including the nullptr.
+      subs x10, x1, 1
+
       b.eq .single%=
       b.lt .no_single%=
 
       .load_pair%=:
+      # Load the source from x12
+      # Store it in to x11
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,29 +166,36 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
+      # x12 - load offset - From array source.
+      # Skipping the first seven arguments.
       add x12, x9, (7 * 8)
 
-      cmp x10, 1
+      # x1 currently contains the number of elements including nullptr.
+      # We want the number of arguments to store, not including the nullptr.
+      subs x10, x1, 1
+
       b.eq .single%=
```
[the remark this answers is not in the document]
Sonicadvance1: These paths are just taking advantage of the hardware.
`load_pair` - is just using ldp+stp to copy 128-bits per step.
`single` - uses ldr+stp to load 64-bit and store 128-bit (the final nullptr) in a step.
`no_single` is the case in which there is only the final nullptr to store. So it does it in a single str.

The passed in array doesn't have a nullptr at the end since there is no reason to eat a stack slot in that case.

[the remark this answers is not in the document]
neobrain: Does this make a measurable difference? Can we use a branchless implementation instead to keep complexity more contained by not taking advantage of the hardware?

ThunkLibs/libX11/libX11_Guest.cpp
Sonicadvance1: Converting this to a vector of pairs introduces a questionable crash.
In addition it adds some bad castings for the `XGet*` handlers which don't use pairs.

ThunkLibs/libX11/libX11_Host.cpp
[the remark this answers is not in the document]
Sonicadvance1: There's definitely going to be a loop still, so it won't be branchless.
Technically it makes a measurable difference, because ldp+stp has the same performance as ldr+str but it is moving double the number of bits. 64 versus 128.

ThunkLibs/libX11/libX11_Guest.cpp
neobrain: > Converting this to a vector of pairs introduces a questionable crash.

What makes it questionable?

> In addition it adds some bad castings for the XGet* handlers which don't use pairs.

How does the current implementation of the function work with these functions? Also, the comment at the top seems very misleading if these functions don't in fact use pairs.

Sonicadvance1: The `XGet*` variants push keys in to a `std::vector<void*>` until they hit a nullptr, then passes that array directly through and is handled in the variadic case like previously. It may or may not be correct, which wasn't the point of this PR.

The questionable crash is that with pairs and then casting the to the variadic handler of void* it crashes. Which will be difficult to nail down.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -124,8 +127,30 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
 
       # Stack objects
       # Count >= 7
+      # The number of these arguments that end up in stack space is slightly different than expected.
```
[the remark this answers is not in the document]
Sonicadvance1: How many lines of the comment do you want moved? It's more detail about how that assembly is laying out the stack frame next to the implementation.
If you want it moved I don't care where it ends up.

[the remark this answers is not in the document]
neobrain: Right now this information is partly redundant in a bad way, since that redundancy isn't helping at understanding what's going on. The redundant bits should be removed, the bits that relate to local code should be kept, and the bits that don't relate to local code should probably be moved to the docstring.

I don't know what you're hoping to clarify by adding this here, so I can't tell you what exactly to change.

ThunkLibs/libX11/libX11_Host.cpp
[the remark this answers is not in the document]
neobrain: This reads to me like the data suggests it makes no practical difference on performance? That would make the loop probably the more maintainable solution. I don't know if it makes a big difference, but I know I'm unhappy with unjustified complexity in this function.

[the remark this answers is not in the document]
Sonicadvance1: It has exactly 2x the difference in performance.

ThunkLibs/libX11/libX11_Host.cpp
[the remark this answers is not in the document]
Sonicadvance1: Removed the comment.

ThunkLibs/libX11/libX11_Host.cpp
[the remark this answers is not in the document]
neobrain: That's referring to the individual instructions, not the overall practical effect. There is little need in micro-optimizing here considering the whole function is wrapped by one heap allocation and potentially more.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -31,6 +33,7 @@ extern "C" {
 #include <utility>
 
 #include "thunkgen_host_libX11.inl"
+#include "X11Common.h"
 
 #ifdef _M_ARM_64
 // This Variadic asm only works for one signature
```
neobrain, line 39: Looks like this never explained what this function actually does. The complexity of this function is growing with this PR, so this is really needed now.

Probably should explain the differences between the AArch64 ABI and the guest ABI (which I suggested be moved here in another comment), and how we resolve that difference.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -46,15 +49,15 @@ extern "C" {
 //
 // Incoming:
 // x0 = XIM
-// x1 = count
-// x2 = array of 64-bit values
+// x1 = count (Not including final nullptr)
+// x2 = array of 64-bit values (not including final nullptr)
```
neobrain, line 53:
```suggestion
// x1 = count (excluding final nullptr)
// x2 = array of 64-bit values (excluding final nullptr)
```

ThunkLibs/libX11/libX11_Host.cpp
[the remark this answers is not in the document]
Sonicadvance1: Removed the optimization. It now just copies one 64-bit value at a time.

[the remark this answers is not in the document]
neobrain: (Note I'm not sure what the single-branch alternative would look like, so for all I know it wouldn't make all that much difference. I'm just pointing out the current code is overly complex for the context it's used in.)

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: This doesn't care about guest ABI at this point. We're already a function call in, this is just packing variadic arguments since there is physically no way to do it without inline ASM.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: Resolved.

ThunkLibs/libX11/libX11_Host.cpp
neobrain: Ah! That sounds like a good comment to add. I expanded it a bit, hopefully I got it right:
```suggestion
// Helper function to call the given variadic function with `a_0` and a variadic argument list packed into an array.
// `Func` must have the same signature as XGetIMValues.
```

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -122,10 +125,11 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
       # x1 = <count>
       # x9 = <list ptr>
 
-      # Stack objects
-      # Count >= 7
+      # The number of arguments on the stack will always be (Count - 7) + 1
+
```
neobrain, line 129: The comment can be simplified a lot.

I would add this under `.stack%=:`:
```suggestion
      # The first 7 variadic arguments fit into reserved registers, but we must copy the rest to the stack. X11 further requires us to add a terminating nullptr, so in total (count - 6) arguments are written to the stack, starting from list[7].
```

Then this entire block code section can be replaced with:
```
      # In x1, store the number of arguments written to stack (including the terminating nullptr).
      # We must round to the nearest pair because TODO
      sub x1, x1, 6
      and x10, x1, 1
      add x10, x10, x1
```

The top comment then also explains why we skip the first 7 elements in x12 below.

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -122,10 +125,11 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
       # Subtract 6 count objects
-      # Leaves us at least 1 (nullptr)
+      # Gives us the total number of objects that need to be stored on to the stack.
+      # If exactly 7 objects then only nullptr ends up on the stack.
       sub x1, x1, 6
 
       # Round up to the nearest pair
```
neobrain, line 135: Why do we need to do this? Is it an expectation by X11 libraries, or something about the specifics in the AArch64 ABI?

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,34 +145,22 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
 
-      # x12 - load offset
+      # x12 - load offset into input array (skipping the first 7 arguments)
       add x12, x9, (7 * 8)
 
-      cmp x10, 1
-      b.eq .single%=
-      b.lt .no_single%=
+      # Compute the number of elements excluding the terminating nullptr
+      subs x10, x1, 1
```
neobrain, line 155: Can this instruction be folded into the `cmp x10, 1` below by changing it to `cmp x1, 2` and using x10 for the `ldr`/`str` instead?

ThunkLibs/libX11/libX11_Host.cpp
```diff
@@ -141,34 +145,22 @@ void *libX11_Variadic_u64(uint64_t a_0, size_t count, unsigned long *list, void
...
-
-      # One variable at most
       .single%=:
-      ldr x1, [x12]
-      stp x1, xzr, [x11]
-      b .top_reg_args%=
+      ldr x1, [x12], 8
+      str x1, [x11], 8
+      sub x10, x10, 1
+      cmp x10, 1
```
neobrain, line 161: This seems to copy one element if `count` is 7 now, but IIUC it should only write the nullptr in that case. Could use `ldr.ge`/`str.ge` instead, I guess?

neobrain (commented):
Should be good to go with the proposed documentation and the simplified implementation. If this causes problems in the future, let's revisit the non-generic C fallback.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: AArch64 ABI requires rsp to be 16-byte aligned.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: Likely just introduced a bug now since we're changing logic.

ThunkLibs/libX11/libX11_Host.cpp
Sonicadvance1: ldr.ge, str.ge aren't real.

ThunkLibs/libX11/libX11_Guest.cpp
neobrain: Further clarified offline: Changing to `std::vector<std::pair<...>>` would benefit the guest side in `ConvertCallbackArguments`, but the host side would have to do a bit of a dance since `XGet*` doesn't use pairs. The huge 64-value switch cases would have to be slightly different for each.

neobrain (approved):

Issue #2707, opened 2023-06-09
Optimize PF calculation


Sonicadvance1:
The PF calculation is very spicy and isn't generating optimal codegen.

Current codegen:
```asm
   0x00000002a10ed054:  and     x24, x22, #0xff
   0x00000002a10ed058:  fmov    d0, x24
   0x00000002a10ed05c:  cnt     v0.8b, v0.8b
   0x00000002a10ed060:  addv    b0, v0.8b
   0x00000002a10ed064:  umov    w24, v0.b[0]
   0x00000002a10ed068:  eor     x24, x24, #0x1
   0x00000002a10ed06c:  ubfx    x24, x24, #0, #1
   0x00000002a10ed070:  strb    w24, [x28, #706]
   ```
What it can be:
```asm
   0x00000002a10ed064:  movi    v1.8b, #1
   0x00000002a10ed058:  fmov    s0, w22
   0x00000002a10ed05c:  cnt     v0.8b, v0.8b
   0x00000002a10ed068:  bic     v0.8b, v1.8b, v0.8b
   0x00000002a10ed070:  str     b0, [x28, #706]
   ```
   
Not a significant reduction in instructions, but removes a FPR->GPR move which is quite an improvement.

For CPUs with CSSC extension this can be even more optimal but even the upcoming X4/A720/A520 doesn't support them.

llvm-mca results:
Before:
```
Iterations:        100
Instructions:      800
Total Cycles:      315
Total uOps:        1000

Dispatch Width:    10
uOps Per Cycle:    3.17
IPC:               2.54
Block RThroughput: 3.0
```

After:
```
Iterations:        100
Instructions:      500
Total Cycles:      408
Total uOps:        600

Dispatch Width:    10
uOps Per Cycle:    1.47
IPC:               1.23
Block RThroughput: 3.0
```

**Comment by dougallj** [comment]
That `BIC` should be an `AND`, right?

I'm personally partial to the variant:

```
    movi    v1.8b, #0
    fmov    s0, w22
    cnt     v0.8b, v0.8b
    sli     v0.8b, v1.8b, #1
    str     b0, [x28, #706]
```

Because `movi v1.8b, #0` can be handled by the renamer on Apple CPUs, that saves a SIMD uop. (Though the SLI may be more expensive than the AND on other microarchitectures, so YMMV. And PF flag calculations probably aren't happening in SIMD-uop bound code anyway?)

[...]

PR #2768, opened 2023-07-13
OpcodeDispatcher: Fix and optimize PF calculation

alyssarosenzweig:
We store garbage in the upper bits. That's ok, but it means we need to
    mask on read for correct behaviour.


alyssarosenzweig:
This takes care of half of #2707. It doesn't get rid of the FPR->GPR move, but I'm not sure whether we want to do that, vs deferring the whole calculation. Punting on that for now.

Sonicadvance1 (approved):
Nice catch on the PF usage having a bug!

PR #2774, opened 2023-07-13
CMake: Fix pkg version extraction for xcb

Sonicadvance1:
Our regex would only ever capture a single digit, so versions that had more than one digit per section would lose additional digits.

Fixes and moves the helper to a cmake file to be shared between GuestLibs and HostLibs.

Uses the fix in xcb because Fedora ships an older version that doesn't have some of FEX's newer symbols.

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -1,5 +1,6 @@
 cmake_minimum_required(VERSION 3.14)
 project(guest-thunks)
+include(${FEX_PROJECT_SOURCE_DIR}/CMakeFiles/version_to_variables.cmake)
 
 option(ENABLE_CLANG_THUNKS "Enable building thunks with clang" FALSE)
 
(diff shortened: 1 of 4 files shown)


lioncash (approved):

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -189,11 +190,7 @@ if (BITNESS EQUAL 64)
 
   find_package(PkgConfig)
   pkg_search_module(X11 REQUIRED x11)
-
-  string(REGEX MATCH "([0-9]*)\.([0-9]*)\.([0-9]*)" _ "${X11_VERSION}")
```
neobrain, line 193: In local testing, `string(REGEX MATCH "([0-9]*)\.([0-9]*)\.([0-9]*)" _ "1.10.430")` extracts the values "1", "10", and "430" fine for me. Is the problem somewhere else?

Sonicadvance1: With `1.13.1` it was pulling out 1.1.1 here, breaking the XCB version check I needed to add.

Sonicadvance1: Additional issue, one PC had version `1.14` which doesn't have a patch version, breaking the define generation as well.

Sonicadvance1: Found the difference between your output and mine. The regex breaks if their is no patch version in the version.
With the regex approach:
`1.14 -> 1.1.0`
`1.14.0 -> 1.14.0`

With the new approach:
`1.14 -> 1.14.0`
`1.14.0 -> 1.14.0`

PR #2783, opened 2023-07-16
Arm64: Optimize {Load,Store}ContextIndexed address generation

Sonicadvance1:
All of these IR operations were being fairly inefficient in their address calculation. All of these are known using power of 2 stride indexing. So all of these can be converted from three instructions to one.

These are always used for x87 stack accesses so each one gets an improvement.

Before:
```asm
0x0000ffff6a800248  d2800200    mov x0, #0x10
0x0000ffff6a80024c  9b007e80    mul x0, x20, x0
0x0000ffff6a800250  8b000380    add x0, x28, x0
0x0000ffff6a800254  fd417805    ldr d5, [x0, #752]
```

After:
```asm
0x0000ffff91e80240  8b141380    add x0, x28, x20, lsl #4
0x0000ffff91e80244  fd417805    ldr d5, [x0, #752]
```

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -10,6 +10,7 @@ $end_info$
 #include "Interface/Core/CPUID.h"
 #include "Interface/Core/JIT/Arm64/JITClass.h"
 #include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/Utils/MathUtils.h>
 
 namespace FEXCore::CPU {
 #define DEF_OP(x) void Arm64JITCore::Op_##x(IR::IROp_Header const *IROp, IR::NodeID Node)
@@ -676,10 +677,7 @@ DEF_OP(LoadContextIndexed) {
     case 2:
     case 4:
     case 8: {
-      LoadConstant(ARMEmitter::Size::i64Bit, TMP1, Op->Stride);
-      mul(ARMEmitter::Size::i64Bit, TMP1, Index, TMP1);
-      add(ARMEmitter::Size::i64Bit, TMP1, STATE, TMP1.R());
-
+      add(ARMEmitter::Size::i64Bit, TMP1, STATE, Index, FEXCore::ARMEmitter::ShiftType::LSL, FEXCore::ilog2(Op->Stride));
       const auto Dst = GetReg(Node);
       switch (OpSize) {
       case 1:
@@ -716,10 +714,7 @@ DEF_OP(LoadContextIndexed) {
     case 8:
     case 16:
     case 32: {
-      LoadConstant(ARMEmitter::Size::i64Bit, TMP1, Op->Stride);
-      mul(ARMEmitter::Size::i64Bit, TMP1, Index, TMP1);
-      add(ARMEmitter::Size::i64Bit, TMP1, STATE, TMP1.R());
-
+      add(ARMEmitter::Size::i64Bit, TMP1, STATE, Index, FEXCore::ARMEmitter::ShiftType::LSL, FEXCore::ilog2(Op->Stride));
       const auto Dst = GetVReg(Node);
 
       switch (OpSize) {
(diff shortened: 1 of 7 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -676,10 +676,7 @@ DEF_OP(LoadContextIndexed) {
     case 2:
     case 4:
     case 8: {
-      LoadConstant(ARMEmitter::Size::i64Bit, TMP1, Op->Stride);
-      mul(ARMEmitter::Size::i64Bit, TMP1, Index, TMP1);
-      add(ARMEmitter::Size::i64Bit, TMP1, STATE, TMP1.R());
-
+      add(ARMEmitter::Size::i64Bit, TMP1, STATE, Index, FEXCore::ARMEmitter::ShiftType::LSL, std::log2(Op->Stride));
```
alyssarosenzweig, line 679: If I'm not mistaken, std::log2 is a *float* logarithm. While this works, it might incur bad code gen if the compiler isn't able to see through this. It also raises all sorts of "questionable semantic" warning bells. We know the stride is a power-of-two, we really just want the `__builtin_ctz` of it.

Should we add a FEX utility function that (in debug builds) asserts the input is a power-of-two and (regardless) returns the `__builtin_ctz`? That way the log2 becomes a single op in release builds even if this doesn't get constant foldd away (and I am unsure if it would be).

alyssarosenzweig (changes requested):

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
alyssarosenzweig: `std::countr_zero` should be even better, actually.

Sonicadvance1: You're right. Created a helper and switched our usage over to it. Including a couple of lines that also used std::log2.
Added a unittest to ensure it returns expected data.

External/FEXCore/include/FEXCore/Utils/MathUtils.h
```diff
@@ -10,4 +14,14 @@ namespace FEXCore {
 [[nodiscard]] constexpr uint64_t AlignDown(uint64_t value, uint64_t size) {
   return value - value % size;
 }
+
+// Returns the ilog2 of a power-of-2 integer.
+// Asserts in the case that the passed in integer is not a power-of-2.
+template<typename T>
+requires(std::is_integral_v<T>)
+[[nodiscard]] constexpr T ilog2(T Value) {
+  LOGMAN_THROW_AA_FMT(std::popcount(Value) == 1, "ilog2 requires popcount to be one");
```
lioncash, line 23:
```suggestion
  LOGMAN_THROW_AA_FMT(std::has_single_bit(Value), "ilog2 requires popcount to be one");
```

External/FEXCore/include/FEXCore/Utils/MathUtils.h
```diff
@@ -10,4 +14,14 @@ namespace FEXCore {
 [[nodiscard]] constexpr uint64_t AlignDown(uint64_t value, uint64_t size) {
   return value - value % size;
 }
+
+// Returns the ilog2 of a power-of-2 integer.
+// Asserts in the case that the passed in integer is not a power-of-2.
+template<typename T>
+requires(std::is_integral_v<T>)
```
lioncash, line 21: perhaps this  should be clamped to only unsigned values with `std::is_unsigned_v`

External/FEXCore/include/FEXCore/Utils/MathUtils.h
Sonicadvance1: Resolved.

External/FEXCore/include/FEXCore/Utils/MathUtils.h
Sonicadvance1: Indeed, switched over to that.

alyssarosenzweig (approved):

PR #2785, opened 2023-07-17
ArmEmitter: Support 32-bit bitmask moves

Sonicadvance1:
Noticed this when inspecting some code that was moving constant `0x80808080` in to a register. Was using two move instructions when it could have used a single bitmask move.

This now checks to see if a constant can be 32-bit encoded in a logical bitmask move and uses that.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -192,6 +192,23 @@ void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, ui
     return;
   }
 
+  if ((Constant >> 32) == 0) {
+    // If the upper 32-bits is all zero, we can now switch to a 32-bit move.
+    s = ARMEmitter::Size::i32Bit;
+    Is64Bit = false;
+    Segments = 2;
+  }
+
+  // If this can be loaded with a mov bitmask.
+  const auto IsImm = vixl::aarch64::Assembler::IsImmLogical(Constant, RegSizeInBits(s));
+  if (IsImm) {
+    orr(s, Reg, ARMEmitter::Reg::zr, Constant);
+    if (NOPPad) {
+      nop(); nop(); nop();
+    }
+    return;
+  }
+
   int NumMoves = 1;
   int RequiredMoveSegments{};
 


External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -192,6 +192,23 @@ void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, ui
...
+    s = ARMEmitter::Size::i32Bit;
+    Is64Bit = false;
+    Segments = 2;
+  }
+
+  // If this can be loaded with a mov bitmask.
+  const auto IsImm = vixl::aarch64::Assembler::IsImmLogical(Constant, RegSizeInBits(s));
+  if (IsImm) {
+    orr(s, Reg, ARMEmitter::Reg::zr, Constant);
+    if (NOPPad) {
```
alyssarosenzweig, line 206: What's this about?

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -192,6 +192,23 @@ void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, ui
...
+    // If the upper 32-bits is all zero, we can now switch to a 32-bit move.
+    s = ARMEmitter::Size::i32Bit;
+    Is64Bit = false;
+    Segments = 2;
+  }
+
+  // If this can be loaded with a mov bitmask.
+  const auto IsImm = vixl::aarch64::Assembler::IsImmLogical(Constant, RegSizeInBits(s));
+  if (IsImm) {
+    orr(s, Reg, ARMEmitter::Reg::zr, Constant);
```
alyssarosenzweig, line 205: I assume arm64 could do "x | 0x80808080" in one op but FEX after this PR will generate two, and we would need an algebraic optimizer to fuse that?

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: For relocation support we will emit the worst case scenario of four instructions, just in-case we need to patch the relocation to something larger than a single instruction. Since relocations aren't commonly used, it doesn't happen frequently.

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: There is some code in the ConstProp pass to determine if it can determine merge logical imm encodings in to an InlineConstant op. So theoretically it should already do that. This just missed the simple case of something like `mov rax, 0x80808080`

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
alyssarosenzweig: Terrifying!

External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
alyssarosenzweig: Got it.

alyssarosenzweig (approved):

lioncash (approved):

PR #2787, opened 2023-07-17
Arm64: Only clear icache for code

Sonicadvance1:
Currently we're clearing icache including the data that lives on the tail of the block. Instead only clear the code that the was emitted and not tail data.

Additionally only disasm the code rather than all the tail data as well, as it gets unwieldy if viewing.


alyssarosenzweig (approved):
Thanks, this should make the vixl disassembly a lot nicer to work with.

PR #2788, opened 2023-07-17
FEXCore/Config: Adds support for enum mask configuration array

Sonicadvance1:
Allows us to consume an array of strings and convert it to an mask of enum values. This is a quality of life change that allows us to specify a mask of options.

The first configuration option added to support this is to control the vixl disassembler. Now by default the vixl disassembler doesn't disassemble any blocks and needs to be enabled individually.

eg:
```
FEXLoader --disassemble=blocks <args>
FEXLoader --disassemble=dispatcher <args>
FEXLoader --disassemble=blocks,dispatcher <args>
```

Has the additional convenience option of just passing in numbers as well.

```
FEXLoader --disassemble=2 <args>
FEXLoader --disassemble=1 <args>
FEXLoader --disassemble=3 <args>
```

Also of course all of this works through environment variables.
```
FEX_DISASSEMBLE=blocks FEXInterpreter <args>
FEX_DISASSEMBLE=dispatcher FEXInterpreter <args>
FEX_DISASSEMBLE=blocks,dispatcher FEXInterpreter <args>
```

While only used fairly sparingly now, this is likely to have some additional configurations using this in the future. Since we already have some configs that are basically using enums, but just by doing string comparisons.

This was asked for by a developer, so I figured I would throw it together quick.

### DIFF
--- a/External/FEXCore/Scripts/config_generator.py
+++ b/External/FEXCore/Scripts/config_generator.py
@@ -27,7 +27,9 @@ def print_header():
 #ifndef OPT_STRARRAY
 #define OPT_STRARRAY(group, enum, json, default) OPT_BASE(fextl::string, group, enum, json, default)
 #endif
-
+#ifndef OPT_STRENUM
+#define OPT_STRENUM(group, enum, json, default) OPT_BASE(uint64_t, group, enum, json, default)
+#endif
 '''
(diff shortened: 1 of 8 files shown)


alyssarosenzweig (approved):
Fantastic, thanks!

PR #2793, opened 2023-07-18
OpcodeDispatcher: Remove unused member variables and reorganize

lioncash:
Removes two member variables that were only assigned to, but never actually used, and moves all members into the private class section. Almost all of the members were only used internally anyway, and for the few accesses outside of the class, we can just add querying accessors for those.

Has the nice effect of putting most of the members in one place.


Sonicadvance1 (approved):
Nice cleanup!

PR #2794, opened 2023-07-18
OpcodeDispatcher: Handle VSIB byte

lioncash:
Ensures that we handle the AVX2 VSIB byte in a decent way.

As is, we can't compute the [index * scale] variant portion of the entire address operand, since the scale needs to act on every element of the vector after sign extension.

What we can do though, is compute the base address and add the displacement to it ahead of time though.


Sonicadvance1 (approved):
Very quirky problems.

PR #2795, opened 2023-07-19
Thunks: Remove weak symbol definitions

### LINKED ISSUE #2754: Canonical PPA build somehow breaking thunks
Behaviour when loading thunks. `Failed to load MakeHostTrampolineForGuestFunction from FEX executable`

Symbol of `MakeHostTrampolineForGuestFunction` in libX11-host when built from the Canonical PPA builders.
```
ryanh@ubuntu:/mnt/Work/Work/work/FEXNew/Build_Lenovo$ objdump -D /usr/lib/fex-emu/HostThunks/libX11-host.so | grep "MakeHostTrampolineForGuestFunction"
000000000002ac64 <_ZN7FEXCore34MakeHostTrampolineForGuestFunctionEPvmm>:
   31044:       97ffe708        bl      2ac64 <_ZN7FEXCore34MakeHostTrampolineForGuestFunctionEPvmm>
  ```
Symbol of `MakeHostTrampolineForGuestFunction` in libX11-host when built locally.
  ```
ryanh@ubuntu:/mnt/Work/Work/work/FEXNew/Build_Lenovo$ objdump -D HostLibs_64/libX11-host.so | grep "MakeHostTrampolineForGuestFunction"
000000000002a890 <_ZN7FEXCore34MakeHostTrampolineForGuestFunctionEPvmm>:
   31340:       94001a88        bl      37d60 <_ZN7FEXCore34MakeHostTrampolineForGuestFunctionEPvmm@plt>
   ```

Somehow this weak symbol is getting converted from a plt to a direct branch with the canonical PPA, breaking FEX's thunks in the process. `-Bsymbolic-functions` seems to be the only additional compile option that stands out but when manually adding it I couldn't reproduce this issue. Do we need to convert it to always be an undefined symbol rather than a weak symbol so this can never occur?

Sonicadvance1:
Fixes #2754

These panicking fallbacks are at times not ending up in as plt calls for some reason that I haven't been able to reproduce locally.

So far the only way I can reproduce is building with Canonical's PPA build system, since rebuilding locally didn't resolve the issue.

This will change the failure mode from these panicking asserts happening at call time, to dlopen failing during relocation when loading the thunk. Which LD_DEBUG=all can be used for debugging relocation failure in that case.

### DIFF
--- a/ThunkLibs/include/common/Host.h
+++ b/ThunkLibs/include/common/Host.h
@@ -16,23 +16,17 @@ $end_info$
 //
 // Note these are statically linked into the FEX executable. The linker hence
 // doesn't know about them when linking thunk libraries. This issue is avoided
-// by declaring the functions as weak symbols. Their implementation in this
-// file serves as a panicking fallback if matching symbols are not found.
+// by declaring the functions as weak symbols.
 namespace FEXCore {
   struct HostToGuestTrampolinePtr;
 
   __attribute__((weak))
   HostToGuestTrampolinePtr*
-  MakeHostTrampolineForGuestFunction(void* HostPacker, uintptr_t GuestTarget, uintptr_t GuestUnpacker) {
-    fprintf(stderr, "Failed to load %s from FEX executable\n", __FUNCTION__);
-    std::abort();
-  }
+  MakeHostTrampolineForGuestFunction(void* HostPacker, uintptr_t GuestTarget, uintptr_t GuestUnpacker);
+
   __attribute__((weak))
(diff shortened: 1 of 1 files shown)


alyssarosenzweig (approved):

Sonicadvance1:
Tested `always_inline` and that also ignores the weak attribute semantics. So it really looks like a bug with clang ignoring weak symbol semantics in the case that it thinks it can direct link to the symbol, or inlining it. So the only way to be safe is to remove the function body.

PR #2796, opened 2023-07-19
Arm64/Emitter: Deduplicate some opcode values

lioncash:
For the unary and unpredicated bitwise ops, we can move the opcode into the implementation function as well as asserts to deduplicate some code.


Sonicadvance1 (approved):
Thanks for cleaning up my mess!

PR #2798, opened 2023-07-19
Jit: Add block links directly through the lookup cache on thread exit

bylaws:
Prevents the code invalidation mutex from being locked as shared recursively, since it is locked before entering ThreadExitFunctionLink and would end up being locked again by ThreadAddBlockLink.

This fixes a deadlock on Windows.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
@@ -384,7 +384,7 @@ static uint64_t X86JITCore_ExitFunctionLink(FEXCore::Core::CpuStateFrame *Frame,
   }
 
   auto LinkerAddress = Frame->Pointers.Common.ExitFunctionLinker;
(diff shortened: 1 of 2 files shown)


External/FEXCore/Source/Interface/Core/JIT/x86_64/JIT.cpp
```diff
@@ -384,7 +384,7 @@ static uint64_t X86JITCore_ExitFunctionLink(FEXCore::Core::CpuStateFrame *Frame,
   }
 
   auto LinkerAddress = Frame->Pointers.Common.ExitFunctionLinker;
-  Context::ContextImpl::ThreadAddBlockLink(Thread, GuestRip, (uintptr_t)record, [record, LinkerAddress]{
+  Thread->LookupCache->AddBlockLink(Thread, GuestRip, (uintptr_t)record, [record, LinkerAddress]{
```
Sonicadvance1, line 387: Looks like you forgot to remove the first argument here.

PR #2799, opened 2023-07-19
Arm64/Emitter: Handle SVE XAR

lioncash:
Now that we have the helper for encoding immediate shifts, we can trivially implement the remaining missing instruction in the bitwise logical unpredicated group.


Sonicadvance1 (approved):
What a silly SHA3 instruction. I love it.

PR #2800, opened 2023-07-19
Arm64/Emitter: Handle SVE FP arithmetic with immediate (predicated) group

lioncash:
Pretty straightforward to handle


Sonicadvance1 (approved):
Interesting encoding limitations for the immediate.

PR #2815, opened 2023-07-24
Draft: Use the host NZCV register

alyssarosenzweig:
This is a very early WIP, not ready for merge at all, but I want to get this on people's radars since it is showing signs of life.

The overall strategy I'm taking is to model the packed 32-bit NZCV register as a SSA value (i.e. pretending it's a GPR). All accesses to SF/ZF/CF/OF become accesses to the packed 32-bit register. As a default path, `OpcodeDispatcher.h` learns to do bitfield insert/extract for these flags. During the block this is treated as a GPR, with msr/mrs to the "real" register only at block boundaries.

That regresses code gen, since now we have piles of extra packing/unpacking emitted. We can get some perf back by rewriting the flag generation code to generate a "synthetic" NZCV register directly, rather than doing 1-bit RMW cycles on the existing values.

The real value comes when we start introducing real arm64 opcodes that operate on NZCV. At this point, I've added support (only) for TST, which perfectly translates the x86 flag behaviour of logical ops. Again this is all modelled as pure SSA. Logically, our IR's "test" opcode returns a 32-bit GPR that happens to match the encoding of the arm64 register -- it could be implemented (less efficiently) with bit operations, e.g. for a risc-v backend.

At this point, `llvm-mca` thinks this is a cycle count win (if not instruction count) for simple logical ops and adds/subs.

The next step is to actually teach the register allocator about these flags, so it can allocate the destination of a TST directly to the NZCV, and then insert an mrs only if it would be get read/clobbered. This will let us delete pointless moves between NZCV<-->GPR. I don't think this PR will get to that, since it's not required to get wins from this and this PR is already big and churny.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -690,6 +690,23 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
       }
     break;
     }
+    case OP_TEST: {
+      auto Op = IROp->CW<IR::IROp_Test>();
+      uint64_t Constant1{};
+      uint64_t Constant2{};
+
+      if (IREmit->IsValueConstant(Op->Header.Args[0], &Constant1) &&
+          IREmit->IsValueConstant(Op->Header.Args[1], &Constant2)) {
+        uint64_t Result = (Constant1 & Constant2) & getMask(Op);
+        bool N = Result & (1ull << ((Op->Header.Size * 8) - 1));
+        bool Z = Result == 0;
+        uint32_t NZVC = (N ? (1 << 31) : 0) | (Z ? (1 << 30) : 0);
+
+        IREmit->ReplaceWithConstant(CodeNode, NZVC);
+        Changed = true;
+      }
+    break;
+    }
     case OP_OR: {
       auto Op = IROp->CW<IR::IROp_Or>();
       uint64_t Constant1{};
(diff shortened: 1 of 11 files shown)


External/FEXCore/Scripts/json_ir_generator.py
Sonicadvance1, line 1: This commit can be in its own PR. I'm hitting it locally as well.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -690,6 +690,23 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
...
+      auto Op = IROp->CW<IR::IROp_Test>();
+      uint64_t Constant1{};
+      uint64_t Constant2{};
+
+      if (IREmit->IsValueConstant(Op->Header.Args[0], &Constant1) &&
+          IREmit->IsValueConstant(Op->Header.Args[1], &Constant2)) {
+        uint64_t Result = (Constant1 & Constant2) & getMask(Op);
+        bool N = Result & (1ull << ((Op->Header.Size * 8) - 1));
+        bool Z = Result == 0;
+        uint32_t NZVC = (N ? (1 << 31) : 0) | (Z ? (1 << 30) : 0);
```
lioncash, line 703:
```suggestion
        uint32_t NZVC = (N ? (1U << 31) : 0) | (Z ? (1U << 30) : 0);
```

otherwise some checks in ubsan will complain about this (specifically the `1 << 31` shift)

PR #2817, opened 2023-07-24
OpcodeDispatcher: Optimize PSAD* to use vuabdl{2,} 

### LINKED ISSUE #443: PSAD could probably be implemented with NEON SABDL
Haven't looked too closely, but would speed that up a bit if it works.

Sonicadvance1:
Optimizes this instruction from 18 down to 10 instructions.

Fixes #443

Before:
```asm
0x00007fffe2080908  10ffffe0            adr x0, #-0x4 (addr 0x7fffe2080904)
0x00007fffe208090c  f9005f80            str x0, [x28, #184]
0x00007fffe2080910  3dc03cc4            ldr q4, [x6, #240]
0x00007fffe2080914  2f08a705            uxtl v5.8h, v24.8b
0x00007fffe2080918  6f08a706            uxtl2 v6.8h, v24.16b
0x00007fffe208091c  2f08a487            uxtl v7.8h, v4.8b
0x00007fffe2080920  6f08a484            uxtl2 v4.8h, v4.16b
0x00007fffe2080924  6e6784a5            sub v5.8h, v5.8h, v7.8h
0x00007fffe2080928  6e6484c4            sub v4.8h, v6.8h, v4.8h
0x00007fffe208092c  4e60b8a5            abs v5.8h, v5.8h
0x00007fffe2080930  4e60b884            abs v4.8h, v4.8h
0x00007fffe2080934  4e71b8a5            addv h5, v5.8h
0x00007fffe2080938  4e71b884            addv h4, v4.8h
0x00007fffe208093c  4ea51ca0            mov v0.16b, v5.16b
0x00007fffe2080940  6e180480            mov v0.d[1], v4.d[0]
0x00007fffe2080944  4ea01c18            mov v24.16b, v0.16b
0x00007fffe2080948  58000040            ldr x0, pc+8 (addr 0x7fffe2080950)
0x00007fffe208094c  d63f0000            blr x0
```

After:
```asm
0x00007fffe2080808  10ffffe0            adr x0, #-0x4 (addr 0x7fffe2080804)
0x00007fffe208080c  f9005f80            str x0, [x28, #184]
0x00007fffe2080810  3dc03cc4            ldr q4, [x6, #240]
0x00007fffe2080814  2e247305            uabdl v5.8h, v24.8b, v4.8b
0x00007fffe2080818  6e247304            uabdl2 v4.8h, v24.16b, v4.16b
0x00007fffe208081c  4e71b8a5            addv h5, v5.8h
0x00007fffe2080820  4e71b884            addv h4, v4.8h
0x00007fffe2080824  4ec438b8            zip1 v24.2d, v5.2d, v4.2d
0x00007fffe2080828  58000040            ldr x0, pc+8 (addr 0x7fffe2080830)
0x00007fffe208082c  d63f0000            blr x0
```

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -3027,6 +3027,37 @@ DEF_OP(VUABDL) {
   }
 }
 
+DEF_OP(VUABDL2) {
+  const auto Op = IROp->C<IR::IROp_VUABDL2>();
+  const auto OpSize = IROp->Size;
+
+  const auto ElementSize = Op->Header.ElementSize;
+  const auto Is256Bit = OpSize == Core::CPUState::XMM_AVX_REG_SIZE;
+
+  const auto Dst = GetVReg(Node);
+  const auto Vector1 = GetVReg(Op->Vector1.ID());
+  const auto Vector2 = GetVReg(Op->Vector2.ID());
+
+  LOGMAN_THROW_AA_FMT(ElementSize == 2 || ElementSize == 4 || ElementSize == 8, "Invalid size");
+  const auto SubRegSize =
+    ElementSize == 2 ? ARMEmitter::SubRegSize::i16Bit :
+    ElementSize == 4 ? ARMEmitter::SubRegSize::i32Bit :
+    ElementSize == 8 ? ARMEmitter::SubRegSize::i64Bit : ARMEmitter::SubRegSize::i8Bit;
+
+  if (HostSupportsSVE && Is256Bit) {
+    // To mimic the behavior of AdvSIMD UABDL, we need to get the
+    // absolute difference of the even elements (UADBLB), get the
+    // absolute difference of the odd elemenets (UABDLT), then
+    // interleave the results in both vectors together.
+
+    uabdlb(SubRegSize, VTMP1.Z(), Vector1.Z(), Vector2.Z());
+    uabdlt(SubRegSize, VTMP2.Z(), Vector1.Z(), Vector2.Z());
+    zip2(SubRegSize, Dst.Z(), VTMP1.Z(), VTMP2.Z());
+  } else {
+    uabdl2(SubRegSize, Dst.Q(), Vector1.Q(), Vector2.Q());
+  }
+}
+
 DEF_OP(VTBL1) {
   const auto Op = IROp->C<IR::IROp_VTBL1>();
   const auto OpSize = IROp->Size;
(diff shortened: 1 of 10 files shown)


Sonicadvance1:
> Looks great! It shouldn't make a difference, but I figured I'd mention that `mov v0.d[1], v4.d[0]` is equivalent to `UZP1.2D` or `TRN1.2D`, in case having a separate destination operand would make it easier to address the extraneous moves.

You're right! I wasn't paying close enough attention, and we can definitely zip these up which our RA understands a lot better.

From 18 instructions down to 10!

dougallj:
Nice!

(And for SVE, I'd maybe aim for something like:

```
UABDLB Ztmp1.H, Zin0.B, Zin1.B
UABDLT Ztmp2.H, Zin0.B, Zin1.B
ADD Ztmp1.H, Ztmp1.H, Ztmp2.H
DUP Ztmp2.H, #1
DUP Zout.D, #0
UDOT Zout.D, Ztmp1.H, Ztmp2.H
```

But that probably shouldn't block this PR, I'm just enjoying the excuse to think about SVE.)

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -3027,6 +3027,37 @@ DEF_OP(VUABDL) {
...
+
+  if (HostSupportsSVE && Is256Bit) {
+    // To mimic the behavior of AdvSIMD UABDL, we need to get the
+    // absolute difference of the even elements (UADBLB), get the
+    // absolute difference of the odd elemenets (UABDLT), then
+    // interleave the results in both vectors together.
+
+    uabdlb(SubRegSize, VTMP1.Z(), Vector1.Z(), Vector2.Z());
+    uabdlt(SubRegSize, VTMP2.Z(), Vector1.Z(), Vector2.Z());
+    zip2(SubRegSize, Dst.Z(), VTMP1.Z(), VTMP2.Z());
```
hthh, line 3055: Oh, actually, what's the plan with regards to 512-bit SVE (or wider)? On 512-bit, zip2 would take bits 256 to 512, when you really want bits 128 to 256, right?

hthh: Yeah... My concern is that, if the plan is to ignore the high half of the registers, zip2 won't work, because it *only* reads from the high halves of its input registers (in this case the result of uabdlt/b, based on the to-be-ignored high-halves of the input registers).

Sonicadvance1: Good catch, we'll have to revisit this once there are 512-bit SVE2 CPUs on the market. We can end up testing using the vixl simulator in the future.
This won't be the only IR operation that is broken because of this.

lioncash (approved):

PR #2820, opened 2023-07-25
Preserve PF across zero shift

alyssarosenzweig:


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1158,8 +1158,8 @@ friend class FEXCore::IR::PassManager;
    * @name These functions are used by the deferred flag handling while it is calculating and storing flags in to RFLAGs.
    * @{ */
   OrderedNode *LoadPF();
-  void CalculatePFUncheckedABI(OrderedNode *Res);
-  void CalculatePF(OrderedNode *Res);
+  void CalculatePFUncheckedABI(OrderedNode *Res, OrderedNode *condition = NULL);
```
Sonicadvance1, line 1161: `nullptr` instead of `NULL` here and the next line. C++-isms.

Sonicadvance1 (approved):

PR #2823, opened 2023-07-26
Mingw: Fixes compiling again

Sonicadvance1:
Getting the CI machines setup to handle this so we can stop breaking it.

### DIFF
--- a/Source/Tools/FEXLoader/WindowsDummyHandlers.cpp
+++ b/Source/Tools/FEXLoader/WindowsDummyHandlers.cpp
@@ -1,0 +1,27 @@
+#ifdef _WIN32
+#include "WindowsDummyHandlers.h"
+
+namespace FEX::WindowsHandlers {
+  thread_local FEXCore::Core::InternalThreadState *TLSThread;
+
+  void DummySignalDelegator::RegisterTLSState(FEXCore::Core::InternalThreadState *Thread) {
+    TLSThread = Thread;
+  }
+
+  void DummySignalDelegator::UninstallTLSState(FEXCore::Core::InternalThreadState *Thread) {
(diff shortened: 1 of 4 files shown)


lioncash (approved):

Source/Tools/FEXLoader/WindowsDummyHandlers.cpp
```diff
@@ -0,0 +1,27 @@
+#ifdef _WIN32
+#include "WindowsDummyHandlers.h"
```
bylaws, line 2: Did you miss adding this file?

Sonicadvance1: Yes I did. I thought this was already there but I had missed it on my first commit of mingw support.

Source/Tools/FEXLoader/WindowsDummyHandlers.h
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  FEXCore::HLE::SyscallABI GetSyscallABI(uint64_t Syscall) override {
+    // Don't do anything
+    return {0, false, 0};
+  }
+
+  // These are no-ops implementations of the SyscallHandler API
+  std::shared_mutex StubMutex;
+  FEXCore::HLE::AOTIRCacheEntryLookupResult LookupAOTIRCacheEntry(FEXCore::Core::InternalThreadState *Thread, uint64_t GuestAddr) override {
```
bylaws, line 22: You can drop this mutex now the LookupAOTIRCacheEntry return type has changed

Sonicadvance1: Indeed. Removed.

PR #2829, opened 2023-07-28
Github: Adds mingw build test workflow

Sonicadvance1:
Currently only does a build, doing a CI run means figuring out why TestHarnessRunner doesn't find libraries correctly.

I want to ensure we don't break building at least while sorting out the rest of this.

### DIFF
--- a/.github/workflows/mingw_build.yml
+++ b/.github/workflows/mingw_build.yml
@@ -1,0 +1,96 @@
+name: Mingw build
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
+
+env:
+  BUILD_TYPE: Debug
+  FEX_ENABLEAVX: 1
+
+jobs:
+  build:
+    runs-on: ${{ matrix.arch }}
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
`buildbot-macmini-2` has been removed from mingw supporting status for now. Something is wrong with it seemingly due to running Ubuntu 20.04

Sonicadvance1:
Turns out this is an outdated cmake version issue.
macmini-2 has cmake version: `cmake version 3.16.3`
macmini-1 has cmake version: `cmake version 3.22.1`
Bisecting to figure out what we need to bump the minimum up to...

Sonicadvance1:
Couldn't reproduce locally by using 3.16.3, maybe Canonical had some patches on top of that version or something.
Just updated the macmini-2 runner's cmake version which resolved this issue.
Reenabled as a mingw runner.

Sonicadvance1:
Rebased to make sure this was still working for the merge.

lioncash (approved):

PR #2831, opened 2023-07-28
Arm64/Emitter: Deduplicate some more SVE implementations

lioncash:
Just moves some base opcodes and asserts into one place. Follow up to #2824


Sonicadvance1:
Looks like you got bit by the race condition that #2830 fixes.

Sonicadvance1 (approved):

PR #2832, opened 2023-07-28
Pack NZCV flags

alyssarosenzweig:
This is the first piece of the NZCV flags puzzle. Instead of having 4 distinct bytes for the four SF/ZF/CF/OF flags, we have a single packed 32-bit word, with SF in bit 31, ZF in bit 30, CF in bit 29, and OF in bit 28. This layout matches the structure of arm64's native NZCV register (semantics diverge with the inverted carry flag, though). That means we can calculate the packed flags with the help of arm64 flag setting registers. In other cases we can calculate this packed format in software, but still end up ahead (compared to 4 `strb`s) with clever use of bitwise operations. In particular, arm's shift-then-OR instructions are invaluable for this.

This is showing instruction count improvements on some synthetic unit tests I'm looking at. I haven't benchmarked it. The really fruitful gains will come later when we teach RA about the host flags, reducing the load/store traffic still incurred here. This PR is a step in that direction: it gets the code in place to calculate flags, the later work will just be about how those calculated values get stored.

There's also a lot of room to calculate flag values more efficiently, especially when FlagM is present, by using instructions like `adds` and `cfinv` but that's for later work.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
@@ -190,19 +190,31 @@ DEF_OP(LoadFlag) {
   uintptr_t ContextPtr = reinterpret_cast<uintptr_t>(Data->State->CurrentFrame);
   ContextPtr += offsetof(FEXCore::Core::CPUState, flags[0]);
   ContextPtr += Op->Flag;
-  uint8_t const *MemData = reinterpret_cast<uint8_t const*>(ContextPtr);
-  GD = *MemData;
+
+  if (Op->Flag == 24 /* NZCV */) {
+    uint32_t const *MemData = reinterpret_cast<uint32_t const*>(ContextPtr);
+    GD = *MemData;
+  } else {
+    uint8_t const *MemData = reinterpret_cast<uint8_t const*>(ContextPtr);
+    GD = *MemData;
+  }
 }
 
 DEF_OP(StoreFlag) {
   auto Op = IROp->C<IR::IROp_StoreFlag>();
-  uint8_t Arg = *GetSrc<uint8_t*>(Data->SSAData, Op->Value);
+  uint32_t Arg = *GetSrc<uint32_t*>(Data->SSAData, Op->Value);
 
   uintptr_t ContextPtr = reinterpret_cast<uintptr_t>(Data->State->CurrentFrame);
   ContextPtr += offsetof(FEXCore::Core::CPUState, flags[0]);
   ContextPtr += Op->Flag;
-  uint8_t *MemData = reinterpret_cast<uint8_t*>(ContextPtr);
-  *MemData = Arg;
+
+  if (Op->Flag == 24 /* NZCV */) {
+    uint32_t *MemData = reinterpret_cast<uint32_t*>(ContextPtr);
+    *MemData = Arg;
+  } else {
+    uint8_t *MemData = reinterpret_cast<uint8_t*>(ContextPtr);
+    *MemData = Arg;
+  }
 }
 
 DEF_OP(LoadMem) {
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -331,6 +331,23 @@ DEF_OP(Or) {
   }
 }
 
+DEF_OP(Orlshl) {
+  auto Op = IROp->C<IR::IROp_Orlshl>();
+  const uint8_t OpSize = IROp->Size;
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+  const auto Src1 = GetReg(Op->Src1.ID());
+
+  uint64_t Const;
+  if (IsInlineConstant(Op->Src2, &Const)) {
+    orr(EmitSize, Dst, Src1, Const << Op->BitShift);
+  } else {
+    const auto Src2 = GetReg(Op->Src2.ID());
+    orr(EmitSize, Dst, Src1, Src2, ARMEmitter::ShiftType::LSL, Op->BitShift);
+  }
+}
+
 DEF_OP(And) {
   auto Op = IROp->C<IR::IROp_And>();
   const uint8_t OpSize = IROp->Size;
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -348,6 +348,23 @@ DEF_OP(Orlshl) {
   }
 }
 
+DEF_OP(Orlshr) {
+  auto Op = IROp->C<IR::IROp_Orlshr>();
+  const uint8_t OpSize = IROp->Size;
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+  const auto Src1 = GetReg(Op->Src1.ID());
+
+  uint64_t Const;
+  if (IsInlineConstant(Op->Src2, &Const)) {
+    orr(EmitSize, Dst, Src1, Const >> Op->BitShift);
+  } else {
+    const auto Src2 = GetReg(Op->Src2.ID());
+    orr(EmitSize, Dst, Src1, Src2, ARMEmitter::ShiftType::LSR, Op->BitShift);
+  }
+}
+
 DEF_OP(And) {
   auto Op = IROp->C<IR::IROp_And>();
   const uint8_t OpSize = IROp->Size;
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -84,6 +84,21 @@ DEF_OP(Add) {
   }
 }
 
+DEF_OP(TestNZ) {
+  auto Op = IROp->C<IR::IROp_TestNZ>();
+  const uint8_t OpSize = Op->Size;
+
+  LOGMAN_THROW_AA_FMT(OpSize == 4 || OpSize == 8, "Unsupported {} size: {}", __func__, OpSize);
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+  const auto ZeroReg = ARMEmitter::Reg::zr;
+  cmn(EmitSize, GetReg(Op->Src1.ID()), ZeroReg);
+
+  // TODO: Optimize this out
+  mrs(Dst, ARMEmitter::SystemRegister::NZCV);
+}
+
 DEF_OP(Sub) {
   auto Op = IROp->C<IR::IROp_Sub>();
   const uint8_t OpSize = IROp->Size;
--- a/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -145,6 +145,7 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
   if (Op->TableInfo->Flags & X86Tables::InstFlags::FLAGS_BLOCK_END) {
     // RIP could have been updated after coming back from the Syscall.
     NewRIP = _LoadContext(GPRSize, GPRClass, offsetof(FEXCore::Core::CPUState, rip));
+    CalculateDeferredFlags();
     _ExitFunction(NewRIP);
   }
 }
@@ -178,6 +179,7 @@ void OpDispatchBuilder::ThunkOp(OpcodeArgs) {
 
   // Store the new stack pointer
   StoreGPRRegister(X86State::REG_RSP, NewSP);
+  CalculateDeferredFlags();
 
   // Store the new RIP
   _ExitFunction(NewRIP);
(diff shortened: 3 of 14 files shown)


External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1040,19 +1060,105 @@ friend class FEXCore::IR::PassManager;
...
+      auto SelectOp = _Select(FEXCore::IR::COND_EQ, Res, Zero, One, Zero);
+      SetRFLAG<FEXCore::X86State::RFLAG_ZF_LOC>(SelectOp);
+    }
+  }
+
+  OrderedNode *InsertNZCV(OrderedNode *NZCV, unsigned BitOffset, OrderedNode *Value) {
+    unsigned Bit = IndexNZCV(BitOffset);
+
+    uint32_t SetBits = PossiblySetNZCVBits;
+    PossiblySetNZCVBits |= (1 << Bit);
```
lioncash, line 1129:
```suggestion
    PossiblySetNZCVBits |= (1U << Bit);
```

since `Bit` can be 31

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1040,19 +1060,105 @@ friend class FEXCore::IR::PassManager;
...
+
+    if (SetBits == 0)
+      return _Lshl(Value, _Constant(Bit));
+    else if (CTX->BackendFeatures.SupportsShiftedBitwise && (SetBits & (1 << Bit)) == 0)
```
lioncash, line 1133:
```suggestion
    else if (CTX->BackendFeatures.SupportsShiftedBitwise && (SetBits & (1U << Bit)) == 0)
```

ditto

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -138,6 +173,10 @@ void OpDispatchBuilder::CalculatePF(OrderedNode *Res, OrderedNode *condition) {
 void OpDispatchBuilder::CalculateDeferredFlags(uint32_t FlagsToCalculateMask) {
   if (CurrentDeferredFlags.Type == FlagsGenerationType::TYPE_NONE) {
     // Nothing to do
+    if (CachedNZCV)
+      _StoreFlag(CachedNZCV, FEXCore::X86State::RFLAG_NZCV_LOC);
+
+    CachedNZCV = NULL;
```
lioncash, line 179:
```suggestion
    CachedNZCV = nullptr;
```

External/FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -313,6 +352,11 @@ void OpDispatchBuilder::CalculateDeferredFlags(uint32_t FlagsToCalculateMask) {
 
   // Done calculating
   CurrentDeferredFlags.Type = FlagsGenerationType::TYPE_NONE;
+
+  if (CachedNZCV)
+    _StoreFlag(CachedNZCV, FEXCore::X86State::RFLAG_NZCV_LOC);
+
+  CachedNZCV = NULL;
```
lioncash, line 359:
```suggestion
  CachedNZCV = nullptr;
```

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -690,6 +690,20 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
...
     break;
     }
+    case OP_TESTNZ: {
+      auto Op = IROp->CW<IR::IROp_TestNZ>();
+      uint64_t Constant1{};
+
+      if (IREmit->IsValueConstant(Op->Header.Args[0], &Constant1)) {
+        bool N = Constant1 & (1ull << ((Op->Size * 8) - 1));
+        bool Z = Constant1 == 0;
+        uint32_t NZVC = (N ? (1 << 31) : 0) | (Z ? (1 << 30) : 0);
```
lioncash, line 700:
```suggestion
        uint32_t NZVC = (N ? (1U << 31) : 0) | (Z ? (1U << 30) : 0);
```

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -82,6 +82,10 @@ friend class FEXCore::IR::PassManager;
 
   void StartNewBlock() {
     flagsOp = SelectionFlag::Nothing;
+
+    // If we loaded flags but didn't change them, invalidate the cached copy and move on.
+    // Changes get stored out by CalculateDeferredFlags.
+    CachedNZCV = NULL;
```
lioncash, line 88:
```suggestion
    CachedNZCV = nullptr;
```

External/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
```diff
@@ -190,19 +190,31 @@ DEF_OP(LoadFlag) {
...
+  } else {
+    uint8_t const *MemData = reinterpret_cast<uint8_t const*>(ContextPtr);
+    GD = *MemData;
+  }
 }
 
 DEF_OP(StoreFlag) {
   auto Op = IROp->C<IR::IROp_StoreFlag>();
-  uint8_t Arg = *GetSrc<uint8_t*>(Data->SSAData, Op->Value);
+  uint32_t Arg = *GetSrc<uint8_t*>(Data->SSAData, Op->Value);
```
Sonicadvance1, line 205: `uint32_t Arg = *GetSrc<uint8_t*>(Data->SSAData, Op->Value);`
should be
`uint32_t Arg = *GetSrc<uint32_t*>(Data->SSAData, Op->Value);`

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1040,19 +1060,105 @@ friend class FEXCore::IR::PassManager;
...
+    CachedNZCV = Value;
+  }
+
+  void ZeroNZCV() {
+    CachedNZCV = _Constant(0);
+    PossiblySetNZCVBits = 0;
+  }
+
+  void SetN_ZeroZCV(unsigned SrcSize, OrderedNode *Res) {
+    assert(IndexNZCV(FEXCore::X86State::RFLAG_SF_LOC) == 31);
```
lioncash, line 1094:
```suggestion
    LOGMAN_THROW_A_FMT(IndexNZCV(FEXCore::X86State::RFLAG_SF_LOC) == 31, "Insert funny message here");
```

Just so it's using our assertion interface

lioncash (commented):
Just a few minor nits on a first pass

External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -84,6 +84,21 @@ DEF_OP(Add) {
...
+DEF_OP(TestNZ) {
+  auto Op = IROp->C<IR::IROp_TestNZ>();
+  const uint8_t OpSize = Op->Size;
+
+  LOGMAN_THROW_AA_FMT(OpSize == 4 || OpSize == 8, "Unsupported {} size: {}", __func__, OpSize);
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+  const auto ZeroReg = ARMEmitter::Reg::zr;
+  adds(EmitSize, ZeroReg, GetReg(Op->Src1.ID()), ZeroReg);
```
Sonicadvance1, line 96: We should add the `adds` alias for `cmn` in to our emitter.
I missed a bunch of aliases when writing it.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
alyssarosenzweig: What's.. what's wrong with assert?

Sonicadvance1: The logman assert interface will actually log the message as to why it failed to FEXServer so we can see why it failed outside of having running through gdb. Which is a nicety.
Also assert doesn't actually fail, because the guest application might capture the signal and try to keep running. Logman will execute the `AssertAndDie` routine that makes it know that FEX /really/ /really/ wants to die.

Sonicadvance1: Although in this case, you could likely change the function signature to be `static inline constexpr unsigned IndexNZCV(unsigned BitOffset)` and then after the function declaration just use `static_asssert` with each of the types, that way it gets enforced at compile time.

alyssarosenzweig: I like that better ^^

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1040,19 +1060,105 @@ friend class FEXCore::IR::PassManager;
   [[nodiscard]] uint32_t GetDstBitSize(X86Tables::DecodedOp Op) const;
   [[nodiscard]] uint32_t GetSrcBitSize(X86Tables::DecodedOp Op) const;
 
+  static inline constexpr unsigned IndexNZCV(unsigned BitOffset) {
+    switch (BitOffset) {
+      case FEXCore::X86State::RFLAG_OF_LOC: return 28;
+      case FEXCore::X86State::RFLAG_CF_LOC: return 29;
+      case FEXCore::X86State::RFLAG_ZF_LOC: return 30;
+      case FEXCore::X86State::RFLAG_SF_LOC: return 31;
+      default: assert(0);
```
Sonicadvance1, line 1069: This should be `FEX_UNREACHABLE` instead of assert, to remove the compiler warning.

Sonicadvance1:
Looks like the issue with unit tests is something specific with 32-bit applications. 64-bit applications in my testing were fine.
Looks like whatever is breaking is happening early in the glibc dynamic linker for 32-bit applications, so it doesn't even get in to loading the real application.
`./Bin/FEXInterpreter /lib/ld-linux.so.2 --help ` is enough to see that the linker's help text is corrupted.
`Usage: /li [OPTION]... EXECUTABLE-FILE [ARGS-FOR-PROGRAM...]`
versus
`Usage: /lib/ld-linux.so.2 [OPTION]... EXECUTABLE-FILE [ARGS-FOR-PROGRAM...] `

alyssarosenzweig:
> Looks like the issue with unit tests is something specific with 32-bit applications. 64-bit applications in my testing were fine. Looks like whatever is breaking is happening early in the glibc dynamic linker for 32-bit applications, so it doesn't even get in to loading the real application. `./Bin/FEXInterpreter /lib/ld-linux.so.2 --help ` is enough to see that the linker's help text is corrupted. `Usage: /li [OPTION]... EXECUTABLE-FILE [ARGS-FOR-PROGRAM...]` versus `Usage: /lib/ld-linux.so.2 [OPTION]... EXECUTABLE-FILE [ARGS-FOR-PROGRAM...] `

Thanks, hopefully I can reproduce the issue and diagnose then.

alyssarosenzweig:
the --help is right locally, uff.

alyssarosenzweig:
The regression was due to inadvertently zeroing the CF when `UpdateCF` was false. Fixed by explicitly preserving it, at the cost of extra instructions. We'll come out ahead once we take advantage of FlagM, though, this isn't a showstopper. Specifically, "set NZ according to the input, zero V, and preserve C" can be expressed as TST followed by RMIF.

External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -331,6 +346,40 @@ DEF_OP(Or) {
...
+DEF_OP(Orlshl) {
+  auto Op = IROp->C<IR::IROp_Orlshl>();
+  const uint8_t OpSize = IROp->Size;
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+  const auto Src1 = GetReg(Op->Src1.ID());
+
+  uint64_t Const;
+  if (IsInlineConstant(Op->Src2, &Const)) {
```
Sonicadvance1, line 358: I don't believe this InlineConstant path in orlshl and orlshr will ever trigger, Since you added constprop support, but not inline constant support.
Probably want to look at ConstProp.cpp and the locations that use the `OP_OR` operation and then do `CreateInlineConstant` after it. Roughly on line 1049 in that file.

External/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -145,6 +145,7 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
   if (Op->TableInfo->Flags & X86Tables::InstFlags::FLAGS_BLOCK_END) {
     // RIP could have been updated after coming back from the Syscall.
     NewRIP = _LoadContext(GPRSize, GPRClass, offsetof(FEXCore::Core::CPUState, rip));
+    CalculateDeferredFlags();
```
Sonicadvance1, line 148: Are all the added `CalculateDeferredFlags` calls in this file necessary? I've noticed a few end up being redundant. Which would be fine but I want to make sure that they weren't thrown around during debugging and no longer necessary.

alyssarosenzweig: Probably some could be removed, but the intention is to call `CalculateDeferredFlags` on any block boundary, since if you try to preserve `CachedNZCV` across block boundaries you're in for extreme badness. Calling `CalculateDeferredFlags` before things like function ends makes all those tricky bugs go away. The no-op calls should be cheap so I'm inclined to leave them in. And someday refactor this mess to make "add a new block" properly factored.

External/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
alyssarosenzweig: So far I can't trigger any case that would inline constants either so meh.

alyssarosenzweig: Yeah, I think for now this may be unreachable. OTOH I don't see a good reason to remove it because this could easily become useful later.

Sonicadvance1 (approved):
Looks good. I did some stability testing on this yesterday and didn't see any regressions for a quick run.

PR #2843, opened 2023-07-30
Linux: Implement {recv,send}mmsg inside of socketcall

Sonicadvance1:
These are the last two missing socketcall operations. Implementing them is mostly using the other implementation just in a helper function.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/x32/Socket.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/x32/Socket.cpp
@@ -315,6 +315,94 @@ namespace FEX::HLE::x32 {
     SYSCALL_ERRNO();
   }
 
+  static uint64_t SendMMsg(int sockfd, compat_ptr<mmsghdr_32> msgvec, uint32_t vlen, int flags) {
+    fextl::vector<iovec> Host_iovec;
+    fextl::vector<struct mmsghdr> HostMmsg(vlen);
+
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/x32/Socket.cpp
```diff
@@ -315,6 +315,94 @@ namespace FEX::HLE::x32 {
...
 
+  static uint64_t SendMMsg(int sockfd, compat_ptr<mmsghdr_32> msgvec, uint32_t vlen, int flags) {
+    fextl::vector<iovec> Host_iovec;
+    fextl::vector<struct mmsghdr> HostMmsg(vlen);
+
+    // Walk the iovec and convert them
+    // Calculate controllen at the same time
+    size_t Controllen_size{};
+    for (size_t i = 0; i < vlen; ++i) {
+      msghdr32 &guest = msgvec[i].msg_hdr;
```
lioncash, line 326: Should we be handling if user code does something silly like pass a null pointer in with a non-zero vlen size? Ditto with RecvMMsg

Sonicadvance1: This falls under the same problem space as if they passed in an invalid pointer entirely. FEX doesn't handle `EFAULT` at all currently which will take a...lot of work to fix for all the syscall handlers.

lioncash (approved):

PR #2845, opened 2023-07-31
Thunks: Set bitness flags for 64-bit guests

Sonicadvance1:
This will help non-multiarch aware distros

### DIFF
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -63,9 +63,10 @@ function(generate NAME SOURCE_FILE)
 
   file(MAKE_DIRECTORY "${OUTFOLDER}")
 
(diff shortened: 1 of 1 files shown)


ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -66,6 +66,8 @@ function(generate NAME SOURCE_FILE)
   set (BITNESS_FLAGS "")
```
neobrain, line 66: This line isn't needed anymore
```suggestion
```

neobrain (approved):

ThunkLibs/GuestLibs/CMakeLists.txt
Sonicadvance1: Resolved.

Issue #2836, opened 2023-07-28
Missing GPR aliases in ARMEmitter


Sonicadvance1:
- [x] tst
- [x] bfc
- [x] bfxil
- [x] cinc
- [x] cinv
- [x] csetm
- [x] stadd{a,al,l,}{b,h,} -mishmash of these
- [x] strclr{a,al,l}{b,h,} -mishmash of these
- [x] streor{a,al,l}{b,h,} -mishmash of these
- [x] strset{a,al,l}{b,h,} -mishmash of these
- [x] stsmax{a,al,l}{b,h,}
- [x] stsmin{a,al,l}{b,h,}
- [x] stumax{a,al,l}{b,h,}
- [x] stumin{a,al,l}{b,h,}
- [x] ngc
- [x] ngcs
- [x] sbfiz
- [x] ubfiz

Seems to be all of the missing GPR aliases from a quick scan of the ARM ARM.

lioncash:
Closing, since this is now completed

PR #2847, opened 2023-07-31
ARMEmitter: Add ngc/ngcs aliases

lioncash:
Adds the Nintendo GameCube aliases for sbc/sbcs.

Crosses two more off #2836


Sonicadvance1 (approved):
So many GameCubes.

PR #2854, opened 2023-08-02
OpcodeDispatcher: Optimize rotates

alyssarosenzweig:
In the non-immediate cases, we can amortize some work between the two flags to come out 1 instruction ahead.

In the immediate case, costs us an extra 2 instructions compared to before we packed NZCV flags, but this mitigates a bigger instr count regression that this PR would otherwise have. Coming out ahead will require FlagM and smarter RA, but is doable.


Sonicadvance1 (approved):
Nice nice nice.

PR #2855, opened 2023-08-02
JIT: Use TST instead of CMN

alyssarosenzweig:
This is more obvious. llvm-mca says TST is half the cycle count of CMN for whatever it's defaulting to. dougallj's reference shows both as the same performance.


Sonicadvance1 (approved):
Likely an incorrect model in llvm-mca for this one but not like it really matters for our end.
cmn(adds) and tst(ands) are both single cycle going all the way back to Cortex-A57 up to Cortex-A715.
Although A57 had some weird performance where the shift or extended instruction encoding would fall down the M pipeline, which maybe that's what llvm-mca is picking up on?

alyssarosenzweig:
> Likely an incorrect model in llvm-mca for this one but not like it really matters for our end.

Yeah, fair enough. TST is a lot more obvious as a translation of TestNZ at any rate :~)

PR #2856, opened 2023-08-03
CMake: Allow overriding linker

Sonicadvance1:
CMake: Allow overriding linker

While the ENABLE_LLD and ENABLE_MOLD options are nice, they don't handle
the case when the linker of `lld` or `mold` doesn't match the compiler.

This particularly crops up when overriding the C compiler to a new
version of clang but the globally installed `ld.lld` is still the old
clang version.
This then causes clang to fail with unusual errors when upstream breaks
compatibility with itself.

Easy enough to use by passing the linker to cmake:
`-DUSE_LINKER=/usr/bin/ld.lld-15`

This also removes the ENABLE_LLD and ENABLE_MOLD options to use
USE_LINKER directly.
- ldd: `-DUSE_LINKER=lld`
- mold: `-DUSE_LINKER=mold`

Example of compiler failure when built with clang-15 but attempting to
link with ld.lld 14:
```bash
ld.lld-14: error: unittests/APITests/CMakeFiles/Filesystem.dir/Filesystem.cpp.o: Opaque pointers are only supported in -opaque-pointers mode (Producer: 'LLVM15.0.7' Reader: 'LLVM 14.0.6')
```


neobrain:
Can we drop `ENABLE_LLD` and `ENABLE_MOLD` in favor of this new setting? IIRC the only reason for adding `ENABLE_MOLD` as a separate variable was that `LD_OVERRIDE` triggered special behavior elsewhere in `CMakeLists.txt`, but that is not the case anymore.

We can also use a simpler variable name then (e.g. `USE_LINKER`), but no strong opinions on that.

Sonicadvance1:
> Can we drop `ENABLE_LLD` and `ENABLE_MOLD` in favor of this new setting? IIRC the only reason for adding `ENABLE_MOLD` as a separate variable was that `LD_OVERRIDE` triggered special behavior elsewhere in `CMakeLists.txt`, but that is not the case anymore.
> 
> We can also use a simpler variable name then (e.g. `USE_LINKER`), but no strong opinions on that.

Removed ENABLE_LLD and ENABLE_MOLD. Switched over to USE_LINKER name.

CMakeLists.txt
```diff
@@ -157,13 +156,9 @@ endif()
...
 set (PTHREAD_LIB pthread)
 
-if (ENABLE_LLD AND ENABLE_MOLD)
-  message (FATAL_ERROR "Cannot enable both lld and mold")
-elseif (ENABLE_LLD)
-  set (LD_OVERRIDE "-fuse-ld=lld")
-  add_link_options(${LD_OVERRIDE})
-elseif (ENABLE_MOLD)
-  add_link_options("-fuse-ld=mold")
+if (NOT USE_LINKER STREQUAL "")
```
neobrain, line 159: Empty strings evaluate to false already:

```suggestion
if (USE_LINKER)
```

neobrain (approved):

CMakeLists.txt
Sonicadvance1: resolved.

PR #2857, opened 2023-08-04
IR: Fixes RAValidation for 32-bit applications

Sonicadvance1:
RAValidation was making an assumption that GPR register class would only have up to 16 registers for either SRA or dynamic registers.

When running a 32-bit application we allow 17 GPRs to be dynamically allocated, since we can take 8 back from SRA in that case.

Just split the two classes in the RAValidation pass since they will never overlap their allocation.

Fixes validation in `32Bit_Secondary/15_XX_0.asm` locally that changed behaviour due to tinkering.


Sonicadvance1:
Rebased on latest main.

lioncash (approved):

PR #2859, opened 2023-08-04
Fix 16-bit popa insertion behaviour

bylaws:
Caught this when debugging Stalker, OpenAL does this amazing little hack https://github.com/OpenXRay/xray-15/blob/1390dfb08ed20997d7e8c95147ea8e8cb71f5e86/cs/3rd%20party/OpenAL/OpenAL-Windows/Router/alc.cpp#L1862 that was broken by the FEX behaviour


Sonicadvance1 (approved):
Nice catch! Bad test wasn't checking that the pops were doing proper 16-bit inserts.

PR #2860, opened 2023-08-05
ARMEmitter: Add missing atomic aliases

lioncash:
Finishes off the missing aliases in #2836


Sonicadvance1 (approved):
Nice, thanks for hitting the missing tests as well.

PR #2861, opened 2023-08-09
FEXCore: Fixes vector shifts by zero

Sonicadvance1:
All vector shifts by zero were broken on the AArch64 JIT since shift operations by constant encoding of "zero" turns in to a different instruction.

Fixes this by making OpcodeDispatcher not emit zero shifts and also making the JIT treat them as nops or moves in that case.

Fixes a bug uncovered by my instruction auditing.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -2474,11 +2474,25 @@ DEF_OP(VUShrI) {
     if (HostSupportsSVE && Is256Bit) {
       const auto Mask = PRED_TMP_32B.Merging();
 
-      // SVE LSR is destructive, so lets set up the destination.
-      movprfx(Dst.Z(), Vector.Z());
-      lsr(SubRegSize, Dst.Z(), Mask, Dst.Z(), BitShift);
+      if (BitShift == 0) {
+        if (Dst.Idx() != Vector.Idx()) {
+          mov(Dst.Z(), Vector.Z());
+        }
+      }
+      else {
+        // SVE LSR is destructive, so lets set up the destination.
+        movprfx(Dst.Z(), Vector.Z());
+        lsr(SubRegSize, Dst.Z(), Mask, Dst.Z(), BitShift);
+      }
     } else {
-      ushr(SubRegSize, Dst.Q(), Vector.Q(), BitShift);
+      if (BitShift == 0) {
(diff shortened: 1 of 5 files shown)


External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -2475,10 +2475,24 @@ DEF_OP(VUShrI) {
       const auto Mask = PRED_TMP_32B.Merging();
 
       // SVE LSR is destructive, so lets set up the destination.
```
lioncash, line 2477: This should be moved by the movprfx

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -2505,10 +2519,24 @@ DEF_OP(VSShrI) {
     const auto Mask = PRED_TMP_32B.Merging();
 
     // SVE ASR is destructive, so lets set up the destination.
```
lioncash, line 2521: ditto

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -2538,11 +2566,24 @@ DEF_OP(VShlI) {
       const auto Mask = PRED_TMP_32B.Merging();
 
       // SVE LSL is destructive, so lets set up the destination.
```
lioncash, line 2568: ditto

lioncash (commented):
Minor nits but looks good overall

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
Sonicadvance1: Resolved.

External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
Sonicadvance1: Resolved.

lioncash (approved):

PR #2864, opened 2023-08-09
Config: Adds support for overriding host features

Sonicadvance1:
This allows use to both enable and disable regardless of what the host supports. This replaces the old `EnableAVX` option.

Unlike the old EnableAVX option which was a binary option which could only disable, each of these options are technically trinary states. Not setting an option gives you the default detection, while explicitly enabling or disabling will toggle the option regardless of what the host supports.

This will be used by the instruction count CI in the future.


Sonicadvance1:
Woops, merged #2865 which brought this in.

PR #2868, opened 2023-08-09
IR: Adds Option to run the IRDumper with more configurations

Sonicadvance1:
This is incredibly useful and I find myself hacking this feature in
every time I am optimizing IR. Adds a new configuration option which
allows dumping IR at various times.

Before any optimization passes has happened
After all optimizations passes have happened
Before and After each IRPass to see what is breaking something.

### DIFF
--- a/External/FEXCore/Source/Interface/IR/Passes/IRDumperPass.cpp
+++ b/External/FEXCore/Source/Interface/IR/Passes/IRDumperPass.cpp
@@ -1,0 +1,84 @@
+/*
+$info$
+tags: ir|debug
+desc: Prints IR
+$end_info$
+*/
+
+#include "Interface/IR/PassManager.h"
+#include "Interface/IR/Passes/RegisterAllocationPass.h"
+#include "Interface/Core/OpcodeDispatcher.h"
+
+#include <FEXCore/IR/IR.h>
+
+namespace FEXCore::IR::Debug {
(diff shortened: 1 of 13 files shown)


External/FEXCore/Source/Interface/IR/Passes/IRDumperPass.cpp
```diff
@@ -0,0 +1,83 @@
...
+
+private:
+  FEX_CONFIG_OPT(DumpIR, DUMPIR);
+  bool DumpToFile{};
+  bool DumpToLog{};
+};
+
+IRDumper::IRDumper() {
+  const auto DumpIRStr = DumpIR();
+  if (DumpIRStr == "stderr" || DumpIRStr == "stdout" || DumpIRStr == "no") {
```
lioncash, line 28: maybe add a comment in the body like:

```cpp
// Intentionally do nothing
```

just to make it super clear the lack of behavior is intentional

lioncash (commented):
Ooh, very nice!

Just one small nit, but looks good otherwise

External/FEXCore/Source/Interface/IR/Passes/IRDumperPass.cpp
Sonicadvance1: comment added.

lioncash (approved):

PR #2872, opened 2023-08-10
FEX: Adds instruction count CI

Sonicadvance1:
Implements CI for tracking instruction counts for generate blocks of
code when transforming from x86 to ARM64 assembly.

This will end up encompassing every instruction in our instruction
tables similarly to how our assembly tests try to test everything in our
instruction tables.

Incidentally, the data for this CI is generated using our assembly
tests. By enabling disassembly and instruction stats when executing a
suite of instructions, this gives the stats that can be added to a json
file.

The current implementation only implements the SecondGroup table of
instructions because it is a relatively small table and has known
inefficiencies in the instruction implementations. As this gets merged I
will be adding more tables of instructions to additional json files for
testing.

These JSON files will support adjusting CPU features regardless of the
host features so it can test implementations depending on different CPU
features. This will let us test things like one instruction having
different "optimal" implementations depending on if it supports SVE128,
SVE256, SVEI8MM, etc.

This initial instruction auditing is what found the bug in our vector
shift instructions by size of zero. If inspecting the result of the CI
run, you can tell that these instructions still aren't "optimal" because
they are doing loads and stores that can be eliminated.

The "Optimal" in the JSON is purely for human readable and grepping
ability to see what is optimal versus not. Same with the "Comment"
section.

According to my auditing spreadsheet, the total number of instructions
that will end up in these json files will be about 1000, but we will
likely end up with more since there will be edge cases that can be more
optimal depending on arguments.

### DIFF
--- a/External/FEXCore/include/FEXCore/Utils/File.h
+++ b/External/FEXCore/include/FEXCore/Utils/File.h
@@ -1,5 +1,6 @@
 #pragma once
 #include <FEXCore/fextl/allocator.h>
+#include <FEXCore/fextl/string.h>
 #include <FEXCore/Utils/EnumOperators.h>
 
 #ifndef _WIN32
@@ -79,6 +80,10 @@ namespace FEXCore::File {
         // Some error, match Linux side.
         return -1;
 #endif
+      }
+
+      ssize_t Write(std::string_view const Data) {
+        return Write(Data.data(), Data.size());
       }
 
       /**
--- a/Source/Tools/CodeSizeValidation/Main.cpp
+++ b/Source/Tools/CodeSizeValidation/Main.cpp
@@ -1,0 +1,432 @@
+#include "DummyHandlers.h"
+#include "FEXCore/Core/Context.h"
+#include "FEXCore/Debug/InternalThreadState.h"
+#include <FEXCore/Config/Config.h>
+#include <FEXCore/Utils/Allocator.h>
+#include <FEXCore/Utils/File.h>
+#include <FEXCore/Utils/FileLoading.h>
+#include <FEXCore/Utils/LogManager.h>
+
+namespace CodeSize {
+  class CodeSizeValidation final {
+    public:
+      struct InstructionStats {
+        uint64_t GuestCodeInstructions{};
+        uint64_t HostCodeInstructions{};
+
+        uint64_t HeaderSize{};
+        uint64_t TailSize{};
+      };
+
+      using CodeLines = fextl::vector<fextl::string>;
+      using InstructionData = std::pair<InstructionStats, CodeLines>;
+
+      bool ParseMessage(char const *Message);
+
+      InstructionData *GetDataForRIP(uint64_t RIP) {
+        return &RIPToStats[RIP];
+      }
+
+      bool InfoPrintingDisabled() const {
+        return SetupInfoDisabled;
+      }
+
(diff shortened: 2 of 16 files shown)


alyssarosenzweig:
Is there a tool to automatically update the expectations (i.e. if implementing an optimization that affects 100s of instructions?)

Sonicadvance1:
> Is there a tool to automatically update the expectations (i.e. if implementing an optimization that affects 100s of instructions?)

Not yet, but I can generate a tool for this.

Sonicadvance1:
> Is there a tool to automatically update the expectations (i.e. if implementing an optimization that affects 100s of instructions?)

Added a tool to update the expected instruction counts after CI has failed.
Flow will basically be:
```bash
ninja instcountci_tests
# Notice that instruction counts has changed
ninja instcountci_update_tests
```

Then afterwards checking the updated files. Care will need to be taken to ensure if something marked "Optimal" has changed that it is no longer optimal. Might need some more metadata to track expected optimal instruction count and updating that automatically in this case, but unsure if necessary for now. Will leave that as an exercise for the future.
Main thing is that if all the flag updating changes, you're not going to need to manually update 500 instructions worth of data manually.

Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -0,0 +1,432 @@
...
+      uint64_t CurrentRIPParse{};
+      bool ConsumingDisassembly{};
+      InstructionData *CurrentStats{};
+      InstructionStats BaseStats{};
+
+      fextl::unordered_map<uint64_t, InstructionData> RIPToStats;
+      bool SetupInfoDisabled{};
+  };
+
+  const fextl::string RIPMessage = "RIP: 0x";
```
lioncash, line 55: Seems like these strings could be constexpr `std::string_view` instances.

External/FEXCore/include/FEXCore/Utils/File.h
```diff
@@ -81,6 +82,10 @@ namespace FEXCore::File {
 #endif
       }
 
+      ssize_t Write(fextl::string const &Data) {
```
lioncash, line 85: Perhaps this should take a std::string_view, so that it can accept all sorts of string types instead of potentially converting to an allocating string

Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -0,0 +1,432 @@
...
+#include <FEXCore/Utils/LogManager.h>
+
+namespace CodeSize {
+  class CodeSizeValidation final {
+    public:
+      struct InstructionStats {
+        uint64_t GuestCodeInstructions{};
+        uint64_t HostCodeInstructions{};
+
+        uint64_t HeaderSize;
```
lioncash, line 17: Should these be initialized to zero as well, like the members above?

Sonicadvance1: Zero initialized.

External/FEXCore/include/FEXCore/Utils/File.h
Sonicadvance1: switched to string_view

Source/Tools/CodeSizeValidation/Main.cpp
Sonicadvance1: switched to constexpr string_view

lioncash (approved):

PR #2873, opened 2023-08-10
Various warning fixes

neobrain:
With these, only three warnings relating to our use of `offsetof` in the dispatchers and one warning in ImGui remain.

@lioncash Could you double-check that the variables in SVEOps.inl are intentionally unused in Release builds?


lioncash:
> @lioncash Could you double-check that the variables in SVEOps.inl are intentionally unused in Release builds?

Yeah, those are intentionally unused. They're just for use in debug assertions

Sonicadvance1 (approved):

PR #2874, opened 2023-08-10
Add WOW64 JIT frontend

bylaws:
- Implements the WOW64 BT API using FEXCore, allowing the unix-side of wine to run as ARM code, with FEX only being called into to run the PE-side x86 application or wine code. WOW64 itself handles most of the dirty work that FEXLoader/LinuxSyscalls end up doing so the implementation here is comparatively small.
- At the moment the boundary between ARM and x86 code is at the syscall/unixcall level, so the resulting performance is likely only a slight improvement over FEXLoader + (hypothetical) 32 bit thunks. However this paves the way for significant improvements in the future, extending the WOW frontend to support ARM64EC/CHPE-style interop with a custom ABI, so that DXVK and wine DLLs can run as native code. Most of this code can also be reused for 64-bit support.
 
- Currently depends on downstream wine patches to allow FEX to link to kernel32 and use TLS:
- - https://github.com/bylaws/wine/commit/0e49a56ae201ce84630a22d3dd36dc17004d1f14
- -  https://github.com/bylaws/wine/commit/593e45860351d34fb58ba1512599928b83d965db
- These patches are unlikely to ever make it upstream, however to not require them would require removing FEX's libc/kernel32 dependency - a fairly large refactor. There's also potential this could be solved via thunks but I'm not quite sure yet as to if that's viable.
- Also depends on downstream patches adding some additional callbacks which I am yet to upstream:
- - https://github.com/bylaws/wine/commit/af9a5991662af72736f1c86f3c77fb2ba59f9b98
- - https://github.com/bylaws/wine/commit/21bd917f01a2a8897692359b3526a47781c0fb8c
- - https://github.com/bylaws/wine/commit/2b2416d1214dc3078984f5d60cdae0ee5cc813c0

- I have local patches for thread suspension, however after researching arm64ec I think it can be done in a better way, so I am not including them for the moment 

Thank you to Andre Zwing for his work on [hangover](https://github.com/AndreRH/hangover), which this was based upon

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -71,11 +71,6 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
     FEXCore::X86State::REG_RSP,
   };
 
-  static constexpr SyscallArray GPRIndexes_Win32 = {
-    FEXCore::X86State::REG_RAX,
-    FEXCore::X86State::REG_RSP,
-  };
-
   SyscallFlags DefaultSyscallFlags = FEXCore::IR::SyscallFlags::DEFAULT;
 
   const auto OSABI = CTX->SyscallHandler->GetOSABI();
@@ -93,8 +88,9 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
     DefaultSyscallFlags = FEXCore::IR::SyscallFlags::NORETURNEDRESULT;
   }
   else if (OSABI == FEXCore::HLE::SyscallOSABI::OS_WIN32) {
-    NumArguments = 2;
-    GPRIndexes = &GPRIndexes_Win32;
+    // Since the whole context is going to be saved at entry anyway, theres no need to do additional work to pass in args
+    NumArguments = 0;
+    GPRIndexes = nullptr;
     DefaultSyscallFlags = FEXCore::IR::SyscallFlags::NORETURNEDRESULT;
   }
   else if (OSABI == FEXCore::HLE::SyscallOSABI::OS_HANGOVER) {
--- a/Source/Windows/WOW64/IntervalList.h
+++ b/Source/Windows/WOW64/IntervalList.h
@@ -1,0 +1,144 @@
+#pragma once
+
+#include <utility>
+#include <algorithm>
+
+#include <FEXCore/fextl/vector.h>
+
+template<typename SizeType>
+class IntervalList {
+public:
+  using DifferenceType = decltype(std::declval<SizeType>() - std::declval<SizeType>());
+
+  struct Interval {
+    SizeType Offset;
+    SizeType End;
+
+    Interval() = default;
+
+    Interval(SizeType Offset, SizeType End) : Offset{Offset}, End{End} {}
+  };
+
+private:
+  fextl::vector<Interval> Intervals; ///< list of intervals sorted by their end offset
+
+public:
+  struct QueryResult {
+    bool Enclosed; ///< If the given offset was enclosed by an interval
+    DifferenceType Size; ///< Size of the interval starting from the query offset, or distance to the next interval if
+                         /// `Enclosed` is false (if there is no next interval, size is 0)
+  };
+
+  void Clear() {
+    Intervals.clear();
+  }
+
+  void Insert(Interval Entry) {
+    if (Entry.Offset == Entry.End) {
+      return;
+    }
+
+    auto [FirstIt, EndIt] = std::equal_range(Intervals.begin(), Intervals.end(), Entry, [](const auto &LHS, const auto &RHS) {
+      return LHS.End <= RHS.Offset;
+    });
+
+    if (FirstIt == EndIt) {
+      // No overlaps
+      Intervals.insert(FirstIt, Entry);
+      return;
+    }
+
+    auto LastIt = std::prev(EndIt);
+    // FirstIt/LastIt are the lowest/highest offset intervals respectively that overlap with the new interval
+
+    const SizeType Offset = std::min(Entry.Offset, FirstIt->Offset);
+    const SizeType End = std::max(LastIt->End, Entry.End);
+
+    // Erase all overlapping entries but the first
+    const auto EraseStartIt = std::next(FirstIt);
+    const auto EraseEndIt = std::next(LastIt);
+    LastIt = Intervals.erase(EraseStartIt, EraseEndIt);
+    FirstIt = std::prev(LastIt);
+
+    FirstIt->Offset = Offset;
+    FirstIt->End = End;
+  }
+
+  void Remove(Interval Entry) {
+    if (Entry.Offset == Entry.End) {
+      return;
+    }
+
+    auto [FirstIt, EndIt] = std::equal_range(Intervals.begin(), Intervals.end(), Entry, [](const auto &LHS, const auto &RHS) {
+      return LHS.End <= RHS.Offset;
+    });
+
+    if (FirstIt == EndIt) {
+      // No intersecting intervals present, nothing more to do
+      return;
+    }
+
+    if (FirstIt->Offset < Entry.Offset && FirstIt->End > Entry.End) {
+      // The interval to be removed is fully enclosed by an existing interval
+      
+      // Break the single interval into two smaller intervals on either side on the interval being removed
+      const auto FirstPredecessorIt = Intervals.insert(FirstIt, *FirstIt);
+      FirstIt = std::next(FirstPredecessorIt);
+      FirstPredecessorIt->End = Entry.Offset;
+      FirstIt->Offset = Entry.End;
+      return;
+    }
+
+    auto LastIt = std::prev(EndIt);
+    // FirstIt/LastIt are the lowest/highest offset intervals respectively that overlap with the new interval
+
+    if (FirstIt->Offset < Entry.Offset) {
+      // The first overlap straddles the start of the interval to be removed
+      FirstIt->End = Entry.Offset;
+      if (FirstIt == LastIt) {
+        // No more overlaps left, nothing more to do
+        return;
+      } else {
+        FirstIt++;
+      }
+    }
+
+    if (LastIt->End > Entry.End) {
+      // The last overlap straddles the end of the interval to be removed
+      LastIt->Offset = Entry.End;
+      if (LastIt == FirstIt) {
+        // No more overlaps left, nothing more to do
+        return;
+      } else {
+        LastIt--;
+      }
+    }
+
+    // Now none of the overlaps straddle the edges of the interval to be removed they can all be erased
+    const auto EraseStartIt = FirstIt;
+    const auto EraseEndIt = std::next(LastIt);
+    Intervals.erase(EraseStartIt, EraseEndIt);
+  }
+
+  QueryResult Query(SizeType Offset) {
+    const auto It = std::upper_bound(Intervals.begin(), Intervals.end(), Offset, [](const auto &LHS, const auto &RHS) {
+      return LHS < RHS.End;
+    }); // Lowest offset interval that (maybe) overlaps with the query offset
+
+    if (It == Intervals.end()) { // No overlaps past offset
+      return {false, {}};
+    } else if (It->Offset > Offset) { // No overlap, return the distance to the next possible overlap
+      return {false, It->Offset - Offset};
+    } else { // Overlap, return the distance to the end of the overlap
+      return {true, It->End - Offset};
+    }
+  }
+
+  bool Intersect(Interval Entry) {
+    const auto It = std::upper_bound(Intervals.begin(), Intervals.end(), Entry, [](const auto &LHS, const auto &RHS) {
+      return LHS.Offset < RHS.End;
+    }); // Lowest offset interval that (maybe) overlaps with the query offset
+
+    return It != Intervals.end() && It->Offset < Entry.End;
+  }
+};
--- a/Source/Windows/WOW64/IntervalList.h
+++ b/Source/Windows/WOW64/IntervalList.h
@@ -1,3 +1,4 @@
+// SPDX-License-Identifier: MIT
 #pragma once
 
 #include <utility>
--- a/Source/Windows/WOW64/Module.cpp
+++ b/Source/Windows/WOW64/Module.cpp
@@ -204,6 +204,21 @@ namespace Context {
     Context::StoreWowContextFromState(TLS.ThreadState(), &TmpWowContext);
     return RtlWow64SetThreadContext(Thread, &TmpWowContext);
   }
+
+  bool HandleUnalignedAccess(CONTEXT *Context) {
+    if (!GetTLS().ThreadState()->CPUBackend->IsAddressInCodeBuffer(Context->Pc)) {
+      return false;
+    }
+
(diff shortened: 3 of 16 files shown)


Sonicadvance1:
Most of the FEX facing changes seem reasonable to me. Although the `ExitFunction` changes should likely be opt-in until we find a better way to handle this. Which I have some ideas but I don't want to regress Linux while trying to solve the wine side.

bylaws:
> Most of the FEX facing changes seem reasonable to me. Although the `ExitFunction` changes should likely be opt-in until we find a better way to handle this. Which I have some ideas but I don't want to regress Linux while trying to solve the wine side.

Ah right yeah I should definitely ifdef it out for now, though what are your ideas for handling this better?

Sonicadvance1:
> > Most of the FEX facing changes seem reasonable to me. Although the `ExitFunction` changes should likely be opt-in until we find a better way to handle this. Which I have some ideas but I don't want to regress Linux while trying to solve the wine side.
> 
> Ah right yeah I should definitely ifdef it out for now, though what are your ideas for handling this better?

Similar to how our deferred signal handling works, we could potentially allocate the page before/after the CPUState context, and then pay a `str wzr, [x28 - 4]` instruction on the block entry after we stored the RIP. When wanting to pause the JIT, just change the mapping permissions and catch the fault.
This will let us know that we have synchronized CoreState and can longjump the code back to the dispatcher.
There's an open question about how this would work with something like syscalls that can hang in the kernel forever.

This is the same sort of problem space that our gdbserver implementation needs to handle if we want to sanely handle pausing the emulation, just haven't gotten around to solving all the problems.

bylaws:
> Similar to how our deferred signal handling works, we could potentially allocate the page before/after the CPUState context, and then pay a `str wzr, [x28 - 4]` instruction on the block entry after we stored the RIP. When wanting to pause the JIT, just change the mapping permissions and catch the fault. This will let us know that we have synchronized CoreState and can longjump the code back to the dispatcher. There's an open question about how this would work with something like syscalls that can hang in the kernel forever.

Ah I see, that would work quite well, syscalls aren't a problem for WOW as context is already spilled on entry and there's some sync around them to handle this (see {Lock,Unlock}JITContext). I'll try and get something working with this approach and push it up here
> 
> This is the same sort of problem space that our gdbserver implementation needs to handle if we want to sanely handle pausing the emulation, just haven't gotten around to solving all the problems.

Makes sense, winegdb is implemented ontop of this same API on the win32 side

bylaws:
@Sonicadvance1 How does that look?

Sonicadvance1:
That's about what I envisioned.
If we wanted to we could do some minor changes.
```asm
 ldr(TMP1, STATE, offsetof(FEXCore::Core::CPUState, InterruptFaultAddress));
  str(ARMEmitter::XReg::zr, TMP1, 0);
  ```
  can get converted to 
  ```asm
  sturb(ARMEmitter::XReg::zr, STATE, -1);
  ```
  
This would remove an instruction and the ldr+str dependency chain. This would make allocation of the InternalThreadState object a little bit more complex.
Would require changing the allocation in Core.cpp
```cpp
FEXCore::Core::InternalThreadState *Thread = new FEXCore::Core::InternalThreadState{};
```

to something like
```diff
diff --git a/External/FEXCore/Source/Interface/Core/Core.cpp b/External/FEXCore/Source/Interface/Core/Core.cpp
index f3297178f..6def38c7b 100644
--- a/External/FEXCore/Source/Interface/Core/Core.cpp
+++ b/External/FEXCore/Source/Interface/Core/Core.cpp
@@ -608,7 +608,9 @@ namespace FEXCore::Context {
   }

   FEXCore::Core::InternalThreadState* ContextImpl::CreateThread(FEXCore::Core::CPUState *NewThreadState, uint64_t ParentTID) {
-    FEXCore::Core::InternalThreadState *Thread = new FEXCore::Core::InternalThreadState{};
+    auto BasePtr = FEXCore::Allocator::VirtualAlloc(4096 + sizeof(FEXCore::Core::InternalThreadState));
+    void *ThreadPtr = (void*)((uint64_t)BasePtr + 4096);
+    auto Thread = new (ThreadPtr) FEXCore::Core::InternalThreadState{};

     // Copy over the new thread state to the new object
     if (NewThreadState) {
diff --git a/External/FEXCore/include/FEXCore/Utils/AllocatorHooks.h b/External/FEXCore/include/FEXCore/Utils/AllocatorHooks.h
index 3d1a3c563..c7ee061ec 100644
--- a/External/FEXCore/include/FEXCore/Utils/AllocatorHooks.h
+++ b/External/FEXCore/include/FEXCore/Utils/AllocatorHooks.h
@@ -135,6 +135,10 @@ namespace FEXCore::Allocator {
       return FEXCore::Allocator::malloc(size);
     }

+    void *operator new(size_t size, void* ptr) {
+      return ptr;
+    }
+
     void operator delete(void *ptr) {
       return FEXCore::Allocator::free(ptr);
     }
```

Plus the rest of the relevant changes. Of course it would likely be something smarter than that so we aren't passing around an InternalThreadState plus magic offset where we always know the page before.

Alternatively always align the internalthreadstate to a page, pad to the next page and access the faulting next page with.
```asm
strb(ARMEmitter::XReg::zr, STATE, 4096);
```
This would make the frontend management less nonsensical.

bylaws:
Ended up doing it with alignas, seemed to be the neatest way

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -791,6 +791,10 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry,
   adr(TMP1, &JITCodeHeaderLabel);
   str(TMP1, STATE, offsetof(FEXCore::Core::CPUState, InlineJITBlockHeader));
 
+  // Trigger a fault if there are any pending interrupts
+  strb(ARMEmitter::XReg::zr, STATE, offsetof(FEXCore::Core::InternalThreadState, InterruptFaultPage) -
```
Sonicadvance1, line 795: I believe you'll want to use 64-bit str here. Since the imm offset has significant offset limitations since it gets scaled by the size of the access. Theoretically this strb won't even work here (and should assert in assertions builds).
- strb imm access range is 0-4095
- str (64-bit) imm access range is 0-32760 (members must be aligned to 64-bit natural alignment).

64-bit will let us keep scaling the state past 4096 once the base CPUState eventually gets larger than 1 page as well.

Sonicadvance1:
Looking very interesting. You'll probably want to fix your rebase so that you're not pulling in commits that were already merged. That way it can run through CI and it's less of a pain to review. Since a lot is duplicated from main.

External/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
bylaws: I understand the range limits, but why would the emitter assert here given it is in range (which I do static assert for)?

Sonicadvance1: Oh, you're right. I forgot that the `CpuStateFrame` isn't at the base of `InternalThreadState` anymore. So yes, you do get away with it for now.
Once we add more stuff to the CpuStateFrame we can easily bump it to an strh or whatever.

Source/Windows/WOW64/Module.cpp
```diff
@@ -0,0 +1,784 @@
+/*
+$info$
+tags: Bin|WOW64
+desc: Implements the WOW64 BT module API using FEXCore
+$end_info$
+*/
+
+// Initially based upon hangover by Andre Zwing (https://github.com/AndreRH/hangover)
```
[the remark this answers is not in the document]
bylaws: It's also MIT so don't see much reason to

Source/Windows/WOW64/IntervalList.h
```diff
@@ -0,0 +1,143 @@
...
+    if (FirstIt == Intervals.end() || FirstIt->Offset >= Entry.End) {
+      Intervals.insert(FirstIt, Entry);
+      return;
+    }
+    // Now FirstIt will always overlap
+
+    auto LastIt = FirstIt; // Highest offset interval that overlaps with the new interval
+    while ((std::next(LastIt) != Intervals.end()) && (std::next(LastIt)->Offset < Entry.End)) {
+      LastIt++;
+    }
```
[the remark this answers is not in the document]
bylaws: No, I don't think it would since I want a strict inequality here on both ends

Source/Windows/WOW64/IntervalList.h
```diff
@@ -0,0 +1,143 @@
...
+    Intervals.clear();
+  }
+
+  void Insert(Interval Entry) {
+    auto FirstIt = std::upper_bound(Intervals.begin(), Intervals.end(), Entry, [](const auto &LHS, const auto &RHS) {
+      return LHS.Offset < RHS.End;
+    }); // Lowest offset interval that (maybe) overlaps with the new interval
+
+    if (FirstIt == Intervals.end() || FirstIt->Offset >= Entry.End) {
+      Intervals.insert(FirstIt, Entry);
```
neobrain, line 42: Does this handle the scenario `Intervals = { [0..10) }; Entry = [10..20)` correctly?

Have you considered adding some Catch2-based unit tests for this class?

bylaws: Yes it does, I considered unit tests but thought the code was probably small enough to be okay without

Source/Windows/WOW64/IntervalList.h
```diff
@@ -0,0 +1,143 @@
...
+      return;
+    }
+    // Now FirstIt will always overlap
+
+    auto LastIt = FirstIt; // Highest offset interval that overlaps with the new interval
+    while ((std::next(LastIt) != Intervals.end()) && (std::next(LastIt)->Offset < Entry.End)) {
```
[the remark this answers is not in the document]
bylaws: (superseded by equal_range now but) why would the loop terminate as you say? It would incr lastIt till the element after it has an offset >= to the entries end, but in this case it wouldn't hit that and would end being prev(Intervals.end()) as expected

Source/Windows/WOW64/IntervalList.h
neobrain: It's not the most critical thing to unit tests indeed, but it's also very easy to write test code for. I would sleep better if we had coverage to ensure the edge cases are handled right :)

Source/Windows/WOW64/Module.cpp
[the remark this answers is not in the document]
neobrain: MIT requires reproduction of the copyright notice, which a casual name drop isn't. You also don't mention what specific file this was taken from, while the hangover repository itself is licensed as LGPL. 

More pragmatically, external repositories can disappear, so it'd be good to document for future reference that somebody has done their due diligence of checking this code's license. Just mentioning it's MIT licensed is better than having a lose mention of some code that can no longer be checked for its original license.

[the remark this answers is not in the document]
neobrain: That's fair, but then the wording for this acknowledgement is very misleading. Maybe something like "Thanks to Andre Zwing, whose ideas from https://github.com/AndreRH/hangover this code is based on" would be better suited?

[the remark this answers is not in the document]
bylaws: Ah sure I'll do that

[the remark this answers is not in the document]
neobrain: (Side note, looks like the correct spelling of their name is `André Zwing`)

Source/Windows/WOW64/WineHelpers.cpp
```diff
@@ -0,0 +1,172 @@
+/*
+ * Copyright 2023 Alexandre Julliard
```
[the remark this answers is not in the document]
bylaws: This should still be fine (resulting license for wowfex would be lgpl) but I plan to replace these before merge anyway with MIT equivalents (and fix some FEXLoader bugs while I'm at it) so there won't be any issues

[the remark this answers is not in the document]
bylaws: Done

Sonicadvance1:
If you open a PR with the FEXCore changes broken out we can accept that early.
Still need to think about the LGPL and MIT interactions. Might need to adopt SPDX licensing or something before frontend Windows stuff can be merged?

bylaws:
> If you open a PR with the FEXCore changes broken out we can accept that early. Still need to think about the LGPL and MIT interactions. Might need to adopt SPDX licensing or something before frontend Windows stuff can be merged?

License wise I've removed the need for all of the winehelpers code so there shouldn't be anything LGPL left, the headers can be re-licensed as we like per LGPL v3 section 3. So I suppose the correct thing to do is just to remove the LGPL header from them, leaving just the copyrights.

Source/Windows/WOW64/IntervalList.h
```diff
@@ -0,0 +1,144 @@
+#pragma once
```
Sonicadvance1, line 1: Needs SPDX

Source/Windows/WOW64/Module.cpp
```diff
@@ -0,0 +1,679 @@
+/*
```
Sonicadvance1, line 1: Needs SPDX

Source/Windows/include/wine/debug.h
```diff
@@ -0,0 +1,33 @@
+/*
```
Sonicadvance1, line 1: Needs SPDX

Source/Windows/include/wine/unixlib.h
```diff
@@ -0,0 +1,35 @@
+/*
```
Sonicadvance1, line 1: Needs SPDX

Source/Windows/include/winternl.h
```diff
@@ -0,0 +1,123 @@
+/*
```
Sonicadvance1, line 1: Needs SPDX

Sonicadvance1 (commented):
Once SPDX is added I don't see any reason why this couldn't be merged.

bylaws:
Rebased

Sonicadvance1 (approved):
woop woop woop

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -93,8 +88,9 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
     DefaultSyscallFlags = FEXCore::IR::SyscallFlags::NORETURNEDRESULT;
   }
   else if (OSABI == FEXCore::HLE::SyscallOSABI::OS_WIN32) {
-    NumArguments = 2;
-    GPRIndexes = &GPRIndexes_Win32;
+    // Since the whole context is going to be saved at entry anyway, theres no need to do additional work to pass in args
+    NumArguments = 0;
+    GPRIndexes = nullptr;
```
pmatos, line 93: Was just looking at this piece of code. I think there's an issue here because we are setting GPRIndexes to null but the next thing we do is to dereference GPRIndexes. I don't understand the WIN32 context but should we have an early exit here?

bylaws: This does end up creating a null reference, however it's never accessed since NumArguments is 0, will probably be cleaner to just keep GPRIndexes as a pointer and avoid the UB

PR #2875, opened 2023-08-11
ARMEmitter: Handle contiguous first fault load (scalar plus scalar) group

lioncash:
Adds the only missing implementation category for the first-faulting loads, making the interface more consistent


Sonicadvance1 (approved):
Noice.

PR #2878, opened 2023-08-11
ARMEmitter: Handle SVE FCPY (predicated)

lioncash:


Sonicadvance1 (approved):
Very nice float encoding

PR #2879, opened 2023-08-11
FEXCore: Fixes bug with 32-bit adcx

Sonicadvance1:
When a 32-bit adcx instruction was encountered, it was getting treated as a 16-bit adcx instruction instead. This is because of the 0x66 prefix required to handle this instruction.

Adds a unit test to ensure it doesn't break again.

Instruction count CI caught this while looking at the assembly output.


lioncash (approved):
Nice catch!

PR #2880, opened 2023-08-11
ARMEmitter: Migrate off vixl float utils

lioncash:
Since we have our own now, we can lessen our dependence on vixl by shifting over to our own equivalents


Sonicadvance1 (approved):
Neato

PR #2882, opened 2023-08-11
ARMEmitter: Handle SVE ADR

lioncash:


dougallj:
Arm's assembly syntax for ADR annoys me, because it isn't a memory operation, it's an add with optional extend and optional left-shift by 1-3. On the scalar side, this is written as `add x0, x1, x2, lsl #1` or `add x0, x1, w2, sxtw #3` or whatever. This is similar to x86's LEA, but without the excuse that it uses the same encoding – none of the SVE scatter/gather instructions can use VectorPlusVector addressing. It's a memory operand that's exclusive to ADR – which isn't a memory operation!

Anyway, that's my rant, sorry. If you want to match the official assembly syntax – that's a sensible goal – and this does, so that's good. But I'd also consider the possibility of using an `add v0.d, v1.d, v2.d, sxtw #3` style encoding and skipping the `SVEMemOperand`.

lioncash:
Yeah, the only reason I initially used SVEMemOperand for it was for expression consistency with other load/stores (even though adr just masquerades as one).

I've split the changing off of SVEMemOperand into another commit, because this also entails moving the modifier enum out of the class (which is also pretty nice, since the qualifiers aren't as noisy anymore).

Sonicadvance1 (approved):

PR #2890, opened 2023-08-14
ConstProp: Fix set-but-not-used mask variable

alyssarosenzweig:
I think this was the intended logic?


neobrain:
Can't comment on the change itself, but was there a compiler warning for this? `-Wunused` should have caught it, but somehow I didn't notice any mention of this file when working on #2873. Are we not setting up warning flags correctly for FEXCore?

Sonicadvance1 (approved):
Nice. This warning has been around for a long time and I haven't looked in to what the logic wanted here.

PR #2891, opened 2023-08-14
Move External/FEXCore/ to FEXCore/

alyssarosenzweig:
It is not an external component, and it makes paths needlessly long. Ryan seemed amenable to this when we discussed on IRC earlier.


Sonicadvance1 (approved):

Sonicadvance1:
Needs a rebase.

PR #2897, opened 2023-08-15
ARMEmitter: Handle SVE load and broadcast element group

lioncash:
These can be used to improve vbroadcast implementations from doing a mem load+dup in the non-GPR case into just directly loading into the destination.


lioncash:
Errors seem related to the instruction CI script blowing up, which is unrelated

Sonicadvance1 (approved):
Curious that nasm is seemingly randomly not output its output file randomly, but only in CI and doesn't reproduce locally.

PR #2898, opened 2023-08-15
InstCountCI: Support encoding expected Arm64 ASM in JSON

Sonicadvance1:
This will allow investigating the Arm64 directly next to the test, plus
publicly linking directly to badly behaving tests.

Perfect for nerdsniping implementations.

Main change is in the first commit. The second commit is purely automated data change.


Sonicadvance1:
Before this is ready I need to regenerate the results and instruction numbers with a real ARM64 platform. These were generated with the vixl simulator which messes with the results.

Sonicadvance1:
Rebased and updated with results from a native ARM devices instead of simulator. Should be good for review now.

Side note, we might want to modify vixl's disassembler to not show target PC on branches, or ensure the tests all get loaded at a fixed PC? Not sure exactly what we want to do there.

Sonicadvance1:
Added sanitization of vixl's address generation for the disassembly so we don't get spurious assembly changes in the json files when regenerated. Just does the sanitization inside of the tool itself rather than modifying vixl or trying to allocate the memory in a fixed location all the time.

Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -59,6 +59,17 @@ namespace CodeSize {
   constexpr std::string_view DisassembleEndMessage = "Disassemble End";
   constexpr std::string_view BlowUpMsg = "Blow-up Amt: ";
 
+  static fextl::string SanitizeDisassembly(char const *Message) {
```
lioncash, line 62:
```suggestion
  static fextl::string SanitizeDisassembly(std::string_view Message) {
```

Would allow its current usage to avoid an extra strlen on the string by passing in the existing MessageView local

Sonicadvance1: Switched over to string_view for both the argument and the return. Makes the sanitization quite clean.

lioncash (approved):

PR #2900, opened 2023-08-15
Arm64/VectorOps: Remove redundant move in VFRSqrt SVE path

lioncash:
We can perform the SQRT first and then broadcast 1.0 into the destination since all the intermediary work is done, meaning we don't have to worry about Dst and Vector aliasing one another

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -1018,10 +1018,9 @@ DEF_OP(VFRSqrt) {
 
   if (HostSupportsSVE256 && Is256Bit) {
     const auto Pred = PRED_TMP_32B.Merging();
-    fmov(SubRegSize.Vector, VTMP1.Z(), 1.0);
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
Nice. So as you can probably see, we don't have InstCountCI fail on instructions that aren't considered "optimal".
Didn't want the behaviour to be /too/ aggressive at the start, but if desired we could have it fail on any change to expected instruction count.

Sonicadvance1:
Although ideally you would run `ninja instcountci_tests && ninja instcountci_update_tests` and commit those changes in that same PR that modified codegen here. Failure on any instruction count change might make that desirable to not forget it.

Sonicadvance1:
Ah, you would need `-DENABLE_VIXL_DISASSEMBLER=True -DENABLE_VIXL_SIMULATOR=True -DENABLE_JIT_ARM64=True -DENABLE_JIT_X86_64=False` cmake options so that the vixl disassembler stuff can run.
Should probably tie those tests to those options being set.

lioncash:
Ah, it was a case of me having the disassembler not enabled. Everything's all good now

PR #2908, opened 2023-08-16
IR/ConstProp: Ensure that BFI with constant bitfields can optimize to Andn or Or

Sonicadvance1:
This optimizes the clc and stc instructions for flag setting and
clearing.


alyssarosenzweig:
Who defines "optimal"? Optimal `CLC` would presumably be `RMIF wzr, 31, 1` or something.

Sonicadvance1:
> Who defines "optimal"? Optimal `CLC` would presumably be `RMIF wzr, 31, 1` or something.

No strict guidelines, it's more of a guide about what our current implementations can handle. A lot of instructions marked "Optimal" are considered that way without considering flag handling even.
Once we have flags living in host flags then these loose guidelines will likely change and we'll need to audit the implementations again.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -47,6 +47,19 @@ uint64_t getMask(IROp_Header* Op) {
...
+// Ensuring that the consecutive bits in the range are entirely 0 or 1.
+static bool HasConsecutiveBits(uint64_t imm, unsigned width) {
+  uint64_t StartingBit = imm & 1;
+  for (size_t i = 1; i < width; ++i) {
+    if (((imm >> i) & 1) != StartingBit) {
+      return false;
+    }
+  }
+
+  return true;
```
dougallj, line 60: No need to optimise this, but if you wanted to, it could look like:

```c
  // assumes width <= 64
  if (width == 0) return true;
  return ((imm ^ (imm >> 1)) & (((uint64_t)1 << (width-1)) - 1)) == 0;
```

https://godbolt.org/z/PdWdx7zna

Sonicadvance1: Very clean implementation! Switched out to your implementation.
Quite a bit nicer, thanks!

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -893,6 +904,33 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
...
         Changed = true;
       }
+      else if (SrcIsConstant) {
+        // If the source is constant in a BFI then it can be converted to an or or bic depending.
+        if (ConstantSrc == 0) {
+          // We are wanting to clear the bitfield.
+          IREmit->SetWriteCursor(CodeNode);
+          uint64_t SourceMask = (1ULL << Op->Width) - 1;
+          if (Op->Width == 64)
+            SourceMask = ~0ULL;
```
dougallj, line 914: I think this'll trip UBSAN? You can swap it around or use a ternary expression to avoid the shift when `Op->Width == 64`

Sonicadvance1: This pattern is actually in a few locations in our source. We probably want to just hammer the idiom all at once if we don't want ubsan to get angry.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -893,6 +904,33 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
         IREmit->ReplaceWithConstant(CodeNode, NewConstant);
         Changed = true;
       }
+      else if (SrcIsConstant) {
+        // If the source is constant in a BFI then it can be converted to an or or bic depending.
+        if (ConstantSrc == 0) {
```
dougallj, line 909: The structure here is:

```
if (ConstantSrc == 0) {
  // AND case
} else {
  if (HasConsecutiveBits(ConstantSrc, Op->Width)) {
    // OR case
  }
}
```

I was expecting 

```
if (HasConsecutiveBits(ConstantSrc, Op->Width)) {
  if (ConstantSrc & 1) {
    // OR case
  } else {
    // AND case
  }
}
```

At the moment, I think there's a bug if you try to insert the low 16 bits of 0x10000. 0x10000 is constant, and the low 16 bits are the same (all zero), but 0x10000 != 0, so we end up in the OR case, which doesn't make sense.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -893,6 +904,33 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
...
+          auto andn = IREmit->_Andn(CurrentIR.GetNode(Op->Header.Args[0]), IREmit->_Constant(NewConstant));
+          andn.first->Header.Size = IROp->Size;
+          IREmit->ReplaceAllUsesWith(CodeNode, andn);
+          Changed = true;
+        }
+        else {
+          // We are trying to insert constant, if it is a bitfield of only set bits then we can orr it.
+          if (HasConsecutiveBits(ConstantSrc, Op->Width)) {
+            IREmit->SetWriteCursor(CodeNode);
+            auto orr = IREmit->_Or(CurrentIR.GetNode(Op->Header.Args[0]), IREmit->_Constant(ConstantSrc << Op->lsb));
```
dougallj, line 927: I think `ConstantSrc << Op->lsb` would also have the wrong value if there are any non-zero bits in `ConstantSrc` past `Op->Width`

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
Sonicadvance1: Refactored ConstProp.cpp to remove this idiom from this file at least. Will need to look around in follow-up PRs to remove this in other files.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
Sonicadvance1: You're right it does end up being a little weird flipped the consecutive bits check around and generates the constant only once since both sides will actually use the same constant regardless.

External/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
Sonicadvance1: Resolved this as mentioned in the other comment. Generating the same constant on both halves of andn/orr paths since it actually uses the same constant. Which removes this weirdness.

dougallj (approved):
Looks good! (But a reminder to everyone else that I don't have write access, so this still needs a review to merge.)

lioncash (approved):

PR #2911, opened 2023-08-16
Arm64: Stop abusing orr in LoadConstant 

Sonicadvance1:
The current implementation uses orr excessively. This has FEX missing
hardware optimization opportunities where some CPU cores will zero-cycle
move constants that fit in to the 16-bits of movz/movk.

First evaluate up front if the number of 16-bit segments is > 1, in
those cases we should check if it is a bitfield that can be moved in one
instruction with orr.

After that point we will use movz for 16-bit constant moves.

Additionally this optimizes the case where a constant of zero is loaded
to be a `mov <reg>, zr` which gets renamed in most hardware.


External/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -268,14 +273,30 @@ void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, ui
...
       if (Part) {
-        movk(s, Reg, Part, i * 16);
+        movk(s, Reg, Part, CurrentSegment * 16);
         ++NumMoves;
       }
     }
+
+    if (NumMoves == 0) {
+      // If we didn't move anything that means this is a zero move. Special case this.
+      mov(s, Reg, ARMEmitter::Reg::zr);
```
dougallj, line 297: Unfortunately `mov reg, zr` is slow on Apple CPUs, but as far as I know, `mov reg, #0x0` should be fast everywhere?

Sonicadvance1: Good idea. Switched over to move of constant zero.

lioncash (approved):

PR #2912, opened 2023-08-17
OpcodeDispatcher: Minor optimization around clearing flags 

Sonicadvance1:
When clearing multiple flags it is more optimal to load the mask
constant in to a register and then clear with a single and/bic.

Back to back bfi is actually less optimal due to dependency tracking.

With https://github.com/FEX-Emu/FEX/pull/2911, this is a total win since this hits an edge case with
constant loading that https://github.com/FEX-Emu/FEX/pull/2911 fixes.


Sonicadvance1:
Rebased for conflicting instcountci json.

lioncash:
This needs a rebase

Sonicadvance1:
Rebased

lioncash (approved):

PR #2914, opened 2023-08-17
IR: Add VBroadcastFromMem opcode 

lioncash:
Allows the implementations of the vbroadcast instructions to perform the load and broadcast in one operation as opposed to doing the load and then broadcast separately.

Notably, the broadcasting loads can also be used on systems that have SVE 128-bit support as well, not only 256-bit. On non-SVE systems, the AdvSIMD equivalent LD1R is used instead.

### DIFF
--- a/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -1489,6 +1489,11 @@ DEF_OP(VBroadcastFromMem) {
       return;
     }
   }
+
+  // Emit a half-barrier if TSO is enabled.
+  if (CTX->IsAtomicTSOEnabled()) {
+    dmb(ARMEmitter::BarrierScope::ISHLD);
+  }
 }
 
 DEF_OP(StoreMem) {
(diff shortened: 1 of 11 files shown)


dougallj:
Even without SVE this should be able to use the SIMD instruction [LD1R](https://docsmirror.github.io/A64/2023-06/ld1r_advsimd.html).

lioncash:
Oh, nice! I completely overlooked it when looking for equivalents on the AdvSIMD side of things. Will update it to use that instead.

External/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1429,6 +1429,68 @@ DEF_OP(VStoreVectorMasked) {
   }
 }
 
+DEF_OP(VBroadcastFromMem) {
```
Sonicadvance1, line 1432: This IR operation needs to have a half-barrier after the load operation when TSO is enabled.
Take a look at that LoadMemTSO operation under the vector class handling to see how it works.
```cpp
    // Half-barrier.
    dmb(FEXCore::ARMEmitter::BarrierScope::ISHLD);
```

InstCountCI doesn't test with TSO enabled yet so this wouldn't be visible there.

lioncash: Added it. Is the way I did it a correct way of implementing that functionality, or is a completely separate IR op necessary?

Sonicadvance1: This works fine, no need for another IR operation in this case.

Sonicadvance1 (approved):

PR #2918, opened 2023-08-17
IR: Allow 128-bit broadcasts in VBroadcastFromMem

lioncash:
Now all vbroadcast implementations go down the more optimal path.

For non-SVE 128-bit cases where we only have 128-bit wide registers, we behave like ld1rqb and just act as a normal 128-bit load for interface convenience.


Sonicadvance1 (approved):
Nice

PR #2919, opened 2023-08-17
OpcodeDispatcher: Improve output of {V}MOVSLDUP/{V}MOVSHDUP

lioncash:
Makes things significantly less churny


Sonicadvance1 (approved):
Holy crap that optimization!

PR #2920, opened 2023-08-17
OpcodeDispatcher: Improve VMOVDDUP output

lioncash:
We can make use of TRN1 here to collapse a bunch of these moves.


Sonicadvance1 (approved):
Those predicate moving improvements are nuts.

PR #2926, opened 2023-08-18
CMake: Add mingw toolchain file

bylaws:
This simplifies the setup needed for WIN32-targeted FEX builds and serves as documentation of the workaround for the LLVM linker bug that would otherwise prevent release builds.


Sonicadvance1 (approved):

Sonicadvance1:
Good job keeping it working with mingw CI.

PR #2927, opened 2023-08-18
FEXCore: Allow for interrupting the JIT on block entry

bylaws:
This takes a similar approach to deferred signal handling and allows any given thread to be interrupted while running JIT code by protecting the appropriate page as RO. Then, when the thread then enters a new block, it will try to access that page and segfault. This can then be caught by the frontend where it can recover/edit the context as required.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
@@ -791,6 +791,13 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry,
   adr(TMP1, &JITCodeHeaderLabel);
   str(TMP1, STATE, offsetof(FEXCore::Core::CPUState, InlineJITBlockHeader));
 
+#ifdef _WIN32
+  // Trigger a fault if there are any pending interrupts
+  // Used only for suspend on WIN32 at the moment
+  strb(ARMEmitter::XReg::zr, STATE, offsetof(FEXCore::Core::InternalThreadState, InterruptFaultPage) -
+                                    offsetof(FEXCore::Core::InternalThreadState, BaseFrameState));
+#endif
+
   if (GDBEnabled) {
     auto GDBSize = CTX->Dispatcher->GenerateGDBPauseCheck(CodeData.BlockEntry, Entry);
     CursorIncrement(GDBSize);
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Is there anything that ensures that the `InternalThreadState` object is allocating as being page aligned?

In CreateThread in Core.cpp it's just a basic new.
```cpp
FEXCore::Core::InternalThreadState *Thread = new FEXCore::Core::InternalThreadState{};
```
Feels a little weird, but I'm just saying that since I don't know if this ensures alignment there.
Does this also not run afoul of Windows/wine page protection apis needing 16k alignments? I don't know if page protections work at 4k page size or 16k since I'm just not familiar with the OS.

FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -791,6 +791,10 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry,
   adr(TMP1, &JITCodeHeaderLabel);
   str(TMP1, STATE, offsetof(FEXCore::Core::CPUState, InlineJITBlockHeader));
 
+  // Trigger a fault if there are any pending interrupts
+  strb(ARMEmitter::XReg::zr, STATE, offsetof(FEXCore::Core::InternalThreadState, InterruptFaultPage) -
```
Sonicadvance1, line 795: Maybe wrap this in WIN32 checks until the Linux side catches up to the ability to pause a thread, and also vetting the performance impact on that side (should fall in to noise for one cycle store but would be nice to validate)

Sonicadvance1:
> > Is there anything that ensures that the `InternalThreadState` object is allocating as being page aligned?
> > In CreateThread in Core.cpp it's just a basic new.
> > ```c++
> > FEXCore::Core::InternalThreadState *Thread = new FEXCore::Core::InternalThreadState{};
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Feels a little weird, but I'm just saying that since I don't know if this ensures alignment there. Does this also not run afoul of Windows/wine page protection apis needing 16k alignments? I don't know if page protections work at 4k page size or 16k since I'm just not familiar with the OS.
> 
> Protection on windows has 4kb granularity, it's just file mappings that are 16k so there's no issue there. The `alignas(FHU::FEX_PAGE_SIZE)` I added in the struct ensures that `InterruptFaultPage` is aligned appropriately as it'll make `new` call the `new (size_t, std::align_t)` variant.

Sounds reasonable then

Sonicadvance1 (approved):

PR #2931, opened 2023-08-19
Optimize PSIGN and VBSL

Sonicadvance1:
Noticed that the PSIGN instructions use quite a few temporaries and didn't look optimal at all. Optimize the implementation to be optimal! Which required some VBSL optimizations as well.

With ASIMD FEX would never optimize BSL out of fear if some registers
overlapped it would break things. So it had previously always moved to a
temporary first and then moved the result back out when done.

Now instead check upfront if any of the source registers overlap the
destination. If the destination register overlaps any of the three
sources we can bsl, bit, or bif depending on which register gets
overlapped.

Worst case the destination doesn't overlap any of the source registers
and still needs these moves.


dougallj:
I love to see those `mov`s disappearing!

Worth double-checking this, but I think it should be possible to do:
```
mul(sub(cmlt(Src2, 0), cmgt(Src2, 0)), Src1)
```
That'd be one less instruction, which is nice for size, and translates to better throughput on the M1. The `mul` might be more expensive elsewhere, but it could also be beneficial to avoid the ternary operations in some places.

dougallj:
(Hmm, or can you even replace `sub(cmlt(Src2, 0), cmgt(Src2, 0))` with `srshr(sqshl(Src2, ElementSize-1), ElementSize-1)`? I posted that [last year](https://twitter.com/dougallj/status/1576031028275466241) and then forgot about it.)

dougallj:
Okay, I guess I shouldn't post that as a question, let me try as an explanation instead:

[PSIGN](https://www.felixcloutier.com/x86/psignb:psignw:psignd) is defined as:
```
    if control<0:
        return -input_val
    elif control==0:
        return 0
    return input_val
```

We can rewrite this as:

```
    if control<0:
        return -1 * input_val
    elif control==0:
        return 0 * input_val
    return 1 * input_val
```

Then we just need a function, which we'll call `sign(x)` which returns 0 if control == 0, -1 if control < 0, and 1 if control > 0, and we can define PSIGN as:
```
    return sign(control) * input_val
```

So, how do we get `sign(x)`? One ASIMD instruction we can use is a signed saturating left shift by width - 1. For example, if the width is 8, and we shift left by 7, everything from -0x80 to -1 will become -0x80, everything from 1 to 0x7F will saturate to 0x7F, and only zero will remain 0.

This is a lot like `sign(x)` – the range has been compressed to three values, depending on whether the value is less then, equal to, or greater than zero, but instead of -1, 0, and 1, we currently have -0x80, 0 and 0x7F.

If we signed-shift-right those by 7, we get -1, 0 and... 0. So that doesn't quite work. However, we also have a signed-**rounding**-shift-right instruction. If we're shifting right by seven, this effectively adds (1<<6) (i.e. 0x40) first, and it does so in integer space, without overflow. So we have -0x40, 0x40 and 0xBF shifted right by 7. Which looks a bit weird, but gives us -1, 0 and 1, exactly as we wanted.

So we can write `sign(x)` as:

```
    return SignedRoundingShiftRight(SignedSaturatingShiftLeft(x, ElementSize-1), ElementSize-1);
```

And we can write PSIGN as:
```
    return SignedRoundingShiftRight(SignedSaturatingShiftLeft(control, ElementSize-1), ElementSize-1) * input_val;
```

Sonicadvance1:
Drafting until I can look at @dougallj's recommendation tomorrow since it has the chance of getting psign down to three instructions.

Sonicadvance1:
New optimization looks like it works pretty well. I just need to implement the shifts in the interpreter and x86 JIT so that CI is green. Will do that later today.

dougallj:
Looks great!

Also, in the "Needs moves" BSL case, we've verified that the destination isn't any of the sources, so it should be safe to do it with only one move:

```
        mov(Dst.Q(), VectorMask.Q());
        bsl(Dst.Q(), VectorTrue.Q(), VectorFalse.Q());
```

(I'd be in favour of dropping the x64 JIT and interpreter – I don't really see that they're worth the friction, and you can use the Vixl Simulator for testing and development on other platforms.)

Sonicadvance1:
> Looks great!
> 
> Also, in the "Needs moves" BSL case, we've verified that the destination isn't any of the sources, so it should be safe to do it with only one move:
> 
> ```
>         mov(Dst.Q(), VectorMask.Q());
>         bsl(Dst.Q(), VectorTrue.Q(), VectorFalse.Q());
> ```
> 
> (I'd be in favour of dropping the x64 JIT and interpreter – I don't really see that they're worth the friction, and you can use the Vixl Simulator for testing and development on other platforms.)

Good shout, I stopped thinking about that path since it was "non-optimal" regardless. Nice to remove that final move.

Sonicadvance1:
Instead of implementing the shifts in the interpreter and x86 JIT, I have added a flag that selects one implementation over the other.
Which is kind of nice since if we start micro-profiling we can compare the two implementations since right now it is hard to tell on Cortex if one is faster than the other. The shifts+mul is obviously faster on Apple.

dougallj:
Looks good to me. 

The shifts do have better throughput on Apple, and better code-size everywhere, but it's slightly worse latency than the `sub(cmlt(x, 0), cmgt(x, 0))` implementation of `sign(x)` – only 8c vs 7c. Then 6c latency for your version. So it's still not entirely black and white – each instruction you save improves throughput, but costs 1c latency.

On Cortex-A715, it looks like the shifts are 12c latency, vs 8c for `sub(cmlt(x, 0), cmgt(x, 0))`, and ~6c for your version. Throughput and code-size should still be better, but that's a much bigger difference in latency, so yeah, definitely worth thinking about.

lioncash (approved):

PR #2932, opened 2023-08-19
OpcodeDispatcher: Handle broadcasting cases in VPERMQ/VPERMPD

lioncash:
We can handle element broadcasts much more efficiently than doing iterative inserts.


Sonicadvance1 (approved):
hah, very fancy.
I wasn't paying too much attention to the immediate encodings that I chose here, funny that it did some silly element duplication.

PR #2934, opened 2023-08-19
OpcodeDispatcher: Remove extraneous moves in {V}CVTSD2SS/{V}CVTSS2SD

lioncash:
Since all we're going to be doing is an insert as the final operation, in the cases where our source is a vector, we can specify the size of the vector rather than the size of the element to avoid doing unnecessary zero-extending.

### DIFF
--- a/unittests/InstructionCountCI/Secondary_REP.json
+++ b/unittests/InstructionCountCI/Secondary_REP.json
@@ -275,14 +275,11 @@
       ]
     },
     "cvtss2sd xmm0, xmm1": {
-      "ExpectedInstructionCount": 5,
-      "Optimal": "No",
+      "ExpectedInstructionCount": 2,
+      "Optimal": "Yes",
       "Comment": "0xf3 0x0f 0x5a",
       "ExpectedArm64ASM": [
-        "movi v0.2d, #0x0",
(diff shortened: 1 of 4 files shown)


unittests/InstructionCountCI/Secondary_REP.json
```diff
@@ -275,14 +275,11 @@
       ]
     },
     "cvtss2sd xmm0, xmm1": {
-      "ExpectedInstructionCount": 5,
+      "ExpectedInstructionCount": 2,
```
Sonicadvance1, line 278: Looks like this can be considered optimal without FEAT_AFP now?

unittests/InstructionCountCI/Secondary_REPNE.json
```diff
@@ -224,15 +224,14 @@
       ]
     },
     "cvtsd2ss xmm0, xmm1": {
-      "ExpectedInstructionCount": 3,
+      "ExpectedInstructionCount": 2,
```
Sonicadvance1, line 227: Looks like this can be considered optimal without FEAT_AFP now?

lioncash:
Marked them as optimal

Sonicadvance1 (approved):

PR #2940, opened 2023-08-20
Arm64/EncryptionOps: Use MOVI reg, #0 to zero vectors

lioncash:
This is a little more optimal than XORing the vector by itself.


Sonicadvance1 (approved):
Ah, thanks. I knew I had missed some of these.

dougallj:
Nice!

If you're curious, in a microbenchmark on Apple M1, `eor v1.16b, v1.16b, v1.16b` takes 2 cycles, because it's latency-bound with a false dependency on v1, whereas `movi v1.2d, #0x0` takes 0.125 cycles (throughput-bound, and handled by rename). So that's about 16x faster.

(That might seem exaggerated, and I'm sure you can find cases where it doesn't make a difference, but the false dependency can easily be worse with other code: `eor v1.16b, v1.16b, v1.16b ; mla v1.16b, v2.16b, v3.16b` takes 5 cycles, `movi v1.2d, #0x0 ; mla v1.16b, v2.16b, v3.16b` takes 0.25 cycles, so that's 20x faster.)

PR #2941, opened 2023-08-20
OpcodeDispatcher: Improve SHA1MSG1 output

lioncash:
We can simplify these inserts down to only two inserts.


Sonicadvance1 (approved):

JosJuice:
I think you might be able to replace the inserts with an EXT instruction.

lioncash:
Very nice. That reduces it down to 2 instructions; thanks!

Sonicadvance1 (approved):
Sick~

PR #2943, opened 2023-08-21
IR: Implements support for wide scalar shifts 

Sonicadvance1:
This matches x86 vector shift behaviour closely for ps{rl,ra,ll}{w,d,q}
where the vector is shifted by a scalar value that is 64-bits wide.
Anything larger than the element size will set that element to zero.

With SVE we have some new wide element shifts that match this behaviour
exactly (except supports wide shift sources rather than scalar).

This is a significant improvement even on platforms that only support
128-bit SVE.


Sonicadvance1:
This is now ready for review. I'll leave it up to @lioncash to remove that spurious moves on the AVX side.

dougallj:
I love this! It's a huge improvement :)

But I'm a bit concerned that keeping the x86 JIT around makes these 1-IR-operation implementations the easiest option.

It's probably out-of-scope of this PR, but supposing you have:

```
psraw xmm0, xmm2
psraw xmm1, xmm2
```
With this PR, I assume you get:
```
uqshl d0, d18, #57
ushr d0, d0, #57
dup v0.8h, v0.h[0]
neg v0.8h, v0.8h
sshl v16.8h, v16.8h, v0.8h

uqshl d0, d18, #57
ushr d0, d0, #57
dup v0.8h, v0.h[0]
neg v0.8h, v0.8h
sshl v17.8h, v17.8h, v0.8h
```
But by splitting it into multiple operations in the IR, assuming some basic optimisations, you should be able to get:
```
uqshl d0, d18, #57
ushr d0, d0, #57
dup v0.8h, v0.h[0]
neg v0.8h, v0.8h
sshl v16.8h, v16.8h, v0.8h
sshl v17.8h, v17.8h, v0.8h
```

Unfortunately, splitting things like "SSHL" into their own operations currently requires doing a lot of the work of writing an A64-on-x86 emulator, just to write your x86-on-A64 emulator.

lioncash (approved):

PR #2946, opened 2023-08-21
OpcodeDispatcher: Optimizes mpsadbw 

Sonicadvance1:
Two optimizations here:
1) The final VInsElement was generating three instructions
   - This itself could have been change to vzip, which would have
     removed two instructions.
2) Optimize how the pairwise elements are calculated to shave one
   instruction off the calculation.
   - addp odd elements and even elemnts first
   - Then transpose those elements
   - Then use one final addp to generate the result in the correct
     order.

The ext+uabdl+addp pairs of operations could be reordered to shave off
one temporary register usage if we really care later.


lioncash:
Could do with a rebase to git rid of the end of file line changes

Sonicadvance1:
Rebased to resolve EOF changes.

lioncash (approved):

PR #2947, opened 2023-08-21
OpcodeDispatcher: Optimize SSE/AVX pmaddubsw

Sonicadvance1:
Can be slightly more optimal with a slightly change algorithm but will
require implementing some IR ops which can be put off. It's only about
an instruction savings.

The AVX savings are hilarious.


lioncash (approved):
damn

PR #2949, opened 2023-08-21
OpcodeDispatcher: Optimize phsub 

Sonicadvance1:
This was...surprisingly bad. I blame myself entirely.


Sonicadvance1:
Fixed the bug. Turns out we have never used i32v2 unzips so they were just broken in the x86 JIT.

lioncash (approved):

PR #2951, opened 2023-08-22
OpcodeDispatcher: Handle zero immediate shifts better

lioncash:
In the SSE and lower cases, we don't need to do anything, since the value is already in the destination.

While super unlikely, we can simply do nothing faster regardless


Sonicadvance1 (approved):
Woo doing nothing faster.

PR #2952, opened 2023-08-22
OpcodeDispatcher: Remove redundant moves in AVXVectorALUOp

lioncash:
We already zero-extend on a store if we have 256-bit vectors and the stored vector is 128-bit.


Sonicadvance1 (approved):
Nice, one set of redundant moves gone.

PR #2953, opened 2023-08-22
OpcodeDispatcher: Remove redundant move in AVXVectorScalarALUOpImpl

lioncash:
Our store will already zero-extend if the vector is 128-bit.


Sonicadvance1 (approved):
Yay more redundant moves.

PR #2960, opened 2023-08-23
Arm64: Optimize SVE VInsElement

lioncash:
This can be done without storing any data to memory and also reducing the amount of instructions used.

Thanks to @dougallj for the optimization suggestions.


Sonicadvance1 (approved):
Whew! What a massive change in number of instructions!

PR #2963, opened 2023-08-23
OpcodeDispatcher: Optimize MOVLP{S,D} loads 

Sonicadvance1:
This now uses the new load element IR operation and makes these
instructions optimal.

LRPCPC3 will introduce instructions in the future for TSO emulation to
help these operations, but that doesn't exist today.

### DIFF
--- a/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
+++ b/FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
@@ -418,6 +418,53 @@ DEF_OP(VStoreVectorMasked) {
   }
 }
 
+DEF_OP(VLoadVectorElement) {
+  const auto Op = IROp->C<IR::IROp_VLoadVectorElement>();
+  const auto OpSize = IROp->Size;
+
+  const auto ElementSize = IROp->ElementSize;
+
+  auto *Mem = *GetSrc<uint8_t**>(Data->SSAData, Op->Addr);
+  const auto *DstSrc = GetSrc<uint8_t const*>(Data->SSAData, Op->DstSrc);
+
+  const auto SetElements = []<typename T>(void* Dst, const T* MemPtr, const auto Index) {
+    std::memcpy(static_cast<uint8_t*>(Dst) + (Index * sizeof(T)), MemPtr, sizeof(T));
+  };
+
+  // Copy the source data first.
+  memcpy(GDP, DstSrc, OpSize);
+
+  switch (ElementSize) {
+    case 1: {
+      SetElements(GDP,
+                  Mem,
+                  Op->Index);
+      return;
+    }
+    case 2: {
+      SetElements(GDP,
+                  reinterpret_cast<const uint16_t*>(Mem),
+                  Op->Index);
+      return;
+    }
+    case 4: {
+      SetElements(GDP,
+                  reinterpret_cast<const uint32_t*>(Mem),
+                  Op->Index);
+      return;
+    }
+    case 8: {
+      SetElements(GDP,
+                  reinterpret_cast<const uint64_t*>(Mem),
+                  Op->Index);
+      return;
+    }
+    default:
+      LOGMAN_MSG_A_FMT("Unhandled {} element size: {}", __func__, ElementSize);
+      return;
+  }
+}
+
 DEF_OP(VBroadcastFromMem) {
   const auto Op = IROp->C<IR::IROp_VBroadcastFromMem>();
   const auto OpSize = IROp->Size;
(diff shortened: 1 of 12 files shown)


FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1429,6 +1429,56 @@ DEF_OP(VStoreVectorMasked) {
...
+  const auto MemReg = GetReg(Op->Addr.ID());
+
+  LOGMAN_THROW_AA_FMT(ElementSize == 1 || ElementSize == 2 ||
+                      ElementSize == 4 || ElementSize == 8 ||
+                      ElementSize == 16, "Invalid element size");
+
+  if (Is256Bit) {
+    LOGMAN_MSG_A_FMT("Unsupported 256-bit VLoadVectorElement");
+  } else {
+    if (Dst.Idx() != DstSrc.Idx() && ElementSize != 16) {
```
lioncash, line 1450:
```suggestion
    if (Dst != DstSrc && ElementSize != 16) {
```

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -1429,6 +1429,56 @@ DEF_OP(VStoreVectorMasked) {
...
+      ld1<ARMEmitter::SubRegSize::i16Bit>(Dst.Q(), Op->Index, MemReg);
+      break;
+    case 4:
+      ld1<ARMEmitter::SubRegSize::i32Bit>(Dst.Q(), Op->Index, MemReg);
+      break;
+    case 8:
+      ld1<ARMEmitter::SubRegSize::i64Bit>(Dst.Q(), Op->Index, MemReg);
+      break;
+    case 16:
+      // Normal load, like ld1rqb with 128-bit regs.
```
lioncash, line 1467: This comment should probably be removed since this isn't really related to vector broadcasting

FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
```diff
@@ -418,6 +418,53 @@ DEF_OP(VStoreVectorMasked) {
...
+    std::memcpy(static_cast<uint8_t*>(Dst) + (Index * sizeof(T)), MemPtr, sizeof(T));
+  };
+
+  // Copy the source data first.
+  memcpy(GDP, DstSrc, OpSize);
+
+  switch (ElementSize) {
+    case 1: {
+      SetElements(GDP,
+                  reinterpret_cast<const uint8_t*>(Mem),
```
lioncash, line 440: The cast here isn't necessary, since Mem is already a uint8_t* pointer

FEXCore/Source/Interface/IR/IR.json
```diff
@@ -491,15 +491,20 @@
...
         "Desc": ["Does a masked store similar to VPMASKMOV/VMASKMOV where the upper bit of each element",
                  "determines whether or not that element will be stored to memory"],
         "HasSideEffects": true,
         "DestSize": "RegisterSize",
         "NumElements": "RegisterSize / ElementSize"
       },
-
+      "FPR = VLoadVectorElement u8:#RegisterSize, u8:#ElementSize, FPR:$DstSrc, u8:$Index, GPR:$Addr": {
+        "Desc": ["Does a memory load to a single element of a vector.",
+                 "Leaves the rest of the vector's data intack.",
```
lioncash, line 503:
```suggestion
                 "Leaves the rest of the vector's data intact.",
```

lioncash (commented):
Just a few nits

FEXCore/Source/Interface/IR/IR.json
Sonicadvance1: Fixed.

FEXCore/Source/Interface/Core/Interpreter/MemoryOps.cpp
Sonicadvance1: Fixed.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: Removed.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: Fixed.

Sonicadvance1:
Converted to a draft while I noticed a bug.
```cpp
OrderedNode *Src = LoadSource(FPRClass, Op, Op->Src[0], Op->Flags, 8, false);
```
Was accidentally loading the memory source as an FPR class.
Which this happens to work currently as it silently gets replaced with a GPRClass for addresses like this, but wanted to ensure the logic in the source is correct. We might have a few other cases that mess this up.

lioncash (approved):

PR #2965, opened 2023-08-23
OpcodeDispatcher: Remove unnecessary moves from AVX conversion operations

lioncash:
These zero-extensions will already happen automatically if necessary.


Sonicadvance1 (approved):
Nice, love seeing all these moves melt away.

PR #2972, opened 2023-08-23
OpcodeDispatcher: Optimizes scalar movd/movq

Sonicadvance1:
MMX and SSE versions are now optimal.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -2256,20 +2256,33 @@ void OpDispatchBuilder::VPMASKMOVOp<true>(OpcodeArgs);
 void OpDispatchBuilder::MOVBetweenGPR_FPR(OpcodeArgs) {
   if (Op->Dest.IsGPR() &&
       Op->Dest.Data.GPR.GPR >= FEXCore::X86State::REG_XMM_0) {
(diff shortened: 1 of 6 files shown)


lioncash (commented):
One minor nit, but looks good overall

Sonicadvance1:
> One minor nit, but looks good overall

Nit wasn't posted?

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -2256,20 +2256,33 @@ void OpDispatchBuilder::VPMASKMOVOp<true>(OpcodeArgs);
...
+    else {
+      // Loading from Memory as a scalar. Zero extend
+      OrderedNode *Src = LoadSource(FPRClass, Op, Op->Src[0], Op->Flags, -1);
+      StoreResult(FPRClass, Op, Op->Dest, Src, -1);
+    }
   }
   else {
-    // Destination is GPR or mem
-    // Extract from XMM first
     auto ElementSize = GetDstSize(Op);
```
lioncash, line 2273: Looks like this can be moved into the GPR case below

lioncash:
Weird, had to repost it

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
Sonicadvance1: Moved.

lioncash (approved):

PR #2975, opened 2023-08-23
Arm64/ConversionOps: Add scalar support to Vector_FToI 

lioncash:
This can be used for the scalar conversions instead of always using the vector variants. Also corrects an oversight on the AdvSIMD side of things where rounding nearest would be using `frinti` (host rounding mode) instead of `frintn`.

While we're in the area, we can also remove unnecessary moves on the AVX side of things, where we were doing a manual zero-extension when we didn't need to.


Sonicadvance1 (approved):
Nice catch! I completely glossed over that with my auditing!

PR #2979, opened 2023-08-23
Arm64/VectorOps: Remove moves from SVE VFDiv if possible

lioncash:
Given the operation is:

```
Dst = Vector1 / Vector2
```

If Dst and Vector1 alias one another, then we can just perform the division as is without any moving of data around.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -773,10 +773,17 @@ DEF_OP(VFDiv) {
   if (HostSupportsSVE256 && Is256Bit) {
     const auto Mask = PRED_TMP_32B.Merging();
 
-    // SVE VDIV is a destructive operation, so we need a temporary.
-    movprfx(VTMP1.Z(), Vector1.Z());
-    fdiv(SubRegSize, VTMP1.Z(), Mask, VTMP1.Z(), Vector2.Z());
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -773,10 +773,17 @@ DEF_OP(VFDiv) {
...
-    movprfx(VTMP1.Z(), Vector1.Z());
-    fdiv(SubRegSize, VTMP1.Z(), Mask, VTMP1.Z(), Vector2.Z());
-    mov(Dst.Z(), VTMP1.Z());
+    if (Dst == Vector1) {
+      // Trivial case where we already have source data to be divided in the
+      // destination register. We can just divide by Vector2 and be done with it.
+      fdiv(SubRegSize, Dst.Z(), Mask, Dst.Z(), Vector2.Z());
+    } else {
+      // SVE FDIV is a destructive operation, so we need a temporary
+      // in the event that Dst and Vector1 alias.
```
Sonicadvance1, line 782: This comment is incorrect. You're handling the case that Dst and Vector1 alias above which is the optimal case.
This is handling the case in which the two /don't/ alias.

lioncash: Ah, good catch, updated the comment

Sonicadvance1 (approved):

PR #2989, opened 2023-08-24
Arm64/ConversionOps: Remove redundant moves in AdvSIMD VInsGPR

lioncash:
If Dst and DestVector alias one another, then we don't need to move the vector unnecessarily.


Sonicadvance1 (approved):
Nice

PR #2990, opened 2023-08-24
OpcodeDispatcher: Optimize PMULH{U,}W using new IR operations

Sonicadvance1:
SSE implementations are now optimal.
SVE-128bit operation makes it more optimal.

Interestingly when SVE-256bit is supported, the 128-bit operation gets worse.


Sonicadvance1:
Massive thanks to @dougallj as usual for calling out my bad use of shrn+shrn2 and showing instead that uzp2 is a better solution.

lioncash (approved):

PR #3002, opened 2023-08-24
OpcodeDispatcher: Optimize 128-bit movmaskpd 

Sonicadvance1:
I'd consider this optimal now.

Thanks to @dougallj for the optimization idea again!


dougallj:
`orr x20, x20, x20, lsl #31` -> `bfi x20, x20, #31, #32`, I think

lioncash (approved):

PR #3012, opened 2023-08-26
OpcodeDispatcher: Optimizes SSE movmaskps 

Sonicadvance1:
This now improves the instruction implementation from 17 instructions
down to 5 or 6 depending on if the host supports SVE.

I would say this is now optimal.


lioncash (approved):

FEXCore/Source/Interface/IR/IR.json
```diff
@@ -1430,15 +1430,15 @@
         "DestSize": "RegisterSize",
         "NumElements": "RegisterSize / (ElementSize << 1)"
       },
-      "FPR = VUShl u8:#RegisterSize, u8:#ElementSize, FPR:$Vector, FPR:$ShiftVector": {
+      "FPR = VUShl u8:#RegisterSize, u8:#ElementSize, FPR:$Vector, FPR:$ShiftVector, i8:$RangeCheck": {
```
lioncash, line 1433: aren't bools supposed to be indicated with i1, or are these synonymous?

Sonicadvance1: You're right, I should switch it over to i1.

Sonicadvance1: Switched to i1.

lioncash (approved):

PR #3014, opened 2023-08-27
IR: Convert all Move+Atomic+ALU ops from implicit to explicit size

Sonicadvance1:
The number of times the implicit size calculation in GPR operations has bit us is immeasurable and was a mistake from the start of the project. The vector based operations never had this problem since they were explicitly sized for a long time now.

This converts the base IR operations to be explicitly sized, but adds implicit sized helpers for the moment while we work on removing implicit usage from the OpcodeDispatcher.

Should be NFC at this moment but it is a big enough change that I want it in before the "real" work starts.

### DIFF
--- a/FEXCore/include/FEXCore/IR/IR.h
+++ b/FEXCore/include/FEXCore/IR/IR.h
@@ -536,6 +536,17 @@ enum IndexNamedVectorConstant : uint8_t {
   INDEXED_NAMED_VECTOR_MAX,
 };
 
+// This must directly match bytes to the named opsize.
+// Implicit sized IR operations does math to get between sizes.
+enum OpSize : uint8_t {
+  i8Bit = 1,
+  i16Bit = 2,
+  i32Bit = 4,
+  i64Bit = 8,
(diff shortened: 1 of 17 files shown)


FEXCore/include/FEXCore/IR/IR.h
```diff
@@ -536,6 +536,17 @@ enum IndexNamedVectorConstant : uint8_t {
   INDEXED_NAMED_VECTOR_MAX,
 };
 
+// This must directly match bytes to the named opsize.
+// Implicit sized IR operations does math to get between sizes.
+enum OpSize : uint8_t {
```
lioncash, line 541: Should this be an enum class?

Sonicadvance1: It will be once the implicit logic is completely gone. Since it gets converted to a u8 inside of the IR data member currently, that would have been a bigger structural undertaking that's easier to do after all the implicit logic is gone.

lioncash (approved):

PR #3017, opened 2023-08-28
Remove implicit sized IR ops part 1

Sonicadvance1:
Needs #3016 merged first.

Converts a significant portion of the easier operations. The next might be getting a bit more complex so I want to split it up in more easily consumed PRs.

I ran InstCountCI on every commit as I was writing this to ensure codegen hasn't changed.


FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5592,14 +5593,14 @@ void OpDispatchBuilder::TZCNT(OpcodeArgs) {
...
-  auto Res = _CountLeadingZeroes(Src);
+  auto Res = _CountLeadingZeroes(IR::SizeToOpSize(GetSrcSize(Op)), Src);
   StoreResult(GPRClass, Op, Res, -1);
   GenerateFlags_LZCNT(Op, Src);
 }
 
 void OpDispatchBuilder::MOVBEOp(OpcodeArgs) {
   OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags, 1);
-  Src = _Rev(Src);
+  Src = _Rev(IR::SizeToOpSize(GetSrcSize(Op)), Src);
```
alyssarosenzweig, line 5603: Can we get a helper for `IR::SizeToOpSize(GetSrcSize(Op))`? It seems to clutter things.

Maybe `IR::OpSizeFromSrc(Op)`?

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1171,7 +1171,7 @@ friend class FEXCore::IR::PassManager;
     if (SetBits == 0)
       return _Lshl(Value, _Constant(Bit));
     else if (CTX->BackendFeatures.SupportsShiftedBitwise && (SetBits & (1u << Bit)) == 0)
-      return _Orlshl(NZCV, Value, Bit);
+      return _Orlshl(IR::SizeToOpSize(std::max(GetOpSize(NZCV), GetOpSize(Value))), NZCV, Value, Bit);
```
alyssarosenzweig, line 1174: This is intended to always be 32-bit operation but it might not matter in practice.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Couldn't add it to `IR::` namespace since `X86Tables::DecodedOp` and IR operation sizes are legally distinct entities, but added the helper to `OpcodeDispatcher` so we can use it all in there.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1133,7 +1133,7 @@ friend class FEXCore::IR::PassManager;
 
     // Shift the sign bit into the N bit
     if (SignBit > NBit)
-      Shifted = _Ashr(Res, _Constant(SignBit - NBit));
+      Shifted = _Ashr(SrcSize == 8 ? OpSize::i64Bit : OpSize::i32Bit, Res, _Constant(SignBit - NBit));
```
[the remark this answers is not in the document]
Sonicadvance1: Ah cool, changed it to be a 64-bit operation and nothing changed with InstCountCI.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: Yea, I noticed that but I'm leaving it alone for now since I don't want codegen to change with these right now. Tackling one problem at a time.
Changing it to 32-bit only operation causes InstCountCI churn half-way through this process but we can make sure to come back to this once implicit size is gone.

Sonicadvance1:
Rebased, addressed current comments.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: https://github.com/FEX-Emu/FEX/issues/3019 To ensure we don't forget

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
alyssarosenzweig: Sounds good, thanks

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
alyssarosenzweig: Seems reasonable, thx

alyssarosenzweig (approved):

PR #3035, opened 2023-08-30
Optimize ADD flag calculation

alyssarosenzweig:
There's An Op For That


alyssarosenzweig:
Const prop is broken. InstCountCI caught that, nice

alyssarosenzweig:
Segment register code is changing. I don't understand why. @Sonicadvance1 ?

Sonicadvance1:
> Segment register code is changing. I don't understand why. @Sonicadvance1 ?

Oop, need to disable a cmake option `-DENABLE_OFFLINE_TELEMETRY=False`, I should probably add a runtime option so this isn't necessary.

FEXCore/Source/Interface/IR/IR.json
```diff
@@ -895,6 +895,10 @@
           "Size == FEXCore::IR::OpSize::i32Bit || Size == FEXCore::IR::OpSize::i64Bit"
         ]
       },
+      "GPR = AddNZCV u8:$Size, GPR:$Src1, GPR:$Src2": {
```
Sonicadvance1, line 898: This should have an EmitValidation section like the IR op above it so the IR validation can also know that it only supports 32bit and 64bit operation.

Sonicadvance1: Probably also use `OpSize:#Size` instead of `u8:$Size`

alyssarosenzweig:
Discussion question: would we rather invert CF as stored in the NZCV register? That would mean adding a cfinv to every add, but saving the cfinv from every subtract/compare. The flag consumers the compare could probably tolerate either form. Given that compares are the dominant source of flags, that might come out ahead.

This is the convention that Rosetta uses. That doesn't mean it's the right call but it's not a bad idea either. I don't think this should be addressed in this PR but it's worth further thought for later.

Sonicadvance1:
> Discussion question: would we rather invert CF as stored in the NZCV register? That would mean adding a cfinv to every add, but saving the cfinv from every subtract/compare. The flag consumers the compare could probably tolerate either form. Given that compares are the dominant source of flags, that might come out ahead.
> 
> This is the convention that Rosetta uses. That doesn't mean it's the right call but it's not a bad idea either. I don't think this should be addressed in this PR but it's worth further thought for later.

This will likely need some statistics gathering from random applications to see which would be the bigger win. Hard to know for sure.

Sonicadvance1:
Did some minor testing on this branch and it seems to break something in Steam. Will want to be careful when merging this.

alyssarosenzweig:
Answer:

```c++
  // TODO: Can use OpSize calculated up front.
  //auto OpSize = SrcSize == 8 ? OpSize::i64Bit : OpSize::i32Bit;
  auto OpSize = IR::SizeToOpSize(std::max<uint8_t>(4, std::max(GetOpSize(Src1), GetOpSize(Src2))));
````

Either way should(?) not be a regression from this particular change. Maybe. Although.. hm

**Comment by alyssarosenzweig** [comment]
Seems like it, at the very least, has the potential to screw up the carry/overflow calculations.

**Comment by alyssarosenzweig** [comment]
Pushed an untested patch on top to find out what happens cuz your CI is faster than my Chromebook

File: unittests/InstructionCountCI/PrimaryGroup.json [hunk 2]
```diff
@@ -3375,43 +3287,36 @@
...
       "Optimal": "No",
       "Comment": "GROUP2 0xf7 /2",
       "ExpectedArm64ASM": [
         "mov w20, #0x0",
         "lsr w21, w7, #0",
         "neg w7, w21",
-        "eor x22, x20, x21",
-        "eor x22, x22, x7",
-        "ubfx x22, x22, #4, #1",
+        "eor w22, w20, w21",
```
alyssarosenzweig, line 3306: Note to self: fix const prop to further improve this code?

unittests/InstructionCountCI/PrimaryGroup.json
```diff
@@ -3375,43 +3287,36 @@
...
-        "mrs x22, nzcv",
-        "cmp x20, x21",
-        "cset x23, lo",
-        "orr x22, x22, x23, lsl #29",
-        "eor x21, x20, x21",
-        "eor x20, x7, x20",
-        "and x20, x21, x20",
-        "ubfx x20, x20, #31, #1",
-        "orr x20, x22, x20, lsl #28",
+        "cmp w20, w21",
```
alyssarosenzweig, line 3312: Note to self: extend const prop to handle this so we translate to a `NEGS` instruction instead

alyssarosenzweig:
CI is happy.. @Sonicadvance1 please retest Steam, hopefully that was the problem

Sonicadvance1:
> CI is happy.. @Sonicadvance1 please retest Steam, hopefully that was the problem

Can confirm, Steam is now happy.

Sonicadvance1 (approved):
Very nice improvements! Nice to see flag generation getting optimized.
Gave this a bunch of testing and I don't see any regressions.

PR #3038, opened 2023-08-30
Defer AF extract

alyssarosenzweig:
Save 1 instruction for all add/sub/cmp-like instructions by deferring part of the AF calculation. on top of #3035

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -227,9 +227,9 @@ namespace FEXCore::Context {
 
     // Currently these flags just map 1:1 inside of the resulting value.
     for (size_t i = 0; i < FEXCore::Core::CPUState::NUM_EFLAG_BITS; ++i) {
-      if (i == X86State::RFLAG_PF_LOC) {
+      if (i == X86State::RFLAG_PF_LOC || i == X86State::RFLAG_AF_LOC) {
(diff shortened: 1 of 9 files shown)


Sonicadvance1:
To be sure that I'm reading this correctly. Currently it isn't deferring beyond the current EFLAGs deferring that upstream already does?
Looks like right now it's adding accessors, and CalculateAF which the deferred system uses instead of each instruction having its own open-coded implementation of calculating AF. Which is why there isn't any InstCountCI changes.

alyssarosenzweig:
> To be sure that I'm reading this correctly. Currently it isn't deferring beyond the current EFLAGs deferring that upstream already does? Looks like right now it's adding accessors, and CalculateAF which the deferred system uses instead of each instruction having its own open-coded implementation of calculating AF. Which is why there isn't any InstCountCI changes.

Sorry for the awkward timing, for 20 minutes, I had dropped the patch to bisect via CI. It's repushed now. It looks like the unit tests are happy but the glibc fault tests aren't.. I'm combing thru this to try to figure out what's going subtly wrong but coming up empty..

Sonicadvance1:
ah okay, no problem.

alyssarosenzweig:
Fixed the signals issue, this should be good to go now! :smile:

Sonicadvance1 (approved):
Awesome stuff!

PR #3039, opened 2023-08-30
IR: Adds printer for OpSize

Sonicadvance1:


alyssarosenzweig (approved):
friends don't let friends C++

PR #3042, opened 2023-08-30
OpcodeDispatcher: Optimize calls with push

Sonicadvance1:
InstCountCI doesn't cover branch instructions so needs manual inspection.

Branch instruction assembly:
```asm
; Header
adr x0, #-0x4 (addr 0xffffe228005c)
str x0, [x28, #184]
; Calculate return address
mov w20, #0xa
movk w20, #0x1, lsl #16
; Calculate target RIP (Only +1 in this instance)
add x21, x20, #0x1 (1)
; Push ret
str x20, [x8, #-8]!
; Load L1 code cache base
ldr x0, [x28, #1864]
; Calculate offset
and x3, x21, #0xfffff
add x0, x0, x3, lsl #4
; Load 128-bit values from base + offset
ldp x1, x0, [x0]
; Compare if L1 target RIP is current RIP
cmp x0, x21
; b.ne to dispatcher
b.ne #+0x8 (addr 0xffffe2280094)
; Jump to block target
br x1
; Dispatcher path. Load dispatcher PTR
ldr x0, [x28, #1776]
; Store current RIP to CPU state so dispatcher can pick it up
str x21, [x28]
; Branch to dispatcher.
br x0
```

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -789,16 +789,13 @@ void OpDispatchBuilder::CALLOp(OpcodeArgs) {
   OrderedNode *JMPPCOffset = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags, -1);
 
   OrderedNode *NewRIP = _Add(IR::SizeToOpSize(GPRSize), ConstantPC, JMPPCOffset);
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Side note. The Call operation was calculating the return address twice before this optimization. So every call gets an additional 2-4 instructions removed from it. So up to 5 instructions removed per call.

lioncash (approved):

PR #3044, opened 2023-08-31
Arm64: Optimize AES operations by caching a zero register 

Sonicadvance1:
A bunch of the AES operations take a zero register upfront and we
currently materialize it for each instruction.
Considering that most AES operations are used back to back, we can
eliminate these materializations by caching it between instructions.

Additionally removes a move in the optimal case when destination matches
the state register, which is exactly what the SSE operation ends up
doing.

AESKeyGenAssist has an edge case that if the destination RA overlaps the
zero register then we still need to eat a move, hopefully doesn't happen
too frequently in practice. This is also the lesser used instruction so
it isn't a big deal. RA constraints could solve that still.

Looking between the disassembly trees for back to back operations:
Original:
```asm
aesenc  xmm2, xmm1
aesenc  xmm3, xmm1
aesenc  xmm4, xmm1
aesenclast xmm2, xmm0
aesenclast xmm3, xmm0
aesenclast xmm4, xmm0
```

Result:
```asm
movi v4.2d, #0x0
unimplemented (Unimplemented)
unimplemented (Unimplemented)
eor v18.16b, v18.16b, v17.16b
unimplemented (Unimplemented)
unimplemented (Unimplemented)
eor v19.16b, v19.16b, v17.16b
unimplemented (Unimplemented)
unimplemented (Unimplemented)
eor v20.16b, v20.16b, v17.16b
unimplemented (Unimplemented)
eor v18.16b, v18.16b, v16.16b
unimplemented (Unimplemented)
eor v19.16b, v19.16b, v16.16b
unimplemented (Unimplemented)
eor v20.16b, v20.16b, v16.16b
```

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -1446,11 +1446,11 @@ OrderedNode* OpDispatchBuilder::InsertPSOpImpl(OpcodeArgs, const X86Tables::Deco
 
   // ZMask happens after insert
   if (ZMask == 0xF) {
-    return _VectorZero(DstSize);
+    return LoadAndCacheNamedVectorConstant(DstSize, FEXCore::IR::NamedVectorConstant::NAMED_VECTOR_ZERO);
   }
 
   if (ZMask) {
-    auto Zero = _VectorZero(DstSize);
+    auto Zero = LoadAndCacheNamedVectorConstant(DstSize, FEXCore::IR::NamedVectorConstant::NAMED_VECTOR_ZERO);
     for (size_t i = 0; i < 4; ++i) {
       if ((ZMask & (1 << i)) != 0) {
         Dest = _VInsElement(DstSize, 4, i, 0, Dest, Zero);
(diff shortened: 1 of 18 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -4451,7 +4456,13 @@ void OpDispatchBuilder::VBLENDPDOp(OpcodeArgs) {
     return;
   }
 
-  OrderedNode *Result = VBLENDOpImpl(*this, DstSize, 8, Src1, Src2, Selector);
+  OrderedNode *Result = VBLENDOpImpl(*this,
+                                     DstSize,
+                                     8,
+                                     Src1,
+                                     Src2,
+                                     LoadAndCacheNamedVectorConstant(DstSize, FEXCore::IR::NamedVectorConstant::NAMED_VECTOR_ZERO),
```
lioncash, line 4464: Maybe hoist this into its own var so everything can stay compact. Ditto for the others

Sonicadvance1: Hoisted.

lioncash (approved):

PR #3047, opened 2023-09-02
SignalDelegator: Fix build with telemetry disabled

alyssarosenzweig:
/home/alyssa/FEX/Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp:1546:7: error: use of undeclared identifier 'CrashMask'
      CrashMask |= (1ULL << Signal);


Sonicadvance1 (approved):
Oops!

PR #3048, opened 2023-09-02
Context: Adds helper to reconstruct and consume packed EFLAGS

Sonicadvance1:
Currently FEX's internal EFLAGS representation is a perfect 1:1 mapping between bit offset and byte offset. This is going to change with #3038. There should be no reason that the frontend needs to understand how to reconstruct the compacted flags from the internal representation.

Adds context helpers and moves all the logic to FEXCore. The locations that previously needed to handle this have been converted over to use this.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -105,6 +105,8 @@ namespace FEXCore::Context {
       void HandleCallback(FEXCore::Core::InternalThreadState *Thread, uint64_t RIP) override;
 
       uint64_t RestoreRIPFromHostPC(FEXCore::Core::InternalThreadState *Thread, uint64_t HostPC) override;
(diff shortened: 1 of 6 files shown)


alyssarosenzweig:
Is `main` broken wrt packed NZCV?

Sonicadvance1:
> Is `main` broken wrt packed NZCV?

Yep.

alyssarosenzweig:
> > Is `main` broken wrt packed NZCV?
> 
> Yep.

:cool:

Do you want to fix that as a second commit in this PR? Or should we merge like this and I'll poke at it?

Sonicadvance1:
> > > Is `main` broken wrt packed NZCV?
> > 
> > 
> > Yep.
> 
> 🆒
> 
> Do you want to fix that as a second commit in this PR? Or should we merge like this and I'll poke at it?

Merge it as is for now and then you can poke at it since you'll be doing NZCV and AF anyway. It's no more broken than it has been at least.

alyssarosenzweig (approved):

PR #3049, opened 2023-09-02
FEXCore: Rework X87 tag word handling

bylaws:
The FXSAVE and FSAVE tag words are written out in different formats, with FXSAVE using an abridged version that lacks the zero/special/valid distinction. This is currently emulated incorrectly, with both FSAVE and FXSAVE writing out the full format. Fix this and switch to using this abridged version internally for simplicity, and to allow the calculation of zero/special/valid distinction to be deferred until an fxsave instruction (in the future, currently the distinction is ignored and only valid/empty states are possible).


bylaws:
Note the massive blowup on save/load here, but considering how rarely save is used i don't see it being a real issue

Sonicadvance1 (approved):
Very cool! I didn't realize there was a difference here. I was definitely confused back when implementing the save function and wondering why it overlapped the "reserved" byte.
Took me a bit but I found where this is defined under section `FXSAVE Format for x87 Tag Word`

Accepting as-is, we can work to optimize some of this in the future. Thanks!

Issue #2688, opened 2023-05-23
The Witcher 3 slow block


Sonicadvance1:
![Image_2023-05-22_20-44-23](https://github.com/FEX-Emu/FEX/assets/1018829/d836405c-10d0-44b8-95f9-a323a51f8702)
![Image_2023-05-22_20-44-35](https://github.com/FEX-Emu/FEX/assets/1018829/97f08c51-35c9-4f72-98d1-8769a0ad8a6e)
![Image_2023-05-22_20-44-41](https://github.com/FEX-Emu/FEX/assets/1018829/812e4b7c-3bb9-4473-9b32-416cef739e97)
Will populate this with more documentation in the future.

dougallj:
Interesting! While the block is dominated by flag calculcations (for `add rdx, rcx`, from `CalculcateFlags_ADD`), I don't think that's the issue. My first clue was that the following block will also have flag calculcations (for `cmp rax, rcx`), and will execute at least as many times as this block.

It looks like you're using a sampling profiler, and it's pointing squarely at the instruction immediately after the `ldaddal`. So I'd say the issue is with the `ldaddal`: a slow atomic operation, presumably either because of the hardware implementation, or because of software issues leading to contention. While it's not my area of expertise, the translation of the atomic instruction looks fine, so I doubt there's much FEX can do to this block to improve performance?

Sonicadvance1:
Indeed, the `lock xadd` instruction translates to the `ldaddal` semantically so that part can't really be improved. I haven't looked too closely at this one but since it gets executed a /lot/ over the period of about a minute, just getting the converted code to a more ideal size should help significantly regardless. Which will end up being multiple optimizations stacked on top of each other.

Sonicadvance1:
Also looked in to why its startup time is now absolute garbage.
Turns out Xess has some function that is generating a 261,432 byte frame on the stack one byte at a time. This causes our RA's `CalculateNodeInterference` to absolutely explode in CPU time at startup.

Example attached of part of the function that is executed at startup. I would need to scroll for minutes to cover it all.
https://github.com/FEX-Emu/FEX/assets/1018829/6fedd138-1f00-4b31-9f54-a2de9a829dfc

Sonicadvance1:
```asm
%ifdef CONFIG
{
  "MemoryRegions": {
    "0x200000000": "0x20000"
  }
}
%endif

mov rsp, 0x200000000
%assign i 0
%rep 0x10000
[...]

PR #3056, opened 2023-09-03
ConstProp: Adds constpool distance heuristic

Sonicadvance1:
FEX has a problem with large blocks that uses a ton of constants spread throughout the block. Once a block gets large enough with enough constants that have large live ranges, FEX slows down to unusable speeds due to the register allocator spending more time calculating node interferences than anything else in the program.

This adds a little heuristic to ensure that constants aren't reused if the previous value is past a certain distance threshold. This threshold works well enough that XeSS's pedantic initialization code doesn't have issues now. See https://github.com/FEX-Emu/FEX/issues/2688 for more information about that.

FEX itself should work to remove bad constant usages to make this pass less necessary anyway. In most cases we are materializing duplicated 0, 1, and masks which could be done without a constant entirely. Maybe once we've improve that enough we could remove this constant pooling entirely.

To note, this doesn't fix the issue that XeSS causes our register allocator, this is purely a heuristic workaround.

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -208,7 +208,11 @@ private:
       OrderedNode* CodeNode, IROp_Header* IROp);
   bool ConstantInlining(IREmitter *IREmit, const IRListView& CurrentIR);
 
-  fextl::unordered_map<uint64_t, OrderedNode*> ConstPool;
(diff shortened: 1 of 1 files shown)


alyssarosenzweig:
Proper RA should be able to chew through this by rematerializing constants, which fits in nicely with the design we've discssued. But as a stop gap, yeah, this seems reasonable

alyssarosenzweig (approved):

PR #3057, opened 2023-09-05
FEXInterpreter: Supports procfs/interpreter

Sonicadvance1:
This is a new procfs symlink path that changes behaviour of binfmt_misc when exposed. We need to check both procfs/exe and procfs/interpreter and see if they exist AND also differ.

Once/if they do then we can disable a bunch of checking of paths once they do. The fallback when none of this is supported has the same behaviour has previously where it still does all the regular checking.

During binfmt_misc install cmake will check the kernel version for the raw binfmt_misc writing. Which will never pass until we have a real kernel version that it is upstreamed in.

For update-binfmts we add a new optional argument where the tool will drop the flag if the host kernel version isn't new enough to handle the option.


alyssarosenzweig (approved):

Amanieu:
Where can I find the kernel patches for this?

Sonicadvance1:
> Where can I find the kernel patches for this?

Patches are on the way. Since I'm not writing the kernel patches myself they aren't yet visible.

Sonicadvance1:
> Where can I find the kernel patches for this?

https://lore.kernel.org/lkml/20230907204256.3700336-1-gpiccoli@igalia.com/
Patches are now on the ML.

PR #3058, opened 2023-09-05
Stop zeroing undefined flags

alyssarosenzweig:


Sonicadvance1:
Good idea! Before merging this I'm going to do some quick testing on it.

Sonicadvance1 (approved):
Looks like everything is good. We could probably go through all the flag setters because I know some are missing their undefined handling.

PR #3059, opened 2023-09-05
Defer second XOR for AF

alyssarosenzweig:
a @dougallj brilliance. on top of #3058 and continues the work logically started in #3038


Sonicadvance1:
Needs a rebase

Sonicadvance1 (approved):
What a fancy optimization!
If you want to remove the `ABINoPF` hack then be my guest, I never liked it to begin with.

PR #3062, opened 2023-09-05
Remove ABINoPF option

alyssarosenzweig:
Now that PF calculation is deferred, the cost of calculating PF correctly should be tolerable. Remove the speed hack to skip PF. It's fundamentally broken, and there are enough broken things in FEX as it is that we don't need to maintain this one ;-)


Sonicadvance1 (approved):
Yay! \o/

PR #3064, opened 2023-09-05
IR: Remove phi nodes

alyssarosenzweig:
It turns out that pure SSA isn't a great choice for the sort of emulation we do. On one hand, it discards information from the guest binary's register allocation that would let us skip stuff. On the other hand, it doesn't have nearly as many benefits in this setting as in a traditional compiler... We really *don't* want to do global RA or really any global optimization. We assume the guest optimizer did its job for x86, we just need to clean up the mess left from going x86 -> arm. So we just need enough SSA to peephole optimize.

My concrete IR proposals are that:

  * SSA values must be killed in the same block that they are defined.
  * Explicit LoadGPR/StoreGPR instructions can be used for global persistence.
  * LoadGPR/StoreGPR are eliminated in favour of SSA within a block.

This has a lot of nice properties for our setting:

  * Except for some internal REP instruction emulation (etc), we already have registers for everything that escapes block boundaries, so this form is very easy to go into -- straightforward local value numbering, not a full into SSA pass.

  * Spilling is entirely local (if it happens at all), since everything is in registers at block boundaries. This is excellent, because Belady's algorithm lets us spill nearly optimally in linear-time for individual blocks. (And the global version of Belady's algorithm is massively more complicated...) A nice fit for a JIT.

    Relatedly, it turns out allowing spilling is probably a decent decision, since the same spiller code can be used to rematerialize constants in a straightforward way. This is an issue with the current RA.

  * Register assignment is entirely local. For the same reason, we can assign registers "optimally" in linear time & memory (e.g. with linear scan). And the impl is massively simpler than a full blown SSA-based tree scan RA. For example, we don't have to worry about parallel copies or coalescing phis or anything. Massively nicer algorithm to deal with.

  * SSA value names can be block local which makes the validation implicit :~)

It also has remarkably few drawbacks, because we didn't want to do CFG global optimization anyway given our time budget and the diminishng returns. The few global optimizations we might want (flag escape analysis?) don't necessarily benefit from pure SSA anyway.

Anyway, we explicitly don't want phi nodes in any of this. They're currently unused. Let's just remove them so nobody gets the bright idea of changing that.


Sonicadvance1 (approved):
bye bye Phis, ye were hardly used.

PR #3066, opened 2023-09-06
FileManagement: Fix inverted boolean check for procfs/interpreter support

neobrain:
Fixes a critical regression from #3057.


Sonicadvance1 (approved):
Reproduced and confirmed this fixes the issue.

PR #3067, opened 2023-09-07
Thunks: Minor restructuring and small cleanups

neobrain:
Just a collection of minor changes with no functional changes. Mostly moves things around.

Parsing of thunk `interface` files is now decoupled from code generation, since both modules will be significantly extended in immediate followup PRs. `analysis.cpp` will take care of data layout analysis (factoring in related annotations), whereas `gen.cpp` will continue to deal with code generation (without direct access to annotations).

677b77f1bb61b05ccf1c7dc0249811c1fdab11df cuts the size of PackedArguments.h's `Invoke` in half by adding a small metaprogramming trick to make `void` appear like a valid variable type. I'm not a fan of adding such arcane tricks en masse, but the savings justify the small addition here.


Sonicadvance1 (approved):
Nice.

PR #3068, opened 2023-09-07
unittests: Add test thunk library

neobrain:
This adds a new FEXLinuxTest written against a custom `libfex_thunk_test` library. `libfex_thunk_test` allows us to define interfaces specifically to trigger thunk generator edge cases that couldn't easily be tested for in actual libraries.

The new test opens a third category of thunk-related tests:
* unittests/ThunkLibs: Unit tests that check the syntax tree of `thunkgen` code output and (in the future) results of data layout analysis (no runtime behavior is checked)
* unittests/ThunkFunctionalTests: Integration tests that runs simple applications (e.g. glxinfo) to verify basic functionality of real-world libraries works
* FEXLinuxTests/thunk_testlib: Mixed unit/integration tests that benefit from us controlling the interfaces in `libfex_thunk_test`. Individual properties of the thunk generator can be checked by co-designing tests and the tested thunk to trigger problematic edge-cases easily.

Ideally the new library and the test should only be enabled if both `BUILD_THUNKS` and `BUILD_FEX_LINUX_TESTS` are set. However, forwarding these settings through our various `ExternalProject`s would be overly complicated. Instead, the new library is always built if `BUILD_THUNKS` is on, and the new test is always built if `BUILD_FEX_LINUX_TESTS` is on (but excluded from the `fex_linux_tests` umbrella target unless `BUILD_THUNKS` is enabled). Since building these targets doesn't do any harm either way, this is a reasonable middle ground.

TODO:
- [x] The tests fail unless thunking of `libfex_thunk_test` is enabled. Should we force-enable thunking of this library in FEX, or is it sufficient to enable it in CI?
    - Resolved: The test targets now pass the `-k` switch to FEXLoader to enable thunking of the test library

### DIFF
--- a/CI/FEXLinuxTestsThunks.json
+++ b/CI/FEXLinuxTestsThunks.json
@@ -1,0 +1,5 @@
+{
+  "ThunksDB": {
+    "fex_thunk_test": 1
+  }
+}
--- a/Data/ThunksDB.json
+++ b/Data/ThunksDB.json
@@ -144,6 +144,12 @@
         "@PREFIX_LIB@/libasound.so.2.0.0"
       ]
     },
+    "fex_thunk_test": {
+      "Library": "libfex_thunk_test-guest.so",
+      "Overlay": [
+        "@PREFIX_LIB@/libfex_thunk_test.so"
+      ]
+    },
     "Xrender": {
       "Library": "libXrender-guest.so",
       "Overlay": [
--- a/ThunkLibs/GuestLibs/CMakeLists.txt
+++ b/ThunkLibs/GuestLibs/CMakeLists.txt
@@ -291,3 +291,6 @@ if (BITNESS EQUAL 32)
   #  338:   0f 0b                   ud2
   target_compile_options(VDSO-guest PRIVATE "-fno-pic")
 endif()
+
+generate(libfex_thunk_test ${CMAKE_CURRENT_SOURCE_DIR}/../libfex_thunk_test/libfex_thunk_test_interface.cpp)
+add_guest_lib(fex_thunk_test "libfex_thunk_test.so")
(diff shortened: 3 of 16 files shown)


neobrain:
@Sonicadvance1 This requires thunks to be built (`BUILD_THUNKS=ON`), installed (`ninja install`), and enabled (in `thunks.json`) for testing to work properly. Is our CI set up for this already, or will this need some adjustments?

Sonicadvance1:
> @Sonicadvance1 This requires thunks to be built (`BUILD_THUNKS=ON`), installed (`ninja install`), and enabled (in `thunks.json`) for testing to work properly. Is our CI set up for this already, or will this need some adjustments?

Thunks are built and installed on all the runners: https://github.com/FEX-Emu/FEX/blob/main/.github/workflows/ccpp.yml#L76

Thunks themselves are only tested on the x86 runner none of the ARM runners has a GPU in it: https://github.com/FEX-Emu/FEX/blob/main/.github/workflows/ccpp.yml#L234

neobrain:
Thunking of the new `libfex_thunk_test` is now always enabled for FEXLinuxTests. This is implemented by passing the `-k` flag when invoking FEXLoader for these tests, similar to how it's done in the ThunkFunctionalTests.

neobrain:
CI is failing on what looks like pre-existing flakes. The actual changes in this PR are good to go.

neobrain:
There we go, now truly passing CI.

> Thunking of the new libfex_thunk_test is now always enabled for FEXLinuxTests. This is implemented by passing the -k flag when invoking FEXLoader for these tests, similar to how it's done in the ThunkFunctionalTests.

This behavior is now only enabled for the thunk test, and only if building thunks are enabled *and* if glibc allocator hooking is disabled.

Sonicadvance1 (approved):

PR #3073, opened 2023-09-09
IR: Changes Select operation to explicit sized operations

Sonicadvance1:
Needs #3071 and #3072 merged first.

This instruction behaviour was very opaque because the size of the
destination was implicitly calculated while the size of the compares
where given explicitly.

Changes both to be explicit and all of the code the requires working
around it.

Quite a large number of changes since select is used all over the place.
Resulting asm output changes is fairly minimal.


Sonicadvance1:
#3099 is starting this instead.

PR #3077, opened 2023-09-11
FEXCore: Implements support for shifted bitwise ops

Sonicadvance1:
This wasn't implemented initially for the interpreter and x86 JIT.

This meant we are maintaining two codepaths. Implement these operations in the interpreter and x86 JIT so we no longer need to do that.

The emitted code in the x86 JIT is hot garbage, but it's only necessary for correctness testing, not performance testing there.

### DIFF
--- a/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
@@ -82,6 +82,91 @@ DEF_OP(Add) {
   }
 }
 
+DEF_OP(AddNZCV) {
+  auto Op = IROp->C<IR::IROp_AddNZCV>();
+  const uint8_t OpSize = Op->Size;
+
+  const uint64_t Src1 = *GetSrc<uint64_t*>(Data->SSAData, Op->Src1);
+  const uint64_t Src2 = *GetSrc<uint64_t*>(Data->SSAData, Op->Src2);
+
+  // Results returned in Arm64 NZCV format
+  // N = Sign bit
+  // Z = Is Zero
+  // C = Carry occured (Unsigned result can't fit within resulting register)
+  // V = Overflow occured (Signed result can't fit in to resulting register)
+  uint32_t NZCV{};
+  switch (OpSize) {
+    case 4: {
+      uint32_t Result = Src1 + Src2;
+      int32_t ResultSigned{};
+      if ((Result >> (sizeof(uint32_t) * 8 - 1)) & 1) {
+        NZCV |= 1U << 31;
+      }
+      if (Result == 0) {
+        NZCV |= 1U << 30;
+      }
+      if (__builtin_uadd_overflow(Src1, Src2, &Result)) {
+        NZCV |= 1U << 29;
+      }
+      if (__builtin_sadd_overflow(Src1, Src2, &ResultSigned)) {
+        NZCV |= 1U << 28;
+      }
+      break;
+    }
+    case 8: {
+      uint64_t Result = Src1 + Src2;
+      int64_t ResultSigned{};
+      if ((Result >> (sizeof(uint64_t) * 8 - 1)) & 1) {
+        NZCV |= 1U << 31;
+      }
+      if (Result == 0) {
+        NZCV |= 1U << 30;
+      }
+      if (__builtin_uaddl_overflow(Src1, Src2, &Result)) {
+        NZCV |= 1U << 29;
+      }
+      if (__builtin_saddl_overflow(Src1, Src2, &ResultSigned)) {
+        NZCV |= 1U << 28;
+      }
+      break;
+    }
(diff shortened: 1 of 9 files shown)


FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -97,6 +182,59 @@ DEF_OP(Sub) {
...
+      }
+      break;
+    }
+    case 8: {
+      uint64_t Result = Src1 - Src2;
+      int64_t ResultSigned{};
+      if ((Result >> (sizeof(uint64_t) * 8 - 1)) & 1) {
+        NZCV |= 1U << 31;
+      }
+      if (static_cast<uint64_t>(Result) == 0) {
```
lioncash, line 222: Ditto

FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -82,6 +82,91 @@ DEF_OP(Add) {
...
+  // V = Overflow occured (Signed result can't fit in to resulting register)
+  uint32_t NZCV{};
+  switch (OpSize) {
+    case 4: {
+      uint32_t Result = Src1 + Src2;
+      int32_t ResultSigned{};
+      if ((Result >> (sizeof(uint32_t) * 8 - 1)) & 1) {
+        NZCV |= 1U << 31;
+      }
+      if (static_cast<uint32_t>(Result) == 0) {
```
lioncash, line 105: This cast isn't necessary

FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -82,6 +82,91 @@ DEF_OP(Add) {
...
+      }
+      if (static_cast<uint32_t>(Src) == 0) {
+        NZCV |= 1U << 30;
+      }
+      break;
+    case 8:
+      if ((Src >> (sizeof(uint64_t) * 8 - 1)) & 1) {
+        NZCV |= 1U << 31;
+      }
+      if (static_cast<uint64_t>(Src) == 0) {
```
lioncash, line 161: Ditto

FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -97,6 +182,59 @@ DEF_OP(Sub) {
   }
 }
 
+DEF_OP(SubNZCV) {
+  auto Op = IROp->C<IR::IROp_SubNZCV>();
+  const uint8_t OpSize = Op->Size;
+
+  const uint64_t Src1 = *GetSrc<int64_t*>(Data->SSAData, Op->Src1);
```
lioncash, line 189: Ditto

FEXCore/Source/Interface/Core/Interpreter/ALUOps.cpp
```diff
@@ -82,6 +82,91 @@ DEF_OP(Add) {
...
+    default: LOGMAN_MSG_A_FMT("Unknown {} Size: {}\n", __func__, OpSize); break;
+  }
+  GD = NZCV;
+}
+
+DEF_OP(TestNZ) {
+  auto Op = IROp->C<IR::IROp_TestNZ>();
+  const uint8_t OpSize = Op->Size;
+
+  const uint64_t Src = *GetSrc<int64_t*>(Data->SSAData, Op->Src1);
```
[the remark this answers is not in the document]
Sonicadvance1: Unintentional. Resolved.

alyssarosenzweig:
I dont love that we're adding NZCV code paths to non arm64 JITs... it doesn't save any complexity because we need the other code path in the dispatcher for 8bit without flagm anyway... otoh i dont have any reason to NAK things either so shrug

Sonicadvance1:
> I dont love that we're adding NZCV code paths to non arm64 JITs... it doesn't save any complexity because we need the other code path in the dispatcher for 8bit without flagm anyway... otoh i dont have any reason to NAK things either so shrug

Indeed, the main thing here is that in most cases we shouldn't need to think about if one backend versus another supports a thing. Trying to keep x86 at feature parity of the arm64 JIT as possible. Emulating the architecture details of NZCV is relatively trivial even if the codegen is a bit ugly.

Once flagm/flagm2 stuff gets added then I'll glue some handlers in to the x86 JIT as well. At some point we'll kill off this JIT, just likely will be once I get my hands on an NVIDIA Grace or AmpereOne system.

alyssarosenzweig:
> Indeed, the main thing here is that in most cases we shouldn't need to think about if one backend versus another supports a thing.

I guess as long as I can continue to add instructions behind a flag without worrying about the x86 side I don't have any real objection to this / you adding x86 paths later

alyssarosenzweig (approved):

PR #3085, opened 2023-09-13
OpcodeDispatcher: Optimize a bunch of shufps variants 

Sonicadvance1:
Hits a whole bunch of common cases, most of which then emit optimal code
generation.
Two cases that use VInsElement hit the RA quirk where the SRA
destination is dead but RA doesn't see it, so it ends up doing a couple
moves. If RA gets fixed then those two moves will go away.

There are definitely still cases that we could emit more optimal code.
Additionally we could implement a TBL2 IR operation to do a LUT approach
for ones we don't cover.

Problem with implementing a TBL2 ir operation is that we have no way to
ensure registers are sequential so we would need to always do moves
```asm
ldr v2, <LUT Table>
mov v0, v16
mov v1, v18
tbl v16.16b, { v0.16b, v1.16b }, v2.16b
```

Which to be fair isn't terrible, and if we're lucky that the guest uses
sequential registers we can naturally get the more optimal code path.
Ideally our RA could push some operations in to sequential registers but
that's not possible currently.

I'll do a follow-up PR that implements TBL2.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -1262,6 +1262,135 @@ OrderedNode* OpDispatchBuilder::SHUFOpImpl(OpcodeArgs, size_t ElementSize,
       Shuffle >>= ShiftAmount;
     }
   } else {
+    if (ElementSize == 4) {
+      // We can shuffle optimally in a lot of cases.
+      // TODO: We can optimize more of these cases.
+      switch (Shuffle) {
+        case 0b01'00'01'00:
+          // Combining of low 64-bits.
+          // Dest[63:0]   = Src1[63:0]
+          // Dest[127:64] = Src2[63:0]
+          return _VZip(DstSize, 8, Src1Node, Src2Node);
+        case 0b11'10'11'10:
+          // Combining of high 64-bits.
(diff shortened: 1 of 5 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -1262,6 +1262,135 @@ OrderedNode* OpDispatchBuilder::SHUFOpImpl(OpcodeArgs, size_t ElementSize,
...
+        case 0b11'11'10'10:
+        case 0b11'11'11'11:
+        {
+          // Duplicate element in upper and lower across each 64-bit segment.
+          auto DupSrc1 = _VDupElement(DstSize, ElementSize, Src1Node, Shuffle & 0b11);
+          auto DupSrc2 = _VDupElement(DstSize, ElementSize, Src2Node, (Shuffle >> 4) & 0b11);
+          return _VZip(DstSize, 8, DupSrc1, DupSrc2);
+        }
+        default:
+          // Intentional fallthrough
```
lioncash, line 1387: Maybe comment this instead as intentionally doing nothing, since there's no fallthrough from the upper cases.

Sonicadvance1: Updated comment.

lioncash (approved):

PR #3100, opened 2023-09-15
OpcodeDispatcher: Optimize cmov 

Sonicadvance1:
cmov was quite terrible in its implementation. Some things of note:
- NZCV cache would cause store for no reason
- {16,32}-bit would zero extend sources for no reason
- 16-bit would zero extend result for no reason

A bunch of flag testing is still doing a ubfx plus compare against zero
when it could end up being a tst instead, but this is a step in the
right direction and switches over to explicit sized selects.

None of the cmovs became optimal yet sadly. That'll be a follow up PR.
~~Needs #3099 merged first.~~


alyssarosenzweig:
> A bunch of flag testing is still doing a ubfx plus compare against zero
when it could end up being a tst instead,

These should actually be csel/cset with the host NZCV flags but this requires RA changes (planned but waiting for the rewrite which is the main reason I've ignored cmov's garbage codegen thus far)

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -869,8 +869,9 @@ friend class FEXCore::IR::PassManager;
     }
   }
 
-  OrderedNode* CachedNZCV = {};
-  uint32_t PossiblySetNZCVBits = 0;
+  OrderedNode* CachedWriteNZCV{};
+  OrderedNode* CachedReadNZCV{};
```
alyssarosenzweig, line 873: I don't know if this is the right approach. The existing code implements a standard pattern *local value numbering* ... would it be practical to simply augment with a bool `NZCVDirty` instead of duplicating the node?

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1275,7 +1275,9 @@ friend class FEXCore::IR::PassManager;
     CachedIndexedNamedVectorConstants.clear();
   }
 
+  // TODO: SelectCC is duplicated SelectCCExplicitSize. Should get removed once all users are removed.
   OrderedNode *SelectCC(uint8_t OP, OrderedNode *TrueValue, OrderedNode *FalseValue);
```
alyssarosenzweig, line 1279: Please redefine this in terms of the explicit size version in this commit to avoid the duplication?

unittests/InstructionCountCI/Secondary.json
```diff
@@ -265,462 +265,388 @@
...
-        "ubfx w23, w22, #28, #1",
-        "cmp x23, #0x0 (0)",
-        "csel w20, w20, w21, ne",
-        "bfxil x4, x20, #0, #16",
-        "str w22, [x28, #728]"
+        "ldr w20, [x28, #728]",
+        "ubfx w20, w20, #28, #1",
+        "cmp w20, #0x0 (0)",
+        "csel w20, w7, w4, ne",
+        "bfxil x4, x20, #0, #16"
```
alyssarosenzweig, line 276: Is this bfxil possible to deal with somehow?

Sonicadvance1: Sadly this bfxil can't be removed since 16-bit operations insert in to the GPR instead of zero extending.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: The duplication is a necessary evil while we take steps to remove the usage of implicit sized SelectCC.
Once all implicit uses are removed then the function will get deleted.
Trying to take the conversion slower than the previous PR so it doesn't turn in to a broken mess in one commit.
Just a duplication mess while stuff is changing.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: I can switch over to a dirty flag instead

unittests/InstructionCountCI/Secondary.json
alyssarosenzweig: Right, ok... that's indeed unfortunate but nothing to be done I suppose

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
alyssarosenzweig: fair enough

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
alyssarosenzweig: Thanks, I suspect that will make more sense, though I'm not super confident tbh

Sonicadvance1: Rewrote to use a dirty flag to track the writes instead.

alyssarosenzweig: Thanks, the new commit makes a lot more sense to me! ... CI doesn't seem to agree yet but once it does this will get my approval :+1:

alyssarosenzweig (approved):

PR #3102, opened 2023-09-15
Inline constant with PF calculation

alyssarosenzweig:
possibly the constant inlining needs a special case to swap the operands of commutative operations?


Sonicadvance1 (approved):

Sonicadvance1:
> possibly the constant inlining needs a special case to swap the operands of commutative operations?

Alternative take would be to introduce an `AndI` IR op that only takes ARM capable immediate encodings and enforce immediates being the right way around to match ARM behaviour

alyssarosenzweig:
> > possibly the constant inlining needs a special case to swap the operands of commutative operations?
> 
> Alternative take would be to introduce an `AndI` IR op that only takes ARM capable immediate encodings and enforce immediates being the right way around to match ARM behaviour

I'm more thinking about the possibility of optimizing guest code that has explicit movs for ARM immediates, but maybe that doesn't occur in practice?

Sonicadvance1:
> > > possibly the constant inlining needs a special case to swap the operands of commutative operations?
> > 
> > 
> > Alternative take would be to introduce an `AndI` IR op that only takes ARM capable immediate encodings and enforce immediates being the right way around to match ARM behaviour
> 
> I'm more thinking about the possibility of optimizing guest code that has explicit movs for ARM immediates, but maybe that doesn't occur in practice?

x86 instructions are a mishmash of being able to encode immediates of any sizes in their second source operand. imm8, imm16, imm32, and like...two imm64?
Luckily it's always in the second source imm similar to ARM. Any route taken is fine though, I don't hold a strong opinion in any regard here.

PR #3105, opened 2023-09-15
OpcodeDispatcher: Handle RORX corner cases better

lioncash:
There are a few cases where we were emitting code when we didn't really need to, or could emit less.

### DIFF
--- a/unittests/InstructionCountCI/VEX_map3.json
+++ b/unittests/InstructionCountCI/VEX_map3.json
@@ -5415,14 +5415,23 @@
       ]
     },
     "rorx eax, ebx, 0": {
-      "ExpectedInstructionCount": 2,
-      "Optimal": "No",
+      "ExpectedInstructionCount": 1,
+      "Optimal": "Yes",
       "Comment": [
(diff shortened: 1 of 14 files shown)


unittests/InstructionCountCI/VEX_map3.json
```diff
@@ -5145,14 +5145,23 @@
...
-      "Optimal": "No",
+      "ExpectedInstructionCount": 1,
+      "Optimal": "Yes",
       "Comment": [
         "Map 3 0b11 0xf0 32-bit"
       ],
       "ExpectedArm64ASM": [
-        "lsr w20, w7, #0",
-        "ror w4, w20, #0"
+        "lsr w4, w7, #0"
```
Sonicadvance1, line 5154: I don't think this is technically optimal since `lsr` of zero doesn't hit CPU renaming mechanisms.
Maybe we should detect these and convert to a move?

lioncash: Opted for adding the handling into the Bfe IR op

Sonicadvance1 (approved):
Nice nice nice.

PR #3107, opened 2023-09-15
Arm64/ALUOps: Remove spills in PEXT

lioncash:
Reduces the number of emitted instructions for a corresponding PEXT instruction.

Notably, we no longer spill for this IR op.


Sonicadvance1 (approved):
Very cool! I didn't quite fully grasp the algorithm details but I'll trust it's correct.

PR #3108, opened 2023-09-16
OpcodeDispatcher: Improve output of MULX

lioncash:
We can cut down on a few of the generated moves. For the case where both destinations alias one another, we can just calculate the high part instead of the low one as well


Sonicadvance1 (approved):
Cute. I think in the future we can use umull for those 32-bit operations.

PR #3111, opened 2023-09-17
Arm64/VectorOps: Fix SVE aliasing-path  move in VSShr

lioncash:
Seems like this was a type from 8d11073, since we'd be moving into a temporary and then never use it.

We currently don't have usages of this shift IR op yet, so better to catch this ahead of time than not at all


Sonicadvance1 (approved):
oop, you're right

PR #3116, opened 2023-09-18
Minor/flag opts

alyssarosenzweig:


alyssarosenzweig:
There is a lot of room for improvement here but most of it requires doing more work than I'm up for this morning. that said, this makes the improvement potential a lot more obvious when reading instcountci

setp should do an ldrb into the FPR directly, and maybe do a VFP compare or at least do ands to avoid the cmp

lahf can use orn+ornlshl to replace and+eor+orlshl, details left to reader.

Sonicadvance1 (approved):

PR #3120, opened 2023-09-19
FEXCore: Support preserve_all ABI for interpreter fallbacks

Sonicadvance1:
This significantly improves performance of x87 instructions when operating in full precision mode. Resulting performance improvement can be seen from microbenchmark results.
Image_2023-09-18_17-20-06

Additionally outside of microbenchmark, this improves Linux native Psychonaut's title screen from 34 FPS to 51FPS. A perfect 50% additional performance.
This doesn't improve performance to the levels of the fp64 precision x87 mode but it is a decent uplift for the default option.

Additionally this `preserve_all` ABI path is only available on clang-17 which isn't out yet. So no one will be able to see the performance improvements yet.


FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -144,6 +236,101 @@ namespace x32 {
     FEXCore::ARMEmitter::VReg::v28, FEXCore::ARMEmitter::VReg::v29,
     FEXCore::ARMEmitter::VReg::v30, FEXCore::ARMEmitter::VReg::v31
   };
+
```
alyssarosenzweig, line 239: Can we deduplicate this chunk of code? failing to deduplicate stuff in FEX has bit me several times. I don't see why this instance should be different

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -532,111 +719,241 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
...
+    // rsp capable move
+    add(ARMEmitter::Size::i64Bit, TmpReg, ARMEmitter::Reg::rsp, 0);
+
+    // Push the vector registers
+    PushVectorRegisters(TmpReg, CanUseSVE, DynamicFPRs);
+
+    // Push the general registers.
+    PushGeneralRegisters(TmpReg, DynamicGPRs);
+  }
+  else {
```
alyssarosenzweig, line 890: surely we can/should dedupe the two sides of the if?

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -532,111 +719,241 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
...
+      DynamicFPRs = x32::PreserveAll_DynamicFPRSVE;
+    }
+
+    // Fill static registers second.
+    FillStaticRegs(true, x32::PreserveAll_SRAMask, CanUseSVE ? x32::PreserveAll_SRAFPRSVEMask : x32::PreserveAll_SRAFPRMask);
+
+    // Pop vectors first
+    PopVectorRegisters(CanUseSVE, DynamicFPRs);
+
+    // Pop GPRs second
```
alyssarosenzweig, line 952: ditto. also, the comments are out of sync with the code

FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -84,9 +84,14 @@ void Arm64JITCore::Op_Unhandled(IR::IROp_Header const *IROp, IR::NodeID Node) {
   } else {
     switch(Info.ABI) {
       case FABI_F80_I16_F32:{
-        SpillStaticRegs(TMP1);
+        if (Info.SupportsPreserveAllABI) {
```
alyssarosenzweig, line 87: probably want to factor out a helper for this

FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
```diff
@@ -97,9 +102,13 @@ void Arm64JITCore::Op_Unhandled(IR::IROp_Header const *IROp, IR::NodeID Node) {
         blr(ARMEmitter::Reg::r1);
 #endif
 
-        PopDynamicRegsAndLR();
-
-        FillStaticRegs();
+        if (Info.SupportsPreserveAllABI) {
```
alyssarosenzweig, line 105: and this

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/F80Fallbacks.h
```diff
@@ -141,14 +155,33 @@ struct OpHandlers<IR::OP_F80CVTTOINT> {
 
 template<>
 struct OpHandlers<IR::OP_F80ROUND> {
+  FEXCORE_PRESERVE_ALL_ATTR
   static X80SoftFloat handle(uint16_t NewFCW, X80SoftFloat Src1) {
-    LoadDeferredFCW(NewFCW);
```
alyssarosenzweig, line 145: unrelated change?

alyssarosenzweig (changes requested):

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/F80Fallbacks.h
Sonicadvance1: Removed

FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
Sonicadvance1: Added helper

FEXCore/Source/Interface/Core/JIT/Arm64/JIT.cpp
Sonicadvance1: Added helper

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Deduplicated

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Deduplicated

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: I have a comment there lamenting that I wish we could constexpr to deduplicate but we lack the library/compiler support for now. We could potentially switch over to parsing a uint32_t mask but that would require refactoring/duplicating a bunch of the push/pop routines to support one or the other.

neobrain:
Nice results!

Maybe I missed some bit of documentation in the diff that explains this, but what does this actually do and how does it help improve x87 performance?

alyssarosenzweig:
> Nice results!
> 
> Maybe I missed some bit of documentation in the diff that explains this, but what does this actually do and how does it help improve x87 performance?

Right now, FEX spills/fills EVERYTHING whenever it calls out to a fallback, due to the ABI mismatch between C code and FEX's SRA. This means x87 code is slooooooooow since every nontrivial x87 instruction is a fallback. This PR changes the calling convention to avoid needing to spill so many registers across call sites, instead requiring the C code to spill just the few registers it actually needs. ==> many fewer instructions executed overall

alyssarosenzweig (approved):

PR #3128, opened 2023-09-20
FEXCore: Gut interpreter

alyssarosenzweig:
It is scarcely used today, and like the x86 jit, it is a significant maintainence burden complicating work on FEXCore and arm64 optimization. Remove it, bringing us down to 2 backends.

1 down, 1 to go.

Some interpreter scaffolding remains for x87 fallbacks. That is not a problem here.


neobrain:
A few obsolete comments that can now be removed:
```
~/FEX $ ack -i "too slow"
unittests/gcc-target-tests-64/CMakeLists.txt
12:  # Interpreter is too slow to run these tests, only generate for jit

unittests/gvisor-tests/CMakeLists.txt
12:  # Interpreter is too slow to run these tests, only generate for jit

unittests/gcc-target-tests-32/CMakeLists.txt
12:  # Interpreter is too slow to run these tests, only generate for jit
```

neobrain:
Also we may be able to re-enable some tests that didn't reliably pass with the interpreter:
```
~/FEX $ ack "flakey on the interpreter"
POSIX/Known_Failures
391:# This test is flakey on the interpreter

POSIX/Disabled_Tests
379:# This test is flakey on the interpreter
```

alyssarosenzweig:
> terminate called after throwing an instance of 'std::bad_function_call'

I think I'm going to need help with this but honestly the value of a PR like this is generating discussion and nerdsnipes, the actual git diff is secondary...

Hopefully we're all on the same page on this one and can make it happen at any rate.

lioncash:
fwiw, personally I'm fine with getting rid of the interpreter, since I usually find the x86 host test runner and comparing it with arm-side tests more valuable for determining whether or not behavior is properly handled

Sonicadvance1:
> > > > terminate called after throwing an instance of 'std::bad_function_call'
> > > 
> > > 
> > > I think I'm going to need help with this
> > 
> > 
> > In gdb you can set a breakpoint on C++ exceptions using `catch throw`, so you'll be able to get a proper backtrace. Does that help?
> > (fwiw, `bad_function_call` sounds like it's trying to call an uninitialized/reset `std::function`)
> 
> how do I run FEX unit tests under gdb?

Checking `<Build Folder>/Testing/Temporary/LastTest.log` usually shows the execution of the commands as they're running that you can copy and paste. Not sure which test you're getting this under.
ASM unit tests could be run something similar like `gdb --args ./Bin/TestHarnessRunner --no-silent -g -c irjit -n 500 ./unittests/ASM/Primary/Primary_8C.asm.bin ./unittests/ASM/Primary/Primary_8C.asm.config.bin`

When under gdb you also then likely need to do `handle SIGSEGV SIGBUS SIGILL SIG63 nostop`

alyssarosenzweig:
> > > > > terminate called after throwing an instance of 'std::bad_function_call'
> > > > 
> > > > 
> > > > I think I'm going to need help with this
> > > 
> > > 
> > > In gdb you can set a breakpoint on C++ exceptions using `catch throw`, so you'll be able to get a proper backtrace. Does that help?
> > > (fwiw, `bad_function_call` sounds like it's trying to call an uninitialized/reset `std::function`)
> > 
> > 
> > how do I run FEX unit tests under gdb?
> 
> Checking `<Build Folder>/Testing/Temporary/LastTest.log` usually shows the execution of the commands as they're running that you can copy and paste. Not sure which test you're getting this under. ASM unit tests could be run something similar like `gdb --args ./Bin/TestHarnessRunner --no-silent -g -c irjit -n 500 ./unittests/ASM/Primary/Primary_8C.asm.bin ./unittests/ASM/Primary/Primary_8C.asm.config.bin`
> 
> When under gdb you also then likely need to do `handle SIGSEGV SIGBUS SIGILL SIG63 nostop`

thanks, got it eventually... CONFIG enum needed some adjustments, let's see what CI thinks now that that's fixed :smile:

alyssarosenzweig:
CI seems happy with this now. @Sonicadvance1 I would prefer to merge this ahead of #3131 if that's ok :lollipop:

Sonicadvance1 (approved):

PR #3131, opened 2023-09-21
OpcodeDispatcher: Optimize lock btr 

Sonicadvance1:
This is an atomicFetchCLR, removes two mvn instructions that are back to
back negating the source.

We didn't have this instruction combination in InstCountCI so will be a
bit hard to see.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/x86_64/AtomicOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/x86_64/AtomicOps.cpp
@@ -399,6 +399,92 @@ DEF_OP(AtomicFetchAnd) {
   }
 }
 
+DEF_OP(AtomicFetchCLR) {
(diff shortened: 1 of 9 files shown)


FEXCore/Source/Interface/Core/JIT/x86_64/AtomicOps.cpp
```diff
@@ -399,6 +399,92 @@ DEF_OP(AtomicFetchAnd) {
   }
 }
 
+DEF_OP(AtomicFetchCLR) {
```
alyssarosenzweig, line 402: sigh

alyssarosenzweig (approved):

alyssarosenzweig:
lgtm but let's land #3128 first and drop the interpreter support here...

would be even better to drop the x86 support too, but,.

Sonicadvance1:
Rebased since #3128 landed.

PR #3134, opened 2023-09-22
FEXCore: Removes x86 JIT.

Sonicadvance1:
This is blocking performance improvements. This backend is almost unilaterally unused except for when I'm testing if games run on Radeon video drivers.

Hopefully AmpereOne and Orin/Grace can fulfill this role when they launch next year.


alyssarosenzweig (approved):
:fire: :purple_heart:

lioncash:
Just glancing over this and wondering how the x86 host test runner CI testing works now. Does the x64 vixl CI now run those now as opposed to the previous build + test x64 CI that we had?

Sonicadvance1:
> Just glancing over this and wondering how the x86 host test runner CI testing works now. Does the x64 vixl CI now run those now as opposed to the previous build + test x64 CI that we had?

The testharnessrunner Host runner runs with the "custom" CPU core still, which means it still executes the code natively.
Might need a new github workflow just so those can still run.

Sonicadvance1:
> Just glancing over this and wondering how the x86 host test runner CI testing works now. Does the x64 vixl CI now run those now as opposed to the previous build + test x64 CI that we had?

Added back the host testharnessrunner for the ASM tests. This way the x86 CI can still get abused appropriately to ensure the ASM implementations are correct.

We will need a new solution for thunks testing, but since this is blocking work that can come up as a follow-up.

lioncash (approved):

alyssarosenzweig:
> > Just glancing over this and wondering how the x86 host test runner CI testing works now. Does the x64 vixl CI now run those now as opposed to the previous build + test x64 CI that we had?
> 
> Added back the host testharnessrunner for the ASM tests. This way the x86 CI can still get abused appropriately to ensure the ASM implementations are correct.
> 
> We will need a new solution for thunks testing, but since this is blocking work that can come up as a follow-up.

Very goofy but looks good, thanks for taking care of this!

PR #3136, opened 2023-09-22
OpcodeDispatcher: Optimize selects of extracts

alyssarosenzweig:
Add AND(N)Z conditions allowing us to optimize selects of extracts (typically of NZCV). Taking full advantage of this will require us to RA the NZCV register, a perenial problem...


alyssarosenzweig:
Will need a rebase once #3134 lands but otherwise this should be good to go.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -843,77 +854,56 @@ OrderedNode *OpDispatchBuilder::SelectCC(uint8_t OP, IR::OpSize ResultSize, Orde
 
   switch (OP) {
     case 0x0: { // JO - Jump if OF == 1
-      auto Flag = GetRFLAG(FEXCore::X86State::RFLAG_OF_LOC);
-      SrcCond = _Select(ResultSize, OpSize::i32Bit, CondClassType{COND_NEQ},
-          Flag, ZeroConst, TrueValue, FalseValue);
+      SrcCond = SelectNZCV(FEXCore::X86State::RFLAG_OF_LOC, 1, ResultSize, TrueValue, FalseValue);
```
lioncash, line 857: A minor nit, but maybe prefer using the boolean literal equivalent to make it evident at the call site that the parameter is a boolean, rather than an integer type

```suggestion
      SrcCond = SelectNZCV(FEXCore::X86State::RFLAG_OF_LOC, true, ResultSize, TrueValue, FalseValue);
```

ditto for the others.

alyssarosenzweig: done

alyssarosenzweig:
-> #3138

PR #3138, opened 2023-09-22
Requiem for the x86 jit

alyssarosenzweig:
In honour of the death of the x86 JIT, have moar fps!!!!

Merge train of today's work.


Sonicadvance1 (approved):
Woop woop, it's the sound of the speed police.

PR #3139, opened 2023-09-23
FEXServerClient: Adds back ServerSocketPath config option 

Sonicadvance1:
This option was disabled a few months ago when we switched the server
socket from a filesystem unix socket to an abstract socket.
This partially broke our chroot scripts which relied on this option
existing.

Readds support for an explicitly named abstract socket named from
config.

This is a workaround for dealing with chroots that change users.
They end up changing a user while doing operations and then can't
connect to the FEXServer anymore because environment variables have been
wiped away.

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -3,6 +3,7 @@
 #include "Common/FEXServerClient.h"
 
 #include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/Utils/FileLoading.h>
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/Utils/NetStream.h>
 #include <FEXCore/fextl/fmt.h>
@@ -143,7 +144,11 @@ namespace FEXServerClient {
   }
 
   fextl::string GetServerSocketName() {
-    return fextl::fmt::format("{}.FEXServer.Socket", ::geteuid());
+    FEX_CONFIG_OPT(ServerSocketPath, SERVERSOCKETPATH);
+    if (ServerSocketPath().empty()) {
+      return fextl::fmt::format("{}.FEXServer.Socket", ::geteuid());
+    }
+    return ServerSocketPath;
   }
 
   int GetServerFD() {


Source/Common/FEXServerClient.cpp
```diff
@@ -143,7 +144,21 @@ namespace FEXServerClient {
   }
 
   fextl::string GetServerSocketName() {
-    return fextl::fmt::format("{}.FEXServer.Socket", ::geteuid());
+    constexpr std::string_view HARDCODED_SOCKET_PATH = "/FEX_Abstract_Socket";
+    if (FHU::Filesystem::Exists(HARDCODED_SOCKET_PATH.data())) {
+      // If this hardcoded filepath exists then use it.
+      // This is a required workaround for chroots changing users.
```
neobrain, line 150: What is the problem, and what is the actual workaround?

Sonicadvance1: Updated the comment.

neobrain: The PR description mentions something about environment variables. How are environment variables related to this? It's probably obvious, but not from the context here since the code below only deals with FEX's config.

Sonicadvance1: The rootfs scripts set the environment variable `FEX_SERVERSOCKETPATH` which typically gets inherited by the child processes: https://github.com/FEX-Emu/RootFS/blob/main/unbreak_chroot.sh#L120

Once an application in the chroot does a sudo/su, that environment variable stops getting inherited by default. Which means we can't rely on environment variables existing to inherit where FEXServer's abstract socket lives.

neobrain: For su/sudo specifically, could we configure the sudoers file to preserve environment variables instead?

If not, do we need the `FEX_SERVERSOCKETPATH` code at all anymore if this is what the rootfs scripts will now be using?

Sonicadvance1: We still need the environment variable for the FEXServer that is starting up.
I used su/sudo as an example, the apt/gnupg or whatever can drop the variable in a multitude of ways.

neobrain: > I used su/sudo as an example, the apt/gnupg or whatever can drop the variable in a multitude of ways.

Does this happen even for variables listed in `/etc/environment`? It's unclear what actual mechanism is triggering this, and if apt/gnupg also run into it, does it have anything to do with user switching at all?

Sonicadvance1: Looks like `/etc/environment` works around the issue today. Removed the hardcoded path, just made sure the environment variable works.

Source/Common/FEXServerClient.cpp
```diff
@@ -3,6 +3,7 @@
 #include "Common/FEXServerClient.h"
 
 #include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/Utils/FileLoading.h>
```
neobrain, line 6: Looks like this header isn't used anymore now.

neobrain (approved):
Commit message / PR writeup need updates now that no new functionality is added anymore. Looks good otherwise!

PR #3147, opened 2023-09-24
More opts to the dispatcher + 1 to the JIT

alyssarosenzweig:

PR #3148, opened 2023-09-24
InstCountCI: Adds negative immediate primary tests

Sonicadvance1:
Noticed these were missing

Will effect #3147


alyssarosenzweig (approved):
ack, didn't review

PR #3149, opened 2023-09-24
InstCountCI: Fail CI if there was any difference.

Sonicadvance1:
Now that InstCountCI has proven itself a workable solution, change the CI solution so that CI fails if there is any difference.
- If instruction count has changed then it can fail directly  in the run step
- CI then runs the `instcountci_update_tests` target which updates the git repo with the new data
- If there is any changes in the git repo then fail CI. This means there was a difference.
- diff file is uploaded as an artifact to quickly see why something changed and the dev can quickly create a patch with it.

This fixes the issue that semi-conflicting instruction changes could previously merge and produce unexpected results and we don't detect it. Plus improves workflow for people that don't directly work with InstCountCI every day.


alyssarosenzweig (approved):

Sonicadvance1:
This is now ready. Tested by uploading a bad commit that changed InstCountCI and CI failed immediately with an artifact showing the change.

PR #3150, opened 2023-09-25
Optimize PF calculation in lahf

alyssarosenzweig:
Do some nonsense to shave an instruction off. I'm feeling silly :laughing:


FEXCore/Source/Interface/Core/OpcodeDispatcher/Flags.cpp
```diff
@@ -162,27 +160,35 @@ OrderedNode *OpDispatchBuilder::GetPackedRFLAG(uint32_t FlagsMask) {
...
+    Original = _Orlshl(OpSize::i64Bit, Original, Flag, FlagOffset);
+  }
 
-    Nonzero = true;
+  // Raw PF value needs to have its bottom bit masked out and inverted. The
+  // naive sequence is and/eor/orlshl. But we can do the inversion implicitly
+  // instead.
+  if (FlagsMask & (1 << FEXCore::X86State::RFLAG_PF_LOC)) {
+    // Set every bit except the bottommost.
+    auto OnesInvPF = _Or(OpSize::i64Bit, LoadPFRaw(), _Constant(~(uint64_t)1));
```
Sonicadvance1, line 185: You can use integer literal suffixes for this to make this nicer. `ULL` suffix is what we landed on in the codebase for uin64_t.
eg: `~1ULL`

alyssarosenzweig: wasn't sure about the portability

Sonicadvance1: We require C++20 everywhere and this is in the spec since...C++11 I think? Used all over the place so this wouldn't be the first :)

alyssarosenzweig: changed regardless

Sonicadvance1 (approved):
Definitely some silliness happening here. Nice to see the instruction reduction from it though.

PR #3153, opened 2023-09-25
Use adcs

alyssarosenzweig:
We could also benefit from adc/sbc but probably need to RA the flags to benefit much.


FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -103,6 +103,54 @@ DEF_OP(AddNZCV) {
...
+
+  LOGMAN_THROW_AA_FMT(OpSize == IR::i32Bit || OpSize == IR::i64Bit, "Unsupported {} size: {}", __func__, OpSize);
+  const auto EmitSize = OpSize == IR::i64Bit ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Dst = GetReg(Node);
+
+  // Carry-in needs to be inverted for subtractions due to carry versus borrow
+  // distinction between x86 and arm.
+  // See below remarks on cfinv
+  eor(ARMEmitter::Size::i32Bit, Dst, GetReg(Op->NZCV.ID()), 1u << 29);
```
Sonicadvance1, line 135: This line contains your bug. You're overwriting Dst when the register allocator has chosen Dst == Src1.
Which causes this eor to overwrite the first source register before the sbcs can occur.
Have this eor and the next msr use TMP1 and the bug will be resolved.
I'm opening a PR with this unit test as well.

alyssarosenzweig: oops, fixed, thanks!

Sonicadvance1 (approved):

Sonicadvance1:
Tested some games. works great

PR #3154, opened 2023-09-25
Optimize 8/16-bit CF calculation

alyssarosenzweig:


alyssarosenzweig:
Question: would it be even better to keep masking the result but instead drop the 2 uxt's of the source? might be opcode dependent....

Sonicadvance1:
> Question: would it be even better to keep masking the result but instead drop the 2 uxt's of the source? This seems opcode dependent.... given the wins here, I think this is the better choice, though we can potentially drop source masking in more cases too

This is mostly operation dependent if I'm remembering correctly. Most ALU operations don't need source or destination masking at < 32-bits. Divides being an exception off the top of my head.

Then result masking is mostly unnecessary since in most cases it will insert in to a GPR so will always be followed with a bfxil, or it is getting stored to memory which will always work on the smaller size and not need a mask.

So the problem ends up being in the weird in-between optimizing state when we have i8 and i16 values getting optimized with potentially garbage data in the upper bits. This is why we are being overly careful with the masking to ensure we don't end up optimizing and generating garbage data.

As an example, clang masks the result because it's one mask versus two and then doesn't need to deal with overflow problems.

As long as behaviour is correct I don't might which route we take.

alyssarosenzweig:
I think we can insert masking on sources retroactively for add/sub in the future. For now this PR should make things strictly better so should probs be merged.

Sonicadvance1 (approved):
Already gave this a bit of real application testing earlier today and didn't notice any regressions. Nice to see the bit of code removal.

PR #3135, opened 2023-09-22
[Not for review] Thunks: Implement data layout analysis and struct repacking. Add 32-bit support to Wayland and Vulkan

neobrain:
I'll split these changes up into smaller PRs for easier review. Having the full set of changes visible may be useful to judge the design during review though.

Full list of changes:
* Data layout analysis pass
* Data layout *comparison* module
* Automatic struct repacking for structs without pointers
* Assisted, semi-automatic struct repacking ("custom repacking")
* New annotations: `ptr_passthrough`/`is_opaque`/`assume_compatible_data_layout`/`custom_repack`
* 32-bit support for Wayland and Vulkan, sufficient to run vkcube / Super Meat Boy / zink
* Tests... _All_ the tests

Related PRs:
* #3067
* #3068
* #3126
* #3156
* #3177
* #3224
* #3225
* #3226
* #3241
* #3309
* #3318
* #3371
* #3373
* #3374
* #3377
* #3391 
* #3392
* #3406
* #3407
* #3422
* #3486
* #3494
* #3487

PR #3156, opened 2023-09-25
Thunks: Analyze data layout to detect platform differences

neobrain:
## Overview

FEX's thunking currently assumes that parameters passed across architecture boundaries use the same data layout on both sides. For example, a function of the form `void CreateObject(ConfigType* config)` can only be handled by FEX if `ConfigType` has the same members (with the same bit sizes) on both x86 and on ARM. This is generally a safe assumption when thunking 64-bit guest applications, but it fully falls apart for 32-bit guests (since pointer members of any structure will be 32-bits instead of 64-bit). However even on 64-bit, it's desirable to have some assurance that we're indeed operating on consistent data.

This PR adds a data layout analysis framework as the first step towards solving these problems. The framework is integrated as a preprocessing pass in the thunk generator. This new pass performs these tasks:
* it gathers all argument types used in any thunked functions
* it queries their data layout from libclang on *both* x86 and ARM to build an internal `type name -> data layout` map for each architecture
* finally, it compares the two maps to detect any potential differences

For simple libraries, the data layout analysis will detect a perfect match such that no additional work is needed. For more complex libraries, future PRs will add *annotations* to the interface files that allow to resolve data type differences semi-automatically via "struct repacking". Since most of our current thunk libraries fall in the latter category, analysis is only enabled for testing.

## Implementation details

Since data layout is queried from libclang, we must run a large chunk of logic twice to get accurate data for *both* the guest *and* the host architecture. The thunk generator will first parse the thunk interface file and query data layouts for the *guest architecture*, after which it repeats the same for the *host architecture*.

It's easy to see that this gets very intertwined quickly, especially if we want a testable design.

To facilitate separation of concerns, a class-hierarchy approach is used for the implementation hence. #3067 already initiated this by splitting interface parsing and boilerplate code emission into an `AnalysisAction` and a `GenerateThunkLibsAction`, respectively. The new functionality affects `AnalysisAction` and adds a number of new modules on top, illustrated in this diagram:

```mermaid
classDiagram
    AnalysisAction
    AnalysisAction <|-- AnalyzeDataLayoutAction
    AnalysisAction <|-- DataLayoutCompareAction
    DataLayoutCompareAction <|-- DataLayoutCompareActionForTest
    DataLayoutCompareAction <|-- GenerateThunkLibsAction
  
    class AnalysisAction {
        - Parses interface files
        - Analyzes data layout factoring in annotations
    }
    class AnalyzeDataLayoutAction {
        - Runs AnalysisAction for the guest

        - Turns any libclang objects into strings so they
        can be used across sessions
    }
    class DataLayoutCompareAction {
        - Runs AnalysisAction for the host
        - Takes the output of an AnalyzeDataLayoutAction
        for the guest arch and detects any differences
        between the guest/host results
    }
    class DataLayoutCompareActionForTest {
        - Thin layer around DataLayoutCompareAction
        used for unit-testing data layout comparison.
        - Mainly deals with lifetime properties
        of libclang objects
    }
    class GenerateThunkLibsAction {
        - Emits thunk boilerplate, factoring in results
        of the DataLayoutCompareAction
    }
```

## Impact

This feature is currently only enabled for unit tests. To enable it for real libraries, new annotations must be added and applied to the corresponding interface files. This has been done in the preview change set in #3135, which supports all of our existing thunk libraries.

These annotations are planned to be added on a per-type basis, so that we won't need to go through each function individually. That helps cut down the work involved tremendously.


unittests/ThunkLibs/common.h
```diff
@@ -42,24 +42,77 @@ class TestDiagnosticConsumer : public clang::TextDiagnosticPrinter {
...
+
+// function parameter annotations: fex_gen_config<fexgen::annotate_parameter<MyFunc, 2>>
+struct ptr_in {};
+struct ptr_out {};
+struct ptr_inout {};
+struct ptr_pointer_passthrough {};
+struct ptr_is_untyped_address {};
+
+template<auto, int, typename = decltype(nullptr)>
+struct annotate_parameter {};
```
neobrain, line 105: These aren't actually used yet, so they should be dropped.

ThunkLibs/include/common/GeneratorInterface.h
```diff
@@ -13,4 +13,24 @@ struct callback_annotation_base {
...
+// The type passed to Host is guest_layout<PointeeType>*.
+// TODO: Update description. "The raw guest_layout<type>* will be passed to the host"
+struct ptr_passthrough {};
+
+// Type / Function parameter annotation.
+// Assume objects of the given type are compatible across architectures,
+// even if the generator can't automatically prove this. For pointers, this refers to the pointee type.
+// NOTE: In contrast to opaque_type, this allows for non-pointer members with the annotated type to be repacked automatically.
+struct assume_compatible_data_layout : type_annotation_base {};
+
```
neobrain, line 35: These aren't used yet either.

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -52,6 +52,9 @@ function(generate NAME SOURCE_FILE)
   add_library(${NAME}-guest-deps INTERFACE)
   target_include_directories(${NAME}-guest-deps INTERFACE "${CMAKE_CURRENT_SOURCE_DIR}/../include")
   target_compile_definitions(${NAME}-guest-deps INTERFACE GUEST_THUNK_LIBRARY)
+  if (BITNESS EQUAL 32)
+    target_compile_definitions(${NAME}-guest-deps INTERFACE IS_32BIT_THUNK)
+  endif ()
```
neobrain, line 57: Changes related to 32-bit support should probably be moved to a future PR, too.

ThunkLibs/Generator/data_layout.h
```diff
@@ -0,0 +1,120 @@
...
+
+    TypeCompatibility GetTypeCompatibility(
+            const clang::ASTContext&,
+            const clang::Type*,
+            const std::unordered_map<const clang::Type*, TypeInfo> host_abi,
+            std::unordered_map<const clang::Type*, TypeCompatibility>& type_compat);
+
+protected: // TODO: Should probably be private
+    // TODO: Make it clearer that this is the guest ABI
+    const ABI& abi;
```
neobrain, line 119: These TODOs should be addressed before merging.

ThunkLibs/Generator/data_layout.h
```diff
@@ -0,0 +1,120 @@
+#pragma once
+
+#include "analysis.h" // TODO: Drop include. Currently needed for ParameterAnnotations though
```
neobrain, line 3: Should drop the TODO.

neobrain (commented):
Adding some notes for personal reference.

ThunkLibs/GuestLibs/CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ endif()
 
 if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_SOURCE_DIR)
   # We've been included using ExternalProject_add, so set up the actual thunk libraries to be cross-compiled
-  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD 20)
```
neobrain, line 20: This is probably not needed, yet.

ThunkLibs/GuestLibs/CMakeLists.txt
neobrain: Actually this is needed for `libfex_thunk_test`, which we're already testing the 32-bit build of. Not including this here would require adding some temporary workaround that would be removed one PR later anyway.

unittests/ThunkLibs/common.h
```diff
@@ -42,15 +42,38 @@ class TestDiagnosticConsumer : public clang::TextDiagnosticPrinter {
...
+    return os;
+}
+
 /**
  * Run the given ToolAction on the input code.
  *
  * The "silent" parameter is used to suppress non-fatal diagnostics in tests that expect failure
  */
-inline void run_tool(clang::tooling::ToolAction& action, std::string_view code, bool silent = false) {
+inline void run_tool(clang::tooling::ToolAction& action, std::string_view code, bool silent = false, std::optional<GuestABI> guest_abi = std::nullopt) {
```
neobrain, line 64: libclang will leak memory in `invocation.run()` when the given `ToolAction` throws an exception. This is annoying but otherwise harmless. We may consider defining something like this here though to disable memory leak checking for thunk unit tests:
```cpp
const char* __asan_default_options() { return "detect_leaks=0"; }
```

unittests/ThunkLibs/abi.cpp
```diff
@@ -0,0 +1,646 @@
...
+        auto action = compute_data_layout(
+            "#include <thunks_common.h>\n"
+            "#include <cstdlib>\n",
+            "struct A { size_t a; };\n"
+            "template<> struct fex_gen_type<A> {};\n", guest_abi);
+
+        INFO(FormatDataLayout(action->host_layout));
+
+        REQUIRE(action->guest_layout->contains("A"));
+        CHECK(action->GetTypeCompatibility("struct A") == (guest_abi == GuestABI::X86_32 ? TypeCompatibility::Repackable : TypeCompatibility::Full));
```
Sonicadvance1, line 220: Looks like you added `compat_full64_repackable32` for this. used below

unittests/ThunkLibs/abi.cpp
```diff
@@ -0,0 +1,646 @@
...
+            "#include <thunks_common.h>\n"
+            "#include <cstdint>\n",
+            "struct A { int64_t a; };\n"
+            "template<> struct fex_gen_type<A> {};\n", guest_abi);
+
+        INFO(FormatDataLayout(action->host_layout));
+
+        REQUIRE(action->guest_layout->contains("A"));
+        CHECK(action->guest_layout->at("A").get_if_struct()->alignment_bits == (guest_abi == GuestABI::X86_32 ? 32 : 64));
+        CHECK(action->GetTypeCompatibility("struct A") == (guest_abi == GuestABI::X86_32 ? TypeCompatibility::Repackable : TypeCompatibility::Full));
```
Sonicadvance1, line 234: Looks like you added `compat_full64_repackable32` for this. used below

unittests/ThunkLibs/abi.cpp
```diff
@@ -0,0 +1,646 @@
...
+            "#include <thunks_common.h>\n"
+            "#include <cstdint>\n",
+            "struct A { int32_t a; int32_t b; int64_t c; };\n"
+            "template<> struct fex_gen_type<A> {};\n", guest_abi);
+
+        INFO(FormatDataLayout(action->host_layout));
+
+        REQUIRE(action->guest_layout->contains("A"));
+        CHECK(action->guest_layout->at("A").get_if_struct()->alignment_bits == (guest_abi == GuestABI::X86_32 ? 32 : 64));
+        CHECK(action->GetTypeCompatibility("struct A") == (guest_abi == GuestABI::X86_32 ? TypeCompatibility::Repackable : TypeCompatibility::Full));
```
Sonicadvance1, line 276: Looks like you added `compat_full64_repackable32` for this. used elsewhere

unittests/ThunkLibs/abi.cpp
```diff
@@ -0,0 +1,646 @@
...
+            "template<> struct fex_gen_type<A> {};\n", guest_abi);
+
+        INFO(FormatDataLayout(action->host_layout));
+
+        REQUIRE(action->guest_layout->contains("A"));
+
+        CHECK(action->GetTypeCompatibility("struct A") == TypeCompatibility::None);
+    }
+
+if (false) // TODO: Currently fails
```
Sonicadvance1, line 375: Curly braces around for if or fix todo I guess.

Sonicadvance1 (commented):
I'm liking what I'm seeing. The unit tests are great to see!

neobrain:
Branch updated ([see changes](https://github.com/FEX-Emu/FEX/compare/899d6a52aea90ea5e27b5a0855ec65c73399fa66..36fedfdb87225b522ca313ff908e0f9423919ebe)):

* Minor cleanups and implemented all easy TODOs (others refer to open questions or will be addressed in follow-up PRs)
* Now disabling ASAN leak detection for tests
* Addressed review feedback
* Fixed and re-enabled tests previously marked as failing

Good to merge from my end 🙂

unittests/ThunkLibs/abi.cpp
neobrain: Checking if the guest layout misses any members from the host layout would be a bit annoying, but I found an elegant way to implement the tested behavior:

The analysis pass already verified that the host layout doesn't missed any members from the guest layout (i.e. the reverse case). To extend this to a "both-way match", we can simply add a check for *equal member count*!

Sonicadvance1 (approved):
Zip zap zoop, in to main it goes

PR #3157, opened 2023-09-25
OpcodeDispatcher: Don't mask logic op inputs

alyssarosenzweig:
Eliminate a pile of moves/uxt.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -4990,7 +4990,7 @@ void OpDispatchBuilder::UpdatePrefixFromSegment(OrderedNode *Segment, uint32_t S
   }
 }
 
-OrderedNode *OpDispatchBuilder::LoadSource_WithOpSize(FEXCore::IR::RegisterClassType Class, FEXCore::X86Tables::DecodedOp const& Op, FEXCore::X86Tables::DecodedOperand const& Operand, uint8_t OpSize, uint32_t Flags, int8_t Align, bool LoadData, bool ForceLoad, MemoryAccessType AccessType) {
+OrderedNode *OpDispatchBuilder::LoadSource_WithOpSize(FEXCore::IR::RegisterClassType Class, FEXCore::X86Tables::DecodedOp const& Op, FEXCore::X86Tables::DecodedOperand const& Operand, uint8_t OpSize, uint32_t Flags, int8_t Align, bool LoadData, bool ForceLoad, MemoryAccessType AccessType, bool AllowUpperGarbage) {
   LOGMAN_THROW_A_FMT(Operand.IsGPR() ||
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -4990,7 +4990,7 @@ void OpDispatchBuilder::UpdatePrefixFromSegment(OrderedNode *Segment, uint32_t S
   }
 }
 
-OrderedNode *OpDispatchBuilder::LoadSource_WithOpSize(FEXCore::IR::RegisterClassType Class, FEXCore::X86Tables::DecodedOp const& Op, FEXCore::X86Tables::DecodedOperand const& Operand, uint8_t OpSize, uint32_t Flags, int8_t Align, bool LoadData, bool ForceLoad, MemoryAccessType AccessType) {
+OrderedNode *OpDispatchBuilder::LoadSource_WithOpSize(FEXCore::IR::RegisterClassType Class, FEXCore::X86Tables::DecodedOp const& Op, FEXCore::X86Tables::DecodedOperand const& Operand, uint8_t OpSize, uint32_t Flags, int8_t Align, bool LoadData, bool ForceLoad, MemoryAccessType AccessType, bool AllowUpperGarbage) {
```
lioncash, line 4993: mental note to self to turn all these options into a struct type or something so we can just use designated initializers for things that need to differ from defaults, this is getting quite long

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5158,16 +5158,19 @@ OrderedNode *OpDispatchBuilder::GetRelocatedPC(FEXCore::X86Tables::DecodedOp con
...
     Size = GPRSize;
   }
   OrderedNode *Reg = _LoadRegister(false, offsetof(FEXCore::Core::CPUState, gregs[GPR]), GPRClass, GPRFixedClass, GPRSize);
 
-  if (Size != GPRSize || Offset != 0) {
+  if ((!AllowUpperGarbage && (Size != GPRSize)) || Offset != 0) {
     // Extract the subregister if requested.
-    Reg = _Bfe(IR::SizeToOpSize(std::max<uint8_t>(4u, Size)), Size * 8, Offset, Reg);
+    if (AllowUpperGarbage)
+      Reg = _Lshr(IR::SizeToOpSize(std::max<uint8_t>(4u, Size)), Reg, _Constant(Offset));
```
lioncash, line 5171: Maybe pull the

```cpp
IR::SizeToOpSize(std::max<uint8_t>(4u, Size))
```

bit into something like:

```cpp
const auto OpSize = IR::SizeToOpSize(std::max<uint8_t>(4u, Size));
```

So it doesn't need to be duplicated across both branches

Sonicadvance1 (approved):
Nice

PR #3163, opened 2023-09-27
Optimize ASCII flags

alyssarosenzweig:


Sonicadvance1 (approved):
Nightmare instructions plus random ptest improvement.

PR #3165, opened 2023-09-29
unittests/ASM: Implements tests for vpgatherdd/vgatherps

Sonicadvance1:
vpgatherdd and vgatherps are effectively the same instructions, so the tests are the same except for the instruction mnemonic.

This adds unit tests for two of the eight gather instructions. Specifically this adds tests for the 32-bit indices loading 32-bit elements instructions.

What it tests:
- Tests all displacement scales
- Tests multiple mask arrangements
- Ensures the mask register is zero'd after the instruction

What it doesn't test:
- Doesn't test address size calculation overflow
   - Only would happen on 32-bit with 32-bit indices, or /really/ high base addresses
   - The instruction should behave as a mask to the address size
   - Effectively behaves like `(uint64_t)(base + index << ilog2(scale))`
   - Better idea is to just not expose AVX to 32-bit applications
- Doesn't test VSIB immediate displacement
   - This just ends up being base_addr + imm so it isn't too interesting
   - We can add more tests in the future if we think we messed that up
- Doesn't test partial fault behaviour
   - Because that's a nightmare.

Specifically keeps each instruction test small and isolated so if a single register fails it is very easily to nail down which operation did it.
I know some of our ASM tests do a chunk of work and spit out a result at the end which can be difficult to debug in some cases. Didn't want to do that which is why the tests are spread out across 16 files for these single class of instructions.


lioncash (approved):

lioncash:
these should be disabled until the handling is added, so the vixl sim doesn't explode

Sonicadvance1:
> these should be disabled until the handling is added, so the vixl sim doesn't explode

Yea, I'm going to add a quick AVX2 feature flag to the asm tests and HostFeatures so it can be checked.

Sonicadvance1:
AVX2 flag check added. Which only these gather instruction tests add currently.
Technically we could move the other AVX2 instruction tests over but since we already support all those instructions it is nicer to keep those working. Maybe once we have full AVX2 we could switch the remainder over at that point.

lioncash:
> Maybe once we have full AVX2 we could switch the remainder over at that point.

Yeah, sounds totally fine to me

PR #3169, opened 2023-09-30
FEXCore: Support CpuState relative vector named constants 

Sonicadvance1:
The motivation towards just having a pointer array in CpuState was that
initialization was fairly cheap and that we have limited space inside
the encoding depending on what we want to do.

Initialization cost is still a concern but doing a memcpy of 128-bytes
isn't that big of a deal.

Limited space in CpuState, while a concern isn't a significant one.
   - Needs to currently be less than 1 page in size
   - Needs to be under the architectural offset limitations of loadstore
     scaled offsets. Which is 65KB for 128-bit vectors

Still keeps the pointer array around for cases when we would need
synthesize an address offset and it's just easier to load the
process-wide table.

The performance improvement here is removing the dependency in the
ldr+ldr chain. In microbenchmarks this has shown to have an improvement
of ~4% by removing this dependency chain on Cortex-X1C.


FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -101,28 +101,44 @@ DEF_OP(LoadNamedVectorConstant) {
...
   // Load the pointer.
-  ldr(TMP1, STATE_PTR(CpuStateFrame, Pointers.Common.NamedVectorConstantPointers[Op->Constant]));
+  auto GenerateMemOperand = [this](uint8_t OpSize, uint32_t NamedConstant, FEXCore::ARMEmitter::Register Base) {
+    const auto ConstantOffset = offsetof(FEXCore::Core::CpuStateFrame, Pointers.Common.NamedVectorConstants[NamedConstant]);
 
+    if (ConstantOffset <= 255) {
+      return ARMEmitter::ExtendedMemOperand(Base.X(), ARMEmitter::IndexType::OFFSET, ConstantOffset);
+    }
+    else if ((ConstantOffset & (OpSize - 1)) == 0 &&  ConstantOffset / OpSize <= 4095 ) {
+      return ARMEmitter::ExtendedMemOperand(Base.X(), ARMEmitter::IndexType::OFFSET, ConstantOffset);
```
alyssarosenzweig, line 111: copypaste fail?

Sonicadvance1: Just a bit tricky here.
The ldr/str code emitters that consume the returned `ARMEmitter::ExtendedMemOperand` support selecting either the 9-bit signed unscaled instruction or 12-bit unsigned scaled instruction depending on constant offset.
I just derped and had if checks to return the same thing.

Combined the if statements so it doesn't look funny when reading it.

FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -101,28 +101,42 @@ DEF_OP(LoadNamedVectorConstant) {
     }
   }
   // Load the pointer.
-  ldr(TMP1, STATE_PTR(CpuStateFrame, Pointers.Common.NamedVectorConstantPointers[Op->Constant]));
+  auto GenerateMemOperand = [this](uint8_t OpSize, uint32_t NamedConstant, FEXCore::ARMEmitter::Register Base) {
+    const auto ConstantOffset = offsetof(FEXCore::Core::CpuStateFrame, Pointers.Common.NamedVectorConstants[NamedConstant]);
 
+    if (ConstantOffset <= 255 || // Unscaled 9-bit signed
+        ((ConstantOffset & (OpSize - 1)) == 0 &&  ConstantOffset / OpSize <= 4095)) /* 12-bit unsigned scaled */ {
```
alyssarosenzweig, line 108: Integer division is potentially expensive, and I don’t expect the compiler knows that op size is a power of two and hence support a more efficient ctz + shift implementation. Do we have a relevant utility function for dividing by powers of 2? If not, perhaps, now is the time.

Sonicadvance1: Added a log2 divider helper and unittest for it.

alyssarosenzweig (approved):

PR #3171, opened 2023-09-30
FEXCore: Merge Arm64Dispatcher in to Dispatcher

Sonicadvance1:
With the removal of the x86 JIT, there is no need to have these be independent classes.

Merges the Arm64Dispatcher in to the base Dispatcher class. No functional change, just moving code.


alyssarosenzweig (approved):
ack, I suppose we can reeval later if there's a case for other backends (riscv or ppc I guess)

PR #3172, opened 2023-10-01
X87F64: Use Bfe for rounding mode, FCHS use float instruction

CallumDev:
Reduces the blowup for FLDCW and FCHS a bit.


Sonicadvance1 (approved):
Nice! More f64 zoomies!

PR #3173, opened 2023-10-01
X87F64: Implement FABS with vector instruction

CallumDev:
- `_VFAbs` added to IR, implemented with fabs arm instruction.
- X87F64 FABS implementation updated to use it.

Reduces fabs instruction count by two :tada:

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -609,6 +609,50 @@ DEF_OP(VAbs) {
   }
 }
 
+DEF_OP(VFAbs) {
+  const auto Op = IROp->C<IR::IROp_VFAbs>();
+  const auto OpSize = IROp->Size;
+
+  const auto ElementSize = Op->Header.ElementSize;
+  const auto Is256Bit = OpSize == Core::CPUState::XMM_AVX_REG_SIZE;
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -609,6 +609,50 @@ DEF_OP(VAbs) {
...
+        }
+        case 4: {
+          fabs(Dst.S(), Src.S());
+          break;
+        }
+        case 8: {
+          fabs(Dst.D(), Src.D());
+          break;
+        }
+        default:
```
lioncash, line 645: Seems like this case would already be covered by the assert doing element-size validation

CallumDev: Is it worth modifying the other instructions to drop this logging then? This is duplicated from e.g. VFAdd

lioncash: Probably. The cases are just old code that happened to be left around when the new assembler/asserts were dropped in

CallumDev: Removed in all

Sonicadvance1 (approved):
Fantastic!

PR #3178, opened 2023-10-04
Minor AVX optimizations

Sonicadvance1:
Some very minor optimizations that I saw we were missing

- Minor optimization in vzeroall to take advantage of `movi` directly to the destination register instead of using a temporary.
- Allow RA to pre-alias and pre-colour AVX registers, which removes a TON of redundant moves in AVX operations.
   - 218 instructions went from non-optimal to optimal


lioncash (approved):
Damn! Very nice!

PR #3179, opened 2023-10-06
FEXCore: Support crypto extensions in HostFeatures override

Sonicadvance1:
Enables in InstCountCI so Pi users can run InstCountCI can run the tests without breaking on crypto operations.

When crypto is enabled or disabled just wholesale change AES, CRC32, and PMULL 128-bit in one step. We don't really care about partial support here.

### DIFF
--- a/FEXCore/Source/Interface/Core/HostFeatures.cpp
+++ b/FEXCore/Source/Interface/Core/HostFeatures.cpp
@@ -125,6 +125,10 @@ static void OverrideFeatures(HostFeatures *Features) {
   const bool EnableFlagM2 = HostFeatures() & FEXCore::Config::HostFeatures::ENABLEFLAGM2;
   LogMan::Throw::AFmt(!(DisableFlagM2 && EnableFlagM2), "Disabling and Enabling CPU features are mutually exclusive");
 
+  const bool DisableCrypto = HostFeatures() & FEXCore::Config::HostFeatures::DISABLECRYPTO;
+  const bool EnableCrypto = HostFeatures() & FEXCore::Config::HostFeatures::ENABLECRYPTO;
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Interface/Core/HostFeatures.cpp
```diff
@@ -125,6 +125,10 @@ static void OverrideFeatures(HostFeatures *Features) {
   const bool EnableFlagM2 = HostFeatures() & FEXCore::Config::HostFeatures::ENABLEFLAGM2;
   LogMan::Throw::AFmt(!(DisableFlagM2 && EnableFlagM2), "Disabling and Enabling CPU features are mutually exclusive");
 
+  const bool DisableCrypto = HostFeatures() & FEXCore::Config::HostFeatures::DISABLECRYPTO;
```
lioncash, line 128: Perhaps these could be extracted into a local helper like:

```cpp
#define ENABLE_DISABLE_OPTION(name, enum_name) \
  const bool Disable##name = (HostFeatures() & FEXCore::Config::HostFeatures::DISABLE##enum_name) != 0; \
  const bool Enable##name = (HostFeatures() & FEXCore::Config::HostFeatures::ENABLE##enum_name) != 0;   \
  LogMan::Throw::AFmt(!(Disable##name && Enable##name), "Disabling and Enabling CPU feature (" #name ") is mutually exclusive");
```

So if an error happens the relevant log will print out which option in particular is making things explode, and condenses all the options a little more

Sonicadvance1: Added a commit to switch over to this.

lioncash (approved):

PR #3184, opened 2023-10-07
ArmEmitter: Adds sized Scalar 1 source and 2 source helpers

Sonicadvance1:
Removes the need for an annoying switch statement with scalar operations for the most part.


Sonicadvance1:
Removing the templated ptrue/ptrues reminded me that I should probably add unittests to the new sized scalar operations.
Which I can do tomorrow.

Sonicadvance1:
Emitter tests added for the new sized scalar emitter functions.

alyssarosenzweig (approved):

PR #3185, opened 2023-10-07
FEXCore/IR: Changes over to automated IR dispatch generation

Sonicadvance1:
Suggested by Alyssa.

Adding an IR operation can be a little tedious since you need to add the definition to JIT.cpp for the dispatch switch, JITClass.h for the function declared, and then actually defining the implementation in the correct file.

Instead support the common case where an IR operation just gets dispatched through to the regular handler. This lets the developer just put the function definition in to the json and the relevent cpp file and it just gets picked up.

Some minor things:
- Needs to support dynamic dispatch for {Load,Store}Register and {Load,Store}Mem
   - This is just a bool in the json
- It needs to not output JIT dispatch for some IR operations
   - SSE4.2 string instructions and x87 operations
   - These go down the "Unhandled" path
- Needs to support a Dispatcher function override
   - This is just for handling NoOp IR operations that get used for other reasons.
- Finally removes VSMul and VUMul, consolidating to VMul
   - Unlike V{U,S}Mull, signed or unsigned doesn't change behaviour here
- Fixed a couple random handler names not matching the IR operation name.


alyssarosenzweig (approved):
Awesome, nice work!

PR #3186, opened 2023-10-08
IR: Adds scalar vector insert operations 

Sonicadvance1:
These IR operations are required to support AFP's NEP mode which does
vector insert in to the destination register. Additionally it gives us
tracking information to allow optimizing out redundant inserts on
devices that don't support AFP natively.

In order to match x86 semantics we need to support binary and unary
scalar operations that do a final insert in to a vector. With optional
zeroing of the top 128-bits for AVX variants.

A tricky thing is that in binary operations this means that the
destination and first source have an intrinsically linked property
depending on if it is SSE or AVX.

SSE example:
- addss xmm0, xmm1
   - xmm0 is both the destination and the first source.
   - This means xmm0[31:0] = xmm0[31:0] + xmm1[31:0]
   - Bits [127:32] are UNMODIFIED.
   - If AVX is supported, Bits[255:128] are also UNMODIFIED.

FEX's JIT jumps through some hoops so that if the destination register
equals the first source register, then it hits the optimal path the
AFP.NEP will insert in to the result. AVX throws a small wrench in to
this due to changed behaviour

AVX example:
- vaddss xmm0, xmm1, xmm2
  - xmm0 is ONLY the destination, xmm1 and xmm2 are the sources
  - This operation copies the bits above the scalar result from the
    first source (xmm1).
  - Additionally this will zero bits above the original 128-bit xmm
    register.
  - xmm0[31:0] = xmm1[31:0] + xmm2[31:0]
  - xmm0[127:32] = xmm1[127:32]
  - ymm0[255:127] = 0

This causes these instructions to support a fairly large table depending
on if the instruction is an SSE or AVX instruction, plus if the host CPU
supports AFP or not.

So while fairly complex, it's handling all the edge cases and gives us
optimization opportunities as we move forward. Currently on non-AFP
supporting devices this has a minor benefit that these IR operations
remove one temporary register, lowering the Register Allocation
overhead.

In the coming weeks I am likely to introduce an optimization pass that
removes redundant inserts because FEX currently does /really/ badly with
scalar code loops.

Needs https://github.com/FEX-Emu/FEX/pull/3184 merged first.


FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/SVEOps.inl
```diff
@@ -1384,6 +1384,12 @@ public:
   }
 
   // SVE predicate initialize
+  void ptrue(SubRegSize size, PRegister pd, PredicatePattern pattern) {
```
lioncash, line 1387: We should probably drop the templated ones if these exist, since they don't really buy us anything that these wouldn't

Sonicadvance1: Removed the templated versions and switched the dispatcher over to it.

FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -14,6 +14,766 @@ tags: backend|arm64
...
+      default:
+        break;
+    }
+    // If the destination is a temporary then it is going to do an insert after the operation.
+    // This means this operation can avoid a redundant insert in this case.
+    const bool DstIsTemp = Dst == VTMP1;
+
+    // Combine results and invert directly in VTMP1.
+    orr(VTMP1.D(), VTMP1.D(), VTMP2.D());
+    mvn(ARMEmitter::SubRegSize::i8Bit, VTMP1.D(), VTMP1.D());
```
alyssarosenzweig, line 675: Ideally we would invert the IR op so we can optimize this out if fed to e.g. selects in the future

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -500,236 +500,492 @@ void OpDispatchBuilder::VectorALUROp<IR::OP_VFSUB, 4>(OpcodeArgs);
...
+
+  StoreResult_WithOpSize(FPRClass, Op, Op->Dest, Dest, DstSize, -1);
+}
+
+OrderedNode* OpDispatchBuilder::InsertCVTGPR_To_FPRImpl(OpcodeArgs,
+                                                  size_t DstSize, size_t DstElementSize,
+                                                  const X86Tables::DecodedOperand& Src1Op,
+                                                  const X86Tables::DecodedOperand& Src2Op,
+                                                  bool ZeroUpperBits) {
+  // We load the full vector width when dealing with a source vector,
```
alyssarosenzweig, line 676: i added a with garbage option to loadsource for this case

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -500,236 +500,492 @@ void OpDispatchBuilder::VectorALUROp<IR::OP_VFSUB, 4>(OpcodeArgs);
...
+
+  static constexpr std::array SourceModes = {
+    FEXCore::IR::Round_Nearest,
+    FEXCore::IR::Round_Negative_Infinity,
+    FEXCore::IR::Round_Positive_Infinity,
+    FEXCore::IR::Round_Towards_Zero,
+    FEXCore::IR::Round_Host,
+  };
+
+  const auto SourceMode = SourceModes[(RoundControlSource << 2) | RoundControl];
```
alyssarosenzweig, line 795: `RoundControlSource ? Round_Host : SourceModes[RCSource]` is a lot less confusing, and drop the host mode from the aarray

alyssarosenzweig (changes requested):

FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
Sonicadvance1: This one is definitely a bit spicy however you slice it right now.
Currently it matches behaviour where the OpcodeDispatcher used to put an MVN.
Theoretically it matches behaviour of ScalarEmitORD, but with a mvn at the end. So we could optimize to ordered.

An additional quirk that gets thrown in here is that SVE adds an unordered fcmp directly in to the ISA, so we could make a more optimal solution in the backend in this case regardless. Since then it turns in to fcmuo + (I think there is an instruction that converts a predicate to a vector mask?)

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
Sonicadvance1: Added XMM support for garbage option and used that

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
Sonicadvance1: Switched over to ternary.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5030,7 +5030,7 @@ OrderedNode *OpDispatchBuilder::LoadSource_WithOpSize(FEXCore::IR::RegisterClass
       // Now extract the subregister if it was a partial load /smaller/ than SSE size
       // TODO: Instead of doing the VMov implicitly on load, hunt down all use cases that require partial loads and do it after load.
```
alyssarosenzweig, line 5031: drop the todo maybe?

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
alyssarosenzweig: not seeing it used..

Sonicadvance1:
```diff
diff --git a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
index 67695dde9..cd3f7d1c5 100644
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -508,10 +508,10 @@ OrderedNode* OpDispatchBuilder::VectorScalarInsertALUOpImpl(OpcodeArgs, IROps IR
   // We load the full vector width when dealing with a source vector,
   // so that we don't do any unnecessary zero extension to the scalar
   // element that we're going to operate on.
-  const auto SrcSize = Op->Src[0].IsGPR() ? GetGuestVectorLength() : GetSrcSize(Op);
+  const auto SrcSize = GetSrcSize(Op);

   OrderedNode *Src1 = LoadSource_WithOpSize(FPRClass, Op, Src1Op, DstSize, Op->Flags, -1);
-  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1);
+  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1, true, false, MemoryAccessType::ACCESS_DEFAULT, true);

   // If OpSize == ElementSize then it only does the lower scalar op
   auto ALUOp = _VFAddScalarInsert(IR::SizeToOpSize(DstSize), ElementSize, Src1, Src2, ZeroUpperBits);
@@ -593,10 +593,10 @@ OrderedNode* OpDispatchBuilder::VectorScalarUnaryInsertALUOpImpl(OpcodeArgs, IRO
   // We load the full vector width when dealing with a source vector,
   // so that we don't do any unnecessary zero extension to the scalar
   // element that we're going to operate on.
-  const auto SrcSize = Op->Src[0].IsGPR() ? GetGuestVectorLength() : GetSrcSize(Op);
+  const auto SrcSize = GetSrcSize(Op);

   OrderedNode *Src1 = LoadSource_WithOpSize(FPRClass, Op, Src1Op, DstSize, Op->Flags, -1);
-  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1);
+  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1, true, false, MemoryAccessType::ACCESS_DEFAULT, true);

   // If OpSize == ElementSize then it only does the lower scalar op
   auto ALUOp = _VFSqrtScalarInsert(IR::SizeToOpSize(DstSize), ElementSize, Src1, Src2, ZeroUpperBits);
@@ -732,10 +732,10 @@ OrderedNode* OpDispatchBuilder::InsertScalar_CVT_Float_To_FloatImpl(OpcodeArgs,
   // We load the full vector width when dealing with a source vector,
   // so that we don't do any unnecessary zero extension to the scalar
   // element that we're going to operate on.
-  const auto SrcSize = Op->Src[0].IsGPR() ? GetGuestVectorLength() : GetSrcSize(Op);
+  const auto SrcSize = GetSrcSize(Op);

   OrderedNode *Src1 = LoadSource_WithOpSize(FPRClass, Op, Src1Op, DstSize, Op->Flags, -1);
-  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1);
+  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1, true, false, MemoryAccessType::ACCESS_DEFAULT, true);

   return _VFToFScalarInsert(IR::SizeToOpSize(DstSize), DstElementSize, SrcElementSize, Src1, Src2, ZeroUpperBits);
 }
@@ -772,27 +772,22 @@ OrderedNode* OpDispatchBuilder::InsertScalarRoundImpl(OpcodeArgs,
   // We load the full vector width when dealing with a source vector,
   // so that we don't do any unnecessary zero extension to the scalar
   // element that we're going to operate on.
-  const auto SrcSize = Op->Src[0].IsGPR() ? GetGuestVectorLength() : GetSrcSize(Op);
+  const auto SrcSize = GetSrcSize(Op);

   OrderedNode *Src1 = LoadSource_WithOpSize(FPRClass, Op, Src1Op, DstSize, Op->Flags, -1);
-  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1);
+  OrderedNode *Src2 = LoadSource_WithOpSize(FPRClass, Op, Src2Op, SrcSize, Op->Flags, -1, true, false, MemoryAccessType::ACCESS_DEFAULT, true);

```

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: I'd say the todo is still relevant because there is code still doing vmovs when they don't need to. Particularly relevant to AVX code that is doing it.

alyssarosenzweig (approved):

PR #3192, opened 2023-10-10
OpcodeDispatcher: Optimizes < 32-bit register push 

Sonicadvance1:
We don't need zero in the upper bits for a push.
Makes a couple variants optimal.

Adds missing tests to the 32-bit file, since only 32-bit can push a
32-bit register.

Removes a spurious Test.json from InstCountCI which is my local testing file.


alyssarosenzweig (approved):
We probably want a helper for load with garbage…

PR #3195, opened 2023-10-11
Thunks: Fixes missing vulkan definitions

Sonicadvance1:
A couple of games were hitting these. Not sure how they were missed in PR #3159 but adds the missing one.

Small rearrangement to make this easier as well. Hopefully thunk stuff lands sooner rather than later to automate this for Vulkan.

Maybe `-isystem` instead of `-I` needs to be used unlike what #2076 described, might depend on what is installed on the host system.


alyssarosenzweig (approved):

neobrain:
> Not sure how they were missed in PR https://github.com/FEX-Emu/FEX/pull/3159 but adds the missing one.

I followed the instructions of https://github.com/FEX-Emu/FEX/pull/2076 word for word, so this is suspicious. Does executing the specific commands mentioned in that PR produce the same result as https://github.com/FEX-Emu/FEX/pull/3159 for you?

Would be good to explicitly note down the proper instructions rather than referring to a chain of old PRs in any case.

Sonicadvance1:
> > Not sure how they were missed in PR https://github.com/FEX-Emu/FEX/pull/3159 but adds the missing one.
> 
> I followed the instructions of #2076 word for word, so this is suspicious. Does executing the specific commands mentioned in that PR produce the same result as #3159 for you?
> 
> Would be good to explicitly note down the proper instructions rather than referring to a chain of old PRs in any case.

Tried both `-i` and `-isystem` and they produced nearly identical results. Just some additional non-vulkan symbols pulled in. So not sure why your PR missed these.

neobrain:
> I followed the instructions of https://github.com/FEX-Emu/FEX/pull/2076 word for word, so this is suspicious. Does executing the specific commands mentioned in that PR produce the same result as https://github.com/FEX-Emu/FEX/pull/3159 for you?

For reference, the instructions in that pull request had a typo (`Externals` instead of `External`) that likely caused system headers to be picked up instead of our externals module. This caused the definitions to be updated to the system version, but not the intended one.

I have edited the comment to fix the typo now.

PR #3199, opened 2023-10-15
OpcodeDispatcher: Put extra LoadSource options in a struct

lioncash:
Allows for easier expansion without needing to expand the function definitons.

Also makes a few usages significantly less verbose and makes specifying options a little more declarative, rather than having to memorize what each argument is specifying.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
@@ -27,6 +27,43 @@ namespace FEXCore::IR {
 class Pass;
 class PassManager;
 
+enum class MemoryAccessType {
+  // Choose TSO or Non-TSO depending on access type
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -27,6 +27,43 @@ namespace FEXCore::IR {
 class Pass;
 class PassManager;
 
+enum class MemoryAccessType {
+  // Choose TSO or Non-TSO depending on access type
+  ACCESS_DEFAULT,
```
neobrain, line 32: No need to use an `ACCESS_` prefix since it's already an enum class.

lioncash: Amended

Sonicadvance1 (approved):

PR #3205, opened 2023-10-18
VectorOps: Handle SVE VSMax/VSMin and VUMax/VUMin paths a little better

lioncash:
When putting the SVE equivalents in, I neglected to handle the case where the destination aliases the second input source, which can also be done with one instruction like the case of the destination aliasing the first input source.


FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -2204,14 +2208,18 @@ DEF_OP(VSMin) {
...
+      smin(SubRegSize, Dst.Z(), Pred, Dst.Z(), Vector1.Z());
     } else {
-      // SVE SMIN is a destructive operation, so we need a temporary.
-      movprfx(VTMP1.Z(), Vector1.Z());
-      smin(SubRegSize, VTMP1.Z(), Pred, VTMP1.Z(), Vector2.Z());
-      mov(Dst.Z(), VTMP1.Z());
+      // SVE SMIN is a destructive operation, but we know nothing is
+      // aliasing the destination by this point, so we can move into
+      // the destination without needing a temporary.
+      mov(Dst.Z(), Vector1.Z());
```
Sonicadvance1, line 2221: Do we still want to use movprfx on these so we get the benefit of the instruction fusing to the next one?

lioncash: Oh, yes we do. Dunno why I thought it wasn't necessary anymore.

Sonicadvance1 (approved):

PR #3207, opened 2023-10-19
VectorOps: Handle SVE VFDiv a little better

lioncash:
In the event no source vectors alias the destination, we can just move the first source vector into it and then perform the divide without needing to move afterword.


Sonicadvance1 (approved):
The fastest float divides in all the lands.

PR #3210, opened 2023-10-19
VectorOps: Handle SVE VFCADD a little better

lioncash:
If no registers alias, then we can move the first source directly into the destination and then perform the FCADD operation as opposed to using a temporary.


Sonicadvance1 (approved):
Those damned moves

PR #3211, opened 2023-10-19
VectorOps: Handle SVE VExtr a little better

lioncash:
If the source registers don't alias the destination, then we can safely move the lower bits over to it without using a temporary.


Sonicadvance1 (approved):
yay those damn moves

PR #3212, opened 2023-10-19
OpcodeDispatcher: Optimize 128-bit DPPS and DPPD 

Sonicadvance1:
These instructions aren't super amazing due to the fact that they have
both a source mask and a destination duplication mask.

Setup a case where we can generate more optimal code in /most/ cases.

There are a few that still fall down a "bad" path for the result
broadcast but in most cases they are optimal. Still to be seen what
games typically use the broadcast mask as.

AVX in its infinite wisdom expanded DPPS to 256-bit, while leaving DPPD
to only support 128-bit still. This leaves the original implementation
alone for 256-bit DPPS since I don't want to break it.

This is another instruction that gets a free optimization when
SVE-128bit is supported!


Sonicadvance1:
Added ASM unittests for this, including one that uses some more aggressive masking tests for dppd even though it is unnecessary, Since it could have been a bug if I didn't change behaviour slightly.
ASM tests now test all broadcast mask options because that is most likely for something to break

lioncash (approved):

PR #3215, opened 2023-10-23
FEXLoader: Query runtime page size

alyssarosenzweig:
This lets most of the ASM tests run on 16K Linux hosts which is good because I have a Mac and I'm bad at computer.


Sonicadvance1 (approved):
Nice. Surprised the mingw stuff didn't complain but fine for me.

PR #3162, opened 2023-09-27
Keep guest SF/ZF/CF/OF flags resident in host NZCV

alyssarosenzweig:
This adds all the infrastructure needed to take advantage of flag manipulation instructions.

PR #3216, opened 2023-10-23
Prep commits for NZCV modelling

alyssarosenzweig:
split from #3162, should be no functional/performance changes


Sonicadvance1 (approved):

FEXCore/include/FEXCore/IR/IREmitter.h
```diff
@@ -332,6 +334,10 @@ friend class FEXCore::IR::PassManager;
       return Ptr;
     }
 
+    virtual void SaveNZCV() {
```
Sonicadvance1, line 337: oop, this should be override at the end instead of virtual. like in OpcodeDispatcher.h

Sonicadvance1 (approved):
Nice little prep work, shows how we need to solve 256-bit shuffles in the future.

PR #3218, opened 2023-10-23
OpcodeDispatcher: Optimize DF pointer offset calculation

Sonicadvance1:
Previously this moved two constant, did a compare and a csel. Four
instructions in total. It also corrupts NZCV which we want to use for
other things.

This new codegen emits one constant and one subtract instruction, two
instructions total and doesn't touch NZCV.

More optimal!

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -187,6 +187,16 @@ DEF_OP(Sub) {
   }
 }
 
+DEF_OP(SubShift) {
+  auto Op = IROp->C<IR::IROp_SubShift>();
+  const uint8_t OpSize = IROp->Size;
+
+  LOGMAN_THROW_AA_FMT(OpSize == 4 || OpSize == 8, "Unsupported {} size: {}", __func__, OpSize);
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  sub(EmitSize, GetReg(Node), GetReg(Op->Src1.ID()), GetReg(Op->Src2.ID()), ConvertIRShiftType(Op->Shift), Op->ShiftAmount);
+}
+
 DEF_OP(SubNZCV) {
   auto Op = IROp->C<IR::IROp_SubNZCV>();
   const IR::OpSize OpSize = Op->Size;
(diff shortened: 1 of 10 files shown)


FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -187,6 +187,21 @@ DEF_OP(Sub) {
...
 }
 
+DEF_OP(SubShift) {
+  auto Op = IROp->C<IR::IROp_SubShift>();
+  const uint8_t OpSize = IROp->Size;
+
+  LOGMAN_THROW_AA_FMT(OpSize == 4 || OpSize == 8, "Unsupported {} size: {}", __func__, OpSize);
+  const auto EmitSize = OpSize == 8 ? ARMEmitter::Size::i64Bit : ARMEmitter::Size::i32Bit;
+
+  const auto Shift =
```
alyssarosenzweig, line 197: I suspect this wants to be a common helper

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -3858,14 +3858,10 @@ void OpDispatchBuilder::STOSOp(OpcodeArgs) {
...
-    auto SizeConst = _Constant(Size);
-    auto NegSizeConst = _Constant(-Size);
-
     // Calculate direction.
     auto DF = GetRFLAG(FEXCore::X86State::RFLAG_DF_LOC);
-    auto PtrDir = _Select(FEXCore::IR::COND_EQ,
-        DF,  _Constant(0),
-        SizeConst, NegSizeConst);
+    auto SizeConst = _Constant(Size);
+    auto PtrDir = _SubShift(IR::SizeToOpSize(CTX->GetGPRSize()), SizeConst, DF, ShiftType::LSL, FEXCore::ilog2(Size) + 1);
```
alyssarosenzweig, line 3864: ........lol.

FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
Sonicadvance1: Moved to common helper.

FEXCore/include/FEXCore/IR/IR.h
```diff
@@ -564,6 +564,12 @@ enum class FloatCompareOp : uint8_t {
   ORD,
 };
 
+enum class ShiftType : uint8_t {
+  LSL = 0,
+  LSR,
+  ASR,
+};
```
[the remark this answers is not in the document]
Sonicadvance1: Added rotate for when bitwise ops with shifted register become a thing.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Lol indeed

alyssarosenzweig (approved):

PR #3220, opened 2023-10-23
InstCountCI: Support disabling flagm extensions 

Sonicadvance1:
This is necessary so https://github.com/FEX-Emu/FEX/pull/3162 can give consistent results


alyssarosenzweig (approved):

alyssarosenzweig:
do we even care to instcountci the non-flagm case...?

Sonicadvance1:
> do we even care to instcountci the non-flagm case...?

Might as well since it's trivial to do.

PR #3227, opened 2023-10-26
IR: Support custom assert messages for IR validation

Sonicadvance1:
When the assert message is empty it is hard to tell what went wrong. Support custom LogMan messages to make this nicer.

Only adds a few custom messages so far.

### DIFF
--- a/FEXCore/Scripts/json_ir_generator.py
+++ b/FEXCore/Scripts/json_ir_generator.py
@@ -253,12 +253,21 @@ def parse_ops(ops):
                 for i in range(len(OpDef.EmitValidation)):
                     # Patch up all the argument names
                     for Arg in OpDef.Arguments:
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (commented):
I’m not sure how I feel about this one. On one hand, custom validate messages can be useful for nontrivial, and non-obvious assertions. But when the thing we’re starting is “size equals 32” and the message is “size must be 32 bits “… It doesn’t seem clear to me that this PR is actually an improvement in developer happiness, because we’ve replaced information that is easy to automatically generate with information that requires manual work to keep up-to-date in the json file… I don’t remember if our FEX assert includes the source code of the assert itself in the output, if not, that would be a good place to start, and then many/most of the messages here would make sense to drop IMHO

Sonicadvance1:
Closing in favour of #3232

PR #3228, opened 2023-10-26
FEXCore: Support forcing all allocations to the upper 48-bits

Sonicadvance1:
This is a debug feature that is not entirely stable but is super useful when trying to profile FEX's memory allocations on 64-bit applications.

Typically FEX just mixes its own allocations with the guest application's allocations for 64-bit, but we can pass the 48-bit allocation to our memory allocator and force everything in to that space with very little effort. This sets the jemalloc hooks when enabled which adds a little bit of overhead.

Once the option is enabled you can look at pmap -x <pid> to see the separation between FEX and guest allocations since all FEX allocations end up in the high bits after VDSO is loaded.

Very handy!

### DIFF
--- a/FEXCore/Source/Utils/Allocator/HostAllocator.h
+++ b/FEXCore/Source/Utils/Allocator/HostAllocator.h
@@ -1,7 +1,11 @@
 // SPDX-License-Identifier: MIT
 #pragma once
+
+#include <FEXCore/Utils/Allocator.h>
+
 #include <FEXCore/fextl/allocator.h>
 #include <FEXCore/fextl/memory.h>
+#include <FEXCore/fextl/vector.h>
 
 #include <cstddef>
 #include <cstdint>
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Utils/Allocator/HostAllocator.h
```diff
@@ -46,5 +50,5 @@ namespace Alloc {
 namespace Alloc::OSAllocator {
 void RegisterTLSData(FEXCore::Core::InternalThreadState *Thread);
 void UninstallTLSData(FEXCore::Core::InternalThreadState *Thread);
-fextl::unique_ptr<Alloc::HostAllocator> Create64BitAllocator();
+fextl::unique_ptr<Alloc::HostAllocator> Create64BitAllocator(fextl::vector<FEXCore::Allocator::MemoryRegion> *MemoryRegion = nullptr);
```
neobrain, line 53: This looks really poorly designed. Previously this function had reasonably obvious behavior, but with this change you really have open up the implementation to see what's going on. Can you come up with a more predictable interface that doesn't use a heavy `std::vector` in-out parameter?

neobrain (changes requested):

Sonicadvance1:
I'll leave this alone for now and keep the branch for reference.

PR #3230, opened 2023-10-26
IR: Optimize unused result atomic fetch mop to just atomic mop

Sonicadvance1:
When the result of an atomic fetch operation is unused then we can
safely convert it to a non-fetching version of the operation.

This happens hundreds of times per process as far as I can tell. No idea
if this actually helps any hardware, but theoretically it can allow CPUs
to not stall waiting for writeback of the atomic operation.
Couldn't detect any performance improvements in the various little
things I was poking at least. Very trivial to support so add it.

Unaligned variants are already handled in our unaligned fault handler
since the only difference is the acquire semantic is dropped and the
destination register is the zero register.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/AtomicOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/AtomicOps.cpp
@@ -193,6 +193,33 @@ DEF_OP(AtomicAnd) {
   }
 }
 
+DEF_OP(AtomicCLR) {
+  auto Op = IROp->C<IR::IROp_AtomicCLR>();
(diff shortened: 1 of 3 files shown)


alyssarosenzweig (approved):
It took me a moment, understand why it was safe to do  the op code switching… Still not a huge fan, given the unsafe to, but I’ve written code of those exactly that so I’m not gonna be a hypocrite and tell you to change it 😝

Sonicadvance1:
Added a comment to the atomic operations that they must match their non-fetch or fetch counterparts so we can can remember this a year later when changing things and trying to figure out why the world broken.

PR #3231, opened 2023-10-26
OpcodeDispatcher: Allow garbage in upper bits for more ALU ops 

Sonicadvance1:
Secondary ALU operations were missed and when the operation is 4-bytes
in size we can also allow garbage upper bits since the JIT will emit a
32-bit operation for this instruction which is safe.

Optimizes some bad codegen around 32-bit ALU operations.


alyssarosenzweig (approved):
Very nice

PR #3234, opened 2023-10-27
Arm64: Minor optimization to bfxil and bfi 

Sonicadvance1:
When the destination doesn't alias the source, we can remove a final mov
from both of these operations.

Does some minor code improvement.


alyssarosenzweig (approved):
nice!

PR #3238, opened 2023-10-27
OpcodeDispatcher: Optimize blendps 

Sonicadvance1:
A bunch of blendps swizzles weren't optimal. This optimizes all swizzles
to be optimal.

Two instructions can be more optimal without a tbx but the rest required
tbx to be optimal since they don't match ARM's swizzle mechanics.


Sonicadvance1:
Weird behaviour here, InstcountCI failed on main because of the previous dpps optimization. This was because somehow the InstCountCI runners managed to miss that `vdpps xmm` should have optimized the dual faddp instructions in to a vaddv.

Not sure how this was missed but fixed in this PR for the map3 file.

lioncash (approved):

PR #3241, opened 2023-10-29
Thunks/xcb: Drop unused and incomplete support for asynchronous callbacks

neobrain:
These have required active maintenance in the past (e.g. #2583 and #2663) and interfere heavily with the upcoming changes from #3135. We know no currently supported use case that actually requires this feature, so just remove it.

Simple detection code is added to error out if the callback is invoked asynchronously. If this gets hit in a use case we care about, #2128 implements an alternative, more generic approach that should be integrated instead.


Sonicadvance1 (approved):
Looks like I can't find the games that this breaks locally on my setup at least. So yolo.

PR #3243, opened 2023-11-01
FEXCore/FileLoading: Updates helper to load file that is backed by memory

Sonicadvance1:
When attempting to read files that aren't backed by a filesystem then
our current read file helpers fail since they query the file size
upfront.

Change the helper so that it doesn't query the size and just reads the file if it
can be opened. This lets us read `/proc/self/maps` using helpers.

### DIFF
--- a/FEXCore/Source/Utils/FileLoading.cpp
+++ b/FEXCore/Source/Utils/FileLoading.cpp
@@ -24,24 +24,48 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
   size_t FileSize{};
   if (FixedSize == 0) {
     struct stat buf;
-    if (fstat(FD, &buf) != 0) {
-      close(FD);
-      return false;
+    if (fstat(FD, &buf) == 0) {
+      FileSize = buf.st_size;
     }
-
-    FileSize = buf.st_size;
   }
   else {
     FileSize = FixedSize;
   }
 
   ssize_t Read = -1;
-  if (FileSize > 0) {
+  bool LoadedFile{};
+  if (FileSize) {
+    // File size is known upfront
     Data.resize(FileSize);
     Read = pread(FD, &Data.at(0), FileSize, 0);
+
+    LoadedFile = Read == FileSize;
+  }
+  else {
+    // The file is either empty or its size is unknown (e.g. procfs data).
+    // Try reading in chunks instead
+    ssize_t CurrentOffset = 0;
+    constexpr size_t READ_SIZE = 4096;
+    Data.resize(READ_SIZE);
+
+    while ((Read = pread(FD, &Data.at(CurrentOffset), READ_SIZE, CurrentOffset)) == READ_SIZE) {
+      CurrentOffset += Read;
+      Data.resize(CurrentOffset + Read);
+    }
+
+    if (Read == -1) {
+      Data.clear();
+      close(FD);
+      return false;
+    }
+
+    // Final resize to ensure there is no garbage data past the end.
+    Data.resize(CurrentOffset + Read);
+
+    LoadedFile = true;
   }
   close(FD);
-  return Read == FileSize;
+  return LoadedFile;
 }
 
 ssize_t LoadFileToBuffer(const fextl::string &Filepath, std::span<char> Buffer) {
--- a/FEXCore/unittests/APITests/FileLoading.cpp
+++ b/FEXCore/unittests/APITests/FileLoading.cpp
@@ -1,0 +1,22 @@
+#include <FEXCore/Utils/FileLoading.h>
+#include <catch2/catch.hpp>
+
+TEST_CASE("LoadFile-Doesn'tExist") {
+  fextl::string MapsFile;
+  auto Read = FEXCore::FileLoading::LoadFile(MapsFile, "/tmp/a/b/c/d/e/z");
+  REQUIRE(MapsFile.size() == 0);
+  REQUIRE(Read == false);
+}
+
+TEST_CASE("LoadFile-procfs") {
+  fextl::string MapsFile;
+  FEXCore::FileLoading::LoadFile(MapsFile, "/proc/self/maps");
+  REQUIRE(MapsFile.size() != 0);
+}
+
+TEST_CASE("LoadFile-Buffer") {
+  fextl::string MapsFile;
+  MapsFile.resize(16);
+  auto Read = FEXCore::FileLoading::LoadFileToBuffer("/proc/self/maps", MapsFile);
+  REQUIRE(MapsFile.size() == Read);
+}


neobrain:
Could you open a PR to illustrate how you intend to use this new helper?

Sonicadvance1:
> Could you open a PR to illustrate how you intend to use this new helper?

```diff
    fextl::string MapsFile;
--  FEXCore::FileLoading::LoadFile(MapsFile, "/proc/self/maps");
++  FEXCore::FileLoading::LoadFileUnsized(MapsFile, "/proc/self/maps");
    fextl::istringstream MapsStream(MapsFile);
```

   This is currently broken today.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -44,6 +44,31 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
   return Read == FileSize;
 }
 
+template<typename T>
+static bool LoadFileUnsizedImpl(T &Data, const fextl::string &Filepath) {
```
neobrain, line 48: This doesn't need to be a template.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -70,6 +95,11 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
   return !f.fail();
 }
 
+template<typename T>
+static bool LoadFileUnsizedImpl(T &Data, const fextl::string &Filepath) {
```
neobrain, line 99: This seems to be defined twice.

neobrain:
> ```diff
>     fextl::string MapsFile;
> --  FEXCore::FileLoading::LoadFile(MapsFile, "/proc/self/maps");
> ++  FEXCore::FileLoading::LoadFileUnsized(MapsFile, "/proc/self/maps");
>     fextl::istringstream MapsStream(MapsFile);
> ```
> 
> This is currently broken today.

What does fstat return for those files? Why not add an internal fallback implementation to LoadFile instead of adding a new, undocumented interface?

neobrain:
For tiny changes like this, I'd prefer if a use case (or at least tests) was included in the PR (or a followup PR were already opened for bigger ones), since I can't reasonably review new interfaces when they're not used.

Sonicadvance1:
Updated the original helper so that if fstat returns zero sized files like procfs that will attempt reading the file incrementally, removing the unsized helper.
Adding a unittest so there is something using this.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -24,24 +24,43 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
...
     FileSize = FixedSize;
   }
 
   ssize_t Read = -1;
-  if (FileSize > 0) {
-    Data.resize(FileSize);
-    Read = pread(FD, &Data.at(0), FileSize, 0);
+  bool LoadedFile{};
+  if (FileSize) {
+    // Filesize is known upfront.
```
neobrain, line 38:
```suggestion
    // File size is known upfront
```

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -24,24 +24,43 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
...
   else {
     FileSize = FixedSize;
   }
 
   ssize_t Read = -1;
-  if (FileSize > 0) {
-    Data.resize(FileSize);
-    Read = pread(FD, &Data.at(0), FileSize, 0);
+  bool LoadedFile{};
+  if (FileSize) {
```
neobrain, line 37: This will handle empty files poorly. Not the most important use case, but easily avoided by changing `FileSize` to be an `std::optional<size_t>`.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -24,24 +24,43 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
...
+    // Filesize is known upfront.
+    if (FileSize > 0) {
+      Data.resize(FileSize);
+      Read = pread(FD, &Data.at(0), FileSize, 0);
+    }
+
+    LoadedFile = Read == FileSize;
+  }
+  else {
+    // Filesize is unknown, try reading.
```
neobrain, line 47:
```suggestion
    // File size is unknown. Try reading in chunks instead
```

FEXCore/Source/Utils/FileLoading.cpp
Sonicadvance1: We can't distinguish empty files from memory backed files, fstat returns success for both, with st_size being zero for both.

neobrain: Oof 😩

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -24,24 +24,43 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
...
   }
 
   ssize_t Read = -1;
-  if (FileSize > 0) {
-    Data.resize(FileSize);
-    Read = pread(FD, &Data.at(0), FileSize, 0);
+  bool LoadedFile{};
+  if (FileSize) {
+    // Filesize is known upfront.
+    if (FileSize > 0) {
```
neobrain, line 39: This is always false.

FEXCore/Source/Utils/FileLoading.cpp
neobrain: In light of https://github.com/FEX-Emu/FEX/pull/3243/files#r1381421301 , this is worth extra emphasis:

```suggestion
    // The file is either empty or its size is unknown (e.g. procfs data).
    // Try reading in chunks instead
```

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -24,24 +24,47 @@ static bool LoadFileImpl(T &Data, const fextl::string &Filepath, size_t FixedSiz
...
+  }
+  else {
+    // The file is either empty or its size is unknown (e.g. procfs data).
+    // Try reading in chunks instead
+    ssize_t CurrentOffset = 0;
+    constexpr size_t READ_SIZE = 4096;
+    Data.resize(READ_SIZE);
+
+    while ((Read = pread(FD, &Data.at(CurrentOffset), READ_SIZE, CurrentOffset)) == READ_SIZE) {
+      if (Read == -1) {
```
neobrain, line 52: This needs to be moved out of the loop, otherwise it will never be true.

neobrain (approved):
The `pread` error check needs to be moved, good to go then :)

PR #3244, opened 2023-11-01
FEXCore: Adds ability to clear InterruptableConditionVariable mutex

Sonicadvance1:
There was no way to change this from the signaled state but there are cases where we need to unsignal and wait on the mutex again.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
+++ b/FEXCore/include/FEXCore/Utils/InterruptableConditionVariable.h
@@ -71,6 +71,9 @@ namespace FEXCore {
         DoNotify(INT_MAX);
       }
 
+      void ClearMutex() {
+        Mutex = UNSIGNALED;
+      }
     private:
       std::atomic<uint32_t> Mutex{};
       constexpr static uint32_t SIGNALED = 1;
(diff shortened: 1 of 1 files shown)


neobrain:
Could you open a PR to illustrate how you intend to use the new interface?

Sonicadvance1:
> Could you open a PR to illustrate how you intend to use the new interface?

```diff
++void GdbServer::Pause() {
++  // GDBServer is requesting that FEX pauses the entire process.
++  // We need to safely signal all the threads to pause.
++  CurrentSIGSEGVStatus = SIGSEGVStatus::PAUSING;
++
++  // Lock the thread creation mutex and update all thread's interrupt fault check pages.
++  {
++    std::lock_guard<std::mutex> lk(CTX->ThreadCreationMutex);
++    for (auto &Thread : CTX->Threads) {
++      void *InterruptFaultPage = Thread->InterruptFaultPage;
++      Thread->StartRunning.ClearMutex(); ///<---------------------- This one right here. Variable likely needs a rename
++      mprotect(InterruptFaultPage, FHU::FEX_PAGE_SIZE, PROT_NONE);
++    }
++
++    CTX->WaitForIdle();
++  }
++
++  CurrentSIGSEGVStatus = SIGSEGVStatus::DEFAULT;
++}

neobrain:
I can't comment on individual lines in a diff comment, and in general I'd like to understand this use case better than by doing guesswork. Could you open a proper PR?

In general, I don't think it makes sense to add unused interfaces in a dedicated PR if it's about such small changes.

PR #3251, opened 2023-11-04
FEXLinuxTests: Compile tests with masm=intel

Sonicadvance1:
ATT is maddening and a unittest I'm writing was going to be a nightmare.

### DIFF
--- a/unittests/FEXLinuxTests/tests/CMakeLists.txt
+++ b/unittests/FEXLinuxTests/tests/CMakeLists.txt
@@ -8,6 +8,11 @@ unset(CMAKE_CXX_FLAGS)
 
 set(GENERATE_GUEST_INSTALL_TARGETS TRUE)
 
+# Use intel masm syntax. ATT style asm syntax is archaic and hard to read.
+add_compile_options(-masm=intel)
+# Override a define so catch2 doesn't use ATT style inline asm
+add_definitions(-D"CATCH_BREAK_INTO_DEBUGGER\(\)"="[]{ if\( Catch::isDebuggerActive\(\) \) { __builtin_trap\(\)\; } }\(\)")
+
 file(GLOB_RECURSE TESTS CONFIGURE_DEPENDS *.cpp)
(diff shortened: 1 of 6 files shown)


unittests/FEXLinuxTests/tests/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@ unset(CMAKE_CXX_FLAGS)
 
 set(GENERATE_GUEST_INSTALL_TARGETS TRUE)
 
+# Use intel masm syntax. ATT style is maddening
+add_compile_options(-masm=intel)
+# Override a define so catch2 doesn't use ATT style inline asm
+add_definitions(-D"CATCH_BREAK_INTO_DEBUGGER\(\)"=)
```
neobrain, line 14: What's the effect of this? Sounds like this disables breakpoint triggering when a test failure is hit, which would be highly undesirable.

unittests/FEXLinuxTests/tests/CMakeLists.txt
```diff
@@ -8,6 +8,11 @@ unset(CMAKE_CXX_FLAGS)
 
 set(GENERATE_GUEST_INSTALL_TARGETS TRUE)
 
+# Use intel masm syntax. ATT style is maddening
```
neobrain, line 11: Please let's keep documentation factual instead of using overly emotional language that doesn't convey any extra information.

unittests/FEXLinuxTests/tests/CMakeLists.txt
Sonicadvance1: This disabled the -b or `shouldDebugBreak` option. Changed it so it uses `__builtin_trap` so it still works.

unittests/FEXLinuxTests/tests/CMakeLists.txt
Sonicadvance1: Changed the comment.

alyssarosenzweig (approved):
...and good riddance

PR #3252, opened 2023-11-05
ARMEmitter: Fix GPR fill mask in `FillStaticRegs`

Sonicadvance1:
This mask was being used incorrectly, it's a GPR spill mask for host
GPRs not an index in to the SRA array. Search the array of SRA registers
for the first one in the mask first to use as a temporary.

Fixes an issue with 32-bit inline syscalls where the first register
being spilled was r8, which was beyond the size of SRA registers on
32-bit processes. This would cause FEX to read the value just after
x32::SRA which is x32::RA. This would mean it would use r20 as a
temporary, corrupting the register in the process.

I noticed this while poking at https://github.com/FEX-Emu/FEX/pull/3162, but also when I was looking at a
memory buffer ownership problem.


alyssarosenzweig (approved):
ooof

PR #3254, opened 2023-11-06
NZCV health checks (for CI only)

alyssarosenzweig:
Debug version of #3162 , I hope


alyssarosenzweig:
This has served its purpose, closing

PR #3255, opened 2023-11-06
FEXLinuxTests: Adds a unittest for eflags and signals around a inlined syscall

Sonicadvance1:
When attempting to debug #3162 I had noticed spurious behaviour around what I assumed to be eflags getting corrupt around inlined syscalls. This turned out to be a red herring but to ensure we are still testing this, create a fully fleshed out unit test.

This test ensures a couple of things.
1) A flag that is set or unset before a syscall doesn't have its data
   corrupt
2) An inline syscall doesn't corrupt the eflags, checking the eflag
   result after returning from the syscall.
3) A signal occuring while in an inline syscall returns the correct
   eflags information in the signal handler information

This test gets accomplished by setting or unsetting a particular flag and then calling the futex syscall in a way that is guaranteed to be inlined and also wait forever. Then the parent thread will signal with a SIGTERM and read back the signal information. It does this multiple times for each flag we care about.

### DIFF
--- a/unittests/FEXLinuxTests/tests/signal/eflags_signal.cpp
+++ b/unittests/FEXLinuxTests/tests/signal/eflags_signal.cpp
@@ -1,0 +1,303 @@
+#include <atomic>
+#include <catch2/catch.hpp>
+#include <fstream>
+#include <sys/syscall.h>
+#include <linux/futex.h>
+#include <signal.h>
+#include <optional>
+#include <thread>
+
+#if __SIZEOF_POINTER__ == 4
+#define DO_ASM(x, y) \
+  __asm volatile( \
+  x \
+  /* Need to late move syscall number since incoming asm will overwrite eax */ \
+  " mov eax, %[Syscall];" \
+  /* Notify we are ready (Without touching flags) */ \
+  "mov dword ptr [%[ReadyNotify]], 1;" \
+  /* Do a futex */ \
+  "int 0x80;" \
+  y \
+  : \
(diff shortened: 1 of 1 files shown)


alyssarosenzweig (approved):

unittests/FEXLinuxTests/tests/signal/eflags_signal.cpp
```diff
@@ -0,0 +1,303 @@
...
+  // Wait for thread to get in to the futex.
+  WaitForThreadAsleep(ThreadID.load());
+
+  tgkill(::getpid(), ThreadID.load(), SIGTERM);
+
+  t.join();
+
+  REQUIRE(from_handler.has_value());
+  REQUIRE(from_handler.value().Signal == SIGTERM);
+  REQUIRE(((from_handler.value().eflags >> FlagLocation) & 1) == FlagValue);
```
neobrain, line 270: REQUIRE will abort any later checks and hence should only be used if those checks can't run properly (e.g. because they'd dereference a nullptr). Using CHECK in other cases will print more information to help diagnose test failures.

```suggestion
  REQUIRE(from_handler.has_value());
  CHECK(from_handler.value().Signal == SIGTERM);
  CHECK(((from_handler.value().eflags >> FlagLocation) & 1) == FlagValue);
```

Sonicadvance1: Resolved.

PR #3256, opened 2023-11-07
unittests/ASM: Adds unittest for implicit flag clobber for #3254

Sonicadvance1:
It looks like currently FEX has a bug around implicit flag clobbering with this pull request where an IR operation that implicitly clobbers flags isn't correctly saving the NZCV flags before doing the operation.

Adds a unit test that specifically captures this issue. RAX will be 1 or 0 depending on if the flags are clobbered incorrectly or not.


alyssarosenzweig:
Looks possible, I will dig into this! Thank you!

alyssarosenzweig (approved):
can confirm this materialized the bug! thank you!

PR #3257, opened 2023-11-07
Add helper for deriving ops by opcode

alyssarosenzweig:
Needed to avoid breaking the world w.r.t nzcv

Not the most idiomatic C++ sol'n in the world but gets the job done.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
@@ -1329,6 +1329,16 @@ private:
     }
   }
 
+  // Helper to derive Dest by a given builder-using Expression with the opcode
+  // replaced with NewOp. Useful for generic building code. Not safe in general.
+  // but does the right handling of ImplicitFlagClobber at least and must be
+  // used instead of raw Op mutation.
+#define DeriveOp(Dest, NewOp, Expr) \
+  if (ImplicitFlagClobber(NewOp))   \
+    SaveNZCV();                     \
+  auto Dest = (Expr);               \
+  Dest.first->Header.Op = (NewOp)
+
   // Named constant cache for the current block.
   // Different arrays for sizes 1,2,4,8,16,32.
   OrderedNode *CachedNamedVectorConstants[FEXCore::IR::NamedVectorConstant::NAMED_VECTOR_MAX][6]{};
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -404,9 +404,7 @@ void OpDispatchBuilder::SecondaryALUOp(OpcodeArgs) {
   }
   else {
     Dest = LoadSource(GPRClass, Op, Op->Dest, Op->Flags, {.AllowUpperGarbage = AllowUpperGarbage || Size >= 4});
-    auto ALUOp = _Add(IR::SizeToOpSize(std::max<uint8_t>(4u, Size)), Dest, Src);
-    // Overwrite our IR's op type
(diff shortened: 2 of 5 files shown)


neobrain:
I'm guessing the issue here is that `ImplicitFlagClobber` needs to run before the `_VAdd` is called?

A more idiomatic C++ version that doesn't have macro hygiene issues would probably look like this:
```cpp
auto WithSavedNZCV(IROps NewOp) {
  if (ImplicitFlagClobber(NewOp))
    SaveNZCV();
  }
  return [NewOp](auto Value) {
    Value.first->Header.Op = NewOp;
  };
};

// Usage
auto ALUOp = WithSavedNZCV(IROp)(_VAdd(Size, ElementSize, Src1, Src2)));
```

More conventional syntax could be achieved by using a proper `struct` instead of a lambda.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1329,6 +1329,12 @@ friend class FEXCore::IR::PassManager;
     }
   }
 
+#define DeriveOp(Dest, NewOp, Expr) \
+  if (ImplicitFlagClobber(NewOp)) \
+    SaveNZCV(); \
+  auto Dest = Expr; \
+  Dest.first->Header.Op = NewOp
```
neobrain, line 1336: Macro hygiene to preempt accidental misuse:

```suggestion
#define DeriveOp(Dest, NewOp, Expr) \
  if (ImplicitFlagClobber((NewOp))) \
    SaveNZCV(); \
  auto Dest = (Expr); \
  Dest.first->Header.Op = (NewOp)
```

alyssarosenzweig:
> I'm guessing the issue here is that ImplicitFlagClobber needs to run before the _VAdd is called?

Correct.

> More conventional syntax could be achieved by using a proper struct instead of a lambda.

I'll let you & Ryan fight it out and then I'll implement whatever is agreed upon, I don't care as long as I don't need to rewrite this several times

Sonicadvance1:
> > More conventional syntax could be achieved by using a proper struct instead of a lambda.
> 
> I'll let you & Ryan fight it out and then I'll implement whatever is agreed upon, I don't care as long as I don't need to rewrite this several times 😇

I don't care which route so I'll leave it up to people suggesting changes. Once the idiom is in the code it'll read the same to me regardless.

neobrain:
I'd personally go with the proposed version for my own work, particularly since it highlights the importance of evaluation order somewhat better.

I mainly posted for informative reasons though; feel free to pick whichever you think is best readable for anyone who might work on the code at some point.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
alyssarosenzweig: fixed, thx

alyssarosenzweig:
I've corrected the macro hygiene issue and added a comment. The closureful version generates noticeably worse code (confirmed with objdump) at least on my local build and TBH it's not being more idiomatic C++ is not a compelling argument to me, especially given our JIT pressure. So I'd prefer to move ahead with the macro version at this point.

Sonicadvance1 (approved):

neobrain:
Minor qualm here: The commit messages don't actually give away you've changed behavior, though it's unclear to me anything actually changes in main or whether this is only a precaution for later changes.

Anyway, no objections to merging.

alyssarosenzweig:
> Minor qualm here: The commit messages don't actually give away you've changed behavior, though it's unclear to me anything actually changes in main or whether this is only a precaution for later changes.
> 
> Anyway, I've only had formal stuff to add here. No objections to merging.

behaviour change will only come with patches from #3162 , nothing changes here

PR #3261, opened 2023-11-08
FEX: Only pass CPU tunables to FEXCore and FEXLoader

Sonicadvance1:
This fixes an issue where CPU tunables were ending up in the thunk generator which means if your CPU doesn't support all the features on the *Builder* then it would crash with SIGILL. This was happening with Canonical's runners because they typically only support ARMv8.2 but we are compiling packages to run on ARMv8.4 devices.

cc: FEX-2311.1


neobrain (approved):
Lots of complexity here to enable `march` by default but then opting-out of all the special cases... has `-march=` been measured to provide a significant benefit that's worth this?

Code change itself is fine, though dropping all of the related code and leaving `march` options in the control of the user seems like a more robust and maintainable approach.

Sonicadvance1:
> Lots of complexity here to enable `march` by default but then opting-out of all the special cases... has `-march=` been measured to provide a significant benefit that's worth this?
> 
> Code change itself is fine, though dropping all of the related code and leaving `march` options in the control of the user seems like a more robust and maintainable approach.

Yes. Especially going from ARMv8.0 to ARMv8.1/v8.2

alyssarosenzweig (approved):

PR #3263, opened 2023-11-09
OpcodeDispatcher: Make "not" not garbage 

alyssarosenzweig:


Sonicadvance1 (approved):
But what about making it even notter garbage by doing eor directly in to the dest? 🤔

Sonicadvance1 (approved):
Noice.

PR #3264, opened 2023-11-10
Optimize selects

alyssarosenzweig:


FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -848,26 +848,58 @@ OrderedNode *OpDispatchBuilder::SelectBit(OrderedNode *Cmp, bool TrueIsNonzero,
                  TrueValue, FalseValue);
 }
 
-OrderedNode *OpDispatchBuilder::SelectCC(uint8_t OP, IR::OpSize ResultSize, OrderedNode *TrueValue, OrderedNode *FalseValue) {
+CondClassType OpDispatchBuilder::DecodeNZCVCondition(uint8_t OP, bool &Complex) {
```
lioncash, line 851: Maybe either use a `std::pair<CondClassType, bool>` for a return type instead of an out param and initialize stuff like:

```cpp
auto [Cond, IsComplex] = DecodeNZCVCondition(CoolValue);
```

or keep the out parameter, but set it to false at the beginning of the function, so that potential uninitialized reads can't happen (through refactors or accidentally in future changes)

e.g. it would prevent:

```cpp
bool Complex;
CondClassType Cond = DecodeNZCVCondition(CoolValue, Complex);
if (!Complex) { // whoops, accidental uninitialized read
  // ...
}
```

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1477,6 +1477,7 @@ friend class FEXCore::IR::PassManager;
     CachedIndexedNamedVectorConstants.clear();
   }
 
+  CondClassType DecodeNZCVCondition(uint8_t OP, bool &Complex);
```
lioncash, line 1480:
```suggestion
  CondClassType DecodeNZCVCondition(uint8_t OP, bool &Complex) const;
```

lioncash (commented):
Just two minor nits

alyssarosenzweig:
gh is being silly, moved to #3266

PR #3265, opened 2023-11-10
GdbServer: Switch over to a unix domain socket

Sonicadvance1:
Resolving the issue that we can only ever have one gdbserver process running consuming port 8086 (even if the port number is cute).

Doesn't give us anything yet but in the future will allow us to have whole process trees running gdbservers that we can attach to.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
@@ -12,6 +12,7 @@ $end_info$
 #include <memory>
 #include <optional>
 
+#include <Common/FEXServerClient.h>
 #include <FEXCore/Config/Config.h>
 #include <FEXCore/Core/CodeLoader.h>
 #include <FEXCore/Core/Context.h>
@@ -32,6 +33,7 @@ $end_info$
 #include <FEXCore/fextl/sstream.h>
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
+#include <FEXHeaderUtils/Filesystem.h>
 
 #include <atomic>
 #include <cstring>
@@ -47,6 +49,7 @@ $end_info$
 #include <stddef.h>
 #include <string_view>
 #include <sys/stat.h>
+#include <sys/un.h>
 #include <unistd.h>
 #include <utility>
 
@@ -1275,6 +1278,7 @@ void GdbServer::GdbServerLoop() {
   }
 
   close(ListenSocket);
+  unlink(GdbUnixSocketPath.c_str());
 }
 static void* ThreadHandler(void *Arg) {
   auto This = reinterpret_cast<FEX::GdbServer*>(Arg);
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1289,46 +1293,45 @@ void GdbServer::StartThread() {
...
-
-  memset(&hints, 0, sizeof(hints));
-  hints.ai_family = AF_UNSPEC;
-  hints.ai_socktype = SOCK_STREAM;
-  hints.ai_flags = AI_PASSIVE;
-
-  if(getaddrinfo(NULL, "8086", &hints, &res) < 0) {
-    perror("getaddrinfo");
+  const auto GdbUnixPath = fextl::fmt::format("{}/FEX_gdbserver/", FEXServerClient::GetServerMountFolder());
+  if (FHU::Filesystem::CreateDirectory(GdbUnixPath) == FHU::Filesystem::CreateDirectoryResult::ERROR) {
```
neobrain, line 1297: This will fail if any of the parent directories doesn't exist. Would `CreateDirectories` be more appropriate to use here?

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1289,46 +1293,45 @@ void GdbServer::StartThread() {
...
+    LogMan::Msg::EFmt("[GdbServer] Couldn't open AF_UNIX socket {} {}", errno, strerror(errno));
+    return;
   }
 
-  if (bind(ListenSocket, res->ai_addr, res->ai_addrlen) < 0) {
-    perror("bind");
+  struct sockaddr_un addr{};
+  addr.sun_family = AF_UNIX;
+  strncpy(addr.sun_path, GdbUnixSocket.data(), sizeof(addr.sun_path));
+  // Include final null character.
```
neobrain, line 1313: How is the final null character included here? Looks like it isn't, or the comment is unclear.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1289,46 +1293,45 @@ void GdbServer::StartThread() {
...
+  if (ListenSocket == -1) {
+    LogMan::Msg::EFmt("[GdbServer] Couldn't open AF_UNIX socket {} {}", errno, strerror(errno));
+    return;
   }
 
-  if (bind(ListenSocket, res->ai_addr, res->ai_addrlen) < 0) {
-    perror("bind");
+  struct sockaddr_un addr{};
+  addr.sun_family = AF_UNIX;
+  strncpy(addr.sun_path, GdbUnixSocket.data(), sizeof(addr.sun_path));
```
neobrain, line 1312: Edge case: `add.sun_path` won't be null-terminated if if the path length equals the byte count. We'll want to use `sizeof(...)-1` if this should always be null-terminated.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1289,46 +1293,45 @@ void GdbServer::StartThread() {
...
+  // Include final null character.
+  size_t SizeOfAddr = sizeof(addr.sun_family) + GdbUnixSocket.size();
```
neobrain, line 1314: Seems like this would be clearer:
```suggestion
  size_t SizeOfAddr = offsetof(sockaddr_un,sun_addr) + GdbUnixSocket.size();
```

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.h
```diff
@@ -93,6 +93,7 @@ class GdbServer {
     std::array<bool, FEXCore::SignalDelegator::MAX_SIGNALS + 1> PassSignals{};
     uint32_t CurrentDebuggingThread{};
     int ListenSocket{};
+    fextl::string GdbUnixSocket{};
```
neobrain, line 96: Subjectively unix sockets seem like a rare thing to be used. I typically think of bsd sockets first when reading the term, which makes it surprising to see it `string`-typed. Calling this variable `GdbUnixSocketPath` would make it a bit less surprising.

neobrain (approved):
Good change, finally able to debug within multiprocess sessions.

Anyway, just added a few minor points on this one that should be easy to sort out.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: It wouldn't be more appropriate. The folder that this ends up in is either `/tmp/` or `/run/user/<uid>/` and we don't want to try and create any of those parent directories.

If the user used the XDG directories to override the path, then they also need to make sure those folders exist otherwise regular applications break.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: oop, old comment. Null-termination doesn't matter here. bind passes a size which the kernel uses instead.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: Old comment confused this, null-termination doesn't matter here.

PR #3268, opened 2023-11-11
More GdbServer improvements

Sonicadvance1:
Nothing too crazy in here, trying to avoid some of the bigger changes.
A couple of stand-out things

- Adds some lldb specific query commands.
- Set the thread name of gdbserver. Easier to pick out if a thread hung.
- Try binding twice, unlinking the unix domain socket path in between
- Poll the listen socket instead of waiting in accept. Removing a hang that could occur.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
@@ -1333,7 +1333,7 @@ void GdbServer::StartThread() {
 }
 
 void GdbServer::OpenListenSocket() {
-  const auto GdbUnixPath = fextl::fmt::format("{}/FEX_gdbserver/", FEXServerClient::GetServerMountFolder());
+  const auto GdbUnixPath = fextl::fmt::format("{}/FEX_gdbserver/", FEXServerClient::GetTempFolder());
   if (FHU::Filesystem::CreateDirectory(GdbUnixPath) == FHU::Filesystem::CreateDirectoryResult::ERROR) {
     LogMan::Msg::EFmt("[GdbServer] Couldn't create gdbserver folder {}", GdbUnixPath);
     return;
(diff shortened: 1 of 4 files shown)


FEXCore/include/FEXCore/Utils/StringUtils.h
```diff
@@ -28,4 +28,41 @@ namespace FEXCore::StringUtils {
...
 
+  namespace StringHasher {
+    struct Hasher {
+      // Jenkins one-at-a-time function.
+      // Based off of https://en.wikipedia.org/wiki/Jenkins_hash_function
+      // citing: https://www.drdobbs.com/database/algorithm-alley/184410284
+      // and citing: https://www.burtleburtle.net/bob/hash/doobs.html
+      //
+      // Good enough constexpr hasher.
+      uint32_t constexpr operator()(const char* str) const {
```
lioncash, line 39:
```suggestion
      size_t constexpr operator()(const char* str) const noexcept {
```

would make this match the interface requirements of std::hash

FEXCore/include/FEXCore/Utils/StringUtils.h
```diff
@@ -28,4 +28,41 @@ namespace FEXCore::StringUtils {
...
+          ++str;
+        }
+        hash += hash << 3;
+        hash ^= hash >> 11;
+        hash += hash << 15;
+        return hash;
+      };
+
+      template<typename T>
+      uint32_t constexpr operator()(const T& str) const {
```
lioncash, line 54: Ditto

FEXCore/include/FEXCore/Utils/StringUtils.h
```diff
@@ -28,4 +28,41 @@ namespace FEXCore::StringUtils {
     return RightTrim(LeftTrim(std::move(String), TrimTokens), TrimTokens);
   }
 
+  namespace StringHasher {
```
neobrain, line 31: This doesn't seem to be related to the other changes. Did you mean to include any other patches?

If not, reviewing this along with code that actually uses it seems more sensible.

Sonicadvance1: This is used in GdbServer.cpp

neobrain: Ah, yes... "Large diffs are not rendered by default" got me once again :(

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -661,84 +671,84 @@ GdbServer::HandledPacketType GdbServer::handleXfer(const fextl::string &packet)
...
-        // Thread id is in hex without 0x prefix
-        const auto ThreadName = getThreadName(Thread->ThreadManager.GetTID());
-        ss << "<thread id=\"" << std::hex << Thread->ThreadManager.GetTID() << "\"";
-        if (!ThreadName.empty()) {
-          ss << " name=\"" << ThreadName << "\"";
-        }
-        ss << "/>\n";
+  using namespace FEXCore::StringUtils;
+  using namespace FEXCore::StringUtils::StringHasher;
+  switch (StringHasher::Hash(object)) {
```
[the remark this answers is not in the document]
Sonicadvance1: I can understand not wanting the idiom. I'll remove it.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1312,13 +1416,21 @@ void GdbServer::OpenListenSocket() {
...
+  };
+
   // Bind the socket to the path
-  int Result = bind(ListenSocket, reinterpret_cast<struct sockaddr*>(&addr), SizeOfAddr);
-  if (Result == -1) {
-    LogMan::Msg::EFmt("[GdbServer] Couldn't bind AF_UNIX socket '{}': {} {}\n", addr.sun_path, errno, strerror(errno));
-    close(ListenSocket);
-    ListenSocket = -1;
-    return;
+  if (!TryBind(ListenSocket, &addr, SizeOfAddr)) {
```
neobrain, line 1425: Using a lambda here for just one repetition adds more noise than it really saves. Maybe a for loop does the trick more easily?

```cpp
for (int attempt = 0; attempt < 2; ++attempt) {
    ...
}
```

neobrain (approved):
Some minor comments, looks good otherwise 👍

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1312,13 +1424,21 @@ void GdbServer::OpenListenSocket() {
...
+    // This can happen periodically with execve. unlink the path and try again.
```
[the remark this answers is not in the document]
Sonicadvance1: Added a comment that the PID is reused on execve and that FEX likely started a gdbserver before execve happened.

PR #3269, opened 2023-11-12
Optimize small negative 32-bit immediates

alyssarosenzweig:


alyssarosenzweig:
separate PR for visiblity, but included with #3264 for merge since other instcountci conflicts hard.

PR #3271, opened 2023-11-13
OpcodeDispatcher: Optimize three sha instructions

Sonicadvance1:
- sha1nexte
   - Takes advantage of sha1h if supported
   - Does the operation in a vector otherwise
- sha1msg2
   - Instead of dumping everything to GPRs, we can do this with vectors
   - Mostly matches ARM's sha1su1 instruction, but it is /just/
     different enough to be annoying.
- sha256msg1
   - Directly matches sha256u0
   - Leaves the previous implementation alone

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ASIMDOps.inl
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ASIMDOps.inl
@@ -60,7 +60,7 @@ public:
   }
   void sha256su1(FEXCore::ARMEmitter::VRegister rd, FEXCore::ARMEmitter::VRegister rn, FEXCore::ARMEmitter::VRegister rm) {
     constexpr uint32_t Op = 0b0101'1110'0000'0000'0000'00 << 10;
-    Crypto3RegSHA(Op, 0b100, rd, rn, rm);
+    Crypto3RegSHA(Op, 0b110, rd, rn, rm);
   }
 
   // Cryptographic two-register SHA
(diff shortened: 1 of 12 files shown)


FEXCore/Source/Interface/Core/ArchHelpers/CodeEmitter/ASIMDOps.inl
```diff
@@ -60,7 +60,7 @@ public:
   }
   void sha256su1(FEXCore::ARMEmitter::VRegister rd, FEXCore::ARMEmitter::VRegister rn, FEXCore::ARMEmitter::VRegister rm) {
     constexpr uint32_t Op = 0b0101'1110'0000'0000'0000'00 << 10;
-    Crypto3RegSHA(Op, 0b100, rd, rn, rm);
+    Crypto3RegSHA(Op, 0b110, rd, rn, rm);
```
alyssarosenzweig, line 63: unit test?

unittests/InstructionCountCI/Crypto/H0F38.json
```diff
@@ -0,0 +1,150 @@
...
+  "Instructions": {
+    "sha1nexte xmm0, xmm1": {
+      "ExpectedInstructionCount": 6,
+      "Optimal": "No",
+      "Comment": [
+        "0x66 0x0f 0x38 0xc8"
+      ],
+      "ExpectedArm64ASM": [
+        "dup v2.4s, v16.s[3]",
+        "unimplemented (Unimplemented)",
```
alyssarosenzweig, line 22: wth?

alyssarosenzweig (commented):
looks reasonable but @lioncash is prolly a better reviewer for this

unittests/InstructionCountCI/Crypto/H0F38.json
Sonicadvance1: Comes from Vixl not supporting crypto instructions. It's a pain yes.

alyssarosenzweig (approved):

PR #3273, opened 2023-11-13
Optimize shifts/rotates

alyssarosenzweig:


alyssarosenzweig:
alright, I'm bored with this now, good enough & time to move onwards. might come back to some of the really obscure stuff (8-bit rcl, etc) later. Ready for review I think :smile_cat:

alyssarosenzweig:
whoops added bonus optimizations

Sonicadvance1 (approved):
Interesting that some instructions regressed due to bad RA selection on BFIs. Oh well.

PR #3274, opened 2023-11-14
Arm64Emitter: Dedicate registers for PF/AF

alyssarosenzweig:
Many flag-generating instructions like cmp need to save calculations for
    deferred PF and AF flag calculation. Currently, they require a store per flag,
    which is prohibitively expensive for hot instructions like cmp. By instead
    pinning PF/AF temporary results to registers (x26/x27 by convention here), we
    eliminate many stores altogether and turn the rest into zero-cycle moves (on
    64-bit at least, this isn't optimal for 32-bit emulation due to CTX->GetGPRSize
    shenanigans, need to check if this requirement can be lifted..).
    
    To implement, we model as SRA and then the existing SRA code is able to generate
    good code with little manual tuning. (Future work will get us to excellent code
    with more tuning ;) ).
    
    The tradeoff is reducing the working dynamic GPR set by 2 registers, which might
    increase spilling in some cases. I think it's worth it in practice, though.


alyssarosenzweig:
now with fcmp optimizations for moar fps

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -21,38 +22,42 @@
 #include <tuple>
 #include <utility>
 
+// We pin r26/r27 as PF/AF respectively, this is internal FEX ABI.
+#define REG_PF 26
```
lioncash, line 26: Using:

```cpp
constexpr auto REG_PF = ARMEmitter::Reg::r26;
constexpr auto REG_AF = ARMEmitter::Reg::r27;
```

Will allow condensing stuff like 

```cpp
str(FEXCore::ARMEmitter::Register(REG_PF).X(), STATE.R(),
    offsetof(FEXCore::Core::CpuStateFrame, State.pf_raw));
```
into

```cpp
str(REG_PF.X(), STATE.R(), offsetof(FEXCore::Core::CpuStateFrame, State.pf_raw));
```

and for cases where you need the index, you can just use its `.Idx()` member function. Also keeps it consistent with the rest of the emitter.

Sonicadvance1 (approved):
Nice nice nice nice

PR #3279, opened 2023-11-19
Scripts: Removes usage of deprecated pkg_resources

Sonicadvance1:
Fairly trivial to convert to packaging.version


Sonicadvance1:
Gah, this isn't provided by default. I'll leave this alone for now.

PR #3283, opened 2023-11-23
FEXCore: Start changing how thread creation works

Sonicadvance1:
First commit from #3282.

The frontend needs to be in control of how threads are created. This is inherent to the fact that OS threads are OS specific. We currently have this weird split that when initializing the FEXCore context, we create a parent thread at all times.

This does some initial cleanup that gets the core initialization nearly decoupled.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -387,18 +387,11 @@ namespace FEXCore::Context {
       StartPaused = true;
     }
 
-    using namespace FEXCore::Core;
-
-    FEXCore::Core::InternalThreadState *Thread = CreateThread(nullptr, 0);
+    FEXCore::Core::InternalThreadState *Thread = CreateThread(InitialRIP, StackPointer, nullptr, 0);
 
     // We are the parent thread
     ParentThread = Thread;
 
-    Thread->CurrentFrame->State.gregs[X86State::REG_RSP] = StackPointer;
-
-    Thread->CurrentFrame->State.rip = InitialRIP;
-
-    InitializeThreadData(Thread);
     return Thread;
   }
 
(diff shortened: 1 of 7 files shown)


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -680,20 +669,23 @@ namespace FEXCore::Context {
...
+    Thread->CurrentFrame->State.gregs[X86State::REG_RSP] = StackPointer;
+    Thread->CurrentFrame->State.rip = InitialRIP;
+
     // Copy over the new thread state to the new object
     if (NewThreadState) {
-      memcpy(Thread->CurrentFrame, NewThreadState, sizeof(FEXCore::Core::CPUState));
+      memcpy(&Thread->CurrentFrame->State, NewThreadState, sizeof(FEXCore::Core::CPUState));
+
+      // Set up the thread manager state
+      Thread->ThreadManager.parent_tid = ParentTID;
```
neobrain, line 683: Why does this need to move into the `if (NewThreadState)`?

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -724,6 +716,13 @@ namespace FEXCore::Context {
       Thread->ExecutionThread->detach();
     }
 
+    if (Thread->DestroyedByParent) {
+#ifndef _WIN32
+      Alloc::OSAllocator::UninstallTLSData(Thread);
+#endif
+      SignalDelegation->UninstallTLSState(Thread);
```
neobrain, line 723: Why does this need to be done now when it seemingly wasn't there before?

Sonicadvance1: Oops, this leaked in from the other changes. This isn't yet required for this changed. Removed for now.

FEXCore/Source/Interface/Core/Core.cpp
Sonicadvance1: This doesn't. but it also doesn't change behaviour. Moved it back.

neobrain (approved):

PR #3284, opened 2023-11-23
FEXCore: Changes ParentThread ownership from the CTX to the frontend 

Sonicadvance1:
Second commit from #3282.

This is the minimal amount of changes to move the ownership from FEXCore
to the frontend. Since the frontends don't yet have a full thread state
tracking, there is an opaque pointer that needs to be managed.

In the followup commits this will be changed to have the syscall handler
to be the thread object manager.


Sonicadvance1:
Closing in favour of #3321

PR #3288, opened 2023-11-24
FEXCore: Fixes imul returning garbage data

Sonicadvance1:
When a 32-bit imul was being executed it had a chance of returning garbage data in the upper 32-bits of the 64-bit result. While this didn't typically cause problems, this gets exacerbated from 32-bit applications executing multiplies for address calculations.

A combination of commits 714669136086cee0d2cc4dfb479e26b204206c37 and d01b457727208fd34511d48e850e3b4a33d76147 exposed this problem where previously there would be multiple moves between the calculation and data use which would have zero'd the upper bits for us previously.

Now that we are no longer doing that, we need to make sure the opcode dispatcher doesn't generate broken code instead.

Fixes Dungeon Defenders, which hasn't worked since FEX-2308.

Adds an ASM test that ensures we don't break it again.


lioncash (approved):
oof, gnarly bug

PR #3289, opened 2023-11-25
FEXCore: Moves debug strings to gdbserver

Sonicadvance1:
These are only used by gdbserver for filling out its XML data structures so just remove them from FEXCore.

Also fixes the ordering on RegNames to match the definition of the enum class definition in CoreState. This has been out of correct order since we reordered registers months ago.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
@@ -58,6 +58,58 @@ $end_info$
 
 namespace FEX
 {
+
+constexpr std::array<std::string_view const, 22> FlagNames = {
+  "CF",
+  "",
+  "PF",
+  "",
+  "AF",
+  "",
+  "ZF",
+  "SF",
+  "TF",
+  "IF",
+  "DF",
+  "OF",
+  "IOPL",
+  "",
+  "NT",
+  "",
+  "RF",
+  "VM",
+  "AC",
+  "VIF",
+  "VIP",
+  "ID",
+};
+
+static std::string_view const& GetFlagName(unsigned Flag) {
+  return FlagNames[Flag];
+}
+
+static std::string_view const GetGRegName(unsigned Reg) {
+  switch (Reg) {
+    case FEXCore::X86State::REG_RAX: return "rax";
+    case FEXCore::X86State::REG_RBX: return "rbx";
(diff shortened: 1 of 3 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -58,6 +58,59 @@ desc: Provides a gdb interface to the guest state
...
+  "ID",
+};
+
+static std::string_view const& GetFlagName(unsigned Flag) {
+  return FlagNames[Flag];
+}
+
+constexpr std::array<std::string_view const, 16> RegNames = {
+  "rax",
+  "rcx",
```
neobrain, line 93: How is this ordered now? The PR description isn't specific enough for me to just find it, so sounds like something that needs a comment here.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -441,7 +494,7 @@ fextl::string buildTargetXML() {
 
       // GPRs
       for (uint32_t i = 0; i < FEXCore::Core::CPUState::NUM_GPRS; i++) {
-        reg(FEXCore::Core::GetGRegName(i), "int64", 64);
+        reg(GetGRegName(i), "int64", 64);
```
neobrain, line 497: Does GDB not expect a specific ordering here? Are we consistent with what other gdbserver implementations return?

Sonicadvance1: gdb doesn't expect a specific ordering but it will end up displaying the order we provide it when you do a `info regs all`

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: Changed over to a switch statement so that even if the enum ordering changes in the future this won't break again. Since it is highly likely to change in the future. This way the ordering matching doesn't matter anymore.

Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
neobrain: Fair enough, no need to worry about it here then.

neobrain (approved):

PR #3290, opened 2023-11-25
FEXCore: Moves more SignalDelegator functions to the frontend

Sonicadvance1:
As we are moving more and more OS specific code to the frontend, this is another set of functions that can be moved to FEXLoader from FEXCore.

No functional change here, only code moved from protected to private and to FEXLoader's SignalDelegator.

Once more thread handling is moved to the frontend we can move even more out of FEXCore. As follows:
- CheckXIDHandler can get moved.
  - First pthread FEX makes would just call this.
- Register/UnregisterTLSState
  - This can happen in the clone/thread handler once the frontend handles it.

This leaves very little in the backend and is mostly an interface for passing signal data to the frontend that it needs once a signal has occured.
It additionally also is used for `SignalThread`.


Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -150,6 +150,38 @@ namespace FEX::HLE {
     return SigInfoLayout::LAYOUT_KILL;
   }
 
+  void SignalDelegator::HandleSignal(int Signal, void *Info, void *UContext) {
+    // Let the host take first stab at handling the signal
+    auto Thread = GetTLSThread();
+    SignalHandler &Handler = HostHandlers[Signal];
```
neobrain, line 156: This shadows `Handler` below. Can it safely be moved down to the 2 lines of code instead? Otherwise, some minor renaming will prevent accidental mixups here.

neobrain (approved):

Source/Tools/FEXLoader/LinuxSyscalls/SignalDelegator.cpp
Sonicadvance1: Moved lower and renamed the temporary in the loop.

PR #3291, opened 2023-11-25
EmulatedFiles: Stop relying on O_TMPFILE

Sonicadvance1:
O_TMPFILE has a few minor problems that I have been thinking about for a while. I just recently got reminded about this and remembered that most problems get resolved by using memfd_create.

- O_TMPFILE is only supported on some filesystems.
   - Supported filesystem must be the one mounted to the pathname being opened.
   - Only a minor inconvenience as tmpfs and all related filesystems support this.
   - An inode is actually created on whatever filesystem is backing the folder.
- `/tmp/` must exist as a directory
   - If this folder happened to not be mounted then these temporary files wouldn't have been created.
   - memfd_create doesn't have a folder that needs to exist.
- We were leaving the files open as read/write
   - While we were rewinding the file offset, an misbehaving application could have wrote garbage to the temp file.
   - memfd sealing allows us to open the FD as RW and then seal its capabilities, making it a read-only FD.
- We were leaking FDs opened with O_CLOEXEC
   - We could have just opened the O_TMPFILE with O_CLOEXEC
   - memfd also just supports this flag, so use it.
   - No real issues, just nice to be sanitary here.

Overall this doesn't really change any behaviour, but it is nice to cleanup some of the edges there.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
@@ -46,9 +46,22 @@ namespace FEX::EmulatedFile {
    *
    * @return A temporary file that we can use
    */
-  static int GenTmpFD() {
-    int fd = open("/tmp", O_RDWR | O_TMPFILE | O_EXCL, S_IRUSR | S_IWUSR);
-    return fd;
+  static int GenTmpFD(const char *pathname, int flags) {
+    uint32_t memfd_flags {MFD_ALLOW_SEALING};
+    if (flags & O_CLOEXEC) memfd_flags |= MFD_CLOEXEC;
+
+    return memfd_create(pathname, memfd_flags);
+  }
+
+  // Seal the tmpfd features by sealing them all.
+  // Makes the tmpfd read-only.
+  static void SealTmpFD(int fd) {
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
```diff
@@ -46,9 +46,22 @@ namespace FEX::EmulatedFile {
...
+  static int GenTmpFD(const char *pathname, int flags) {
+    uint32_t memfd_flags {MFD_ALLOW_SEALING};
+    if (flags & O_CLOEXEC) memfd_flags |= MFD_CLOEXEC;
+
+    return memfd_create(pathname, memfd_flags);
+  }
+
+  static void SealTmpFD(int fd) {
+    // Seal the tmpfd features by sealing them all.
+    // Makes the tmpfd read-only.
```
neobrain, line 58: Function documentation should go above the function, since it's easy to assume it's just implementation details otherwise.

neobrain (approved):

Source/Tools/FEXLoader/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
Sonicadvance1: Moved.

PR #3296, opened 2023-11-29
FEXCore: Moves OS thread creation to the frontend 

Sonicadvance1:
~~Needs #3294 merged first.~~

Fairly lightweight since it is almost 1:1 transplanting the code from
FEXCore in to the SyscallHandler's thread creation code.

Minor changes:
- ExecutionThreadHandler gets freed before executing the thread
   - Saves 16-bytes of memory per thread
- Start all threads paused by default
   - Since I moved the code to the frontend, I noticed we needed to do
     some post thread-creation setup.
   - Without the pause we were racing code execution with TLS setup and
     a few other things.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
@@ -40,6 +40,21 @@ $end_info$
 ARG_TO_STR(idtype_t, "%u")
 
 namespace FEX::HLE {
+
+  struct ExecutionThreadHandler {
+    FEXCore::Context::Context *CTX;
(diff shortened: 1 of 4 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -86,6 +89,27 @@ namespace FEX::HLE {
       x32::AdjustRipForNewThread(NewThread->CurrentFrame);
     }
 
+    // We need to do some post-thread creation setup.
+    NewThread->StartPaused = true;
+
+    // Initialize a new thread for execution,
```
neobrain, line 95: Pending comma

neobrain (approved):

PR #3301, opened 2023-11-29
InstCountCI: Moves Sonic Mania code to 32-bit file

Sonicadvance1:
Sonic Mania is a 32-bit game, it shouldn't be running its assembly as 64-bit.


alyssarosenzweig:
D'oh!

alyssarosenzweig:
Well, that would make #3300 pretty useless then...

alyssarosenzweig (approved):

PR #3302, opened 2023-11-29
Fold flag generation into the producers

alyssarosenzweig:
should be a small perf boost.


alyssarosenzweig:
With the distance I think backpatching is a bad idea. Closing this, hoping to revisit the problem again with a safer approach.

PR #3303, opened 2023-11-29
OpcodeDispatcher: Initial support for runtime long-mode switch 

Sonicadvance1:
This has the Frontend and OpcodeDispatcher select their operating mode
depending on the incoming code segment long-mode flag.

Adds some asserts since currently it is unexpected if the configuration
changes at runtime.

This is fairly straightforward for an initial setup but isn't fully
fleshed out.

Right now FEX's x86 tables aren't setup in a way to support choosing a
different instruction decoding depending on runtime operating mode
change, so that would break in interesting ways.

Primarily this just gets FEX setup to start piping the operating mode
through from the frontend to the backend. This is a long term task, so
it is going to take a long time to iron out all the issues.


alyssarosenzweig:
What uses a runtime long mode switch?

alyssarosenzweig:
The changes look good to me (I like making the dispatcher less boilerplatey), but I don't understand the background so I don't feel good about approving in this form. This PR is prep work -- What's the end goal?

alyssarosenzweig:
got it, didn't realize WINE wanted this. Seems reasonable then :+1:

alyssarosenzweig (approved):

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Thread.cpp
```diff
@@ -68,29 +68,29 @@ namespace FEX::HLE::x32 {
 
     // Now we need to update the thread's GDT to handle this change
     auto GDT = &Frame->State.gdt[u_info->entry_number];
-    GDT->base = u_info->base_addr;
+    Frame->State.SetGDTBase(GDT, u_info->base_addr);
```
ChengmingTangCN, line 71: Does the GDT limit need to be updated here? Otherwise, it will allocate an already allocated TLS entry later?

PR #3304, opened 2023-11-30
Dispatcher: Fixes corruption when spilling SRA registers

Sonicadvance1:
These functions only want the GPRs returned for SRA. This is because the signal handler needs this map to relation between x86 GPRs and AArch64 GPRs.

When we added AF and PF to the SRA array we accidentally started returning two more GPRs to the frontend. This caused the signal delegator to start corrupting the members after GPRs in FEX's CoreState.

Corrupting 16-bytes after the gregs[] array.
This included corrupting:
   - es_idx, cs_idx, ss_idx, ds_idx, gs_idx, fs_idx, _pad[]


alyssarosenzweig (approved):
oof

PR #3305, opened 2023-12-02
JIT: Fixes crash in TestNZ

Sonicadvance1:
In some situations TestNZ is generated with a constant that is using a
constant that can't fit inside of the tst instruction.

This was found in libGLX with virgl, crashing invalid instruction
generation and crashing steamwebhelper

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -122,7 +122,7 @@ DEF_OP(SbbNZCV) {
 
 DEF_OP(TestNZ) {
   auto Op = IROp->C<IR::IROp_TestNZ>();
-  const uint8_t OpSize = Op->Size;
+  const uint8_t OpSize = IROp->Size;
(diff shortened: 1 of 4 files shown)


alyssarosenzweig:
this doesn't seem right, testnz is morally the same as and. if we're inlining unencodable immediates, then presumbly https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp#L1014  is the place to fix that, no?

Sonicadvance1:
> this doesn't seem right, testnz is morally the same as and. if we're inlining unencodable immediates, then presumbly https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp#L1014 is the place to fix that, no?

You're right. That's where the bug was. It was using IROp->Size instead of Op->Size for that operation. Which means it was thinking the operating size was 0 because it wasn't ever setting the destination size, and since it doesn't return an SSA value, this wasn't an error.

Changed the IR to store the size in the IROp->Size member instead of its own internal op size. This now means the ConstProp isn't incorrectly creating an inline constant and fixes the problem.

PR #3309, opened 2023-12-04
Thunks: Add workarounds for pointers not readable by 32-bit guests

neobrain:
## Overview

32-bit guests cannot access memory stored in the upper 32-bit address space of memory, which is why we strongly control [how FEX-internal memory is allocated](https://github.com/FEX-Emu/FEX/blob/64276dbd0c3be786004f08dc2d40753f72d8e8b1/docs/ProgrammingConcerns.md). Despite all the effort we go through, two major problems still remain:

## Problem 1: Internal jemalloc reserves

Some allocations are made out of our control during FEX startup. The allocated data is never passed to the guest, but `jemalloc` will optimistically over-allocate and hold on to unused memory as a future reserve. This means that even after FEX blocks the upper memory range, future allocations may still use those jemalloc-internal reserves.

Ideally we'd patch up jemalloc to seal any such reserves properly.

As a transitional measure we spam `malloc(1)` to exhaust those reserves by brute force instead.

## Problem 2: Host stack pointer when forwarding APIs with callbacks

Many functions that we forward from x86 to the host library have callback parameters, i.e. function pointer arguments provided by the guest application. When the host calls into the guest-function through this callback parameter, the arguments for the guest functions are packed onto the host stack. Critically, the initial placement of the host stack is generally out of our control, so this data may not be in a location readable by the 32-bit guest.

As a transitional measure, we allocate a temporary memory page and relocate argument data to it whenever a host->guest callback is invoked. **This has a huge performance impact and makes forwarding of 32-bit libraries prohibitively slow.** It does, however, allow us to continue bringing up 32-bit support.

## Side problem: 32-bit pointers passed across architecture boundaries will be read as if they were 64-bit

For 32-bit guests, guest-side argument packers always write 32-bit pointers whereas the host-side unpackers always read 64-bits. This breaks execution of forwarding-related 32-bit FEXLinuxTests (see #3201). This issue will be resolved soon in a future change, but temporarily disabling affected test cases allows us to re-enable the test as a whole.


Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -438,6 +438,15 @@ int main(int argc, char **argv, char **const envp) {
...
     }
+
+    // Now that the upper 32-bit address space is blocked for future allocations,
+    // exhaust all of jemalloc's remaining internal allocations that it reserved before.
+    // TODO: It's unclear how reliably this exhausts those reserves
+    void* data;
+    do {
+      data = malloc(0x1);
+    } while (reinterpret_cast<uintptr_t>(data) >> 32 != 0);
+    free(data);
```
neobrain, line 449: Do we have a good way of enabling this only when thunking is enabled? A check for the presence of a thunk config file seems pointlessly generic, whereas a concrete check for enabled thunks would require peeking into ThunkDB internals AFAIK.

Not sure to what extent it matters overall, since the `malloc(1)` spam will generally hit jemalloc's fast path and hence return quickly.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -438,6 +438,15 @@ int main(int argc, char **argv, char **const envp) {
     else {
       Allocator = FEX::HLE::CreatePassthroughAllocator();
     }
+
+    // Now that the upper 32-bit address space is blocked for future allocations,
+    // exhaust all of jemalloc's remaining internal allocations that it reserved before.
+    // TODO: It's unclear how reliably this exhausts those reserves
+    void* data;
+    do {
+      data = malloc(0x1);
```
neobrain, line 447: I'm guessing this should be `FEXCore::Allocator::malloc`.

neobrain, line 447: This fails the glibc allocator CI checks. Is `FEXCore::Allocator::malloc` the right alternative to use here?

Sonicadvance1: The fault enabling would need to be temporarily disabled around this, since it is expected that FEX is abusing the glibc allocator functions here.

neobrain:
The glibc fault test runners kept failing because the infinite-looped on the added `malloc(1)` workaround. It looks like the FORCE32BITALLOCATOR path doesn't actually block the upper address space; is this by design or a bug?

For the time being, I disabled the `malloc(1)` workaround on that path.

FEXCore/Source/Interface/HLE/Thunks/Thunks.cpp
```diff
@@ -180,8 +180,25 @@ namespace FEXCore {
             Thread->CurrentFrame->State.gregs[FEXCore::X86State::REG_RDI] = (uintptr_t)arg0;
             Thread->CurrentFrame->State.gregs[FEXCore::X86State::REG_RSI] = (uintptr_t)arg1;
           } else {
+            // The argument pointer comes from the host stack, so it's not
+            // accessible by the guest. Allocate a thread-local page of memory
+            // to relocate the argument data.
+            // TODO: Directly store the arguments in a guest-accessible location instead
+            thread_local void* local_args = std::invoke([]() {
+              return mmap(
+                  0, 1000,
```
Sonicadvance1, line 189: comment says page but allocating only 1000 bytes through mmap instead of 0x1000.

neobrain: This remark got me reconsidering why I was using `mmap` here in the first place... Annoyingly enough, `FEXCore::Allocator::malloc` still returns something like `0xffffxxxxxxxx` here. mmap seemed to be the only reliable way to allocate guest-accessible memory.

I'll keep using mmap here, but changed over to the same amount copied below. (Also removed the unnecessary layer of immediately-invoked lambda.)

neobrain:
CI fixed and patch history cleaned up. Dropping draft status.

Sonicadvance1 (approved):

PR #3310, opened 2023-12-05
FEXLoader: Temporarily disable CLONE_CLEAR_SIGHAND

Sonicadvance1:
This flag breaks FEX heavily for now.
glibc 2.38 started using this flag as an optimization for posix_spawn. It will fall back to a "non-optimized" implementation if the clone syscall returns EINVAL. For now do this while we investigate a more proper implementation.

Should be backported to 2312.1.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
@@ -525,6 +525,14 @@ static uint64_t Clone3Handler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clo
 uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args *args) {
   uint64_t flags = args->args.flags;
 
+  if (flags & CLONE_CLEAR_SIGHAND) {
+    // CLONE_CLEAR_SIGHAND was added in kernel 5.5. FEX doesn't properly support this.
+    // glibc started using this flag in 2.38 as an optimization for posix_spawn.
+    // If clone returns EINVAL or ENOSYS then it will fallback to the non-optimized path.
+    LogMan::Msg::IFmt("CLONE_CLEAR_SIGHAND passed to clone3. Returning EINVAL.");
+    return -EINVAL;
+  }
+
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
```diff
@@ -525,6 +525,14 @@ static uint64_t Clone3Handler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clo
 uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args *args) {
   uint64_t flags = args->args.flags;
 
+  if (flags & CLONE_CLEAR_SIGHAND) {
+    // CLONE_CLEAR_SIGHAND was added in kernel 5.5. FEX doesn't properly support this.
+    // glibc started using this flag in 2.38 as an optimization for posix_spawn.
+    // If clone returns EINVAL or ENOSYS then it will fallback to the non-optimized path.
+    // Temporarily disable clone3 if this flag is used.
```
neobrain, line 532: > Temporarily disable clone3 if this flag is used.

What does this mean? We don't seem to temporarily disable anything here.

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
```diff
@@ -525,6 +525,14 @@ static uint64_t Clone3Handler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clo
...
+    return -EINVAL;
```
neobrain, line 533: Might be worth printing a warning log here?

neobrain (approved):

Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
neobrain: That's *permanently* though as far as behavior goes.

I'm guessing what you meant with temporary is either that we hope to drop this workaround eventually OR the fact that glibc will retry after seeing the EINVAL return value.

The rest of the comment is clearly written, but the last line adds ambiguity. Just removing it seems better to me, since it's reasonably obvious that we shouldn't normally return EINVAL but do so here to make use of glibc's fallback path. Alternatively, a TODO could stress that this is just a workaround.

PR #3315, opened 2023-12-06
Linux: Adjust when clone allocates stack memory

Sonicadvance1:
clone needs to allocate memory before the fork locks are held, otherwise they will hang forever waiting on a locked mutex.

This wasn't previously seen since nothing was using fork with clone3.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
@@ -385,7 +385,6 @@ struct StackFrameData {
   FEXCore::Context::Context *CTX{};
   FEXCore::Core::CpuStateFrame NewFrame{};
   FEX::HLE::clone3_args GuestArgs{};
-  void *NewStack;
   size_t StackSize;
 };
 
@@ -399,7 +398,7 @@ struct StackFramePlusRet {
 static void Clone3HandlerRet() {
   StackFrameData *Data = (StackFrameData*)alloca(0);
   uint64_t Result = FEX::HLE::HandleNewClone(Data->Thread, Data->CTX, &Data->NewFrame, &Data->GuestArgs);
-  FEX::LinuxEmulation::Threads::DeallocateStackObject(Data->NewStack);
+  FEX::LinuxEmulation::Threads::DeallocateStackObject(Data->GuestArgs.NewStack);
   // To behave like a real clone, we now just need to call exit here
   exit(Result);
   FEX_UNREACHABLE;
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/Syscalls.cpp
```diff
@@ -597,6 +585,12 @@ uint64_t CloneHandler(FEXCore::Core::CpuStateFrame *Frame, FEX::HLE::clone3_args
 
       args->SignalMask = ~0ULL;
       ::syscall(SYS_rt_sigprocmask, SIG_SETMASK, &args->SignalMask, &args->SignalMask, sizeof(args->SignalMask));
+
+      // In the case of thread, we need a new stack
+      // Needs to be allocated before the locks occur.
```
neobrain, line 590: Why does it hang otherwise?

Sonicadvance1: Context fork locking locks the allocator mutex, so no allocations can occur inside of that.

neobrain: I'm having trouble finding which locks are interfering here. I only looked into LockBeforeFork, which seems to lock a different mutex than AllocateStackObject.

Sonicadvance1: Lock Side:
- Thread->CTX->LockBeforeFork(Frame->Thread);
- Allocator::LockBeforeFork(Thread);
- Alloc64->LockBeforeFork(Thread);
- AllocationMutex.lock();

Stack Allocation side:
- FEX::LinuxEmulation::Threads::AllocateStackObject();
- FEXCore::Allocator::mmap(nullptr, FEX::LinuxEmulation::Threads::STACK_SIZE, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
- FEX_DEFAULT_VISIBILITY extern MMAP_Hook mmap;
- FEXCore::Allocator::mmap = FEX_mmap;
- void *Result = Alloc64->Mmap(addr, length, prot, flags, fd, offset);
- auto lk = FEXCore::GuardSignalDeferringSectionWithFallback(AllocationMutex, TLSThread);

neobrain: Ah, I was looking at the wrong LockBeforeFork before.

This should be a bit clearer:
```suggestion
      // LockBeforeFork grabs the allocator mutex to block allocations temporarily, so this must be allocated before
```

neobrain (approved):

PR #3318, opened 2023-12-08
Library Forwarding: Emit layout helpers to allow repacking struct data

neobrain:
## Overview

This PR makes the thunk generator emit helper structures to allow repacking `struct`s that have different data layout on the guest than on the host. The new `guest_layout` and `host_layout` wrappers are emitted for each type used in APIs of forwarded libraries. These structures have the same members as the original structure but they are defined in an architecture-agnostic way. This allows the host to read guest structures without having to worry about member offsets, type size differences, or other sorts of differences in the data layout.

There is no *automatic* struct repacking yet, but the wrappers allow for *manual* repacking by converting a `guest_layout` is converted to a `host_layout`, or vice-versa.

See #3156 for the data layout analysis framework that gathers the required metadata.

## Implementation: Layout wrappers

The layout wrappers are only generated for types that don't have consistent data layout between guest and host **and** if they considered "repackable" by the data analysis pass. For non-repackable types, layout wrappers must explicitly be enabled using the new `emit_layout_wrappers` annotation (*no* repacking code will be generated in that case).

`guest_layout` and `host_layout` are C++ templates specialized for each type. This allows a number of utility functions to be defined generically, so that the generator only needs to emit the bare minimum of type-specific information. `guest_layout` maps built-in types to themselves, and it recursively maps structs to structs of `guest_layout`s. For example:
```cpp
template<>
struct guest_layout<VkDisplayModeCreateInfoKHR> {
    struct type {
        guest_layout<VkStructureType>             sType;
        guest_layout<const void*>                 pNext;
        guest_layout<VkDisplayModeCreateFlagsKHR> flags;
        guest_layout<VkDisplayModeParametersKHR>  parameters;
    };
    type data;
};

template<>
struct guest_layout<VkDisplayModeParametersKHR> {
    struct type {
        guest_layout<VkExtent2D> visibleRegion;
        guest_layout<uint32_t>   refreshRate;
    };
    type data;
};

// ...
```

Since these structures are only ever used on the host, `host_layout` is a simple wrapper around the unchanged type *and* a constructor from `guest_layout`. This layer of indirection adds type-safety and makes sure any conversions are intentional:
```cpp
template<>
struct host_layout<VkDisplayModeCreateInfoKHR> {
  using type = VkDisplayModeCreateInfoKHR;
  type data;

  host_layout(const guest_layout<VkDisplayModeCreateInfoKHR>& from) :
    data { from.data } {
  }
};

// ...
```

## Implementation: Struct repacking

Repacking happens whenever a `guest_layout` is converted to a `host_layout`, or vice-versa. To avoid performance overhead or compatibility problems from "accidental" repacking, strong typing is used consistently for these layout wrappers.

Conversion between the unwrapped type and the wrappers is done using the `host_layout` constructor and two utility functions (`to_guest` (host->guest), `to_host_layout` (Type -> host_layout)). This diagram illustrates the different pathways of conversion:
```mermaid
flowchart TD
    Type -->|to_host_layout| A["host_layout&lt;Type&gt"]
    A["host_layout&lt;Type&gt;"] -->|to_guest| G["guest_layout&lt;Type&gt;"]
    G["guest_layout&lt;Type&gt;"] -->|"host_layout()"| A["host_layout&lt;Type&gt"]
    A["host_layout&lt;Type&gt"] -->|.data| Type
```

## Impact

The layout wrappers aren't automatically used anywhere yet, so this has no immediate impact on compatibility or performance.

## Future

The repacking machinery added here is the basis for automatic struct repacking, where parameters passed to forwarded APIs are automatically repacked as needed.

Furthermore, `guest_layout` does not yet take care of size differences of built-in types (`size_t`, ...). A future change will map these to fixed-size integers.

## TODO

- [ ] Assess potential future performance impact
- [x] Fix build problems with old libclang

### DIFF
--- a/ThunkLibs/include/common/Host.h
+++ b/ThunkLibs/include/common/Host.h
@@ -48,27 +48,6 @@ struct ExportEntry { uint8_t* sha256; void(*fn)(void *); };
 
 typedef void fex_call_callback_t(uintptr_t callback, void *arg0, void* arg1);
 
-/**
- * Opaque wrapper around a guest function pointer.
- *
- * This prevents accidental calls to foreign function pointers while still
- * allowing us to label function pointers as such.
- */
-struct fex_guest_function_ptr {
-private:
-    void* value = nullptr;
-
-public:
-    fex_guest_function_ptr() = default;
-
-    template<typename Ret, typename... Args>
-    fex_guest_function_ptr(Ret (*ptr)(Args...)) : value(reinterpret_cast<void*>(ptr)) {}
-
-    inline operator bool() const {
-      return value != nullptr;
-    }
-};
-
 #define EXPORTS(name) \
   extern "C" { \
     ExportEntry* fexthunks_exports_##name() { \
--- a/ThunkLibs/include/common/Host.h
+++ b/ThunkLibs/include/common/Host.h
@@ -97,8 +97,63 @@ struct guest_layout {
   static_assert(!std::is_enum_v<T>, "No guest layout defined for this enum type. This is a bug in the thunk generator.");
   static_assert(!std::is_void_v<T>, "Attempted to get guest layout of void. Missing annotation for void pointer?");
 
-  using type = T;
+  using type = std::enable_if_t<!std::is_pointer_v<T>, T>;
   type data;
+
+  guest_layout& operator=(const T from) {
+    data = from;
+    return *this;
+  }
+};
+
+template<typename T>
+struct guest_layout<T*> {
+#ifdef IS_32BIT_THUNK
+  using type = uint32_t;
+#else
+  using type = uint64_t;
+#endif
+  type data;
+
+  // Allow implicit conversion for function pointers, since they disallow use of host_layout
+  guest_layout& operator=(const T* from) requires (std::is_function_v<T>) {
+    // TODO: Assert upper 32 bits are zero
+    data = reinterpret_cast<uintptr_t>(from);
+    return *this;
+  }
+
+  guest_layout<T>* get_pointer() {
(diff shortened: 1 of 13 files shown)


ThunkLibs/include/common/Host.h
```diff
@@ -111,9 +92,155 @@ struct ParameterAnnotations {
...
+
+  // TODO: Make this conversion explicit
+  guest_layout& operator=(const T from) {
+    data = from;
+    return *this;
+  }
+
+  // Allow conversion of integral types of same size and sign to each other.
+  // This is useful for handling "long"/"long long" on 64-bit, as well as uint8_t/char.
+  // TODO: Make this conversion explicit
```
neobrain, line 111: Drop this one. These are constraint anyway, and a future change needs this implicit conversion behavior between "similar" types.

ThunkLibs/include/common/Host.h
```diff
@@ -210,6 +337,14 @@ struct GuestWrapperForHostFunction<Result(Args...)> {
   }
 };
 
+// TODO: Move?
```
neobrain, line 340: Double-check this. (Should just be moved to a different position in the same file)

ThunkLibs/include/common/Host.h
neobrain: Had an epiphany the other day on how to handle this more cleanly. For now, this extra constructor is fine though, but it will be replaced once we land full 32-bit support.

neobrain: Actually this won't be needed at all for a while, so I've dropped it from this PR.

Sonicadvance1 (approved):
I'm sure the todos will get resolved over time.

PR #3319, opened 2023-12-08
GdbServer: Fixes crash on gdb detach

Sonicadvance1:
GdbServer object was getting deleted before the socket thread was shutdown which was causing a crash on detach.

Now on destructor it will wait for the thread to thread to exit.

### DIFF
--- a/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
@@ -127,8 +127,12 @@ void GdbServer::WaitForThreadWakeup() {
 }
 
 GdbServer::~GdbServer() {
+  CloseListenSocket();
   CoreShuttingDown = true;
-  close(ListenSocket);
+
+  if (gdbServerThread->joinable()) {
+    gdbServerThread->join(nullptr);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXLoader/LinuxSyscalls/GdbServer.cpp
```diff
@@ -127,8 +127,12 @@ void GdbServer::WaitForThreadWakeup() {
 }
 
 GdbServer::~GdbServer() {
+  CloseListenSocket();
```
neobrain, line 130: Since OpenListenSocket is running at the start of the GDB server thread, is there potential for a race condition for short-lived FEX processes here?

Sonicadvance1: Not currently since gdbserver always forces processes to wait for gdb attach currently.

neobrain: Does it? I see `GDBServer::GDBServer()` ending in a call to StartThread, so the caller can immediately close it after. Is there external logic that does this waiting?

Sonicadvance1: > Is there external logic that does this waiting?
Yes.
https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/Core/Core.cpp#L331

neobrain: Fair enough, thanks for clarifying!

neobrain (approved):

PR #3320, opened 2023-12-08
X86Tables: Converts tables to be mostly consteval

Sonicadvance1:
Reduces the ELF's VM size from 9.8MB down to 9.37MB and should reduce initialization time a smidge.

Slammed this out while waiting for other PRs to get reviewed.

### DIFF
--- a/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
+++ b/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
@@ -507,26 +507,30 @@ using U8U8InfoStruct = X86TablesInfoStruct<uint8_t>;
 using U16U8InfoStruct = X86TablesInfoStruct<uint16_t>;
 
 template<typename OpcodeType>
-static inline void GenerateTable(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize) {
+constexpr static inline void GenerateTable(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize) {
   for (size_t j = 0; j < TableSize; ++j) {
     X86TablesInfoStruct<OpcodeType> const &Op = LocalTable[j];
     auto OpNum = Op.first;
     X86InstInfo const &Info = Op.Info;
     for (uint32_t i = 0; i < Op.second; ++i) {
-      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+      if (FinalTable[OpNum + i].Type != TYPE_UNKNOWN) {
+        ERROR_AND_DIE_FMT("Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+      }
       FinalTable[OpNum + i] = Info;
     }
   }
 };
 
 template<typename OpcodeType>
-static inline void GenerateTableWithCopy(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize, X86InstInfo *OtherLocal) {
+constexpr static inline void GenerateTableWithCopy(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize, X86InstInfo *OtherLocal) {
   for (size_t j = 0; j < TableSize; ++j) {
     X86TablesInfoStruct<OpcodeType> const &Op = LocalTable[j];
     auto OpNum = Op.first;
     X86InstInfo const &Info = Op.Info;
     for (uint32_t i = 0; i < Op.second; ++i) {
-      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+      if (FinalTable[OpNum + i].Type != TYPE_UNKNOWN) {
+        ERROR_AND_DIE_FMT("Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+      }
       if (Info.Type == TYPE_COPY_OTHER) {
         FinalTable[OpNum + i] = OtherLocal[OpNum + i];
       }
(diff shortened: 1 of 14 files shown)


lioncash (commented):
Nice! Are some of the tables variables themselves also able to be made const/constexpr as well? (rather than just the lambda being consteval?)

Looks good either way though

neobrain:
> Nice! Are some of the tables variables themselves also able to be made const/constexpr as well? (rather than just the lambda being consteval?)

I wrote a version of this a while ago, though I wasn't happy with the executable size overhead: https://github.com/FEX-Emu/FEX/compare/main...neobrain:FEX:opt_x86tables_init_next

I think the tables carry slightly less information than back then, so maybe this has changed.

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -507,26 +507,24 @@ using U8U8InfoStruct = X86TablesInfoStruct<uint8_t>;
...
 
 template<typename OpcodeType>
-static inline void GenerateTable(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize) {
+constexpr static inline void GenerateTable(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize) {
   for (size_t j = 0; j < TableSize; ++j) {
     X86TablesInfoStruct<OpcodeType> const &Op = LocalTable[j];
     auto OpNum = Op.first;
     X86InstInfo const &Info = Op.Info;
     for (uint32_t i = 0; i < Op.second; ++i) {
-      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
```
neobrain, line 516: Not sure how I feel about dropping self-consistency checks for the sake of saving a few kilobytes of executable memory. How about re-adding the checks as a post=init verification step in `InitializeInfoTables`?

Luckily this can even be generically [as I've done here](https://github.com/FEX-Emu/FEX/compare/main...neobrain:FEX:opt_x86tables_init_next).

neobrain:
I wrote a patch very similar to this but wasn't happy with the binary size overhead. Do you know what changed meanwhile, since I measured a size *increase* of relative 5% whereas you're mentioning a size decrease?

I vaguely recall some flags being removed recently, but not sure that make such a huge difference.

For reference, my patch is here: https://github.com/FEX-Emu/FEX/compare/main...neobrain:FEX:opt_x86tables_init_next

Sonicadvance1:
> I wrote a patch very similar to this but wasn't happy with the binary size overhead. Do you know what changed meanwhile, since I measured a size _increase_ of relative 5% whereas you're mentioning a size decrease?
> 
> I vaguely recall some flags being removed recently, but not sure that make such a huge difference.
> 
> For reference, my patch is here: [main...neobrain:FEX:opt_x86tables_init_next](https://github.com/FEX-Emu/FEX/compare/main...neobrain:FEX:opt_x86tables_init_next)

The file size increased a small amount (Half a megabyte I think?) but the VM size reduced, since it doesn't need to maintain both tables.

neobrain:
> The file size increased a small amount (Half a megabyte I think?) but the VM size reduced, since it doesn't need to maintain both tables.

That's interesting. Where did the binary increase go to? Why is VM size the decisive metric here?

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
Sonicadvance1: Post initialization verification is hard to do since there will be table entries that remain having their type set to `TYPE_UNKNOWN`.
We're checking to ensure the values that we're setting are still `TYPE_UNKNOWN` before overwriting it. 
We can't do consteval asserts to match this behaviour as far as I can tell.

For your implementation of `CheckForInternalConflicts` it is implemented with the comment of `// TODO: Implement?`

Sonicadvance1:
> > The file size increased a small amount (Half a megabyte I think?) but the VM size reduced, since it doesn't need to maintain both tables.
> 
> That's interesting. Where did the binary increase go to? Why is VM size the decisive metric here?

VM size is more interesting since that means how much memory will be allocated and consumed in a new process. File size is less interesting since when binfmt_misc is installed, the kernel is going to keep the file in memory regardless, it won't be accessing the disk.

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
neobrain: On second look I think `CheckForInternalConflicts` only ever intended to check for conflicts within each *individual* `LocalTable` but not between *different* once. The former makes the check trivial, hence the TODO. 

>  We can't do consteval asserts to match this behaviour as far as I can tell.

Yeah, that would require one `.cpp` file to have full visibility of all the various LocalTables, upon which it could concatenate them and then check for duplicates. And that's clearly contrary to the whole point of splitting these tables into different files in the first place.

I realized one thing though: Have you tried simply leaving the `LOGMAN_THROW_AA_FMT(...)` in? You should actually be able to have non-constexpr code in a `consteval` lambda as long as the code doesn't execute. It will only fail if the assert indeed is failing then. If for some reason it doesn't compile right now, I think there should be an easy way to make it work.

Sonicadvance1: Yes I tried leaving them in and it results in a compile failure.
```
/mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/X86Tables/SecondaryGroupTables.cpp:15:80: error: call to consteval function 'FEXCore::X86Tables::(anonymous class)::operator()' is not a constant expression
std::array<X86InstInfo, MAX_INST_SECOND_GROUP_TABLE_SIZE> SecondInstGroupOps = []() consteval {
                                                                               ^
/mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/X86Tables/X86Tables.h:516:7: note: non-constexpr function 'AFmt<const char *, const char *>' cannot be used in a constant expression
      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
      ^
/mnt/Work/Work/work/FEXNew/FEXCore/include/FEXCore/Utils/LogManager.h:75:45: note: expanded from macro 'LOGMAN_THROW_AA_FMT'
#define LOGMAN_THROW_AA_FMT(pred, ...) do { LogMan::Throw::AFmt(pred, __VA_ARGS__); } while (0)
                                            ^
/mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/X86Tables/SecondaryGroupTables.cpp:490:3: note: in call to 'GenerateTable(&Table._M_elems[0], &SecondaryExtensionOpTable[0], 384)'
  GenerateTable(&Table.at(0), SecondaryExtensionOpTable, std::size(SecondaryExtensionOpTable));
  ^
/mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/X86Tables/SecondaryGroupTables.cpp:15:80: note: in call to '&[]() {
```

neobrain:
> VM size is more interesting since that means how much memory will be allocated and consumed in a new process. File size is less interesting since when binfmt_misc is installed, the kernel is going to keep the file in memory regardless, it won't be accessing the disk.

I'm not following. Could you explain what VM size actually is and how you measure it? It's not a web-searchable term, and it sounds like I guessed wrong about its meaning.

Regarding my original question, do you know which of the ELF sections in the FEX binary specifically has grown?

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
neobrain: Looks like it fails because that unconditionally calls `LogMan::Throw::AFmt`. What happens if you wrap the calls in the actual assertion condition?

```cpp
if (FinalTable[OpNum + i].Type != TYPE_UNKNOWN) {
      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
}
```

(Obviously we might as well just use an `ERROR_AND_DIE_FMT` or similar at that point)

Sonicadvance1: Turns out that does work. Thought I tried that.

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -538,21 +542,40 @@ static inline void GenerateTableWithCopy(X86InstInfo *FinalTable, X86TablesInfoS
...
+template<typename OpcodeType>
+constexpr static inline void GenerateX87Table(X86InstInfo *FinalTable, X86TablesInfoStruct<OpcodeType> const *LocalTable, size_t TableSize) {
   for (size_t j = 0; j < TableSize; ++j) {
     X86TablesInfoStruct<OpcodeType> const &Op = LocalTable[j];
     auto OpNum = Op.first;
     X86InstInfo const &Info = Op.Info;
     for (uint32_t i = 0; i < Op.second; ++i) {
-      LOGMAN_THROW_AA_FMT(FinalTable[OpNum + i].Type == TYPE_UNKNOWN, "Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+      if (FinalTable[OpNum + i].Type != TYPE_UNKNOWN) {
+        LOGMAN_MSG_A_FMT("Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
```
neobrain, line 567: There's no need to use debug-only primitives here anymore. Currently, this won't trigger a build failure in Release builds, since LOGMAN_MSG_A_FMT is no-opped there.

neobrain (approved):

PR #3321, opened 2023-12-11
FEXCore: Changes ParentThread ownership from the CTX to the frontend, take 2

Sonicadvance1:
Similar to #3284 but works around some of the bugs that one introduced.

This is the minimal amount of changes to move the ownership from FEXCore to the frontend. Since the frontends don't yet have a full thread state tracking, there is an opaque pointer that needs to be managed.

In the followup commits this will be changed to have the syscall handler to be the thread object manager.

### DIFF
--- a/FEXCore/include/FEXCore/Core/Context.h
+++ b/FEXCore/include/FEXCore/Core/Context.h
@@ -124,7 +124,7 @@ namespace FEXCore::Context {
        *
        * @return true if we loaded code
        */
-      FEX_DEFAULT_VISIBILITY virtual FEXCore::Core::InternalThreadState* InitCore(uint64_t InitialRIP, uint64_t StackPointer) = 0;
+      FEX_DEFAULT_VISIBILITY virtual bool InitCore() = 0;
 
       FEX_DEFAULT_VISIBILITY virtual void SetExitHandler(ExitHandler handler) = 0;
       FEX_DEFAULT_VISIBILITY virtual ExitHandler GetExitHandler() const = 0;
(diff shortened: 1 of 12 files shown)


FEXCore/include/FEXCore/Core/Context.h
```diff
@@ -248,17 +248,24 @@ namespace FEXCore::Context {
...
+       * @param WhoManages The flag to determine what manages ownership of the InternalThreadState object
        * @param NewThreadState The thread state to inherit from if not nullptr.
        * @param ParentTID The thread ID that the parent is inheriting from
        *
        * @return A new InternalThreadState object for using with a new guest thread.
        */
-      FEX_DEFAULT_VISIBILITY virtual FEXCore::Core::InternalThreadState* CreateThread(uint64_t InitialRIP, uint64_t StackPointer, FEXCore::Core::CPUState *NewThreadState = nullptr, uint64_t ParentTID = 0) = 0;
+
+      // TODO: This is a temporary construct and will be removed once the frontend has full ownership of InternalThreadState objects.
+      enum class ManagedBy {
```
neobrain, line 259: Sounds like a call for add-as-deprecated?

```suggestion
      // TODO: This is a temporary construct and will be removed once the frontend has full ownership of InternalThreadState objects.
      enum class [[deprecated]] ManagedBy {
```

neobrain (commented):
Looks like a lot of "temporaries" being added in this PR, pending on the frontends doing full thread state tracking.

What's the transition plan on that?
Is this an immediate next step or do you have a timeline in mind?

Sonicadvance1:
> Looks like a lot of "temporaries" being added in this PR, pending on the frontend doing full thread state tracking.
> 
> What's the transition plan on that? Is this an immediate next step or do you have a timeline in mind?

https://github.com/FEX-Emu/FEX/pull/3326 This is the transition plan.

lioncash (approved):

PR #3326, opened 2023-12-12
FEXLoader: Moves thread management to the frontend 

Sonicadvance1:
Lots going on here.

This moves OS thread object lifetime management and internal thread
state lifetime management to the frontend. This causes a bunch of thread
handling to move from the FEXCore Context to the frontend.

Looking at `FEXCore/include/FEXCore/Core/Context.h` really shows how
much of the API has moved to the frontend that FEXCore no longer needs
to manage. Primarily this makes FEXCore itself no longer need to care
about most of the management of the emulation state.

A large amount of the behaviour moved wholesale from Core.cpp to
LinuxEmulation's ThreadManager.cpp. Which this manages the lifetimes of
both the OS threads and the FEXCore thread state objects.

One feature lost was the instruction stepping, but this was already
buggy and is going to be rewritten/fixed when gdbserver work continues.

Now that all of this management is moved to the frontend, the gdbserver
can start improving since it can start managing all thread state
directly.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.cpp
+++ b/FEXCore/Source/Interface/Context/Context.cpp
@@ -34,20 +34,12 @@ namespace FEXCore::Context {
     return CustomExitHandler;
   }
 
-  void FEXCore::Context::ContextImpl::Stop() {
-    Stop(false);
-  }
-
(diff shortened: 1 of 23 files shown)


lioncash:
Needs a rebase

lioncash (approved):

PR #3327, opened 2023-12-13
Arm64: Removes a vtable indirection in syscalls 

Sonicadvance1:
We can safely call virtual functions through the JIT with a little bit
of work.

FEX's JIT has quite a few steps before it gets to a syscall handler.

Before this commit:
JIT->static HandleSyscall->SyscallHandler::HandleSyscall->SyscallHandler

After this commit:
JIT->SyscallHandler::HandleSyscall->SyscallHandler

A bit hard to notice this when this interface can spin at 67-million
calls per second though.

### DIFF
--- a/FEXCore/Source/Utils/MemberFunctionToPointer.h
+++ b/FEXCore/Source/Utils/MemberFunctionToPointer.h
@@ -17,13 +17,15 @@ class MemberFunctionToPointerCast final {
   public:
     MemberFunctionToPointerCast(PointerToMemberType Function) {
       memcpy(&PMF, &Function, sizeof(PMF));
+    }
 
+    uintptr_t GetConvertedPointer() const {
 #ifdef _M_X86_64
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Utils/MemberFunctionToPointer.h
```diff
@@ -17,24 +17,52 @@ class MemberFunctionToPointerCast final {
   public:
     MemberFunctionToPointerCast(PointerToMemberType Function) {
```
alyssarosenzweig, line 18: This seems to lose the assertions. Is that intended?

alyssarosenzweig (approved):

FEXCore/Source/Utils/MemberFunctionToPointer.h
Sonicadvance1: Intended, the asserts get moved to the accessor functions `GetConvertedPointer` and `GetVTableOffset` instead.

PR #3331, opened 2023-12-16
Tools: Moves IRLoader to independent folder

Sonicadvance1:
Which requires moving LinuxEmulation to its own independent folder as well. Since both IRLoader and FEXLoader rely on it.

No functional change, just moves the the code around.


Source/Tools/IRLoader/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+
+set (DEFINES)
+if (ENABLE_VIXL_SIMULATOR)
+  list(APPEND DEFINES -DVIXL_SIMULATOR=1)
+endif()
+
+add_executable(IRLoader
+  IRLoader.cpp
+)
+target_compile_definitions(IRLoader PRIVATE ${DEFINES})
```
neobrain, line 11: Dropping use of the variable keeps control flow straightforward:

```suggestion
if (ENABLE_VIXL_SIMULATOR)
    target_compile_definitions(IRLoader PRIVATE "-DVIXL_SIMULATOR=1")
endif()
```

neobrain (approved):

PR #3337, opened 2023-12-17
FEXCore: Implements an efficient spin-loop API

Sonicadvance1:
This will only be used internally inside of FEXCore for efficient shared codecach backpatch spin-loops.

### DIFF
--- a/FEXCore/Source/Utils/FutexSpinWait.cpp
+++ b/FEXCore/Source/Utils/FutexSpinWait.cpp
@@ -1,0 +1,27 @@
+#include "Utils/FutexSpinWait.h"
+
+namespace FEXCore::Utils::FutexSpinWait {
+#ifdef _M_ARM_64
+  constexpr uint64_t NanosecondsInSecond = 1'000'000'000ULL;
+
+  static uint32_t GetCycleCounterFrequency() {
+    uint64_t Result{};
+    __asm("mrs %[Res], CNTFRQ_EL0"
+        : [Res] "=r" (Result));
+    return Result;
+  }
+
+  static uint64_t CalculateCyclesPerNanosecond() {
+    // Snapdragon devices historically use a 19.2Mhz cycle counter frequency
+    // This means that the number of cycles per nanosecond ends up being 52.0833...
+    //
+    // ARMv8.6 and ARMv9.1 requires the cycle counter frequency to be 1Ghz.
+    // This means the number of cycles per nanosecond ends up being 1.
+    uint64_t CounterFrequency = GetCycleCounterFrequency();
+    return NanosecondsInSecond / CounterFrequency;
+  }
+
+  uint32_t CycleCounterFrequency = GetCycleCounterFrequency();
+  uint64_t CyclesPerNanosecond = CalculateCyclesPerNanosecond();
+#endif
+}
--- a/FEXCore/Source/Utils/FutexSpinWait.h
+++ b/FEXCore/Source/Utils/FutexSpinWait.h
@@ -1,0 +1,266 @@
+#include <atomic>
+#include <chrono>
+#include <mutex>
+#include <type_traits>
+
+namespace FEXCore::Utils::FutexSpinWait {
+  /**
+   * @brief This provides routines to implement implement an "efficient spin-loop" using ARM's WFE and exclusive monitor interfaces.
+   *
+   * Spin-loops on mobile devices with a battery can be a bad idea as they burn a bunch of power. This attempts to mitigate some of the impact
+   * by putting the CPU in to a lower-power state using WFE.
+   * On platforms tested, WFE will put the CPU in to a lower power state for upwards of 52ns per WFE. Which isn't a significant amount of time
+   * but should still have power savings. Ideally WFE would be able to keep the CPU in a lower power state for longer. This also has the added benefit
+   * that atomics aren't abusing the caches when spinning on a cacheline, which has knock-on powersaving benefits.
+   *
+   * FEAT_WFxT adds a new instruction with a timeout, but since the spurious wake-up is so aggressive it isn't worth using.
+   *
+   * It should be noted that this implementation has a few dozen cycles of start-up time. Which means the overhead for invoking this implementation is
+   * slightly higher than a true spin-loop. The hot loop body itself is only three instructions so it is quite efficient.
+   *
+   * On non-ARM platforms it is truly a spin-loop, which is okay for debugging only.
+   */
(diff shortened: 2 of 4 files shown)


FEXCore/Source/Utils/FutexSpinWait.cpp
```diff
@@ -0,0 +1,39 @@
+#include "Utils/FutexSpinWait.h"
+
+namespace FEXCore::Utils {
+#ifdef _M_ARM_64
+  std::atomic<bool> FEXCore::Utils::FutexSpinWait::StaticDataInitialized;
+  std::mutex FEXCore::Utils::FutexSpinWait::StaticInitMutex;
+  uint32_t FEXCore::Utils::FutexSpinWait::CycleCounterFrequency;
+  uint64_t FEXCore::Utils::FutexSpinWait::CyclesPerNanosecond;
```
neobrain, line 8: Why not initialize these directly?

```suggestion
  uint32_t FEXCore::Utils::FutexSpinWait::CycleCounterFrequency = GetCycleCounterFrequency();
  uint64_t FEXCore::Utils::FutexSpinWait::CyclesPerNanosecond = CalculateCyclesPerNanosecond();
```

In fact, doing so allows for dropping CalculateCyclesPerNanosecond entirely and just moving its comment up here. Then you don't need the `Fairly trivial calculation but broken out for additional information.` disclaimer either, which was a bit of a bumpy introduction anyway.

FEXCore/Source/Utils/FutexSpinWait.h
```diff
@@ -0,0 +1,282 @@
...
+   * On platforms tested, WFE will put the CPU in to a lower power state for upwards of 52ns per WFE. Which isn't a significant amount of time
+   * but should still have power savings. Ideally WFE would be able to keep the CPU in a lower power state for longer. This also has the added benefit
+   * that atomics aren't abusing the caches when spinning on a cacheline, which has knock-on powersaving benefits.
+   *
+   * FEAT_WFxT adds a new instruction with a timeout, but since the spurious wake-up is so aggressive it isn't worth using.
+   *
+   * It should be noted that this implementation has a few dozen cycles of start-up time. Which means the overhead for invoking this implementation is
+   * slightly higher than a true spin-loop. The hot loop body itself is only three instructions so it is quite efficient.
+   *
+   * On non-ARM platforms it is truly a spin-loop. Which is okay for debugging only.
```
neobrain, line 54: Relative clauses are connected using comma:

```suggestion
   * On non-ARM platforms it is truly a spin-loop, which is okay for debugging only.
```

FEXCore/Source/Utils/FutexSpinWait.h
```diff
@@ -0,0 +1,282 @@
...
+   */
+  class FutexSpinWait final {
+    public:
+#ifdef _M_ARM_64
+      template<typename T, typename TT>
+      static void Wait(T *Futex, TT ExpectedValue) {
```
neobrain, line 60: Since this is a stateless class with static member functions, `FutexSpinWait` is better-suited to be a namespace. That has the same syntax at call-site, but doesn't require the reader to scan for potential class variables or non-static functions.

FEXCore/Source/Utils/FutexSpinWait.h
```diff
@@ -0,0 +1,282 @@
...
+
+namespace FEXCore::Utils {
+#define SPINLOOP_8BIT " \
+  /* Prime the exclusive monitor with the passed in address. */ \
+  ldaxrb %w[Tmp], [%[Futex]]; \
+  /* WFE will wait for either the memory to change or spurious wake-up. */ \
+  wfe; \
+  /* Load with acquire to get the result of memory. */ \
+  ldarb %w[Result], [%[Futex]]; \
+"
```
neobrain, line 14: Why not have a single macro that takes `ldaxrb` and `ldarb` as arguments? That's less explicit than having the `_8BIT` suffix, but since the macros are only used in `sizeof(T)==1` branches the meaning is clear anyway.

FEXCore/Source/Utils/FutexSpinWait.h
```diff
@@ -0,0 +1,282 @@
...
+              : "memory");
+          }
+          else {
+            static_assert(!std::is_same_v<T, T>, "Invalid");
+          }
+        } while (Result != ExpectedValue);
+      }
+
+      template<typename T, typename TT>
+      static bool Wait(T *Futex, TT ExpectedValue, std::chrono::nanoseconds const &Timeout) {
```
neobrain, line 110: I'm guessing these are implemented directly to allow inlining, however it's very difficult to actually see what interface is being exposed that way. It would help to forward-declare all of these functions in a single block hence.

FEXCore/Source/Utils/FutexSpinWait.h
```diff
@@ -0,0 +1,282 @@
...
+   *
+   * It should be noted that this implementation has a few dozen cycles of start-up time. Which means the overhead for invoking this implementation is
+   * slightly higher than a true spin-loop. The hot loop body itself is only three instructions so it is quite efficient.
+   *
+   * On non-ARM platforms it is truly a spin-loop. Which is okay for debugging only.
+   */
+  class FutexSpinWait final {
+    public:
+#ifdef _M_ARM_64
+      template<typename T, typename TT>
```
neobrain, line 59: Why do we need different integer-sizes for this?

neobrain (commented):
Went over this for a general first pass.

As usual newly added abstractions can be reviewed more effectively when there's code that actually uses them, so a follow-up draft PR for the adapted backpatching would help here.

FEXCore/Source/Utils/FutexSpinWait.h
Sonicadvance1: Because the following idiom is the expected use-case and enforcing an integer literal on the second argument is a pita.
```cpp
uint8_t *VariablePtr;
FEXCore::Utils::FutexSpinWait::Wait(VariablePtr, 0);
```

Sonicadvance1:
> Went over this for a general first pass.
> 
> As usual newly added abstractions can be reviewed more effectively when there's code that actually uses them, so a follow-up draft PR for the adapted backpatching would help here.

https://github.com/FEX-Emu/FEX/pull/3338
There you go.

neobrain:
> #3338 There you go.

Thanks, but there was no confusion on the intended usage of the new API. Before adding new code for a primitive, I think it's reasonable to have a concrete use case where it's demonstrably a better solution than the alternatives to justify its addition.

PR #3341, opened 2023-12-19
IREmitter: Slightly optimize linked list appending

Sonicadvance1:
If we set `CurrentWriteCursor` to the invalid node immediately then we can remove the per operation check in `CreateNode` which wasn't getting optimized away. This was causing checks to occur for every node emitted even in opdispatch handlers that emitted several IR operations unconditionally.

Since we always need to emit the `InvalidNode` upfront anyway, we know we will always have a linked list object to append to, eliminating the check at all times.

This flattens a bunch of code generated in the opcode dispatcher that I had presumed would only do a single check upfront. The amount of code actually emitted is still very similar so this won't be a significant savings.

Before:
Image_2023-12-18_19-44-30

After:
Image_2023-12-18_19-44-41


Sonicadvance1:
Oop okay, this is going to be harder than expected. Maybe I'll return to this later.

PR #3344, opened 2023-12-21
Externals: Update xbyak to v7.02 and switch away from fork

Sonicadvance1:
The last few patches we need have been upstreamed so we shouldn't need our downstream fork anymore.


alyssarosenzweig (approved):
Awesome, moving closer to upstream makes me happy to see :smile_cat:

PR #3345, opened 2023-12-21
OpcodeDispatcher: Fixes syscall rcx/r11 generation

### LINKED ISSUE #3343: Syscall instruction doesn't set RCX and R11 correctly
Syscall is supposed to set RCX to the RIP following the instruction and R11 to the EFLAGS.
This is so the kernel can then use `Sysret` which moves these registers back to RIP and EFLAGS.
Noticed this in a new unittest I'm writing.

Sonicadvance1:
Noticed this while writing #3342.

Fixes #3343

The syscall instruction is defined in the documentation that it will set RCX to the next instruction's RIP and R11 to be RFLAGS. We entirely skipped this which I noticed while writing unit tests.

Adds unittests to test both 32-bit and 64-bit behaviour because our helper shares code with both.

I don't know if anything actually relied on this behaviour but we should definitely support it.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -36,6 +36,7 @@ using X86Tables::OpToIndex;
 
 #define OpcodeArgs [[maybe_unused]] FEXCore::X86Tables::DecodedOp Op
 
+template<bool IsSyscallInst>
 void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
   constexpr size_t SyscallArgs = 7;
   using SyscallArray = std::array<uint64_t, SyscallArgs>;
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -35,6 +35,7 @@ using X86Tables::OpToIndex;
 
 #define OpcodeArgs [[maybe_unused]] FEXCore::X86Tables::DecodedOp Op
 
+template<bool IsInt80>
```
bylaws, line 38: Probably worth choosing a different name for this, as it will be true even on windows, where the syscall immediate is 0x2E

unittests/FEXLinuxTests/tests/signal/Syscall_state.32.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+struct CPUState {
+  uint32_t Registers[8];
+  uint32_t eflags;
+};
+
+CPUState CapturedState{};
+extern "C" uint64_t HLT_INST;
+
+const void *HaltLocation = &HLT_INST;
```
neobrain, line 17: This doesn't seem to be used.

unittests/FEXLinuxTests/tests/signal/Syscall_state.64.cpp
```diff
@@ -0,0 +1,174 @@
...
+struct CPUState {
+  uint64_t Registers[16];
+
+  uint64_t eflags;
+};
+
+CPUState CapturedState{};
+extern "C" uint64_t HLT_INST;
+
+const void *HaltLocation = &HLT_INST;
```
neobrain, line 18: Looks like you meant to make this pointer constant, but it's actually a pointer to constant (untyped) data.

```suggestion
void * const HaltLocation = &HLT_INST;
```

Yes, I'm just as annoyed by this 🙃

unittests/FEXLinuxTests/tests/signal/Syscall_state.64.cpp
```diff
@@ -0,0 +1,174 @@
...
+#include <sys/wait.h>
+
+struct CPUState {
+  uint64_t Registers[16];
+
+  uint64_t eflags;
+};
+
+CPUState CapturedState{};
+extern "C" uint64_t HLT_INST;
```
neobrain, line 16: Could probably do with a comment explaining that this refers to the label defined in the inline assembly below.

alyssarosenzweig (approved):

PR #3350, opened 2023-12-25
FEXCore: Optimize HostFeatures and CPUID feature calculation

Sonicadvance1:
~~Need https://github.com/FEX-Emu/FEX/pull/3348 merged first.~~

As I was casually thinking, this code made me realize that it was quite
branch heavy and could likely be optimized to logic.

The previous code generated some fairly nasty branch heavy code. This
can be optimized to be branchless and take roughly five instructions
per flag. Using a bitfield for each feature would turn each calculation
in to 3-4 instructions but that seems overkill.

Very minor thing.

Before:
Image_2023-12-25_05-03-14

After:
Image_2023-12-25_05-03-22

### DIFF
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -441,6 +441,30 @@ def print_parse_envloader_options(options):
                 output_argloader.write("}\n")
     output_argloader.write("#endif\n")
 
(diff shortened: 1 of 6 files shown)


alyssarosenzweig (approved):
nice!

PR #3349, opened 2023-12-24
Revert "FEXLoader: Moves thread management to the frontend"

Sonicadvance1:
This reverts commit 58f26939548cba8b26ec4771aeed6b54e3700046.

This is causing crashes in Proton games right now. So I need to spend a bit more time to figure out where this went wrong.

PR #3355, opened 2023-12-28
Linux: Converts passthrough syscalls to generator script 

Sonicadvance1:
~~Needs #3349 and #3354 merged first.~~

After adding the past few Linux versions of syscalls I noticed that this
could be better.

Walks through all of our syscalls and converts them to a json format
that is parsed and then generated at compile time.

This also fixes some mislabeling of system calls that were getting
inlined on accident.

This really helps clean up the syscall implementation, all syscalls that
can get passed through to the host syscall handler is no longer visible
in the implementation.


neobrain:
I like the prospect of cleaning up the syscalls handling and making it less prone to small bugs. Not sure I'm a fan of the implementation chosen here though. You've added a layer of code generation (Python) on top of another (C macros), which in turn uses a third code-generating mechanism (C++ templates). The syscalls implementation cleanup comes at the cost of increased architectural complexity.

Generally I'm not a fan of involving more Python scripts into our build process, since we're already at a point where minor changes trigger almost full rebuilds since everything depends on everything. This has tangible effects on the developer workflow, since I often find myself waiting for lengthy recompiles after rebasing even though I didn't touch any FEXCore code.

Since the state captured in SyscallDescription.json is super simple, have you considered any C++-based alternatives? The code we use today seems straightforward enough to turn into a single template with simple `if constexpr`s.

(As a side note, it looks like e1c6cb2173bf42c5e5bd9f5ec25dfecc1b14bf0e accidentally slipped into this PR.)

Sonicadvance1:
Thank you for reviewing the code.

PR #3356, opened 2023-12-29
Jitarm64: Implements spin-loop futex for JIT blocks 

Sonicadvance1:
~~Needs #3339 merged first.~~

This will ensure that multiple concurrent SIGBUS handlers in the same
code block doesn't modify the same code.

### DIFF
--- a/FEXCore/Source/CMakeLists.txt
+++ b/FEXCore/Source/CMakeLists.txt
@@ -5,9 +5,9 @@ set (FEXCORE_BASE_SRCS
   Utils/Allocator.cpp
   Utils/CPUInfo.cpp
   Utils/FileLoading.cpp
-  Utils/FutexSpinWait.cpp
   Utils/ForcedAssert.cpp
   Utils/LogManager.cpp
+  Utils/SpinWaitLock.cpp
   )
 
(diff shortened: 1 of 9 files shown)


bylaws:
I feel like FutexSpinWait probably isn't the best name for this, I haven't see futex used to refer to anything but the Linux syscall, which this explicitly doesn't use

Sonicadvance1:
> I feel like FutexSpinWait probably isn't the best name for this, I haven't see futex used to refer to anything but the Linux syscall, which this explicitly doesn't use

Renamed to SpinWaitLock

alyssarosenzweig:
There's nothing appropriate in the stl? or provided by Linux? Why do we need to roll our own?

Sonicadvance1:
> There's nothing appropriate in the stl? or provided by Linux? Why do we need to roll our own?

Nothing in the STL, libc, or Linux similar to a spin-loop.
Closest things end up being `std::condition_variable` or `pthread_cond_t`, or a kernel futex. All of these are too heavy for what is getting accomplished here, which is a spin-loop/spin-lock.

lioncash (approved):

PR #3358, opened 2023-12-29
JIT: Fixes broken register in VTBX1

wannacu:
If the Dst register is allocated as VectorIndices or VectorTable, using Dst as an operand to perform the tbx operation will result in an error. For example:
%131(FPR0) i128 = LoadNamedVectorIndexedConstant u8:Tmp:RegisterSize, #0x6, #0xaa0 
%132(FPR0) i128 = VTBX1 u8:Tmp:RegisterSize, %129(FPRFixed6) i32v4, %126(FPRFixed10) i16v8, %131(FPR0) i128 
Since the tbx instruction's destination register is also the original operand, this is consistent with the semantics of VTBX1. Therefore, directly using VectorSrcDst as the destination operand for the tbx instruction is safe.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -5009,42 +5009,50 @@ DEF_OP(VTBX1) {
   if (Dst != VectorSrcDst) {
     switch (OpSize) {
       case 8: {
-        mov(Dst.D(), VectorSrcDst.D());
+        mov(VTMP1.D(), VectorSrcDst.D());
+        tbx(VTMP1.D(), VectorTable.Q(), VectorIndices.D());
+        mov(Dst.D(), VTMP1.D());
         break;
       }
       case 16: {
-        mov(Dst.Q(), VectorSrcDst.Q());
+        mov(VTMP1.Q(), VectorSrcDst.Q());
+        tbx(VTMP1.Q(), VectorTable.Q(), VectorIndices.Q());
+        mov(Dst.Q(), VTMP1.Q());
         break;
       }
       case 32: {
-        mov(Dst.Z(), VectorSrcDst.Z());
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (changes requested):
I see the bug you're trying to fix here and this actually introduces a new bug where this will overwrite the value in `VectorSrcDst` even when `Dst != VectorSrcDst`. So it needs to be fixed in a different way.

Sonicadvance1:
> > I see the bug you're trying to fix here and this actually introduces a new bug where this will overwrite the value in `VectorSrcDst` even when `Dst != VectorSrcDst`. So it needs to be fixed in a different way.
> 
> According to the semantics of the VTBX1 instruction, overwriting VectorSrcDst is safe.

According to VTBX1 it is safe yes, but according to the rules of our IR it is not. Overwriting an incoming SSA value that doesn't alias/overlap the destination isn't safe.

FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
```diff
@@ -5009,42 +5009,50 @@ DEF_OP(VTBX1) {
   if (Dst != VectorSrcDst) {
     switch (OpSize) {
       case 8: {
-        mov(Dst.D(), VectorSrcDst.D());
+        mov(VTMP1.D(), VectorSrcDst.D());
+        tbx(VectorSrcDst.D(), VectorTable.Q(), VectorIndices.D());
```
Sonicadvance1, line 5013: This tbx and the following two which are using the VTMP1 are incorrectly still using `VectorSrcDst` as their destination. Should use VTMP1 to be correct here.

Sonicadvance1: Yes, we want to improve our RA to support constraints to handle these cases. It's on the roadmap.

Sonicadvance1 (approved):
Thanks for the fix. I completely missed this.

PR #3361, opened 2024-01-07
FEXCore: Decompose some std::function usage to regular pointers

Sonicadvance1:
The delinker step of the JIT was using std::function with capture lambdas that required memory allocation when unnecessary. Because the compiler can't see through our std::function usage it could never decompose these by itself.

By passing the Thread's frame and record to the function as arguments then we can have the signature be a raw function pointer.

This fixes an area of concern from:
https://github.com/FEX-Emu/FEX/blob/main/docs/ProgrammingConcerns.md#stdfunction-and-lambdas


alyssarosenzweig (approved):
I like less C++

PR #3362, opened 2024-01-09
Fixes some new glibc allocations that cropped up

Sonicadvance1:
I guess this was handled by brk things before.

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
@@ -21,6 +21,7 @@
 #endif
 
 #include <FEXCore/Config/Config.h>
+#include <FEXCore/fextl/vector.h>
 
 #include <array>
 #include <cstddef>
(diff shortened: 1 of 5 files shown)


alyssarosenzweig (approved):

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
```diff
@@ -233,7 +233,9 @@ class Arm64Emitter : public FEXCore::ARMEmitter::Emitter {
 #endif
 
 #ifdef VIXL_DISASSEMBLER
-  vixl::aarch64::Disassembler Disasm;
+  std::vector<char> DisasmBuffer;
```
neobrain, line 236: Should this be a `fextl::vector`?

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -362,7 +362,7 @@ namespace FEX::VDSO {
   }
 
   void LoadHostVDSO() {
-
+    FEXCore::Allocator::YesIKnowImNotSupposedToUseTheGlibcAllocator glibc;
```
neobrain, line 365: Should have a comment mentioning why it's needed here.

neobrain (approved):

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
Sonicadvance1: Yes

PR #3363, opened 2024-01-09
ArmEmitter: Support single use forward labels

Sonicadvance1:
~~Needs #3361 merged first.~~

Previously forward labels needed to be pessimistic about multiple instructions binding to a label. This results in forward labels having a vector that always allocates memory. Instead if we know the data is only going to be used by a single instruction then we don't need to allocate any memory. This is the common case in FEX.

In particular the delinker step for the JIT blocks was always allocating memory due to the forward label. Now it no longer allocates any memory since it uses a single use forward label.


alyssarosenzweig (approved):
nice!

PR #3374, opened 2024-01-15
Library Forwarding: Test interaction of struct repacking and assume_compatible_data_layout

neobrain:
~~Needs https://github.com/FEX-Emu/FEX/pull/3373 merged first.~~

This just adds one more test to test the interaction between automatic struct repacking (added in #3371) and the previously added `assume_compatible_data_layout` annotation (#3177). This serves mainly as a consistency check to ensure other tests don't just pass "by accident".

While developing this, I also found that we aren't setting up `GUEST_THUNK_LIBRARY` properly, which the test library needs to enable guest-specific behavior however. A fix for this is included in the PR now as well.


neobrain:
Rebased now that the custom repacking code is merged.

Sonicadvance1 (approved):

PR #3375, opened 2024-01-17
Linux: Implements a fault safe memcpy routine

Sonicadvance1:
We are required in our syscall emulation to handle cases where pointers
are invalid. This means we need to pessimistically assume a memcpy will
fault when reading application memory.

This implements a signal handler based approach to catching the SIGSEGV
on memcpy and returning an EFAULT if it faults.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FaultSafeMemcpy.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FaultSafeMemcpy.cpp
@@ -1,0 +1,72 @@
+// SPDX-License-Identifier: MIT
+#include "LinuxSyscalls/Syscalls.h"
+
+namespace FEX::HLE::FaultSafeMemcpy {
+#ifdef _M_ARM_64
+__attribute__((naked))
+size_t CopyFromUser(void *Dest, const void* Src, size_t Size) {
+  __asm volatile(R"(
+  // Early exit if a memcpy of size zero.
+  cbz x2, 2f;
+
+  1:
+  .globl CopyFromUser_FaultInst
+  CopyFromUser_FaultInst:
+    ldrb w3, [x1], 1; // <- This line can fault.
+    strb w3, [x0], 1;
+    sub x2, x2, 1;
+    cbnz x2, 1b;
+2:
+    mov x0, 0;
+    ret;
+  )"
+  ::: "memory");
+}
+
+__attribute__((naked))
+size_t CopyToUser(void *Dest, const void* Src, size_t Size) {
+  __asm volatile(R"(
+  // Early exit if a memcpy of size zero.
+  cbz x2, 2f;
+
+  1:
+    ldrb w3, [x1], 1;
+  .globl CopyToUser_FaultInst
+  CopyToUser_FaultInst:
+    strb w3, [x0], 1; // <- This line can fault.
+    sub x2, x2, 1;
+    cbnz x2, 1b;
+2:
+    mov x0, 0;
+    ret;
+  )"
+  ::: "memory");
+}
+
+extern "C" uint64_t CopyFromUser_FaultInst;
+void * const CopyFromUser_FaultLocation = &CopyFromUser_FaultInst;
+
+extern "C" uint64_t CopyToUser_FaultInst;
+void * const CopyToUser_FaultLocation = &CopyToUser_FaultInst;
+
+bool IsFaultLocation(uint64_t PC) {
+  return reinterpret_cast<void*>(PC) == CopyFromUser_FaultLocation ||
+    reinterpret_cast<void*>(PC) == CopyToUser_FaultLocation;
+}
+
+#else
+size_t CopyFromUser(void *Dest, const void* Src, size_t Size) {
+  memcpy(Dest, Src, Size);
+  return Size;
+}
+
+size_t CopyToUser(void *Dest, const void* Src, size_t Size) {
+  memcpy(Dest, Src, Size);
+  return Size;
+}
+
+bool IsFaultLocation(uint64_t PC) {
+  return false;
+}
+#endif
+}
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
@@ -1471,10 +1471,17 @@ namespace FEX::HLE {
       else if (Signal == SIGSEGV &&
                SigInfo.si_code == SEGV_ACCERR &&
                FaultSafeMemcpy::IsFaultLocation(ArchHelpers::Context::GetPc(UContext))) {
-        // Return from the subroutine, returning EFAULT.
-        ArchHelpers::Context::SetArmReg(UContext, 0, EFAULT);
-        ArchHelpers::Context::SetPc(UContext, ArchHelpers::Context::GetArmReg(UContext, 30));
-        return;
+        // If you want to emulate EFAULT behaviour then enable this if-statement.
+        // Do this once we find an application that depends on this.
+        if constexpr (false) {
+          // Return from the subroutine, returning EFAULT.
+          ArchHelpers::Context::SetArmReg(UContext, 0, EFAULT);
+          ArchHelpers::Context::SetPc(UContext, ArchHelpers::Context::GetArmReg(UContext, 30));
+          return;
+        }
+        else {
+          LogMan::Msg::AFmt("Received invalid data to syscall. Crashing now!");
+        }
       }
       else {
         if (IsAsyncSignal(&SigInfo, Signal) && MustDeferSignal) {
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
@@ -7,6 +7,7 @@ $end_info$
 */
 
 #include "LinuxSyscalls/SignalDelegator.h"
+#include "LinuxSyscalls/Syscalls.h"
 
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Core/CoreState.h>
@@ -1466,6 +1467,14 @@ namespace FEX::HLE {
           ERROR_AND_DIE_FMT("X86 shouldn't hit this DeferredSignalFaultAddress");
 #endif
         }
+      }
+      else if (Signal == SIGSEGV &&
+               SigInfo.si_code == SEGV_ACCERR &&
+               FaultSafeMemcpy::IsFaultLocation(ArchHelpers::Context::GetPc(UContext))) {
+        // Return from the subroutine, returning EFAULT.
+        ArchHelpers::Context::SetArmReg(UContext, 0, EFAULT);
+        ArchHelpers::Context::SetPc(UContext, ArchHelpers::Context::GetArmReg(UContext, 30));
+        return;
       }
       else {
         if (IsAsyncSignal(&SigInfo, Signal) && MustDeferSignal) {
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
@@ -531,6 +531,16 @@ static bool HasSyscallError(const void* Result) {
 
 template<bool IncrementOffset, typename T>
 uint64_t GetDentsEmulation(int fd, T *dirp, uint32_t count);
+
+namespace FaultSafeMemcpy {
+// These are little helper functions for cases when FEX needs to copy data to or from the application in a robust fashion.
+// CopyFromUser and CopyToUser are memcpy routines that expect to safely SIGSEGV when reading or writing application memory respectively.
+// Returns zero if the memcpy completed, or crashes with SIGABRT and a log message if it faults.
+[[nodiscard]] size_t CopyFromUser(void *Dest, const void* Src, size_t Size);
+[[nodiscard]] size_t CopyToUser(void *Dest, const void* Src, size_t Size);
+bool IsFaultLocation(uint64_t PC);
+}
+
 }
 
 // Registers syscall for both 32bit and 64bit
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
@@ -263,7 +263,12 @@ namespace FEX::HLE::x32 {
       struct timespec tp64{};
       struct timespec *timed_ptr{};
       if (timeout_ts) {
-        tp64 = *timeout_ts;
+        struct timespec32 timeout{};
+        if (FaultSafeMemcpy::CopyFromUser(&timeout, timeout_ts, sizeof(timeout)) == EFAULT) {
+          return -EFAULT;
+        }
+
+        tp64 = timeout;
         timed_ptr = &tp64;
       }
 
@@ -275,7 +280,13 @@ namespace FEX::HLE::x32 {
         sigsetsize);
 
       if (timeout_ts) {
-        *timeout_ts = tp64;
+        struct timespec32 timeout{};
+        timeout = tp64;
+
+        if (FaultSafeMemcpy::CopyToUser(timeout_ts, &timeout, sizeof(timeout)) == EFAULT) {
+          // Write to user memory failed, this can occur if the timeout is defined in read-only memory.
+          // This is okay to happen, kernel continues happily.
+        }
       }
 
       SYSCALL_ERRNO();
--- a/unittests/FEXLinuxTests/Known_Failures
+++ b/unittests/FEXLinuxTests/Known_Failures
@@ -19,3 +19,6 @@ sigtest_samask.32
 sigtest_samask.64
 sigtest_sigmask.32
 sigtest_sigmask.64
+
+# Disabled since FEX's FaultSafeMemcpy is intentionally stub-implemented
+syscalls_efault.32
--- a/unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
+++ b/unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
@@ -1,0 +1,169 @@
+#include <catch2/catch.hpp>
+
+#include <cstdint>
+#include <unistd.h>
+#include <sys/mman.h>
+#include <sys/syscall.h>
+#include <poll.h>
+#include <signal.h>
+
+TEST_CASE("poll") {
+  // poll can return EFAULT if first argument is pointed to invalid pointer.
+  // Using mmap specifically for allocating with PROT_NONE.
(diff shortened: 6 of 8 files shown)


unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+TEST_CASE("poll") {
+  // poll can return EFAULT if first argument is pointed to invalid pointer.
+  struct pollfd *invalid_fds = reinterpret_cast<struct pollfd *>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  auto ret = ::syscall(SYS_poll, invalid_fds, 1, 0);
+  REQUIRE(ret == -1);
+  CHECK(errno == EFAULT);
+}
+
+TEST_CASE("ppoll", "[!mayfail]") {
```
neobrain, line 18: Why is this one marked as `mayfail` but the others aren't? Shouldn't all of them fail since we don't emulate the kernel-internal recovery in the latest version?

Sonicadvance1: Because only this one may fail and only on 32-bit.

neobrain: Why do the tests pass otherwise, even though we don't emulate the tested property? What do the tests test?

Sonicadvance1: The 64-bit tests pass successfully because we pass the syscalls to the kernel. The 32-bit tests need the emulation, thus the EFAULT handling. If the EFAULT capturing is disabled then Catch2 will catch the SIGSEGV resulting in a fail, if the EFAULT capturing is enabling then we capture and result in a SIGABRT that Catch2 doesn't catch.

Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -1471,9 +1471,16 @@ namespace FEX::HLE {
       else if (Signal == SIGSEGV &&
                (reinterpret_cast<void*>(ArchHelpers::Context::GetPc(UContext)) == FaultSafeMemcpy::CopyFromUser_FaultLocation ||
                reinterpret_cast<void*>(ArchHelpers::Context::GetPc(UContext)) == FaultSafeMemcpy::CopyToUser_FaultLocation)) {
-        // Jump to the copy fault handler return trampoline to safely return EFAULT.
-        ArchHelpers::Context::SetPc(UContext, reinterpret_cast<uint64_t>(&FaultSafeMemcpy::CopyFaultTrampoline));
-        return;
+        // If you want to emulate EFAULT behaviour then enable this if-statement.
```
alyssarosenzweig, line 1474: Why are we making FEX less accurate?

Sonicadvance1: @neobrain Is concerned about papering over emulation bugs uncovered through capturing EFAULT correctly. So at least now we can detect it until we find a game/application that requires it.

alyssarosenzweig: Yeah, OK, fair enough

Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -1467,6 +1468,20 @@ namespace FEX::HLE {
 #endif
         }
       }
+      else if (Signal == SIGSEGV &&
+               (reinterpret_cast<void*>(ArchHelpers::Context::GetPc(UContext)) == FaultSafeMemcpy::CopyFromUser_FaultLocation ||
+               reinterpret_cast<void*>(ArchHelpers::Context::GetPc(UContext)) == FaultSafeMemcpy::CopyToUser_FaultLocation)) {
```
neobrain, line 1473: Wrapping these two checks into a single helper function (like `FaultSafeMemcpy::IsFaultLocation(GetPc(UContext))`) keeps implementation details more contained, reducing noise here, in the header file, and in the x86 placeholder path.

Source/Tools/LinuxEmulation/LinuxSyscalls/FaultSafeMemcpy.cpp
```diff
@@ -0,0 +1,69 @@
...
+#include "LinuxSyscalls/Syscalls.h"
+
+namespace FEX::HLE::FaultSafeMemcpy {
+#ifdef _M_ARM_64
+__attribute__((naked))
+size_t CopyFromUser(void *Dest, const void* Src, size_t Size) {
+  __asm volatile(R"(
+  1:
+  CopyFromUser_FaultInst:
+    ldrb w3, [x1], 1; // <- This line can fault.
```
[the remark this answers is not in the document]
Sonicadvance1: You're right

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
neobrain: > It will exit with SIGABRT because of the log message in SignalDelegator when it catches a fault           LogMan::Msg::AFmt("Received invalid data to syscall. Crashing now!");

You're now referring to the case I was *not* asking about. I probably misinterpreted what you meant with EFAULT capturing, and it's hard to interpret your explanation based on guesswork.

What I do understand now is why the tests pass on 64-bit regardless of whether we enable the new emulation code or not. You fully lost me on what's going on for 32-bit though. Could you try explaining again why some of the tests are marked as `mayfail` for 32-bit and why others are not, while clearly mentioning which configurations you're referring to?

Source/Tools/LinuxEmulation/LinuxSyscalls/FaultSafeMemcpy.cpp
```diff
@@ -0,0 +1,69 @@
...
+    strb w3, [x0], 1; // <- This line can fault.
+    sub x2, x2, 1;
+    cbnz x2, 1b;
+    mov x0, 0;
+    ret;
+  )"
+  ::: "memory");
+}
+
+__attribute__((naked))
```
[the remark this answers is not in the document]
Sonicadvance1: We need to fully control the function, due to random preambles that can get added to functions. But also this function doesn't need to exist so I removed it.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -531,6 +531,25 @@ static bool HasSyscallError(const void* Result) {
 
 template<bool IncrementOffset, typename T>
 uint64_t GetDentsEmulation(int fd, T *dirp, uint32_t count);
+
+namespace FaultSafeMemcpy {
+// These are little helper functions for cases when FEX needs to copy data to or from the application in a robust fashion.
+// CopyFromUser and CopyToUser are memcpy routines that expect to safely SIGSEGV when reading or writing application memory respectively.
+// Returns zero if the memcpy completed, or EFAULT if the memcpy faulted.
```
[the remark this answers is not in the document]
Sonicadvance1: Updated.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
```diff
@@ -275,7 +280,11 @@ namespace FEX::HLE::x32 {
         sigsetsize);
 
       if (timeout_ts) {
-        *timeout_ts = tp64;
+        struct timespec32 timeout{};
+        timeout = tp64;
+
+        // If the copy fails to update the timeout, then this is safe to ignore the result.
```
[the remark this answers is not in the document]
Sonicadvance1: Updated comment.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
Sonicadvance1: I'll make this easier for everyone. I removed the tag. The test is supposed to pass. It'll correctly pass when we return EFAULT instead of crashing.

neobrain: I don't understand what disconnect we're having here. The main feature of the revised PR is to detect if system calls are called with invalid arguments. How is it possible we have tests that intentionally trigger this faulty behavior, yet our detection logic isn't making them *all of them* fail on 32-bit? Is the detection logic really working properly?

Sonicadvance1: We're likely talking past each other because of test bifurcation. 32-bit, 64-bit, efault handling returning efault or just aborting.
Fundamentally there are three tests
- poll
- ppoll
- ppoll_time64 (This one is duplicated on 64-bit since it maps to the ppoll syscall)

Each test broken down.
- poll test
   - This one doesn't really matter, we pass the arguments untouched to the kernel where it correctly returns EFAULT for us. It's just coverage of the poll family of syscalls.

- ppoll (32-bit)
   - The pollfds and sigset_t argument are passed to kernel where it handles EFAULT for us
   - The timespec argument does `CopyFromUser`
   - If EFAULT handling will return EFAULT or abort depending on the constexpr
      - Current code it will SIGABRT which kills the entire Catch2 test process
   - Once EFAULT handling inevitably gets switched over to returning EFAULT, this test can be removed from the Known_Failures

- ppoll (64-bit)
   - Everything is passed to the kernel directly. We don't need to handle shit.

- ppoll_time64 (32-bit only because the previously mentioned syscall aliasing)
  - Same as  `ppoll (64-bit)`, Every argument is passed directly to the kernel without userspace touching it.

As you can see, our detection logic can't make them all fail, because we are using the kernel's mechanisms to handle most of the arguments.

And yes the detection logic is working properly.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
```diff
@@ -0,0 +1,163 @@
...
+TEST_CASE("poll") {
+  // poll can return EFAULT if first argument is pointed to invalid pointer.
+  struct pollfd *invalid_fds = reinterpret_cast<struct pollfd *>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  auto ret = ::syscall(SYS_poll, invalid_fds, 1, 0);
+  REQUIRE(ret == -1);
+  CHECK(errno == EFAULT);
+}
+
+TEST_CASE("ppoll") {
+  // ppoll can return EFAULT for arguements 1, 3, 4
```
neobrain, line 19:
```suggestion
  // ppoll can return EFAULT for arguments 1, 3, 4
```

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
```diff
@@ -0,0 +1,163 @@
...
+  struct pollfd *invalid_fds = reinterpret_cast<struct pollfd *>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  struct timespec *invalid_timespec = reinterpret_cast<struct timespec*>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  sigset_t *invalid_sigset = reinterpret_cast<sigset_t*>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+
+  {
+    auto ret = ::syscall(SYS_ppoll, invalid_fds, 1, 0, nullptr, nullptr);
+    REQUIRE(ret == -1);
+    CHECK(errno == EFAULT);
+  }
```
neobrain, line 28: Since each of these blocks is an independent test, these are a good fit for SECTION:
```cpp
  SECTION("invalid FDs") {
    auto ret = ::syscall(SYS_ppoll, invalid_fds, 1, 0, nullptr, nullptr);
    REQUIRE(ret == -1);
    CHECK(errno == EFAULT);
  }
```

This will restart the test case for each SECTION, such that a failing REQUIRE won't cause the other sections to be skipped.

unittests/FEXLinuxTests/Known_Failures
```diff
@@ -19,3 +19,5 @@ sigtest_samask.32
 sigtest_samask.64
 sigtest_sigmask.32
 sigtest_sigmask.64
+
+syscalls_efault.32
```
neobrain, line 23: Needs a comment indicating that the corresponding emulation code is intentionally not enabled currently.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
neobrain: Thanks, that helps. The term "efault handling" was my main source of confusion (considering EFAULT isn't involved in FEX's new behavior at all), but with the breakdown I could understand what you're referring to.

With this explained, I would've been happy to keep the original `mayfail` tag, but it doesn't make a big difference where we declare the potential test failure.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
```diff
@@ -275,7 +280,12 @@ namespace FEX::HLE::x32 {
         sigsetsize);
 
       if (timeout_ts) {
-        *timeout_ts = tp64;
+        struct timespec32 timeout{};
+        timeout = tp64;
+
+        // We read the timeout correct, but now we can't write it. This can happen when then timeout is in a read-only memory segment.
```
neobrain, line 286: This comment explains exceptional behavior as if it were the common case. Normally, you *can* write the timeout.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
```diff
@@ -0,0 +1,163 @@
...
+#define SYS_ppoll_time64 SYS_ppoll
+#endif
+  struct timespec64 {
+    uint64_t tv_sec, tv_nsec;
+  };
+
+  // ppoll can return EFAULT for arguements 1, 3, 4
+  struct pollfd *invalid_fds = reinterpret_cast<struct pollfd *>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  timespec64 *invalid_timespec = reinterpret_cast<timespec64*>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
+  sigset_t *invalid_sigset = reinterpret_cast<sigset_t*>(mmap(nullptr, sysconf(_SC_PAGESIZE), PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0));
```
neobrain, line 99: What makes these invalid? Should these be `memset` to zero?

neobrain, line 99: I'm guessing `mmap` is used here (instead of `new`/`unique_ptr`) to allow `mprotect` later in the test? Seems like a quirk that deserves a short comment.

Alternatively, making `invalid_timespec` a global `const` variable will also place it in read-only memory and allows us to use `unique_ptr` for everything else.

Sonicadvance1: Protections make them invalid. Since it is PROT_NONE we can't memset, which is also fine since mmap is guaranteed to return zero'd out memory pages.

Sonicadvance1: mmap is used here to ensure we can fiddle with memory protections, which is why they all have PROT_NONE protections. And as you noted, so we can set a timeout and then set read-only for one of the tests.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
Sonicadvance1: Added sections

unittests/FEXLinuxTests/Known_Failures
Sonicadvance1: Added comment.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -531,6 +531,16 @@ static bool HasSyscallError(const void* Result) {
 
 template<bool IncrementOffset, typename T>
 uint64_t GetDentsEmulation(int fd, T *dirp, uint32_t count);
+
+namespace FaultSafeMemcpy {
+// These are little helper functions for cases when FEX needs to copy data to or from the application in a robust fashion.
+// CopyFromUser and CopyToUser are memcpy routines that expect to safely SIGSEGV when reading or writing application memory respectively.
+// Returns zero if the memcpy completed, or crashes with SIGABRT and a log message if it faults.
+size_t CopyFromUser(void *Dest, const void* Src, size_t Size);
+size_t CopyToUser(void *Dest, const void* Src, size_t Size);
```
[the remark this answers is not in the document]
Sonicadvance1: Added `[[nodiscard]]`

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
Sonicadvance1: moved comment.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
Sonicadvance1: Moved the read-only timespec value to a const global and double-checked that it was in a RO data section and ensuring that it still got captured as expected.

Sonicadvance1: Added a comment.

unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
```diff
@@ -0,0 +1,169 @@
...
+  .tv_sec = 1,
+  .tv_nsec = 0,
+};
+
+
+TEST_CASE("ppoll_64") {
+#ifndef SYS_ppoll_time64
+#define SYS_ppoll_time64 SYS_ppoll
+#endif
+  // ppoll can return EFAULT for arguements 1, 3, 4
```
neobrain, line 104: arguments

unittests/FEXLinuxTests/Known_Failures
neobrain: As mentioned in the other discussion, "EFAULT" is misleading here. Nothing in our current handling involves EFAULT, which makes linking this comment back to the actual piece of FEX code impossible.

A better wording could be "Disabled since FEX's FaultSafeMemcpy is intentionally stub-implemented".

neobrain (approved):

PR #3376, opened 2024-01-19
Improvements to the Dockerfile

Guztaver:
I added the `pkg-config` that is missing in the Dockerfile, and make the git clone the repository direct from the image, so the user don't need to clone it manually and after that pulls inside the Dockerfile!

If the git change is too much, tell me that I remove it and keep only the dependecies update, because without it, it's not compiling at all.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,12 @@ FROM ubuntu:20.04 as builder
 
 RUN DEBIAN_FRONTEND="noninteractive" apt-get update
 RUN DEBIAN_FRONTEND="noninteractive" apt install -y cmake \
-clang-10 llvm-10 nasm ninja-build \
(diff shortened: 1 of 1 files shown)


Guztaver:
I was thinking to add something to the `git clone` section, if goes toward. (I don't added it because a don't know if it will be included)

The changes are basicly add cache to the git, to not take too much time when compiling the Dockerfile mutiple times:

```Dockerfile
RUN --mount=type=cache,target=/opt/FEX \
  git clone --recurse-submodules https://github.com/FEX-Emu/FEX.git /opt/FEX; \
  cd /opt/FEX \ 
  git pull \

```

Sonicadvance1:
Seems reasonable to me. I'm surprised this docker file works at all since it is random user's whims to implement it.

Sonicadvance1 (approved):

PR #3378, opened 2024-01-22
Misc little opts

alyssarosenzweig:
trying to get back in the swing of things[](url)


FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -713,64 +713,55 @@ DEF_OP(PDep) {
...
-  cbz(EmitSize, Mask, &EarlyExit);
-  mov(EmitSize, IndexReg, ZeroReg);
-
-  // We sadly need to spill regs for this for the time being
-  // TODO: Remove when scratch registers can be allocated
-  //       explicitly.
-  SpillStaticRegs(TMP1, false, SpillCode);
 
+  // Handle early exit case
+  mov(EmitSize, Dest, ZeroReg);
```
Sonicadvance1, line 746: This has the issue that if the destination register overlaps either the input or mask then it corrupts the data before doing any work.

alyssarosenzweig: Ooooooooooooooooooooooooooof

alyssarosenzweig:
Mega PR has been split up for easier review

PR #3383, opened 2024-01-29
Optimize PTEST and VTESTP

alyssarosenzweig:
Use umaxv+tst to accelerate vector tests


lioncash (approved):
love seeing those compacts disappear

PR #3384, opened 2024-01-29
Optimize CDQOp

pmatos:
Optimize `cwde` and `cbw`.

Change OpcodeDispatcher CDQOp to load directly from GPR, and recognize instruction width.

Also change Constant Propagation to synthesize 32bits constant when necessary.

Update instcountci files.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -1395,11 +1395,12 @@ void OpDispatchBuilder::XCHGOp(OpcodeArgs) {
 }
 
 void OpDispatchBuilder::CDQOp(OpcodeArgs) {
-  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags);
   uint8_t DstSize = GetDstSize(Op);
   uint8_t SrcSize = DstSize >> 1;
+  OrderedNode *Src = LoadGPRRegister(X86State::REG_RAX, SrcSize, 0, true);
 
-  Src = _Sbfe(OpSize::i64Bit, SrcSize * 8, 0, Src);
+  Src = _Sbfe(DstSize <= 4 ? OpSize::i32Bit : OpSize::i64Bit, SrcSize * 8, 0,
+              Src);
 
   StoreResult_WithOpSize(GPRClass, Op, Op->Dest, Src, DstSize, -1);
 }
--- a/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -833,12 +833,17 @@ bool ConstProp::ConstantPropagation(IREmitter *IREmit, const IRListView& Current
       uint64_t Constant;
       if (IREmit->IsValueConstant(Op->Src, &Constant)) {
         // SBFE of a constant can be converted to a constant.
-        uint64_t SourceMask = Op->Width == 64 ? ~0ULL : ((1ULL << Op->Width) - 1);
+        uint64_t SourceMask =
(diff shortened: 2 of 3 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1395,11 +1395,12 @@ void OpDispatchBuilder::XCHGOp(OpcodeArgs) {
 }
 
 void OpDispatchBuilder::CDQOp(OpcodeArgs) {
-  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags);
   uint8_t DstSize = GetDstSize(Op);
   uint8_t SrcSize = DstSize >> 1;
+  OrderedNode *Src = LoadGPRRegister(X86State::REG_RAX, SrcSize, 0, true);
```
alyssarosenzweig, line 1400: Why did you change this? We typically use LoadSource

alyssarosenzweig: If you  want to eliminate the masking, what you want is `{.AllowUpperGarbage = true}` which opts into loading garbage values in the upper bits.

pmatos: > If you want to eliminate the masking, what you want is `{.AllowUpperGarbage = true}` which opts into loading garbage values in the upper bits. See #3381 as an example

But `LoadSource` takes `const LoadSourceOptions& Options`, while `LoadGPRRegister` takes a `bool AllowUpperGarbage`.

pmatos:
instcountci is failing - apparently it's using the old json file, not the one I just committed. Is there any procedure to properly submit a json file to change instcountci?

Sonicadvance1:
> instcountci is failing - apparently it's using the old json file, not the one I just committed. Is there any procedure to properly submit a json file to change instcountci?

Looks like in this case you didn't get the full json modified. Specifically the `ExpectedInstructionCount` fields of each should have been changed.
Doing a `ninja instcountci_tests` and then `ninja instcountci_update_tests` should change that?

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: Either route is okay to take. In a lot of cases for hardcoded register usage inside of an instruction we will use `LoadGPRRegister`. Since the cwd family of instructions hardcode their register usage to RAX then I think it is fine to use `LoadGPRRegister`.

If the instruction wasn't hardcoded to a register then `LoadSource` is definitely preferred if possible.

pmatos:
> > instcountci is failing - apparently it's using the old json file, not the one I just committed. Is there any procedure to properly submit a json file to change instcountci?
> 
> Looks like in this case you didn't get the full json modified. Specifically the `ExpectedInstructionCount` fields of each should have been changed. Doing a `ninja instcountci_tests` and then `ninja instcountci_update_tests` should change that?

Odd - you're right. instruction counts were not modified. Pretty sure I already did update_tests... I mean, that's how Primary.json got modified. Let me try that again.

pmatos:
> > > instcountci is failing - apparently it's using the old json file, not the one I just committed. Is there any procedure to properly submit a json file to change instcountci?
> > 
> > 
> > Looks like in this case you didn't get the full json modified. Specifically the `ExpectedInstructionCount` fields of each should have been changed. Doing a `ninja instcountci_tests` and then `ninja instcountci_update_tests` should change that?
> 
> Odd - you're right. instruction counts were not modified. Pretty sure I already did update_tests... I mean, that's how Primary.json got modified. Let me try that again.

Now it worked - maybe I ran things improperly first time around.

lioncash (approved):

PR #3387, opened 2024-01-29
Optimize rotates

alyssarosenzweig:
Lots of low-hanging optimizations around rotates, especially rotate-thru-carry.


lioncash (approved):
After looking this over a few times, it looks good!

Issue #3388, opened 2024-01-30
Fix steamwebhelper panic spill


Sonicadvance1:
As seen in #3389

Quite a large amount of panic spilling:
```
[DEBUG] Panic spilling %61, Live Range[61, 128)
[DEBUG] Panic spilling %185, Live Range[185, 246)
[DEBUG] Panic spilling %350, Live Range[350, 411)
[DEBUG] Panic spilling %467, Live Range[467, 528)
[DEBUG] Panic spilling %629, Live Range[629, 690)
[DEBUG] Panic spilling %746, Live Range[746, 807)
[DEBUG] Panic spilling %908, Live Range[908, 969)
[DEBUG] Panic spilling %1025, Live Range[1025, 1086)
[DEBUG] Panic spilling %1187, Live Range[1187, 1248)
[DEBUG] Panic spilling %1304, Live Range[1304, 1365)
[DEBUG] Panic spilling %1466, Live Range[1466, 1527)
[DEBUG] Panic spilling %1583, Live Range[1583, 1644)
[DEBUG] Panic spilling %1745, Live Range[1745, 1806)
[DEBUG] Panic spilling %1862, Live Range[1862, 1923)
[DEBUG] Panic spilling %2024, Live Range[2024, 2085)
[DEBUG] Panic spilling %2141, Live Range[2141, 2202)
[DEBUG] Panic spilling %2303, Live Range[2303, 2364)
[DEBUG] Panic spilling %2420, Live Range[2420, 2481)
[DEBUG] Panic spilling %2582, Live Range[2582, 2643)
[DEBUG] Panic spilling %2699, Live Range[2699, 2760)
[DEBUG] Panic spilling %2861, Live Range[2861, 2922)
[DEBUG] Panic spilling %2978, Live Range[2978, 3039)
[DEBUG] Panic spilling %3140, Live Range[3140, 3201)
[DEBUG] Panic spilling %3257, Live Range[3257, 3318)
[DEBUG] Panic spilling %3429, Live Range[3429, 3490)
[DEBUG] Panic spilling %3546, Live Range[3546, 3607)
[DEBUG] Panic spilling %3662, Live Range[3662, 3723)
[DEBUG] Panic spilling %3779, Live Range[3779, 3840)
[DEBUG] Panic spilling %3895, Live Range[3895, 3956)
[DEBUG] Panic spilling %4012, Live Range[4012, 4073)
[DEBUG] Panic spilling %4128, Live Range[4128, 4189)
[...]

PR #3389, opened 2024-01-30
instcountci: Adds panicspill test from steamwebhelper

Sonicadvance1:
This block of code causes our RA to panic spill and it gets hit on steamwebhelper.
We should be able to spill without panic on this code at some point.

#3388 open to track the panic spill in this code.


alyssarosenzweig:
This isn't really a unit test. Any chance we could get this in InstCountCI instead? :smile_cat:

Sonicadvance1:
> This isn't really a unit test. Any chance we could get this in InstCountCI instead? 😸

Moved to instcountci

alyssarosenzweig:
Oof that's a lot of asm. but yeah thanks :+1:

alyssarosenzweig (approved):

unittests/InstructionCountCI/FEXOpt/libnss.json
```diff
@@ -0,0 +1,4532 @@
+{
+  "Features": {
+    "Bitness": 64,
+    "EnabledHostFeatures": [
+      "CRYPTO"
```
alyssarosenzweig, line 5: explicitly enable flagm? rmif is used

Sonicadvance1: Explicitly enabled.

lioncash (approved):

PR #3390, opened 2024-01-30
Check that path arguments to TestHarnessRunner exist

pmatos:
Previously missing files would result in a segmentation fault.

### DIFF
--- a/Source/Tools/TestHarnessRunner/TestHarnessRunner.cpp
+++ b/Source/Tools/TestHarnessRunner/TestHarnessRunner.cpp
@@ -18,22 +18,23 @@ $end_info$
 #endif
 
 #include "Common/ArgumentLoader.h"
-#include "TestHarnessRunner/HostRunner.h"
 #include "HarnessHelpers.h"
+#include "TestHarnessRunner/HostRunner.h"
 
 #include <FEXCore/Config/Config.h>
+#include <FEXCore/Core/CPUBackend.h>
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Core/CoreState.h>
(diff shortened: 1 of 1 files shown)


pmatos (commented):
I spent some good 5 minutes wondering why I was getting segfaults on a `memcpy` initializing the configs... until I noticed my paths didn't exist. :)

Source/Tools/TestHarnessRunner/TestHarnessRunner.cpp
```diff
@@ -198,13 +198,25 @@ int main(int argc, char **argv, char **const envp) {
...
+  auto Filename = Args[0];
+  auto ConfigFile = Args[1];
 
   if (Args.size() < 2) {
     LogMan::Msg::EFmt("Not enough arguments");
     return -1;
   }
 
-  FEX::HarnessHelper::HarnessCodeLoader Loader{Args[0], Args[1]};
+  if (!std::filesystem::exists(Filename)) {
```
Sonicadvance1, line 209: These cause the glibc fault CI test to fail because of these concerns here: https://github.com/FEX-Emu/FEX/blob/main/docs/ProgrammingConcerns.md#stdfilesystemexists

Easy enough switch will be to switch both use cases over to `FHU::Filesystem::Exists`

pmatos: TIL - following your suggestion.

pmatos:
@Sonicadvance1 Thanks - I have squashed the changes. Can you pls merge?

Sonicadvance1 (approved):
Will do, thanks for the quality of life improvement there.

PR #3392, opened 2024-01-30
Library Forwarding: Handle cross-architecture differences of integer types

neobrain:
## Preface

Of all the library forwarding PRs, this has been the most tricky one to implement due to the myriad edge cases that must be considered. Some trade-offs had to be made to progress at the expense of not having a perfectly sound implementation, but overall I'm happy this turned out reasonably clean.

## Overview

Integer types can be surprisingly inconsistent between guest and host architectures:
* A given type may be differently sized on the guest than on the host (e.g. `long` or `size_t` on 32-bit)
* libclang may choose slightly different "canonical" types (e.g. `long` vs `long long` on 64-bit)
* C's silly type system hacks get in the way (e.g. `char` is a different type than `unsigned char` and `signed char`)
* All of the above issues also translate to *pointers of the involved types* (e.g. `wchar_t*` and `uint32_t*` must be detected as having the same data layout)

To address these problems, this PR changes `thunkgen` to map *all* guest integer types to fixed-size integers. This means we must selectively soften type-safety when converting between `guest_layout` and `host_layout`. Also, function pointer support must be adapted to be data-layout-aware.

## Implementation details: `host_to_guest_convertible`

Converting function return values from host layout to guest layout gets surprisingly tricky: For example, consider the functions `uint64_t MakeU64()` and `size_t MakeST()`. On the 64-bit host, these both return a 64-bit value. However, the value type that must be returned to a 32-bit guest differs: `MakeU64` still returns a 64-bit value, but `MakeST` returns a 32-bit value. We can't use `guest_layout<size_t>` to resolve this difference, since that's compiled in a host context (where `size_t` itself is just a typedef for `uint64_t`).

Instead, a new abstraction is added to act as an intermediary for return values: `host_to_guest_convertible<size_t>` wraps the `host_layout<size_t>` and provides *implicit conversion operators* to either `guest_layout<uint32_t>` or `guest_layout<uint64_t>`. This way, we can selectively enable specific conversion paths without sacrificing type-safety entirely.

## Implementation details: Making function pointers data-layout-aware

Since types like `size_t` (on 32-bit) have different sizes on the guest than on the host, `GuestWrapperForHostFunction` must be able to tell what bitness each of the function parameters has both on the host and on the guest. This information cannot be inferred from the parameter list, since the C++ type system doesn't distinguish between `size_t` (as written) and `uint64_t`.

The chosen solution is to add a second parameter pack to `GuestWrapperForHostFunction`: This pack lists all the types *as seen by the guest* but using fixed-size types. The existing `guest_layout`/`host_layout`/`repack_wrapper` helpers then take care of pairwise conversion of the guest argument type to the host argument type. 

## Impact

Issues in the new code could have a performance impact by adding unintended repacking between integer types. In practice, `repack_wrapper` is still not used in any of our forwarded libraries on 64-bit outside of test code, so performance should be unchanged.

## Future

This PR concludes the struct repacking work. A few minor PRs are needed to start enabling 32-bit Wayland thunks.


neobrain:
CI fixed & now ready for review.

Clang <16.0 had troubles building this PR previously due to incomplete support for C++20 concepts. I changed the code slightly to avoid the shortcomings of these older versions.

Sonicadvance1 (approved):
Spicy

PR #3393, opened 2024-01-31
InstCountCI: Fixes test to not use relative data

Sonicadvance1:
JIT code and loaded code memory offsets aren't consistent run to run which causes code generation consistency issues.
Instead use a register plus an immediate that can't fit in a single 16-bit encoding to get similar register allocation to RIP relative addresses.

At some point it will probably be nice to support consistent placement of both guest and host code so RIP relative can be used, but we don't have support in our backend to do that.

@alyssarosenzweig This will fix your consistency woes.


alyssarosenzweig (approved):
thanks :)

PR #3394, opened 2024-01-31
Add cmake option DISABLE_CLANG_PRESERVE_ALL

pmatos:
Forces disabling use of __attribute__((preserve_all)). Until CI uses clang17, where this attribute was added, instcountci fails when FEX is compiled with clang>=17.


Sonicadvance1 (approved):
Nice

PR #3397, opened 2024-02-01
Improve XCHG operations

pmatos:
Update InstCountCI

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -1367,14 +1367,17 @@ void OpDispatchBuilder::XCHGOp(OpcodeArgs) {
     // So x86-64 spec mandates this special case that even though it is a 32bit instruction and
     // is supposed to zext the result, it is a true no-op
     if (Op->Flags & FEXCore::X86Tables::DecodeFlags::FLAG_REP_PREFIX) {
-      // If this instruction has a REP prefix then this is architectually defined to be a `PAUSE` instruction.
-      // On older processors this ends up being a true `REP NOP` which is why they stuck this here.
+      // If this instruction has a REP prefix then this is architecturally
+      // defined to be a `PAUSE` instruction. On older processors this ends up
+      // being a true `REP NOP` which is why they stuck this here.
       _Yield();
     }
     return;
   }
 
-  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags);
+  // AllowUpperGarbage: OK to allow as it will be overwritten by StoreResult.
+  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags,
+                                {.AllowUpperGarbage = true});
   if (DestIsMem(Op)) {
     HandledLock = Op->Flags & FEXCore::X86Tables::DecodeFlags::FLAG_LOCK;
     OrderedNode *Dest = LoadSource(GPRClass, Op, Op->Dest, Op->Flags, {.LoadData = false});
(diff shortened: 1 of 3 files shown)


pmatos:
Still in draft mode - trying to frustratingly remove the second of the move in 
```
        "mov x20, x5",
        "mov x5, x20",
        "bfxil x5, x7, #0, #8",
        "bfxil x7, x20, #0, #8"
```

Basically, I would have expected DCE pass to remove this but it doesn't. It sounds like something it should do, given that x5 is rewritten straightaway.

alyssarosenzweig:
> Still in draft mode - trying to frustratingly remove the second of the move in
> 
> ```
>         "mov x20, x5",
>         "mov x5, x20",
>         "bfxil x5, x7, #0, #8",
>         "bfxil x7, x20, #0, #8"
> ```
> 
> Basically, I would have expected DCE pass to remove this but it doesn't. It sounds like something it should do, given that x5 is rewritten straightaway.

bfxil does a read-modify-write on its destination, it's a hidden source here. DCE is doing its job, the actual issue is that RA doesn't yet do affinities for bfi/bfxil/etc. This is planned for a future rewrite of RA, the current RA is a :wastebasket: :fire: and I don't want to touch the jenga tower before the big rewrite.

pmatos:
> > Still in draft mode - trying to frustratingly remove the second of the move in
> > ```
> >         "mov x20, x5",
> >         "mov x5, x20",
> >         "bfxil x5, x7, #0, #8",
> >         "bfxil x7, x20, #0, #8"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Basically, I would have expected DCE pass to remove this but it doesn't. It sounds like something it should do, given that x5 is rewritten straightaway.
> 
> bfxil does a read-modify-write on its destination, it's a hidden source here. DCE is doing its job, the actual issue is that RA doesn't yet do affinities for bfi/bfxil/etc. This is planned for a future rewrite of RA, the current RA is a 🗑️ 🔥 and I don't want to touch the jenga tower before the big rewrite.

Oh no... :( I missed the fact that the destination is a hidden source. Not much I can do here atm then, I think. I will leave this for now.

What's the plan regarding the big RA rewrite?

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1367,14 +1367,16 @@ void OpDispatchBuilder::XCHGOp(OpcodeArgs) {
...
+      // defined to be a `PAUSE` instruction. On older processors this ends up
+      // being a true `REP NOP` which is why they stuck this here.
       _Yield();
     }
     return;
   }
 
-  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags);
+  OrderedNode *Src = LoadSource(GPRClass, Op, Op->Src[0], Op->Flags,
+                                {.AllowUpperGarbage = true});
```
alyssarosenzweig, line 1379: Usually I try to justify with a comment or assert why garbage is ok ("it will be masked away by the following And" or whatever)

pmatos: Very good point!

alyssarosenzweig (approved):

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
alyssarosenzweig: The correctness of this patch relies on `swpalb` ignoring the upper 24-bits of its source, etc. (Carefully look at the instcountci diff :smile: )

lioncash (approved):

PR #3400, opened 2024-02-02
Revert #3303

Sonicadvance1:
Sadly this breaks running proton applications for no benefit currently.
It's hard to figure out why it breaks, so revert for now and we can bring it back later.

Test case for this is running Sonic Mania and watching the process crash before this revert.


lioncash (approved):
Unfortunate. But yeah, this is quite a lot of code to sift through, so reverting for now definitely seems best to keep things moving.

alyssarosenzweig (approved):

PR #3401, opened 2024-02-02
HostFeatures: Supports runtime disabling of preserve_all

Sonicadvance1:
This is used for instcountci to ensure instruction counts don't change when a compiler supports this feature or not. Always runtime disable when running in instcountci.

CMake option from #3394 can still be useful so leaving that in place.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -77,7 +77,9 @@
           "ENABLECRYPTO": "enablecrypto",
           "DISABLECRYPTO": "disablecrypto",
           "ENABLERPRES": "enablerpres",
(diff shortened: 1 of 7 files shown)


neobrain:
What use cases does #3394 serve that this doesn't? Do we really need both?

Our CMakeLists.txt is already exploding as far as number of possible configurations go, so any unneeded complexity we can avoid on top of that would be appreciated.

alyssarosenzweig:
> What use cases does #3394 serve that this doesn't? Do we really need both?
> 
> Our CMakeLists.txt is already exploding as far as number of possible configurations go, so any unneeded complexity we can avoid on top of that would be appreciated.

I'd be in favour of reverting #3394

pmatos:
I agree we can revert 3394.

PR #3409, opened 2024-02-05
Revert "Add cmake option DISABLE_CLANG_PRESERVE_ALL"

neobrain:
Reverts FEX-Emu/FEX#3394. A runtime-check for this was added in #3401.


Sonicadvance1 (approved):
Probably good to not create more branches on the upstream repo though.

PR #3408, opened 2024-02-05
Optimize TST

alyssarosenzweig:
A few commits I had lying around on random branches, longing for merge.

PR #3410, opened 2024-02-05
Add NZCV+PF/AF optimization pass

alyssarosenzweig:
Finally felt good enough to type this guy up. This addresses the Sonic silly block and lays the foundation for even more juicy NZCV work. On top of #3408 , that one is smol and will land first

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
@@ -157,7 +157,6 @@ DeadFlagCalculationEliminination::Classify(IROp_Header *IROp)
       return {.Read = FLAG_NZCV};
 
     case OP_RMIFNZCV: // TODO: Optimize?
-    case OP_CONDADDNZCV:
       return {
         .Read = FLAG_NZCV,
         .Write = FLAG_NZCV,
@@ -183,6 +182,15 @@ DeadFlagCalculationEliminination::Classify(IROp_Header *IROp)
         break;
 
       return {.Read = FlagsForCondClassType(Op->Cond)};
+    }
+
+    case OP_CONDADDNZCV: {
+      auto Op = IROp->CW<IR::IROp_CondAddNZCV>();
+      return {
+        .Read = FlagsForCondClassType(Op->Cond),
+        .Write = FLAG_NZCV,
+        .CanEliminate = true,
+      };
     }
 
     case OP_INVALIDATEFLAGS: {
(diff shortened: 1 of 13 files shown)


FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -17,54 +17,203 @@ desc: This is not used right now, possibly broken
...
+
 class DeadFlagCalculationEliminination final : public FEXCore::IR::Pass {
 public:
   bool Run(IREmitter *IREmit) override;
+
+private:
+  struct FlagInfo Classify(IROp_Header *Node);
 };
 
+struct FlagInfo
```
lioncash, line 60:
```suggestion
FlagInfo
```

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -17,54 +17,203 @@ desc: This is not used right now, possibly broken
...
+DeadFlagCalculationEliminination::Classify(IROp_Header *IROp)
+{
+  switch (IROp->Op) {
+    case OP_ANDWITHFLAGS:
+      return (struct FlagInfo){
```
lioncash, line 65: The tag name can be omitted:

```suggestion
      return {
```

Ditto for the other returns

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -17,54 +17,203 @@ desc: This is not used right now, possibly broken
...
+  // Replacement if the flag writes can all be eliminated.
+  IROps Replacement;
+};
+
 class DeadFlagCalculationEliminination final : public FEXCore::IR::Pass {
 public:
   bool Run(IREmitter *IREmit) override;
+
+private:
+  struct FlagInfo Classify(IROp_Header *Node);
```
lioncash, line 57:
```suggestion
  FlagInfo Classify(IROp_Header *Node);
```

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -17,54 +17,203 @@ desc: This is not used right now, possibly broken
 #include <array>
 #include <memory>
 
+// Flag bit flags
+#define FLAG_N (1 << 0)
```
lioncash, line 21:
```suggestion
#define FLAG_N (1U << 0)
```

Maybe, just to enforce these as unsigned, since they're bit flags

lioncash (commented):
Just a few minor nits for a first pass

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
alyssarosenzweig: Neat

alyssarosenzweig:
I think there's enough here to prove the foundations are good, so I think this is ready for review. Time for me to do a latency-hiding context switch :grin:

Sonicadvance1 (approved):
Looks good. I also threw the change at a bunch of games and haven't noticed a regression.

PR #3412, opened 2024-02-06
Clang Format file proposal

pmatos:
This is a proposal to add a `.clang-format` file to be used as the basis for a code formatting standard inside FEX.

This PR is meant for us to discuss if we are open to a `.clang-format` and how to adopt it. 
If I run [clang-unformat](https://github.com/alandefreitas/clang-unformat) on different files and different directories inside FEX, I get different configuration files since there is a number of coding guidelines being used.

Therefore I took the two top level directories: `Source`, and `FEXCore` and created for each, a clang-format file that minimizes the amount of necessary changes. The version included in this PR, is the version generated by `clang-unformat` for the FEXCore directory.

I also created two branches that format the whole FEX tree by each of these files. The branches are:

* Format according to FEXCore: https://github.com/pmatos/FEX/tree/ClangFormatFEXCore
* Format according to Source: https://github.com/pmatos/FEX/tree/ClangFormatSource

My initial feeling is, whatever `.clang-format` is chosen, we could just use it to format the snippets we edit without doing a full tree format and slowly make it the adopted code formatting convention. The nuclear option is to reformat the whole tree and add the commit to a `.git-blame-ignore-revs` file (see for the example the one for LLVM [here](https://github.com/llvm/llvm-project/blob/main/.git-blame-ignore-revs)).

If you have suggestions about a different convention, and would like to see it applied to the tree or a specific file, let me know and I can generate that for you if you wish.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,109 @@
+Language: Cpp
+BasedOnStyle: WebKit
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: Consecutive
+AlignConsecutiveDeclarations: None
+AlignConsecutiveMacros: None
+AlignEscapedNewlines: DontAlign
+AlignOperands: DontAlign
+AlignTrailingComments: false
+AllowShortCaseLabelsOnASingleLine: true
+AllowShortEnumsOnASingleLine: true
+AllowShortFunctionsOnASingleLine: All
+AllowShortIfStatementsOnASingleLine: WithoutElse
+AllowShortLambdasOnASingleLine: Inline
+AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: No
+BinPackArguments: true
+BinPackParameters: true
+BitFieldColonSpacing: Both
+BreakBeforeBraces: Custom
+BraceWrapping:
+  AfterCaseLabel: false
+  AfterClass: false
+  AfterControlStatement: Never
+  AfterEnum: false
+  AfterFunction: false
+  AfterNamespace: false
+  AfterStruct: false
+  AfterUnion: false
+  AfterExternBlock: false
+  BeforeElse: false
+  BeforeLambdaBody: true
+  BeforeWhile: false
+  IndentBraces: false
+BreakBeforeBinaryOperators: None
+BreakBeforeTernaryOperators: false
+BreakConstructorInitializers: BeforeComma
+BreakInheritanceList: BeforeColon
+ColumnLimit: 140
+CompactNamespaces: false
+ConstructorInitializerIndentWidth: 2
+ContinuationIndentWidth: 0
+Cpp11BracedListStyle: true
+DeriveLineEnding: true
+DerivePointerAlignment: true
+EmptyLineAfterAccessModifier: Leave
+EmptyLineBeforeAccessModifier: Leave
+ExperimentalAutoDetectBinPacking: false
+FixNamespaceComments: false
+IncludeBlocks: Preserve
+IndentAccessModifiers: false
+IndentCaseBlocks: false
+IndentCaseLabels: false
+IndentExternBlock: AfterExternBlock
+IndentGotoLabels: false
+IndentPPDirectives: None
+IndentRequires: false
+IndentWidth: 2
+KeepEmptyLinesAtTheStartOfBlocks: true
+LambdaBodyIndentation: OuterScope
+MaxEmptyLinesToKeep: 2
+NamespaceIndentation: Inner
+PackConstructorInitializers: Never
+PenaltyBreakAssignment: 2
+PenaltyBreakBeforeFirstCallParameter: 2
+PenaltyBreakOpenParenthesis: 2
+PenaltyBreakTemplateDeclaration: 8
+PenaltyExcessCharacter: 2
+PenaltyIndentedWhitespace: 1024
+PenaltyReturnTypeOnItsOwnLine: 16
+PointerAlignment: Left
+QualifierAlignment: Leave
+RemoveBracesLLVM: false
+ReflowComments: false
+RequiresClausePosition: WithPreceding
+SeparateDefinitionBlocks: Leave
+SortIncludes: Never
+SpaceAfterCStyleCast: false
+SpaceAfterLogicalNot: false
+SpaceAfterTemplateKeyword: false
+SpaceAroundPointerQualifiers: Default
+SpaceBeforeAssignmentOperators: true
+SpaceBeforeCaseColon: false
+SpaceBeforeCpp11BracedList: false
+SpaceBeforeInheritanceColon: true
+SpaceBeforeParens: Custom
+SpaceBeforeParensOptions:
+  AfterControlStatements: true
+  AfterFunctionDeclarationName: false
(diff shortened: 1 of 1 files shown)


alyssarosenzweig:
I'm in favour of a single bulk reformat with a single treewide style, if we can find a style we like. git-blame-ignore-revs helps a lot here. There are ways to rebase in-flight branches if we need.

Enforcing with CI would probably be ok for FEX, it's not viable with Mesa but I think we can get away with it here.

Ideally you configure your editor to format automatically on save so you can just type utter slop and get good formatted code out :smile_cat:

lioncash:
Very nice that this is being done . My only suggestion before doing the final formatting pass, would be placing

```cpp
// clang-format off
```
and
```cpp
// clang-format on
```

around most of the tables in the X86Tables directory, so that it doesn't get potentially mangled (either now, or in the future if the format file gets changed), so it stays mostly tabular

pmatos:
Thanks for all your comments.

Firstly, I tried to find a style + options that match more or less what we have been doing. If you want a single style, both WebKit and LLVM seem to be the closest. I have been using the LLVM style for quite a few years now in LLVM so my opinion is biased. Also, if we use a single style we don't need to be bothered by `clang-unformat`.

I think if we go for a full tree rebuild we can mark the revision where we reformat as to be ignored through `.git-blame-ignore-revs`.

Regarding process, we can use format on save on our editors. I currently use format modified code on save - instead of whole file.
Then we could have a PR CI action to check that the modified code was formatted. There's something from LLVM that we can use. The PR author would be pinged if he had forgotten to format the code appropriately. See the action [here](https://github.com/llvm/llvm-project/blob/main/.github/workflows/pr-code-format.yml), generating a comment like [this](https://github.com/llvm/llvm-project/pull/80908#issuecomment-1930831187).

What do you think on this process?

pmatos:
> Great work!
> 
> We probably want a round or two of tweaking `clang-unformat`'s output later, but I agree with the general course of action you've outlined. However, I would prefer we bite the bullet and reformat everything to a _single_ style instead of using multiple different format definitions. I don't have strong opinions on whether to do this all-in-one-go or more incrementally, as long as we don't end up with 50 reformatting PRs.
> 
> What are your thoughts on enforcing formatting longer-term? Should we just regularly run clang-format manually and create dedicated "update formatting" PRs, or would you enforce it automatically instead? I'm a bit wary of the latter approach if it's based on commit hooks (since they disturb local, iterative development), but something that's part of CI could work.

I assumed that when you say `a single style instead of using multiple different format definitions`, you mean a single style like:

```
BasedOnStyle: LLVM
```

for example rather than having a more complex format. The other possibly reading is to have a different style per FEX sub-tree but I never meant to suggest that. 

My two approaches to clang-unformat was to use two different sub-tree indeed and see which style is suggested but my goal was always to have a single style for the whole FEX tree. I was just trying to give two options.

neobrain:
> My two approaches to clang-unformat was to use two different sub-tree indeed and see which style is suggested but my goal was always to have a single style for the whole FEX tree. I was just trying to give two options.

Ah yes, I misunderstood this part of the proposal then. I thought you were suggesting a different .clang-format file for each subtree, and I didn't mean to counterpropose sticking to a vanilla base style. Using a single .clang-format file with a few customizations for the entire project is fine of course.

Either of the 2 styles in your branches seem fine to me as a base, though as mentioned before we should have a dedicated pass over this to tweak the specific rules. For example, I'd be happy to drop a couple of the 100 customized individual settings at the expense of producing a slightly bigger reformat, assuming the end result still looks acceptable of course.

> Regarding process, we can use format on save on our editors. I currently use format modified code on save - instead of whole file.
> Then we could have a PR CI action to check that the modified code was formatted. T[here](https://github.com/llvm/llvm-project/blob/main/.github/workflows/pr-code-format.yml?rgh-link-date=2024-02-07T08%3A56%3A31Z)'s something from LLVM that we can use. The PR author would be pinged if he had forgotten to format the code appropriately. See the action here, generating a comment like https://github.com/llvm/llvm-project/pull/80908#issuecomment-1930831187.
>
> What do you think on this process?

Sounds good. Relying on format-on-save is effective but also unreliable, so having the CI enforcement on top (like these LLVM checks) would definitely be appreciated.

pmatos:
> > My two approaches to clang-unformat was to use two different sub-tree indeed and see which style is suggested but my goal was always to have a single style for the whole FEX tree. I was just trying to give two options.
> 
> Ah yes, I misunderstood this part of the proposal then. I thought you were suggesting a different .clang-format file for each subtree, and I didn't mean to counterpropose sticking to a vanilla base style. Using a single .clang-format file with a few customizations for the entire project is fine of course.
> 
> Either of the 2 styles in your branches seem fine to me as a base, though as mentioned before we should have a dedicated pass over this to tweak the specific rules. For example, I'd be happy to drop a couple of the 100 customized individual settings at the expense of producing a slightly bigger reformat, assuming the end result still looks acceptable of course.

Re-reading my original post I can understand how it can be confusing. Apologies about that.

> 
> > Regarding process, we can use format on save on our editors. I currently use format modified code on save - instead of whole file.
> > Then we could have a PR CI action to check that the modified code was formatted. T[here](https://github.com/llvm/llvm-project/blob/main/.github/workflows/pr-code-format.yml?rgh-link-date=2024-02-07T08%3A56%3A31Z)'s something from LLVM that we can use. The PR author would be pinged if he had forgotten to format the code appropriately. See the action here, generating a comment like [llvm/llvm-project#80908 (comment)](https://github.com/llvm/llvm-project/pull/80908#issuecomment-1930831187).
> > What do you think on this process?
> 
> Sounds good. Relying on format-on-save is effective but also unreliable, so having the CI enforcement on top (like these LLVM checks) would definitely be appreciated.

Yes, I think with the addition of the clang-format file commit and the reformat whole tree commit, I will add a commit to add appropriate PR CI.

Next step is to look at the tree, for example here:

https://github.com/pmatos/FEX/tree/ClangFormatFEXCore

and see if there's anything we would like to change. I think one way would be for everyone to pick a favourite file and point out what they _do not_ like about the reformatting. If you pick `OpcodeDispatcher.cpp`, you'd be looking at https://github.com/pmatos/FEX/commit/af5bf9712a997af4bec5221328fb81aa4878489d#diff-69995770a39cb5462186c29b1864b038c570ea5db22218cbbb566c82373f7578 .

As @lioncash pointed out we might have to safeguard a few things to keep a table look to the code. If you see any of those besides the already mentioned X86Tables stuff, feel free to drop them here as well and I will start preparing a commit with the reformat and clang-format file.

neobrain:
> Next step is to look at the tree, for example here:
>
> [pmatos/FEX@ClangFormatFEXCore](https://github.com/pmatos/FEX/tree/ClangFormatFEXCore?rgh-link-date=2024-02-07T10%3A44%3A37Z)
> 
> and see if there's anything we would like to change

Maybe it's easier to open a separate PR for these branches? That way we can isolate the detail-oriented discussion from higher-level planning, and we can just comment on individual lines instead of copy-pasting the context here.

pmatos:
> > Next step is to look at the tree, for example here:
> > [pmatos/FEX@ClangFormatFEXCore](https://github.com/pmatos/FEX/tree/ClangFormatFEXCore?rgh-link-date=2024-02-07T10%3A44%3A37Z)
> > and see if there's anything we would like to change
> 
> Maybe it's easier to open a separate PR for these branches? That way we can isolate the detail-oriented discussion from higher-level planning, and we can just comment on individual lines instead of copy-pasting the context here.

Good point.

pmatos:
Closing this one as we don't need it any more. Work continues in other PRs - main one being #3413.

PR #3414, opened 2024-02-08
Fixes one mutex hang

Sonicadvance1:
When code invalidation is happening we currently have the issue that a thread can acquire the code invalidation mutex in the middle of invalidation. This is due to us acquiring and releasing the mutex between each thread's code invalidation.

We need to hold the mutex for the entire duration for all thread's code invalidation.
This fixes a rare hang on proton startup and resolves a consistent hang on Proton application shutdown.

This now puts us on par with FEX-2312.1 with hanging.

This does not fix a relatively rare hang on fork (which also existed with FEX-2312.1).

This also does not fix the issue that the intersection of our mutexes between frontend and backend are very convoluted. In part of the work that is going to fix the rare fork mutex hang will change more of this.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -174,6 +174,10 @@ namespace FEXCore::Context {
       void ClearCodeCache(FEXCore::Core::InternalThreadState *Thread) override;
       void InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length) override;
       void InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length, CodeRangeInvalidationFn callback) override;
+      FEXCore::ForkableSharedMutex &GetCodeInvalidationMutex() override {
+        return CodeInvalidationMutex;
+      }
+
       void MarkMemoryShared(FEXCore::Core::InternalThreadState *Thread) override;
 
       void ConfigureAOTGen(FEXCore::Core::InternalThreadState *Thread, fextl::set<uint64_t> *ExternalBranches, uint64_t SectionMaxAddress) override;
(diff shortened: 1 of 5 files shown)


neobrain:
It would be great if when submitting fixes, you could more clearly highlight whether the underlying bugs were only recently introduced or whether they've persisted for a long time. This comes up from time to time and it's making it hard to recognize truly forward progress.

> This now puts us on par with FEX-2312.1 with hanging.

Does this mean this is another regression in addition to #3399? What PR does this one trace back to?

> This does not fix a relatively rare hang on fork (which also existed with FEX-2312.1).

Is this also a regression? Any idea where it's coming from yet?

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -174,6 +174,10 @@ namespace FEXCore::Context {
       void ClearCodeCache(FEXCore::Core::InternalThreadState *Thread) override;
       void InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length) override;
       void InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState *Thread, uint64_t Start, uint64_t Length, CodeRangeInvalidationFn callback) override;
+      FEXCore::ForkableSharedMutex *GetCodeInvalidationMutex() override {
```
neobrain, line 177: There is no ownership transfer here and the returned pointer is unconditionally assumed to be non-null, so returning a reference here is a better fit.

neobrain (commented):
Looks reasonable. Since you expressed discontent about passing backend-state to the frontend, do you have any plans to refactor for less state leakage?

Sonicadvance1:
> Looks reasonable. Since you expressed discontent about passing backend-state to the frontend, do you have any plans to refactor for less state leakage?

Yea, there are plans for this. I just need to get the whole model of these three interacting mutexes in my head. Potentially with the single recursive one playing a role in FEXCore as well.
Ideally the frontend is the only thing that cares about the global code invalidation mutex and FEXCore can only care about its thread local mutexes, but I don't know yet if it will allow that clean of an interface.

FEXCore/Source/Interface/Context/Context.h
Sonicadvance1: Switched over to a reference

neobrain (approved):

PR #3415, opened 2024-02-08
Optimize test -1

alyssarosenzweig:


Sonicadvance1 (approved):
Probably reasonable in a future PR to add more immediate encoded operations to instcountci in order to ensure we hit the bitfield encoded ARM instructions in all instances that we can.

PR #3416, opened 2024-02-08
Optimize 8/16-bit adds/subs

alyssarosenzweig:
If we shift first, NZCV matches up exactly :+1:


Sonicadvance1 (approved):
Nice. Almost all the instructions affected have a big win, only a few that are worse off.

PR #3424, opened 2024-02-14
Linux: More safe stack cleanup for clone

Sonicadvance1:
Previously: Would keep one clone thread's stack active for teardown delaying.

With aggressive cloning and teardown, this was unsafe. Only reap the stack when told it is safe to do so.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
@@ -394,21 +394,26 @@ struct StackFramePlusRet {
 };
 
 [[noreturn]]
+static void CloneBody(StackFrameData *Data, bool NeedsDataFree) {
+  uint64_t Result = FEX::HLE::HandleNewClone(Data->Thread, Data->CTX, &Data->NewFrame, &Data->GuestArgs);
+  auto Stack = Data->GuestArgs.NewStack;
+  if (NeedsDataFree) {
+    FEXCore::Allocator::free(Data);
+  }
+
+  FEX::LinuxEmulation::Threads::DeallocateStackObjectAndExit(Stack, Result);
+  FEX_UNREACHABLE;
+}
+
+[[noreturn]]
 static void Clone3HandlerRet() {
   StackFrameData *Data = (StackFrameData*)alloca(0);
-  uint64_t Result = FEX::HLE::HandleNewClone(Data->Thread, Data->CTX, &Data->NewFrame, &Data->GuestArgs);
-  FEX::LinuxEmulation::Threads::DeallocateStackObject(Data->GuestArgs.NewStack);
-  // To behave like a real clone, we now just need to call exit here
-  exit(Result);
-  FEX_UNREACHABLE;
+  CloneBody(Data, false);
 }
 
 static int Clone2HandlerRet(void *arg) {
   StackFrameData *Data = (StackFrameData*)arg;
-  uint64_t Result = FEX::HLE::HandleNewClone(Data->Thread, Data->CTX, &Data->NewFrame, &Data->GuestArgs);
-  FEX::LinuxEmulation::Threads::DeallocateStackObject(Data->GuestArgs.NewStack);
-  FEXCore::Allocator::free(arg);
-  return Result;
+  CloneBody(Data, true);
 }
 
 // Clone3 flags
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 #include "LinuxSyscalls/Utils/Threads.h"
+#include "LinuxSyscalls/Syscalls.h"
 
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Utils/Threads.h>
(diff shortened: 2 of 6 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
```diff
@@ -393,21 +393,43 @@ struct StackFramePlusRet {
...
+    , [Result] "r" (Result) \
+    : "memory", "x0", "x8");
+#else
+#define ExitWithoutStack(Result) \
+  __asm volatile( \
+    "mov %[Result], %%rdi;" \
+    "syscall;" \
+    :: "a" (SYSCALL_DEF(exit)) \
+    , [Result] "r" (Result) \
+    : "memory", "rdi");
```
neobrain, line 412: These can be `[[noreturn]]` functions instead, can't they?

Sonicadvance1: They cannot since debug builds will try to use the stack and we can't guarantee tail-call optimization.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
```diff
@@ -393,21 +393,43 @@ struct StackFramePlusRet {
...
+#endif
 [[noreturn]]
 static void Clone3HandlerRet() {
   StackFrameData *Data = (StackFrameData*)alloca(0);
```
neobrain, line 416: Since these functions have become increasingly complex, do we actually need Clone2HandlerRet and Clone3HandlerRet to be separate functions? If we change the `arg` parameter to be e.g. `std::optional<void*>`, we can just use `StackFrameData* Data = (StackFrameData*)(arg ? arg : alloca(0)) and keep the rest of the functions shared.

The ExitWithoutStack thing doesn't need to be a separate macro/function either then, since it's just used once.

Sonicadvance1: Implementations merged.

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
```diff
@@ -12,33 +12,53 @@ namespace FEX::LinuxEmulation::Threads {
...
+
+    if (Ptr == nullptr) {
+      Ptr = FEXCore::Allocator::mmap(nullptr, FEX::LinuxEmulation::Threads::STACK_SIZE, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    }
+
+    return Ptr;
   }
 
-  void AddStackToDeadPool(void *Ptr) {
+  bool *AddStackToDeadPool(void *Ptr) {
```
neobrain, line 58: Returning a C pointer to some boolean in a nontrivial (and undocumented) internal data structure is a massive footgun compared to what you're actually trying to do here. What's the reason for doing it like this?

Sonicadvance1: We could return a struct with a named member that does the same thing. Ensuring it is smaller than 128-bits to ensure it gets placed in to registers X0/X1.
We can't guarantee that another thread will receive a SIGCHLD to cleanup the thread's stack, so that's out.
We could have the Threads API add the stack to the thread pool and also do the exit syscall itself to keep the implementation details internal.

We really can't do much when the thread's stack is dead.

Sonicadvance1: Yes, we don't control the function signature of clone2, while we do with clone3 (until glibc adds a wrapper anyway) Which is why the code for the entrypoint needs to differ.

We need to have the exit syscall usage right next to whatever takes away ownership of the stack from the thread. Since returning from a function that takes ownership away means using a stack which the thread no longer owns.

neobrain: Fair enough. Then let's move everything into `Threads` - an interface that deals performs a process exit still fits reasonably well into that module while keeping the implementation simpler.

Sonicadvance1: Implementation moved over to Threads

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
```diff
@@ -61,6 +82,32 @@ namespace FEX::LinuxEmulation::Threads {
     AddStackToDeadPool(Ptr);
   }
 
+  [[noreturn]]
+  void DeallocateStackObjectAndExit(void *Ptr, int Status) {
+    RemoveStackFromLivePool(Ptr);
+    auto ReadyToBeReaped = AddStackToDeadPool(Ptr);
+    *ReadyToBeReaped = true;
+
+  // To behave like a real clone, we now just need to call exit here
```
neobrain, line 91: Comment doesn't apply anymore

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
```diff
@@ -61,6 +82,32 @@ namespace FEX::LinuxEmulation::Threads {
     AddStackToDeadPool(Ptr);
   }
 
+  [[noreturn]]
+  void DeallocateStackObjectAndExit(void *Ptr, int Status) {
+    RemoveStackFromLivePool(Ptr);
+    auto ReadyToBeReaped = AddStackToDeadPool(Ptr);
+    *ReadyToBeReaped = true;
```
neobrain, line 89: There's no other logic in here, and the code reading `ReadyToBeReaped` is protected by a mutex. Can't we move setting the boolean into AddStackToDeadPool and avoid the obscure bool pointer return altogether hence? This could be optional behavior controlled by a function parameter. (This might also make it clearer why ReadyToBeReaped must be set here but not in the other call site of AddStackToDeadPool.)

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.h
```diff
@@ -21,7 +21,8 @@ namespace FEX::LinuxEmulation::Threads {
    *
    * @param Ptr The stack base from `AllocateStackObject`
    */
-  void DeallocateStackObject(void *Ptr);
+  [[noreturn]]
+  void DeallocateStackObjectAndExit(void *Ptr, int Status);
```
neobrain, line 25: Docstring needs updating

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
```diff
@@ -61,6 +82,32 @@ namespace FEX::LinuxEmulation::Threads {
     AddStackToDeadPool(Ptr);
```
neobrain, line 82: Just to make sure, is it intentional that this code path doesn't set `ReadyToBeReaped`?

Sonicadvance1: Intentional stack memory leak that is happening elsewhere and is a pre-existing condition. Subject to refactoring that has yet to occur. It's one of the reasons why I've spent months moving thread management to the frontend.

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
Sonicadvance1: We can move the ReadyToBeReaped store in to `AddStackToDeadPool` if we continue passing the status all the way to that function so we can call exit there.
At the point of `ReadyToBeReaped` being set to true, we no longer have ownership of the stack. We can't return from a function, we can't call a function, we must do the syscall immediately.
The mutex guarding this doesn't matter at all, we are dancing around the thread no longer having a stack.

neobrain: The stack ownership is released when setting ReadyToBeReaped indeed, but this ownership isn't re-assigned until `AllocateStackObject` observes `ReadyToBeReaped==1`. This observation can't happen as long as `DeadStackPoolMutex` is locked. Until then, accessing the stack should still be safe.

So instead of the boolean pointer dance, why not keep `DeadStackPoolMutex` locked until after we return from `AddStackDeadPool`? This could be achieved for example by moving the mutex locking out of `AddStackToDeadPool` and locking/unlocking the mutex here manually. (To preserve current behavior elsewhere, the function could be renamed to "AddStackToDeadPoolInternal" and a helper function could be added that behaves like the current code).

neobrain: As discussed externally, this won't work because unlocking the mutex itself requires a stack. (Leaving discussion open for visibility.)

PR #3425, opened 2024-02-14
FEXCore: Add a frontend pointer to InternalThreadState

Sonicadvance1:
FEXCore is guaranteed to not touch this pointer and can be used by frontends to store thread-specific data.


FEXCore/include/FEXCore/Debug/InternalThreadState.h
```diff
@@ -139,6 +139,9 @@ namespace FEXCore::Core {
     // Async signals aren't guaranteed to be delivered in any particular order, but FEX treats them as FILO.
     fextl::vector<DeferredSignalState> DeferredSignalFrames;
 
+    ///< Pointer that FEXCore will never overwrite.
```
neobrain, line 142: The name `FrontendPtr` is ambiguous, better clear it up with the comment you already mentioned in the PR description:

```suggestion
    ///< Data pointer for exclusive use by the frontend
```

lioncash (approved):

neobrain (approved):

PR #3426, opened 2024-02-16
Linux: Make sure to destroy thread object when thread shuts down

Sonicadvance1:
This fixes a fairly large memory leak.


lioncash (approved):
Nice!

PR #3427, opened 2024-02-18
Fixes VDSO crash in 64-bit code

Sonicadvance1:
Ever since #3406 this has been crashing. Struct tail padding was saving this before.


neobrain (approved):
Good catch! I did notice VDSO-related crashes during development as well, but couldn't reproduce them after a couple of fixes. Great to have these sorted out now.

PR #3430, opened 2024-02-19
Implement small TSC scaling

Sonicadvance1:
Games engines are expecting >1Ghz cycle counters. Scale them to work around the issue.

Resolves the excessive busy waiting in Unreal Engine 5 games.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -3591,12 +3591,31 @@ void OpDispatchBuilder::SGDTOp(OpcodeArgs) {
   _StoreMemAutoTSO(GPRClass, GDTStoreSize, _Add(OpSize::i64Bit, DestAddress, _Constant(2)), _Constant(GDTAddress));
 }
 
-void OpDispatchBuilder::RDTSCOp(OpcodeArgs) {
+
+OpDispatchBuilder::CycleCounterPair OpDispatchBuilder::CycleCounter() {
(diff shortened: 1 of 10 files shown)


alyssarosenzweig (approved):
approved, but I'll likely move the shift into the dispatcher in a future PR to optimize it

(lsl + mov + lsr => lsl + ubfx or something like that)

Sonicadvance1:
> approved, but I'll mayb move the shift into the dispatcher in a future PR to optimize it
> 
> (lsl + mov + lsr => lsl + ubfx or something like that)

Switched the implementation over to that.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -3593,8 +3593,18 @@ void OpDispatchBuilder::SGDTOp(OpcodeArgs) {
 
 void OpDispatchBuilder::RDTSCOp(OpcodeArgs) {
   auto Counter = _CycleCounter();
-  auto CounterLow = _Bfe(OpSize::i64Bit, 32, 0, Counter);
-  auto CounterHigh = _Bfe(OpSize::i64Bit, 32, 32, Counter);
+  OrderedNode *CounterLow{};
```
alyssarosenzweig, line 3596: Instead of duplicating, pull out a helper CycleCounter() which does _CycleCounter + math and outputs two OrderedNodes for low/high

Sonicadvance1: Implementation merged in to a helper.

PR #3435, opened 2024-02-21
Fuse sub+cmp -> subs

alyssarosenzweig:
The one we've all been waiting . This notably fuses cmp, as well as sub/neg/dec

PR #3436, opened 2024-02-21
Simplify CalculateAF

alyssarosenzweig:
Prep for #3435


alyssarosenzweig:
Draft since I think I can do this nicer

Sonicadvance1 (approved):

PR #3438, opened 2024-02-21
Moves JITSymbol allocation

Sonicadvance1:
This isn't actually using TLS allocations. Instead it is an allocation tied to the InternalThreadState object.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -347,13 +347,6 @@ namespace FEXCore::Context {
     Thread->ThreadManager.TID = FHU::Syscalls::gettid();
     Thread->ThreadManager.PID = ::getpid();
 
-    if (Config.BlockJITNaming() ||
-        Config.GlobalJITNaming() ||
-        Config.LibraryJITNaming()) {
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -419,6 +412,13 @@ namespace FEXCore::Context {
     Thread->CurrentFrame->State.DeferredSignalRefCount.Store(0);
     Thread->CurrentFrame->State.DeferredSignalFaultAddress = reinterpret_cast<Core::NonAtomicRefCounter<uint64_t>*>(FEXCore::Allocator::VirtualAlloc(4096));
 
+    if (Config.BlockJITNaming() ||
+        Config.GlobalJITNaming() ||
+        Config.LibraryJITNaming()) {
+      // Allocate a TLS JIT symbol buffer only if enabled.
```
alyssarosenzweig, line 418: if it's not TLS, does the comment need adjustment?

Sonicadvance1: Updated comment

lioncash (approved):

PR #3439, opened 2024-02-21
FEXLoader: Allocate the second 4GB of virtual memory when executing 32-bit

Sonicadvance1:
Spurred on by https://github.com/FEX-Emu/FEX/pull/3421. To ensure that applications don't take advantage of
small address wrap around, allocate the second 4GB of virtual memory.

Some context. Linux always reserves the first 16KB of virtual address
space (unless you tinker with some settings which nobody should do).

Example of 32-bit code:
lea eax, [0xffff_0000]
mov ebx, [eax + 0x1_0000]

The address calculated by the mov will wrap around to 0x0 which will
result in SIGSEGV. If FEX messes up zero extensions then it would try to
access 0x1_0000_0000 instead.

This could result in a 32-bit application potentially accessing some FEX
memory instead of crashing.
Add this safety net which will still SIGSEGV and we will be able to see
the crash.

### DIFF
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -417,11 +417,17 @@ int main(int argc, char **argv, char **const envp) {
 
   fextl::unique_ptr<FEX::HLE::MemAllocator> Allocator;
   fextl::vector<FEXCore::Allocator::MemoryRegion> Base48Bit;
+  fextl::vector<FEXCore::Allocator::MemoryRegion> Low4GB;
 
   if (Loader.Is64BitMode()) {
     // Destroy the 48th bit if it exists
     Base48Bit = FEXCore::Allocator::Steal48BitVA();
   } else {
+    // Reserve [0x1_0000_0000, 0x2_0000_0000).
+    // Safety net if 32-bit address calculation overflows in to 64-bit range.
+    constexpr uint64_t First64BitAddr = 0x1'0000'0000ULL;
+    Low4GB = FEXCore::Allocator::StealMemoryRegion(First64BitAddr, First64BitAddr + First64BitAddr);
+
     // Setup our userspace allocator
     FEXCore::Allocator::SetupHooks();
     Allocator = FEX::HLE::CreatePassthroughAllocator();
@@ -578,6 +584,8 @@ int main(int argc, char **argv, char **const envp) {
 
   FEXCore::Allocator::ClearHooks();
   FEXCore::Allocator::ReclaimMemoryRegion(Base48Bit);
+  FEXCore::Allocator::ReclaimMemoryRegion(Low4GB);
+
   // Allocator is now original system allocator
   FEXCore::Telemetry::Shutdown(Program.ProgramName);
   FEXCore::Profiler::Shutdown();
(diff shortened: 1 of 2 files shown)


alyssarosenzweig:
It's really the *second* 4GB we're allocating, no?

Sonicadvance1:
> It's really the _second_ 4GB we're allocating, no?

Maybe it should be phrased that way. As usual my English phrasing is derp.

alyssarosenzweig (approved):

pmatos:
LGTM - curious if there a reason why you decided to allocate a full 4GB rather than a smaller window?

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -417,11 +417,15 @@ int main(int argc, char **argv, char **const envp) {
...
   fextl::unique_ptr<FEX::HLE::MemAllocator> Allocator;
   fextl::vector<FEXCore::Allocator::MemoryRegion> Base48Bit;
+  fextl::vector<FEXCore::Allocator::MemoryRegion> Low4GB;
 
   if (Loader.Is64BitMode()) {
     // Destroy the 48th bit if it exists
     Base48Bit = FEXCore::Allocator::Steal48BitVA();
   } else {
+    constexpr uint64_t First64BitAddr = 0x1'0000'0000ULL;
+    Low4GB = FEXCore::Allocator::StealMemoryRegion(First64BitAddr, First64BitAddr + First64BitAddr);
```
neobrain, line 427: Shouldn't this have a comment explaining why we're stealing this particular memory region?

Sonicadvance1:
> LGTM - curious if there a reason why you decided to allocate a full 4GB rather than a smaller window?

Primary to ensure if we mess up adding together two 32-bit values, it still captures the overflow. Bigger safety net.
And allocating 16KB versus 4GB has the same overhead so it doesn't really matter.

Amanieu:
You should reserve the first 64G so that you can use shifted-add addressing modes on AArch64 with 32-bit zero-extended values: `ldr w0, [x1, x2, sxtw #3]` instead of needing a separate 32-bit `add` for zero-extended address calculation.

This will almost always work and return an address in the low 4G, and you can handle the cases where it doesn't by catching the signal and backpatching. See section 3.3 of [my paper](https://pure.manchester.ac.uk/ws/portalfiles/portal/56078084/pldi_16.pdf) for more details.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Added a coment.

Sonicadvance1:
> You should reserve the first 64G so that you can use shifted-add addressing modes on AArch64 with 32-bit zero-extended values: `ldr w0, [x1, x2, sxtw #3]` instead of needing a separate 32-bit `add` for zero-extended address calculation.
> 
> This will almost always work and return an address in the low 4G, and you can handle the cases where it doesn't by catching the signal and backpatching. See section 3.3 of [my paper](https://pure.manchester.ac.uk/ws/portalfiles/portal/56078084/pldi_16.pdf) for more details.

Very good information, we can look at increasing the number to 64GB once/if we land that optimization.

pmatos:
> You should reserve the first 64G so that you can use shifted-add addressing modes on AArch64 with 32-bit zero-extended values: `ldr w0, [x1, x2, sxtw #3]` instead of needing a separate 32-bit `add` for zero-extended address calculation.
> 
> This will almost always work and return an address in the low 4G, and you can handle the cases where it doesn't by catching the signal and backpatching. See section 3.3 of [my paper](https://pure.manchester.ac.uk/ws/portalfiles/portal/56078084/pldi_16.pdf) for more details.

I only just noticed this comment and I think it's related to the patch I am working on #3421. Thanks for this.

PR #3440, opened 2024-02-22
InstcountCI: Adds addressing limitations to instcountci

Sonicadvance1:
Spurred on by #3421

This does a bunch of GPR and vector loads to showcase addressing limitations between ARM and x86.

Tests:
- 8/16/32/64-bit GPR loads
   - Both 32-bit and 64-bit addressing modes
- 32/64/128-bit Vector loads
   - Both 32-bit and 64-bit addressing modes
- Duplicate the tests for 32-bit addressing mode with a 32-bit process
   - Since it should change behaviour.

Untested:
- 8/16-bit vector loadstores since those don't exist on x86
   - 16-bit x87 integer load exists but that doesn't go through a vector load in FEX.

### DIFF
--- a/unittests/InstructionCountCI/FEXOpt/AddressingLimitations.json
+++ b/unittests/InstructionCountCI/FEXOpt/AddressingLimitations.json
@@ -1,0 +1,860 @@
+{
+  "Features": {
+    "Bitness": 64,
+    "EnabledHostFeatures": [],
+    "DisabledHostFeatures": [
+      "SVE128",
+      "SVE256",
+      "RPRES",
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):
lots of good stuff here, looking forward to progress here :smile:

pmatos:
Thanks for this - this is very useful. I will keep the momentum and see if I can do progress here.

pmatos:
> lots of good stuff here, looking forward to progress here 😄

will you be working on this as well? don't want to step on you toes, so let me know what you prefer.

PR #3442, opened 2024-02-22
InstCountCI: add The Witcher 3 block

alyssarosenzweig:
Closes: #2688

### DIFF
--- a/unittests/InstructionCountCI/FlagM/HotBlocks.json
+++ b/unittests/InstructionCountCI/FlagM/HotBlocks.json
@@ -1,0 +1,40 @@
+{
+  "Features": {
+    "Bitness": 64,
+    "EnabledHostFeatures": [
+      "FLAGM",
+      "FLAGM2"
+    ],
+    "DisabledHostFeatures": [
+      "SVE128",
+      "SVE256"
+    ]
+  },
+  "Instructions": {
+    "The Witcher 3": {
+      "ExpectedInstructionCount": 11,
+      "x86Insts": [
+        "mov eax, 0x1",
+        "lock xadd qword [rcx], rax",
+        "mov rdx, rax",
+        "and edx, 0x1f",
+        "inc rdx",
+        "shl rdx, 0x6",
+        "add rdx, rcx"
+      ],
+      "ExpectedArm64ASM": [
+        "mov w4, #0x1",
+        "ldaddal x4, x4, [x5]",
+        "mov x6, x4",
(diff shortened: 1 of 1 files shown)


unittests/InstructionCountCI/FlagM/HotBlocks.json
```diff
@@ -0,0 +1,40 @@
...
+        "mov rdx, rax",
+        "and edx, 0x1f",
+        "inc rdx",
+        "shl rdx, 0x6",
+        "add rdx, rcx"
+      ],
+      "ExpectedArm64ASM": [
+        "mov w4, #0x1",
+        "ldaddal x4, x4, [x5]",
+        "mov x6, x4",
```
alyssarosenzweig, line 28: We should really be able to DCE this move, I guess SRA is getting in the way of regular DCE doing it, so it'd need to be either RCLSE or something like the flag elim pass. @Sonicadvance1 thoughts which is more appropriate?

Sonicadvance1: Not sure, ldaddal will update the value in rax but then it also copies the value in to rdx there. Can it actually be eliminated unless we know this is a looping block or if rax later gets overwritten?

Sonicadvance1 (approved):

unittests/InstructionCountCI/FlagM/HotBlocks.json
alyssarosenzweig: yes, x6 is overwritten later in the block

alyssarosenzweig: Tracking in #3444

Sonicadvance1: x6 is overwritten yes, which corresponds to rdx. But x4 which corresponds to rax is not. So rax and rdx contain the same returned value from ldaddal once the move occurs

alyssarosenzweig: ..so? The sequence is

        "mov x6, x4",
        "and w6, w4, #0x1f",

The mov can be deleted without affecting either register after the and.

Sonicadvance1: Oop, yep. I was on my phone so I didn't capture the full context that rdx is overwritten so it can remove the mov in favour of using the ALU operation on rax directly.

PR #3443, opened 2024-02-22
Fuse add + cmn -> adds

alyssarosenzweig:
On top of #3442 and #3435 which is in turn on top of #3436. This applies the ideas that #3435 did for sub and applies it to add.


Sonicadvance1 (approved):
Nice nice nice
Image_2024-02-24_15-05-20

PR #3446, opened 2024-02-22
Syscalls: Fix SourcecodeMap generation for GDB JIT integration

neobrain:
This fixes a regression from 9dd715573, which accidentally changed the filename and set up incorrect file opening flags.

For personal reference, it seems that Ubuntu's `gdb` package enables assertions that break with FEX's GDBJITReader, however it's working fine when building gdb from source with the default configuration.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
@@ -975,7 +975,7 @@ fextl::unique_ptr<FEXCore::HLE::SourcecodeMap> SyscallHandler::GenerateMap(const
     fextl::istringstream Stream(SourceData);
 
     constexpr int USER_PERMS = S_IRWXU | S_IRWXG | S_IRWXO;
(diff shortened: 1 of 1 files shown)


pmatos:
How can I test this? Are there any docs online on using the sourcecode map generation in gdb?

Sonicadvance1 (approved):

neobrain:
> How can I test this? Are there any docs online on using the sourcecode map generation in gdb?

The corresponding gdb feature is [documented here](https://sourceware.org/gdb/current/onlinedocs/gdb.html/Custom-Debug-Info.html#Custom-Debug-Info), but that's not too helpful for testing FEX's implementation.

To test out use in FEX:
* Enable the CMake setting `ENABLE_GDB_SYMBOLS=1`
* Run `FEXLoader --gdbsymbols <program>` to generate the source map for the program and the libraries it uses
* Re-run FEX, this time in gdb: `gdb --args FEXLoader --gdbsymbols <program>`
* In gdb, run `jit-reader-load libFEXGDBReader.so` and `layout src`
* Ideally, x86 assembly will now be visible when halting executing within JIT code

neobrain:
(rebased without changes to drop the auto-generated merge commit added in last rebase)

PR #3447, opened 2024-02-22
Fix instcountci

Sonicadvance1:
Looks like #3436 had an interaction that was unexpected.

Not sure why the scripts aren't showing this diff, should be looked at separately.


alyssarosenzweig:
This is the problem with not rerunning CI when rebasing... Marge Bot fixes this over in gitlab land but yeah.

PR #3448, opened 2024-02-23
Further address mode optimizations

pmatos:
Motivated by #3440


alyssarosenzweig:
32-bit zero extension  is not a noop

pmatos:
> 32-bit zero extension is not a noop

Right. That's why I shouldn't create PRs on a Friday evening. :)

pmatos:
Going to look at the x87 stack machine opt, so closing this for now.

PR #3450, opened 2024-02-24
InstcountCI: Adds a vector addition loop from bytemark

Sonicadvance1:
This was interesting because it caught that we are failing loadstores with reg+reg address generation on vectors.

Additionally the address generation is duplicated twice, but once reg+reg gets added that won't be an issue.


alyssarosenzweig (approved):
approved, but it'l conflict with #3443

PR #3451, opened 2024-02-25
Fixes zero register flag generation

Sonicadvance1:
Fixes 140976d322dc5e26c129d1e6448f25f6b2378341

Adds a unit test to ensure it keeps working.

Before fix:
Screenshot 2024-02-24 15-33-39

After fix:
Screenshot 2024-02-24 16-34-35


alyssarosenzweig:
ngl I like the before better

alyssarosenzweig (approved):
heh, right, this fixes PF too for this case, whoopsies & thank you

PR #3459, opened 2024-02-26
Capture a 64-bit process trying to jump to 32-bit syscall handler

### LINKED ISSUE #591: Capture x86-64 applications jumping to 32bit syscalls
x86-64 applications can still call in to the 32bit x86 syscall interface with `int 0x80`
Capture this case.

Sonicadvance1:
Fixes #591

Adds a simple unittest

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -5218,6 +5218,9 @@ void OpDispatchBuilder::INTOp(OpcodeArgs) {
     constexpr uint8_t SYSCALL_LITERAL = 0x2E;
 #endif
     if (Literal == SYSCALL_LITERAL) {
+      if (CTX->Config.Is64BitMode()) [[unlikely]] {
+        ERROR_AND_DIE_FMT("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");
+      }
       // Syscall on linux
       SyscallOp<false>(Op);
       return;
--- a/unittests/ASM/FEX_bugs/32bit_syscall.asm
+++ b/unittests/ASM/FEX_bugs/32bit_syscall.asm
@@ -1,0 +1,32 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0"
+  }
+}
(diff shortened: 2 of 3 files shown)


lioncash (approved):

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -5218,6 +5218,9 @@ void OpDispatchBuilder::INTOp(OpcodeArgs) {
     constexpr uint8_t SYSCALL_LITERAL = 0x2E;
 #endif
     if (Literal == SYSCALL_LITERAL) {
+      if (CTX->Config.Is64BitMode()) [[unlikely]] {
+        LogMan::Msg::AFmt("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");
```
neobrain, line 5222: What will happen in this case? Is there a chance this will behave correctly at all?

Sonicadvance1: It will not behave correctly. It will call an incorrect syscall since we only install syscall handlers for the architecture that was originally started.
This will force a SIGABRT with the log message.

neobrain: Isn't `AFmt` a no-op on Release/RelWithDebInfo builds?

Sonicadvance1: `LOGMAN_MSG_A_FMT` is a no-op. `LogMan::Msg::AFmt` is not.

Sonicadvance1: er wait, made a fool. That was legacy behaviour. I need to change this

Sonicadvance1: Swapped to `ERROR_AND_DIE_FMT`

neobrain: I wasn't entirely sure either, but glad it's fixed now :)

neobrain (approved):

PR #3460, opened 2024-02-26
Adds a unittest for a bug from #3421

Sonicadvance1:
When the source arguments for LoadMem/StoreMem have bit 31 set then they are incorrectly sign extending in some instances.

Detected this when testing #3421 but I don't have a proper fix for it.

### DIFF
--- a/unittests/32Bit_ASM/FEX_bugs/SignExtendBug.asm
+++ b/unittests/32Bit_ASM/FEX_bugs/SignExtendBug.asm
@@ -1,0 +1,20 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x41424344"
+  },
+  "MemoryRegions": {
+    "0xf0000000": "4096"
+  },
+  "MemoryData": {
+    "0xf0000000": "0x41424344"
+  },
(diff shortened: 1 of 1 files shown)


pmatos:
Nice - thanks for adding this. I will take a look.

alyssarosenzweig (approved):

alyssarosenzweig:
> I don't have a proper fix for it.

Simply ignoring sign ext in 32-bit mode doesn't work?

Sonicadvance1:
> > I don't have a proper fix for it.
> 
> Simply ignoring sign ext in 32-bit mode doesn't work?

It may matter if the sign extension wants a negative number or a large offset. Since behaviour differs there if your VA wraps around in 32-bit space or not.

pmatos:
> > I don't have a proper fix for it.
> 
> Simply ignoring sign ext in 32-bit mode doesn't work?

I added sign extension to deal with this case:
https://github.com/FEX-Emu/FEX/pull/3421/files#diff-8f8264c0e8c14c4d9e2bbbb7d7c71c7ad8d4113090517942e7a83ede596064da

PR #3461, opened 2024-02-26
Optimize SBC

alyssarosenzweig:
the natural follow on to ADC. divided up into smaller commits because this one is a bit more delicate, thanks to the silly `salc` op.


Sonicadvance1:
Nice to see SALC turn in to two instructions. Will give this some testing in a bit to ensure it doesn't break anything but is looking pretty good.

Sonicadvance1:
Image_2024-02-27_00-28-48
Looks like a couple of benches are outside of noise.

Sonicadvance1 (approved):
Some good instruction reductions here.

PR #3463, opened 2024-02-27
Update xxhash to v0.8.2

Sonicadvance1:
Update xxhash to v0.8.2

Switches to using upstream cmake files. Which means we don't need to use our own fork now.

There has been some good ARM ASIMD/SVE optimizations since our last update. I have also seen some additional changes merged, so when they do a version bump we should probably update again.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -41,7 +41,7 @@
 	url = https://github.com/FEX-Emu/drm-headers.git
 [submodule "External/xxhash"]
 	path = External/xxhash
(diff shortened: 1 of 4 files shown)


lioncash (approved):
([changelog](https://github.com/Cyan4973/xxHash/releases/tag/v0.8.2) for any other readers)

Nice. ~20% improvement for the neon path

lioncash (approved):

PR #3464, opened 2024-02-27
InstcountCI: Add a monster of a game block

Sonicadvance1:
Doing very little work with a bunch of instructions. Hottest block in the Windows version of Psychonauts, it's just doing a matrix swizzle but in the worst possible way.


lioncash (approved):
I can't believe "worst" actually comes out to be a bit of an understatement here

pmatos:
Then block is gets so blown up that it's funny. :)

PR #3465, opened 2024-02-27
InstCountCI: enable preserve_all

alyssarosenzweig:
This is what we'll actually ship (I hope), so that's the config we want to
    track long-term. It's also a lot more managable resulting asm.


Sonicadvance1 (approved):
vroom

PR #3466, opened 2024-02-28
OpcodeDispatcher: Don't use AddShift with no shift

Sonicadvance1:
This accidentally removed optimizations elsewhere that was only checking
for Add.

Adds a unit test that gets unveiled with this that #3421 ends up breaking.


alyssarosenzweig:
At some point I want to merge Add and AddShift into a single instruction, but for now this seems fine

alyssarosenzweig (approved):

Sonicadvance1:
> At some point I want to merge Add and AddShift into a single instruction, but for now this seems fine

I like that idea. Looking forward to it.

PR #3468, opened 2024-02-28
Misc little opts

alyssarosenzweig:
Just browsing instcountci for low hanging fruit


Sonicadvance1 (approved):
Nice little improvements.

Issue #3472, opened 2024-02-29
Skyrim SE: Hottest blocks


Sonicadvance1:
Skyrim SE's hottest blocks are really hot atomic loop blocks:
1- 6.83% of primary thread
```
6ffffd37ad33  89f8               mov     eax, edi  {0x1}
6ffffd37ad35  8706               xchg    dword [rsi], eax  {0x1}
6ffffd37ad37  85c0               test    eax, eax
6ffffd37ad39  7555               jne     0x6ffffd37ad90
```

2- 5.12% of primary thread
```
6ffffd4a8300  488d410c           lea     rax, [rcx+0xc]
6ffffd4a8304  f0ff08             lock dec dword [rax]
6ffffd4a8307  7407               je      0x6ffffd4a8310
```

3- 5.03% of primary thread
```
6ffffd4405f7  lea     rax, [rcx+0xc]
6ffffd4405fb  lock dec dword [rax]
6ffffd4405fe  jne     0x6ffffd44060d
```
lock dec at the very least can optimize away a mov+neg combination in to a mov. Looking at instcountci.
https://github.com/FEX-Emu/FEX/blob/d24446ed139b6740d5e606ddc05b7028b5fecaf8/unittests/InstructionCountCI/FlagM/Atomics.json#L1412

These three blocks are totaling 16.98% of CPU time on the primary thread on Cortex-A78AE. All coming from d3d11.dll in DXVK.

4- 2.7% - RIP: 0x1414df7c7 - SkyrimSE.exe
- Encrypted EXE, needs investigation

alyssarosenzweig:
closing because there's not really anything actionable here.

PR #3475, opened 2024-02-29
Optimize lock dec

alyssarosenzweig:
for #3472


Sonicadvance1 (approved):
Simple little optimization

PR #3477, opened 2024-02-29
Library Forwarding: Update Vulkan definitions to v1.3.278

### LINKED ISSUE #3476: Need to update Vulkan Headers for thunks.
Newer Vulkan drivers (like Mesa 24.0) ships a newer version of Vulkan than what our thunks supports. This causes DXVK to use features that our thunks don't support and results in games having early initialization failure.

We currently ship: v1.3.261
Mesa radv ships as of writing: 1.3.274
Current header version as of writing: 1.3.278

neobrain:
Should fix #3476, though I didn't explicitly check DXVK.

Updated as described in #2076.


Sonicadvance1:
I'll give this a try today to see if it fixes the issue.

Sonicadvance1 (approved):
Sadly didn't fix my problem but doesn't seem to cause any new problems.
Nice to have the update although.

PR #3480, opened 2024-03-01
Use SETF8/16 for 8/16-bit INC/DEC

alyssarosenzweig:
This isn't as useful as you'd like, since the definition of SETF seems to be broken (due to a fundamental misunderstanding at Arm about how OF works?). But it's still a win in this one case.


Sonicadvance1 (approved):
Fun little optimization. Weird overflow behaviour is weird.

alyssarosenzweig:
Realized on my walk tonight that I forgot to update the flag opt pass. Fixed, and added an instcountci test that would've flagged the problem (pun intended)

Sonicadvance1 (approved):
Nice catch on the oversight there

PR #3474, opened 2024-02-29
Fix reserving range check

Sonicadvance1:
Fixes an issue where TestHarnessRunner was managing to reserve the space below stack again, resulting in stack growth breaking. Would typically only show up when using the vixl simulator under gdb for some reason.

This is likely the last bandage on this code before it gets completely rewritten to be more readable.

PR #3483, opened 2024-03-05
Allocator: Cleanup StealMemoryRegions implementation

neobrain:
**Marked as draft seeking general feedback and a nod-off for a minor functional change.**

This throws the completely overblown handwritten state-machine that emulated regex parsing before. Instead, `/proc/self/maps` is parsed line-by-line, which flattens control flow and enables use of `std::from_chars`.

To minimize risk for regressions, interfaces have been changed to allow unit-testing this functionality. Special care was taken around the StackLocation case (see e.g. #3474) by uglifying `StealMemoryRegions`'s interface temporarily. This ensures each individual patch is correct, but the final patch drops this test coverage in the interest of keeping the interface clean.

A minor functional change was encountered during this testing: If the range end falls into the stack mapping, the new version now applies the stack special handling (when the previously version did not). I believe this is the more sensible behavior.

### DIFF
--- a/FEXCore/Source/Utils/Allocator.cpp
+++ b/FEXCore/Source/Utils/Allocator.cpp
@@ -274,7 +274,7 @@ namespace FEXCore::Allocator {
         if (c == ' ') {
           STEAL_LOG("[%d] ParseEnd; RegionBegin: %016lX RegionEnd: %016lX\n", __LINE__, RegionBegin, RegionEnd);
 
-          if (RegionEnd > End) {
+          if (RegionEnd >= End) {
             // Early return if we are completely beyond the allocation space.
             return Regions;
           }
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
This approach looks reasonable to me, much easier to read. Did some testing to ensure the world wasn't broken, also small benchmarking shows that this new approach is 25%-30% faster. Average of 62k nanosecond down to 50k nanoseconds.

neobrain:
> This approach looks reasonable to me, much easier to read. Did some testing to ensure the world wasn't broken, also small benchmarking shows that this new approach is 25%-30% faster. Average of 62k nanosecond down to 50k nanoseconds.

Thanks for testing!

Un-drafted after having cleaned up the tests. Implementation is unchanged compared to the draft.

Sonicadvance1 (approved):

Issue #3455, opened 2024-02-25
Vulkan thunks rendering issue with Detroit Become Human.


Sonicadvance1:
Tested with Mesa 23.3.6 in both the rootfs and the host.
System:
SoC: NVIDIA Orin
GPU: AMD Radeon Pro W7500
Driver: RADV 23.3.6

Detroit: Become Human has rendering artifacts with thunks enabled but go away with thunks disabled.
Hard to diagnose exactly what the issue is.

Thunks disabled:
![Screenshot 2024-02-24 18-24-02](https://github.com/FEX-Emu/FEX/assets/1018829/c38a2166-fbe2-452e-8c95-dbbf86f25466)

Thunks Enabled:
![Screenshot 2024-02-24 18-11-41](https://github.com/FEX-Emu/FEX/assets/1018829/d94867c6-3151-4cd7-90a4-c3ab7a809255)

PR #3488, opened 2024-03-08
Revert "Library Forwarding/gen: Map integers to fixed-size equivalents on guest"

Sonicadvance1:
This reverts commit d04c94fe802a09288a24ea68c7f951474bb0235a.

This isn't necessarily expected to be merged but instead opening a discussion about why this is breaking rendering with thunking enabled.

In particular this revert fixes "Broforce". Before the revert the game looked like it was rendering with a bad anaglyph 3D renderer.

This might also be related to #3455 but I haven't had a chance to retest it.

This could be related to sign/zero extension semantics being different between AArch64 and x86?

Original commit comes from PR #3392

### DIFF
--- a/ThunkLibs/Generator/analysis.cpp
+++ b/ThunkLibs/Generator/analysis.cpp
@@ -436,17 +436,7 @@ void AnalysisAction::ParseInterface(clang::ASTContext& context) {
                             check_struct_type(param_type.getTypePtr());
                             types.emplace(context.getCanonicalType(param_type.getTypePtr()), RepackedType { });
                         } else if (param_type->isPointerType()) {
-                            auto pointee_type = param_type->getPointeeType();
-
-                            if (pointee_type->isIntegerType()) {
-                                // Add builtin pointee type to type list
-                                if (!pointee_type->isEnumeralType()) {
-                                    types.emplace(pointee_type.getTypePtr(), RepackedType { });
-                                } else {
-                                    types.emplace(context.getCanonicalType(pointee_type.getTypePtr()), RepackedType { });
-                                }
-                            }
-
+                            auto pointee_type = param_type->getPointeeType()->getLocallyUnqualifiedSingleStepDesugaredType();
                             if (data.param_annotations[param_idx].assume_compatible) {
                                 // Nothing to do
                             } else if (types.contains(context.getCanonicalType(pointee_type.getTypePtr())) && LookupType(context, pointee_type.getTypePtr()).assumed_compatible) {
@@ -519,12 +509,6 @@ void AnalysisAction::ParseInterface(clang::ASTContext& context) {
 }
 
 void AnalysisAction::CoverReferencedTypes(clang::ASTContext& context) {
-    // Add common fixed-size integer types explicitly
-    for (unsigned size : { 8, 32, 64 }) {
-        types.emplace(context.getIntTypeForBitwidth(size, false).getTypePtr(), RepackedType {});
-        types.emplace(context.getIntTypeForBitwidth(size, true).getTypePtr(), RepackedType {});
-    }
-
     // Repeat until no more children are appended
     for (bool changed = true; std::exchange(changed, false);) {
         for ( auto next_type_it = types.begin(), type_it = next_type_it;
--- a/ThunkLibs/Generator/analysis.h
+++ b/ThunkLibs/Generator/analysis.h
@@ -199,15 +199,3 @@ inline std::string get_type_name(const clang::ASTContext& context, const clang::
     }
     return type_name;
 }
-
-inline std::string get_fixed_size_int_name(bool is_signed, int size) {
-    return (!is_signed ? "u" : "") + std::string { "int" } + std::to_string(size) + "_t";
-}
-
-inline std::string get_fixed_size_int_name(const clang::Type* type, int size) {
-    return get_fixed_size_int_name(type->isSignedIntegerType(), size);
-}
-
-inline std::string get_fixed_size_int_name(const clang::Type* type, const clang::ASTContext& context) {
-    return get_fixed_size_int_name(type, context.getTypeSize(type));
-}
--- a/ThunkLibs/Generator/data_layout.cpp
+++ b/ThunkLibs/Generator/data_layout.cpp
@@ -87,7 +87,6 @@ std::unordered_map<const clang::Type*, TypeInfo> ComputeDataLayout(const clang::
                 .array_size = array_size,
                 .is_function_pointer = field_type->isFunctionPointerType(),
                 .is_integral = field->getType()->isIntegerType(),
-                .is_signed_integer = field->getType()->isSignedIntegerType(),
             };
 
             // TODO: Process types in dependency-order. Currently we skip this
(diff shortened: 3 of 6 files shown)


neobrain:
Very interesting this would break things on 64-bit guests. For those, in theory the underlying types shouldn't really change. I'll try and see if I can pin something down by adding some `static_assert`s locally, otherwise a renderdoc capture recorded from guest-lavapipe might be handy for debugging.

> This could be related to sign/zero extension semantics being different between AArch64 and x86?

What are the differences?

neobrain:
I couldn't find any obvious difference in code generation. The types used throughout libvulkan for 64-bit guests after the change almost fully match the ones used before. The only difference is that some `const char*` pointers are turned into `const uint8_t*`. For reference, this is the patch I used to check:
```patch
diff --git ThunkLibs/Generator/gen.cpp ThunkLibs/Generator/gen.cpp
index e47e335bc..a9ee37a65 100644
--- ThunkLibs/Generator/gen.cpp
+++ ThunkLibs/Generator/gen.cpp
@@ -549,7 +549,7 @@ void GenerateThunkLibsAction::OnAnalysisComplete(clang::ASTContext& context) {
                 }
             }

-            auto get_guest_type_name = [this](clang::QualType type) {
+            auto get_guest_type_name2 = [this](clang::QualType type) {
                 if (type->isBuiltinType() && !type->isFloatingType()) {
                     auto size = guest_abi.at(type.getUnqualifiedType().getAsString()).get_if_simple_or_struct()->size_bits;
                     return get_fixed_size_int_name(type.getTypePtr(), size);
@@ -560,6 +560,9 @@ void GenerateThunkLibsAction::OnAnalysisComplete(clang::ASTContext& context) {
                     return type.getUnqualifiedType().getAsString();
                 }
             };
+            auto get_guest_type_name = [&](clang::QualType type) {
+                return fmt::format("std::enable_if_t<std::is_same_v<{}, const char*> || std::is_same_v<{}, {}>, {}>", type.getAsString(), type.getAsString(), get_guest_type_name2(type), get_guest_type_name2(type));
+            };

             // Forward declarations for user-provided implementations
             if (thunk.custom_host_impl) {
diff --git ThunkLibs/include/common/Host.h ThunkLibs/include/common/Host.h
index 3783dab56..b650cce41 100644
--- ThunkLibs/include/common/Host.h
+++ ThunkLibs/include/common/Host.h
@@ -467,6 +467,8 @@ struct GuestWrapperForHostFunction<Result(Args...), GuestArgs...> {
     static_assert(sizeof...(Annotations) == sizeof...(Args));
     static_assert(sizeof...(GuestArgs) == sizeof...(Args));

+    static_assert((std::is_same_v<GuestArgs, Args> && ...));
+
     auto args = reinterpret_cast<PackedArguments<Result, guest_layout<GuestArgs>..., uintptr_t>*>(argsv);
     constexpr auto CBIndex = sizeof...(GuestArgs);
     uintptr_t cb;
```

So the next step would be either to do a more finely-grained revert, or to test against a renderdoc capture.

Sonicadvance1:
[Thunked.zip](https://github.com/FEX-Emu/FEX/files/14565421/Thunked.zip)
[Thunked_reverted.zip](https://github.com/FEX-Emu/FEX/files/14565446/Thunked_reverted.zip)
[Unthunked.zip](https://github.com/FEX-Emu/FEX/files/14565422/Unthunked.zip)
Here are apitrace captures (from revision a2cfbdf876b86af7665184f8c08b328f1c4f6a66 if it matters).
Including Thunked (with broken output), Thunked with this revert (correct rendering), and unthunked running through the x86 libraries. Should allow some inspection to see why the rendering has changed.

The corruption can be seen immediately with the first logo on the screen having a magenta background instead of a black background.

Title

alyssarosenzweig:
> The only difference is that some const char* pointers are turned into const uint8_t*

I wonder if the change in signedness is affecting anything...?

Sonicadvance1:
> Very interesting this would break things on 64-bit guests. For those, in theory the underlying types shouldn't really change. I'll try and see if I can pin something down by adding some `static_assert`s locally, otherwise a renderdoc capture recorded from guest-lavapipe might be handy for debugging.
> 
> > This could be related to sign/zero extension semantics being different between AArch64 and x86?
> 
> What are the differences?

As for the differences between 8-bit and 16-bit sign extension between x86 and aarch64.
x86 systemv ABI spec under-specifies this, which means compilers have made their own ABI. Which has become both the callee AND the caller will sign/zero extend the smaller value to be 32-bit.

AAPCS64 is a bit nicer in this regard, it is specified that only the Callee needs to sign/zero extend the value to 32-bit. Thus allowing the optimization that if the upper bits can be ignored then the function doesn't need to sign/zero extend at all.

neobrain:
> I wonder if the change in signedness is affecting anything...?

These pointers (all used as strings as far as I could see) aren't dereferenced before passing them to the native host function, so the underlying type doesn't play into semantics. Other pointers have their signedness preserved, but `char` is a special case since `char`/`unsigned char`/`signed char` are three distinct types (thanks, C).

Obviously can't rule out something more subtle is going on, but at that point I'd be reading tea leaves.

neobrain:
> [Thunked.zip](https://github.com/FEX-Emu/FEX/files/14565421/Thunked.zip) [Thunked_reverted.zip](https://github.com/FEX-Emu/FEX/files/14565446/Thunked_reverted.zip) [Unthunked.zip](https://github.com/FEX-Emu/FEX/files/14565422/Unthunked.zip) Here are apitrace captures (from revision a2cfbdf876b86af7665184f8c08b328f1c4f6a66 if it matters). Including Thunked (with broken output), Thunked with this revert (correct rendering), and unthunked running through the x86 libraries. Should allow some inspection to see why the rendering has changed.
> 
> The corruption can be seen immediately with the first logo on the screen having a magenta background instead of a black background.

Thanks! The issue did not reproduce for me on playback, but the API dump clearly showed a difference in `glTexParameterf`. Could you check if changing https://github.com/FEX-Emu/FEX/blob/e33a76a2ad960d650dc46666eb9d7cb53911ec08/ThunkLibs/Generator/data_layout.cpp#L194 from `if (type->isBuiltinType()) {` to `if (type->isBuiltinType() && type->isInteger()) {` fixes the issue?

I'll need to do some further testing, but if my working theory is correct that should resolve the problem.

Sonicadvance1:
> if (type->isBuiltinType() && ((clang::BuiltinType*)type)->isInteger()) {

Couldn't quite use that exact patch since it couldn't compiled.

Did this instead.
```diff
diff --git a/ThunkLibs/Generator/data_layout.cpp b/ThunkLibs/Generator/data_layout.cpp
index 0076a69d7..8d0697ce3 100644
--- a/ThunkLibs/Generator/data_layout.cpp
+++ b/ThunkLibs/Generator/data_layout.cpp
@@ -191,7 +191,7 @@ static std::array<uint8_t, 32> GetSha256(const std::string& function_name) {
 };

 std::string GetTypeNameWithFixedSizeIntegers(clang::ASTContext& context, clang::QualType type) {
-    if (type->isBuiltinType()) {
+    if (type->isBuiltinType() && type->isIntegerType()) {
         auto size = context.getTypeSize(type);
         return fmt::format("uint{}_t", size);
     } else if (type->isPointerType() && type->getPointeeType()->isBuiltinType() && context.getTypeSize(type->getPointeeType()) > 8) {
```

This fixed the rendering issues in Broforce.

Sonicadvance1:
Looks like this will also fix #3455, just retested.

PR #3491, opened 2024-03-11
RCLSE: Optimize store-after-store

### LINKED ISSUE #3444: DCE/RCLSE should remove dead register moves
In multiblock

              We should really be able to DCE this move, I guess SRA is getting in the way of regular DCE doing it, so it'd need to be either RCLSE or something like the flag elim pass. @Sonicadvance1 thoughts which is more appropriate?

_Originally posted by @alyssarosenzweig in https://github.com/FEX-Emu/FEX/pull/3442#discussion_r1498517641_

alyssarosenzweig:
Closes #3444


Sonicadvance1 (approved):
Image_2024-03-13_17-42-11

I'm fine with merging this as-is, but we should work to resolve some of the instructions that were made worse off as we go forward.

alyssarosenzweig:
> Image_2024-03-13_17-42-11
> 
> I'm fine with merging this as-is, but we should work to resolve some of the instructions that were made worse off as we go forward.

Woof, those results are all over the place.

RA rewrite should fix the regressions but yeah...

PR #3497, opened 2024-03-18
JIT: Optimize pmovmaskb with a named vector constant

Sonicadvance1:
I was looking at some other JIT overheads and this cropped up as some
overhead. Instead of materializing a constant using mov+movk+movk+movk,
load it from the named vector constant array.

In a micro-benchmark this improved performance by 34%.
In bytemark this improved on subbench by 0.82%


alyssarosenzweig (approved):
lgtm. It feels like we "ought" to be able to generate the constant with a bytewise vectorized `(1 << lane ID)` but I can't figure out how to actually do that in neon any faster than what we had here so... approved given the positive benchmark data :)

JosJuice:
I can't think of a good way to do that in NEON either. SVE does have the INDEX instruction, but even with that, generating the 64-bit constant would take three instructions. So it doesn't seem like it would be better than this PR's approach.

PR #3478, opened 2024-02-29
FEXCore: Add non-atomic Memcpy and Memset IR fast paths

bylaws:
When TSO is disabled, vector LDP/STP can be used for a two instruction 32 byte memory copy which is significantly faster than the current byte-by-byte copy. Performing two such copies directly after oneanother also marginally increases copy speed for all sizes >=64.
Before:
```
-----------------------------------------------------------
Benchmark                 Time             CPU   Iterations
-----------------------------------------------------------
BM_SyscallBasic        20.2 ns         20.3 ns     35000000
BM_repmovsb/8          3.72 ns         3.70 ns    189189189 bytes_per_second=2.01367Gi/s
BM_repmovsb/16         6.73 ns         6.60 ns    100000000 bytes_per_second=2.22405Gi/s
BM_repmovsb/32         12.9 ns         12.8 ns     53846154 bytes_per_second=2.32571Gi/s
BM_repmovsb/64         25.5 ns         25.6 ns     26923077 bytes_per_second=2.32571Gi/s
BM_repmovsb/128        50.6 ns         50.0 ns     10000000 bytes_per_second=2.38419Gi/s
BM_repmovsb/256       100.0 ns         98.6 ns      7000000 bytes_per_second=2.41874Gi/s
BM_repmovsb/512         190 ns          191 ns      3500000 bytes_per_second=2.49094Gi/s
BM_repmovsb/1024        378 ns          375 ns      1842105 bytes_per_second=2.54604Gi/s
BM_repmovsb/2048        746 ns          740 ns      1000000 bytes_per_second=2.5775Gi/s
BM_repmovsb/4096       1473 ns         1479 ns       466667 bytes_per_second=2.57999Gi/s
BM_repmovsb/8192       2905 ns         2900 ns       241379 bytes_per_second=2.63082Gi/s
BM_repstosb/8          3.68 ns         3.70 ns    189189189 bytes_per_second=2.01367Gi/s
BM_repstosb/16         6.36 ns         6.34 ns    116666667 bytes_per_second=2.34928Gi/s
BM_repstosb/32         11.7 ns         11.8 ns     63636364 bytes_per_second=2.52868Gi/s
BM_repstosb/64         22.4 ns         22.6 ns     31818182 bytes_per_second=2.63404Gi/s
BM_repstosb/128        44.2 ns         44.0 ns     15909091 bytes_per_second=2.7093Gi/s
[...]

PR #3499, opened 2024-03-20
unittests/ASM: Adds a test for overlapping memcpy using rep movs

Sonicadvance1:
Caused by #3478

This was missed in the review that it could cause issues. bylaws already has a fix incoming that will get this unit test working.


alyssarosenzweig (approved):
approved, didn't check the details but this has the right shape

PR #3500, opened 2024-03-20
Thunks: Disable 32-bit thunks unless a config option is set

Sonicadvance1:
This will be necessary before #3487 is merged.
Once #3487 is merged then 32-bit Vulkan thunk libraries is built which results in crashing in every 32-bit vulkan/dxvk game.

Explicitly disable the thunking until this config option is set. Once X11 is as stable on 32-bit as 64-bit then this option can get removed.

For testing purposes it's easy to set the environment variable `FEX_THUNKS32BITENABLED=1`

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -137,6 +137,8 @@ jobs:
     - name: FEXLinuxTests
       working-directory: ${{runner.workspace}}/build
       shell: bash
+      env:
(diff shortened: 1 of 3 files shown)


neobrain:
Not sure adding user-facing features as temporary workaround is the right way forward here. Maybe this can instead be covered by the "forwarding groups" idea that we'll need for reliable window system integration either way. I'll do some more thinking in this direction.

PR #3507, opened 2024-03-22
FEXLoader: Add some debug-only tracking for FEX owned FDs

Sonicadvance1:
I remember seeing some application last year where they closed a FEX owned FD but now I don't remember what it was. This can really mess us up so add some debug tracking so we can try and find it again.

Might be something specifically around flatpack, appimage, or chrome's sandbox. I have some ideas about how to work around these problems if they crop up but need to find the problem applications again.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.h
@@ -88,7 +88,53 @@ public:
 
   bool SupportsProcFSInterpreterPath() const { return SupportsProcFSInterpreter; }
 
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+  void TrackFEXFD(int FD) noexcept {
+    std::lock_guard lk(FEXTrackingFDMutex);
+    FEXTrackingFDs.emplace(FD);
+  }
+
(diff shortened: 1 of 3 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.h
```diff
@@ -88,7 +88,60 @@ class FileManager final {
...
+  void RemoveFEXFD(int FD) noexcept {
+    std::lock_guard lk(FEXTrackingFDMutex);
+    FEXTrackingFDs.erase(FD);
+  }
+
+  void RemoveFEXFDRange(int begin, int end) noexcept {
+    std::lock_guard lk(FEXTrackingFDMutex);
+
+    // Just linear scan since the number of tracking FDs is low.
+    for (auto it = FEXTrackingFDs.begin(); it != FEXTrackingFDs.end(); ) {
```
lioncash, line 106: Seems like this can be:

```cpp
std::erase_if(FEXTrackingFDs, [begin, end](int FD) {
  return FD >= begin && (FD <= end || end == -1);
});
```

if I'm reading this right

Sonicadvance1: You're right, changed.

alyssarosenzweig (approved):

PR #3514, opened 2024-03-25
rewrite Demon Addition Adjust (DAA) and other demonic opcodes

alyssarosenzweig:
yakshave for new ra


Sonicadvance1 (approved):
Such a massive improvement. I assume we will also end up seeing DAS with a similar change.

alyssarosenzweig:
> Such a massive improvement. I assume we will also end up seeing DAS with a similar change.

yeah, I think I'm going to rewrite all the demon ops to get them out of the way for RA rework

alyssarosenzweig:
@Sonicadvance1 I added a bunch of new commits to rework the whole set of demons, you probably want to rerevview

Sonicadvance1 (approved):
Definitely some nice cleanups here.
Bit of trusting the process on these instructions, I did some testing to ensure nothing breaks, but I really don't have anything that absolutely abuses these.

I know our unit tests are a bit lacking for these instructions, but even if something is missed then we can deal with that when it happens.

alyssarosenzweig:
> Definitely some nice cleanups here. Bit of trusting the process on these instructions, I did some testing to ensure nothing breaks, but I really don't have anything that absolutely abuses these.
> 
> I know our unit tests are a bit lacking for these instructions, but even if something is missed then we can deal with that when it happens.

Yeah... I exhaustively tested the DAS logic with a python script but should maybe do similar tests for the others... since that *did* find bugs 😅

PR #3518, opened 2024-03-26
Put <20M in double quotes to avoid truncate error

pmatos:
Avoids bash assuming < is redirection.
See error in https://github.com/FEX-Emu/FEX/actions/runs/8434045431/job/23096421079#step:18:16


Sonicadvance1 (approved):
Oh, hah. Wacky.

PR #3522, opened 2024-03-27
OpcodeDispatcher: eliminate branch in cmpxchg pair

alyssarosenzweig:
In the old case:
    
    * if we take the branch, 1 instruction
    * if we don't take the branch, 3 instruction
    * branch predictor fun
    * 3 instructions of icache pressure
    
    In the new case:
    
    * unconditionally 2 instructions
    * no branch predictor dependence
    * 2 instructions of icache pressure
    
    This should not be non-neglibly worse, and it simplifies things for RA.


Sonicadvance1 (approved):
Nice!
I like how this looks compared to the previous implementation.

PR #3530, opened 2024-03-29
Optimize cmpxchg with flagm

alyssarosenzweig:


Sonicadvance1 (approved):
Nice! This'll feed in to the future `FEAT_THE` and `FEAT_D128` extensions in the future as well!

PR #3532, opened 2024-03-29
Elim xblock live for rol/ror

alyssarosenzweig:


alyssarosenzweig:
Just realized this approach is bogus because of how deferred flags work. Will need to noodle harder.

alyssarosenzweig:
superseded by #3539

PR #3538, opened 2024-03-31
Fix reference to out of bounds address in offsetof

pmatos:


Sonicadvance1:
Looks like the two std::array vectors can drop their fixed length count by one. 44 to 42.

`static constexpr std::array<uint64_t, 44> OffsetArrayAVX = {{`

Nice catch although.

pmatos:
> Looks like the two std::array vectors can drop their fixed length count by one. 44 to 43.
> 
> `static constexpr std::array<uint64_t, 44> OffsetArrayAVX = {{`
> 
> Nice catch although.

Good point, thanks.

Sonicadvance1 (approved):

pmatos:
Squashed - ready to merge.

pmatos:
@Sonicadvance1 added the off-by-one printout fix to this patch.

Sonicadvance1 (approved):

PR #3545, opened 2024-04-01
Use scalar integer code to calculate PF

alyssarosenzweig:
* This is probably fewer cycles for the same # of instructions (?). And thanks to FPU shenanigans, PF calculation isn't completely legacy :upside_down_face: 
* Lets us fold in an invert to save an instruction in `setp` :clown_face:


Sonicadvance1 (approved):
Nice. I love the improvement.
fmov+cnt+umov is quite spicy with both fmov and umov having terrible throughput.
Meanwhile we have 4-8 integer pipelines that can all consume eor/eon.

PR #3547, opened 2024-04-02
x87 Stack Optimization

pmatos:
This effort tries to implement a framework to optimize the x87 stack.

Currently draft, it's a work in progress and many, many tests are still failing. However, it feels like there's a direction so I am opening a PR.

I have been writing a sort of thoughts/design document about this which you can access through:
https://docs.google.com/document/d/1ZIWNuu6h6EuAkMTL70PxfMwx9cZxOZ_OAK849e8y4tM/edit?usp=sharing


pmatos:
**This is not ready for merging.**

Having said that the non-reduced precision work is complete. Things that need to happen before a merge:

* Testing in places other than Psychonauts.
  * HalfLife and Oblivion not running atm under FEX (independently of this branch).
* Integrating reduced precision mode into pass.
* Code cleanup - code has quite a bit of logs and so on that can be removed once the work has come to a close.
  * Still wondering if it's ok to do some simplifications like transforming a pure memcopy into a load/store pair, therefore bypassing the stack. This might not exactly duplicate what's happening natively due to native->80bit and 80bit->native conversions that happen if you go through the stack.
  * There are surely other possibilities for improvement but this is already quite a speedup in my experience.

CallumDev:
Double precision mode in master affects 80-bit and 32-bit loadstores, as it basically switches the "native" operating mode to 64-bit. Conceivably if the stack optimization eliminates stack operations for memcpy-style operations, it could be extended to reduced precision by simply eliminating the stack operation/conversion pairs? As an FLD 80-bit plus FST 80-bit translates to (very basically):

- F80 to Double
- Stack store

- Stack load
- Double to F80

pmatos:
Update: Reduced precision integrated into pass and currently all asm_tests are passing! ✔️ 

There are still a few issues. The plan is:
* Benchmark a few games;
* Code cleanup / refactoring where necessary;
* Land;

It would be interesting to have a few more pairs of eyes over the code. 
@alyssarosenzweig you offered to look at the code a few weeks ago. I think if you have any cleanup/refactoring suggestions, it would be great. I know there's quite a few things that can be improved. We have sort of 4 branches in the stack optimization patch like:

```
if (SlowPath) {
  if (ReducedPrecisionMode) {
    ...
  } else {
    ...
  }
} else {
  if (ReducedPrecisionMode) {
    ...
  } else {
    ...
  }
}
```

We should be able to improve the readability here. Any specific suggestions are very welcome! :) I also have a few TODO/FIXME comments in-code that I need to sort out but otherwise it's going in the right direction.

alyssarosenzweig:
Happy to hear you're making good progress! I'll definitely be reviewing this, although I might not get to it for a bit. (I'm prioritizing AVX review, and I'm out-of-office next week - Monday is a Canadian holiday.)

Sonicadvance1:
Woo! Good job!

pmatos:
While testing this branch on a few games, I noticed a bug on this branch with reduced precision. Currently trying to figure out why we are having a segfault in generated code.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+#include <cstddef>
+#include <cstdint>
+#include <stdint.h>
+
+namespace FEXCore::IR {
+
+// FixedSizeStack is a model of the x87 Stack where each element in this
+// fixed size stack lives at an offset from top. The top of the stack is at
+// index 0.
+template<typename T, size_t MaxSize>
```
alyssarosenzweig, line 19: Why is this a template, instead of just adding a `const size_t MaxSize = 8` and hardcoding the type.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  void dump() const {
+    for (size_t i = 0; i < MaxSize; i++) {
+      const auto& [Valid, Element] = buffer[i];
+      if (Valid == Slot::VALID) {
+        LogMan::Msg::DFmt("ST{}: 0x{:x}", i, (uintptr_t)(Element.StackDataNode));
+      }
+    }
+  }
+
+  constexpr size_t size() const {
```
alyssarosenzweig, line 82: Drop this method, just use a MaxStackSize const throughout the file

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    return MaxSize;
+  }
+
+  int8_t getTopOffset() const {
+    return TopOffset;
+  }
+
+  void setTopOffset(int8_t Offset) {
```
alyssarosenzweig, line 90: This getter/setter pair doesn't seem to add any value instead of just making TopOffset public. Could we drop them?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  void setValidTag(size_t Index, bool Valid) {
+    buffer[Index].first = Valid ? Slot::VALID : Slot::INVALID;
+  }
+
+  // Returns a mask to set in AbridgedTagWord
+  uint8_t getValidMask() {
+    uint8_t Mask = 0;
+    for (size_t i = 0; i < buffer.size(); i++) {
+      const bool Valid = buffer[i].first == Slot::VALID;
+      if (Valid) {
```
alyssarosenzweig, line 103: `if (buffer[i].first == Slot::VALID)`, no need for the variable

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    }
+    return Mask;
+  }
+
+  // Returns a mask to set in AbridgedTagWord
+  uint8_t getInvalidMask() {
+    uint8_t Mask = 0;
+    for (size_t i = 0; i < buffer.size(); i++) {
+      const bool Valid = buffer[i].first == Slot::INVALID;
+      if (Valid) {
```
alyssarosenzweig, line 115: Likewise here

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+      const bool Valid = buffer[i].first == Slot::VALID;
+      if (Valid) {
+        Mask |= 1U << i;
+      }
+    }
+    return Mask;
+  }
+
+  // Returns a mask to set in AbridgedTagWord
+  uint8_t getInvalidMask() {
```
alyssarosenzweig, line 111: Could merge getValidMask and getInvalidMask into a single `getMask(Slot)` helper?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  [[nodiscard]]
+  uint32_t MMBaseOffset() {
+    return static_cast<uint32_t>(offsetof(Core::CPUState, mm[0][0]));
+  }
+  std::tuple<Ref, Ref> SplitF64SigExp(Ref Node);
+
+  // Top Management Helpers
+  /// Set the valid tag for Value as valid (if Valid is true), or invalid (if Valid is false).
+  void SetX87ValidTag(Ref Value, bool Valid);
+  // Generates slow code to load/store a value from the top of the stack
```
alyssarosenzweig, line 152: What is "slow" versus "fast"? If I run the slow path on an M2 Ultra does it become fast 😉 ?

It's ok to introduce language for the pass, but it needs to be defined in a comment towards the top of the file so we're all on the same page.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  Ref LoadStackValueAtTop_Slow();
+  void StoreStackValueAtTop_Slow(Ref Value, bool SetValid = true);
+  // Generates slow code to load/store a value from an offset from the top of the stack
+  Ref LoadStackValueAtOffset_Slow(uint8_t Offset);
+  void StoreStackValueAtOffset_Slow(uint8_t Offset, Ref Value, bool SetValid = true);
+  // Update Top value in slow path for a pop
+  void UpdateTop4Pop_Slow();
```
alyssarosenzweig, line 159: Should this be `UpdateTopForPop`? Or is this really updating the top four entries?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  // Top Cache Management
+  Ref GetTopWithCache_Slow();
+  Ref GetOffsetTopWithCache_Slow(uint8_t Offset);
+  void SetTopWithCache_Slow(Ref Value);
+  Ref GetX87ValidTag_Slow(uint8_t Offset);
+  // Resets fields to initial values
+  void Reset(bool AlsoSlowPath = true);
+
+  struct StackMemberInfo {
+    IR::OpSize SourceDataSize; // Size of SourceDataNode
```
alyssarosenzweig, line 176: You might need a `using IR` at the top of the file, but there should not be any need for `IR::` in the pass.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    IR::OpSize StackDataSize;  // Size of the data in the stack
+    IR::Ref SourceDataNode;    // Reference to the source location of the data.
+                               // In the case of a load, this is the source node of the load.
+                               // If it's not a load, then it's nullptr.
+    IR::Ref StackDataNode;     // Reference to the data in the Stack.
+    bool InterpretAsFloat {};  // True if this is a floating point value, false if integer
+  };
+
+  // StackData, TopCache need to be always properly set to ensure
+  // it reflects the current state of the FPU. This sync only makes sense while
```
alyssarosenzweig, line 186: "it reflects" -> "they reflect"?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  // taking the fast path. Once in the slow path, these don't make sense anymore
+  // and we are syncing everything.
+
+  // Index on vector is offset to top value at start of block
+  // If slow path is true, then StackData is always empty.
+  FixedSizeStack<StackMemberInfo, 8> StackData;
+  using StackSlot = FixedSizeStack<StackMemberInfo, 8>::Slot;
```
alyssarosenzweig, line 193: Could we just define the `enum class StackSlot` at top-level instead of nested in FixedSizeStack? That avoids special renaming.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+
+  void InvalidateCaches();
+  void InvalidateTopOffsetCache();
+
+  // Cache for Constants
+  // ConstantPoll[i] has IREmit->_Constant(i);
+  constexpr static size_t ConstantPoolSize = 8;
```
alyssarosenzweig, line 200: In this case, we use `ConstantPool.size()` so we don't need the extra constant here, we can just hardcode the8 into the ConstantPool on the next line.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+  // we can implement that but I would expect that there would be very few cases where that's necessary.
+  // On the slow path TopCache is always the last obtained version of top.
+  // TopOffset is ignored
+  bool SlowPath = false;
+  // Keeping IREmitter not to pass arguments around
+  IREmitter* IREmit = nullptr;
+};
+
+inline void X87StackOptimization::InvalidateCaches() {
+  LogMan::Msg::DFmt("Invalidating caches");
```
alyssarosenzweig, line 219: I expect most of the debug prints should be removed, but I don't have a good sense of this pass yet so not sure.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    // not dealt by pool
+    LogMan::Msg::DFmt("Generating uncacheable constant for {}", Offset);
+    return IREmit->_Constant(Offset);
+  }
+  if (ConstantPool[Offset] == nullptr) {
+    LogMan::Msg::DFmt("Generating and caching constant for {}", Offset);
+
+    ConstantPool[Offset] = IREmit->_Constant(Offset);
+  } else {
+    LogMan::Msg::DFmt("Returning cached constant for {}", Offset);
```
alyssarosenzweig, line 248: drop the prints in this function

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+
+inline void X87StackOptimization::Reset(bool AlsoSlowPath) {
+  if (AlsoSlowPath) {
+    SlowPath = false;
+  }
+  StackData.clear();
+  InvalidateCaches();
+}
+
+inline Ref X87StackOptimization::GetConstant(ssize_t Offset) {
```
alyssarosenzweig, line 237: We already have a general-purpose constant pooling optimization in ConstProp, this is strictly a specific case of it. Does it make sense to have both? (for example to improve compile-time?)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+
+inline void X87StackOptimization::MigrateToSlowPathIf(bool ShouldMigrate) {
+  if (SlowPath) {
+    return;
+  }
+  if (!ShouldMigrate) {
+    return;
+  }
+
+  LogMan::Msg::DFmt("Migrating to SlowPath");
```
alyssarosenzweig, line 261: Wrap in `if (ShouldMigrate && !SlowPath)`, no need for any `return`s.

alyssarosenzweig (changes requested):
(Not done reviewing yet but left you some comments at the beginning of the file)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+public:
+  enum class Slot { UNUSED, INVALID, VALID };
+  FixedSizeStack() {
+    buffer.resize(MaxSize, {Slot::UNUSED, T()});
+  }
+
+  void push(const T& value) {
+    rotate();
+    buffer.front() = {Slot::VALID, value};
+  }
```
neobrain, line 30: This all seems very contrived, seemingly with the intent of improving performance, but `std::deque` is a heavyweight container that interferes with data caching and branch prediction.

Have you considered just using an `std::array` and reversing the order elements are stored internally? That would make the implementation a lot more straightforward while having little (if any) impact on performance: Indexing in `top` will be somewhat faster, and `rotate` slightly slower (not by much since we're talking about tiny amounts of data though).

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
neobrain: Making it a template highlights that this is a generic container that doesn't peek into any of `StackMemberInfo`'s internal state. This is a simple template use so I don't see much of a downside.

alyssarosenzweig: fair, I don't feel strongly on this one either way.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    if (Mask == 0xff) {
+      IREmit->_StoreContext(1, GPRClass, GetConstant(Mask), offsetof(FEXCore::Core::CPUState, AbridgedFTW));
+    } else if (Mask != 0) {
+      if (std::popcount(Mask) == 1) {
+        uint8_t BitIdx = __builtin_ctz(Mask);
+        SetX87ValidTag(GetOffsetTopWithCache_Slow(BitIdx), true);
+      } else {
+        // perform a rotate right on mask by top
+        // since we must operate on 32bits as a minimum:
+        // ror (mask, top) = ((M << 8) | M) >> TOP
```
alyssarosenzweig, line 403: Any reason you can't _Ror?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+
+std::tuple<Ref, Ref> X87StackOptimization::SplitF64SigExp(Ref Node) {
+  Ref Gpr = IREmit->_VExtractToGPR(8, 8, Node, 0);
+
+  Ref Exp = IREmit->_And(OpSize::i64Bit, Gpr, GetConstant(0x7ff0000000000000LL));
+  Exp = IREmit->_Lshr(OpSize::i64Bit, Exp, GetConstant(52));
+  Exp = IREmit->_Sub(OpSize::i64Bit, Exp, GetConstant(1023));
+  Exp = IREmit->_Float_FromGPR_S(8, 8, Exp);
+  Ref Sig = IREmit->_And(OpSize::i64Bit, Gpr, GetConstant(0x800fffffffffffffLL));
+  Sig = IREmit->_Or(OpSize::i64Bit, Sig, GetConstant(0x3ff0000000000000LL));
```
alyssarosenzweig, line 451: We know that none of the constants in this function will hit this pass's constant cache, might as well use _Constant() directly

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1892 @@
...
+    Ref MaskC = GetConstant(Mask);
+    LogMan::Msg::DFmt("Writing invalid tags 0x{:02x}", Mask);
+    if (Mask == 0xff) {
+      IREmit->_StoreContext(1, GPRClass, GetConstant(0), offsetof(FEXCore::Core::CPUState, AbridgedFTW));
+    } else if (Mask != 0) {
+      if (std::popcount(Mask)) {
+        uint8_t BitIdx = __builtin_ctz(Mask);
+        SetX87ValidTag(GetOffsetTopWithCache_Slow(BitIdx), false);
+      } else {
+        // Same rotate right as above but this time on the invalid mask
```
alyssarosenzweig, line 427: _Ror here too..?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: If i do that then it makes no sense having FixedSizeStack as a template. What's the downside here of using `size()`?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Yes, that's true - they will be removed. But I will leave those until before the merge. :)

Just please ignore them until we are close to merge.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Not really - we could potentially abstract into a ConstantPool class, what do you think?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: My initial impl used an array but for some reason that is escaping me now, I then switched to a deque. I will need to think a bit more about why I switched and the considerations here. There are quite a few corner cases that made using a deque easier.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: The thought was that if the Constant cache is increased, to a certain point then this would be automatically caught in the future. And if it doesn't then I don't expect it to have performance impact.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: My ignorance maybe. :) 
Can we do an 8bit ror? I assumed it was only 32 or 64 bits rotate that were possible.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: No, the logic is different anyway. I don't mean that "we shouldn't have 2 separate implementations of constant pooling logic". Rather that "we're going to have constant pooling happen anyway if the pass order is right, so is it worth also pooling constants here ahead of time?"

(The answer may well be yes.)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: Per discord - indeed we can't do an 8-bit ror. But we can use `_Orlshl` to cut an instruction.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: I guess this duplicates the thread of "why do we even have a constant cache here, does this actually make a ton of sense?" then..

alyssarosenzweig:
What happened to Comiss()?

pmatos:
> What happened to Comiss()?

I had to break Comiss() apart - sorry Comiss...
The issue I had was that I need to do part of what Comiss is doing in `X87F64.cpp`. So in `X87F64.cpp` I call `ComissFlags` which is the flag dealing part of Comiss. The `_FCmp` part is dealt with in Stack optimization. I think I can keep `Comiss` actually but still have it call `ComissFlags`.

pmatos:
> > What happened to Comiss()?
> 
> I had to break Comiss() apart - sorry Comiss... The issue I had was that I need to do part of what Comiss is doing in `X87F64.cpp`. So in `X87F64.cpp` I call `ComissFlags` which is the flag dealing part of Comiss. The `_FCmp` part is dealt with in Stack optimization. I think I can keep `Comiss` actually but still have it call `ComissFlags`.

@alyssarosenzweig ok - `Comiss` is whole again. :)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: `getValidMask` was never used so it's gone. :)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Had to laugh at that comment. :) Renamed it. No more 4 instead of 'For' or 2 instead of 'To'.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Actually I didn't even need that cause the namespace is already `FEXCore::IR`.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Done.

pmatos:
Part of the PR now contains #3872 since it's rebased on it.
Once that lands, I can clean it up. 

GH, where's stacked PR support?

alyssarosenzweig:
Just merged the other PR for you

pmatos:
> Just merged the other PR for you

Rebased on main now - ready for review.

alyssarosenzweig:
There are still open coded `CurrentHeader->HasX87 = true;`, these should all be deleted I think?

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -72,757 +55,313 @@ void OpDispatchBuilder::SetX87FTW(Ref FTW) {
...
+  if (Op->Src[0].IsNone()) {
+    const auto offset = Op->OP & 7;
+    const auto st0 = 0;
 
-  auto orig_top = GetX87Top();
-  // if we are popping then we must first mark this location as empty
-  SetX87ValidTag(orig_top, false);
-  auto top = _And(OpSize::i32Bit, _Add(OpSize::i32Bit, orig_top, _Constant(1)), _Constant(7));
-  SetX87Top(top);
+    if (Reverse) {
```
alyssarosenzweig, line 256: can combine the conditions: `if (Reverse ^ (ResInST0 == OpResult::RES_STI))`

and then drop the latter half

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -72,757 +55,313 @@ void OpDispatchBuilder::SetX87FTW(Ref FTW) {
...
-  Ref data = LoadAndCacheNamedVectorConstant(16, NamedVectorConstant::NAMED_VECTOR_X87_ONE);
-
-  // TODO: ACCURACY: should check source is in range –2^63 to +2^63
-  SetRFLAG<FEXCore::X86State::X87FLAG_C2_LOC>(_Constant(0));
-
-  // Write to ST[TOP]
-  _StoreContextIndexed(result, orig_top, 16, MMBaseOffset(), 16, FPRClass);
-  _StoreContextIndexed(data, top, 16, MMBaseOffset(), 16, FPRClass);
-}
+    if (Reverse) {
```
alyssarosenzweig, line 307: likewise here

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
alyssarosenzweig: I guess you'd still need a `Ref Result = (ResInST0 == OpR....) ? offset : st0;` or something. But given my eyes glazed over that difference the first time, that'd probably be more readable anyhow.

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -970,43 +553,52 @@ void OpDispatchBuilder::X87FNSAVE(OpcodeArgs) {
...
-  auto NewFSW = _LoadMem(GPRClass, Size, Mem, _Constant(Size * 1), Size, MEM_OFFSET_SXTX, 1);
-  auto Top = ReconstructX87StateFromFSW(NewFSW);
+  Ref MemLocation = _Add(OpSize::i64Bit, Mem, _Constant(Size * 1));
+  auto NewFSW = _LoadMem(GPRClass, Size, MemLocation, Size);
+  Ref Top = ReconstructX87StateFromFSW_Helper(NewFSW);
   {
     // FTW
-    SetX87FTW(_LoadMem(GPRClass, Size, Mem, _Constant(Size * 2), Size, MEM_OFFSET_SXTX, 1));
+    Ref MemLocation = _Add(OpSize::i64Bit, Mem, _Constant(Size * 2));
+    SetX87FTW(_LoadMem(GPRClass, Size, MemLocation, Size));
```
alyssarosenzweig, line 580: Why are we changing this? I specifically folded in the Adds into these instructions a few months ago, this is a functional revert of that. Why?

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87F64.cpp
```diff
@@ -228,468 +180,267 @@ void OpDispatchBuilder::FISTF64(OpcodeArgs, bool Truncate) {
...
-  auto low = _Constant(0);
-  Ref data = _VCastFromGPR(8, 8, low);
+  CurrentHeader->HasX87 = true;
 
   // We are going to clobber NZCV, make sure it's in a GPR first.
   GetNZCV();
 
   // Now we do our comparison.
-  _FCmp(8, a, data);
+  _F80StackTest(0, 0 /*No flags in reduced precision mode needed*/);
```
alyssarosenzweig, line 387: Why is that?

alyssarosenzweig: Never mind. I get it. #3875

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
pmatos: Nice catch!

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
pmatos: I went through your commit at the time and found many more places where this happened, so I fixed those as well. :-(

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1338 @@
...
+// stack overview. However, if we encounter a value that's not in the virtual stack - maybe it was added
+// to the stack in a previous block, we move onto the slow path which loads and stores values to the stack
+// registers.
+// Once in a slow path, we won't return to the fast pass until the beginning of the following block.
+
+namespace FEXCore::IR {
+
+// Similar helper to the one in OpcodeDispatcher.h except we do not
+// need to handle flags, etc.
+#define DeriveOp(RefV, NewOp, Expr)                                                   \
```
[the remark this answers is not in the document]
pmatos: Yes, that makes sense. I have changed that now.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1338 @@
...
+    for (size_t i = 0; i < buffer.size(); i++) {
+      if (buffer[i].first == StackSlot::INVALID) {
+        Mask |= 1U << i;
+      }
+    }
+    return Mask;
+  }
+
+private:
+  fextl::deque<std::pair<StackSlot, T>> buffer;
```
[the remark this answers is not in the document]
pmatos: Good point - I changed this now to use std::array. 
Initially I was trying to use array but for some reason it didn't work as I expected or I thought rotation was too much code - can't recall. In any case, using std::array + std::rotate seems to do the trick.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1331 @@
...
+      if (IsBlockExit(IROp->Op)) {
+        // Set write cursor to previous instruction
+        IREmit->SetWriteCursor(IREmit->UnwrapNode(CodeNode->Header.Previous));
+        ExitFound = true;
+        break;
+      }
+    }
+    LOGMAN_THROW_A_FMT(ExitFound, "No block exit found is bad!");
+    SynchronizeStackValues();
+    IREmit->SetWriteCursor(OriginalWriteCursor);
```
alyssarosenzweig, line 1322: drop this, no need to save/restore cursor

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1331 @@
...
+
+      default: break;
+      }
+    }
+
+    // We need to write the registers before any branch in the block,
+    // so loop until a branch instruction is found. Add instructions _before_
+    // the branch instruction.
+    [[maybe_unused]] bool ExitFound = false; // used for assertion
+    for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
```
alyssarosenzweig, line 1312: A block exit is always the last instruction in a block before the block exit - no need to loop the instructions to find it, that's O(n) when we can just grab the instruction O(1). Something like:

```cpp
      auto Last = CurrentIR.at(BlockIROp->Last);
      --Last;
      LOGMAN_THROW_A_FMT(IsBlockExit(Last->Op), "must be exit");
      IREmit->SetWriteCursorBefore(Last);
```

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: TBH, with a view to future IR work, I think I'd weakly prefer if we drop the pooling and just use `_Constant` throughout the file to remain consistent with the OpcodeDispatcher & the rest of FEX. This feels like a premature optimization.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: That does not seem to be the case?

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -1011,36 +540,203 @@ void OpDispatchBuilder::X87FRSTOR(OpcodeArgs) {
...
+// Top is 3 bits at bit 11.
+// C0 is 1 bit at bit 8.
+// C1 is 1 bit at bit 9.
+// C2 is 1 bit at bit 10.
+// C3 is 1 bit at bit 14.
+// Optionally we can pass a pre calculated value for Top, otherwise we calculate it
+// during the function runtime.
+Ref OpDispatchBuilder::ReconstructFSW_Helper(Ref T) {
+  // We must construct the FSW from our various bits
+  Ref FSW = _Constant(0);
```
alyssarosenzweig, line 685: I optimized this yesterday, please backport the change :)

pmatos: 8xc51s

pmatos: Sorry about this - fixed.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Didn't know I could do that. Thanks.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Perfect!

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      default: break;
+      }
+    }
+
+    // We need to write the registers before any branch in the block,
+    // so loop until a branch instruction is found. Add instructions _before_
```
alyssarosenzweig, line 1307: stale comment

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        }
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_DECSTACKTOP: {
+        if (SlowPath) {
+          UpdateTopForPush_Slow();
+        } else {
+          StackData.rotate(true);
```
alyssarosenzweig, line 1242: combine with INCSTAKCKTOP and just `rotate(op == OP_DECSTACKTOP)`

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+      }
+
+      case OP_F80XTRACTSTACK: {
+        const auto& [Valid, StackMember] = StackData.top();
+
+        MigrateToSlowPathIf(Valid != StackSlot::VALID);
+
+        if (SlowPath) {
+          Ref st0 = LoadStackValueAtTop_Slow();
+          Ref Exp {};
```
alyssarosenzweig, line 1175: Probably want to `st0 = SlowPath ? Load() : StackDataNode` outside the if, then you can deduplicate the splitting logic and just have the if for writing it out

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        IREmit->ReplaceUsesWithAfter(CodeNode, CmpNode, CodeNode);
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+
+      case OP_F80CMPVALUE: {
+        const auto* Op = IROp->C<IROp_F80CmpValue>();
+        const auto& Value = CurrentIR.GetNode(Op->X80Src);
+        const auto& [Valid, StackMember] = StackData.top();
```
alyssarosenzweig, line 1143: I think we want to factor out a little helper that does

```cpp
const auto& [Valid, StackMember] = StackData.top();
MigrateToSlowPathIf(Valid != StackSlot::VALID);
return [Valid, StackMember];
```

then use that throughout the file. this should make most of the migrates go away

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+
+        MigrateToSlowPathIf(Valid != StackSlot::VALID);
+
+        Ref CmpNode {};
+        if (SlowPath) {
```
alyssarosenzweig, line 1148: I think we want to pull out a helper `LoadStackValue()` that returns

```
SlowPath ? LoadStackValueAtTop_Slow() : StackMember.StackDataNode
```

Then use that throughout the file. It means we can totally deduplicate/collapse lots of if-statements, like this one here

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        if (SlowPath) { // slow path
+          Ref ResultNode {};
+          Ref Value = LoadStackValueAtTop_Slow();
+          // Negate value
+          if (ReducedPrecisionMode) {
+            ResultNode = IREmit->_VFNeg(8, 8, Value);
+          } else {
+            ResultNode = IREmit->_VXor(16, 1, Value, HelperNode);
+          }
+          StoreStackValueAtTop_Slow(ResultNode, false);
```
alyssarosenzweig, line 1029: I think we want to extract a `SetTop(Ref)` helper that does

```
if (SlowPath) {
   StoreStackValueAtTop_Slow(Ref, false);
} else {
   StackData.setTop(StackMemberInfo {
            .SourceDataSize = StackMember.SourceDataSize,
            .SourceDataNode = nullptr,
            .StackDataNode = Ref
   });
}
```

Along with the `LoadStackTop()` helper I proposed below, this will make this if/else go away and deduplicate the logic.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        auto offset = Op->SrcStack;
+        const auto& [Valid1, StackMember1] = StackData.top();
+        const auto& [Valid2, StackMember2] = StackData.top(offset);
+
+        MigrateToSlowPathIf(Valid1 != StackSlot::VALID || Valid2 != StackSlot::VALID);
+
+        Ref CmpNode {};
+        if (SlowPath) {
+          Ref StackValue1 = LoadStackValueAtTop_Slow();
+          Ref StackValue2 = LoadStackValueAtOffset_Slow(offset);
```
alyssarosenzweig, line 1093: Similarly the  LoadStackValue helper should take an optional (default = 0) offset so we can use it here too. Again, that either does the slow path or the fast top(offset). And of course that top(offset) call would migrate automatically. So putting it all together, this switch case will look like:


```cpp
      case OP_F80CMPSTACK: {
        const auto* Op = IROp->C<IROp_F80CmpStack>();
        const auto Value1 = LoadStackValue();
        const auto Value2 = LoadStackValue(Op->SrcStack);

        Ref CmpNode = ReducedPrecisionMode ? IREmit->_FCmp(8, Value1, Value2) : 
                                                                                   IREmit->_F80Cmp(Value1, Value2;
        IREmit->ReplaceUsesWithAfter(CodeNode, CmpNode, CodeNode);
        IREmit->Remove(CodeNode);
        break;
      }
```

...which I'm sure you'll agree is a lot nicer!

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+          Ref Sig {};
+          if (ReducedPrecisionMode) {
+            std::tie(Exp, Sig) = SplitF64SigExp(st0);
+          } else {
+            Exp = IREmit->_F80XTRACT_EXP(st0);
+            Sig = IREmit->_F80XTRACT_SIG(st0);
+          }
+          // Write exp to top, update top for a push and set sig at new top.
+          StoreStackValueAtTop_Slow(Exp, false);
+          UpdateTopForPush_Slow();
```
alyssarosenzweig, line 1185: Likewise probably want a StackPush helper for something like:

```
if (Slow) {
          UpdateTopForPush_Slow();
          StoreStackValueAtTop_Slow(Ref);
} else {
         StackData.push(StackMemberInfo {.SourceDataSize = StackMember.SourceDataSize,
                                          .SourceDataNode = nullptr,
                                          .StackDataNode = Ref,
                                          .InterpretAsFloat = StackMember.InterpretAsFloat});
```

and then with the other helpers the branching here all goes away

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+        if (SlowPath) {
+          UpdateTopForPush_Slow();
+        } else {
+          StackData.rotate(true);
+        }
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_F80ROUNDSTACK: {
```
alyssarosenzweig, line 1248: This is going to look a *lot* shorter once it's converted to the helpers :+1:

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+      }
+
+      case OP_STORESTACKMEMORY: {
+        const auto* Op = IROp->C<IROp_StoreStackMemory>();
+
+        const auto& [Valid, Value] = StackData.top();
+        MigrateToSlowPathIf(Valid != StackSlot::VALID);
+
+        Ref StackNode {};
+        if (SlowPath) {
```
alyssarosenzweig, line 893: Here we're open coding that `LoadStackValue()` helper again

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+            .SourceDataSize = StackMember.SourceDataSize, .SourceDataNode = nullptr, .StackDataNode = SinValue, .InterpretAsFloat = true});
+          StackData.push(StackMemberInfo {
+            .SourceDataSize = StackMember.SourceDataSize, .SourceDataNode = nullptr, .StackDataNode = CosValue, .InterpretAsFloat = true});
+        }
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_INITSTACK: {
+
```
alyssarosenzweig, line 775: drop this blank line (there are a handful of cases like this throughout the patch)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+    // Each time we deal with a new block we need to start over.
+    // The optimization should run per-block
+    Reset();
+
+    for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
+      IREmit->SetWriteCursor(CodeNode);
+
+      switch (IROp->Op) {
+
+        // Binary ops on the stack
```
alyssarosenzweig, line 583: I think now that they're going through named helpers (HandleBinopStack) we can drop this comment and the blank line above it

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+          UpdateTopForPop_Slow();
+        } else {
+          StackData.pop();
+        }
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+        // Binary ops with a non-stack argument
+
```
alyssarosenzweig, line 653: Likewise here

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+      }
+
+      case OP_F80MULVALUE: {
+        const auto* Op = IROp->C<IROp_F80MulValue>();
+        HandleBinopValue(OP_VFMUL, true, OP_F80MUL, 0, true, Op->SrcStack, CurrentIR.GetNode(Op->X80Src));
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      // Unary Operations
```
alyssarosenzweig, line 684: Likewise

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+      case OP_F80PTANSTACK: {
+        HandleUnop(OP_F64TAN, false, OP_F80TAN);
+        Ref OneConst {};
+        if (ReducedPrecisionMode) {
+          OneConst = IREmit->_VCastFromGPR(8, 8, GetConstant(0x3FF0000000000000));
+        } else {
+          OneConst = IREmit->_LoadNamedVectorConstant(16, NamedVectorConstant::NAMED_VECTOR_X87_ONE);
+        }
+
+        if (SlowPath) {
```
alyssarosenzweig, line 719: Here we're open coding that StackPush helper again :)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+                                          .InterpretAsFloat = false});
+        }
+
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+        // Miscellaneous Ops
+        // These are handled on a case-by-case basis
+      case OP_F80SINCOSSTACK: {
```
alyssarosenzweig, line 735: This guy's going to be cut in half with the new helpers :)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1323 @@
...
+
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_PUSHSTACK: {
+        const auto* Op = IROp->C<IROp_PushStack>();
+        auto* SourceNode = CurrentIR.GetNode(Op->X80Src);
+
+        if (SlowPath) {
```
alyssarosenzweig, line 814: The prototypical `StackPush()` :>

alyssarosenzweig:
Over in #3876, I've added an autogenerated `LoweredX87()` helper that will tell you if an opcode is marked `"X87": true`. The reason I did this is two fold:

1. Correctness. We need to tag IR ops iff they are lowered here. By putting in an `if (!LoweredX87(IROp->Op)) continue;` in front of the switch, we can then assert that we never hit the `default` - if we did, the pass is out of sync with the JSON. So there's a bit of safety there.
2. Compactness. Once we put in that little `if`, we know we are guaranteed to lower the instruction. Now you can do a single `Remove()` after the switch, instead of needing a separate Remove call in every single case.

So making those changes should let you delete a few dozen LOC with little effort :)

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Nice - yes, that makes total sense.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Lots of these indeed.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Think this is sorted.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Yes to these and all the others like it. :)

pmatos:
> Over in #3876, I've added an autogenerated `LoweredX87()` helper that will tell you if an opcode is marked `"X87": true`. The reason I did this is two fold:
> 
> 1. Correctness. We need to tag IR ops iff they are lowered here. By putting in an `if (!LoweredX87(IROp->Op)) continue;` in front of the switch, we can then assert that we never hit the `default` - if we did, the pass is out of sync with the JSON. So there's a bit of safety there.
> 2. Compactness. Once we put in that little `if`, we know we are guaranteed to lower the instruction. Now you can do a single `Remove()` after the switch, instead of needing a separate Remove call in every single case.
> 
> So making those changes should let you delete a few dozen LOC with little effort :)

OK - thanks for this. I have finished revising stuff. I think things are looking much better and simpler. I will run some games to make sure that the refactoring didn't break anything. There might be a couple of methods that might not be used anymore which I will still remove if that's the case. Otherwise, let me know if you have further comments.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+inline std::optional<X87StackOptimization::StackMemberInfo> X87StackOptimization::MigrateToSlowPath_IfInvalid(uint8_t Offset) {
+  const auto& [Valid, StackMember] = StackData.top(Offset);
+  MigrateToSlowPathIf(Valid != StackSlot::VALID);
+  if (Valid == StackSlot::VALID) {
+    return StackMember;
+  }
+  return {};
+}
+
+inline Ref X87StackOptimization::LoadStackValue(std::optional<StackMemberInfo>& StackValue, uint8_t Offset) {
```
alyssarosenzweig, line 366: Can we change the signature of this to simply

```
inline Ref X87StackOptimization::LoadStackValue(uint8_t Offset) {
```

and then include a call of `MigrateToSlowPath_IfInvalid` inside? Then all the Migrate calls goa way

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+        // If we are doing:
+        // fld dword [rax]
+        // fst dword [rbx]
+        // We can optimize this to:
+        // ldr w2, [x0]
+        // str w2, [x1]
+        // or similar. As long as the source size and dest size are one and the same.
+        // This will avoid any conversions between source and stack element size and conversion back.
+        if (!SlowPath && Value->Source && Value->Source->first == Op->StoreSize && Value->InterpretAsFloat) {
+          LogMan::Msg::DFmt("Performing memcpy optimization from stack");
```
Sonicadvance1, line 850: Make sure this log is removed before merging since it shows up very frequently when running 32-bit games

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+    for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
+      if (!LoweredX87(IROp->Op)) {
+        continue;
+      }
+      IREmit->SetWriteCursor(CodeNode);
+      switch (IROp->Op) {
+      case OP_F80ADDSTACK: {
+        const auto* Op = IROp->C<IROp_F80AddStack>();
+        HandleBinopStack(OP_VFADD, true, OP_F80ADD, Op->SrcStack1, Op->SrcStack1, Op->SrcStack2);
+        IREmit->Remove(CodeNode);
```
alyssarosenzweig, line 571: All of these Remove()s can be removed, in favour of a single Remove() after the switch.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+        break;
+      }
+
+      case OP_F80FYL2XSTACK: {
+        HandleBinopStack(OP_F64FYL2X, false, OP_F80FYL2X, 1, 0, 1);
+        if (SlowPath) {
+          UpdateTopForPop_Slow();
+        } else {
+          StackData.pop();
+        }
```
alyssarosenzweig, line 620: This should be a helper

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+          HelperNode = IREmit->_VInsGPR(16, 8, 1, HelperNode, High);
+        }
+
+        // Negate value
+        Ref ResultNode {};
+        if (ReducedPrecisionMode) {
+          ResultNode = IREmit->_VFNeg(8, 8, Value);
+        } else {
+          ResultNode = IREmit->_VXor(16, 1, Value, HelperNode);
+        }
```
alyssarosenzweig, line 957: You'll want to combine these ifs now that we've deduplicated things.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+          HelperNode = IREmit->_VCastFromGPR(16, 8, Low);
+          HelperNode = IREmit->_VInsGPR(16, 8, 1, HelperNode, High);
+        }
+
+        Ref ResultNode {};
+        if (ReducedPrecisionMode) {
+          ResultNode = IREmit->_VFAbs(8, 8, Value);
+        } else {
+          ResultNode = IREmit->_VAnd(16, 1, Value, HelperNode);
+        }
```
alyssarosenzweig, line 981: Likewise here

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+      case OP_STACKFORCESLOW: {
+        MigrateToSlowPathIf(true);
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_INCSTACKTOP:
+      case OP_DECSTACKTOP: {
+        if (SlowPath) {
+          if (IROp->Op == OP_DECSTACKTOP) {
```
alyssarosenzweig, line 1085: Oops, I didn't see that there was a push vs pop distinction. I guess I shouldn't've said to combine these, my mistake! 😿

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+        auto* VecCond = CurrentIR.GetNode(Op->VectorMask);
+        auto StackOffset1 = Op->SrcStack1;
+        auto StackOffset2 = Op->SrcStack2;
+        auto ValueOffset1 = MigrateToSlowPath_IfInvalid(StackOffset1);
+        auto ValueOffset2 = MigrateToSlowPath_IfInvalid(StackOffset2);
+        Ref Value1 = LoadStackValue(ValueOffset1, StackOffset1);
+        Ref Value2 = LoadStackValue(ValueOffset2, StackOffset2);
+
+        Ref StackNode = IREmit->_VBSL(16, VecCond, Value1, Value2);
+        StoreStackValue(StackNode, 0, StackOffset1 && StackOffset2);
```
alyssarosenzweig, line 1125: I don't understand the && here.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_F80VBSLSTACK: {
+        const auto* Op = IROp->C<IROp_F80VBSLStack>();
+
+        // Multiplies two elements in the stack by offset.
+        auto* VecCond = CurrentIR.GetNode(Op->VectorMask);
+        auto StackOffset1 = Op->SrcStack1;
```
alyssarosenzweig, line 1117: No sense making temporaries for these IMHO

alyssarosenzweig:
It seems like a bunch of InterpretAsFloats might've gotten lost in the latest refactor. The memcpy instcountci blocks, for example, had a bunch of conversions in them yesterday that are not there now. Was that intended?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: That's a good sign. :) I mean - that games actually use that, not necessarily that you're flooded with messages.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1148 @@
...
+        }
+        StoreStackValue(Value);
+        IREmit->Remove(CodeNode);
+        break;
+      }
+
+      case OP_F80VBSLSTACK: {
+        const auto* Op = IROp->C<IROp_F80VBSLStack>();
+
+        // Multiplies two elements in the stack by offset.
```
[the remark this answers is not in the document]
pmatos: Yep - thanks

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Sounds good. Only one case I left as is because we needed the value from `Migrate...`.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: ah - because now we continue at the top if it's not lowered. Nice!

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: I think we only have 3 of these. Does it really make sense here?

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: I guess it's `StackOffset1 != 0 && StackOffset2 != 0`. The point is, we are storing to the top of the stack. We only really need to mark it as valid if none of the values come from the top of the stack. Because if they did, we would know it's already valid and there's no point in re-marking it.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: VecCond can be removed even though I think it aid readability. The others are used in a couple of places so they make sense to stay.

pmatos:
> It seems like a bunch of InterpretAsFloats might've gotten lost in the latest refactor. The memcpy instcountci blocks, for example, had a bunch of conversions in them yesterday that are not there now. Was that intended?

I simplified a few things yesterday but I am surprised you say that some conversions are gone. I will take a closer look but when I skimmed through the instcounci results, it looked good to me.

neobrain:
> I simplified a few things yesterday but I am surprised you say that some conversions are gone. I will take a closer look but when I skimmed through the instcounci results, it looked good to me.

GitHub often drops conversations if you update the underlying code. There's nothing really you can do about it (other than enabling email notifications so you can look things up manually), annoyingly.

pmatos:
> > It seems like a bunch of InterpretAsFloats might've gotten lost in the latest refactor. The memcpy instcountci blocks, for example, had a bunch of conversions in them yesterday that are not there now. Was that intended?
> 
> I simplified a few things yesterday but I am surprised you say that some conversions are gone. I will take a closer look but when I skimmed through the instcounci results, it looked good to me.

I know what happened here. I add the memcpy insts in the commit where I add the tests and the third commit with instcounci modifies those. A recent change correctly removed the fcvt. I think I will add these instcountci tests in a separate PR instead.

pmatos:
> > > It seems like a bunch of InterpretAsFloats might've gotten lost in the latest refactor. The memcpy instcountci blocks, for example, had a bunch of conversions in them yesterday that are not there now. Was that intended?
> > 
> > 
> > I simplified a few things yesterday but I am surprised you say that some conversions are gone. I will take a closer look but when I skimmed through the instcounci results, it looked good to me.
> 
> I know what happened here. I add the memcpy insts in the commit where I add the tests and the third commit with instcounci modifies those. A recent change correctly removed the fcvt. I think I will add these instcountci tests in a separate PR instead.

Now in #3880

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
alyssarosenzweig: Sometimes helpers are useful even if they're only called once for the sake of expressing the intention clearly and separating the "manage the stack" logic from the "implement all the instructions" 🤷

pmatos: sgtm - added helper.

alyssarosenzweig:
For FixedSizeStack, why are we using a fextl::vector and not simply a static array? We know it will always be  8 elements - no need for the dynamic allocation, right?

alyssarosenzweig (approved):
Approved. There are still little things that could be improved, but at this point, the code looks structurally sound and we're hitting diminishing returns with prolonging the review process versus the benefit of getting this in-tree and soak-tested. Let's go :+1:

pmatos:
To be honest, it's I started by attempting to use an std::array but had issues with allocation due to being unable to construct an initial array with the values I needed. I then decided to move to an std vector where I could use reserve and fill methods.

pmatos:
Nice! Thanks for all the time you spent carefully reviewing this and providing great feedback.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -0,0 +1,1069 @@
...
+  // update top straight away the TopOffset is 1.
+  // If SlowPath is true, then TopOffset is always zero.
+  int8_t TopOffset = 0;
+
+  FixedSizeStack()
+    : buffer(FixedSizeStack::size, {StackSlot::UNUSED, T()}) {}
+
+  void push(const T& Value) {
+    rotate();
+    buffer.front() = {StackSlot::VALID, Value};
```
neobrain, line 65: Now that we're using `vector`, we can invert the storage order of the elements to simplify the implementation. Instead of `rotate` + `assign to front()`, we can just use this:
```suggestion
    buffer.push_back({StackSlot::VALID, Value});
```
Similarly, pop becomes `pop_back()`, and `StackSlot::UNUSED` can be entirely removed.

Internally, the buffer won't be fixed-size anymore then, but that's fine. Calling `reserve(size)` on construction ensures there will never be more than one heap allocation.

pmatos: I understand the suggestion but the location where the items are valid correspond to an offset to TOP, which will stop being the case. I understand that this optimization is possible, but I think it's going to complicate the code readability elsewhere. I will take a look to see exactly what such a change would look like and report back. Thanks.

pmatos: Another thing is that we cannot really rmeove StackSlot::UNUSED. This is because the stack doesn't need to be contiguous, i.e. some elements might be unused. So you do need to mark that. For example:
```
fld qword [addr] ; 1.0
fst st2
```

In this case the stack will look like:
```
MM1  1.0      ST2
MM0  UNUSED   ST1
MM7  1.0      ST0 <- TOP
```

So we need to mark these in-between slots as unused.

neobrain: Are we baking pointers within the FixedSizeStack into the generated assembly code? Otherwise I don't understand why you're bringing up assembly code here.

The change I'm proposing is an implementation detail and shouldn't affect any of the users of the FixedSizeStack interface.

pmatos: Maybe I am confused about your suggestion, but I cannot see how your suggestion would work. The "stack" is not really a stack, which makes it harder to implement as you suggest. This is because I can push an element to the stack but then set the element that's two elements above that. With your proposed solution where the vector starts empty a push would be an emplace_back but then setting ST2, would mean a rotate right by one and emplacing back. It's certainly possible but I am not sure it's an improvement.

neobrain: > The "stack" is not really a stack

I think that's what confused me here. I see now that with the offset-parameter in top/setTop, we really always need 8 valid elements (no more, no less), which means you can't just push elements without also dropping the last element.

Bummer :(

neobrain:
I don't really have the expertise to review this beyond what I already commented on, so Alyssa's +1 is good enough.

Good job on having continued pushing this forward despite how tricky it was to get working!

PR #3548, opened 2024-04-02
Eliminate xblock liveness for shifts

alyssarosenzweig:
The last one~


FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -863,6 +863,69 @@ DEF_OP(Ashr) {
...
+  ARMEmitter::SingleUseForwardLabel Done;
+  cbz(EmitSize, Src2, &Done);
+  {
+    // PF/SF/ZF/OF
+    if (OpSize >= 4) {
+      ands(EmitSize, StaticRegisters[16], Dst, Dst);
+    } else {
+      unsigned Shift = 32 - (OpSize * 8);
+      cmn(EmitSize, ARMEmitter::Reg::zr, Dst, ARMEmitter::ShiftType::LSL, Shift);
+      mov(EmitSize, StaticRegisters[16], Dst);
```
Sonicadvance1, line 884: Using hardcoded offset `16` is incorrect here for 32-bit. If anything it should be using the named register `REG_PF`

alyssarosenzweig: nice!

FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
```diff
@@ -863,6 +863,69 @@ DEF_OP(Ashr) {
...
+        neg(EmitSize, TMP1, Src2);
+      } else {
+        mov(EmitSize, TMP1, OpSize * 8);
+        sub(EmitSize, TMP1, TMP1, Src2);
+      }
+    } else {
+      sub(ARMEmitter::Size::i64Bit, TMP1, Src2, 1);
+    }
+
+    lsrv(EmitSize, TMP1, Src1, TMP1);
```
Sonicadvance1, line 899: Any reason why we stick with lsrv for all operations instead of using lsrv, lslv, and asrv for the chosen shift operation?

Sonicadvance1: Oh, this is entirely calculating flags for shifts, nevermind misread.

Sonicadvance1:
This PR causes Steam to crash early. Will need to bisect to find out why.

alyssarosenzweig:
> This PR causes Steam to crash early. Will need to bisect to find out why.

I've split the PR up into atomic commits that are as small as is reasonable. I don't think I fixed whatever the bug is (and I did audit instcountci without finding any fault), but this should make the bisect results much more useful.

Sonicadvance1:
Looks like the first commit that breaks Steam is `OpcodeDispatcher: use _ShiftFlags for SHL`

Sonicadvance1:
With latest changes this still crashes. First commit breaking steam is still `OpcodeDispatcher: use _ShiftFlags for SHL`

Sonicadvance1 (approved):
Congrats! it no longer explodes!

PR #3552, opened 2024-04-02
RFC: Negate 32-bit constants to inline

alyssarosenzweig:
RFC because this regresses instruction count in certain cases with complex addressing modes, I'm not sure where this is going wrong / whether it's an easy fix. See instcountci.


unittests/InstructionCountCI/FlagM/HotBlocks_32Bit.json
```diff
@@ -103,26 +103,27 @@
       "ExpectedArm64ASM": [
         "ldr w6, [x4, #12]",
         "ldr w4, [x4, #8]",
-        "mov w20, #0xffffffcc",
-        "str w10, [x9, w20, sxtw]",
+        "sub w20, w9, #0x34 (52)",
+        "str w10, [x20]",
```
Sonicadvance1, line 107: I think we need to teach a pass that stores with small offset in the range of `[-256, 256)` can inline the immediate. Which it looks like it doesn't understand the negative half of that equation yet.

alyssarosenzweig: Mm.. I'm going to close this for now since RA is more pressing, can maybe revisit later if @pmatos doesn't beat me to it

PR #3555, opened 2024-04-05
Add nix flake

billyb2:
This makes setting up the dependencies for building the project much easier. To use, you need to run:
`nix build '.?submodules=1'`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,61 @@
+{
+  "nodes": {
+    "flake-utils": {
+      "inputs": {
+        "systems": "systems"
+      },
+      "locked": {
+        "lastModified": 1710146030,
+        "narHash": "sha256-SZ5L6eA7HJ/nmkzGG7/ISclqe6oZdOZTNoesiInkXPQ=",
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "rev": "b1d9ab70662946ef0850d488da1c9019f3a9752a",
+        "type": "github"
+      },
+      "original": {
+        "owner": "numtide",
+        "repo": "flake-utils",
+        "type": "github"
+      }
+    },
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1711523803,
+        "narHash": "sha256-UKcYiHWHQynzj6CN/vTcix4yd1eCu1uFdsuarupdCQQ=",
+        "owner": "nixos",
+        "repo": "nixpkgs",
+        "rev": "2726f127c15a4cc9810843b96cad73c7eb39e443",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nixos",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "flake-utils": "flake-utils",
+        "nixpkgs": "nixpkgs"
+      }
+    },
+    "systems": {
+      "locked": {
+        "lastModified": 1681028828,
+        "narHash": "sha256-Vy1rq5AaRuLzOxct8nz4T6wlgyUR7zLU309k9mBC768=",
+        "owner": "nix-systems",
+        "repo": "default",
+        "rev": "da67096a3b9bf56a91d16901293e51ba5b49a27e",
+        "type": "github"
+      },
+      "original": {
+        "owner": "nix-systems",
+        "repo": "default",
+        "type": "github"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
+}
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,59 @@
+{
+  description = " A fast usermode x86 and x86-64 emulator for Arm64 Linux";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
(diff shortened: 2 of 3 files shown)


billyb2:
If this PR is accepted, is there a docs repo I can make a follow up to?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+  description = " A fast usermode x86 and x86-64 emulator for Arm64 Linux";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
+  };
+
+  outputs = { self, nixpkgs, flake-utils, ... }:
+    let
+      supportedSystems = [ "x86_64-linux" "aarch64-linux" ];
```
alyssarosenzweig, line 11: why x86?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+          fex = pkgs.stdenv.mkDerivation rec {
+            name = "FEXBash";
+            src = self;
+            nativeBuildInputs = deps ++ [ pkgs.stdenv.cc.cc.lib ];
+            dontUseCmakeConfigure = true;
+            dontUseCmakeInstall = true;
+            configurePhase = ''
+              mkdir Build
+              cd Build
+              CC=clang CXX=clang++ cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DUSE_LINKER=lld -DENABLE_LTO=True -DBUILD_TESTS=False -DENABLE_ASSERTIONS=False -G Ninja ..
```
alyssarosenzweig, line 45: Is the prefix right for nix...?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+            '';
+            buildPhase = ''
+              ninja -j $(nproc)
```
alyssarosenzweig, line 48: I think ninja should default to nproc so just `ninja` probably works

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+        let
+          pkgs = nixpkgsFor.${system};
+          deps = with pkgs; [
+            git
+            cmake
+            ninja
+            clang
+            binutils
+            lld
+            llvm
```
alyssarosenzweig, line 26: do we depend on llvm proper or just clang and lld?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+            gcc-unwrapped
+            python311Packages.setuptools
+          ];
+        in {
+          fex = pkgs.stdenv.mkDerivation rec {
+            name = "FEXBash";
+            src = self;
+            nativeBuildInputs = deps ++ [ pkgs.stdenv.cc.cc.lib ];
+            dontUseCmakeConfigure = true;
+            dontUseCmakeInstall = true;
```
Sonicadvance1, line 41: Why not use cmake's install step?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+              mkdir Build
+              cd Build
+              CC=clang CXX=clang++ cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DUSE_LINKER=lld -DENABLE_LTO=True -DBUILD_TESTS=False -DENABLE_ASSERTIONS=False -G Ninja ..
+            '';
+            buildPhase = ''
+              ninja -j $(nproc)
+            '';
+            installPhase = ''
+              mkdir -p $out/bin
+              mv ./Bin/* $out/bin/
```
Sonicadvance1, line 52: We have more install components than just what is in Bin/. Also some of the binaries in Bin shouldn't be installed as some are utility applications used during the build process.
Missing files would include config files, man pages, thunk libraries, and the like.
Would using the cmake install step resolve this?

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+            libepoxy
+            nasm
+            pkg-config
+            python3
+            gcc-unwrapped
+            python311Packages.setuptools
+          ];
+        in {
+          fex = pkgs.stdenv.mkDerivation rec {
+            name = "FEXBash";
```
Sonicadvance1, line 37: Is this the package name? FEXBash is just one utility application. If anything it should match our PPA package name of `fex-emu` https://github.com/FEX-Emu/FEX-ppa/blob/main/deb_base/control#L57

flake.lock
Sonicadvance1, line 1: As someone that has never used nix. What's a flake lock file for?

Sonicadvance1 (commented):
What an interesting package description format.

Sonicadvance1:
> If this PR is accepted, is there a docs repo I can make a follow up to?

We also have a `docs/` directory in this repo if a markdown file is enough there

flake.nix
```diff
@@ -0,0 +1,59 @@
...
+            SDL2
+            libepoxy
+            nasm
+            pkg-config
+            python3
+            gcc-unwrapped
+            python311Packages.setuptools
+          ];
+        in {
+          fex = pkgs.stdenv.mkDerivation rec {
```
bylaws, line 36: Can't you use clangStdenv here to avoid the need to set cc and manually install llvm?

neobrain:
Closing due to inactivity.

We can reopen it if interest picks up again, though personally I don't currently see a net benefit in adding poorly maintained build system integrations into `main`.

PR #3556, opened 2024-04-05
FEXCore: Fixes priority of FEX_APP_CONFIG

Sonicadvance1:
This environment variable had an incorrect priority on the configuration system. The expectation was higher priority than most other layers.

Now the only layer that has higher priority is the environment variables.


lioncash (approved):

Source/Common/Config.cpp
```diff
@@ -188,6 +190,12 @@ namespace JSON {
     , Config{std::move(ConfigFile)} {
   }
 
+
+  MainLoader::MainLoader(FEXCore::Config::LayerType Type, const char* ConfigFile)
+    : OptionMapper(Type)
+    , Config{std::move(ConfigFile)} {
```
lioncash, line 196: Noticed just after approving, but it doesn't seem like std::move would do anything here.

Sonicadvance1: Dropped the move

lioncash (approved):
but for real for real this time

PR #3559, opened 2024-04-06
LookupCache: Track ARM64EC page state in the code cache

bylaws:
Rather than checking the actual EC bitmap in the dispatcher (~6 instrs), this indirection through the code cache allows just 1 instr for the hot path of calling repeated EC code/x64 code.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -777,6 +777,14 @@ namespace FEXCore::Context {
     FEXCORE_PROFILE_SCOPED("CompileBlock");
     auto Thread = Frame->Thread;
 
+#ifdef _M_ARM_64EC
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -774,6 +774,14 @@ namespace FEXCore::Context {
   }
 
   uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame *Frame, uint64_t GuestRIP, uint64_t MaxInst) {
+#ifdef _M_ARM_64EC
```
Sonicadvance1, line 777: This block of code should probably be move at least after `FEXCORE_PROFILE_SCOPED` if not after pulling the thread object or after the CodeInvalidationMutex gets locked.

Sonicadvance1 (approved):

PR #3568, opened 2024-04-09
X87: Simplify constant loading for FLD family

lioncash:
We can just extract these into our named constant pool and load them out of there so that we don't need to rematerialize them every time one of these instructions are executed.

Was also able to apply this to FYL2XP1 and FPTAN as well


Sonicadvance1 (approved):
Oops, needs a rebase.

PR #3569, opened 2024-04-09
FHU: Switch over win32 file operations to std::filesystem

bylaws:
These were broken to varying degrees across the board on win32, just switch to their std::filesystem as windows doesn't have the same allocator issues that require their reimplementation.

NOTE: using std::filesystem within syscall callbacks (i.e. invalidation related code) would still cause issues, however I don't see that ever happening on the FEX side?

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
@@ -19,6 +19,18 @@
 #include <unistd.h>
 
 namespace FHU::Filesystem {
+  enum class CreateDirectoryResult {
+    CREATED,
+    EXISTS,
+    ERROR,
+  };
+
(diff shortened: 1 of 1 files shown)


FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -19,6 +19,12 @@
 #include <unistd.h>
 
 namespace FHU::Filesystem {
+#ifdef _WIN32
```
neobrain, line 22: These preprocessor branches add a lot of unnecessary noise that make the interfaces in this file difficult to read.

Instead of adding one such branch per function, could you group them together? (Doesn't need to be all in one block, but it might make sense to reorder some function definitions for readability as well)

Sonicadvance1 (approved):

PR #3572, opened 2024-04-12
Clang Format file and script

pmatos:
Adds paths to ignore to clang-format, .clang-format file. Also a wrapper script to clang-format to read .clang-format-ignore.

To format the whole tree at the root of the repository run: `find . -iname '*.h' -o -iname '*.cpp' -exec python3 Scripts/clang-format.py -i \{\} \;`

### DIFF
--- a/Scripts/clang-format.py
+++ b/Scripts/clang-format.py
@@ -1,0 +1,73 @@
+# Imported from LLVM
+# It's basically a clang-format wrapper with .clang-format-ignore support.
+# Can be removed once we adopt Clang19, which supports this out of the box.
+import subprocess
+import sys
+import os
+import re
+import fnmatch
+
+# Wrapper globals
+project_root = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
+ignore_file_path = os.path.join(project_root, ".clang-format-ignore")
+clang_format_command = "clang-format"
+
+
+def glob_to_regex(pattern):
+    # Normalize directory separators
+    pattern = pattern.replace("\\", "/")
+    return fnmatch.translate(pattern)
+
+
+def load_ignore_patterns(ignore_file_path):
+    with open(ignore_file_path, "r") as file:
+        lines = file.readlines()
+
+    patterns = []
+    for line in lines:
+        line = line.strip()
+        if line and not line.startswith("#"):  # Ignore empty lines and comments
+            pattern = glob_to_regex(line)
+            patterns.append(re.compile(pattern))
+    return patterns
+
+
+def normalize_path(file_path):
+    absolute_path = os.path.abspath(file_path)
+    normalized_path = absolute_path.replace("\\", "/")
+    return normalized_path
+
+
+def should_ignore(file_path, ignore_patterns):
+    normalized_path = normalize_path(file_path)
+    relative_path = os.path.relpath(normalized_path, start=project_root).replace(
+        "\\", "/"
+    )
+    for pattern in ignore_patterns:
+        if pattern.match(relative_path):
+            return True
+    return False
+
+
+def find_valid_file_paths(args):
+    return [arg for arg in args if os.path.isfile(arg)]
+
+
+def main():
+    ignore_patterns = load_ignore_patterns(ignore_file_path)
+    valid_paths = find_valid_file_paths(sys.argv[1:])
+
+    if len(valid_paths) != 1:
+        print("Error: Expected exactly one valid file path as argument.")
+        sys.exit(1)
+
+    file_path = valid_paths[0]
+    if should_ignore(file_path, ignore_patterns):
+        print(f"Ignoring {file_path} based on ignore patterns.")
+        return
+
+    subprocess.run([clang_format_command] + sys.argv[1:], check=True)
+
+
+if __name__ == "__main__":
+    main()
--- a/Scripts/reformat.sh
+++ b/Scripts/reformat.sh
@@ -1,0 +1,14 @@
+#! /bin/bash
+
+# Save current directory
+DIR=$(pwd)
+
+# Change to the directory passed as argument if any
+if [ $# -eq 1 ]; then
+    cd $1
+fi
+
+# Reformat whole tree.
+# This is run by the reformat target.
+git ls-files -z '*.cpp' '*.h' | xargs -0 -n 1 -P $(nproc) python3 Scripts/clang-format.py -i  
+cd $DIR
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,109 @@
+Language: Cpp
+BasedOnStyle: WebKit
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: Consecutive
+AlignConsecutiveDeclarations: None
+AlignConsecutiveMacros: None
+AlignEscapedNewlines: DontAlign
+AlignOperands: Align
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: false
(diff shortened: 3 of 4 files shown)


pmatos:
**why clang-format.py?**

for older versions of clang-format there was no way to ignore certain files in the tree. clang-format 19 adds this feature: https://clang.llvm.org/docs/ClangFormat.html#clang-format-ignore

Once we can all move to clang-format 19, we can drop the script which emulates this feature in python.

CMakeLists.txt
```diff
@@ -398,6 +398,10 @@ if (ENABLE_CLANG_FORMAT)
   set(CMAKE_CXX_CLANG_TIDY ${CLANG_TIDY_EXE} "${CLANG_TIDY_FLAGS}")
 endif()
 
+add_custom_target(reformat
+    COMMAND find ${CMAKE_SOURCE_DIR} -iname '*.h' -o -iname '*.cpp' -exec python3 ${CMAKE_SOURCE_DIR}/Scripts/clang-format.py -i \\{\\} \\\\;
```
neobrain, line 402: Could you limit this to our actual source directories?

I'd prefer not having this command make assumptions about what name the build folder has (I actually use multiple) or what other files are in the source folder.

Instead of `find`, maybe `git ls-tree` would be a better tool to use here?

neobrain: Ah, `git ls-files` is an even better match, since it has a `-c` toggle with which the index is included as well.

pmatos: yes - git ls-files is definitely better, but why is the difference so large between what you reported on discord and my reformatting? Let me look at it.

Scripts/clang-format.py
```diff
@@ -0,0 +1,70 @@
...
+    if len(valid_paths) != 1:
+        print("Error: Expected exactly one valid file path as argument.")
+        sys.exit(1)
+
+    file_path = valid_paths[0]
+    if should_ignore(file_path, ignore_patterns):
+        print(f"Ignoring {file_path} based on ignore patterns.")
+        return
+
+    subprocess.run([clang_format_command] + sys.argv[1:], check=True)
```
neobrain, line 66: Adding `--verbose` here might be useful to get some feedback (1 line per file). Maybe my VM is IO bottlenecked, but reformatting the entire tree takes ~15s without printing anything otherwise.

neobrain: Actually realized you can just pass this in manually since argv is passed along from the script to `clang-format`. Let's leave it as is :)

CMakeLists.txt
pmatos: I have updated this now. I was a bit worried about putting a bash script in the CMakeLists command so I moved that to a `reformat.sh` script. What do you think?

CMakeLists.txt
```diff
@@ -346,6 +346,10 @@ if (ENABLE_IWYU)
   endif()
 endif()
 
+add_custom_target(reformat
+    COMMAND Scripts/reformat.sh ${CMAKE_SOURCE_DIR}
```
Sonicadvance1, line 350: This needs to be changed to
```diff
diff --git a/CMakeLists.txt b/CMakeLists.txt
index 6f5e9081a..49c1a2972 100644
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -347,7 +347,7 @@ if (ENABLE_IWYU)
 endif()

 add_custom_target(reformat
-    COMMAND Scripts/reformat.sh ${CMAKE_SOURCE_DIR}
+  COMMAND ${CMAKE_CURRENT_SOURCE_DIR}/Scripts/reformat.sh ${CMAKE_SOURCE_DIR}
 )

 add_compile_options(-Wall)
```
Otherwise when executing from a build folder it doesn't find the script

Scripts/reformat.sh
```diff
@@ -0,0 +1,18 @@
...
+# Change to the directory passed as argument if any
+if [ $# -eq 1 ]; then
+    cd $1
+fi
+
+# Reformat whole tree.
+# This is run by the reformat target.
+for file in `git ls-files '*.cpp' '*.h'`
+do 
+    python3 Scripts/clang-format.py -i ${file}
```
Sonicadvance1, line 15: Changing this to
```diff
diff --git a/Scripts/reformat.sh b/Scripts/reformat.sh
index 957d7256c..7c37f3a6e 100755
--- a/Scripts/reformat.sh
+++ b/Scripts/reformat.sh
@@ -10,9 +10,6 @@ fi

 # Reformat whole tree.
 # This is run by the reformat target.
-for file in `git ls-files '*.cpp' '*.h'`
-do
-    python3 Scripts/clang-format.py -i ${file}
-done
+git ls-files -z '*.cpp' '*.h' | xargs -0 -n 1 -P $(nproc) python3 Scripts/clang-format.py -i

-cd $DIR
\ No newline at end of file
+cd $DIR
```
Allows the script to be run threaded which significantly improves execution time.

pmatos: You're right - done.

CMakeLists.txt
pmatos: Thanks.

Sonicadvance1 (approved):

pmatos:
@Sonicadvance1 landing the CI without this PR will format incorrectly because this PR is the one with the `.clang-format` file for the project. 

Sorry for not mentioning this PR dependency. :(

Sonicadvance1:
Nice, no worries. I don't see anything blocking this aside from final pending review.

neobrain (approved):
Woops, still had one pending comment left I thought I'd already submitted.

Does the CMake target serve much of a purpose now that there's a shell script? It's longer to type `ninja reformat` and you can't actually run it without a build folder, so it doesn't really seem to add any value.

Anyway, looks fine otherwise. Tested out the command again and the unexpected changes I reported on Discord are now fully gone since the fixup PR #3577.

neobrain (changes requested):
(changing review status to avoid confusion)

pmatos:
> Woops, still had one pending comment left I thought I'd already submitted.
> 
> Does the CMake target serve much of a purpose now that there's a shell script? It's longer to type `cmake reformat` and you can't actually run it without a build folder, so it doesn't really seem to add any value.
> 
> Anyway, looks fine otherwise. Tested out the command again and the unexpected changes I reported on Discord are now fully gone since the fixup PR #3577. 
> 

It was @Sonicadvance1 who wanted a ninja target for reformatting. Will leave it up to him.

Sonicadvance1:
I would still prefer the cmake target but other people have stronger opinions than me so if they are adamant about it then just delete it.

pmatos:
> I would still prefer the cmake target but other people have stronger opinions than me so if they are adamant about it then just delete it.

Personally, I wouldn't use either so I don't care. @neobrain ?

Scripts/clang-format.py
```diff
@@ -0,0 +1,70 @@
+import subprocess
```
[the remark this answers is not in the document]
pmatos: Done.

pmatos:
Ignore the check code formatting fail until this lands. :) It's the one that contains the one true .clang-format file.

neobrain:
> I would still prefer the cmake target but other people have stronger opinions than me so if they are adamant about it then just delete it.
> 
> Personally, I wouldn't use either so I don't care. @neobrain ?

Great, let's proceed with deleting it then. The CMakeLists.txt is already bloated enough as is :)

pmatos:
> > I would still prefer the cmake target but other people have stronger opinions than me so if they are adamant about it then just delete it.
> > Personally, I wouldn't use either so I don't care. @neobrain ?
> 
> Great, let's proceed with deleting it then. The CMakeLists.txt is already bloated enough as is :)

Done. Happy to see this landing soon.

neobrain (approved):

PR #3579, opened 2024-04-15
Replace deprecated  parse_version from pkg_resources

pmatos:
Uses recommended version usage from packaging.
Silences cmake warning:
```
/home/pmatos/dev/FEX/Scripts/aarch64_fit_native.py:5: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
```

Requires lib packaging to be installed.


pmatos:
CI failure because `packaging` library is not builtin so we'll need to install it. However, it's still the recommended replacement for `parse_version` which is deprecated.

PR #3580, opened 2024-04-15
AppConfig: Disable libGL forwarding for steamwebhelper

neobrain:
This app bypasses the glX functions exported by libGL and instead sends GLX requests directly via xcb. FEX won't support forwarding this usage pattern in the foreseeable future.

### DIFF
--- a/Data/AppConfig/steamwebhelper.json
+++ b/Data/AppConfig/steamwebhelper.json
@@ -1,0 +1,6 @@
+{
+  "Comment": "Bypasses libGL's glX and instead sends GLX requests directly via xcb",
+  "ThunksDB": {
+    "GL": 0
+  }
+}


Sonicadvance1:
Does this still work when the global config has GL and Vulkan thunking enabled? Not sure if Vulkan will play a role but it may still redirect xcb things.

neobrain:
> Does this still work when the global config has GL and Vulkan thunking enabled? Not sure if Vulkan will play a role but it may still redirect xcb things.

What problem are you anticipating?

Sonicadvance1:
> > Does this still work when the global config has GL and Vulkan thunking enabled? Not sure if Vulkan will play a role but it may still redirect xcb things.
> 
> What problem are you anticipating?

I'm anticipating global Vulkan thunking still being enabled for steamwebhelper to pull in its xcb dependency, then steamwebhelper using xcb in some fashion to break.

neobrain:
> I'm anticipating global Vulkan thunking still being enabled for steamwebhelper to pull in its xcb dependency, then steamwebhelper using xcb in some fashion to break.

Maybe, I haven't checked. The intent of this PR is to set the stage for the next PR (which will transition to libGL/libvulkan forwarding based on guest-libX11) rather than making it safer to enable library forwarding as-is globally. So long-term, GL is the library to list in this AppConfig.

Sonicadvance1 (dismissed):

Data/AppConfig/steamwebhelper.json
```diff
@@ -0,0 +1,6 @@
+{
+  "Comment": "Bypasses libGL's glX and instead sends GLX requests directly via xcb"
```
Sonicadvance1, line 2: This line needs a comma at the end, otherwise the json parser will explode.

Sonicadvance1 (approved):

PR #3581, opened 2024-04-15
Library Forwarding: Fix issues with libGL's fake X11 dependency

neobrain:
The library's soname is changed to libX11.so.6 and the CMake target is
renamed to libPlaceholderX11. This fixes two issues:
* Steam and mangohud can't find libX11 during startup if the library doesn't include a version suffix.
* Calling the CMake target libX11 overrode the true host X11 library used by unrelated targets (such as FEXConfig), which could cause link errors


neobrain:
Looks like non-LLD builds are failing currently, will investigate why.

neobrain:
The fake libX11 dependency broke the build since FEXConfig linked against it instead of using the true host libX11, which is why it couldn't find any of the actual X11 symbols. Not sure why this didn't already show up before, but it's fixed now either way.

Sonicadvance1 (approved):

PR #3583, opened 2024-04-15
Remove some x86 leftovers

awulicja:
As far as I know the x86 hosts are no longer supported, there is an x86 specific option (that crashes FEX on x86 with `terminate called after throwing an instance of 'std::bad_function_call'`) and an old description in readme. This PR cleans it up a bit.

The CONFIG_IRJIT isn't completely removed, but I actually think that it could be removed. That will mean removing the CustomCPUFactory concept (I guess) which might not be ideal. For now removing an option to use the unused choice might be okay.

I also noticed that there are some _M_X86_64 defines that hide x86 specific code, should they go as well? Or maybe everything that isn't in FEXCore should keep the #ifdef _M_X86_64?

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -6,12 +6,11 @@
         "Default": "FEXCore::Config::ConfigCore::CONFIG_IRJIT",
         "TextDefault": "irjit",
         "ShortArg": "c",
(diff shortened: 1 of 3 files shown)


alyssarosenzweig (approved):

Sonicadvance1:
This breaks the hostrunner which our CI does still use.

Sonicadvance1:
Closing since this fundamentally breaks some of our testing

PR #3535, opened 2024-03-30
WIP: Moves Thread state tracking to frontend

Sonicadvance1:
The FEXCore ThreadManagement object has been a byproduct of leaking Linux details in to FEXCore for a long time. Since we have moved most of the thread management to the frontend, we can finally move this out of FEXCore and start taking advantage of the `FrontendPtr`.

PR #3592, opened 2024-04-19
GDBServer fixes

Sonicadvance1:
This gets gdbserver working for simpler cases again. Some of the work is working around #3535 not being complete since some of the thread state management is still split between FEXCore and the frontend.

I'm still working on fixing that code, but until then lets at least get gdbserver working again so we can get x86 debugging.


neobrain (commented):
Thanks for looking into the problem! It seems that I still can't do anything useful with `gdb-multiarch` connecting to `FEXLoader -G ...`. Is there anything specifically that should be working now that I can test?

Otherwise, looks like this PR adds a few similarly named interfaces to already sparely documented APIs. That should be sorted out to keep the interfaces somewhat accessible.

Sonicadvance1:
Pulled a couple minor patches from this in to #4170, but none of the bigger logic changes.
Will be changing enough code that this won't matter much now.

PR #3599, opened 2024-04-25
JIT: fix neon vec4 faddv

alyssarosenzweig:
We were previously genrating nonsense code if the destination != source:

         faddp v2.4s, v4.4s, v4.4s
         faddp s2, v4.2s

The result of the first faddp is ignored, so the second merely calculates the sum of the first 2 sources (not all 4 as needed).

The correct fix is to feed the first add into the second, regardless of the final destination:

         faddp v2.4s, v4.4s, v4.4s
         faddp s2, v2.2s

Hit in an ASM test with new RA.


Sonicadvance1 (approved):
And I oop.

PR #3601, opened 2024-04-29
Fix 8/16-bit ADC/SBC

alyssarosenzweig:
alternative to #3600


Sonicadvance1:
Interestingly, while this solves the unittest, it doesn't solve the rendering issue that FF7 is encountering.

alyssarosenzweig:
> Interestingly, while this solves the unittest, it doesn't solve the rendering issue that FF7 is encountering.

Curious. Guess we're going to want a second unit test too, lest we optimize this later 😅 

Racking my brain for what behaviour change could possibly happen here.

alyssarosenzweig:
Hey, if it's good enough for me to repro, that'll do! thanks 😸

alyssarosenzweig:
Oh, derp, I know what this is. Fix coming. thanks!

Sonicadvance1:
With the latest changes FF7 now renders correctly again!

Sonicadvance1 (approved):
Nice little optimization and nice bug fix!

PR #3606, opened 2024-05-02
CPUID: Fix inverted RDTSCP check

Sonicadvance1:
This was inverted and always enabling the RDTSCP cpuid bit for wine. Thus always disabling it elsewhere.


lioncash (approved):
oops

PR #3607, opened 2024-05-02
Pass compulsory `mode` argument to `open` when `O_CREAT` is used

teohhanhui:
From `man 2 open`:

> The mode argument must be supplied if O_CREAT or O_TMPFILE is
> specified in flags; if it is not supplied, some arbitrary bytes
> from the stack will be applied as the file mode.

(Fix broken build on Fedora Rawhide: https://download.copr.fedorainfracloud.org/results/teohhanhui/fex-emu/fedora-rawhide-aarch64/07399392-fex-emu/builder-live.log.gz)


Sonicadvance1 (approved):
Nice catch! Thanks!

PR #3612, opened 2024-05-05
FEXLoader: Changes frontend thread management to wrap FEXCore thread objects

Sonicadvance1:
A bit of refactoring necessary before we can move the remaining Linux specific code to the frontend.

Most of this taken from #3535 but attempting to be NFC as much as possible.


lioncash (approved):
Looks good as far as I can tell

PR #3613, opened 2024-05-05
CPUInfo: Optimize CalculateNumberOfCPUs

Sonicadvance1:
Caches the result for future requests since we call this multiple times.

Additionally instead of using fmt::format_to_n, it instead uses std::to_chars for just the index at the end of the filename.

This causes the function to go from around 146k ns to 55k ns, nearly 1/3rd the amount of time to execute.

### DIFF
--- a/FEXCore/Source/Utils/CPUInfo.cpp
+++ b/FEXCore/Source/Utils/CPUInfo.cpp
@@ -3,6 +3,8 @@
 
 #include <fmt/format.h>
 
+#include <atomic>
+#include <charconv>
 #include <cstddef>
 #include <cstdint>
 #ifdef _WIN32
@@ -14,13 +16,24 @@
 namespace FEXCore::CPUInfo {
 #ifndef _WIN32
 uint32_t CalculateNumberOfCPUs() {
-  char Tmp[PATH_MAX];
+  static std::atomic<uint32_t> NumCPUs {};
+  if (NumCPUs.load(std::memory_order_relaxed)) {
+    return NumCPUs.load(std::memory_order_relaxed);
+  }
+
+  // Support up to 8 digits of CPU cores plus the null terminator.
+  constexpr std::string_view CPUString = "/sys/devices/system/cpu/cpu";
+  char Tmp[CPUString.size() + 1 + 8];
+  std::ranges::copy(CPUString, Tmp);
   size_t CPUs = 1;
 
   for (;; ++CPUs) {
-    auto Size = fmt::format_to_n(Tmp, sizeof(Tmp), "/sys/devices/system/cpu/cpu{}", CPUs);
-    Tmp[Size.size] = 0;
+    auto Res = std::to_chars(&Tmp[CPUString.size()], &Tmp[sizeof(Tmp)], CPUs);
+    // null terminate the string.
+    Res.ptr[0] = '\0';
+
     if (!FHU::Filesystem::Exists(Tmp)) {
+      NumCPUs.store(CPUs, std::memory_order_relaxed);
       break;
     }
   }


FEXCore/Source/Utils/CPUInfo.cpp
```diff
@@ -14,17 +15,30 @@
...
+  if (NumCPUs) {
+    return NumCPUs;
+  }
+
+#define CPUSTRING "/sys/devices/system/cpu/cpu"
+  constexpr size_t CPUStringLength = std::char_traits<char>::length(CPUSTRING);
+
+  // Support up to 8 digits of CPU cores plus the null terminator.
+  char Tmp[CPUStringLength + 1 + 8] = CPUSTRING;
+  static_assert(std::char_traits<char>::length(CPUSTRING) == 27, "What");
```
lioncash, line 28:
```suggestion
  static_assert(CPUStringLength == 27, "What");
```

FEXCore/Source/Utils/CPUInfo.cpp
```diff
@@ -14,17 +15,30 @@
 namespace FEXCore::CPUInfo {
 #ifndef _WIN32
 uint32_t CalculateNumberOfCPUs() {
-  char Tmp[PATH_MAX];
+  static uint32_t NumCPUs {};
+  if (NumCPUs) {
+    return NumCPUs;
+  }
+
+#define CPUSTRING "/sys/devices/system/cpu/cpu"
```
lioncash, line 23: Should this be undef'd after the assignment into the char buffer?

lioncash (commented):
Two minor nits but seems reasonable overall

FEXCore/Source/Utils/CPUInfo.cpp
Sonicadvance1: undef'd

FEXCore/Source/Utils/CPUInfo.cpp
```diff
@@ -14,17 +15,31 @@
...
+  constexpr size_t CPUStringLength = std::char_traits<char>::length(CPUSTRING);
+
+  // Support up to 8 digits of CPU cores plus the null terminator.
+  char Tmp[CPUStringLength + 1 + 8] = CPUSTRING;
+#undef CPUSTRING
```
neobrain, line 28: This got overly complex. `std::string_view` makes it more straightforward:

```suggestion
  // Support up to 8 digits of CPU cores plus the null terminator.
  char Tmp[CPUString.size() + 1 + 8];
  constexpr std::string_view CPUString = "/sys/devices/system/cpu/cpu";
  std::ranges::copy(CPUString, Tmp);
```

FEXCore/Source/Utils/CPUInfo.cpp
```diff
@@ -14,17 +15,31 @@
 namespace FEXCore::CPUInfo {
 #ifndef _WIN32
 uint32_t CalculateNumberOfCPUs() {
-  char Tmp[PATH_MAX];
+  static uint32_t NumCPUs {};
```
neobrain, line 18: IIRC `uint32_t` writes are atomic on arm64 anyway, but might as well be explicit that it must be so for thread-safety:
```suggestion
  static std::atomic<uint32_t> NumCPUs {};
```

FEXCore/Source/Utils/CPUInfo.cpp
```diff
@@ -14,17 +15,31 @@
...
+    // null terminate the string.
+    Res.ptr[0] = '\0';
+
     if (!FHU::Filesystem::Exists(Tmp)) {
       break;
     }
   }
 
+  NumCPUs = CPUs;
   return CPUs;
```
neobrain, line 43: To keep control flow straightforward, this can just be moved inside of the failed `if (Exists)` check above.

neobrain:
> This causes the function to go from around 146k ns to 55k ns, nearly 1/3rd the amount of time to execute.

Does this have tangible impact in practice on hardware FEX is commonly used on? How often are we calling this function?

FEXCore/Source/Utils/CPUInfo.cpp
Sonicadvance1: constexpr also has the academic benefit of knowing the string size at compile time instead of strlen getting called at runtime with this suggested implementation.

```
        mov     x0, sp
        ldp     q0, q1, [x8]
        mov     x20, sp
        stp     q0, q1, [sp]
        ldp     q0, q2, [x8, #32]
        stp     q0, q2, [sp, #32]
        bl      strlen
 ```

neobrain: Did you enable compiler optimizations? That would be the relevant scenario to reference when optimizing in the nano-second range.

Sonicadvance1: This is with -O3, so yes.

neobrain: Interesting, I tested against gcc and it optimizes away strlen there while clang doesn't. Since it's just one-time setup, this probably has no significant overhead compared to the filesystem accesses below, but happy to just go with the former proposal if you prefer.

FEXCore/Source/Utils/CPUInfo.cpp
Sonicadvance1: Switched to atomic

FEXCore/Source/Utils/CPUInfo.cpp
Sonicadvance1: Moved

FEXCore/Source/Utils/CPUInfo.cpp
Sonicadvance1: Not quite as optimal since clang zero pads the ending of the string when it is a char array which lets it generate the whole string in a single pair of 256-bit ldp and stp instructions but converted over to the string_view anyway.

neobrain:
> We call this function twice in our code currently. So 146k ns times two, down to roughly 55k ns with the second access getting the cached result. Minor startup time uplift.

I assumed it would be orders of magnitude higher. This seems like a really weak case for the complexity introduced here, especially considering the majority of those gains (146->73) would come simply from the caching. Am I missing something that makes the extra 18 µs more critical than they appear to be? If not, I think we're better off dropping the non-caching changes here.

PR #3616, opened 2024-05-06
SRA controlled burn

alyssarosenzweig:
Our current SRA intrinsics are way too expressive, with very complicated semantics (especially with sve256/avx). This makes a mess out of the whole codebase:

* too many random arguments in the dispatcher
* too many arguments to validate in passes
* too much complicated handling in RA
* too much complicated handling in the JIT

Fortunately, we don't actually need any of the complexity. This PR aims to simplify the intrinsics to just the subset of functionality we actually use in the dispatch, allowing us to drop a lot of code from the JIT now and avoid writing a pile of new dead code in the new RA.


FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
```diff
@@ -105,42 +48,14 @@ struct GPRInfo {
...
-}
-
-uint32_t GPRBit(uint32_t Offset) {
-  if (!IsGPR(Offset)) {
-    return 0;
-  }
-
-  return 1 << ((Offset - 8) / 8);
+uint32_t GPRBit(RegisterClassType Class, uint32_t Reg) {
+  return (Class == GPRClass) ? (1 << Reg) : 0;
```
lioncash, line 52: Perhaps:

```suggestion
  return (Class == GPRClass) ? (1U << Reg) : 0;
```

Just so it's always well defined, even if bit 31 isn't pragmatically hit in practice

lioncash (approved):
One very small nit. Feel free to punt on it, since it was in the original code as well

Overall though, this looks nice

FEXCore/Source/Interface/IR/Passes/DeadStoreElimination.cpp
alyssarosenzweig: Sure

PR #3617, opened 2024-05-06
FEXCore: ARM64EC x64 entry/exit support 

bylaws:
This drops any need for frontend assembly code to be aware of FEX's SRA layout at the cost of the dispatcher needing to access the TEB. Also includes a patch to disable frontend callbacks during code compilation or block linking


alyssarosenzweig:
This looks plausible... wondering if there's any expected perf change (positive or negative)?

alyssarosenzweig:
> This drops any need for frontend assembly code to be aware of FEX's SRA layout

I suppose you have follow on patches that simplify the frontend? in isolation I'm not sure what the benefit of this PR is.

bylaws:
The perf change here should be pretty much negligible, ec exit/entry isn't hot enough of a path for an extra instruction or two to realistically matter. This is a prerequisite for the frontend to make it upstream, rather than submitting it as a mix of fexcore changes and new code

alyssarosenzweig:
> The perf change here should be pretty much negligible, ec exit/entry isn't hot enough of a path for an extra instruction or two to realistically matter. This is a prerequisite for the frontend to make it upstream, rather than submitting it as a mix of fexcore changes and new code

Sounds good. Ack from me on principle, will defer review to someone more knowledgable in this area.

Sonicadvance1 (approved):

PR #3621, opened 2024-05-08
RegisterAllocationPass: drop AVX flag

alyssarosenzweig:
RA should not depend on whether we support AVX, that's a huge layering
    violation! and fortunately, it does not.

Follow on from SRA work


Sonicadvance1 (approved):
I guess this no longer mattered because we ensure that Load/StoreRegister IR ops now do full accesses instead of anything partial now. So it was sort of working away from needing to know anyway.

alyssarosenzweig:
> I guess this no longer mattered because we ensure that Load/StoreRegister IR ops now do full accesses instead of anything partial now. So it was sort of working away from needing to know anyway.

Yep, exactly. And now since #3616, that requirement is baked into the IR, which makes the deadcodeness much more obvious.

PR #3622, opened 2024-05-08
FEXCore: Moves CodeEmitter to own module 

Sonicadvance1:
The frontend needs to have a microJIT for BPF/seccomp. It can't use the CodeEmitter from FEXCore in its current state.

- Removes the vixl dependency from the CodeEmitter
   - Only vixl uses were its Float16 class. Replaced with a custom implementation
   -  And IsImmLogical, which was added to its own header with vixl's license maintained. Reimplementing the logic is hard.
- Moves it to its own module
- Renames the namespace

No functional change


FEXHeaderUtils/FEXHeaderUtils/CodeEmitter/VixlUtils.inl
```diff
@@ -0,0 +1,311 @@
...
+
+static inline uint64_t LowestSetBit(uint64_t value) {
+  return value & UnsignedNegate(value);
+}
+
+template <typename V>
+static inline int CountLeadingZeros(V value, int width = (sizeof(V) * 8)) {
+#if COMPILER_HAS_BUILTIN_CLZ
+  if (width == 32) {
+    return (value == 0) ? 32 : __builtin_clz(static_cast<unsigned>(value));
```
lioncash, line 269: Seems like this ifdef and fallback can be replaced with `std::countl_zero` from the `<bit>` header

FEXHeaderUtils/FEXHeaderUtils/CodeEmitter/VixlUtils.inl
```diff
@@ -0,0 +1,311 @@
...
+  }
+
+  return true;
+}
+
+private:
+
+template <typename V>
+static inline bool IsPowerOf2(V value) {
+  return (value != 0) && ((value & (value - 1)) == 0);
```
lioncash, line 250: `std::has_single_bit`

FEXHeaderUtils/FEXHeaderUtils/CodeEmitter/VixlUtils.inl
Sonicadvance1: These two helpers are directly pulled from vixl. Didn't want to mess with them especially since they have some prickly behaviours like passing in a uint8_t but casting to different widths.

neobrain:
Why doesn't this live in its own module? FHU makes sense for things that heavily inter-depend on each other and that are used virtually everywhere, but this is a fairly isolated piece of code.

CodeEmitter/CodeEmitter/Emitter.h
```diff
@@ -759,15 +756,91 @@ class Emitter : public FEXCore::ARMEmitter::Buffer {
...
+        RawBits = (Sign << 15) | (0b1'1111 << 10);
+        break;
+      case FP_SUBNORMAL:
+      case FP_NAN:
+      // TODO: Can be encoded if necessary, but only handle when necessary.
+      default: LOGMAN_MSG_A_FMT("Invalid FP type");
+      }
+    }
+
+    uint8_t ToImm8() {
```
neobrain, line 806:
```suggestion
    uint8_t ToImm8() const {
```

neobrain:
Looks fine to me in the grand scheme now that there's a dedicated CMake target.

> Only vixl uses were its Float16 class. Replaced with a custom implementation

Is there an upside to this other than making the dependency graph simpler (without removing the use of vixl elsewhere)? The new class has no unit tests, so its upsides must be weighed against this added risk and the additional time required for the more careful review.

Sonicadvance1:
> Looks fine to me in the grand scheme now that there's a dedicated CMake target.
> 
> > Only vixl uses were its Float16 class. Replaced with a custom implementation
> 
> Is there an upside to this other than making the dependency graph simpler (without removing the use of vixl elsewhere)? The new class has no unit tests, so its upsides must be weighed against this added risk and the additional time required for the more careful review.

An upside is that the vixl dependency isn't leaked to the frontend.

Sonicadvance1:
To what end? This is the easiest route.

neobrain:
> To what end? This is the easiest route.

The PR replaces battle-proven code with a custom reimplementation of the same interface. Continuing to use vixl avoids risks associated with that, though I'd be happy to instead see tests for the new Float16 code as well.

Sonicadvance1:
At this point I might as well as just delete all Float16 references since we don't even use it.

neobrain:
> At this point I might as well as just delete all Float16 references since we don't even use it.

I didn't know the code using Float16 wasn't hit. If it can just be dropped, that's even better.

Sonicadvance1:
> > At this point I might as well as just delete all Float16 references since we don't even use it.
> 
> I didn't know the code using Float16 wasn't hit. If it can just be dropped, that's even better.

Deleted.

FEXCore/unittests/Emitter/SVE_Tests.cpp
```diff
@@ -2336,69 +2336,69 @@ TEST_CASE_METHOD(TestDisassembler, "Emitter: SVE: SVE broadcast integer immediat
 }
 
 TEST_CASE_METHOD(TestDisassembler, "Emitter: SVE: SVE broadcast floating-point immediate (predicated)") {
-  TEST_SINGLE(fcpy(SubRegSize::i16Bit, ZReg::z30, PReg::p6.Merging(), -0.125), "fmov z30.h, p6/m, #0xc0 (-0.1250)");
+  // TEST_SINGLE(fcpy(SubRegSize::i16Bit, ZReg::z30, PReg::p6.Merging(), -0.125), "fmov z30.h, p6/m, #0xc0 (-0.1250)");
```
neobrain, line 2339: Probably good to keep these around in case 16-bit support is enabled again later, but code shouldn't be commented out without indication why.

Might make most sense to add a preprocessor define at the top of the file to be more explicit about it, or to move together the various i16Bit tests so we don't need to repeat the same docstring everywhere.

FEXCore/unittests/Emitter/Scalar_Tests.cpp
```diff
@@ -757,9 +757,64 @@ TEST_CASE_METHOD(TestDisassembler, "Emitter: Scalar: Floating-point compare") {
   TEST_SINGLE(fcmpe(HReg::h30), "fcmpe h30, #0.0");
 }
 TEST_CASE_METHOD(TestDisassembler, "Emitter: Scalar: Floating-point immediate") {
-  TEST_SINGLE(fmov(ScalarRegSize::i16Bit, VReg::v30, 1.0), "fmov h30, #0x70 (1.0000)");
+  // TEST_SINGLE(fmov(ScalarRegSize::i16Bit, VReg::v30, 1.0), "fmov h30, #0x70 (1.0000)");
   TEST_SINGLE(fmov(ScalarRegSize::i32Bit, VReg::v30, 1.0), "fmov s30, #0x70 (1.0000)");
   TEST_SINGLE(fmov(ScalarRegSize::i64Bit, VReg::v30, 1.0), "fmov d30, #0x70 (1.0000)");
+
+#if 0
+  float Decoding[] = {
```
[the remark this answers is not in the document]
Sonicadvance1: This is testing the float->float16->imm8 encoding correctness

FEXCore/unittests/Emitter/SVE_Tests.cpp
Sonicadvance1: Wrapped in a preprocessor define.

neobrain: There's now a single point of reference for the disabled code (good!), but it's still not indicated why the code is disabled. Could you add a comment to the preprocessor define?

Sonicadvance1: Added a comment

neobrain (approved):

FEXCore/unittests/Emitter/SVE_Tests.cpp
neobrain: Great, thanks!

PR #3624, opened 2024-05-10
FEXCore: Get rid of DeferredSignalFaultAddress and use the InterruptFaultPage

Sonicadvance1:
Arm64ec introduced the InterruptFaultPage which is lower overhead since instead of ldr+str it just turns in to a single str. We were already allocating the space, FEXCore and the frontend signal delegator just needed to be updated to understand the new location.

We can additionally use this in the future if we want to make deferred async signals INSIDE the JIT only cost a single str as well.

Side-benefit: Shaving 1 page off each thread's state object allocation size. From 3 pages down to 2. Not much but eh, isn't the point of this PR.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
@@ -1233,7 +1233,7 @@ bool SignalDelegator::HandleSIGILL(FEXCore::Core::InternalThreadState* Thread, i
       // If we have more deferred frames to process then mprotect back to PROT_NONE.
       // It will have been RW coming in to this sigreturn and now we need to remove permissions
       // to ensure FEX trampolines back to the SIGSEGV deferred handler.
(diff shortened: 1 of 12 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
```diff
@@ -1405,7 +1404,7 @@ void SignalDelegator::HandleGuestSignal(FEXCore::Core::InternalThreadState* Thre
 #else
         // X86 should always be doing a refcount compare and branch since we can't guarantee instruction size.
```
alyssarosenzweig, line 1405: Guessing this comment needs an update ?

alyssarosenzweig (approved):
Very clever. Nice!

Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
Sonicadvance1: Nah, x86 implementation still does a refcount compare and branch. That's over in SignalScopeGuards.h

PR #3628, opened 2024-05-13
WIP: Seccomp

Sonicadvance1:
Let's see what CI says.


Source/Tools/LinuxEmulation/LinuxSyscalls/SeccompEmulator.cpp
```diff
@@ -0,0 +1,890 @@
...
+
+    break;
+  }
+  case BPF_NEG:
+    if (SrcType == BPF_X) {
+      ///< Only BPF_K supported on NEG.
+      return -EINVAL;
+    }
+
+    mvn(ARMEmitter::Size::i32Bit, REG_A, REG_A);
```
alyssarosenzweig, line 570: I believe this should be `neg`, not `mvn`. (Arithmetic inverse instead of bitwise inverse)

alyssarosenzweig: Per the reference implementation - https://github.com/torvalds/linux/blob/master/arch/arm64/net/bpf_jit_comp.c#L970-L974

Sonicadvance1: You're right. I'll need to fix that.

alyssarosenzweig:
Is 64-bit bpf interesting for FEX? Seems potentially better to do now than later given how much churn it would be.

Sonicadvance1:
> Is 64-bit bpf interesting for FEX? Seems potentially better to do now than later given how much churn it would be. (Unless it's never going to be supported for some good reason)

I haven't seen any usage of BPF outside of seccomp currently. Not even sure if the bpf syscall is even typically accessible outside of root. Also haven't checked if we can just forward the syscall, I don't know if they expose an architecture or anything inside the VM. So it's just kind of eeh.

alyssarosenzweig:
> Also haven't checked if we can just forward the syscall

Didn't we talk about this in the meeting last week? I thought there was some reason FEX had to do these shenanigans instead of simply forwarding like @neobrain hoped...? I'd rather not merge 1kloc of new JIT code if we could just... not 😅

Sonicadvance1:
> > Also haven't checked if we can just forward the syscall
> 
> Didn't we talk about this in the meeting last week? I thought there was some reason FEX had to do these shenanigans instead of simply forwarding like @neobrain hoped...? I'd rather not merge 1kloc of new JIT code if we could just... not 😅

Sorry, "bpf" the syscall is different from seccomp. Seccomp uses a highly restricted form of BPF which exposes an architecture constant which means we can't forward it.
"bpf" might not have that same architecture specific data and /might/ be able to get forwarded, unsure and needs investigation. Ideally that syscall is still restricted in userspace and can continue to get ignored.

Sonicadvance1:
jetson-orin-1 is still on Clang 12.x which doesn't support the C++20 feature of lambda expressions in unevaluated operands. Clang 13.x added support for this. Time to raise the minspec to Clang-13.

neobrain:
> Sorry, "bpf" the syscall is different from seccomp. Seccomp uses a highly restricted form of BPF which exposes an architecture constant which means we can't forward it.
"bpf" might not have that same architecture specific data and /might/ be able to get forwarded, unsure and needs investigation. Ideally that syscall is still restricted in userspace and can continue to get ignored.

I'm also interested in picking up this conversation again. Currently the PR description is very sparse on details, but I hope some background on this change can be provided once the WIP label is dropped to enable a sensible review.

Sonicadvance1:
This has now been cleaned up and rebased. It's now ready for review.

It requires a config option for the feature to be enabled because there are a few remaining issues around seccomp that needs to be resolved first. This will allow it to get merged while I work through the remaining issues that aren't directly related to the seccomp feature itself.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
```diff
@@ -759,6 +787,91 @@ uint32_t SyscallHandler::CalculateGuestKernelVersion() {
 }
 
 uint64_t SyscallHandler::HandleSyscall(FEXCore::Core::CpuStateFrame* Frame, FEXCore::HLE::SyscallArguments* Args) {
+  // Need to reconstruct RIP from the return address. Which will be inside the FEX JIT.
+  const uint64_t RIP =
```
Sonicadvance1, line 791: This handling that was added to this function should be moved in to the `ExecuteFilter` and `RestoreRIPFromHostPC` only called if RIP is actually needed. To keep seccomp handling consolidated in the seccomp handler and reduce overhead for when seccomp isn't enabled.

neobrain:
Thanks for the detailed writeup in the PR description!

> FEX needs to implement seccomp because there are multiple applications using the feature, the primary one being Chromium which some games embed without disabling the sandbox. WINE also uses seccomp for capturing games that do raw Windows system calls.

How can I test seccomp support in Chromium and Wine? Are there any other suitable tests?

> Seccomp uses a highly restricted form of BPF which exposes an architecture constant which means we can't forward it.

Are there any other important features that could hinder forwarding to the host kernel?

I briefly looked into the arch constant thing and it's not clear to me it's truly a blocker. Have you considered patching seccomp data loads (redirecting them to a data buffer set up in a FEX-internal preamble) so that we can leverage the host kernel's BPF runtime instead of reimplementing *all of it*? (Also, is the arch constant used in practice for workloads important to FEX?)

neobrain:
> For testing Wine, enable the config option and run Red Dead Redemption 2. Proton will stop complaining about seccomp in the logs

This isn't possible without a separate Rockstar account, right?

> For testing Chromium, build it from source, run it without --no-sandbox, watch it explode from things unrelated to seccomp for now. I had some hacks to work around the remaining sandbox problems to test it.

So practically speaking it's not possible to verify for the time being?

> For another tool, use the Linux unittest linux/tools/testing/selftests/seccomp/seccomp_bpf which runs through a test suite.

Fair. Is it expected to pass with the current implementation?

> So....No. It won't work.

It's already evident this is your preferred implementation strategy since you opened the PR as-is. Does it hurt to explore the design space? I would feel more comfortable with dropping almost 1600 new lines of code with a huge gap in test coverage if I understood that other options have been thoroughly explored.

> The language (BPF) is effectively turing complete with access to temporary memory.

I thought the architecture constant was part of a dedicated, seccomp-specific storage area that is not accessible through normal load/store instructions. Is this incorrect? (If it is, what prevents us from adding runtime-checks to catch it?)

> syscall numbers don't match between x86, x86-64, and aarch64. So....No. It won't work.

How are system call numbers used in BPF? A quick look through the documentation didn't bring up anything obvious.

>  Additionally, true seccomp on top of FEX means it can break syscalls that FEX itself needs and then there is literally no way to work around that.

Do we *need* to install guest seccomp filters as host seccomp filter? Can't we run the recompiled seccomp as a normal BPF program that we call as part of each system call?

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
```diff
@@ -0,0 +1,397 @@
...
+    EMIT_INST(ldr(DestReg, REG_SECCOMP_DATA, Offset););
+    break;
+  }
+  case BPF_MEM:
+    if (Inst->k >= 16) {
+      ///< Larger than scratch space size.
+      RETURN_ERROR(-EINVAL);
+    }
+
+    EMIT_INST(ldr(DestReg, REG_SECCOMP_DATA, offsetof(WorkingBuffer, ScratchMemory[Inst->k])););
```
alyssarosenzweig, line 58: `EMIT_INST(ldr(..););` isn't too pretty, and EMIT_INST requires that there's exactly 1 instruction emitted anyway for correct size calculation. Could we insert the trailing `;` in EMIT_INST itself and then `s/;);/);/g` throughout this file?

neobrain:
> ```c++
> uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
>   if (Data->instruction_pointer >= 0x10'0000) return SECCOMP_RET_KILL;
>   return SECCOMP_RET_ALLOW;
> }
> ```
> 
> How do we resolve this? Remove the check?

What's the problem?

> ```c++
> uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
>   if (Data->arch != ARCH_i386) return SECCOMP_RET_KILL;
>   if (Data->nr != SYS_socketcall) return SECCOMP_RET_KILL;
>   return SECCOMP_RET_ALLOW;
> }
> ```
> 
> How do we resolve this? socketcall doesn't even exist on arm64.

If FEX provided a replacement data block for `seccomp_data` that has the same content as if the filter ran on an x86 machine, this program wouldn't need any specific treatment.

> ```c++
> uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
>   Global_Tmp[0] = Data->arch;
>   Global_Tmp[0] += 0x41424344;
>   if (Global_Tmp[0] != (ARCH_i386 + 0x41424344) return SECCOMP_RET_KILL;
>   return SECCOMP_RET_ALLOW;
> }
> ```
> 
> Are we planning to write a sufficiently advanced static analysis pass that can parse through arbitrary ALU operations, branching, and whatever else a turing complete language can muster to ensure we can handle arch number differences, syscall number differences, syscalls missing entirely, syscall argument order differences, argument size differences, and anything else?

In this case also, replacing the `seccomp_data` data block should allow the filter to run without further adaptions.

I might be missing something that is obvious to someone who has worked with BPF, but I don't see any fundamental issue with these examples per se.

Sonicadvance1:
> > For testing Wine, enable the config option and run Red Dead Redemption 2. Proton will stop complaining about seccomp in the logs
> 
> This isn't possible without a separate Rockstar account, right?
The launcher itself enables seccomp. Rockstar accounts are free, the game is available on Steam.

> 
> > For testing Chromium, build it from source, run it without --no-sandbox, watch it explode from things unrelated to seccomp for now. I had some hacks to work around the remaining sandbox problems to test it.
> 
> So practically speaking it's not possible to verify for the time being?
With Chrome no, because chromium fails with things unrelated to seccomp.

> 
> > For another tool, use the Linux unittest linux/tools/testing/selftests/seccomp/seccomp_bpf which runs through a test suite.
> 
> Fair. Is it expected to pass with the current implementation?

The pass results are currently in a comment at the top of SeccompEmulator.cpp
> 
> > So....No. It won't work.
> 
> It's already evident this is your preferred implementation strategy since you opened the PR as-is. Does it hurt to explore the design space? I would feel more comfortable with dropping almost 1600 new lines of code with a huge gap in test coverage if I understood that other options have been thoroughly explored.

I did explore the design space and determined that static analysis of code (bpf filters) won't cover the use cases.

> 
> > The language (BPF) is effectively turing complete with access to temporary memory.
> 
> I thought the architecture constant was part of a dedicated, seccomp-specific storage area that is not accessible through normal load/store instructions. Is this incorrect? (If it is, what prevents us from adding runtime-checks to catch it?)
> 

seccomp-bpf has a load instruction which can load any piece of data from `struct seccomp_data`. This data structure has the arch number, the syscall number, the RIP, and all six syscall arguments passed in to it.
Additionally it has generic 32-bit load and store instructions in to a scratch space which is an array of 16 32-bit values. Which works in an accumulator or index register.
It also has ALU operations, branches, comparison operations.

> > syscall numbers don't match between x86, x86-64, and aarch64. So....No. It won't work.
> 
> How are system call numbers used in BPF? A quick look through the documentation didn't bring up anything obvious.
> 

Comes from the `struct seccomp_data` structure that the kernel fills out.

> > Additionally, true seccomp on top of FEX means it can break syscalls that FEX itself needs and then there is literally no way to work around that.
> 
> Do we _need_ to install guest seccomp filters as host seccomp filter? Can't we run the recompiled seccomp as a normal BPF program that we call as part of each system call?

Without root we don't have access to a "normal BPF program", the `bpf` system call returns EPERM and is disjoint from seccomp-bpf, it's just reusing part of the language.




> > ```c++
> > uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
> >   if (Data->instruction_pointer >= 0x10'0000) return SECCOMP_RET_KILL;
> >   return SECCOMP_RET_ALLOW;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How do we resolve this? Remove the check?
> 
> What's the problem?
RIP of FEX's syscalls will never match the RIP of the x86 `syscall` or `int 0x80` location in memory.

> 
> > ```c++
> > uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
> >   if (Data->arch != ARCH_i386) return SECCOMP_RET_KILL;
> >   if (Data->nr != SYS_socketcall) return SECCOMP_RET_KILL;
> >   return SECCOMP_RET_ALLOW;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > How do we resolve this? socketcall doesn't even exist on arm64.
> 
> If FEX provided a replacement data block for `seccomp_data` that has the same content as if the filter ran on an x86 machine, this program wouldn't need any specific treatment.

If we use a host seccomp-bpf filter, we don't have control of a replacement `seccomp_data` structure.

> 
> > ```c++
> > uint64_t filter(uint32_t Acc, uint32_t Index, uint32_t Tmp, uint32_t Tmp2, seccomp_data* Data) {
> >   Global_Tmp[0] = Data->arch;
> >   Global_Tmp[0] += 0x41424344;
> >   if (Global_Tmp[0] != (ARCH_i386 + 0x41424344) return SECCOMP_RET_KILL;
> >   return SECCOMP_RET_ALLOW;
> > }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Are we planning to write a sufficiently advanced static analysis pass that can parse through arbitrary ALU operations, branching, and whatever else a turing complete language can muster to ensure we can handle arch number differences, syscall number differences, syscalls missing entirely, syscall argument order differences, argument size differences, and anything else?
> 
> In this case also, replacing the `seccomp_data` data block should allow the filter to run without further adaptions.
> 
> I might be missing something that is obvious to someone who has worked with BPF, but I don't see any fundamental issue with these examples per se.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
neobrain: Macros that don't end on `;` at use site combine poorly with clang-format, see #3878.

alyssarosenzweig: I'm talking about the other `;`

neobrain: Ah, the thought crossed my mind but then I read "trailing" and thought you meant the latter. Apologies :)

neobrain:
> seccomp-bpf has a load instruction which can load any piece of data from `struct seccomp_data`. This data structure has the arch number, the syscall number, the RIP, and all six syscall arguments passed in to it. Additionally it has generic 32-bit load and store instructions in to a scratch space which is an array of 16 32-bit values. Which works in an accumulator or index register. It also has ALU operations, branches, comparison operations.

This sounds compatible with my idea of patching this load instruction and redirect any access to the host `seccomp_data` to a FEX-managed substitute buffer.

> > What's the problem?

> RIP of FEX's syscalls will never match the RIP of the x86 `syscall` or `int 0x80` location in memory.

With a substitute for `seccomp_data`, we can fill in the guest RIP instead.

> > If FEX provided a replacement data block for `seccomp_data` that has the same content as if the filter ran on an x86 machine, this program wouldn't need any specific treatment.

> If we use a host seccomp-bpf filter, we don't have control of a replacement `seccomp_data` structure.

We can patch the instructions that load data from `seccomp_data`, which will effectively hide the host's `seccomp_data` from the guest and ensure it only sees the data it's supposed to see.

> Without root we don't have access to a "normal BPF program", the `bpf` system call returns EPERM and is disjoint from seccomp-bpf, it's just reusing part of the language.

That sounds more tricky to resolve. Sounds like we have to refer to host seccomp indeed then. With a reliable way to detect whether the filter is handling a guest-triggered syscall or a FEX-internal one, we could simply make FEX's seccomp preamble return early unless handling a guest handle, though.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
Sonicadvance1: Moved the semi-colon in to the define so we don't have the double semi-colon thing.

Sonicadvance1:
> That sounds more tricky to resolve. Sounds like we have to refer to host seccomp indeed then. With a reliable way to detect whether the filter is handling a guest-triggered syscall or a FEX-internal one, we could simply make FEX's seccomp preamble return early unless handling a guest syscall, though.

You'll need to explain the implementation details of how you are envisioning this will work. I can't wrap my head around it.

neobrain:
> You'll need to explain the implementation details of how you are envisioning this will work. I can't wrap my head around it.

There's a number of ways it could be done, but here's one option: Provide a thread-local boolean to indicate whether the filter is running as part of handling a guest syscall. This boolean would be used in two places:
* Read in the host seccomp filter: A FEX-internal preamble checks the boolean and early-returns `SECCOMP_RET_ALLOW` if `false` or continues to run the guest-provided BPF code if `true`.
* Written in FEX's guest syscall handlers (i.e. the lambda functions in Syscalls.cpp): Before calling the host system call, we set the boolean to `true` and back to `false` right after. For example, consider `pipe`:
```cpp
  REGISTER_SYSCALL_IMPL_FLAGS(pipe, SyscallFlags::OPTIMIZETHROUGH | SyscallFlags::NOSYNCSTATEONENTRY,
                              [](FEXCore::Core::CpuStateFrame* Frame, int pipefd[2]) -> uint64_t {
                                SetGuestSeccompFilterEnabled(Frame, true);
                                uint64_t Result = ::pipe(pipefd);
                                // NOTE: Needs further plumbing if SIGSYS is raised
                                SetGuestSeccompFilterEnabled(Frame, false);
                                SYSCALL_ERRNO();
                              });
```

One concern might be how exactly to store the boolean. Conveniently, assuming the `seccomp_data` problem is taken care of, the seccomp filter will already have access to a thread-local buffer of data under control of FEX. We can just extend this buffer with the new boolean then.

Does that make it clearer?

Sonicadvance1:
> * Read in the host seccomp filter: A FEX-internal preamble checks the boolean and early-returns `SECCOMP_RET_ALLOW` if `false` or continues to run the guest-provided BPF code if `true`.

Is this making an assumption that seccomp filters can read arbitrary memory? I don't understand how we would provide the ability to read a boolean inside of a host seccomp filter when we only have access to loadstore from two locations
- `struct seccomp_data`- read-only - kernel fills in the data
- `uint32_t Scratch[16];` - read-write, temporary data the kernel sets to zero between each installed seccomp filter.

> Conveniently, assuming the `seccomp_data` problem is taken care of, the seccomp filter will already have access to a thread-local buffer of data under control of FEX. We can just extend this buffer with the new boolean then.

This sounds like where the disconnect is coming from, since FEX can't control the seccomp_data that is fed in to a seccomp filter. It's fully under the control of the Linux kernel.

neobrain:
> Is this making an assumption that seccomp filters can read arbitrary memory?

Not arbitrary memory, but a specific memory location set up externally. Is this not possible?

> This sounds like where the disconnect is coming from, since FEX can't control the seccomp_data that is fed in to a seccomp filter. It's fully under the control of the Linux kernel.

I outlined how to address this before, but it's hinging on the previous question.

Sonicadvance1:
> > Is this making an assumption that seccomp filters can read arbitrary memory?
> 
> Not arbitrary memory, but a specific memory location set up externally. Is this not possible?

This is not possible. The load instructions in BPF can only constant index in to `struct seccomp_data` or scratchpad, and the stores can only constant index in to scratchpad.

neobrain:
> This is not possible. The load instructions in seccomp-bpf can only constant index in to struct seccomp_data or scratchpad, and the stores can only constant index in to scratchpad.

Is there any upstream documentation on this type of constraint?

Sonicadvance1:
> > This is not possible. The load instructions in seccomp-bpf can only constant index in to struct seccomp_data or scratchpad, and the stores can only constant index in to scratchpad.
> 
> Is there any upstream documentation on this type of constraint?

https://www.kernel.org/doc/Documentation/networking/filter.txt
Look for references to `seccomp_data` and`M[`. `M[]` referring to the scratchpad.
Also `BPF engine and instruction set` has a description of all the instructions. Not all of which are available in seccomp-bpf.

neobrain:
> Not all of which are available in seccomp-bpf.

Yes, I meant specifically the restrictions of seccomp-bpf. The BPF documentation isn't difficult to find, but I can't find anything on seccomp-bpf.

Sonicadvance1:
> > Not all of which are available in seccomp-bpf.
> 
> Yes, I meant specifically the restrictions of seccomp-bpf. The BPF documentation isn't difficult to find, but I can't find anything on seccomp-bpf.

I don't know if there is a good location for the restrictions other than the kernel validation code. The main restriction for the instructions from that documentation's list is that 8-bit, 16-bit, and 64-bit loads aren't allowed. And all the `addressing modes` aren't allowed other than constant encodings.

Sonicadvance1:
For completeness, here is the validation function in the kernel that directly limits seccomp-bpf's scope https://github.com/torvalds/linux/blob/master/kernel/seccomp.c#L280-L349

https://www.kernel.org/doc/Documentation/bpf/verifier.rst Also refers directly to source as documentation.

neobrain:
> For completeness, here is the validation function in the kernel that directly limits seccomp-bpf's scope https://github.com/torvalds/linux/blob/78eb4ea25cd5fdbdae7eb9fdf87b99195ff67508/kernel/seccomp.c#L280-L348
> 
> https://www.kernel.org/doc/Documentation/bpf/verifier.rst Also refers directly to source as documentation.

Normal BPF supports `BPF_MEM` or `BPF_CALL` (and `bpf_probe_read_user`) for reading memory, but these are disallowed for seccomp :(

I'd like to get back to this idea:

> > Do we need to install guest seccomp filters as host seccomp filter? Can't we run the recompiled seccomp as a normal BPF program that we call as part of each system call?

> Without root we don't have access to a "normal BPF program", the bpf system call returns EPERM and is disjoint from seccomp-bpf, it's just reusing part of the language.

Is this really true anymore? [Documentation](https://www.man7.org/linux/man-pages/man2/bpf.2.html) suggests recent kernels let unprivileged users create programs of type `BPF_PROG_TYPE_SOCKET_FILTER` specifically, which can then be ran in isolation using `BPF_PROG_RUN`. That's still restricting, but we get the ability to pass in a custom context, which we can then use for our custom `seccomp_data`.

This approach would require some more plumbing due to not leveraging host seccomp directly, but at least it would sort out the BPF interpretation.

alyssarosenzweig:
It's not clear to me how "piles of magic tricks to translate run x86 bpf programs on arm64 bpf kernel" is actually going to be simpler / more maintainable / faster / ...? than the straightforward JIT here.

Sonicadvance1:
> Is this really true anymore? [Documentation](https://www.man7.org/linux/man-pages/man2/bpf.2.html) suggests recent kernels let unprivileged users create programs of type `BPF_PROG_TYPE_SOCKET_FILTER` specifically, which can then be ran in isolation using `BPF_PROG_RUN`. That's still restricting, but we get the ability to pass in a custom context, which we can then use for our custom `seccomp_data`.

Documentation is outdated, `unprivileged_bpf_disabled` defaults to `2` now which means unprivileged users can't use it at all.

```bash
$ strace ./bpf
<...>
bpf(BPF_PROG_LOAD, <...>) = -1 EPERM (Operation not permitted)

$ sudo strace ./bpf
<...>
bpf(BPF_PROG_LOAD, <...>) = 3
```

neobrain:
> It's not clear to me how "piles of magic tricks to translate run x86 bpf programs on arm64 bpf kernel" is actually going to be simpler / more maintainable / faster / ...? than the straightforward JIT here.

Discussed this on Discord. The 1.6kloc in this PR have poor test coverage (only one reproducible real-world use case; the seccomp-bpf selftests don't cover the actual ISA). Leveraging the host kernel's BPF engine would address this, which is why it's useful to explore alternatives even if the JIT is deemed straightforward.

> Documentation is outdated, `unprivileged_bpf_disabled` defaults to `2` now which means unprivileged users can't use it at all.

Ah, sounds like that settles it indeed. I don't see any other options either then, since pure static analysis as you mentioned isn't viable.

Sonicadvance1:
Poking around at Chrome's sandbox with this PR
- Seems to work successfully
- Chrome hits two unrelated bugs when running the sandbox that are unrelated to this PR
  - CLONE_NEWNET is currently unsupported in FEX, as it loses connection to FEXServer. I have a WIP branch locally to support this
  - FEX hits some `EDEADLK` futex locks for some reason, potentially due to forking Chrome does. With some massaging of our various mutexes, I was able to still get Chrome rendering with the sandbox enabled.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
```diff
@@ -0,0 +1,397 @@
...
+
+#include <linux/bpf_common.h>
+#include <linux/filter.h>
+#include <linux/seccomp.h>
+
+namespace FEX::HLE {
+template<bool CalculateSize>
+uint64_t BPFEmitter::HandleLoad(uint32_t BPFIP, const sock_filter* Inst) {
+  [[maybe_unused]] size_t OpSize {};
+  switch (BPF_SIZE(Inst->code)) {
```
alyssarosenzweig, line 19: This switch would be probably better as:

```
if (BPF_SIZE() != W) {
   RET_ERR(EINVAL);
}
```

alyssarosenzweig:
Could we get a macro:

```
#define VALIDATE(cond) if (!(cond)) RETURN_ERROR(-EINVAL)
````

Then porting the file to this would let us write straightline code instead which should be a lot shorter and imho simpler. e.g.:

```
+  case BPF_JA: {
+    // Only BPF_K supported on JA.
+    VALIDATE((SrcType != BPF_X);
+
+    // BPF IP register is effectively only 32-bit. Treat k constant like a signed integer.
+    // This allows it to jump anywhere in the program.
+    // But! Loops are EXPLICITLY disallowed inside of BPF programs.
+    // This is to prevent DOS style attacks through BPF programs.
+    uint64_t Target = BPFIP + Inst->k + 1;
+
+    // Can't jump past the end.
+    VALIDATE (Target < NumInst);
+
+    fextl::unordered_map<uint32_t, ARMEmitter::ForwardLabel>::iterator TargetLabel {};
+
+    if constexpr (!CalculateSize) {
+      TargetLabel = JumpLabels.try_emplace(Target, ARMEmitter::ForwardLabel {}).first;
+    }
+
+    EMIT_INST(b(&TargetLabel->second));
+    break;
+  }
```

Could even take a string as a second arg to match our logman asserts.

**Comment by Sonicadvance1** [comment]
> Could we get a macro:
> 
> ```
> #define VALIDATE(cond) if (!(cond)) RETURN_ERROR(-EINVAL)
> ```
> 
> Then porting the file to this would let us write straightline code instead which should be a lot shorter and imho simpler. e.g.:
> 
> ```
> +  case BPF_JA: {
> +    // Only BPF_K supported on JA.
> +    VALIDATE((SrcType != BPF_X);
> +
> +    // BPF IP register is effectively only 32-bit. Treat k constant like a signed integer.
> +    // This allows it to jump anywhere in the program.
> +    // But! Loops are EXPLICITLY disallowed inside of BPF programs.
> +    // This is to prevent DOS style attacks through BPF programs.
> +    uint64_t Target = BPFIP + Inst->k + 1;
> +
> +    // Can't jump past the end.
> +    VALIDATE (Target < NumInst);
> +
> +    fextl::unordered_map<uint32_t, ARMEmitter::ForwardLabel>::iterator TargetLabel {};
> +
> +    if constexpr (!CalculateSize) {
> +      TargetLabel = JumpLabels.try_emplace(Target, ARMEmitter::ForwardLabel {}).first;
> +    }
> +
> +    EMIT_INST(b(&TargetLabel->second));
> +    break;
> +  }
> ```
> 
> Could even take a string as a second arg to match our logman asserts.

Added a `VALIDATE` define. No need for logging because these are non-fatal errors, they are ensuring that seccomp-bpf applications don't try to do something invalid.

File: Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp (hunk 4 above)
Comment (Sonicadvance1) [reply Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp]: Used the `VALIDATE` define

**Comment by alyssarosenzweig** [comment]
Those switches should probably all have a default: EINVAL on them

**Comment by alyssarosenzweig** [comment]
> Added a `VALIDATE` define. No need for logging because these are non-fatal errors, they are ensuring that seccomp-bpf applications don't try to do something invalid.

The problem is that 

```
+  // Larger than scratch space size.
+  VALIDATE(Inst->k < 16);
```

is very confusing. In some ways the comment makes this *more* confusing. since we validate the opposite. Clearer would be


```
+  // Must be smaller than scratch space size.
+  VALIDATE(Inst->k < 16);
```

or

```
+  VALIDATE(Inst->k < 16, "Larger than scratch space size");
```

**Comment by alyssarosenzweig** [comment]
(That goes for all the validation I think)

File: Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp [hunk 5]
```diff
@@ -0,0 +1,378 @@
...
+      auto Const = ConstPool.try_emplace(Inst->k, ARMEmitter::ForwardLabel {});
+      EMIT_INST(ldr(CompareSrcReg, &Const.first->second));
+    }
+    uint32_t TargetTrue = BPFIP + Inst->jt + 1;
+    uint32_t TargetFalse = BPFIP + Inst->jf + 1;
+
+    // Trying to jump past the end.
+    VALIDATE(TargetTrue < NumInst && TargetFalse < NumInst);
+
+    auto CompareResultOp = ARMEmitter::Condition::CC_EQ;
```
alyssarosenzweig, line 185: Default value is confusing here, how about just `ARMEmitter::Condition CompareResultOp;`

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
```diff
@@ -0,0 +1,378 @@
...
+    }
+    case BPF_ST:
+    case BPF_STX: {
+      Result = HandleStore<CalculateSize>(i, Inst);
+      HadError = PredFunc(Result);
+      break;
+    }
+    case BPF_ALU: {
+      Result = HandleALU<CalculateSize>(i, Inst);
+      HadError = PredFunc(Result);
```
alyssarosenzweig, line 287: Pop this line out of the switch. Also, does this need a default?

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/Dumper.cpp
```diff
@@ -0,0 +1,147 @@
...
+#include "LinuxSyscalls/Seccomp/SeccompEmulator.h"
+
+#include <linux/bpf_common.h>
+#include <linux/filter.h>
+#include <linux/seccomp.h>
+
+namespace FEX::HLE {
+void SeccompEmulator::DumpProgram(const sock_fprog* prog) {
+  auto Parse_Class_LD = [](uint32_t BPFIP, const sock_filter* Inst) {
+    auto DestName = [](sock_filter const* Inst) {
```
alyssarosenzweig, line 17: These nested lambdas seem gratuitous, what is this buying us over

```
auto DestName = ( (BPF_CLASS(Inst->code) == BPF_LD) ? "A" : "X";
```

alyssarosenzweig: I guess just massively nicer syntax of return vs assign+break. Fair.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/Dumper.cpp
```diff
@@ -0,0 +1,147 @@
...
+  for (size_t i = 0; i < prog->len; ++i) {
+    const sock_filter* Inst = &prog->filter[i];
+    const uint16_t Code = Inst->code;
+    const uint16_t Class = BPF_CLASS(Code);
+    switch (Class) {
+    case BPF_LD:
+    case BPF_LDX: Parse_Class_LD(i, Inst); break;
+    case BPF_ST:
+    case BPF_STX: Parse_Class_ST(i, Inst); break;
+    case BPF_ALU: break;
```
alyssarosenzweig, line 140: Why are we dumping load/store/jumps but not alu?

alyssarosenzweig:
All the non-JIT stuff seems reasonable, though I'm obviously less familiar

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/Dumper.cpp
Sonicadvance1: True, I didn't finish that off. Added the two handlers.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
Sonicadvance1: Popped out. default handler switched to FEX_UNREACHABLE since we handle all instruction classes.

Source/Tools/LinuxEmulation/LinuxSyscalls/Seccomp/BPFEmitter.cpp
Sonicadvance1: Default value removed.

Sonicadvance1:
> > Added a `VALIDATE` define. No need for logging because these are non-fatal errors, they are ensuring that seccomp-bpf applications don't try to do something invalid.
> 
> The problem is that
> 
> ```
> +  // Larger than scratch space size.
> +  VALIDATE(Inst->k < 16);
> ```
> 
> is very confusing. In some ways the comment makes this _more_ confusing. since we validate the opposite. Clearer would be
> 
> ```
> +  // Must be smaller than scratch space size.
> +  VALIDATE(Inst->k < 16);
> ```
> 
> or
> 
> ```
> +  VALIDATE(Inst->k < 16, "Larger than scratch space size");
> ```

Updated the comments around the VALIDATES

Sonicadvance1:
> Those switches should probably all have a default: EINVAL on them

`default:RETURN_ERROR(-EINVAL);` added to the switches that missed them.

alyssarosenzweig (approved):

PR #3632, opened 2024-05-16
Use erase-remove idiom to remove element

### LINKED ISSUE #3631: Blocks are not physically removed
I haven't looked into the algo in detail but it feels to me like we should not ignore the return value of `std::remove`. This does not physically remove the Predecessor, simply moves it to another place. We should probably use `std::erase` here. See https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

https://github.com/FEX-Emu/FEX/blame/c5f8ea58e91df8ad73a8efb6ae2b0206662ea24c/FEXCore/Source/Interface/IR/Passes/RAValidation.cpp#L246

pmatos:
Fixes #3631


neobrain (approved):
Good catch!

Sonicadvance1 (approved):

PR #3633, opened 2024-05-16
Fix left shift undefined behaviour

pmatos:
Here BitOffset can have values higher than 32.


pmatos:
There's actually a weird situation here and I might be missing expectations... `IsNZCV` is called with `BitOffset` higher than 32 through `GetRFLAG` and `SetRFLAG`. But the result, is passed as a 32bit integer to `ContainsNZCV`. Maybe because NZCV Mask is within the lower 32bits, we can just return false if the BitOffset is higher than 32.

alyssarosenzweig:
Nice catch!

alyssarosenzweig (approved):

PR #3637, opened 2024-05-17
Rewrite RA

alyssarosenzweig:
I recommend viewing the new source file as the diff is quite messy.
    
    ---
    
    The old RA commits every "how not to write an RA" sin in the book.
    
    Chaitin spill-one loop? Check.
    
    Potential spilling caused by alignment issues since there's no live range
    splitting? Check.
    
    Panic spilling? Check.
    
    Generating an interference graph with linear live ranges, so you get the code
    quality of linear scan with the cost of graph colouring? Check.
    
    ...
    
    It is wholly unsuitable to any application, and specifically unsuitable for FEX.
    
    ---
    
    The new RA exploits a key IR invariant unique to FEX: no values are live across
    block boundaries. This is validated.
    
    Because of this invariant, all RA is block local. This lets us use a dead simple
    2 pass RA that generates ~optimal code in linear time.
    
    The first pass walks the IR backwards, analyzing the IR. This is a souped up
    analogue to liveness analysis.
    
    The second pass walks the IR forward, blasting out registers. If necessary, it
    will insert spill and/or shuffle code on the fly. Spilling uses the well-known
    furthest-first heuristic, which has excellent results for straight line code.
    
    That's it :-)


FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,676 @@ tags: ir|opts
...
+using namespace FEXCore;
 
-#include <algorithm>
-#include <cstddef>
-#include <cstdint>
-#include <cstring>
-#include <optional>
-#include <strings.h>
-#include <utility>
+#define foreach_arg(IROp, index, arg) \
```
neobrain, line 22: Could we avoid control-flow altering macros like these?

The file-specific idioms they add make reading source code more difficult/surprising when the same thing could be expressed much more clearly with slightly lengthier code. Tooling also tends to handle this poorly (such as debuggers that can't reliably step through code, or IDEs that don't know what line to show when jumping to definitions, etc).

The C++ way to do this equally concisely is to implement a custom iterator (or use C++23 range views), but since both macros don't seem to be used very often we can equally well just use plain index-based loops and initialize Arg in the loop body.

alyssarosenzweig: Frankly, the C++ ways of doing things have made a mess of FEXCore and have made my work improving FEXCore needlessly challenging. By contrast, the C ways of doing things have served us well in Mesa. I don't think this is a coincidence.

I'm not inclined to make my code worse because it's not C++ enough, I'd rather make the rest of FEXCore more like the competent compilers we have in Mesa. I care more about code being  simple, short, and efficient.. than about it being C++ish.

neobrain: > Frankly, the C++ ways of doing things have made a mess of FEXCore and have made my work improving FEXCore needlessly challenging. By contrast, the C ways of doing things have served us well in Mesa. I don't think this is a coincidence.

It might be worth having a broader discussion about what idioms you've found challenging, but let's focus on this specific instance for the time being.

As mentioned I'm happy with (and actually recommend) a macro-less "C way" too. I wasn't arguing out of pure principle:

> The file-specific idioms they add make reading source code more difficult/surprising when the same thing could be expressed much more clearly with slightly lengthier code.

> Tooling also tends to handle this poorly (such as debuggers that can't reliably step through code, or IDEs that don't know what line to show when jumping to definitions, etc).

To illustrate the first point, it's impossible for me to predict what `foreach_arg(Header, _, Arg) {` will do without jumping between the macro definition and its use several times (even more so since Header is declared without an explicit type). Is this code *simple*? I would consider *local* code that uses standard language syntax simpler, since it clearly and succinctly expresses what variables are being added to the scope.

If the macro were widely used and implemented heavily complicated and repeated logic, I could see some merit in this approach. But when the macros are only used 2/3 times each, what value are they really adding here?

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,676 @@ tags: ir|opts
...
+  fextl::vector<uint32_t> NextUses;
 
-  Graph = AllocateRegisterGraph(ClassCount);
+  unsigned SpillSlotCount;
 
-  // Add identity conflicts
-  for (uint32_t Class = 0; Class < INVALID_CLASS; Class++) {
-    for (uint32_t Reg = 0; Reg < INVALID_REG; Reg++) {
-      AddRegisterConflict(RegisterClassType {Class}, Reg, RegisterClassType {Class}, Reg);
+  bool IsValidArg(auto Arg) {
```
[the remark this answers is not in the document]
alyssarosenzweig: will fix, didn't mean for auto's to slip in, not used to driving a monster truck

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
alyssarosenzweig: I typed a big response to this then Firefox ate it 😭 

Consider some code samples:

* https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/asahi/compiler/agx_opt_empty_else.c
* https://gitlab.freedesktop.org/mesa/mesa/-/blob/main/src/asahi/compiler/agx_register_allocate.c#L599
* https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/IR/Passes/LongDivideRemovalPass.cpp
* https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/IR/Passes/InlineCallOptimization.cpp

The C code is compact with almost no boilerplate. It expresses complex logic in the language of compiler theory, emphasizing the algorithms and eschewing the specific arrays and lists in the IR. It was easy and fast to write. It is easy and fast to read, facilitated by individually simple helpers that compose to form a rhythm greater than the sum of the parts. It performs well. It will be easy to maintain later. It follows AGX compiler idioms: it is not standard C but it is standard "AGX compiler C".

The C++ code is less compact with piles of pointless class boilerplate. It has fewer expressive helpers, so there is more emphasis on the underlying IR than just the algorithm. Judging by my experiences on this RA and the previous NZCV work, it would have been harder and slower (= more expensive) to write - at least for me. It is - at least for me - harder and slower to read. It is not standard C++ but it is standard "FEXCore C++".

FEXCore has significant performance problems. I want to fix them. Doing so will require rewriting much of FEXCore, redesigning FEXCore, establishing new idioms. Already this RA pass is made far more complex (= slower *and* much more code *and* more expensive to write) because of entrenched FEXCore problems -- Map/Unmap/Remap and Swap2 are just to work around IR deficiencies. The former slows down RA, the latter impedes us from efficiently implementing `pop` instructions. Fixing this is hard ... in part because of the problematic iterators that FEXCore *does* have!

Given the experiences, my inclination is to run as far from C++ as possible. Maybe it's not C++'s fault, but I don't know how to write a good compiler in idiomatic C++ and I do know how to write a good one in non-idiomatic C. I've seen what happens when our team (including me) tries in idiomatic C++, and it isn't working well. Time to try something better. I'm happy to keep `vector` and tuples, but "idiomatic C++" or "idiomatic C" are explicit non-goals for me, in favour of "clean, obvious, fast, compact code" and eventually "idiomatic FEXCore C++".

In my (apparently quixotic) quest to fix FEXCore, I can either attempt a wholesale rewrite or I can improve incrementally. But I can't make incremental progress if the objection is that it doesn't match how we do things today. Well, yeah, that's kind of the point - what we're doing now isn't working but I know something else that does.

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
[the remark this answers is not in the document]
alyssarosenzweig: I think all the auto args are gone now, much better.

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
neobrain: > I typed a big response to this then Firefox ate it 😭

Having been there and done that, I appreciate you've taken the time to engage in the discussion :)

> Consider some code samples:

Granted, one of these was created by an experienced compiler engineer (you) and uses proven Mesa idioms that passed the test of time, whereas the other was written without this specific expertise and for the longest time focused on getting things running at all more so than on running well.

However, what's clear from your reported experience is that FEX's JIT is not a poster child of good C++. Maybe C++ doesn't even apply all that well to compilers in general. I'm not convinced the latter is the case, but it is what it is regarding the former.

> Given the experiences, my inclination is to run as far from C++ as possible.

If specific patterns fail to apply well to FEX's JIT we should avoid them and stick to simpler approaches. I reject this as an argument for automatically resorting to non-idiomatic code just because the alternative uses C++ however. It helps no one if we fragment FEX into a mix of 20 different coding styles, because I certainly am not convinced that extending Mesa's C-only patterns into FEX's non-JIT modules is a net benefit.

Be that as it may, maybe your approach is a pragmatic solution to dealing with the accumulated debt in the FEX JIT. In this particular instance, this pragmatism leads to a counter-factual decision though:

> Fixing this is hard ... in part because of the problematic iterators that FEXCore does have!

What is the concrete benefit of the `foreach` macros you've added here? The best point I can think of is refactoring, but that's a mood point when - again - we're talking about 5 instances.

I see you've at least dropped the use of `std::tuple` + structured binding from the for-loop to make the macro definitions themselves easier on the eyes. Meanwhile it's still a non-local, file-specific idiom that saves you a minimal number of keystrokes when the alternative of spelling out the for-loops directly code is already simple and and equally performant.

> I can either attempt a wholesale rewrite or I can improve incrementally. But I can't make incremental progress if the objection is that it doesn't match how we do things today

I'm open to incremental improvements, though they should indeed be improvements - otherwise we're just pushing code around for individual convenience. Maybe your ideas can be illustrated more effectively?

Sonicadvance1: I'm backing Alyssa's decision here. The paradigms in the code are changing and if C++ is getting in the way then just get rid of the problems for the developer. We are always going to have multiple coding styles in a project, there's no getting around that, we are all squishy humans that write things differently.

We can't stomp on momentum of one person working on new features/implementations because the idioms are closer to C than C++, we'd never get any work done. FEXCore (and more specifically the JIT) is going to be moving more towards C because the people working on the code is more familiar with C, that's completely fine.

Trying to strictly enforce that the codebase must be homogenous in coding style and behaviour is a foolish endeavour and will only lead to burn-out.

neobrain: > I'm backing Alyssa's decision here. The paradigms in the code are changing

As mentioned multiple times, I'm not arguing against change.

> and if C++ is getting in the way then just get rid of the problems for the developer.

This *is* what I argued for. At the same time, C++ is *not* getting in the way of the specific point I commented on.

> Trying to strictly enforce that the codebase must be homogenous in coding style and behaviour is a foolish endeavour and will only lead to burn-out.

No one was trying to enforce a homogenous coding style.

We can agree this general debate isn't useful if it happens on the side of an actual change request. Meanwhile I have raised specific points related to the proposed changes that have not been discussed. Could we stick to the specific change I highlighted about this PR?

Sonicadvance1:
Still reviewing the code but it looks like there is some validation disabled. Does that need to get re-enabled, changed, deleted, or something?

Sonicadvance1:
> All of this waffling is to say: I don't know whether I should spend the time now to fix up a validation pass that will ultimately be deleted. So I punted on the question for now, since there's enough other code to review meanwhile anyhow.

Sounds good to me

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
alyssarosenzweig: Dropped the macros, I'm not interested in further litigating this here.

cwabbott0:
> Where does that leave us? In truth, I'm not sure yet. RA validation was undoubtedly useful when bringing up this code, I don't know if I'm ready to close the door behind me, and I don't know how to have our cake and eat it too. (The cake was a lie.)

FWIW, in pretty much all Mesa backends, we make this work without much of a performance penalty by storing both the SSA pointer and the assigned register in the IR source/destination itself. Every other framework I know of, like LLVM, does this too. RA validation has been invaluable in ir3 for catching things that would be hard or impossible to track down bugs without it.

alyssarosenzweig:
> > Where does that leave us? In truth, I'm not sure yet. RA validation was undoubtedly useful when bringing up this code, I don't know if I'm ready to close the door behind me, and I don't know how to have our cake and eat it too. (The cake was a lie.)
> 
> FWIW, in pretty much all Mesa backends, we make this work without much of a performance penalty by storing both the SSA pointer and the assigned register in the IR source/destination itself. Every other framework I know of, like LLVM, does this too. RA validation has been invaluable in ir3 for catching things that would be hard or impossible to track down bugs without it.

Hmm, ok. I think we can make this work without too much bloat in the future. Thanks for chiming in 👍

I was hoping to make sources/dests 16-bit - which we can get away with because we choose the block size, and there's diminishing returns for FEX optimization past a few dozen instructions anyway. FEX is not supposed to be a real compiler. We need "just enough" IR to translate straight line code and that's it. Pure SSA with no phi nodes 🤡. Making the IR as thin as possible is desireable to keep JIT time in check, since I think we're far more JIT time sensitive than even a shader compiler.

Though maybe 32-bit source/dests, partitioned as "24-bit SSA index + 8-bit register" is the happy medium here, and then we don't have to artificially slice up large blocks. Either way, storing the register inside the instruction instead of in a sideband should get us most of the benefit for the relevant issues.

cwabbott0:
> Making the IR as thin as possible is desireable to keep JIT time in check, since I think we're far more JIT time sensitive than even a shader compiler.

I'm not so familiar with FEX, so speculation incoming. But I would expect that the IR is allocated using a single-threaded linear allocator or something equally cheap, and then immediately thrown away after translation, so making things larger won't matter that much, as long as the IR is still simple (i.e. without too much pointer chasing).

alyssarosenzweig:
> > Making the IR as thin as possible is desireable to keep JIT time in check, since I think we're far more JIT time sensitive than even a shader compiler.
> 
> I'm not so familiar with FEX, so speculation incoming. But I would expect that the IR is allocated using a single-threaded linear allocator or something equally cheap, and then immediately thrown away after translation, so making things larger won't matter that much, as long as the IR is still simple (i.e. without too much pointer chasing).

Yeah, hope so. There's a question of "how many instructions can we fit in a cache line" and such. But you're right that between flat 8-byte ops and flat 16-byte ops.. either way it's cheap. Anyway, thanks for the input. Will plan on the 24/8 split for the new IR (and hence 16-byte for typical ALU binops), so we keep the RA validation around.

alyssarosenzweig:
I've weakened a single assert in RA validation, but otherwise the hack commit is gone. Thanks all

Sonicadvance1 (approved):
Very fancy stuff. Thanks for rewriting RA, it has needed this for a long time.

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,713 @@ tags: ir|opts
...
+    LOGMAN_THROW_AA_FMT(Reg.Reg == BestReg, "Invariant6");
 
-            SRA_DEBUG("ssa{} cannot be a pre-write because ssa{} reads from sra{} before storereg", ID, Node, -1 /*vreg*/);
-            (*StaticMap)->PrefferedRegister = PhysicalRegister::Invalid();
-            (*StaticMap)->PreWritten.Invalidate();
-            SetNodeClass(Graph, ID, Op->Class);
-          }
+    IROp_Header* Header = IR->GetOp<IROp_Header>(Candidate);
+    uint32_t Value = IR->GetID(Candidate).Value;
+    bool Spilled = SpillSlots.size() > 0 && SpillSlots[Value] != 0;
```
neobrain, line 312:
```suggestion
    bool Spilled = !SpillSlots.empty() && SpillSlots[Value] != 0;
```

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,713 @@ tags: ir|opts
...
+      LOGMAN_THROW_AA_FMT(Reg.Class == GetRegClassFromNode(IR, Header), "Consistent");
 
-            NodeLiveRange.PrefferedRegister = Reg;
-            (*StaticMap) = &NodeLiveRange;
-            SetNodeClass(Graph, Node, RegisterClassType {NodeLiveRange.PrefferedRegister.Class});
-            SRA_DEBUG("Marking ssa{} as allocated to sra{}\n", Node, -1 /*vreg*/);
-          }
-        }
+      // SpillSlots allocation is deferred.
+      if (SpillSlots.size() == 0) {
```
neobrain, line 320:
```suggestion
      if (SpillSlots.empty()) {
```

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,713 @@ tags: ir|opts
...
-          // This would ensure something will spill earlier if its previous use and next use are farther away
-          auto InterferenceNodeNextUse = FindFirstUse(IREmit, InterferenceOrderedNode, NodeOpBeginIter, InterferenceNodeOpEndIter);
-          auto InterferenceNodePrevUse = FindLastUseBefore(IREmit, InterferenceOrderedNode, InterferenceNodeOpBeginIter, NodeOpBeginIter);
-          LOGMAN_THROW_A_FMT(InterferenceNodeNextUse != IR::NodeIterator::Invalid(), "Couldn't find next usage of op");
-          // If there is no use of the interference op prior to our op then it only has initial definition
-          if (InterferenceNodePrevUse == IR::NodeIterator::Invalid()) {
-            InterferenceNodePrevUse = InterferenceNodeOpBeginIter;
-          }
+    // Next-use distance relative to the block end of each source, last first.
+    fextl::vector<uint32_t> SourcesNextUses;
```
neobrain, line 521: How many blocks does RA typically run over? Might make sense to move the declaration of this vector out of the loop and just `.clear()` it at the loop entry to minimize memory reallocations.

FEXCore/Source/Interface/IR/Passes/RAValidation.cpp
```diff
@@ -173,16 +173,13 @@ bool RAValidation::Run(IREmitter* IREmit) {
         const auto Value = BlockRegState.Unspill(FillRegister->Slot);
 
         // TODO: This only proves that the Spill has a consistent SSA value
-        //       In the future we need to prove it contains the correct SSA value
+        // In the future we need to prove it contains the correct SSA value. For
+        // this we need to analysis copies/swaps properly. As a hot fix, don't
```
neobrain, line 177:
```suggestion
        // this we need to analyze copies/swaps properly. As a hot fix, don't
```

neobrain (approved):
Gave this a review pass out of curiosity. I didn't dive too much into the details but I really like the code explanations and the explicit use of pre/post-condition checks!

Just leaving some general suggestions otherwise.

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -5,1239 +5,713 @@ tags: ir|opts
...
+  AllocData = RegisterAllocationData::Create(SSAToReg.size());
+  AllocData->SpillSlotCount = SpillSlotCount;
+  memcpy(AllocData->Map, SSAToReg.data(), sizeof(PhysicalRegister) * SSAToReg.size());
+
+  PreferredReg.clear();
+  SSAToNewSSA.clear();
+  NewSSAToSSA.clear();
+  SSAToReg.clear();
+  SpillSlots.clear();
+  NextUses.clear();
```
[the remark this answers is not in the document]
alyssarosenzweig: I would like to bulldoze the whole pass framework at some point because persisting these allocations is terrifying. In the mean time, this is probably fine.

Sonicadvance1:
Something changed and is now causing crashes when trying to run Steam with this PR.

alyssarosenzweig:
> Something changed and is now causing crashes when trying to run Steam with this PR.

Oh no! Any idea when that regressed? I haven't changed logic today except for responding to the review commens

Sonicadvance1:
> > Something changed and is now causing crashes when trying to run Steam with this PR.
> 
> Oh no! Any idea when that regressed? I haven't changed logic today except for responding to the review commens

Not sure, steamwebhelper is crashing very quickly on startup.

Sonicadvance1:
Needs instcountci rebase

alyssarosenzweig:
rebased

PR #3638, opened 2024-05-18
JIT/VectorOps: deduplicate common implementations

alyssarosenzweig:
Part of my project to clean up FEXCore in prep. for IR go-fast

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -16,6 +16,31 @@ namespace FEXCore::CPU {
 #define GRS(Node) (IROp->Size <= 4 ? GetReg<RA_32>(Node) : GetReg<RA_64>(Node))
 
 #define DEF_OP(x) void Arm64JITCore::Op_##x(IR::IROp_Header const* IROp, IR::NodeID Node)
+
+#define DEF_BINOP_WITH_CONSTANT(FEXOp, VarOp, ConstOp)                                         \
+  DEF_OP(FEXOp) {                                                                              \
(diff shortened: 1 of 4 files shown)


Sonicadvance1 (approved):
Nice

.clang-format
```diff
@@ -7,7 +7,7 @@ AlignConsecutiveAssignments: None
 AlignConsecutiveBitFields: Consecutive
 AlignConsecutiveDeclarations: None
 AlignConsecutiveMacros: None
-AlignEscapedNewlines: DontAlign
+AlignEscapedNewlines: Right
```
neobrain, line 10: This makes side-by-side diffs impossible to read in code review. Can we keep the old behavior or use `Left` here instead?

Also, as a side note, please don't sneak in project-wide formatting changes into otherwise highly specific PRs.

alyssarosenzweig: > This makes side-by-side diffs impossible to read in code review.

How so...?

> Can we keep the old behavior or use Left here instead?

Left would be a big improvement on the old behaviour, but Right is definitely less noisy (since the \'s then go away when reading code).

> Please don't sneak in project-wide formatting changes into otherwise highly specific PRs.

OK. This one seemed trivial but I guess not.

neobrain: > How so...?

The `Right` style apparently always pads up to the line character limit (140), so even on my MacBook with a 2560x1600 resolution, GitHub's side-by-side diff breaks each macro line into two.

> Left would be a big improvement on the old behaviour, but Right is definitely less noisy (since the 's then go away when reading code).

Arguably things for which `\` noise is a distinguishing factor shouldn't be macros in the first place.

Sonicadvance1 (approved):

PR #3639, opened 2024-05-18
FEXLoader: Cleanup FD extraction from environment variables

Sonicadvance1:
In preparation for seccomp execve inheritance where we need to extract another FD from a different environment variable.

- Small function to extract the FD and also unset the environment variable in the same place.
   - Keeping the fetch and unset together instead of spreading to another location in the source.
- Extract the FD upfront instead of passing the string_view around, since we are unsetting the environment variable at the same place.

Future seccomp inheritance will get the FD just after the FEXFD
   - `int FEXSeccompFD {GetFEXFDFromEnv("FEX_SECCOMPFD")};`

### DIFF
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -245,21 +245,38 @@ void SetupTSOEmulation(FEXCore::Context::Context* CTX) {
 }
 } // namespace FEX::TSO
 
+/**
+ * @brief Get an FD from an environment variable and then unset the environment variable.
+ *
+ * @param Env The environment variable to extract the FD from.
+ *
+ * @return -1 if the variable didn't exist.
+ */
+static int StealFEXFDFromEnv(const char* Env) {
+  int FEXFD {-1};
+  const char* FEXFDStr = getenv(Env);
+  if (FEXFDStr) {
+    const std::string_view FEXFDView {FEXFDStr};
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
Looks like `buildbot-solidrun` is flaking instcountci `Sekiro spill block` for some reason?

Sonicadvance1:
ping

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -245,21 +245,38 @@ void SetupTSOEmulation(FEXCore::Context::Context* CTX) {
...
 } // namespace FEX::TSO
 
+/**
+ * @brief Get an FD from an environment variable and then unset the environment variable.
+ *
+ * @param Env The environment variable to extract the FD from.
+ *
+ * @return -1 if the variable didn't exist.
+ */
+static int GetFEXFDFromEnv(const char* Env) {
```
alyssarosenzweig, line 255: Nit: could we rename this to `StealFEXFEDFromEnv`, because of the `unsetenv`? To me, "get" is a pure function whereas "steal" or "take" or "pop" destroys the input it returns.

alyssarosenzweig (approved):

alyssarosenzweig:
Clean up looks great, approved! Just one little naming quibble, take it or leave it.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Renamed.

PR #3640, opened 2024-05-18
LinuxSyscalls: Cleanup envp copying in execve

Sonicadvance1:
In preparation for seccomp execve inheritance.

We are going to need to add a new environment variable earlier in the execve sequence to handle inheritance in the case of binfmt_misc.

No functional change in regards to envp handling.

Minor change around execveat with FD without binfmt_misc. In the case that execveat returned an error and we did a `dup` of the FD then we would have an FD leak. Make sure to close the duplicated FD in that instance.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
@@ -259,38 +259,7 @@ uint64_t ExecveHandler(const char* pathname, char* const* argv, char* const* env
 
   fextl::vector<const char*> EnvpArgs {};
   char* const* EnvpPtr = envp;
-  bool AlreadyCopiedEnvp {};
-
-  auto AppendEnvpVariables = [&EnvpArgs, &AlreadyCopiedEnvp, envp, &EnvpPtr]() {
-    if (AlreadyCopiedEnvp) {
-      ///< If the environment variables were already copied then early return.
-      return;
-    }
-
-    if (envp) {
-      auto OldEnvp = envp;
-      while (*OldEnvp) {
-        ///< Copy the pointers to our own vector of environment variables.
-        EnvpArgs.emplace_back(*OldEnvp);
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
ping

alyssarosenzweig:
I'm not sure I'm qualified to review this one. But if you think so and nobody else knows this code I guess I can try...?

Sonicadvance1:
> I'm not sure I'm qualified to review this one. But if you think so and nobody else knows this code I guess I can try...?

This change is relatively straightforward, just moves code upward and fixes a potential FD leak.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
```diff
@@ -256,36 +257,71 @@ uint64_t ExecveHandler(const char* pathname, char* const* argv, char* const* env
     return -ENOEXEC;
   }
 
+  fextl::vector<const char*> EnvpArgs {};
+  char* const* EnvpPtr = envp;
+  bool AlreadyCopiedEnvp {};
+
+  auto AppendEnvpVariables = [&EnvpArgs, &AlreadyCopiedEnvp, envp, &EnvpPtr]() {
```
neobrain, line 264: This lambda is unnecessary since it's called only once.

More generally, lambda functions are a good fit for short snippets of code with few side effects. Anything more complicated would be better solved with a non-local refactor (to a global function or class) instead.

Sonicadvance1: Indeed, the visibility in to what is going to change in the future is hard here. Future PRs (seccomp and FEXServer FD inheritance) are going to call this up to three times.

Sonicadvance1: I removed the unnecessary lambda usage.

lioncash (approved):

PR #3644, opened 2024-05-20
clang-format: left-align escaped newlines

alyssarosenzweig:
alternative to #3638. this is theoretically better for side-by-side diffs. in practice it may make other diffs worse since all the \'s change when part of the macro change.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -7,7 +7,7 @@ AlignConsecutiveAssignments: None
 AlignConsecutiveBitFields: Consecutive
 AlignConsecutiveDeclarations: None
 AlignConsecutiveMacros: None
-AlignEscapedNewlines: DontAlign
+AlignEscapedNewlines: Left
 AlignOperands: Align
 AlignTrailingComments: true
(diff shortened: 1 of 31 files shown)


Sonicadvance1 (approved):

neobrain:
> in practice it may make other diffs worse since all the 's change when part of the macro change.

This is a good point. I'm actually inclined to give this a `-0` for that reason, but I ultimately care more that long code blocks for which that is a problem shouldn't need to be macros in the first place.

Sonicadvance1:
I fully support this since the spacing change is trivial. It's also a good enough reference that if anyone were to actually care about changing defines to something that isn't a define then they can walk through each individual formatting change in future PRs.

neobrain:
> I fully support this since the spacing change is trivial. It's also a good enough reference that if anyone were to actually care enough about changing defines to something that isn't a define then they can walk through each individual formatting change in future PRs.

Yes, to clarify my `-0` is only a formal note, not an objection to merging the change.

Sonicadvance1:
I'll formally click the merge button since it's trivial then.

PR #3649, opened 2024-05-21
Simplify/fix our validation passes

alyssarosenzweig:
Now that we don't have cross block liveness, we get big delete the code.


lioncash (approved):
you love to see it

Sonicadvance1 (approved):

PR #3652, opened 2024-05-22
InstructionCountCI: add bytemark hot block

alyssarosenzweig:
this was basically practice for using perf with FEX, but a few things do stand out in the assembly as suboptimal.

### DIFF
--- a/unittests/InstructionCountCI/FlagM/HotBlocks.json
+++ b/unittests/InstructionCountCI/FlagM/HotBlocks.json
@@ -224,6 +224,29 @@
         "subs x26, x11, x4",
         "cfinv"
       ]
+    },
+    "bytemark num sort": {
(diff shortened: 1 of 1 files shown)


alyssarosenzweig:
Ideally, this would be:

```
        "ldr x17, [x10, x13, lsl #3]",
        "orr x15, x13, #0x1",
        "ldr x20, [x10, x15, lsl #3]",
        "eor w27, w17, w20",
        "subs x26, x17, x20",
        "cfinv"
```

or even (with multiblock flag optimizations)

```
        "ldr x17, [x10, x13, lsl #3]",
        "orr x15, x13, #0x1",
        "ldr x20, [x10, x15, lsl #3]",
        "cmp x17, x20",
```

lioncash (approved):

PR #3660, opened 2024-05-24
Delete a big chunk of IR/Passes/*

alyssarosenzweig:
Grab bag of simplifying and pruning to the IR passes, aimed at improving compile-times and reducing complexity.


Sonicadvance1 (commented):
Looks good, I just want to do a testing pass over it.

Sonicadvance1 (approved):

PR #3661, opened 2024-05-25
Removes warnings

Sonicadvance1:
NFC

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
@@ -105,7 +105,7 @@ private:
 
   // Return the New node (if it exists) for an Old node, else the Old node.
   OrderedNode* Map(OrderedNode* Old) {
-    LOGMAN_THROW_AA_FMT(IsOld(Old), "Pre-condition");
+    LOGMAN_THROW_A_FMT(IsOld(Old), "Pre-condition");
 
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -611,8 +611,8 @@ bool ConstrainedRAPass::Run(IREmitter* IREmit_) {
         if (Class->Allocated & (1u << Reg.Reg)) {
           OrderedNode* Old = Class->RegToSSA[Reg.Reg];
 
-          LOGMAN_THROW_AA_FMT(IsOld(Old), "RegToSSA invariant");
-          LOGMAN_THROW_AA_FMT(IsOld(Node), "Haven't remapped this instruction");
+          LOGMAN_THROW_A_FMT(IsOld(Old), "RegToSSA invariant");
```
alyssarosenzweig, line 614: remind me the difference?

Sonicadvance1: AA_FMT will do a `__builtin_assume` in release builds. A_FMT is just an "assert format".
Since `IsOld` has a side-effect it throws a warning.

Sonicadvance1: AA = `Assume + Assert`
A = `Assert`

alyssarosenzweig (approved):

Issue #3662, opened 2024-05-25
Darwinia Linux bugs


Sonicadvance1:
Darwinia doing incorrect chdir with proc/self?

Seems to break the game from Lina's stream. Needs some poking.

tycho:
The texture "sprites/virii.bmp" does exist though?

(I agree the null pointer checks should probably be there, but at the same time, that should *never* fail if it's loading archives properly because there's no rational circumstance that the texture would be missing, so it'd just be a `DarwiniaReleaseAssert(!=nullptr)` anyway.)

Sonicadvance1:
> The texture "sprites/virii.bmp" does exist though?
> 
> (I agree the null pointer checks should probably be there, but at the same time, that should _never_ fail if it's loading archives properly because there's no rational circumstance that the texture would be missing, so it'd just be a `DarwiniaReleaseAssert(!=nullptr)` anyway.)

That's where the mystery lies. I assume it's checking some filesystem path first and then falling back to some memory mapped file (If I'm reading the assembly and the strace correctly).

```
191678 1644226 newfstatat(AT_FDCWD, "data/sprites/viriifull.bmp", 0xffffeefa95a0, 0) = -1 ENOENT (No such file or directory)
191679 1644226 newfstatat(AT_FDCWD, "data/sprites/virii.bmp", 0xffffeefa95a0, 0) = -1 ENOENT (No such file or directory)
191680 1644226 --- SIGSEGV {si_signo=SIGSEGV, si_code=SEGV_MAPERR, si_addr=NULL} ---
```

So then the question then comes down to _why_ is this `Resource::GetBinaryReader` returning nullptr under emulation.
I haven't yet dove in to the details of the assembly yet since it's doing some sort of map(vector?) lookup and it's relatively complex. Initial inspection looks like it always allocates some memory on the heap and only returns nullptr in a couple of edge-case conditions.

Hopefully not some mimalloc quirk or brk thing, but maybe FEX is somehow managing to mess up that lookup somehow.

tycho:
Yeah, it checks filesystem first, because anything under `<game dir>/data` takes precedence.

The archives and their contents are all decompressed and loaded into system memory at game startup, and the file blobs are stored in an `ankerl::unordered_dense::map<std::string, MemoryMappedFile *>` [1]. The key is a std::string of the path (e.g. `data/sprites/virii.bmp`) and the `MemoryMappedFile` is effectively just a pointer to some heap allocated memory and a length.

[...]

PR #3666, opened 2024-05-28
FileManagement: Fix fstatat/statx with self and NOFOLLOW

Sonicadvance1:
When asked to not follow the symlink, FEX needs to return data about the symlink itself rather than following to the target executable. In that case we need to return symlink information otherwise games that sanity check can break.

This is what happened with Darwinia in #3662.

We return the FEXInterpreter symlink information in this case since it doesn't return any information that is relevent to leaking emulator state. Once the application asks to follow through to the symlink target is when we will replace.

Also adds a unit test to ensure we don't break it.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -503,6 +503,24 @@ std::pair<int, const char*> FileManager::GetEmulatedFDPath(int dirfd, const char
   return std::make_pair(RootFSFD, &SubPath[1]);
 }
 
+///< Returns true if the pathname is self and symlink flags are set NOFOLLOW.
+bool FileManager::IsSelfNoFollow(const char* Pathname, int flags) const {
+  const bool Follow = (flags & AT_SYMLINK_NOFOLLOW) == 0;
+  if (Follow) {
+    // If we are following the self symlink then we don't care about this.
+    return false;
+  }
+
+  if (!Pathname) {
+    return false;
+  }
+
+  char PidSelfPath[50];
+  snprintf(PidSelfPath, sizeof(PidSelfPath), "/proc/%i/exe", CurrentPID);
+
+  return strcmp(Pathname, "/proc/self/exe") == 0 || strcmp(Pathname, "/proc/thread-self/exe") == 0 || strcmp(Pathname, PidSelfPath) == 0;
+}
+
 std::optional<std::string_view> FileManager::GetSelf(const char* Pathname) {
   if (SupportsProcFSInterpreter) {
     // FEX doesn't need to track procfs/exe if this is supported.
@@ -514,7 +532,7 @@ std::optional<std::string_view> FileManager::GetSelf(const char* Pathname) {
   }
 
   char PidSelfPath[50];
-  snprintf(PidSelfPath, 50, "/proc/%i/exe", CurrentPID);
+  snprintf(PidSelfPath, sizeof(PidSelfPath), "/proc/%i/exe", CurrentPID);
 
   if (strcmp(Pathname, "/proc/self/exe") == 0 || strcmp(Pathname, "/proc/thread-self/exe") == 0 || strcmp(Pathname, PidSelfPath) == 0) {
     return Filename();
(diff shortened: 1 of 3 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -503,6 +503,21 @@ std::pair<int, const char*> FileManager::GetEmulatedFDPath(int dirfd, const char
   return std::make_pair(RootFSFD, &SubPath[1]);
 }
 
+bool FileManager::IsSelf(const char* Pathname) const {
+  if (!Pathname) {
+    return false;
+  }
+
+  char PidSelfPath[50];
+  snprintf(PidSelfPath, 50, "/proc/%i/exe", CurrentPID);
```
alyssarosenzweig, line 512: s/50/sizeof(PidSelfPath)/

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -503,6 +503,21 @@ std::pair<int, const char*> FileManager::GetEmulatedFDPath(int dirfd, const char
   return std::make_pair(RootFSFD, &SubPath[1]);
 }
 
+bool FileManager::IsSelf(const char* Pathname) const {
+  if (!Pathname) {
```
alyssarosenzweig, line 507: Can Pathname actually be null..?

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -503,6 +503,21 @@ std::pair<int, const char*> FileManager::GetEmulatedFDPath(int dirfd, const char
...
+    return false;
+  }
+
+  char PidSelfPath[50];
+  snprintf(PidSelfPath, 50, "/proc/%i/exe", CurrentPID);
+
+  if (strcmp(Pathname, "/proc/self/exe") == 0 || strcmp(Pathname, "/proc/thread-self/exe") == 0 || strcmp(Pathname, PidSelfPath) == 0) {
```
alyssarosenzweig, line 514: Instead of

```
if (foo) { return true; }

return false;
```

simply `return foo`.

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -870,6 +892,13 @@ uint64_t FileManager::Statfs(const char* path, void* buf) {
 }
 
 uint64_t FileManager::NewFSStatAt(int dirfd, const char* pathname, struct stat* buf, int flag) {
+  const bool Follow = (flag & AT_SYMLINK_NOFOLLOW) == 0;
+  if (!Follow && IsSelf(pathname)) {
+    // If we aren't following the symlink for self then we need to return data about the symlink itself.
```
alyssarosenzweig, line 897: Drop the duplicated comments, or maybe replace with simply `// See Statx`

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -829,6 +844,13 @@ uint64_t FileManager::Openat2(int dirfs, const char* pathname, FEX::HLE::open_ho
 }
 
 uint64_t FileManager::Statx(int dirfd, const char* pathname, int flags, uint32_t mask, struct statx* statxbuf) {
+  const bool Follow = (flags & AT_SYMLINK_NOFOLLOW) == 0;
+  if (!Follow && IsSelf(pathname)) {
```
alyssarosenzweig, line 848: Would it be better for IsSelf to also take an `int flags` so we deduplicate the `NOFOLLOW` check?

alyssarosenzweig (changes requested):
Looks good, just would like some of the duplicated code tightened up first :) Nice debugging on the issue!

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
alyssarosenzweig: (Feel free to ignore this one if you think it's silly. It's a bit silly.)

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Yes, it is allowed for some syscalls (like statx) to pass a nullptr pathname, which then it uses the dirfd passed in.

alyssarosenzweig: Fun, thanks

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Did it.

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: Passed in the flags instead

alyssarosenzweig (approved):

PR #3669, opened 2024-05-29
ConstProp fixes for Darwinia

Sonicadvance1:
Alongside #3666 This will fix Darwinia's Linux executable.

Two bugs here.
ConstProp with AddShift was transposing Base and Offset registers, which means we were scaling the Base and not scaling Offset.

Additionally we were accidentally allowing GPR loadstore limits on Vector loadstores which is invalid.

Both of these combined were causing fault behaviour in Darwinia after #3666 was fixed.

### DIFF
--- a/FEXCore/Source/Interface/IR/IR.json
+++ b/FEXCore/Source/Interface/IR/IR.json
@@ -1001,7 +1001,8 @@
       },
       "GPR = AddShift OpSize:#Size, GPR:$Src1, GPR:$Src2, ShiftType:$Shift{ShiftType::LSL}, u8:$ShiftAmount{0}": {
         "Desc": [ "Integer Add with shifted register",
-                  "Will truncate to 64 or 32bits"
+                  "Will truncate to 64 or 32bits",
+                  "Dest = Src1 + (Src2 << ShiftAmount)"
                 ],
         "DestSize": "Size",
         "EmitValidation": [
--- a/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -105,9 +105,9 @@ static std::optional<MemExtendedAddrResult> MemExtendedAddressing(IREmitter* IRE
       auto Scale = 1U << AddShift->ShiftAmount;
       if (IsMemoryScale(Scale, AccessSize)) {
         // remove shift as it can be folded to the mem op
-        return MemExtendedAddrResult {MEM_OFFSET_SXTX, (uint8_t)Scale, IREmit->UnwrapNode(AddShift->Src2), IREmit->UnwrapNode(AddShift->Src1)};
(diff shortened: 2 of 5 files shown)


alyssarosenzweig:
Looks you fixed 2 bugs, but the unit test only tests for 1 one of them. Do we need a second unit test covering the other case?

alyssarosenzweig (approved):

Sonicadvance1:
> Looks you fixed 2 bugs, but the unit test only tests for 1 one of them. Do we need a second unit test covering the other case?

I'm not too terribly concerned about it, Seems like the confusion happened because we were returning a tuple and misunderstanding if the first or second OrderedNode* referred to base or offset. Now with a struct with named members it shouldn't occur, especially since I renamed the terrible `Arg0, Arg1` thing going on there.

alyssarosenzweig:
> > Looks you fixed 2 bugs, but the unit test only tests for 1 one of them. Do we need a second unit test covering the other case?
> 
> I'm not too terribly concerned about it, Seems like the confusion happened because we were returning a tuple and misunderstanding if the first or second OrderedNode* referred to base or offset. Now with a struct with named members it shouldn't occur, especially since I renamed the terrible `Arg0, Arg1` thing going on there.

Yes, but if i'm going to rewrite all of this logic anyway, I'd like a way to test it that's more reliable than "does Darwinia crash?" 😉

Sonicadvance1:
> > > Looks you fixed 2 bugs, but the unit test only tests for 1 one of them. Do we need a second unit test covering the other case?
> > 
> > 
> > I'm not too terribly concerned about it, Seems like the confusion happened because we were returning a tuple and misunderstanding if the first or second OrderedNode* referred to base or offset. Now with a struct with named members it shouldn't occur, especially since I renamed the terrible `Arg0, Arg1` thing going on there.
> 
> Yes, but if i'm going to rewrite all of this logic anyway, I'd like a way to test it that's more reliable than "does Darwinia crash?" 😉

I can do a unit test that tries to tell the difference between scaling on the wrong register once I've got a bit of time either later today or tomorrow

Sonicadvance1:
> > > > Looks you fixed 2 bugs, but the unit test only tests for 1 one of them. Do we need a second unit test covering the other case?
> > > 
> > > 
> > > I'm not too terribly concerned about it, Seems like the confusion happened because we were returning a tuple and misunderstanding if the first or second OrderedNode* referred to base or offset. Now with a struct with named members it shouldn't occur, especially since I renamed the terrible `Arg0, Arg1` thing going on there.
> > 
> > 
> > Yes, but if i'm going to rewrite all of this logic anyway, I'd like a way to test it that's more reliable than "does Darwinia crash?" 😉
> 
> I can do a unit test that tries to tell the difference between scaling on the wrong register once I've got a bit of time either later today or tomorrow

Added a unit test that tests the multiple different optimizations that FEX does in the const-prop for SIB scaling. Should be relatively encompassing, if not all.

alyssarosenzweig (approved):
Awesome, thanks a ton

PR #3673, opened 2024-05-29
Track tied sources in the IR

alyssarosenzweig:
Many FEX IR ops want their dests assigned to the same register as a particular source. Denote this in the IR.

Then, add a very simple heuristic in the RA to take advantage of this in typical cases. The "proper" RA handling will hopefully come soon(tm) in a follow up PR, that will guarantee coalescing in all cases to let us delete a chunk of JIT code. But this is a step towards that anyhow.

PR #3674, opened 2024-05-30
Optimize logical flags

alyssarosenzweig:
fuse in PF move. on top of #3673


Sonicadvance1 (approved):
Nice little optimization

PR #3678, opened 2024-05-31
Fix segfault when starting TestHarnessRunner with missing arguments

marysaka:
Found while trying to discover FEX codebase a bit.

Since https://github.com/FEX-Emu/FEX/pull/3390, launching `TestHarnessRunner` with missing arguments would cause a crash.


Sonicadvance1 (approved):
Nice catch!

PR #3680, opened 2024-05-31
Rewrite address mode handling

### LINKED ISSUE #3672: ConstProp is missing a bunch of AddShift optimizations for loadstore addresses
A bunch of the ConstProp pass with loadstore address generation is looking for OP_ADD instead of OP_ADD and OP_ADDSHIFT. This has made loadstores a bit less effectively in the case where semantics match.

Remembered this when writing #3669, and @alyssarosenzweig said they were going to look more in to it a while ago.

alyssarosenzweig:
Both x86 and arm64 support a rich set of addressing modes, broadly allowing instructions to encode modes of the form:

    base + (index << shift) + offset

There are restrictions on both ISAs, and naturally x86 is both more expressive and more nonsense. To get good performance, it's essential that we map x86 addressing modes to native arm64 equivalents.

Currently, FEX solves this problem in 2 parts:

* First, the OpcodeDispatcher spits out simple RISC instructions for the decoded x86 address mode (e.g.: LoadMem + Add + Constant) 
* Then ConstProp pattern matches the IR to fuse trees of instructions together into single arm64 loads

To be clear, there's nothing fundamentally *wrong* with this approach. It would allow fusing arm64 modes that don't have x86 equivalents by way of interinstruction optimization. In practice, this is unnecessary as the x86 modes are almost a superset of the arm64 modes.

This PR rewrites the handling in favour of a much simpler approach:

* OpcodeDispatcher generates fused arm64 loads directly based on the decoded x86 mode  

No pattern matching is performed.

This has a number of advantages:

* It is *much* simpler. Several hundred lines of code are removed.
* It is easier to reason about. The quirks of arm64 addressing are isolated to a single generative "select address mode" routine, instead of implicitly constraining a complex pattern match. This means we can handle arm64's funny corner cases (e.g. around atomic TSO emulation) simply. 
* It is theoretically faster. Generating IR instructions has a compile-time cost, and then optimizing them has a second compile-time cost. Generated optimized code in one shot (splatter-style) is more efficient. Although in practice, this is noise, nodejs start time is about the same across the series. 

One little "gotcha" is that OpcodeDispatcher *must* generate good code, it can no longer rely on the optimizer to clean up sloppy output. To avoid performance regressions, that requires reworking a bunch of ops that open-coded address arithmetic, poorly. All instructions tracked in instcountci are regression free except prefetch (which is nontrivial to rework and hard to get me to care right now). 

Demonstrating the capabilities, this PR has efficient handling for several x86 addressing modes missed by the old pattern matching (closes #3672), despite being a massive reduction in code. This massively speeds up several of our hot blocks.

bytemark scores before/after:

```
    NUMERIC SORT        :  Iterations/sec.:       1904.81  Index:  48.85
    STRING SORT         :  Iterations/sec.:       1413.04  Index: 631.39
    BITFIELD            :  Iterations/sec.:  473769108.23  Index:  81.27
    FP EMULATION        :  Iterations/sec.:        887.89  Index: 426.05
    FOURIER             :  Iterations/sec.:     101382.31  Index: 115.30
    ASSIGNMENT          :  Iterations/sec.:         89.00  Index: 338.66
    IDEA                :  Iterations/sec.:      11572.91  Index: 177.00
    HUFFMAN             :  Iterations/sec.:       4536.55  Index: 125.80
    NEURAL NET          :  Iterations/sec.:        148.59  Index: 238.69
    LU DECOMPOSITION    :  Iterations/sec.:       3142.32  Index: 162.79
    ==========================ORIGINAL BYTEMARK RESULTS==========================
    INTEGER INDEX       : 187.188
    FLOATING-POINT INDEX: 164.845

---

    NUMERIC SORT        :  Iterations/sec.:       2136.98  Index:  54.80
    STRING SORT         :  Iterations/sec.:       1476.12  Index: 659.57
    BITFIELD            :  Iterations/sec.:  483957639.05  Index:  83.02
    FP EMULATION        :  Iterations/sec.:        900.42  Index: 432.06
    FOURIER             :  Iterations/sec.:     102748.69  Index: 116.86
    ASSIGNMENT          :  Iterations/sec.:         91.67  Index: 348.84
    IDEA                :  Iterations/sec.:      11557.96  Index: 176.78
    HUFFMAN             :  Iterations/sec.:       4484.66  Index: 124.36
    NEURAL NET          :  Iterations/sec.:        145.18  Index: 233.23
    LU DECOMPOSITION    :  Iterations/sec.:       3241.80  Index: 167.94
    ==========================ORIGINAL BYTEMARK RESULTS==========================
    INTEGER INDEX       : 192.911
    FLOATING-POINT INDEX: 166.025
```

Numeric sort in particular is up **12%**, bringing that test case to **85%** of native on M1.


unittests/InstructionCountCI/SecondaryGroup.json
```diff
@@ -1640,19 +1639,19 @@
...
-        "add x21, x4, #0x18 (24)",
         "str w20, [x4, #24]",
         "mov w20, #0xffff",
-        "str w20, [x21, #4]",
-        "ubfx x20, x4, #0, #3",
-        "str x20, [x4, #512]"
+        "str w20, [x4, #28]",
+        "add x20, x4, #0x200 (512)",
+        "ubfx x21, x4, #0, #3",
+        "str x21, [x20]"
```
Sonicadvance1, line 1654: This is an interesting change. Should be able to still do `str x21, [x4, #512]`

alyssarosenzweig: Nice, fixed.

FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -940,6 +705,8 @@ void ConstProp::ConstantInlining(IREmitter* IREmit, const IRListView& CurrentIR)
 
       uint64_t Constant2 {};
       if (Op->OffsetType == MEM_OFFSET_SXTX && IREmit->IsValueConstant(Op->Offset, &Constant2)) {
+        Constant2 *= Op->OffsetScale;
```
Sonicadvance1, line 708: Is this correct? Wouldn't the Operation still have OffsetScale set to its scale and now the constant has been scaled? So if the backend starts caring about scale+imm, the logic is now disjoint and odd from the original representation.

alyssarosenzweig: Oh, yeah, good point - ConstProp should set OffsetScale = 1. Will fix.

alyssarosenzweig: rewrote this commit

Sonicadvance1 (commented):
Looking pretty good. I'll want to throw some testing at this one as well.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -1198,6 +1209,10 @@ class OpDispatchBuilder final : public IREmitter {
 
   OrderedNode* GetRelocatedPC(const FEXCore::X86Tables::DecodedOp& Op, int64_t Offset = 0);
 
+  struct AddressMode AddSegmentToAddress(struct AddressMode A, uint32_t Flags);
```
neobrain, line 1212: Struct types don't need to be prefixed with `struct ` in C++.

(I'm actually surprised clang-format doesn't format it away, since it really looks out of place.)

Sonicadvance1 (approved):
No issues detected, looks good to me.

PR #3681, opened 2024-06-01
optimize shld

alyssarosenzweig:
on top of #3680 for instcountci conflicts


Sonicadvance1 (approved):
Last two commits look good.

PR #3682, opened 2024-06-03
Find-and-replace OrderedNode* with Ref

alyssarosenzweig:
This is a step towards the new IR, where we don't have pointerful data structures and instead have flat handles only. Naming inspired by `ibc_ref`, it represents a reference to a source or destination (which are not explicitly stored).

### DIFF
--- a/FEXCore/Source/Interface/IR/IR.h
+++ b/FEXCore/Source/Interface/IR/IR.h
@@ -360,6 +360,17 @@ static_assert(std::is_trivially_copyable_v<OrderedNode>);
 static_assert(offsetof(OrderedNode, Header) == 0);
 static_assert(sizeof(OrderedNode) == (sizeof(OrderedNodeHeader) + sizeof(uint32_t)));
 
+// This is temporary. We are transitioning away from OrderedNode's in favour of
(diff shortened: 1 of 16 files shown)


FEXCore/Source/Interface/IR/IR.h
```diff
@@ -360,6 +360,17 @@ static_assert(std::is_trivially_copyable_v<OrderedNode>);
...
+// This is temporary. We are transitioning away from OrderedNode's in favour of
+// flat Ref words. To ease porting, we have this typedef. Eventually OrderedNode
+// will be removed and this typedef will be replaced by something like:
+//
+//  struct Ref {
+//     uint Flags : 1;
+//     uint ID : 23;
+//     uint Reg : 8;
+//  };
+typedef OrderedNode* Ref;
```
Sonicadvance1, line 372: `using Ref = OrderedNode*;`

alyssarosenzweig: done

Sonicadvance1 (approved):

pmatos:
I like this! :) Nice.

PR #3683, opened 2024-06-03
SMCTracking: Fix incorrect mprotect tracking

### LINKED ISSUE #3675: Sifu and RDR break FEX's SMC tracking logic
Both Sifu and RDR break FEX's SMC tracking logic due to the ordering it does mprotects in. Nailed down where it is happening, now I just need to fix it.

small unittest which reproduces without running the full game.
```cpp
#include <cstdint>
#include <sys/mman.h>
#include <unistd.h>
#include <sys/syscall.h>

int main() {

  // Original:
  // Base: 0x6ffffc0b8000
  // PtrHigh: 0x6ffffc0be000 (+0x6000)
  // PtrLow: 0x6ffffc0bd000 (+0x5000)
  const uint64_t PtrBase = 0x10'0000;
  const uint64_t PtrLow = PtrBase + 0x1000;
  const uint64_t PtrLowSize = 0x2000;
  const uint64_t PtrHigh = PtrBase + 0x2000;
  const uint64_t PtrHighSize = 0x1000;
  const uint64_t TotalSize = PtrHigh + 0x1000 - PtrBase;

  (void)mmap((void*)PtrBase, TotalSize, PROT_NONE, MAP_PRIVATE | MAP_ANONYMOUS | MAP_FIXED, -1, 0);

  // Order matters.
  mprotect((void*)PtrHigh, PtrHighSize, PROT_READ);
  mprotect((void*)PtrLow, PtrLowSize, PROT_READ);

  ::syscall(SYS_exit_group, 0);
  return 0;
}
```

Sonicadvance1:
Fixes #3675

This was the first time I've ever actually dived in to this code and this function melted my brain a bit while reading it. It was trying to be too smart in tracking VMA splits, but if it split right at the end of a VMA range it would then add a new range at the end where one already existed. This then caused us to have overlapping VMA ranges and it would completely break our SMC tracking since all tracking in the map must not overlap.

Instead of being too smart, just break it down in to 4 merge strategies, three of which are one shot. This is significantly easier to reason about and each strategy is mostly self-contained. The fourth strategy in the list is the most complex since it requires multiple steps since it needs to walk multiple VMAs.

To help test this I added some sanity checking code that proved invaluable to ensure everything was correct. That's not getting merged since the overhead is too much to run, but good to have available. Diff for this is at
  https://gist.github.com/Sonicadvance1/1ee60101ed0742b971a476fadbb51083

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsVMATracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsVMATracking.cpp
@@ -263,6 +263,11 @@ void SyscallHandler::VMATracking::ChangeUnsafe(uintptr_t Base, uintptr_t Length,
     MappingIter--;
 
     auto Current = &MappingIter->second;
+
+    if (Current->Base <= Base || Current->Base + Current->Length < Top) {
+      break;
+    }
+
     const auto CurrentBase = Current->Base;
     const auto CurrentTop = CurrentBase + Current->Length;
     const auto CurrentFlags = Current->Flags;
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Tested a few dozen games so far and haven't found any regressions yet. So I think this is okay.

neobrain:
Nice work finding that bug and clarifying the implementation!

One striking difference between merge strategy 4 and all the others is the iteration behavior. 4 will continue iterating, whereas 1-3 are loop-terminating events. It's currently a bit unfortunate that strategy 4 comes last, since it's the first one to be applied in general.

We can reshuffle the control flow a bit to make this more logical:

```cpp
auto MappingIter = VMAs.lower_bound(Top);

// "Strategy 4": Process all mappings partially covering the input range
while (MappingIter != VMAs.begin()) {
    MappingIter--;

    auto Current = &MappingIter->second;
    if (Current->Base <= Base || Current->Base + Current->Length < Top) {
        break;
    }

    // Previous logic for Strategy 4 goes here
}

// Process final mapping covering the start (if any)
if (CurrentTop <= Base) {
    // Nothing to do
} else {
    // Strategies 1-3
}
```

As a side note, it seems advisable to refactor the code such that it can be unit-tested. Extensive manual testing and careful code review is good, but it seems like a less effective use of our time than having a reliable set of tests.

Sonicadvance1:
> It's currently a bit unfortunate that strategy 4 comes last, since it's the first one to be applied in general.

In general strategy 1 and 2 is hit from testing actually. Then 3 for what I assume is linear arena allocations with protections. The final strategy is actually fairly rare, which is why it took so long for us to find things that broke us.

neobrain:
> In general strategy 1 and 2 is hit from testing actually. Then 3 for what I assume is linear arena allocations with protections. The final strategy is actually fairly rare, which is why it took so long for us to find things that broke us.

What I mean is that when it runs, it's the first one to be applied. This is not apparent from the current ordering in the code, which makes understanding the general control flow more difficult.

It's good to know it's fairly rare though. If we add a comment saying strategy 4 is rare, my change as proposed will also make it clear that the common case won't run more than one of the strategies at all.

Sonicadvance1:
I don't feel comfortable significantly changing the logic further since it already took me this long to write this change. Since I didn't write this SMC tracking code I don't have the whole thing in my head.

neobrain:
> Since this entire interface is opaque, it would require significant restructuring to punch that data through to unit tests today.

Does it require that? The changed code is already part of a dedicated class, and the function-to-be-tested doesn't access any state other than the `VMAs` vector. In the worst case we could split off most of `VMATracking::ChangeUnsafe` into a dedicated, standalone function.

Sonicadvance1:
> > Since this entire interface is opaque, it would require significant restructuring to punch that data through to unit tests today.
> 
> Does it require that? The changed code is already part of a dedicated class, and the function-to-be-tested doesn't access any state other than the `VMAs` vector. In the worst case we could split off most of `VMATracking::ChangeUnsafe` into a dedicated, standalone function.

It isn't a clean interface since it is hidden behind the syscalls interface entirely. So it would require making parts of that interface public to expose its data. It's going to take time I don't have to make it presentable enough to pass code review.

alyssarosenzweig (approved):
Let's unblock the release, we can always improve this later

neobrain:
> Let's unblock the release, we can always improve this later

Ryan and I discussed this PR in the last meet. The lack of unit tests isn't necessarily a blocker, but I'm not happy to merge a change that we don't trust enough to execute an almost trivial refactor on.

alyssarosenzweig:
> > Let's unblock the release, we can always improve this later
> 
> Ryan and I discussed this PR in the last meet. The lack of unit tests isn't necessarily a blocker, but I'm not happy to merge a change that we don't trust enough to execute an almost trivial refactor on.
> 
> It's also worth noting that this PR needn't block a new FEX release. #3675 is not a regression caused by FEX itself, so a new FEX release without this fix will still be a strict improvement over the May release.

Ok, I missed that #3675 wasn't a regression. In that case @Sonicadvance1 maybe we should release without this for now and target this change for FEX-2406?

Sonicadvance1:
> Nice work finding that bug and clarifying the implementation!
> 
> One striking difference between merge strategy 4 and all the others is the iteration behavior. 4 will continue iterating, whereas 1-3 are loop-terminating events. It's currently a bit unfortunate that strategy 4 comes last, when actually it's the first one to be applied in general.
> 
> We can reshuffle the control flow a bit to make this more logical:
> 
> ```c++
> auto MappingIter = VMAs.lower_bound(Top);
> 
> // "Strategy 4": Process all mappings partially covering the input range
> while (MappingIter != VMAs.begin()) {
>     MappingIter--;
> 
>     auto Current = &MappingIter->second;
>     if (Current->Base <= Base || Current->Base + Current->Length < Top) {
>         break;
>     }
> 
>     // Previous logic for Strategy 4 goes here
> }
> 
> // Process final mapping covering the start (if any)
> if (CurrentTop <= Base) {
>     // Nothing to do
> } else {
>     // Strategies 1-3
> }
> ```
> 
> As a side note, it seems advisable to refactor the code such that it can be unit-tested. Extensive manual testing and careful code review is good, but it seems like a less effective use of our time than having a reliable set of tests.

Changed the logic to match what was desired here.

PR #3688, opened 2024-06-10
unittests/bextr: add SrcSize tests

catfella:
dougallj mentioned that adding these tests might expose a bug in bextr. Since bextr implementation was changed apparently it now works correctly, that's good.


Sonicadvance1 (approved):
Nice. Thanks for adding a test for this!

PR #3689, opened 2024-06-10
Scripts/InstallFEX: update PPA URL

catfella:
After running the script my Ubuntu 24.04 lists https://ppa.launchpadcontent.net/fex-emu/fex/ubuntu in the output of `apt-cache policy` (which means GetPPAStatus would always fail).

Now, maybe older Ubuntus use http://ppa.launchpad.net/fex-emu/fex/ubuntu? If not then I guess the fix is just this commit, otherwise some sort of `or` with strings would be in order.

### DIFF
--- a/Scripts/InstallFEX.py
+++ b/Scripts/InstallFEX.py
@@ -3,6 +3,7 @@ import os
 import subprocess
 import sys
 import tempfile
+import re
 
 _Arch = None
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Very cool. I'm fine with this solution.

PR #3692, opened 2024-06-13
Fixes AFP.NEP handling on scalar insertions

### LINKED ISSUE #3690: asm_tests failing on HDK8650
Currently, and since 2e40da3d6b448598f6d86033b33f2a630357520b , the following asm_tests are failing:

```
The following tests FAILED:
    4977 - jit_1/Test_64Bit_REP/F3_C2.asm (Failed)
    4978 - jit_500/Test_64Bit_REP/F3_C2.asm (Failed)
    4979 - jit_500_m/Test_64Bit_REP/F3_C2.asm (Failed)
    5052 - jit_1/Test_64Bit_REPNE/F2_C2.asm (Failed)
    5053 - jit_500/Test_64Bit_REPNE/F2_C2.asm (Failed)
    5054 - jit_500_m/Test_64Bit_REPNE/F2_C2.asm (Failed)
    5394 - jit_1/Test_64Bit_TwoByte/0F_2A.asm (Failed)
    5395 - jit_500/Test_64Bit_TwoByte/0F_2A.asm (Failed)
    5396 - jit_500_m/Test_64Bit_TwoByte/0F_2A.asm (Failed)
```

This seems to be related to enabling AFP and RPRES in 2e40da3d6b448598f6d86033b33f2a630357520b.

pmatos:
Fixes #3690

### DIFF
--- a/unittests/InstructionCountCI/AFP/Secondary_REP.json
+++ b/unittests/InstructionCountCI/AFP/Secondary_REP.json
@@ -47,7 +47,7 @@
       ]
     },
     "rsqrtss xmm0, xmm1": {
-      "ExpectedInstructionCount": 3,
+      "ExpectedInstructionCount": 4,
       "Comment": [
         "FEAT_FPRES could make this more optimal",
         "0xf3 0x0f 0x52"
@@ -55,7 +55,8 @@
       "ExpectedArm64ASM": [
         "fmov s0, #0x70 (1.0000)",
         "fsqrt s1, s17",
-        "fdiv s16, s0, s1"
+        "fdiv s0, s0, s1",
+        "mov v16.s[0], v0.s[0]"
       ]
     },
     "rcpss xmm0, xmm1": {
@@ -144,7 +145,7 @@
         "0xf3 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq s16, s16, s17"
+        "fcmeq s16, s17, s16"
       ]
     },
     "cmpss xmm0, xmm1, 1": {
@@ -184,7 +185,7 @@
         "0xf3 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq s0, s16, s17",
+        "fcmeq s0, s17, s16",
         "mvn v0.8b, v0.8b",
         "mov v16.s[0], v0.s[0]"
       ]
--- a/unittests/InstructionCountCI/AFP/Secondary_REPNE.json
+++ b/unittests/InstructionCountCI/AFP/Secondary_REPNE.json
@@ -136,7 +136,7 @@
         "0xf2 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq d16, d16, d17"
+        "fcmeq d16, d17, d16"
       ]
     },
     "cmpsd xmm0, xmm1, 1": {
@@ -176,7 +176,7 @@
         "0xf2 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq d0, d16, d17",
+        "fcmeq d0, d17, d16",
         "mvn v0.8b, v0.8b",
         "mov v16.d[0], v0.d[0]"
       ]
--- a/unittests/InstructionCountCI/AFP/VEX_map1.json
+++ b/unittests/InstructionCountCI/AFP/VEX_map1.json
@@ -30,7 +30,7 @@
       ]
     },
     "vrsqrtss xmm0, xmm1, xmm2": {
-      "ExpectedInstructionCount": 4,
+      "ExpectedInstructionCount": 5,
       "Comment": [
         "FEAT_FPRES could make this more optimal",
         "Map 1 0b10 0x52 128-bit"
@@ -39,7 +39,8 @@
         "mov v16.16b, v17.16b",
         "fmov s0, #0x70 (1.0000)",
         "fsqrt s1, s18",
-        "fdiv s16, s0, s1"
+        "fdiv s0, s0, s1",
+        "mov v16.s[0], v0.s[0]"
       ]
     },
     "vrcpss xmm0, xmm1, xmm2": {
@@ -61,7 +62,7 @@
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "fcmeq s16, s17, s18"
+        "fcmeq s16, s18, s17"
       ]
     },
     "vcmpss xmm0, xmm1, xmm2, 0x01": {
@@ -105,7 +106,7 @@
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "fcmeq s0, s17, s18",
+        "fcmeq s0, s18, s17",
         "mvn v0.8b, v0.8b",
         "mov v16.s[0], v0.s[0]"
       ]
@@ -154,7 +155,7 @@
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "fcmeq d16, d17, d18"
+        "fcmeq d16, d18, d17"
       ]
     },
     "vcmpsd xmm0, xmm1, xmm2, 0x01": {
@@ -198,7 +199,7 @@
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "fcmeq d0, d17, d18",
+        "fcmeq d0, d18, d17",
         "mvn v0.8b, v0.8b",
         "mov v16.d[0], v0.d[0]"
       ]
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/VectorOps.cpp
@@ -224,16 +224,18 @@ DEF_FBINOP_SCALAR_INSERT(VFSubScalarInsert, fsub)
 DEF_FBINOP_SCALAR_INSERT(VFMulScalarInsert, fmul)
 DEF_FBINOP_SCALAR_INSERT(VFDivScalarInsert, fdiv)
 
+
+// VFScalarOperation performs the operation described through ScalarEmit between Vector1 and Vector2,
+// storing it into Dst. This is a scalar operation, so the only lowest element of each vector is used for the operation.
+// The result is stored into the destination. The untouched bits of the destination come from Vector1, unless it's a 256 vector
+// and ZeroUpperBits is true, in which case the upper bits are zero.
 void Arm64JITCore::VFScalarOperation(uint8_t OpSize, uint8_t ElementSize, bool ZeroUpperBits, ScalarBinaryOpCaller ScalarEmit,
                                      ARMEmitter::VRegister Dst, ARMEmitter::VRegister Vector1, ARMEmitter::VRegister Vector2) {
   const auto Is256Bit = OpSize == Core::CPUState::XMM_AVX_REG_SIZE;
-  if (!Is256Bit) {
-    LOGMAN_THROW_A_FMT(ZeroUpperBits == false, "128-bit operation doesn't support ZeroUpperBits in {}", __func__);
-  }
+  LOGMAN_THROW_A_FMT(Is256Bit || !ZeroUpperBits, "128-bit operation doesn't support ZeroUpperBits in {}", __func__);
 
   // Bit of a tricky detail.
-  // The upper bits of the destination comes from the first source.
-
+  // The upper bits of the destination comes from Vector1.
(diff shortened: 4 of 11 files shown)


pmatos:
This is **not** ready to merge - still failing `0F_2A.asm` asm test.

pmatos:
Interesting - CI is failing instcountci - but it's passing locally! :-/

pmatos:
> Interesting - CI is failing instcountci - but it's passing locally! :-/

I wonder if this is because I am running instcountci with HW with AFP support? @Sonicadvance1 ?
I actually expected instcountci to be HW indep.

pmatos:
It might be that we are assuming AFP is on and turning it on as needed... Shall we add explicit disable AFP when we expect AFP to be off?

pmatos:
> It might be that we are assuming AFP is on and turning it on as needed... Shall we add explicit disable AFP when we expect AFP to be off?

Yep - confirmed that this was the issue.

Anyway, there's some pessimization ongoing that I still need to understand.

pmatos:
I think I need a few extras pairs of eyes here.

There are still some pessimizations but these are fixes for handling AFP. We had assumed that upper bits were comming from first arg for most ops but that's not the case. For example, for fcm<cmp> it comes from last arg. For frsqrte, it comes from destination. But for fdiv, it comes from first source as initially expected. Since for most (all?) intel sse scalar inserts, top bits come from first source, if arm64 differs then we pessimize.

I am open to suggestions and code review. First time looking at this code and this part of the arm spec so maybe something can be improved codewise.

unittests/InstructionCountCI/AFP/Secondary_REP.json
```diff
@@ -139,30 +139,33 @@
...
     },
     "cmpss xmm0, xmm1, 0": {
-      "ExpectedInstructionCount": 1,
+      "ExpectedInstructionCount": 2,
       "Comment": [
         "0xf3 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq s16, s16, s17"
+        "fcmeq s0, s16, s17",
```
Sonicadvance1, line 147: I'm confused with the changes to fcmeq, fcmgt, fcmge here. According to the ARM manual's pseudocode, the scalar version of these instructions should properly merge when NEP is enabled.
```
boolean merge = elements == 1 && IsMerging(FPCR); ///< IsMerging checking FPCR.NEP
bits(128) result = if merge then V[m, 128] else Zeros(128);
```

It just uses vm as the merge register, so we need to be careful.

- cmpss - EQ - swap the arguments
- cmpss - LT - swap arguments and convert to fcmge?
- cmpss - LE - swap arguments and convert to fcmgt?

unittests/InstructionCountCI/AFP/Secondary_REPNE.json
```diff
@@ -131,30 +131,33 @@
...
     },
     "cmpsd xmm0, xmm1, 0": {
-      "ExpectedInstructionCount": 1,
+      "ExpectedInstructionCount": 2,
       "Comment": [
         "0xf2 0x0f 0xc2"
       ],
       "ExpectedArm64ASM": [
-        "fcmeq d16, d16, d17"
+        "fcmeq d0, d16, d17",
```
Sonicadvance1, line 139: Same as the 32-bit version, same concerns for the 64-bit versions.

unittests/InstructionCountCI/AFP/VEX_map1.json
```diff
@@ -10,27 +10,29 @@
...
     "vsqrtss xmm0, xmm1, xmm2": {
-      "ExpectedInstructionCount": 2,
+      "ExpectedInstructionCount": 3,
       "Comment": [
         "Map 1 0b10 0x51 128-bit"
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "fsqrt s16, s18"
+        "fsqrt s0, s18",
```
Sonicadvance1, line 19: Also confused by this change. Pseudocode of fsqrt
```
boolean merge = IsMerging(FPCR);
bits(128) result = if merge then V[d, 128] else 0<127:0>;
```
We already move xmm1 in to xmm0 at the start, so then `fsqrt s16 (xmm0), s18 (xmm2)` should insert correctly here?

Same situation for the 64-bit version

unittests/InstructionCountCI/AFP/VEX_map1.json
```diff
@@ -39,7 +41,8 @@
         "mov v16.16b, v17.16b",
         "fmov s0, #0x70 (1.0000)",
         "fsqrt s1, s18",
-        "fdiv s16, s0, s1"
+        "fdiv s0, s0, s1",
```
Sonicadvance1, line 44: This one looks like it was a legitimate problem.
Pseudocode:
```
boolean merge = IsMerging(FPCR);
bits(128) result = if merge then V[n, 128] else Zeros(128);
```
Since vn here is s0 which contains 1.0, it wasn't doing correct insert mechanics. AFP+RPRES would be necessary to get correct semantics.
:+1:

unittests/InstructionCountCI/AFP/VEX_map1.json
```diff
@@ -241,43 +250,47 @@
...
     "vcvtsi2ss xmm0, xmm1, eax": {
-      "ExpectedInstructionCount": 2,
+      "ExpectedInstructionCount": 3,
       "Comment": [
         "Map 1 0b10 0x2A 128-bit"
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "scvtf s16, w4"
+        "scvtf s0, w4",
```
Sonicadvance1, line 259: Similar to `vsqrtss` from before, this looks like this should have worked with the previous version.
We first move xmm1 in to xmm0 at the start, then the `scvtf s16 (xmm0), w4 (eax)` should work because pseudocode:
```
constant boolean merge = IsMerging(FPCR);
fltval = if merge then V[d, fltsize] else Zeros(fltsize);
```
Using vd as the destination as the merge register (It's really confusing that instructions choose between vd, vn, and vm depending).

Same thought process for all the following vcvt* instructions after this.

unittests/InstructionCountCI/RPRES/VEX_map1_AFP.json
```diff
@@ -29,14 +29,15 @@
...
-      "ExpectedInstructionCount": 2,
+      "ExpectedInstructionCount": 3,
       "Comment": [
         "AFP can make this more optimal",
         "Map 1 0b10 0x52 128-bit"
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
-        "frsqrte s16, s18"
+        "frsqrte s0, s18",
```
Sonicadvance1, line 39: Following from the previous comment about `vrsqrtss` with only AFP and not rpres. Theoretically the original version of this should have correct semantics.
We are first moving xmm1 in to xmm0
We then `frsqrte s16 (xmm0), s18 (xmm2)` and following the pseudocode:
```
boolean merge = elements == 1 && IsMerging(FPCR);
bits(128) result = if merge then V[d, 128] else Zeros(128);
```
vd is used as the merge source, which should follow semantics correctly.
Same thing about the `vrcpss` down below.

unittests/InstructionCountCI/AFP/Secondary_REP.json
pmatos: You're right - we can do that. I was so focused into the details of where the top bits were coming from that it didn't occur to me that swapping the arguments would allow us to continue generating the same code we had before. 🤦🏼 

Thanks for the feedback on the patch.

pmatos: @Sonicadvance1 For cmpss, swapping seems to work but I get problems with cmpsd as expected. I will see if there's anything that can be done in that case.

Sonicadvance1: Curious, the pseudocode for fcmeq doesn't claim there is a behaviour change between single and double precision scalar.

pmatos: The comment for `IsMerging` in the spec sounds like it's in contradiction to what the NEP bit part says. The IsMerging says:
```
// IsMerging()
// ===========
// Returns TRUE if the output elements other than the lowest are taken from
// the destination register.
```

but the NEP.bit part actually has a bunch of expections where the output elements other than the lowest bit are taken from different places.

Screenshot from 2024-06-12 16-45-09

Sonicadvance1: And the pseudocode for the instruction which shows expected behaviour.
`elements = 1` for the scalar variants.
Screenshot 2024-06-17 at 02 12 58

pmatos: That's odd - but there are couple of other odd things that I still need to wrap my head around. For example, 
1. I can't see a different in execution path between cmpss and cmpsd. 
2. We are already swapping argumnets by converting cmp-LT into fcmgt and cmp-LE into fcmge. In my head, this cannot work. `LT a1, a2 ==> GE a2, a1` and `LE a1, a2 ==> GT a2, a1`. I am probably missing something but need to understand what.

Sonicadvance1: Actually, this transformation might be because of this confusing table.
Image_2024-06-17_03-09-02
So that sounds like you could have found a legitimate bug in our implementation.

pmatos: OK - I initially had some misunderstandings about the code but I got it now so I removed comment. :)

unittests/InstructionCountCI/AFP/VEX_map1.json
pmatos: humm, I might have been too eager to get the proper upper bits. Also I misunderstood the spec to what happens without AFP.

pmatos: I am preparing a simpler patch for this.

unittests/InstructionCountCI/AFP/Secondary_REP.json
pmatos: > Actually, this transformation might be because of this confusing table. Image_2024-06-17_03-09-02 So that sounds like you could have found a legitimate bug in our implementation.

The bug in fcmeq was what triggered the test failures once AFP was enabled. I assume you're referring to that? But just changing the order of the ops, fixes it.

pmatos:
@Sonicadvance1 let me know what you think.

unittests/InstructionCountCI/AFP/VEX_map1.json
```diff
@@ -30,13 +30,14 @@
...
     "vrsqrtss xmm0, xmm1, xmm2": {
-      "ExpectedInstructionCount": 4,
+      "ExpectedInstructionCount": 5,
       "Comment": [
         "FEAT_FPRES could make this more optimal",
         "Map 1 0b10 0x52 128-bit"
       ],
       "ExpectedArm64ASM": [
         "mov v16.16b, v17.16b",
+        "mov v0.16b, v16.16b",
```
Sonicadvance1, line 40: This mov looks extraneous. The `fmov s0, #0x70 (1.0000)` afterwords will always zero the full register even with AFP.NEP. Same situations for the other locations that add this move.

Sonicadvance1:
Other than the one comment now, looking pretty good! Nice work fixing these AFP.NEP bugs.

unittests/InstructionCountCI/AFP/VEX_map1.json
pmatos: This specific case looks like:
```
   "vrsqrtss xmm0, xmm1, xmm2": {
      "ExpectedInstructionCount": 5,
      "Comment": [
        "FEAT_FPRES could make this more optimal",
        "Map 1 0b10 0x52 128-bit"
      ],
      "ExpectedArm64ASM": [
        "mov v16.16b, v17.16b",
        "mov x0, #0x3f800000",
        "mov v16.s[0], w0",
        "fsqrt s1, s18",
        "fdiv s16, s16, s1"
      ]
    },
```

By inserting 0x3f80000 (1.0f) directly into the bottom of v16, we can then use that as the destination and first source of fdiv. WDYT?

pmatos: This looks ok:

```
        "mov v16.16b, v17.16b",
        "fmov s0, #0x70 (1.0000)",
        "fsqrt s1, s18",
        "fdiv s0, s0, s1",
        "mov v16.s[0], v0.s[0]"
```

Sonicadvance1: Yea, that matches expectations for my mental model.

Sonicadvance1 (approved):
Nice ncie ncei nice nice.

PR #3694, opened 2024-06-13
FEX: Consolidate JSON allocators and fix 3691

### LINKED ISSUE #3691: Unhandled exception when starting any FEX* command
After error during `FEXRootFSFetcher`'s "unpacking fs image" phase
```
Do you wish to extract the erofs file or use it as-is?
Options:
	0: Cancel
	1: Extract
	2: As-Is
	
Response {1-2} or 0 to cancel
1
Extracting Erofs. This might take a few minutes.
<E> erofs: I/O error occurred when verifying data chunk @ nid 31192463
Do you wish to set this RootFS as default?
Response {y,yes,1} or {n,no,0}
y
Ubuntu_23_04.ero set as default RootFS
```

 any Fex* command return this error:
```
> FEXRootFSFetcher 
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
fish: Job 1, 'FEXRootFSFetcher' terminated by signal SIGABRT (Abort)
```
FEXConfig spawns window, but then fail with same error
```
> FEXConfig 
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
fish: Job 1, 'FEXConfig' terminated by signal SIGABRT (Abort)
```

#### Build
```sh
mkdir build
cd build
env CC=clang CXX=clang++ cmake ..  -G Ninja -D ENABLE_LTO=False -D BUILD_TESTS=False -D CMAKE_BUILD_TYPE=Release -D INTERPROCEDURAL_OPTIMIZATION=False

sudo ninja install
```
#### Run

FEXRootFSFetcher
Select Ubuntu 23 erofs image

### Addition info

os-release
```
NAME="Manjaro ARM"
ID="manjaro-arm"
```

lscpu
```
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 6
  On-line CPU(s) list:  0-5
Vendor ID:              ARM
  Model name:           Cortex-A53
    Model:              4
    Thread(s) per core: 1
    Core(s) per socket: 4
    Socket(s):          1
    Stepping:           r0p4
    CPU(s) scaling MHz: 58%
    CPU max MHz:        1416.0000
    CPU min MHz:        408.0000
    BogoMIPS:           48.00
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
  Model name:           Cortex-A72
    Model:              2
    Thread(s) per core: 1
    Core(s) per socket: 2
    Socket(s):          1
    Stepping:           r0p2
    CPU(s) scaling MHz: 100%
    CPU max MHz:        1800.0000
    CPU min MHz:        408.0000
    BogoMIPS:           48.00
    Flags:              fp asimd evtstrm aes pmull sha1 sha2 crc32 cpuid
Vulnerabilities:        
  Gather data sampling: Not affected
  Itlb multihit:        Not affected
  L1tf:                 Not affected
  Mds:                  Not affected
  Meltdown:             Not affected
  Mmio stale data:      Not affected
  Retbleed:             Not affected
  Spec rstack overflow: Not affected
  Spec store bypass:    Vulnerable
  Spectre v1:           Mitigation; __user pointer sanitization
  Spectre v2:           Vulnerable
  Srbds:                Not affected
  Tsx async abort:      Not affected
```

Sonicadvance1:
Fixes #3691

We weren't checking if the file was empty before using its `at` function member. This was causing an early crash if the config file existed but was empty.

Consolidates the three locations that copy and pasted the json allocator tools and adds an empty check for all of them.

Also adds two missing checks to the ThunksDB handler that could have resulted in the same crash if ThunksDB was an empty file.

### DIFF
--- a/Source/Common/JSONPool.h
+++ b/Source/Common/JSONPool.h
@@ -1,0 +1,25 @@
+// SPDX-License-Identifier: MIT
+
+#include <FEXCore/fextl/memory.h>
+#include <FEXCore/fextl/list.h>
+
+#include <tiny-json.h>
+
+namespace FEX::JSON {
+struct JsonAllocator {
+  jsonPool_t PoolObject;
+  fextl::list<json_t> json_objects;
+
+  JsonAllocator();
+};
+static_assert(offsetof(JsonAllocator, PoolObject) == 0, "This needs to be at offset zero");
+
+template<typename T>
+const json_t* CreateJSON(T& Container, JsonAllocator& Allocator) {
+  if (Container.empty()) {
+    return nullptr;
+  }
+
+  return json_createWithPool(&Container.at(0), &Allocator.PoolObject);
+}
+} // namespace FEX::JSON
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -8,6 +8,7 @@ $end_info$
 
 #include "Common/Config.h"
 #include "Common/FDUtils.h"
+#include "Common/JSONPool.h"
 
 #include "FEXCore/Config/Config.h"
 #include "LinuxSyscalls/FileManagement.h"
@@ -42,25 +43,6 @@ $end_info$
 
 #include <tiny-json.h>
 
-namespace JSON {
-struct JsonAllocator {
-  jsonPool_t PoolObject;
-  fextl::unique_ptr<fextl::list<json_t>> json_objects;
-};
-static_assert(offsetof(JsonAllocator, PoolObject) == 0, "This needs to be at offset zero");
-
-json_t* PoolInit(jsonPool_t* Pool) {
-  JsonAllocator* alloc = reinterpret_cast<JsonAllocator*>(Pool);
-  alloc->json_objects = fextl::make_unique<fextl::list<json_t>>();
-  return &*alloc->json_objects->emplace(alloc->json_objects->end());
-}
-
-json_t* PoolAlloc(jsonPool_t* Pool) {
-  JsonAllocator* alloc = reinterpret_cast<JsonAllocator*>(Pool);
-  return &*alloc->json_objects->emplace(alloc->json_objects->end());
-}
-} // namespace JSON
-
 namespace FEX::HLE {
 bool FileManager::RootFSPathExists(const char* Filepath) {
   LOGMAN_THROW_A_FMT(Filepath && Filepath[0] == '/', "Filepath needs to be absolute");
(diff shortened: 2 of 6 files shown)


Source/Common/JSONPool.h
```diff
@@ -0,0 +1,36 @@
...
+
+#include <FEXCore/fextl/memory.h>
+#include <FEXCore/fextl/list.h>
+
+#include <tiny-json.h>
+
+namespace FEX::JSON {
+struct JsonAllocator {
+  jsonPool_t PoolObject;
+  fextl::unique_ptr<fextl::list<json_t>> json_objects;
```
neobrain, line 11: Unrelated to this PR, but why does this wrap `fextl::list` in a `unique_ptr`?

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -258,16 +237,14 @@ FileManager::FileManager(FEXCore::Context::Context* ctx)
...
-          {
-            .init = JSON::PoolInit,
-            .alloc = JSON::PoolAlloc,
-          },
-      };
+      FEX::JSON::JsonAllocator Pool = FEX::JSON::GetPool();
 
       // If a thunks DB property exists then we pull in data from the thunks database
-      const json_t* json = json_createWithPool(&FileData.at(0), &Pool.PoolObject);
+      const json_t* json = FEX::JSON::CreateJSON(FileData, Pool);
```
neobrain, line 243: Isn't another issue here that `FileData` won't be null-terminated? One of the other callees explicitly uses `push_back(0)` to add a terminator.

Instead, we can just change all `FileData`s to `fextl::string` instead (and drop the explicit `push_back`) and change `&FileData.at(0)` to `FileData.c_str()`. This will avoid the need for a wrapper function too, since the pointer [will always be valid](https://en.cppreference.com/w/cpp/string/basic_string/data).

Source/Common/JSONPool.h
```diff
@@ -0,0 +1,36 @@
...
+
+inline JsonAllocator GetPool() {
+  return FEX::JSON::JsonAllocator {
+    .PoolObject =
+      {
+        .init = FEX::JSON::PoolInit,
+        .alloc = FEX::JSON::PoolAlloc,
+      },
+  };
+}
```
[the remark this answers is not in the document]
Sonicadvance1: Sure. Moved it to the cpp file and added a default constructor instead.

Source/Common/JSONPool.h
Sonicadvance1: Removed the unique_ptr

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
Sonicadvance1: The json library doesn't care about null terminator, so I removed that extraneous push_back instead.

alyssarosenzweig (approved):
nice, I like deduplication :)

Source/Common/JSONPool.h
```diff
@@ -0,0 +1,28 @@
...
+struct JsonAllocator {
+  jsonPool_t PoolObject;
+  fextl::list<json_t> json_objects;
+
+  JsonAllocator();
+};
+static_assert(offsetof(JsonAllocator, PoolObject) == 0, "This needs to be at offset zero");
+
+json_t* PoolInit(jsonPool_t* Pool);
+json_t* PoolAlloc(jsonPool_t* Pool);
```
neobrain, line 18: These functions don't need to be declared in the header file.

neobrain (approved):
Looks good other than not needing public declarations for PoolInit/PoolAlloc.

Source/Common/JSONPool.h
alyssarosenzweig: Fixed, thanks

PR #3700, opened 2024-06-14
Externals: Update vixl submodule

lioncash:
Updates vixl to track the latest upstream changes that fix erroneous non-zeroing behavior for 256-bit vectors


lioncash:
Not sure how to address the formatting error, since I've never gotten an error like that before

Sonicadvance1 (approved):
Me either.

Sonicadvance1:
Let's see what that does.

PR #3701, opened 2024-06-14
FEXCore: Fixes Call with 32-bit displacement and address size override

Sonicadvance1:
FEX had a bug with this instruction where it was incorrectly using both the address size override and operand size override to truncate the immediate offset. This isn't how the instruction should behave as it should actually ignore the address size override.

This now puts it correctly inline with how the jump instruction works and adds a unit test to ensure it doesn't break again.

This fixes a crash from the Arch rootfs from the glibc dynamic linker being compiling in a way where a call instruction was getting aligned using this prefix (Since the compiler knew it does nothing).

As a side-note, FEX doesn't handle 16-bit displacement at all in both the jump and call instructions. This is because the operand size qualifier controls both the immediate size and the RIP size. This is because the override will truncate the resulting rip to 16-bit, which falls in to the 16-bit guard pages at the start of the process. Theoretically an application can abuse this with fault behaviour but wiring up unittests for this would be a nightmare. I'm opting to continue not supporting this behaviour.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -672,10 +672,13 @@ void OpDispatchBuilder::CALLOp(OpcodeArgs) {
   }
 
   auto ConstantPC = GetRelocatedPC(Op);
+  // Call instruction only uses up to 32-bit signed displacement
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):
nice catch!

PR #3703, opened 2024-06-15
Remove RCLSE

alyssarosenzweig:
RCLSE is one of our most complex compiler passes. It is responsible for removing redundant context load/stores, and by extension register & flag load/stores. Its problems are well known:

* It's based on the context model, which requires awkward translations for registers.
* It's a *lot* of code, and not especially efficient code. It has a lot of complexity for things like partial tracking, which just don't matter *that* much.
* It has no awareness of instructions that are invalid to optimize across, if we're trying to guarantee precise exception handling to the JIT.
* Its effect on codegen is somewhat random, helping some blocks while hurting others due to the laissez-faire approach to SRA in FEX.

In a step to addressing these problems, this PR replaces RCLSE with a simple register cache in the OpcodeDispatcher, reusing loads and deferring stores until the cache is explicitly flushed. We treat a few important context entries (like AbridgedFTW) as pseudo-registers to get RCLSE-like behaviour there too. This has some tradeoffs:

* It is a *lot* simpler, a *lot* less code, and a *lot* faster.
* It will never optimize across a FlushRegisterCache boundary. For precise exceptions later on, we can simply insert flushes around instructions that might fault and then we get correct behaviour throughout the compiler. Exception handling behaviour is now completely controlled by the OpcodeDispatcher instead of being implicitly sprinkled throughout FEXCore.
* If/when we introduce formal clauses into the IR, this is the first step to get there. FlushRegisterCache would then be a "change clause" signal in the dispatcher.
* Not all context load/stores are tracked here. This could be improved, I let instcountci be my guide for what matters.
* This gives SRA heartburn, especially around tied registers. This will be solved by the SRA rework that I have planned for clauses, but we need to ripoff the bandage first to get there because I don't want to port RCLSE to new IR data structures.

So... I think this will be better long term. It's a bit of a toss-up short term. JIT time is reduced **6%** across the series. instcountci is a net win, with some outliers in both directions. bytemark scores are regressed a few % unfortunately, but we should be able to recover that perf in due time.


unittests/InstructionCountCI/Secondary.json
```diff
@@ -868,11 +868,12 @@
       ]
     },
     "packsswb mm0, mm0": {
-      "ExpectedInstructionCount": 4,
+      "ExpectedInstructionCount": 5,
       "Comment": "0x0f 0x63",
       "ExpectedArm64ASM": [
         "ldr d2, [x28, #768]",
-        "zip1 v2.2d, v2.2d, v2.2d",
+        "ldr d3, [x28, #768]",
```
Sonicadvance1, line 875: Looks like redundant MMX registers aren't tracked currently. Should be an easy fix.

unittests/InstructionCountCI/FEXOpt/MultiInst_AFP.json
```diff
@@ -24,8 +24,10 @@
         "addss xmm0, xmm2"
       ],
       "ExpectedArm64ASM": [
-        "fadd s16, s16, s17",
-        "fadd s16, s16, s18"
+        "mov v2.16b, v16.16b",
```
Sonicadvance1, line 27: This one looks pretty bad. Looks like it will completely break scalar SSE operations. It is zeroing the upper bits insert of doing an insert. CI isn't hitting it since nothing in CI supports AFP yet.

unittests/InstructionCountCI/Secondary.json
alyssarosenzweig: Thanks, will look into that

unittests/InstructionCountCI/FEXOpt/MultiInst_AFP.json
alyssarosenzweig: Yes, that's unfortunately fallout from the tied register issue. I could try to frontload some RA work now to deal with it but it's very much nontrivial. Alternatively, I could maybe see how much of this PR can land without regression and then deal with the rebasing for the next few months... It's not critical things land asap, just trying to avoid ending up with a gigantic "rewrite the world" patch set if I can help it!

alyssarosenzweig: I've split out #3705 and #3704 as changes that can land now, then what's left here shouldn't be too bad to rebase for a while until I have RA in better shape. Will close this for now while I shave RA shaped yaks / IR reworks.

unittests/InstructionCountCI/Secondary.json
alyssarosenzweig: MMX tracking added locally.

PR #3704, opened 2024-06-16
RA: priorize remat over spilling

alyssarosenzweig:
No instcountci changes yet, since nothing currently spills in instcountci. This
    mitigates spilling later seen with #3703, and should help for certain
    pathological blocks even without those changes (maybe we should try to get some
    of those blocks in instcountci?).


Sonicadvance1 (approved):

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -265,37 +269,42 @@ class ConstrainedRAPass final : public RegisterAllocationPass {
     }
   };
 
+#define foreach_bit(b, x) for (uint32_t __x = (x), b; ((b) = __builtin_ffs(__x) - 1, __x); __x &= ~(1 << (b)))
```
Sonicadvance1, line 272: This should probably be commented about the magic bit-walking that this uses. I know what it is doing but the magic may be mysterious for anyone that isn't used to walking discontiguous bits in a mask.

I like it.

alyssarosenzweig: done.

Sonicadvance1 (approved):

Issue #3677, opened 2024-05-31
Begin migrating from fmt to c++20 format


Sonicadvance1:
libstdc++ and libc++ on all of our supported distros is now new enough to support <format>.

This should be mostly straightforward to implement, the only weird one will be fextl::fmt::vformat since it uses the fmt::basic_memory_buffer for custom allocator support. There's no direct 1:1 replacement in the C++ standard so will need some massaging.

This will remove one more external project dependency in a relatively straight forward way. Low-prio but shouldn't be terrible to do.

Sonicadvance1:
Looks like this would require upgrading to at least libstdc++ 13.1, which Ubuntu 22.04 doesn't ship.
Once we drop 22.04 it should be safe but until then it is likely too new still.

PR #3709, opened 2024-06-16
Switch tools depending on FEXCore to use fextl::fmt

marysaka:
~~This contribute a little bit to #3677 (I explicitly left `fextl::fmt` internals out this PR for now)~~
Ubuntu 22.04 currently doesn't ship a version of libstdc++ new enough to support `std::fmt`, those changes were removed from that PR.

This make tools that directly depends on FEXCore to use `fextl::fmt` when possible.

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -493,13 +493,13 @@ bool DownloadToPathWithZenityProgress(const fextl::string& URL, const fextl::str
   // -# for progress bar
   // -o for output file
   // -f for silent fail
-  std::string CurlPipe = fmt::format("curl -C - -#f {} -o {} 2>&1", URL, PathName);
+  fextl::string CurlPipe = fextl::fmt::format("curl -C - -#f {} -o {} 2>&1", URL, PathName);
   const std::string StdBuf = "stdbuf -oL tr '\\r' '\\n'";
   const std::string SedBuf = "sed -u 's/[^0-9]*\\([0-9]*\\).*/\\1/'";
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
I think the build failures are coming from the c++ standard being set to gnu++20 rather than c++20?
Might need to add `set(CMAKE_CXX_EXTENSIONS OFF)` at https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L97 to get that fixed. Hopefully won't introduce other problems.

neobrain:
I'm -0 on the latest iteration of this PR.

fextl was explicitly added to deal with FEXLoader's esoteric allocator requirements (i.e. keeping allocations within the guest-accessible 32-bit address space). This is not needed for other tools, and I would prefer those continued to live outside the `fextl::` world to reduce overall code weirdness and surprises. (Granted, existing code is already a bit mixed here due to involving FEXCore APIs.)

As for the general goal of reducing the number of external dependencies, I'm not sure `fmt` is a good place to start. `fmt` is self-contained and fairly well-packaged among various distributions.

Meanwhile, C++20 (even if Ubuntu 22.04 did have a recent-enough libstdc++) does not provide a full replacement, since useful utilities like `fmt::print` and `fmt::join` are missing (the latter of which I'm not sure we can provide a replacement for). If `fmt` were our last remaining external dependency it would make sense to push for a compromise, but currently I don't see a big value-gain to justify the transition.

neobrain:
Closing for now - let's pick this up again once `Externals` is slimmed down more generally and/or `std::format` is more feature-complete.

marysaka:
Oh yeah forgot to close it, sorry about that :+1:

PR #3710, opened 2024-06-17
CoreState: Move `InlineJITBlockHeader` to the start of the struct

Sonicadvance1:
This currently doesn't do much but soon this will be very important to ensure the data prefetcher of Cortex keeps the cachelines following this variable in L1.


alyssarosenzweig:
RCLSE can't be deleted soon enough..

alyssarosenzweig (approved):

Sonicadvance1:
> RCLSE can't be deleted soon enough..

I look forward to it :)

PR #3711, opened 2024-06-17
FEXCore: Disentangle the SVE256 feature from AVX

Sonicadvance1:
In quite a few locations we are mixing the case that SVE256 == AVX or that AVX means the guest register size is 256-bit.

While this is true today, this is entanglement is going to change very quickly and cause confusion in follow-up PRs.

Now we have SVE128, SVE256, and SVE2 HostFeatures to disambiguate the different features which mean different things.

This PR keeps the alias that `SupportsAVX` = `SupportsSVE256 && SupportsSVE2` but that alias is going to very quickly change its definition.

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -624,7 +624,7 @@ void Arm64Emitter::SpillStaticRegs(ARMEmitter::Register TmpReg, bool FPRs, uint3
   }
 
   if (FPRs) {
-    if (EmitterCTX->HostFeatures.SupportsAVX) {
+    if (EmitterCTX->HostFeatures.SupportsSVE256) {
       for (size_t i = 0; i < StaticFPRegisters.size(); i++) {
         const auto Reg = StaticFPRegisters[i];
 
@@ -709,11 +709,11 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
     // We don't bother spilling these in SpillStaticRegs,
     // since all that matters is we restore them on a fill.
     // It's not a concern if they get trounced by something else.
-    if (EmitterCTX->HostFeatures.SupportsSVE) {
+    if (EmitterCTX->HostFeatures.SupportsSVE128) {
       ptrue(ARMEmitter::SubRegSize::i8Bit, PRED_TMP_16B, ARMEmitter::PredicatePattern::SVE_VL16);
     }
 
-    if (EmitterCTX->HostFeatures.SupportsAVX) {
+    if (EmitterCTX->HostFeatures.SupportsSVE256) {
       ptrue(ARMEmitter::SubRegSize::i8Bit, PRED_TMP_32B, ARMEmitter::PredicatePattern::SVE_VL32);
 
       for (size_t i = 0; i < StaticFPRegisters.size(); i++) {
@@ -779,8 +779,8 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
   }
 }
 
-void Arm64Emitter::PushVectorRegisters(ARMEmitter::Register TmpReg, bool SVERegs, std::span<const ARMEmitter::VRegister> VRegs) {
-  if (SVERegs) {
+void Arm64Emitter::PushVectorRegisters(ARMEmitter::Register TmpReg, bool SVE256Regs, std::span<const ARMEmitter::VRegister> VRegs) {
+  if (SVE256Regs) {
     size_t i = 0;
 
     for (; i < (VRegs.size() % 4); i += 2) {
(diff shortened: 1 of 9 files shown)


alyssarosenzweig:
For theoretical hardware that supports SVE but not SVE2, would FEX use SVE for anything? Or do we assume SVE2 is the baseline?

If the latter - could we drop the separate SVE2 feature and just roll it into the SVE check?

Sonicadvance1:
> For theoretical hardware that supports SVE but not SVE2, would FEX use SVE for anything? Or do we assume SVE2 is the baseline?
> 
> If the latter - could we drop the separate SVE2 feature and just roll it into the SVE check?

We could likely drop the SVE1 check. I don't think we really care about Neoverse-V1 or Fujitsu A64X that much.

alyssarosenzweig:
> > For theoretical hardware that supports SVE but not SVE2, would FEX use SVE for anything? Or do we assume SVE2 is the baseline?
> > If the latter - could we drop the separate SVE2 feature and just roll it into the SVE check?
> 
> We could likely drop the SVE1 check. I don't think we really care about Neoverse-V1 or Fujitsu A64X that much.

Ok, if we're going to be munging all the feature flags anyway, I think it makes sense to squash that into here then. Either way lgtm

Sonicadvance1:
> > > For theoretical hardware that supports SVE but not SVE2, would FEX use SVE for anything? Or do we assume SVE2 is the baseline?
> > > If the latter - could we drop the separate SVE2 feature and just roll it into the SVE check?
> > 
> > 
> > We could likely drop the SVE1 check. I don't think we really care about Neoverse-V1 or Fujitsu A64X that much.
> 
> Ok, if we're going to be munging all the feature flags anyway, I think it makes sense to squash that into here then. Either way lgtm

Removed checks for SVE1 and now we only check for SVE2.

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -923,8 +923,8 @@ void Arm64Emitter::PopDynamicRegsAndLR() {
 }
 
 void Arm64Emitter::SpillForPreserveAllABICall(ARMEmitter::Register TmpReg, bool FPRs) {
-  const auto CanUseSVE = EmitterCTX->HostFeatures.SupportsAVX;
-  const auto FPRRegSize = CanUseSVE ? Core::CPUState::XMM_AVX_REG_SIZE : Core::CPUState::XMM_SSE_REG_SIZE;
+  const auto CanUseSVE256 = EmitterCTX->HostFeatures.SupportsSVE256;
+  const auto FPRRegSize = CanUseSVE256 ? Core::CPUState::XMM_AVX_REG_SIZE : Core::CPUState::XMM_SSE_REG_SIZE;
```
alyssarosenzweig, line 927: This should probably changed to `? 32 : 8` but shrug.

FEXCore/Source/Interface/Core/HostFeatures.cpp
```diff
@@ -143,11 +150,13 @@ HostFeatures::HostFeatures() {
...
-  SupportsSVE = true;
+  SupportsSVE128 = true;
+  SupportsSVE256 = true;
   SupportsAVX = true;
 #else
-  SupportsSVE = Features.Has(vixl::CPUFeatures::Feature::kSVE);
-  SupportsAVX = Features.Has(vixl::CPUFeatures::Feature::kSVE2) && vixl::aarch64::CPU::ReadSVEVectorLengthInBits() >= 256;
+  SupportsSVE128 = Features.Has(vixl::CPUFeatures::Feature::kSVE2);
+  SupportsSVE256 = Features.Has(vixl::CPUFeatures::Feature::kSVE2) && vixl::aarch64::CPU::ReadSVEVectorLengthInBits() >= 256;
+  SupportsAVX = SupportsSVE256;
```
alyssarosenzweig, line 159: Pop this assignment out of the #if and then drop the `AVX = true` from the simulator

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: I'll make a note of it, but I haven't gotten to rewriting the logic of these yet.

alyssarosenzweig: I just mean, `CanAVX ? AVXSize : SSESize` makes sense, and `Can256 ? 256 : 128` makes sense, but `Can256 ? AVXSize : SSESize` doesn't make sense, so if we're renaming this feature now we should replace the constants with just 256 (32) and 128 (16) to match.

alyssarosenzweig (approved):

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
Sonicadvance1: Sure, once I return from a nap I will replace it with numeric constants.

Sonicadvance1: Switched to numeric constants.

alyssarosenzweig (approved):

PR #3715, opened 2024-06-17
FXCH should set C1 to zero

pmatos:


pmatos:
Pessimization unfortunately is a consequence of the missing flag setting.

pmatos:
Updated instcountci - ready for review.

Sonicadvance1 (approved):
Nice

PR #3719, opened 2024-06-18
OpcodeDispatcher: Handle F16C operations

lioncash:
Preliminary PR, just to get CI on this. Some of the tests still need expanding.


Sonicadvance1 (commented):
Very nice. So it looks like what's left is to write some tests for the four rounding modes? So generating some various precise 32-bit float values both positive and negative and just run through all eight rounding mode variants? 4 imm, 4 MXCSR?

lioncash:
Yeah, expanding rounding mode coverage is the main thing left

Sonicadvance1:
> Yeah, expanding rounding mode coverage is the main thing left

I guess this might actually be spicy since vixl sim only supports one round mode. might need to put them on the known failures list once created and only get fully tested with AVX128.

lioncash:
This should be all good to go now

alyssarosenzweig:
Looks good other than the thing Ryan mentioned. I can look into optimizing vcvtps2ph after this merges.

Sonicadvance1 (approved):
Nice nice nice nice

PR #3721, opened 2024-06-18
Scripts: add update_instcountci.sh script

alyssarosenzweig:
This is helpful for devs working on FEXCore, I've been using this locally but it might make sense to stick it in tree.

### DIFF
--- a/Scripts/update_instcountci.sh
+++ b/Scripts/update_instcountci.sh
@@ -1,0 +1,14 @@
+#! /bin/bash
+set -e
+
+# Make sure we actually build
+ninja
+
+# Run tests, ignoring the retval since there will be changes.
+ninja instcountci_tests || true
+
+# Now we can update.
+ninja instcountci_update_tests
+
+# Commit the result in bulk.
+git commit -sam "InstCountCI: Update"


pmatos:
SGTM I wonder if it would make sense to connect the instcounci commit to the commit that caused the change? Maybe it could be an optional argument to the script?

alyssarosenzweig:
> SGTM I wonder if it would make sense to connect the instcounci commit to the commit that caused the change? Maybe it could be an optional argument to the script?

There's not really a good way to do that, I typically have a single instcountci update commit at the end of a PR with many commits. But also it hasn't been a problem so... 🤷‍♀️? If you wanted to add that after this merges I'm open to it but I probably wouldn't use it myself 😅

pmatos:
> > SGTM I wonder if it would make sense to connect the instcounci commit to the commit that caused the change? Maybe it could be an optional argument to the script?
> 
> There's not really a good way to do that, I typically have a single instcountci update commit at the end of a PR with many commits. But also it hasn't been a problem so... 🤷‍♀️? If you wanted to add that after this merges I'm open to it but I probably wouldn't use it myself 😅

Fair! :) Not meaning to make a huge deal about this process.

Sonicadvance1 (approved):

PR #3722, opened 2024-06-18
InstCountCI: explicitly disable AFP everywhere

alyssarosenzweig:
(except for when we explicitly enable AFP).

Since AFP gets saved/restored, we get `msr fpcr` garbage in random instructions when AFP is enabled. Explicitly disable everywhere since it's not worth our time to triage which files might hit that path. Fixes instcountci on AFP-supporting hosts now that we have AFP enabled.


pmatos:
I think you missed `InstructionCountCI/FlagM/HotBlocks_32Bit.json`

pmatos:
Tested! All good. Lets merge.

lioncash (approved):

PR #3724, opened 2024-06-18
First few commits from Ryan's AVX branch

alyssarosenzweig:
I feel comfortable Approve'ing these as-is, but not the last commit there yet. I think these can land early now though which gets the big instcountci patch out of the way.


Sonicadvance1 (approved):
🤔

PR #3726, opened 2024-06-18
HostFeatures: Work around Qualcomm Oryon RNG errata

Sonicadvance1:
The Oryon is the first CPU we know of that implemented support for the RNG extension. It also has an errata where reading the RNDRRS register never returns success. X86's RDSEED guarantees forward progress with enough retries.

When an x86 processor messed this up at one point, some Linux systems would infinite loop (presumably when something in boot was filling an entropy pool). This required a microcode change to fix that processor.

The rdseed unittest infinite loops on this platform if RNG was exposed.

### DIFF
--- a/FEXCore/Source/Interface/Core/HostFeatures.cpp
+++ b/FEXCore/Source/Interface/Core/HostFeatures.cpp
@@ -44,6 +44,13 @@ static uint32_t GetFPCR() {
 static void SetFPCR(uint64_t Value) {
   __asm("msr FPCR, %[Value]" ::[Value] "r"(Value));
 }
+
+static uint32_t GetMIDR() {
+  uint64_t Result {};
+  __asm("mrs %[Res], MIDR_EL1" : [Res] "=r"(Result));
+  return Result;
+}
+
 #else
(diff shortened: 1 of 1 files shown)


alyssarosenzweig (approved):
Ouch

FEXCore/Source/Interface/Core/HostFeatures.cpp
```diff
@@ -44,6 +44,13 @@ static uint32_t GetFPCR() {
 static void SetFPCR(uint64_t Value) {
   __asm("msr FPCR, %[Value]" ::[Value] "r"(Value));
 }
+
+uint32_t GetMIDR() {
```
lioncash, line 48: Seems like this should be static

lioncash (commented):
One minor nit, but looks good overall

FEXCore/Source/Interface/Core/HostFeatures.cpp
Sonicadvance1: There we go.

lioncash (approved):

PR #3727, opened 2024-06-19
VAES support

Sonicadvance1:
Not directly related to AVX128 but we will be gaining support for VAES through it, so good to add tests before the rest.

1. Implement support for querying VAES through HostFeatures
2. Add unittests for the four new instructions
3. Expose the feature in CPUID if supported


Sonicadvance1:
@pmatos Looks like the code formatting script isn't properly ignoring python files. This PR will need to ignore the code formatting failure until that gets resolved.

pmatos:
> @pmatos Looks like the code formatting script isn't properly ignoring python files. This PR will need to ignore the code formatting failure until that gets resolved.

I will get to it today.

alyssarosenzweig (approved):
cute lil extension 👍

pmatos:
> > @pmatos Looks like the code formatting script isn't properly ignoring python files. This PR will need to ignore the code formatting failure until that gets resolved.
> 
> I will get to it today.

@Sonicadvance1 fixed #3728

PR #3728, opened 2024-06-19
Ignore python files for clang-format

pmatos:


neobrain:
Why do Python files need to be special-cased here in the first place? Do we need similar changes for json and other file types that aren't C/C++ sources?

pmatos:
> Why do Python files need to be special-cased here in the first place? Do we need similar changes for json and other file types that aren't C/C++ sources?

Nope - my mistake.  Initially I thought we wanted to format python so I kept the python formatter and use of darker. I removed it now.

Sonicadvance1 (approved):

PR #3729, opened 2024-06-19
OpcodeDispatcher: Refactor address modes

alyssarosenzweig:
Extract some helpers to deduplicate the code to decode address modes, so when AVX128 adds its own load/store helpers we won't end up with 4 (!) copies of this logic.

[I'm now porting the AVX128 helpers to use this, don't worry about that]


Sonicadvance1:
Oh no! My code duplication!

Sonicadvance1 (approved):

PR #3730, opened 2024-06-20
Vector: Helper refactorings

Sonicadvance1:
Refactors some of the base vector helpers so that AVX128 can reuse them.


alyssarosenzweig:
What are those instcountci changes from?!

Sonicadvance1:
> What are those instcountci changes from?!

Oopsies, had a `FEX_HOSTFEATURES` override set.

alyssarosenzweig (approved):

PR #3731, opened 2024-06-20
HostFeatures: Always disable AVX in 32-bit mode to protect from stack overflows

Sonicadvance1:
As it says on the tin.

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -624,7 +624,7 @@ void Arm64Emitter::SpillStaticRegs(ARMEmitter::Register TmpReg, bool FPRs, uint3
   }
 
   if (FPRs) {
-    if (EmitterCTX->HostFeatures.SupportsSVE256) {
+    if (EmitterCTX->HostFeatures.SupportsAVX && EmitterCTX->HostFeatures.SupportsSVE256) {
       for (size_t i = 0; i < StaticFPRegisters.size(); i++) {
         const auto Reg = StaticFPRegisters[i];
 
@@ -715,7 +715,9 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
 
     if (EmitterCTX->HostFeatures.SupportsSVE256) {
       ptrue(ARMEmitter::SubRegSize::i8Bit, PRED_TMP_32B, ARMEmitter::PredicatePattern::SVE_VL32);
+    }
 
+    if (EmitterCTX->HostFeatures.SupportsAVX && EmitterCTX->HostFeatures.SupportsSVE256) {
       for (size_t i = 0; i < StaticFPRegisters.size(); i++) {
         const auto Reg = StaticFPRegisters[i];
         if (((1U << Reg.Idx()) & FPRFillMask) != 0) {
--- a/FEXCore/Source/Interface/Core/HostFeatures.cpp
+++ b/FEXCore/Source/Interface/Core/HostFeatures.cpp
@@ -58,7 +58,7 @@ static uint32_t GetDCZID() {
 }
 #endif
 
-static void OverrideFeatures(HostFeatures* Features) {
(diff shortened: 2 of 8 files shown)


Sonicadvance1:
This one might be getting caught by one of the other PR bug fixes, luckily it is fairly freestanding so it can live independently for a bit before I investigate fully.

neobrain:
What does one have to do with the other?

Sonicadvance1:
> What does one have to do with the other?

When AVX256 is enabled, signal frames start using significantly more stack space.
- 16bytes * 16 registers = 256 bytes for XMM registers
- 32bytes * 16 registers = 512 bytes for YMM registers

There are known game failures on real x86 hardware where a 32-bit game is running up against the wall on stack space on non-AVX hardware and then explodes when run on AVX hardware. This is to guard against that.

alyssarosenzweig:
> > What does one have to do with the other?
> 
> When AVX256 is enabled, signal frames start using significantly more stack space.
> 
> * 16bytes * 16 registers = 256 bytes for XMM registers
> * 32bytes * 16 registers = 512 bytes for YMM registers
> 
> There are known game failures on real x86 hardware where a 32-bit game is running up against the wall on stack space on non-AVX hardware and then explodes when run on AVX hardware. This is to guard against that.

Could you put that in the comment for the next person who tries to "helpfully" enable the feature?

Sonicadvance1:
> There are known game failures on real x86 hardware where a 32-bit game is running up against the wall on stack space on non-AVX hardware and then explodes when run on AVX hardware. This is to guard against that.

Good idea.

alyssarosenzweig (approved):

alyssarosenzweig:
> So, to me that seems as solving a software bug/forward-compatibility-problem by doing something different than the hardware you emulate? Or I'm misunderstanding it?
> 
> Wouldn't that break compatibility with other software? Maybe there aren't a lot of AVX 32-bit programs, but still, are you sure you want to introduce divergence/incompatibility?
> 
> * [Intel AVX instructions will be available on both 32bit and 64bit flavors of Processor and OS](https://community.intel.com/t5/Intel-ISA-Extensions/Are-AVX-Intrinsics-supported-on-32-bit-processors/m-p/771334)
> * [VEX can be used in 32-bit mode with the some restrictions](https://en.wikipedia.org/wiki/VEX_prefix)

In theory, yes. In practice... if we *know* of games that break with 32-bit AVX (including on real hardware), but we don't know of any 32-bit games that hard require AVX... then disabling AVX for 32-bit seems like the right call. FEX isn't designed to be bug-for-bug compatible with any particular Intel or AMD CPU.

alyssarosenzweig (approved):

Sonicadvance1:
Rebased on current main. This disablement code is still before the `FEX_HOSTFEATURES` override. So if a user wants to try 32-bit games with AVX, they can still set the override which will enable it.

alyssarosenzweig (approved):
+1, if ever we find a 32-bit AVX-only title we can appconfig it, this seems like the saner default for now.

Sonicadvance1:
vixl simulator bug getting caused by #3733. Investigating if that needs to go in first and what it actually fixes...

alyssarosenzweig:
> vixl simulator bug getting caused by #3733. Investigating if that needs to go in first and what it actually fixes...

I'm not hard opposed to merging that if it's going to be a release blocker.  (But I don't like reverts that I don't understand.)

Sonicadvance1:
This was a smidge quirkier to fix than expected. Since I had to detangle a few locations that still assumed AVX = SVE256, which is kind of funny

alyssarosenzweig (approved):
kind of wonder if we shouldn't just smash SupportSVE256 to false when AVX is disabled but this is probably cleaner.

alyssarosenzweig:
> vixl simulator bug getting caused by #3733. Investigating if that needs to go in first and what it actually fixes...

With the fixes here, do we still need #3733?

Sonicadvance1:
> > vixl simulator bug getting caused by #3733. Investigating if that needs to go in first and what it actually fixes...
> 
> With the fixes here, do we still need #3733?

I believe #3733 is now irrelevant, since it was an unrelated bug

PR #3732, opened 2024-06-20
unittests: Split up vtestps unittest to accumulate flags in independent registers.

Sonicadvance1:
Makes it easier to see what is failing on the 128-bit side versus 256-bit side.

### DIFF
--- a/unittests/ASM/VEX/vtestps.asm
+++ b/unittests/ASM/VEX/vtestps.asm
@@ -2,7 +2,8 @@
 {
   "HostFeatures": ["AVX"],
   "RegData": {
-    "R15":  "0x0000000EDDFFB77F",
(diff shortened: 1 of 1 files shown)


alyssarosenzweig (approved):

unittests/ASM/VEX/vtestps.asm
```diff
@@ -12,7 +13,7 @@
 
 ; Uses AX and BX and stores result in r15
```
lioncash, line 14: Perhaps remove the r15 bit of the comment, since this is no longer true

Sonicadvance1: You're right, I'll change that in a bit.

PR #3733, opened 2024-06-20
Arm64: Revert #2865 optimization

Sonicadvance1:
This was causing a regression in the `vpmaddubsw` unittest on AVX128. Not sure if this is a vixl simulator bug or misunderstanding how the instructions work.

Remove the optimization until we can investigate further.

Needs to be tested on actual SVE-128 hardware.


alyssarosenzweig:
I'm mostly iffy on merging a revert without any sense of what's broken or even if anything is broken since it could just as easily be papering over some issue elsewhere that could bite later. As agreed let's let this hang out for a while while focusing review/merge effort on the other commits in the merge party. -0 from me.

PR #3736, opened 2024-06-21
AVX128: Some pun pickles, moves and conversions

Sonicadvance1:
First use case of the binary operation helper. Some are related, some are not.

Removing the `SupportsSaturatingRoundingShifts` option because it's always true now.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
@@ -63,11 +63,11 @@ void OpDispatchBuilder::InstallAVX128Handlers() {
     {OPD(1, 0b00, 0x2B), 1, &OpDispatchBuilder::AVX128_MOVVectorNT},
     {OPD(1, 0b01, 0x2B), 1, &OpDispatchBuilder::AVX128_MOVVectorNT},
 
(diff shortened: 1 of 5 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -812,4 +812,41 @@ void OpDispatchBuilder::AVX128_MOVVectorUnaligned(OpcodeArgs) {
...
+    // then it is more optimal to load in to the FPR register directly and convert there.
+    auto Src2 = AVX128_LoadSource_WithOpSize(Op, Op->Src[1], Op->Flags, false);
+    // Always signed
+    Result.Low = _VSToFVectorInsert(IR::SizeToOpSize(DstSize), DstElementSize, DstElementSize, Src1.Low, Src2.Low, false, false);
+  }
+
+  if (Is128Bit) {
+    Result.High = LoadZeroVector(OpSize::i128Bit);
+  } else {
+    LOGMAN_MSG_A_FMT("Programming Error: This should never occur!");
```
alyssarosenzweig, line 846: drop the control flow, LOGMAN_THROW_A_FMT(!Is128Bit, "...") instead.

Sonicadvance1: Dropped and added assert

alyssarosenzweig (approved):

PR #3740, opened 2024-06-21
AVX128: More various instructions

Sonicadvance1:
A smattering of scalar operations and vector shifts.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
@@ -181,8 +181,8 @@ void OpDispatchBuilder::InstallAVX128Handlers() {
 
     {OPD(1, 0b00, 0xC2), 1, &OpDispatchBuilder::AVX128_VFCMP<4>},
     {OPD(1, 0b01, 0xC2), 1, &OpDispatchBuilder::AVX128_VFCMP<8>},
-    // TODO: {OPD(1, 0b10, 0xC2), 1, &OpDispatchBuilder::AVXInsertScalarFCMPOp<4>},
-    // TODO: {OPD(1, 0b11, 0xC2), 1, &OpDispatchBuilder::AVXInsertScalarFCMPOp<8>},
+    {OPD(1, 0b10, 0xC2), 1, &OpDispatchBuilder::AVX128_InsertScalarFCMP<4>},
+    {OPD(1, 0b11, 0xC2), 1, &OpDispatchBuilder::AVX128_InsertScalarFCMP<8>},
 
     // TODO: {OPD(1, 0b01, 0xC4), 1, &OpDispatchBuilder::VPINSRWOp},
     // TODO: {OPD(1, 0b01, 0xC5), 1, &OpDispatchBuilder::PExtrOp<2>},
@@ -948,6 +948,30 @@ void OpDispatchBuilder::AVX128_VFCMP(OpcodeArgs) {
   AVX128_VectorBinaryImpl(Op, GetSrcSize(Op), ElementSize, [this, Op, CompType](size_t _ElementSize, Ref Src1, Ref Src2) {
     return VFCMPOpImpl(Op, _ElementSize, Src1, Src2, CompType);
   });
+}
+
+template<size_t ElementSize>
+void OpDispatchBuilder::AVX128_InsertScalarFCMP(OpcodeArgs) {
+  // We load the full vector width when dealing with a source vector,
+  // so that we don't do any unnecessary zero extension to the scalar
+  // element that we're going to operate on.
+  const auto SrcSize = GetSrcSize(Op);
+
+  auto Src1 = AVX128_LoadSource_WithOpSize(Op, Op->Src[0], Op->Flags, false);
+  RefPair Src2 {};
+
+  if (Op->Src[1].IsGPR()) {
+    Src2 = AVX128_LoadSource_WithOpSize(Op, Op->Src[1], Op->Flags, false);
+  } else {
+    Src2.Low = LoadSource_WithOpSize(FPRClass, Op, Op->Src[1], SrcSize, Op->Flags);
+  }
+
+  const uint8_t CompType = Op->Src[2].Literal();
+
+  RefPair Result {};
+  Result.Low = InsertScalarFCMPOpImpl(OpSize::i128Bit, OpSize::i128Bit, ElementSize, Src1.Low, Src2.Low, CompType, false);
+  Result.High = LoadZeroVector(OpSize::i128Bit);
+  AVX128_StoreResult_WithOpSize(Op, Op->Dest, Result);
 }
 
 } // namespace FEXCore::IR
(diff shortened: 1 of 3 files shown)


alyssarosenzweig:
I was going to flag the copypaste in UCOMIS, instead of doing that I wrote #3741 to use 😄

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -948,4 +948,73 @@ void OpDispatchBuilder::AVX128_VectorScalarInsertALU(OpcodeArgs) {
   AVX128_StoreResult_WithOpSize(Op, Op->Dest, RefPair {.Low = Result_Low, .High = High});
 }
 
+Ref OpDispatchBuilder::AVX128_VFCMPImpl(size_t ElementSize, Ref Src1, Ref Src2, uint8_t CompType) {
+  Ref Result {};
+  switch (CompType) {
```
alyssarosenzweig, line 953: Unless there's a strong reason otherwise, I'd rather switch on `CompType & 7` which lets you delete 2/3 the code here.

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -948,4 +948,73 @@ void OpDispatchBuilder::AVX128_VectorScalarInsertALU(OpcodeArgs) {
...
+  case 0x00:
+  case 0x08:
+  case 0x10:
+  case 0x18: // EQ
+    return _VFCMPEQ(OpSize::i128Bit, ElementSize, Src1, Src2);
+    break;
```
alyssarosenzweig, line 959: No need to break after a return (likewise throughout)

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
Sonicadvance1: Fixed

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
Sonicadvance1: Also fixed

Sonicadvance1:
> I was going to flag the copypaste in UCOMIS, instead of doing that I wrote #3741 to use 😄

Cherry-picked and used.

alyssarosenzweig:
Wrote #3743 for all the Literals flying around here, but I don't mind merging without porting to that, I can do the (fully mechanical) changes after merge if you prefer.

Sonicadvance1:
> Wrote #3743 for all the Literals flying around here, but I don't mind merging without porting to that, I can do the (fully mechanical) changes after merge if you prefer.

Yeesss, that would be good to do in one lump commit once the 66 remaining commits are merged. Since there will be a decent amount of them after the fact and it will be painful to go through each individual commit

alyssarosenzweig:
> > Wrote #3743 for all the Literals flying around here, but I don't mind merging without porting to that, I can do the (fully mechanical) changes after merge if you prefer.
> 
> Yeesss, that would be good to do in one lump commit once the 66 remaining commits are merged. Since there will be a decent amount of them after the fact and it will be painful to go through each individual commit

Sure, that's fine with me. Although https://github.com/torbiak/git-autofixup is basically tailormade for this problem - you do all the changes at the end, run `git autofixup` and then it creates a whole slew of fixup! commits for you to `git rebase` and then it's all fixed up. Magic!

alyssarosenzweig:
I see the fcmp funniness is an existing funniness. So I wrote you #3744

alyssarosenzweig:
VPSRA/VPSLL/VPSRL look like candidates to use the lambda helper. If that doesn't work for some reason (I don't see such a reason but maybe ther'es a subtlelty I'm missing here), we can at least template between the 3

alyssarosenzweig:
oops accidental close

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -1270,4 +1270,235 @@ void OpDispatchBuilder::AVX128_VPINSRDQ(OpcodeArgs) {
...
+  const auto Is128Bit = DstSize == Core::CPUState::XMM_SSE_REG_SIZE;
+
+  LOGMAN_THROW_A_FMT(Op->Src[1].IsLiteral(), "Src1 needs to be literal here");
+  const uint64_t ShiftConstant = Op->Src[1].Data.Literal.Value;
+
+  auto Src = AVX128_LoadSource_WithOpSize(Op, Op->Src[0], Op->Flags, !Is128Bit);
+  RefPair Result {};
+
+  if (ShiftConstant == 0) [[unlikely]] {
+    Result.Low = Src.Low;
```
alyssarosenzweig, line 1367: `Result = Src`?

alyssarosenzweig:
I just pushed a version with all my comments addressed. I'm happy to merge it in this state if you are. Diff between what I pushed and what you pushed rebased on #3743, for review convenience: https://rosenzweig.io/diff.txt

Sonicadvance1:
> I just pushed a version with all my comments addressed. I'm happy to merge it in this state if you are. Diff between what I pushed and what you pushed rebased on #3743, for review convenience: https://rosenzweig.io/diff.txt

Looks good to me

alyssarosenzweig (approved):

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -901,4 +950,402 @@ void OpDispatchBuilder::AVX128_VPSIGN(OpcodeArgs) {
...
+
+  AVX128_StoreResult_WithOpSize(Op, Op->Dest, Result);
+}
+
+void OpDispatchBuilder::AVX128_VPSRLDQ(OpcodeArgs) {
+  AVX128_ShiftDoubleImm(Op, true /* Right */);
+}
+
+void OpDispatchBuilder::AVX128_VPSLLDQ(OpcodeArgs) {
+  AVX128_ShiftDoubleImm(Op, false /* Right */);
```
lioncash, line 1348:
```suggestion
  AVX128_ShiftDoubleImm(Op, false /* Left */);
```

alyssarosenzweig: I think that might be more confusing. it is false that this is Right. it is true that this is Left.

lioncash: Conversely I find that extremely confusing. The programmatic part should absolutely not be inverting a comment lol, and it genuinely comes across as a copy-paste typo to me. This can be more explicit by just making the comments "Right shift" and "Left shift".

Sonicadvance1: Easy fix for this is to do a simple enum
```cpp
enum class ShiftDirection {
  RIGHT, LEFT,
};

void AVX128_ShiftDoubleImm(OpcodeArgs, ShiftDirection Dir);

void OpDispatchBuilder::AVX128_VPSRLDQ(OpcodeArgs) {
  AVX128_ShiftDoubleImm(Op, ShiftDirection::RIGHT);
}

void OpDispatchBuilder::AVX128_VPSLLDQ(OpcodeArgs) {
  AVX128_ShiftDoubleImm(Op, ShiftDirection::LEFT);
}
```

alyssarosenzweig: We have a winner!

PR #3742, opened 2024-06-21
FEXCore: Implement AVX reconstruction helpers

Sonicadvance1:
Due to the distinct view of AVX registers changing depending on if the host supports SVE256 or not, introduce some helpers to reconverge the view in the ways that Linux and Wow64 wants.

The TestHarnessRunner has a bit of a kludge but rewriting it to not read the InternalThreadState object directly will take longer than currently desired. So fetch the register state, then reconverge the AVX register state. Once the InternalThreadState object changes its view to cut out those excess 256bytes then this will get refactored at that point.

### DIFF
--- a/FEXCore/include/FEXCore/Core/Context.h
+++ b/FEXCore/include/FEXCore/Core/Context.h
@@ -194,6 +194,11 @@ public:
   ///< Sets FEX's internal EFLAGS representation to the passed in compacted form.
   FEX_DEFAULT_VISIBILITY virtual void SetFlagsFromCompactedEFLAGS(FEXCore::Core::InternalThreadState* Thread, uint32_t EFLAGS) = 0;
 
+  FEX_DEFAULT_VISIBILITY virtual void
+  ReconstructXMMRegisters(const FEXCore::Core::InternalThreadState* Thread, __uint128_t* XMM_Low, __uint128_t* YMM_High) = 0;
+  FEX_DEFAULT_VISIBILITY virtual void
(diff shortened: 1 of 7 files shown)


FEXCore/include/FEXCore/Core/Context.h
```diff
@@ -194,6 +194,16 @@ class Context {
   ///< Sets FEX's internal EFLAGS representation to the passed in compacted form.
   FEX_DEFAULT_VISIBILITY virtual void SetFlagsFromCompactedEFLAGS(FEXCore::Core::InternalThreadState* Thread, uint32_t EFLAGS) = 0;
 
+  enum class XMMRecoveryMode {
```
alyssarosenzweig, line 197: Does it make sense to drop this enum, and instead infer the recovery mode inside the helpers by checking if `YMM_High != nullptr`? This would simplify the helper signatures a fair bit and remove a failure mode (passing a mode that doesn't match whether you passed a YMM_High pointer)

alyssarosenzweig (changes requested):
.

FEXCore/include/FEXCore/Core/Context.h
Sonicadvance1: Indeed, that might make sense

alyssarosenzweig: I've gone ahead and made this change, IMO it's a lot nicer now, see diff https://rosenzweig.io/diff2.txt ... approving, I'm good with this to go in if you're ok with that diff.

alyssarosenzweig (approved):

Sonicadvance1 (commented):
I can't approve my own PR but the changes look good

alyssarosenzweig:
> I can't approve my own PR but the changes look good

You can't approve it but you can merge it 😄

PR #3743, opened 2024-06-21
X86Tables: add Literal() helper

alyssarosenzweig:
Any time we get the value of Literal, we want to assert that it's actually a
    literal. We've been open coding this pattern sporadically throughout the
    opcodedispatcher. Let's add an ergonomic helper to fetch the value of literal,
    asserting that the value is indeed literal.

This is very noticeable in the AVX changes.


Sonicadvance1 (approved):
Let the cpp flow through you!

PR #3744, opened 2024-06-21
Tweak vector FCMP helpers so we can reuse in AVX128

alyssarosenzweig:
on top of #3743 for conflicts, for #3740


alyssarosenzweig:
We'll merge this with #3740 itself, closing

PR #3745, opened 2024-06-23
Adds back cmake error on x86-64 hosts

Sonicadvance1:
This is a footgun for users to allow this configuration through without anything stopping them.
Had a user try this configuration today and not understand that it isn't supported.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -118,6 +118,12 @@ else()
 endif()
 
 if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
+  option(ENABLE_X86_HOST_DEBUG "Enables compiling on x86_64 host" FALSE)
+  if (NOT ENABLE_X86_HOST_DEBUG)
+    message(FATAL_ERROR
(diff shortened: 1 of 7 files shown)


CMakeLists.txt
```diff
@@ -118,6 +118,13 @@ else()
 endif()
 
 if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86_64")
+  option(ENABLE_X86_HOST_DEBUG "Enables compiling on x86_64 host" FALSE)
+  if (NOT ENABLE_X86_HOST_DEBUG)
+    message(FATAL_ERROR
+    " FEX-Emu doesn't support compiling for x86-64 hosts!"
+    " This is /only/ a supported configuration for FEX CI and nothing else!"
+    " Pass -DENABLE_X86_HOST_DEBUG=True to bypass this message!")
```
alyssarosenzweig, line 126: tbh drop this line, as long as CI knows the option better not tempt folk...

alyssarosenzweig (approved):

CMakeLists.txt
Sonicadvance1: Removed that line

PR #3747, opened 2024-06-24
AVX128: More instructions Part 3

Sonicadvance1:
Based on top of #3746

Adds AES instructions, SSE4.2 string ops, silly minposuw, and rounds.

PR #3748, opened 2024-06-24
AVX128: More instructions Part 4

Sonicadvance1:
Based on top of #3747

We're getting to the point of wrapping up all the AVX specific instruction implementations here.


alyssarosenzweig:
I've done a *lot* of surgery on this, but I think I'm happy with it. https://rosenzweig.io/avx15.diff

Sonicadvance1 (commented):
The changes look reasonable to me!

alyssarosenzweig (approved):

PR #3749, opened 2024-06-24
Arm64: Remove contiguous masked element optimization

Sonicadvance1:
This was a premature optimization and currently breaks. Just remove it for now.


alyssarosenzweig:
Breaks... how?

Sonicadvance1:
Not actually sure yet. qemu tcg tests showed the difference though, so reproducing will be straightforward when adding it back

lioncash (approved):

PR #3750, opened 2024-06-24
OpcodeDispatcher: Fixes bug in pshuf{lw,hw} 

Sonicadvance1:
This optimization was incorrect. Updates unittests to ensure it keeps
working.


alyssarosenzweig:
OOI what's wrong with this code? It 'seems' like it should be correct to my naive eyes.

Sonicadvance1:
It's not respecting the upper-bits of the XMM register in respect to how zip behaviour works. Noticed it with the qemu tcg tests again.

lioncash (approved):

PR #3751, opened 2024-06-24
InstcountCI: Support AVX flag

Sonicadvance1:


alyssarosenzweig:
I don't understand. We already have hostfeature flags for SVE. Why do we need an AVX flag if we're going to expose AVX on ASIMD unconditionally anyway?

Sonicadvance1:
> I don't understand. We already have hostfeature flags for SVE. Why do we need an AVX flag if we're going to expose AVX on ASIMD unconditionally anyway?

The problem right now is that FEX doesn't currently enable AVX by unconditionally (even with ASIMD and SVE) the only way to enable it is with the HostFeatures flag to set the enableavx.

This is because we don't yet know the full extent of how performance is going to change and needs investigation to determine if we can enable it unconditionally.

alyssarosenzweig:
I see... Neither approving nor nak'ing this now then. Will keep reviewing the other patches while you do perf testing and we'll see where we land at the end of the review wave.

alyssarosenzweig:
Since it sounds like we're going to enable AVX unconditionally (per #3768), I think we can close this.

PR #3755, opened 2024-06-24
AVX128: Fix vmovntdqa failing to zero upper 128-bits

Sonicadvance1:


lioncash (approved):

Sonicadvance1:
Oop, we also use this handler for `vmovntpd` which is a store. This was converting a 128-bit store in to a 256-bit store.
Only zero `Src.High` if destination is a GPR and is 128-bit, otherwise it is a store and we're only storing 128-bits.
Extends the `vmovntpd` test to ensure this

PR #3756, opened 2024-06-25
Fix VMOVLHPS instruction

Sonicadvance1:
We were missing a unittest for this and both the SVE256 and AVX128 implementations were incorrect. A bit confusing since x86 encodes VMOVLHPS and VMOVLPS as the same instruction, it just changes behaviour depending on if the destination or source is memory.


lioncash (commented):
nice catch

lioncash (approved):
ok, it helps if I actually approve the thing, rather than comment on it

PR #3757, opened 2024-06-25
AVX128: Implement support for gathers

Sonicadvance1:
Based on #3748


FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -940,6 +940,192 @@ DEF_OP(VStoreVectorMasked) {
...
+  ///< If the host supports SVE and the offset scale matches SVE limitations then it can do an SVE style load.
+  const bool SupportsSVELoad = (HostSupportsSVE128 || HostSupportsSVE256) && (OffsetScale == 1 || OffsetScale == VectorIndexSize) &&
+                               (VectorIndexSize == IROp->ElementSize);
+
+  const auto PerformSMove = [this](size_t ElementSize, const ARMEmitter::Register Dst, const ARMEmitter::VRegister Vector, int index) {
+    switch (ElementSize) {
+    case 1: smov<ARMEmitter::SubRegSize::i8Bit>(Dst.X(), Vector, index); break;
+    case 2: smov<ARMEmitter::SubRegSize::i16Bit>(Dst.X(), Vector, index); break;
+    case 4: smov<ARMEmitter::SubRegSize::i32Bit>(Dst.X(), Vector, index); break;
+    case 8: umov<ARMEmitter::SubRegSize::i64Bit>(Dst.X(), Vector, index); break;
```
alyssarosenzweig, line 987: I don't think it actually matters, but why is this umov?

Sonicadvance1: ARM doesn't have a 64-bit smov because it would be identical to a 64-bit umov

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -940,6 +940,192 @@ DEF_OP(VStoreVectorMasked) {
...
+      SVEScale = FEXCore::ilog2(OffsetScale);
+    }
+
+    ARMEmitter::Register AddrReg = TMP1;
+
+    if (BaseAddr.has_value()) {
+      AddrReg = GetReg(Op->AddrBase.ID());
+    } else {
+      ///< OpcodeDispatcher didn't provide a Base address while SVE requires one.
+      LoadConstant(ARMEmitter::Size::i64Bit, AddrReg, 0);
```
alyssarosenzweig, line 1021: I guess you can't pass xzr to ld1b? Not really sure how to find that out in the docs.

Sonicadvance1: Nah, it would try loading off the stack if we gave it xzr because of the register aliasing.

To find it we can look at the instruction definition:
`LD1B { <Zt>.D }, <Pg>/Z, [<Xn|SP>, <Zm>.D]`

Which states `Xn|SP` which means it treats the zero register as stack register :)

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -940,6 +940,192 @@ DEF_OP(VStoreVectorMasked) {
...
+    case 8: {
+      ld1d(TempDst.Z(), CMPPredicate.Zeroing(), MemDst);
+      break;
+    }
+    default: break;
+    }
+
+    ///< Merge elements based on predicate.
+    sel(SubRegSize, Dst.Z(), CMPPredicate, TempDst.Z(), IncomingDst.Z());
+  } else if (!Is256Bit) {
```
alyssarosenzweig, line 1056: Better to `} else { LOGMAN_ASSERT(!Is256Bit); ...`

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -450,7 +455,6 @@ OpDispatchBuilder::RefPair OpDispatchBuilder::AVX128_LoadSource_WithOpSize(
     AddressMode HighA = A;
     HighA.Offset += 16;
 
-    ///< TODO: Implement VSIB once we get there.
     if (Operand.IsSIB()) {
       const bool IsVSIB = (Op->Flags & X86Tables::DecodeFlags::FLAG_VSIB_BYTE) != 0;
       LOGMAN_THROW_AA_FMT(!IsVSIB, "VSIB currently unsupported");
```
alyssarosenzweig, line 460: assert msg is stale, change to something like "VSIB uses LoadVSIB instead"

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -2275,4 +2304,97 @@ void OpDispatchBuilder::AVX128_VPCLMULQDQ(OpcodeArgs) {
...
+      DataElementOffset = OpSize::i128Bit / ElementLoadSize / 2;
+    }
+
+    ///< Calculate the high-half.
+    auto ResultHigh = _VLoadVectorGatherMasked(OpSize::i128Bit, ElementLoadSize, DestReg, MaskReg, BaseAddr, AddrAddressing.Low,
+                                               AddrAddressing.High, AddrElementSize, VSIB.Scale, DataElementOffset, IndexElementOffset);
+
+    if (AddrElementSize == OpSize::i64Bit && ElementLoadSize == OpSize::i32Bit) {
+      // If we only fetched 128-bits worth of data then the upper-result is all zero.
+      Result.Low = ResultHigh;
```
alyssarosenzweig, line 2366: `Result = AVX128_Zext(ResultHigh)`

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
alyssarosenzweig: Ah, I didn't know about that. Thanks, got it!

alyssarosenzweig:
This all looks plausible to me but is such a mind-bend that I don't feel able to give this a competent review.

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
Sonicadvance1: Changed

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
Sonicadvance1: Changed.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
Sonicadvance1: Changed

alyssarosenzweig:
Rebased.

lioncash (approved):
As far as I can tell, this looks good to me

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -940,6 +940,191 @@ DEF_OP(VStoreVectorMasked) {
...
+    if (OffsetScale != 1) {
+      SVEScale = FEXCore::ilog2(OffsetScale);
+    }
+
+    ARMEmitter::Register AddrReg = TMP1;
+
+    if (BaseAddr.has_value()) {
+      AddrReg = GetReg(Op->AddrBase.ID());
+    } else {
+      ///< OpcodeDispatcher didn't provide a Base address while SVE requires one.
```
alyssarosenzweig, line 1020: Wondering if it wouldn't be better to make the Base mandatory for the IR, and have the OpcodeDispatcher plug in a `_Constant(0)` if the original x86 op didn't have a Base. That way, constant pooling can kick in.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
```diff
@@ -940,6 +940,191 @@ DEF_OP(VStoreVectorMasked) {
...
+    ARMEmitter::SVEModType ModType = ARMEmitter::SVEModType::MOD_NONE;
+    uint8_t SVEScale = 0;
+    if (VectorIndexSize == 4) {
+      ModType = ARMEmitter::SVEModType::MOD_SXTW;
+    } else if (VectorIndexSize == 8 && OffsetScale != 1) {
+      ModType = ARMEmitter::SVEModType::MOD_LSL;
+    }
+
+    if (OffsetScale != 1) {
+      SVEScale = FEXCore::ilog2(OffsetScale);
```
alyssarosenzweig, line 1012: ilog2(1) = 0, so we might as well drop the if and save 4 lines.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
alyssarosenzweig: Oh, that's worse for ASIMD. Never mind that.

FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
alyssarosenzweig: I've changed this.

alyssarosenzweig (approved):
This is a mind bend (especially all the size changing stuff). But it looks.. plausible at least.

Sonicadvance1:
Definitely a bit tricky to keep all the logic in your brain as the instruction's behaviour is quite complex.

PR #3758, opened 2024-06-25
AVX128: FMA3

Sonicadvance1:
Uses the FMA instructions that were added in #3752


alyssarosenzweig:
This all looks reasonable 👍

Sonicadvance1:
Rebased on main, avoiding the gather operations for now.
This will cause a minor merge conflict on the gather side, but since that one only has one function it will be a trivial resolve.

lioncash (approved):

PR #3759, opened 2024-06-25
AVX: Fixes PCMUL with weird selectors and zero-extend

Sonicadvance1:
We had a bug where we weren't correctly ignoring the non-used bits in the selector. This was causing an assert in the ARM backend.

Also fixes AVX128 failing to zero extend on 128-bit operations.


alyssarosenzweig:
I've cherry-picked the non-AVX bits of this into #3760

PR #3763, opened 2024-06-26
Remove RCLSE

alyssarosenzweig:
RCLSE is one of our most complex compiler passes. It is responsible for removing redundant context load/stores, and by extension register & flag load/stores. Its problems are well known:

* It's based on the context model, which requires awkward translations for registers.
* It's a lot of code, and not especially efficient code. It has a lot of complexity for things like partial tracking, which just don't matter that much.
* It has no awareness of instructions that are invalid to optimize across, if we're trying to guarantee precise exception handling to the JIT.
* Its effect on codegen is somewhat random, helping some blocks while hurting others due to the laissez-faire approach to SRA in FEX.

In a step to addressing these problems, this PR replaces RCLSE with a simple register cache in the OpcodeDispatcher, reusing loads and deferring stores until the cache is explicitly flushed. We treat a few important context entries (like AbridgedFTW) as pseudo-registers to get RCLSE-like behaviour there too. This has some tradeoffs:

* It is a lot simpler, a lot less code, and a lot faster.
* It will never optimize across a FlushRegisterCache boundary. In this iteration, we do a partial flush around every instruction, which should give us mostly precise exception handling behaviour at a performance cost. Exception handling behaviour is now completely controlled by the OpcodeDispatcher instead of being implicitly sprinkled throughout FEXCore.
* If/when we introduce formal clauses into the IR, this is the first step to get there. FlushRegisterCache would then be a "change clause" signal in the dispatcher.
* Not all context load/stores are tracked here. This could be improved, I let instcountci be my guide for what matters.
* Without the constant flushing, this gives SRA heartburn, especially around tied registers. This will be solved by the SRA rework that I have planned for clauses, but we need to ripoff the bandage first to get there because I don't want to port RCLSE to new IR data structures. Instead of tackle that now, this PR does extra flushing to make RCLSE.

 JIT time is reduced **5%** across the series. instcountci is a wash.


unittests/InstructionCountCI/FlagM/Secondary.json
```diff
@@ -1059,13 +1059,15 @@
...
     "cmpxchg rcx, rbx": {
-      "ExpectedInstructionCount": 5,
+      "ExpectedInstructionCount": 7,
       "ExpectedArm64ASM": [
         "eor w27, w4, w5",
         "subs x26, x4, x5",
         "cfinv",
-        "mov x4, x5",
-        "csel x5, x7, x5, eq"
+        "csel x20, x7, x5, eq",
```
Sonicadvance1, line 1067: Definitely something to glare at in a follow-up PR

Sonicadvance1 (approved):
I like the route this is going. Significantly less terrible worst case compared to previous attempts. I like that it is moving more towards a splatter style JIT.

PR #3766, opened 2024-06-26
Optimize vcvtps2ph

alyssarosenzweig:
We can avoid a LOT of pointless work with some dedicated IR ops for specifically
overriding the round mode.

Small behaviour change here: we no longer reset FTZ. I think this is a bug fix?
But if it's not it's not hard to fix.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/MiscOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/MiscOps.cpp
@@ -120,6 +120,39 @@ DEF_OP(SetRoundingMode) {
   msr(ARMEmitter::SystemRegister::FPCR, TMP1);
 }
 
+DEF_OP(PushRoundingMode) {
+  auto Op = IROp->C<IR::IROp_PushRoundingMode>();
+  auto Dest = GetReg(Node);
+
+  // Save the old rounding mode
+  mrs(Dest, ARMEmitter::SystemRegister::FPCR);
+
+  // vixl simulator doesn't support anything beyond ties-to-even rounding
+  if (!CTX->Config.DisableVixlIndirectCalls) [[unlikely]] {
+    return;
+  }
+
+  // Insert the rounding flags, reversing the mode bits as above
+  if (Op->RoundMode == 3) {
+    orr(ARMEmitter::Size::i64Bit, TMP1, Dest, 3 << 22);
+  } else if (Op->RoundMode == 0) {
+    and_(ARMEmitter::Size::i64Bit, TMP1, Dest, ~(3 << 22));
+  } else {
+    LOGMAN_THROW_AA_FMT(Op->RoundMode == 1 || Op->RoundMode == 2, "expect a valid round mode");
+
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
> We can avoid a LOT of pointless work with some dedicated IR ops for specifically overriding the round mode.
> 
> Small behaviour change here: we no longer reset FTZ. I think this is a bug fix? But if it's not it's not hard to fix.

Yes, Resetting FTZ (Or any other set FPCR state!) would have been a bug. Good catch.

Sonicadvance1 (approved):
10/10, would implement dedicated IR ops for things again.

FEXCore/Source/Interface/Core/JIT/Arm64/MiscOps.cpp
```diff
@@ -120,6 +120,37 @@ DEF_OP(SetRoundingMode) {
...
+    return;
+  }
+
+  // Insert the rounding flags, reversing the mode bits as above
+  if (Op->RoundMode == 3) {
+    orr(ARMEmitter::Size::i64Bit, TMP1, Dest, 3 << 22);
+  } else if (Op->RoundMode == 0) {
+    and_(ARMEmitter::Size::i64Bit, TMP1, Dest, ~(3 << 22));
+  } else {
+    and_(ARMEmitter::Size::i64Bit, TMP1, Dest, ~(Op->RoundMode << 22));
```
lioncash, line 141: Perhaps have an assert here ensuring the rounding mode is either 1 or 2, so it prevents this from carrying on in the event of invalid values being passed in (either through refactoring in the future, or using this with other things in the future)

alyssarosenzweig: Done. Good idea, thanks.

Sonicadvance1: Could also use the EmitValidation section in IR.json to ensure `RoundMode >= 0 && RoundMode <= 3`

FEXCore/Source/Interface/Core/JIT/Arm64/MiscOps.cpp
```diff
@@ -120,6 +120,39 @@ DEF_OP(SetRoundingMode) {
...
+    return;
+  }
+
+  // Insert the rounding flags, reversing the mode bits as above
+  if (Op->RoundMode == 3) {
+    orr(ARMEmitter::Size::i64Bit, TMP1, Dest, 3 << 22);
+  } else if (Op->RoundMode == 0) {
+    and_(ARMEmitter::Size::i64Bit, TMP1, Dest, ~(3 << 22));
+  } else {
+    LOGMAN_THROW_AA_FMT(Op->RoundMode == 1 || Op->RoundMode == 3, "expect a valid round mode");
```
lioncash, line 141:
```suggestion
    LOGMAN_THROW_AA_FMT(Op->RoundMode == 1 || Op->RoundMode == 2, "expect a valid round mode");
```

alyssarosenzweig: This is awkward... thx

Sonicadvance1 (approved):

PR #3767, opened 2024-06-26
AVX128: Fixes wide shifts

Sonicadvance1:
During refactoring this was missed and rerunning unittests locally caught it. 256-bit operations get their shift only from the lower half of the vector register.


lioncash (approved):

alyssarosenzweig (approved):
Whoops, sorry!

Sonicadvance1:
SNIPED

PR #3768, opened 2024-06-26
AVX128: Enable all the things

Sonicadvance1:
Here we are, at the summit of the mountain.


unittests/ASM/VEX/bextr2.asm
```diff
@@ -0,0 +1,23 @@
+%ifdef CONFIG
```
alyssarosenzweig, line 1: Drop this file, it duplicates the FEX_bug/ test I added for this today

Sonicadvance1: Oops, yep, added some ASM tests on accident.

unittests/InstructionCountCI/AVX128/VEX_map1.json
```diff
@@ -0,0 +1,6318 @@
+{
+  "Features": {
+    "Bitness": 64,
+    "EnabledHostFeatures": [],
+    "DisabledHostFeatures": [
+      "FCMA",
+      "RPRES",
+      "AFP",
+      "FLAGM",
+      "FLAGM2",
```
alyssarosenzweig, line 10: Could we enable flagm for the avx128 instcountci? It should save a few instructions on vptest.

(I've considered removing non-flagm instcountci. Even if we're not going to raise our minspec for FEX doesn't mean we need to have instcountci for it... But that's a tangential discussion to be had)

unittests/InstructionCountCI/AVX128/VEX_map1.json
```diff
@@ -0,0 +1,6318 @@
...
+        "Map 1 0b01 0x12 128-bit"
+      ],
+      "ExpectedArm64ASM": [
+        "mov w20, #0xde4",
+        "movk w20, #0x1, lsl #16",
+        "str x20, [x28, #272]",
+        "mov w1, #0x401",
+        "str x1, [x28, #1328]",
+        "ldr x0, [x28, #2296]",
+        "br x0"
```
alyssarosenzweig, line 353: I think something is very wrong here.

alyssarosenzweig: I think the instcountci commit needs to be interchanged with the "enable AVX" commit.

unittests/InstructionCountCI/AVX128/VEX_map1.json
Sonicadvance1: Added the same flagm map2 file that we had for SVE256, but this time for asimd

unittests/InstructionCountCI/AVX128/VEX_map1.json
Sonicadvance1: Oops, instcountci was disabling AVX if SVE256 was explicitly disabled. Weird interaction that no longer needs to exist.

unittests/InstructionCountCI/AVX128/VEX_map1.json
alyssarosenzweig: v(u)comiss notably missing

Sonicadvance1: You're right, that was in the other map file. Added that one as well.

Sonicadvance1:
Alright CI is green. OpenSSL is still angry with AVX overall but I hope that I can find the bug in their bignum library sooner rather than later.

alyssarosenzweig (approved):
Let's gooooooooo

unittests/InstructionCountCI/AVX128/VEX_map1.json
alyssarosenzweig: Looks good, thx

PR #3772, opened 2024-06-27
FEXCore: Fixes address size override on GPR sources and destinations

Sonicadvance1:
When the source or destination is a register, the address size override doesn't apply. We were accidentally applying it on all sources regardless of type which was causing us to zero extend on operations that aren't affected by address size override.

This fixes the OpenSSL cert error in every application, but most importantly Steam.


lioncash (approved):

alyssarosenzweig (approved):
Jeepers. I wonder if I introduced this bug when I did the address size stuff. But also, what?!?!?!

Gg.

PR #3774, opened 2024-06-27
FEXCore ARM64EC CI support

bylaws:
Dep on #3773, the toolchain artifact from https://github.com/bylaws/llvm-mingw/releases/download/20240628/llvm-mingw-20240627-ucrt-ubuntu-20.04-aarch64.tar.xz  will also need to be installed on the runner

### DIFF
--- a/.github/workflows/mingw_build.yml
+++ b/.github/workflows/mingw_build.yml
@@ -16,7 +16,7 @@ jobs:
     runs-on: ${{ matrix.arch }}
     strategy:
       matrix:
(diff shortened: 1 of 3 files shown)


bylaws:
https://github.com/bylaws/llvm-mingw/releases/download/20240630/llvm-mingw-20240630-ucrt-ubuntu-20.04-aarch64.tar.xz

should work with this new toolchain

Sonicadvance1 (approved):
Nice nice, runners updated

PR #3776, opened 2024-06-28
Frontend: Fixes invalid VSIB Index problem

Sonicadvance1:
In regular SIB land the index register encoding of 0b100 encodes to "no register", this feature lets you get SIB encodings without an index register for flexibility.

In VSIB encoding this isn't expected behaviour and instead there are no encodings where an index register is missing. Allowing you to encode all sixteen registers as an index register.

This was causing an abort in `AVX128_LoadVSIB` because the index turned in to an invalid register.

Working instruction:
`vgatherdps ymm2, dword [eax+ymm5*4], ymm7`

Broken instruction:
`vgatherdps ymm0, dword [eax+ymm4*4], ymm7`

This fixes a crash in libfmod where it is using gathers in the wild. Fixing a crash in Ender Lilies.


alyssarosenzweig:
can we get a unit test? thx

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -221,14 +221,16 @@ void Decoder::DecodeModRM_64(X86Tables::DecodedOperand* Operand, X86Tables::ModR
     {
       // If we have a VSIB byte (as opposed to SIB), then the index register is a vector.
       const bool IsIndexVector = (DecodeInst->TableInfo->Flags & InstFlags::FLAGS_VEX_VSIB) != 0;
+      uint8_t InvalidSIBIndex = 0b100; ///< SIB Index where there is no register encoding.
       if (IsIndexVector) {
         DecodeInst->Flags |= X86Tables::DecodeFlags::FLAG_VSIB_BYTE;
+        InvalidSIBIndex = 16; ///< No Invalid SIB Index with Index Vectors.
```
alyssarosenzweig, line 227: Maybe ~0 or -1 would be more to the point?

Sonicadvance1: Switched to ~0

Sonicadvance1:
> can we get a unit test? thx

Added a unittest

alyssarosenzweig (approved):

PR #3777, opened 2024-06-28
Add debug logging to server connections and config loads

pmatos:
I have found these messages to be useful to understand what's happening when debugging FEX, both to understand if a connection actually happened and to understand which config files are being used. Is it ok to merge it?

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -168,6 +168,7 @@ MainLoader::MainLoader(FEXCore::Config::LayerType Type, const char* ConfigFile)
   , Config {ConfigFile} {}
 
 void MainLoader::Load() {
+  LogMan::Msg::DFmt("[MainLoader] Loading config file: {}", Config.c_str());
   JSON::LoadJSonConfig(Config, [this](const char* Name, const char* ConfigString) { MapNameToOption(Name, ConfigString); });
 }
 
@@ -181,6 +182,7 @@ AppLoader::AppLoader(const fextl::string& Filename, FEXCore::Config::LayerType T
 }
 
 void AppLoader::Load() {
+  LogMan::Msg::DFmt("[AppLoader] Loading config file: {}", Config.c_str());
   JSON::LoadJSonConfig(Config, [this](const char* Name, const char* ConfigString) { MapNameToOption(Name, ConfigString); });
 }
 
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -182,7 +182,7 @@ AppLoader::AppLoader(const fextl::string& Filename, FEXCore::Config::LayerType T
 }
 
 void AppLoader::Load() {
-  LogMan::Msg::DFmt("[AppLoader] Loading config file: {}", Config.c_str());
+  LogMan::Msg::DFmt("[AppLoader] Loading config file: {}", Config);
   JSON::LoadJSonConfig(Config, [this](const char* Name, const char* ConfigString) { MapNameToOption(Name, ConfigString); });
 }
 
--- a/Source/Tools/FEXServer/ProcessPipe.cpp
+++ b/Source/Tools/FEXServer/ProcessPipe.cpp
@@ -449,6 +449,7 @@ void WaitForRequests() {
               int NewFD = accept(ServerSocketFD, reinterpret_cast<struct sockaddr*>(&Addr), &AddrSize);
 
               // Add the new client to the temporary array
+              LogMan::Msg::DFmt("[FEXServer] New connection accepted");
               NewPollFDs.emplace_back(pollfd {
                 .fd = NewFD,
                 .events = POLLIN | POLLPRI | POLLRDHUP,


alyssarosenzweig (approved):

alyssarosenzweig:
Seems reasonable to me at least

Source/Common/Config.cpp
```diff
@@ -168,6 +168,7 @@ MainLoader::MainLoader(FEXCore::Config::LayerType Type, const char* ConfigFile)
   , Config {ConfigFile} {}
 
 void MainLoader::Load() {
+  LogMan::Msg::DFmt("[MainLoader] Loading config file: {}", Config.c_str());
```
Sonicadvance1, line 171: This message goes nowhere at application launch since the LogManager handlers aren't setup at the start so its outputFD ends up being -1 so the writes do EBADFD. Same as the apploader message below.

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -449,6 +449,7 @@ void WaitForRequests() {
               int NewFD = accept(ServerSocketFD, reinterpret_cast<struct sockaddr*>(&Addr), &AddrSize);
 
               // Add the new client to the temporary array
+              LogMan::Msg::DFmt("[FEXServer] New connection accepted");
```
Sonicadvance1, line 452: This gets /really/ chatty when running any real workloads and will spam hundreds of messages every time pressure-vessel launches plus hundreds more each time dxsetup runs. We have configurations running FEXServer with its output log going to the journal so we want to be mindful of logging.

I am going to need to nak from this line alone, but if the other two lines were actually getting their logs out then I would do it thrice over.

alyssarosenzweig (changes requested):
per Ryan

Source/Common/Config.cpp
```diff
@@ -181,6 +182,7 @@ AppLoader::AppLoader(const fextl::string& Filename, FEXCore::Config::LayerType T
 }
 
 void AppLoader::Load() {
+  LogMan::Msg::DFmt("[AppLoader] Loading config file: {}", Config.c_str());
```
neobrain, line 185: fmt can format strings, so there's no need for `.c_str()`:
```suggestion
  LogMan::Msg::DFmt("[AppLoader] Loading config file: {}", Config);
```

Source/Common/Config.cpp
pmatos: I am slightly confused about this. I am probably missing something but I can see these messages being output to server. There are no errors. In which case is the LogManager not set up when we load the configs?

Source/Tools/FEXServer/ProcessPipe.cpp
pmatos: Since this is a DFmt, doesn't it mean that you need to enable Debug messaging for it to actually spam the logs? I would expect that for normal runs, the log level would be info or something like that. I don't think Debug logging in general should be enabled for production executation.

Sonicadvance1: Log levels don't mean anything in FEX's context atm, they were more a qualifier for the type of log than anything else atm. FEXInterpreter will write them all, and FEXServer will display them all regardless.

Source/Common/Config.cpp
Sonicadvance1: Interesting, shouldn't do. The default logmanager handler is setup by time this loads but it defaults to SilentLog (through uninitialized bool but eh) and default output file of -1. stracing showed these trying to write to -1 locally

Installer getting installed:
https://github.com/FEX-Emu/FEX/blob/main/Source/Tools/FEXLoader/FEXLoader.cpp#L275
LogManager getting switched over to Server significantly later:
https://github.com/FEX-Emu/FEX/blob/main/Source/Tools/FEXLoader/FEXLoader.cpp#L346

Source/Tools/FEXServer/ProcessPipe.cpp
pmatos: Ufff - I didn't know that. :-/
That's unfortunate. So, I guess the issue is that we always set the log level to debug? We sometimes do use IFmt and other times DFmt...

Source/Common/Config.cpp
pmatos: OK, maybe we are doing things differently... I start a FEXServer with `-f -p`. Then I start FEX and get in the FEXServer window:
```
[DEBUG] [MainLoader] Loading config file: /FEX/FEXDeb-d8c2fcb77/share/fex-emu/Config.json
[DEBUG] [MainLoader] Loading config file: /home/steamvr/.fex-emu/Config.json
[DEBUG] [FEXServer] New connection accepted
[INFO][436755.473149275][242833.242833] [242833][steam-launch-wrapper] Sleeping for 5 seconds
[INFO][436760.623612865][242833.242833] IR-pre 0x7f9c2e3de0:
```

which... is quite useful tbh.

Sonicadvance1: We would need to change how logging occurs if we want to block debug message on release builds, since a potential for 3 messages in the server log when pressure-vessel and proton spin up hundreds of processes isn't viable.

pmatos: OK - lets close this for now then.

PR #3778, opened 2024-06-28
Largest x87 blocks of code from games

pmatos:
These are largest blocks of x87 code for some of game benchmarks we know of. They are not necessarily hot blocks but should give us an overview how changes in x87 optimizations affect the generated code.


alyssarosenzweig (approved):

Sonicadvance1 (approved):
whew, them some spicy blocks

Joshua-Ashton:
Oblivion mention :raised_hands:

PR #3803, opened 2024-07-02
Use number of jobs as defined by TEST_JOB_COUNT

pmatos:
At the moment we always run ctest with max number of cpus. If
undefined, it will keep current behaviour, otherwise it will
honour TEST_JOB_COUNT.

Therefore to run ctest one test at a time, use
`cmake ... -DTEST_JOB_COUNT=1`


pmatos:
Thanks to @neobrain for the discussion on this.

CMakeLists.txt
```diff
@@ -368,6 +368,15 @@ if (BUILD_TESTS)
   include(CTest)
   enable_testing()
   message(STATUS "Unit tests are enabled")
+
+  # If CTEST_PARALLEL_LEVEL is not defined, lets define it to ncpus to keep current behavior
+  execute_process(COMMAND "nproc" OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE CORES)
+  if (NOT TEST_JOB_COUNT)
+    set(TEST_JOB_COUNT ${CORES})
+  endif()
```
neobrain, line 376: Promoting `TEST_JOB_COUNT` to a CACHE variable will make it show up in CMakeCache.txt and allow IDEs to display a GUI for it:

```cmake
set (TEST_JOB_COUNT "" CACHE STRING "Override number of parallel jobs to use while running tests")
if (NOT TEST_JOB_COUNT)
  execute_process(COMMAND "nproc" OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE TEST_JOB_COUNT)
endif()
```

CMakeLists.txt
```diff
@@ -368,6 +368,15 @@ if (BUILD_TESTS)
   include(CTest)
   enable_testing()
   message(STATUS "Unit tests are enabled")
+
+  # If CTEST_PARALLEL_LEVEL is not defined, lets define it to ncpus to keep current behavior
```
neobrain, line 372: Outdated comment? `CTEST_PARALLEL_LEVEL` doesn't seem to be involved anymore.

CMakeLists.txt
```diff
@@ -368,6 +368,15 @@ if (BUILD_TESTS)
...
+  execute_process(COMMAND "nproc" OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE CORES)
+  if (NOT TEST_JOB_COUNT)
+    set(TEST_JOB_COUNT ${CORES})
+  endif()
+  message(STATUS "Running tests with ${TEST_JOB_COUNT} jobs")
+  # Set TEST_JOB_FLAG to -j$TEST_JOB_COUNT
+  set(TEST_JOB_FLAG "-j${TEST_JOB_COUNT}")
```
neobrain, line 379: Any reason for not wrapping this in `if (${CMAKE_VERSION} VERSION_LESS "3.29")` to enable using `CTEST_PARALLEL_LEVEL` on newer CMake versions?

The difference doesn't matter much in our current setup, but technically the build machine could be a different one than the one running ctest, which `nproc` cannot predict.

pmatos: What's the advantage of `CTEST_PARALLEL_LEVEL`?

neobrain: `CTEST_PARALLEL_LEVEL` isn't the important part, but rather it's to make `ctest` infer the core count itself when `TEST_JOB_COUNT` is not set. Alternatively, we could default `TEST_JOB_COUNT` to 0 for CMake 3.29 instead of calling `nproc`.

Having `ctest` infer parallelization itself is better than guessing the core count at generation-time, since the latter can produce incorrect results (e.g. if the tests run on a different machine than the build host, or if hardware is upgraded). Neither of these are critical things to be concerned with, but it's equally simple to add a version check to address them.

neobrain: ... indeed, if we leave `TEST_JOB_COUNT` untouched on CMake 3.29+, `TEST_JOB_FLAG` will end up as `-j`, which would also have the desired behavior. We just need to change `else()` to `elif(version check)` for this.

Sonicadvance1:
The CI runners need to be updated to take advantage of this new flag, or to pass in -j <whatever> because this is a CI time worsening atm. Hostrunner running the asm tests takes 37 seconds in this PR versus 6 seconds on main and of course other CI steps are affected as well.
Like vixl sim taking 3m21s on this PR and 44s on main.

CMakeLists.txt
pmatos: Thanks!

pmatos:
> The CI runners need to be updated to take advantage of this new flag, or to pass in -j because this is a CI time worsening atm. Hostrunner running the asm tests takes 37 seconds in this PR versus 6 seconds on main and of course other CI steps are affected as well. Like vixl sim taking 3m21s on this PR and 44s on main.



> The CI runners need to be updated to take advantage of this new flag, or to pass in -j because this is a CI time worsening atm. Hostrunner running the asm tests takes 37 seconds in this PR versus 6 seconds on main and of course other CI steps are affected as well. Like vixl sim taking 3m21s on this PR and 44s on main.

Wait - I need to check this cause this shouldn't affect CI. The whole point of my dance with @neobrain  here is that we are backwards compatible.

pmatos:
> The CI runners need to be updated to take advantage of this new flag, or to pass in -j because this is a CI time worsening atm. Hostrunner running the asm tests takes 37 seconds in this PR versus 6 seconds on main and of course other CI steps are affected as well. Like vixl sim taking 3m21s on this PR and 44s on main.

@Sonicadvance1 OK - CI time worsening is gone. I missed a couple of CMake rules earlier on.

Sonicadvance1 (approved):
i'm now fine with this change

CMakeLists.txt
```diff
@@ -368,6 +368,16 @@ if (BUILD_TESTS)
   include(CTest)
   enable_testing()
   message(STATUS "Unit tests are enabled")
+
+  set (TEST_JOB_COUNT "" CACHE STRING "Override number of parallel jobs to use while running tests")
+  if (TEST_JOB_COUNT)
+    message(STATUS "Running tests with ${TEST_JOB_COUNT} jobs")
+  endif()
+  if (CMAKE_VERSION VERSION_LESS "3.29")
```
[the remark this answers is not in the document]
pmatos: I don't think so... see: https://cmake.org/cmake/help/v3.29/command/if.html#variable-expansion

Although I am the last person to be confident about cmake file writing. :)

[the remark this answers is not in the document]
neobrain: Huh, TIL. Looks fine then :)

CMakeLists.txt
```diff
@@ -368,6 +368,16 @@ if (BUILD_TESTS)
...
+    execute_process(COMMAND "nproc" OUTPUT_STRIP_TRAILING_WHITESPACE OUTPUT_VARIABLE TEST_JOB_COUNT)
+  endif()
+  # Set TEST_JOB_FLAG to -j$TEST_JOB_COUNT
```
neobrain, line 379: This comment seems very redundant.

neobrain (approved):
Added one more comment, looks good otherwise!

pmatos:
> Added one more comment, looks good otherwise!

Thanks. 👍🏼 
Fixed.

PR #3804, opened 2024-07-02
Revert removing RCLSE

Sonicadvance1:
Keeping this ready if I can't solve the issue. There's some sort of gnarly interaction here that breaks [Peggle Deluxe](https://store.steampowered.com/app/3480/Peggle_Deluxe/).
Might just be an interaction between MMX and X87 register states causing confusion, but if I can't nail down the exact issue then it'll need a revert.

At the very least fxsave/xsave fxrstor/rstor is broken because it is only saving the bottom 64-bits of the MMX/x87 registers. Which I didn't catch in instcountci before.

To repro in Peggle Deluxe, start a level and the fire the ball, it should immediately crash.

I'm also tracking an issue with Portal 2 which may be related but I'm still testing.

Keeping this as a draft until more investigation can be done.


alyssarosenzweig:
Pity.

alyssarosenzweig (approved):

PR #3808, opened 2024-07-03
Try to delete RCLSE again

alyssarosenzweig:


unittests/InstructionCountCI/Secondary_REP.json
```diff
@@ -612,7 +612,7 @@
       "ExpectedInstructionCount": 1,
       "Comment": "0xf3 0x0f 0xd6",
       "ExpectedArm64ASM": [
-        "ldr d16, [x28, #1040]"
+        "ldr q16, [x28, #1040]"
```
Sonicadvance1, line 615: This is incorrect behaviour now and we don't have a unit test to ensure correct behaviour

alyssarosenzweig: It should be bug for bug compatible though, filed #3850 about the preexisting MMX broken

Sonicadvance1: This one isn't as it will move broken x87 data in to the XMM register in this instance, where before it correctly zero extended the lower 64-bits.

alyssarosenzweig: Or maybe not? I'm kinda confused tbh

Sonicadvance1: Here's a test case to show what goes wrong.
```asm
%ifdef CONFIG
{
  "RegData": {
    "XMM0": ["0x5152535455565758", "0"]
  }
}
%endif

fninit
; Load all test values
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]
fld tword [rel .test_value]

; Setup for MMX usage
emms

; Load XMM value
movups xmm0, [rel .test_xmm_value]

; Load MMX value
movq mm0, [rel .test_mmx_value]

jmp .test
.test:
; Move MMX register in to XMM
; Should set the upper 64-bits of xmm0 to zero
movq2dq xmm0, mm0

hlt
align 32

.test_value:
dq 0x4142434445464748
dw 0x7fff

.test_mmx_value:
dq 0x5152535455565758

.test_xmm_value:
dq 0x6162636465666768
dq 0x7172737475767778
```

alyssarosenzweig: Thanks for the unit test, repro'd, indeed this is more MMX brokenness. Uff. Not really sure how to address this without more perf hits. Wondering if I need to go back to the drawing board with MMX :sob:

Sonicadvance1: This one is more reasonable to eat the `_VMov(OpSize::i64Bit` because it is actually doing a MMX->SSE move. Just kind of silly that it does an ldr+mov instead of ldr but is a whatever burger

Sonicadvance1 (approved):
Okay with the latest changes I'm not feeling significantly better about it not breaking the world. And if we do happen to find breakage it'll likely be easier to fix rather than revert this time around.

PR #3811, opened 2024-07-04
RA: fix interaction between SRA & shuffles

alyssarosenzweig:


Sonicadvance1 (approved):
Nice. Fixed the issue I ran into.

PR #3812, opened 2024-07-04
OpcodeDispatcher: Fixes rotates with zero not zero extending 32-bit result

Sonicadvance1:
For all the 32-bit rotates (except for RORX) we were failing to zero
extend the 32-bit result to the destination register when the rotate was
masked to zero.

Ensure we do this.


FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -2100,15 +2100,21 @@ class OpDispatchBuilder final : public IREmitter {
...
+  template<typename F, typename FF>
+  void Calculate_ShiftVariable(Ref Shift, F&& Calculate, FF&& ZeroShiftResult, bool NeedsZeroShiftResult = true) {
     // RCR can call this with constants, so handle that without branching.
     uint64_t Const;
     if (IsValueConstant(WrapNode(Shift), &Const)) {
       if (Const) {
         Calculate();
       }
 
+      if (Const == 0 && NeedsZeroShiftResult) {
```
alyssarosenzweig, line 2114: `} else if (NeedsZeroShiftResult) {`

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -2100,15 +2100,21 @@ class OpDispatchBuilder final : public IREmitter {
     return CurrentDeferredFlags.Type == FlagsGenerationType::TYPE_NONE;
   }
 
-  template<typename F>
-  void Calculate_ShiftVariable(Ref Shift, F&& Calculate) {
+  void ZeroShiftResult() {}
+
+  template<typename F, typename FF>
+  void Calculate_ShiftVariable(Ref Shift, F&& Calculate, FF&& ZeroShiftResult, bool NeedsZeroShiftResult = true) {
```
alyssarosenzweig, line 2106: Why not make ZeroShiftResult nullable, calling it if present and ignoring it if absent? Then the boolean flag goes away and the two can't go out of sync. And we also don't need the DoNothing thunk below

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2285,10 +2310,20 @@ void OpDispatchBuilder::RCLOp(OpcodeArgs) {
   Ref Src = LoadSource(GPRClass, Op, Op->Src[1], Op->Flags, {.AllowUpperGarbage = true});
   const auto OpSize = OpSizeFromSrc(Op);
 
+  auto CalculateZeroResult = [this, Op]() {
```
alyssarosenzweig, line 2313: Can we promote this to `OpDispatchBuilder::` space so we don't duplicate it across each rotate instruction?

Sonicadvance1: promoted

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
Sonicadvance1: nullable indeed

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -2355,7 +2365,8 @@ void OpDispatchBuilder::RCLOp(OpcodeArgs) {
     SetRFLAG<FEXCore::X86State::RFLAG_OF_RAW_LOC>(NewOF, Size - 1, true);
 
     StoreResult(GPRClass, Op, Res, -1);
-  });
+    },
+    GetSrcSize(Op) == OpSize::i32Bit ? std::make_optional(&OpDispatchBuilder::ZeroShiftResult) : std::nullopt);
```
alyssarosenzweig, line 2369: ZeroShiftResult checks the src size anyway, so I think we can drop the ternary no?

alyssarosenzweig (changes requested):
per discord - I'd like a chance to try an alternate approach first, let's hold off on merge for now

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
Sonicadvance1: The issue comes down to the non-const shift case generating an empty block for !32-bit cases in that case, so it needs to be passed nothing to avoid that.

alyssarosenzweig: Right ok

alyssarosenzweig (approved):
Doing this more efficiently is nontrivial because of the RA requirements. given how rare rotates are, I don't really mind the changes here. merging, thanks

PR #3814, opened 2024-07-04
unittests: Adds test for flags when shifting by zero

Sonicadvance1:


alyssarosenzweig (approved):

Sonicadvance1:
This will fail CI until the instruction is actually fixed of course

alyssarosenzweig:
#3818

PR #3815, opened 2024-07-04
unittests: Adds rotate with carry test for large rotates

Sonicadvance1:
FEX-Emu currently doesn't do large rotates for small data sources correctly. This will fail CI until fixed in OpcodeDispatcher


alyssarosenzweig:
#3819

PR #3816, opened 2024-07-04
Arm64: Fixes long signed divide 

Sonicadvance1:
The two halves are provided as two uint64_t values that shouldn't be
sign extended between them. Treat them as uint64_t until combined in to
a single int128_t. Fixes long signed divide.


lioncash (approved):
nice catch

PR #3821, opened 2024-07-05
unittests: Adds a test for small shift flags calculation

Sonicadvance1:
Currently we calculate CF incorrectly in the case of small shifts with large offsets.

Will fail CI until fixed


alyssarosenzweig:
#3823

PR #3822, opened 2024-07-05
unittests: More rotate with carry unit tests

Sonicadvance1:
Looks like we missed some edge cases with small carry rotate. Adds even more unit tests.

This will fail CI until fixed


alyssarosenzweig:
#3824

PR #3823, opened 2024-07-05
Fix CF with small shifts

alyssarosenzweig:


Sonicadvance1 (approved):
pfft, I wasn't expecting an optimization to come out of fixing this bug. That's kind of fun.

alyssarosenzweig:
> pfft, I wasn't expecting an optimization to come out of fixing this bug. That's kind of fun.

:cowboy_hat_face:

PR #3824, opened 2024-07-05
OpcodeDispatcher: Fix 8/16-bit rcr masking

alyssarosenzweig:


unittests/InstructionCountCI/FlagM/PrimaryGroup.json
```diff
@@ -1602,23 +1602,24 @@
...
-        "bfi x21, x21, #18, #18",
-        "bfi x21, x21, #36, #9",
-        "lsr x20, x21, x5",
-        "bfxil x4, x20, #0, #8",
-        "sub w22, w5, #0x1 (1)",
-        "lsr w21, w21, w22",
-        "rmif x21, #63, #nzCv",
-        "eor w20, w20, w20, lsr #1",
+        "cbz x20, #+0x3c",
+        "and w20, w5, #0x1f",
```
Sonicadvance1, line 1610: Is this and actually needed? We also and above, we could reuse the previous and and save and instruction. Looks like before the small sizes were actually just incorrectly using w5 directly instead of w20 from before.

Sonicadvance1: Or does this fall under the issue that only SRA is allowed to be live across a block edge?

alyssarosenzweig: >  Or does this fall under the issue that only SRA is allowed to be live across a block edge?

It's exactly that. There are ways we could work around this, but given how much time I wasted optimizing obscure rotate-with-carry optimizations with little ROI ... I'm not inclined to spend more time on these demon instructions beyond what's needed to pass TCG.

Sonicadvance1: Sounds reasonable to me

Sonicadvance1 (approved):
Nice trivial fix

Issue #3806, opened 2024-07-03
AVX128: Extend 32-bit indices in the case of `VPGATHERDQ ymm1, vm32x, ymm2`


Sonicadvance1:
For the particular code path that `VPGATHERDQ ymm1, vm32x, ymm2` hits, we can sign-extend the indices provided in vm32x to be 64-bit indices.
This only really works because we split the IR operation in to two halves anyway, so 4 32-bit indices in vm32x loading 4 64-bit indices in ymm1 works out to two SVE gather loads using 64-bit indices quite well.

We can also then use the scaling optimization that #3805 proposes since the 32-bit signed indices are guaranteed to overflow correctly in this case.

PR #3826, opened 2024-07-06
AVX128: Extend 32-bit address indices when possible 

Sonicadvance1:
When loading 256-bits of data with only 128-bits of address indices, we
can sign extend the source indices to be 64-bit. Thus falling down the
ideal path for SVE where each 128-bit lane is loading the data to
addresses in a 1:1 element ratio.

This means we use the SVE path more often because of this.

Based on top of https://github.com/FEX-Emu/FEX/pull/3825 because the prescaling behaviour was introduced
there. This implements its own prescaling when the sign extension occurs
because ARM's SSHLL{,2} instruction gives us that for free.

This additionally fixes a bug where we were accidentally loading the top
128-bit half of the addresses for gathers when it was unnecessary, and
on the AVX256 side it was duplicating and doing some additional work
when it shouldn't have.

It'll be good to walk the commits when looking at this one, as there are
a couple of incremental changes that are easier to follow that way.

Fixes https://github.com/FEX-Emu/FEX/issues/3806


lioncash (approved):
Very nice

Issue #3827, opened 2024-07-06
AVX128: `vgatherqps` and `vpgatherqd` might be able to be optimized with `LD1W { <Zt>.D }, <Pg>/Z, [<Xn|SP>, <Zm>.D]`


Sonicadvance1:
Currently these operations fall down the ASIMD path universally. We might be able to use this weirdo zero extending gather from SVE and then swizzle the components down to where we want them instead.

Need to noodle on it a bit.

PR #3828, opened 2024-07-06
AVX128: Optimize QPS/QD variant of gather loads! 

Sonicadvance1:
SVE has a special version of their gather instruction that gets similar
behaviour to x86's VGATHERQPS/VPGATHERQD instructions.

The quirk of these instructions that the previous SVE implementation
didn't handle and required ASIMD fallback, was that most gather
instructions require the data element size and address element size to
match. This x86 instruction uses a 64-bit address size while loading 32-bit
elements. This matches this specific variant of the SVE instruction, but
the data is zero-extended once loaded, requiring us to shuffle the data
after it is loaded.

This isn't the worst but the implementation is different enough that
stuffing it in to the other gather load will cause headaches.

Basically gets 32 instruction variants to use the SVE version!

Fixes https://github.com/FEX-Emu/FEX/issues/3827

Based on #3826 until that is merged.


Sonicadvance1:
Rebased for your viewing pleasure

lioncash (approved):

PR #3831, opened 2024-07-06
Path resolution fixes

asahilina:
* FEXLoader: Strip rootfs path from CONFIG_APP_FILENAME
* FileManagement: Trace symlink paths across rootfs

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -536,6 +536,11 @@ uint64_t FileManager::Close(int fd) {
   }
 #endif
 
+  if (fd == RootFSFD) {
+    LogMan::Msg::EFmt("Process attempting to close RootFS FD, ignoring.");
+    return 0;
+  }
+
   return ::close(fd);
 }
 
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -330,40 +330,36 @@ FileManager::~FileManager() {
 }
 
 fextl::string FileManager::GetEmulatedPath(const char* pathname, bool FollowSymlink) {
-  if (!pathname ||                  // If no pathname
-      pathname[0] != '/' ||         // If relative
-      strcmp(pathname, "/") == 0) { // If we are getting root
-    return {};
-  }
 
-  auto thunkOverlay = ThunkOverlays.find(pathname);
-  if (thunkOverlay != ThunkOverlays.end()) {
-    return thunkOverlay->second;
-  }
+  FDPathTmpData TmpFilename;
 
   const auto& RootFSPath = LDPath();
   if (RootFSPath.empty()) { // If RootFS doesn't exist
     return {};
   }
 
-  fextl::string Path = RootFSPath + pathname;
-  if (FollowSymlink) {
-    char Filename[PATH_MAX];
-    while (FEX::HLE::IsSymlink(AT_FDCWD, Path.c_str())) {
-      auto SymlinkSize = FEX::HLE::GetSymlink(AT_FDCWD, Path.c_str(), Filename, PATH_MAX - 1);
-      if (SymlinkSize > 0 && Filename[0] == '/') {
-        Path = RootFSPath;
-        Path += std::string_view(Filename, SymlinkSize);
-      } else {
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
:+1:

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -356,41 +356,37 @@ FileManager::~FileManager() {
...
   constexpr auto NoEntry = std::make_pair(-1, nullptr);
 
+  // The two temporary paths.
+  const std::array<char*, 2> TmpPaths = {
+    TmpFilename[0],
+    TmpFilename[1],
+  };
+
+  // Current index for the temporary path to use.
+  uint32_t CurrentIndex {};
```
andy9a9, line 388: What's the value here? Shouldn't be `= 0` used instead?

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -356,41 +356,37 @@ FileManager::~FileManager() {
...
-    while (FEX::HLE::IsSymlink(AT_FDCWD, Path.c_str())) {
-      auto SymlinkSize = FEX::HLE::GetSymlink(AT_FDCWD, Path.c_str(), Filename, PATH_MAX - 1);
-      if (SymlinkSize > 0 && Filename[0] == '/') {
-        Path = RootFSPath;
-        Path += std::string_view(Filename, SymlinkSize);
-      } else {
-        break;
-      }
-    }
+  auto FdPath = GetEmulatedFDPath(AT_FDCWD, pathname, FollowSymlink, TmpFilename);
```
andy9a9, line 367: Maybe `FdPath` and `Path` shouldn't be declared while their values aren't used :).

andy9a9 (commented):
Great work!

asahilina:
@Sonicadvance1 Note that [d9a8309](https://github.com/Sonicadvance1/FEX/commit/d9a83092c768164fc5702713ddf44664c532f4fb) is not enough to fix wine, you still need b483c800 here since it's needed to make the /etc/alternatives stuff work with wine's path resolution.

Sonicadvance1:
> @Sonicadvance1 Note that [d9a8309](https://github.com/Sonicadvance1/FEX/commit/d9a83092c768164fc5702713ddf44664c532f4fb) is not enough to fix wine, you still need [b483c80](https://github.com/FEX-Emu/FEX/commit/b483c800f485900a1094d3c37dc4683e1c430589) here since it's needed to make the /etc/alternatives stuff work with wine's path resolution.

Indeed, my PR is only a partial fix. I still need to work through the rest of the problems with rootfs symlink problem.

asahilina:
Rebased this on top of your change and fixed a bug that broke Steam

asahilina:
Added another patch to refuse to close the RootFS FD. This was already a hidden potential problem, but more specifically regresses Proton with the change to `GetEmulatedPath` to call `GetEmulatedFDPath`, since that uses the fd where it wasn't previously used.

asahilina:
Fixed the formatting (I think)

Sonicadvance1:
> Fixed the formatting (I think)

fyi, this command can reformat the project for you `sh -c "CLANG_FORMAT=clang-format-16 ./Scripts/reformat.sh . > /dev/null 2>&1"`

alyssarosenzweig:
This seems plausible to me but @Sonicadvance1 should probably review

neobrain:
What's the path (heh) forward for this? Does a proper fix need more looking into or are we considering integrating this PR as-is? I have reservations against the latter, but not sure they need to be elaborated.

asahilina:
My main worry about this is the performance implications of doing a lot more syscalls for some things, but I don't know that that is solvable without a major overhaul introducing caching or something like that...

What I was going to explore next is actually completely bypassing this whole mess and introducing an alternate mode where FEX *only* does RootFS lookups, using a kernel-side overlayfs and overlaid mounts to create the complete x86 view of the filesystem. I suspect that'll work better for our (muvm) use case and solve a large number of path resolution/rootfs related issues in FEX all at once (there are just about enough tricks with path resolution and POSIX/Linux syscalls available to make path resolution correctly implementable in this scenario, without any userspace path resolution at the component level).

But that only works for that setup. Assuming there isn't interest in moving to that solution fully as a baseline, I imagine we still want to improve the internal overlay logic as in this PR. And I think for that I need some FEX folks to weigh in, since I don't really know where to go from here.

PR #3832, opened 2024-07-06
WOW64: Mark the FEX dll as a wine builtin

bylaws:
Allows it to be automatically picked up by wine during prefix setup, without a manual dll override.

Thanks to AndreRH for pointing me to this.


alyssarosenzweig (approved):
utterly magic

Sonicadvance1 (approved):
Image_2024-07-06_17-55-54
A quick pick for showing what's in the random binary. Kind of wacky this magic.

PR #3834, opened 2024-07-06
Windows: Report as an AMD64 processor when targeting ARM64EC

bylaws:


Source/Windows/Common/CPUFeatures.cpp
```diff
@@ -5,7 +5,11 @@
 
 namespace FEX::Windows {
 CPUFeatures::CPUFeatures(FEXCore::Context::Context& CTX) {
+#ifdef _M_ARM_64EC
```
lioncash, line 8: Maybe leave a comment that explains why the distinction matters here, just as a helping aide. e.g. I don't really know the specifics of ARM64EC, so it's not really clear why this is necessary to me

PR #3835, opened 2024-07-06
AllocatorHooks: Allocate from the top down on windows

bylaws:
FEX allocations can get in the way of allocations that are 4gb-limited even in 64-bit mode (i.e. those from LuaJIT), so allocate starting from the top of the AS to prevent conflicts.


alyssarosenzweig (approved):

alyssarosenzweig:
approved, but this probably warrants a comment

Sonicadvance1 (approved):
Very cute. I'm surprised that FEX VA is intersecting with guest VA but I guess just a quirk of arm64ec currently.

Issue #3829, opened 2024-07-06
AVX128: vpgatherdd/vgatherdps asimd fallbacks should be able to utilize SVE


Sonicadvance1:
The versions of vpgatherdd/vgatherdps that fallback to the ASIMD variant should be able to use the QPS variant of the gather load introduced in #3826.

Gets a bit spicy with the 256-bit destination and 256-bit source versions. I believe we need to sign extend the addr elements to 64-bit, effectively generating 512-bits of address vectors. So it won't be as efficient as the other versions but a heck of a lot better than ASIMD fallback.

PR #3836, opened 2024-07-07
AVX128: Optimize the vpgatherdd/vgatherdps cases that would fall back to ASIMD

Sonicadvance1:
With the introduction of the wide gathers in https://github.com/FEX-Emu/FEX/pull/3828 this has opened new
avenues for optimizing these cases that would typically fall back to
ASIMD. In the cases that 32-bit SVE scaling doesn't fit, we can instead
sign extend the elements in to double-width address registers.

This then feeds naturally in to the SVE path even though we end up
needing to allocate 512-bits worth of address registers. This ends up
being significantly better than the ASIMD path still.

Relies on https://github.com/FEX-Emu/FEX/pull/3828 to be merged first
Fixes https://github.com/FEX-Emu/FEX/issues/3829


lioncash (approved):

Sonicadvance1:
Rebased

lioncash (approved):

PR #3837, opened 2024-07-07
AVX128: Extends 32-bit indexes path for 128-bit operations 

Sonicadvance1:
The codepath from https://github.com/FEX-Emu/FEX/pull/3826 was only targeting 256-bit sized operations.
This missed the vpgatherdq/vgatherdpd 128-bit operations. By extending
the codepath to understand 128-bit operations, we now hit these
instruction variants.

With this PR, we now have SVE128 codepaths that handle ALL variants of
x86 gather instructions! There are zero ASIMD fallbacks used in this
case!

Of course depending on the instruction, the performance still leaves a
lot to be desired, and there is no way to emulate x86 TSO behaviour
without an ASIMD fallback, which we will likely need to add as a
fallback at some point.

Based on https://github.com/FEX-Emu/FEX/pull/3836 until that is merged.


Sonicadvance1:
Easy rebase for the final gather support

lioncash (approved):

Issue #2995, opened 2023-08-24
cvttpd2dq is probably broken


dougallj:
According to InstCountCI,
```
cvttpd2dq xmm0, xmm1
```
Translates to:
```
fcvtn v4.2s, v17.2d
fcvtzs v16.4s, v4.4s
```

`fcvtn` converts from a double to a float, which is a lossy operation. For example, the double 100000001.0 turns into the float 100000000.0f. (These are safely within the integer range, so the result will be the integer 100000000 instead of 100000001.)

I suspect the tests are inadequate, mostly testing small, positive numbers: https://github.com/FEX-Emu/FEX/blob/c5f358b47adfc7afe00f7aeeddbd8fb28b5f601c/unittests/ASM/VEX/vcvttpd2dq.asm#L47-L55

As well as a value that is too small to see the problem (0x4142434445464748 -> 2393736.541207228), and a value that is too large (0x5152535455565758 -> 5.562560877255032e+83) (and -1.0 and -2.0):

https://github.com/FEX-Emu/FEX/blob/c5f358b47adfc7afe00f7aeeddbd8fb28b5f601c/unittests/ASM/OpSize/66_E6.asm#L25-L28

I'd recommend testing with positive and negative powers of two (2^0 through 2^33 for a 32-bit operation like this) ± {1, 0.5, 0.25, 0.75} for tests, and adding zero, positive and negative infinity, and NaN:

* 2**30±1 covers loss of precision
* 2**n±0.5 ensures that numbers half-way between integers are rounding correctly
* 2**n±0.25/0.75 test for rounding up/down/nearest.
* Going up to 2**33 tests the out-of-range result.
* Including negative numbers tests negative-out-of-range, and negative-rounding (e.g. rounding towards zero vs towards negative-infinity are indistinguishable when testing only positive numbers).
* Infinity, NaN and zero are common floating point special cases.

Python code to generate these:
```python
import struct

l = set()
for i in range(34):
	for j in [0, 1, 0.5, 0.25, 0.75]:
		l.add(float(2**i+j))
		l.add(float(2**i-j))

for i in list(l):
	l.add(-i)

for i in [float('-inf')] + sorted(l) + [float('inf'), float('nan')]:
	print('dq 0x%016X ; %r' % (struct.unpack('<Q', struct.pack('<d',i))[0], i))
```

PR #3842, opened 2024-07-08
VCVT{T,}PD2DQ fixes and optimization

Sonicadvance1:
A few things here.

- Fixes a bug with VCVT{T,}PD2DQ where it was failing to zero the upper 128-bit lane in AVX128
  - Updating the unittests to ensure it happens
- Adds a new IR path to take advantage of new SVE instructions which can directly F64->I32
- Fixes a /minor/ precision issue reported in #2995 due to truncating F64->F32 before rounding which could have resulting in rounding the wrong direction.

This doesn't change behaviour around overflow problems with conversion. If we want to handle that then we need to enable FP exceptions and capture the overflow and continue running. Which most ARM hardware doesn't support those exceptions anyway. x86 always returns its "indefinite integer" value of 0x8000_0000, where ARM will return the saturated minimum or maximum int32_t depending on sign.


JosJuice:
I think you still have a rounding problem in the non-SVE path. Using FRINTI before FCVTN does ensure that the value sent to FCVTN is an integer, but there's a possibility that it's an integer that can't be represented exactly in f32 but can be represented in i32. You have to avoid intermediate conversion to f32 in order to solve this, and as far as I know, the best way to do that in ASIMD is to use FCVTZS (scalar, integer) in place of FCVTN (double to single-precision) + FCVTZS (vector, integer).

JosJuice:
I guess there's also the option of converting f64->i64 and then manually narrowing it to i32. This could be convenient if you want to emulate overflow accurately, but manually narrowing would take a few extra instructions.

Sonicadvance1:
> FCVTZS

Indeed, I have thought about reducing down to using the scalar operations but I'm not quite ready to take that leap without more testing. This PR gets us setup to take that plunge if need be

Sonicadvance1:
For future excavation the diff for implementing that code path is as follows:
```diff
diff --git a/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp b/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp
index 6156e07b9..f08bfb145 100644
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp
@@ -505,11 +505,14 @@ DEF_OP(Vector_F64ToI32) {
     case FEXCore::IR::Round_Host.Val: frinti(ARMEmitter::SubRegSize::i64Bit, Dst.Q(), Vector.Q()); break;
     }

-    // Now narrow from f64 to f32.
-    fcvtn(ARMEmitter::SubRegSize::i32Bit, Dst.Q(), Dst.Q());
-
-    ///< Convert the two F32 integrals to real integers.
-    fcvtzs(ARMEmitter::SubRegSize::i32Bit, Dst.D(), Dst.D());
+    ///< Narrow from F64->I32
+    dup(ARMEmitter::SubRegSize::i64Bit, VTMP1.Q(), Dst.Q(), 1);
+    fcvtzs(ARMEmitter::Size::i32Bit, TMP1.W(), Dst.D());
+    fcvtzs(ARMEmitter::Size::i32Bit, TMP2.W(), VTMP1.D());
+
+    ///< Insert back in to vector register.
+    fmov(ARMEmitter::Size::i32Bit, Dst.S(), TMP1.W());
+    ins(ARMEmitter::SubRegSize::i32Bit, Dst, 1, TMP2.W());
   }
 }
 ```

File: FEXCore/Source/Interface/Core/JIT/Arm64/ConversionOps.cpp [hunk 1]
```diff
@@ -449,5 +450,68 @@ DEF_OP(Vector_FToI) {
...
+  const auto Dst = GetVReg(Node);
+  const auto Vector = GetVReg(Op->Vector.ID());
+  if (HostSupportsSVE128 || HostSupportsSVE256) {
+    const auto Mask = Is256Bit ? PRED_TMP_32B.Merging() : PRED_TMP_16B.Merging();
+    // First step is to round the f64 values to integrals (frint*)
+    // Then convert to integers using fcvtzs.
+    switch (Round) {
+    case IR::Round_Nearest.Val:
+      frintn(ARMEmitter::SubRegSize::i64Bit, Dst.Z(), Mask, Vector.Z());
+      fcvtzs(Dst.Z(), ARMEmitter::SubRegSize::i32Bit, Mask, Dst.Z(), ARMEmitter::SubRegSize::i64Bit);
```
alyssarosenzweig, line 469: Pop the fcvtzs out of the switch for deduplication across all the cases

Sonicadvance1: Popped

alyssarosenzweig (approved):

lioncash:
Perhaps document the rounding problem Jos brought up in the ASIMD path so it isn't totally forgotten about. Otherwise this seems fine as an improvement

Sonicadvance1:
> Perhaps document the rounding problem Jos brought up in the ASIMD path so it isn't totally forgotten about. Otherwise this seems fine as an improvement

Added a comment

lioncash (approved):

PR #3845, opened 2024-07-08
Use nproc only if TEST_JOB_COUNT not specified

pmatos:


pmatos:
@neobrain I think a mistake went through in the last patch. We should have used elseif, otherwise in cmake < 3.29, TEST_JOB_COUNT is ignored.

neobrain (approved):
Yup, thanks for catching this!

PR #3846, opened 2024-07-09
x87: Make xam more robust

Sonicadvance1:
Had this floating


FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -1134,26 +1134,12 @@ void OpDispatchBuilder::X87FRSTOR(OpcodeArgs) {
...
   auto a = _LoadContextIndexed(top, 16, MMBaseOffset(), 16, FPRClass);
-  Ref Result = _VExtractToGPR(16, 8, a, 1);
-
-  // Extract the sign bit
-  Result = _Bfe(OpSize::i64Bit, 1, 15, Result);
-  SetRFLAG<FEXCore::X86State::X87FLAG_C1_LOC>(Result);
-
-  // Claim this is a normal number
-  // We don't support anything else
-  auto TopValid = GetX87ValidTag(top);
```
Sonicadvance1, line 1145: Needs Valid flag pumped through for empty/nonempty state

alyssarosenzweig:
Are we sure this is the right way forward? It doesn't seem that tricky to handle this properly in the opdispatcher, no?

Sonicadvance1:
> Are we sure this is the right way forward? It doesn't seem that tricky to handle this properly in the opdispatcher, no?

This actually gets spicy in the OpcodeDispatcher for this because of branches and no live values across block boundaries. In the CPU backend it would be trivial though.
Image_2024-07-09_20-02-57

alyssarosenzweig:
> > Are we sure this is the right way forward? It doesn't seem that tricky to handle this properly in the opdispatcher, no?
> 
> This actually gets spicy in the OpcodeDispatcher for this because of branches and no live values across block boundaries. In the CPU backend it would be trivial though.

Right, yeah... I think I assumed it would all flatten to csel's. (on the gpu it would :package:)

PR #3847, opened 2024-07-09
Enable coverage configuration for FEX

pmatos:
I am using this to understand our asm testsuite coverage of x87 stuff. I wonder if others would be interested in adding it to the CMakefile upstream.

The way to get it to work is:
1. Compile with `-DENABLE_COVERAGE` and `-DCMAKE_BUILD_TYPE=Debug`
2. `LLVM_PROFILE_FILE="default-%p.profraw" ninja asm_tests`
3. Run
```
llvm-profdata merge --sparse -o coverage.profdata `find . -name 'default-*.profraw'`
```
4. Run
```
llvm-cov show Bin/TestHarnessRunner -instr-profile=coverage.profdata -format=html -output-dir=./coverage-report -show-line-counts-or-regions -show-expansions -show-instantiations <PATH_TO_FEX_SRCROOT>
```

This will generate an html report at coverage report.  There are of course other ways to use coverage but this is way I am playing with atm.


Sonicadvance1 (approved):
Looks completely reasonable to me

PR #3848, opened 2024-07-09
Tests for X87 FTST

pmatos:


pmatos:
Noticed from #3847 we had no test for ftst, so here it is.

Sonicadvance1 (approved):
Nice!

PR #3852, opened 2024-07-10
 OpcodeDispatcher: Avoid template monomorphization to reduce FEXLoader binary size

neobrain:
Template functions are compiled once per set of template arguments (as if they were entirely separate functions). Moving all logic to an inner function and turning the templates into thin wrappers avoids binary size overhead.

In total, reduces FEXLoader size by 5.5% (45.1 MB -> 42.6 MB).

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -35,8 +35,7 @@ using X86Tables::OpToIndex;
 
 #define OpcodeArgs [[maybe_unused]] FEXCore::X86Tables::DecodedOp Op
 
-template<bool IsSyscallInst>
-void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
+void OpDispatchBuilder::SyscallOp(OpcodeArgs, bool IsSyscallInst) {
   constexpr size_t SyscallArgs = 7;
   using SyscallArray = std::array<uint64_t, SyscallArgs>;
 
@@ -128,6 +127,11 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
     NewRIP = _LoadContext(GPRSize, GPRClass, offsetof(FEXCore::Core::CPUState, rip));
     ExitFunction(NewRIP);
   }
+}
+
+template<bool IsSyscallInst>
+void OpDispatchBuilder::SyscallOp(OpcodeArgs) {
+  SyscallOp(Op, IsSyscallInst);
 }
 
 void OpDispatchBuilder::ThunkOp(OpcodeArgs) {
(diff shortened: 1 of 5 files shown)


Sonicadvance1:
> Template functions are compiled once per set of template arguments (as if they were entirely separate functions). Moving all logic to an inner function and turning the templates into thin wrappers avoids binary size overhead.
> 
> In total, reduces FEXLoader size by 5.5% (45.1 MB -> 42.6 MB).

What's the reduction on a stripped release build for reference?

alyssarosenzweig:
nodejs bench is ~similar, so I have no concerns about this from a performance perspective. And of course I'll never complain about fewer templates 😉 Good stuff, thanks Tony

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -1740,6 +1743,11 @@ void OpDispatchBuilder::RotateOp(OpcodeArgs) {
     GetSrcSize(Op) == OpSize::i32Bit ? std::make_optional(&OpDispatchBuilder::ZeroShiftResult) : std::nullopt);
 }
 
+template<bool Left, bool IsImmediate, bool Is1Bit>
```
alyssarosenzweig, line 1746: Is there a long term reason to have all these 'thunk templates', or is this just short term to avoid rewriting all the call sites?

neobrain: The dispatcher uses function pointer tables that must be filled with functions of the same signature. The new functions have extra parameters not compatible with that signature, whereas the templates allow "hiding" these parameters such that the function pointers can be written to the table.

We could instead hide the extra parameters as state somewhere (ew) so they don't need to be passed directly to the functions. In comparison the template wrappers seem like the lesser evil for now :)

neobrain:
> What's the reduction on a stripped release build for reference?

About the same: 5.6% (4.7MB->4.4MB)

The original numbers were measured for RelWithDebInfo. Debug builds actually get virtually no improvement (0.3%).

Sonicadvance1:
> > What's the reduction on a stripped release build for reference?
> 
> About the same: 5.6% (4.7MB->4.4MB)
> 
> The original numbers were measured for RelWithDebInfo. Debug builds actually get virtually no improvement (0.3%).

Very neat

Sonicadvance1 (approved):

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
alyssarosenzweig: Ah, gotcha, ok. That seems reasonable then (and indeed the lesser evil given the rest of the design this slots into ... the whole structure of the dispatcher indeed assumes templates, a C project would probably structure things differently to the same effect.) Anyway, seems reasonable to me, thanks for the explanation 👍

PR #3853, opened 2024-07-10
Fix all the warnings

neobrain:
Fixes all warnings I found in my local build configuration, excluding those from External/.

Two few patches are worth highlighting (others are trivial):
* de258ced20f9351b55281fd87accc5cbe91e0938: Adds an **internal** `unique_ptr` replacement, side-stepping the issues mentioned in a prior attempt in #3618 . By dropping movability, this becames a mere auto-deleting wrapper for non-transferable memory, which makes it much easier to reliably implement while allowing the rest of the code to continue using `std::unique_ptr`.
* 68936a0dce59e34fccb5819bc29e8776978676e2: Structs embedding `compat_ptr` can't use custom constructors without triggering warnings. Luckily we only need the constructors when *initializing data* or when *calling functions*, which can both be served by a utility class instead.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/MemoryOps.cpp
@@ -99,7 +99,7 @@ DEF_OP(LoadRegister) {
       }
     }
   } else if (Op->Class == IR::FPRClass) {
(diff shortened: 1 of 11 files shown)


neobrain:
Waiting for #3854 to be merged, otherwise good to go.

Sonicadvance1 (approved):
🎉

pmatos:
Argh! Thanks for this - that warning about invalid-offsetof was a real annoyance. 

❤️

Sonicadvance1 (approved):

PR #3854, opened 2024-07-10
fextl: Properly handle nullptr arguments in fextl::default_delete

neobrain:
`fextl::default_delete<...>{}(nullptr)` must be a no-op, reflecting behavior of std::default_delete.

Previously, `std::destroy_at` would invoke the destructor on null objects.


neobrain:
Reference for `destroy_at` not supporting nullptr: https://github.com/llvm/llvm-project/blob/8d800e6c9014e55ff729f0237efeba246750b12c/libcxx/include/__memory/construct_at.h#L73

lioncash (approved):

PR #3858, opened 2024-07-11
Implement support for SSE4.1/AVX NT loads

### LINKED ISSUE #3790: AVX128: Implement SSE4.1 non-temporal loads with SVE LDNT1B
SSE4.1 introduced the 128-bit `MOVNTDQA` instruction
This can be implemented with the SVE LDNT1B instruction

Additionally AVX2 extends this to a 256-bit version, which can be implemented with ASIMD LDNP.
Should be fairly similar to the store non-temporal version, just doesn't need to workaround the scalar and mmx variants since they don't exist.

Sonicadvance1:
Implements support for both SSE4.1 128-bit NT load and AVX 128/256-bit NT loads

Only feature missing is doing a 256-bit load with AVX128 using LDNP for discontiguous registers, but we don't have a sane way to return results in two different vector registers today.

Fixes #3790

### DIFF
--- a/CodeEmitter/CodeEmitter/SVEOps.inl
+++ b/CodeEmitter/CodeEmitter/SVEOps.inl
@@ -2569,7 +2569,19 @@ public:
   }
 
   // SVE contiguous non-temporal load (scalar plus immediate)
-  // XXX:
+  void ldnt1b(ZRegister zt, PRegister pg, Register rn, int32_t Imm = 0) {
+    SVEContiguousNontemporalLoad(0b00, zt, pg, rn, Imm);
+  }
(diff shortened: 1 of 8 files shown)


alyssarosenzweig (approved):
This looks solid. Just to confirm, by stripping the non-temporalness in the ASIMD fallback, we potentially lose performance but there's no impact to correctness?

Sonicadvance1:
> This looks solid. Just to confirm, by stripping the non-temporalness in the ASIMD fallback, we potentially lose performance but there's no impact to correctness?

Correct

lioncash (approved):

PR #3859, opened 2024-07-11
OpcodeDispatcher: Replace hand-written wrapper templates with a generic utility

neobrain:
Instead of writing template wrappers by hand and manually instantiating them (for making them visible in other .cpp files), a small generic template (`&OpDispatchBuilder::Bind`) does the job with much less code.

This also lowers the burden to adapt the pattern in new code.

Also sneaking in a 2-character fix for clang-format [mangling Lioncash's beautiful AVX documentation](https://github.com/FEX-Emu/FEX/blob/8d2b15665d893571919db4e6ec163121c287a867/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp#L4638-L4640).


neobrain:
Gave this a very close look before checking CI results. With CI passing, I'm reasonably confident I didn't slip any typos in :)

Sonicadvance1 (approved):

alyssarosenzweig:
This makes me very happy :3

pmatos:
Won't the removal of template arguments incur in some performance hit? For example, for X87ModifySTP, Inc can be a template argument, however you make it a function argument so the `if` on Inc, which could be a compile time `if`, cannot be one anymore.

neobrain:
> Won't the removal of template arguments incur in some performance hit? For example, for X87ModifySTP, Inc can be a template argument, however you make it a function argument so the `if` on Inc, which could be a compile time `if`, cannot be one anymore.
> 
> The same applies to other functions.

This PR doesn't change things in this regard:
* Previously: `X87ModifySTP<Inc>(OpcodeArgs)` was a thin wrapper around `X87ModifySTP(OpcodeArgs, bool Inc)`
* Now: `Bind<X87ModifySTP, Inc>` is a thin wrapper around `X87ModifySTP(OpcodeArgs, bool Inc)`

The motivation for moving logic from templates into non-template functions is given in #3852. In principle there can be a performance penalty, but it's unlikely the overhead of a few runtime branches has measurable effect on compile-time. (Conversely and similarly impactful, the non-templated functions may have better icache locality and may fare better in branch prediction.)

alyssarosenzweig:
> Won't the removal of template arguments incur in some performance hit? For example, for X87ModifySTP, Inc can be a template argument, however you make it a function argument so the `if` on Inc, which could be a compile time `if`, cannot be one anymore.
> 
> The same applies to other functions.

Just benchmarked this, no discernible difference. My :+1: stands.

lioncash (approved):

pmatos:
> > Won't the removal of template arguments incur in some performance hit? For example, for X87ModifySTP, Inc can be a template argument, however you make it a function argument so the `if` on Inc, which could be a compile time `if`, cannot be one anymore.
> > The same applies to other functions.
> 
> This PR doesn't change things in this regard:
> 
> * Previously: `X87ModifySTP<Inc>(OpcodeArgs)` was a thin wrapper around `X87ModifySTP(OpcodeArgs, bool Inc)`
> * Now: `Bind<X87ModifySTP, Inc>(OpcodeArgs)` is a thin wrapper around `X87ModifySTP(OpcodeArgs, bool Inc)`
> 
> The motivation for moving logic from templates into non-template functions is given in #3852. In principle there can be a performance penalty, but it's unlikely the overhead of a few runtime branches has measurable effect on JIT-runtime. (Conversely and similarly impactful, the non-templated functions may have better icache locality and may fare better in branch prediction.)

Very nice @neobrain . Thanks for the explanation on the motivation. Hadn't noticed #3852 yet.

pmatos:
> > Won't the removal of template arguments incur in some performance hit? For example, for X87ModifySTP, Inc can be a template argument, however you make it a function argument so the `if` on Inc, which could be a compile time `if`, cannot be one anymore.
> > The same applies to other functions.
> 
> Just benchmarked this, no discernible difference. My 👍 stands.

Thanks for confirming!

PR #3861, opened 2024-07-12
VDSO: Stop using a vector for a static

Sonicadvance1:
This causes a global initializer that registers an atexit handler.

Be smarter, use an std::array and pass its data around using a span instead.

Removes the global initializer and removes the atexit installation

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -195,7 +195,7 @@ public:
   // returns false if a handler was already registered
   CustomIRResult AddCustomIREntrypoint(uintptr_t Entrypoint, CustomIREntrypointHandler Handler, void* Creator = nullptr, void* Data = nullptr);
 
-  void AppendThunkDefinitions(const fextl::vector<FEXCore::IR::ThunkDefinition>& Definitions) override;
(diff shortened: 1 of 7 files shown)


FEXCore/Source/Interface/Context/Context.h
```diff
@@ -195,7 +195,7 @@ class ContextImpl final : public FEXCore::Context::Context {
   // returns false if a handler was already registered
   CustomIRResult AddCustomIREntrypoint(uintptr_t Entrypoint, CustomIREntrypointHandler Handler, void* Creator = nullptr, void* Data = nullptr);
 
-  void AppendThunkDefinitions(const fextl::vector<FEXCore::IR::ThunkDefinition>& Definitions) override;
+  void AppendThunkDefinitions(const std::span<FEXCore::IR::ThunkDefinition> Definitions) override;
```
neobrain, line 198: `const` has no effect for value parameters. Instead, the element references can be made const, since we copy them around anyway: `std::span<const ThunkDefinition>`

neobrain (approved):

FEXCore/Source/Interface/Context/Context.h
Sonicadvance1: changed

PR #3862, opened 2024-07-12
LogManager: Removes fextl::vector usage

Sonicadvance1:
We never use more than one logging method at a time so this was overengineered for what it is doing.

Instead only allow one handler for messages and throw messages each which just is a pointer.

Removes a global initializer and an atexit handler being installed

### DIFF
--- a/FEXCore/Source/Utils/LogManager.cpp
+++ b/FEXCore/Source/Utils/LogManager.cpp
@@ -8,27 +8,21 @@ $end_info$
 #include <FEXCore/Utils/CompilerDefs.h>
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/fextl/fmt.h>
-#include <FEXCore/fextl/vector.h>
-
-#include <cstdarg>
-#include <cstdio>
-#include <malloc.h>
 
 namespace LogMan {
 
 namespace Throw {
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Utils/LogManager.cpp
```diff
@@ -10,25 +10,22 @@ tags: glue|log-manager
 #include <FEXCore/fextl/fmt.h>
 #include <FEXCore/fextl/vector.h>
```
neobrain, line 11: This header is probably not needed anymore.

FEXCore/Source/Utils/LogManager.cpp
```diff
@@ -10,25 +10,22 @@ tags: glue|log-manager
...
 
-#include <cstdarg>
-#include <cstdio>
 #include <malloc.h>
```
neobrain, line 13: Do we still need this header?

neobrain (approved):

FEXCore/Source/Utils/LogManager.cpp
Sonicadvance1: Removed

FEXCore/Source/Utils/LogManager.cpp
Sonicadvance1: Removed

PR #3863, opened 2024-07-12
Ioctl32: Removes static fextl::vector in ioctlemulation

Sonicadvance1:
Removes a global static initializer for the vector and its atexit handler.

This handler array can be consteval similar to the x86 tables so it can be generated entirely at compile time.


lioncash (approved):
Nice!

PR #3866, opened 2024-07-12
ArgumentLoader: Removes static fextl::vector usage

Sonicadvance1:
Removes a global initializer and atexit registration

Ownership of this data has always been the frontend and the config system, we just used these static vectors as a side-channel.

### DIFF
--- a/Source/Common/ArgumentLoader.cpp
+++ b/Source/Common/ArgumentLoader.cpp
@@ -11,13 +11,16 @@
 #include <stdint.h>
 
 namespace FEX::ArgLoader {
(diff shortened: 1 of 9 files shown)


Sonicadvance1:
Okay, this change is causing tests that are on the known_failures list to start passing for some reason. This will need some investigation because this shouldn't cause a change to JIT logic at all...

Sonicadvance1:
Easy fix, just wasn't parsing arguments in one path, oops.

lioncash (approved):

PR #3867, opened 2024-07-12
Remove Disabled_Tests file

pmatos:
It's exactly the same as Known_Failures.

### DIFF
--- a/Scripts/guest_test_runner.py
+++ b/Scripts/guest_test_runner.py
@@ -120,7 +120,7 @@ if (mode == "guest"):
 for i in range(len(sys.argv) - StartingFEXArgsOffset):
     RunnerArgs.append(sys.argv[StartingFEXArgsOffset + i])
 
-#print(RunnerArgs)
+# print(RunnerArgs)
 
 ResultCode = 0
 
@@ -130,7 +130,8 @@ if (flake_tests.get(test_name)):
     TryCount = 5
 
 if (disabled_tests.get(test_name)):
-    ResultCode = -73
+    print(f"Test {test_name} is disabled")
+    sys.exit(0)
 
 # expect zero by default
 if (not test_name in expected_output):
@@ -170,4 +171,3 @@ else:
     else:
         # passed and expected to pass -- pass the test
         sys.exit(0)
-
--- a/unittests/gcc-target-tests-64/Disabled_Tests
+++ b/unittests/gcc-target-tests-64/Disabled_Tests
@@ -1,45 +1,4 @@
-# This has invalid asm generated for its test
-# 'foo' gets the argument in edi
-# Passes the argument to bar in eax
-# 'bar' accepts the argument in edi
-# inline asm is incorrect, needs to use =d and d on the value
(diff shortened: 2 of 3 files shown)


pmatos:
hummm, both 
```
The following tests FAILED:
	2273 - mcount_pic.c.gcc-target-test-64.jit.gcc-target-64 (Failed)
	2346 - pr57275.c.gcc-target-test-64.jit.gcc-target-64 (Failed)
``` 

were supposed to fail but actually pass now. Disabled_Tests was masking this.

Sonicadvance1:
> hummm, both
> 
> ```
> The following tests FAILED:
> 	2273 - mcount_pic.c.gcc-target-test-64.jit.gcc-target-64 (Failed)
> 	2346 - pr57275.c.gcc-target-test-64.jit.gcc-target-64 (Failed)
> ```
> 
> were supposed to fail but actually pass now. Disabled_Tests was masking this.

Very fancy

Sonicadvance1:
Oh yes, if one is related to SIGPROF then that should stay disabled because it's just a race condition, even if we're doing everything correctly.
AVX starting working makes sense.

pmatos:
@Sonicadvance1 I did something here that makes total sense - in my head - but maybe I am missing context. I added the racy test to `Disabled_Tests` file and removed it from `Known_Failures`. Then I noticed it still fails. Looking at `guest_test_runner.py`, it seems that if a test is Disabled, it just gets the ResultCode -73 (what random number is this) and keeps going, i.e. since we expect a 0 ReturnCode by default. I added a commit to just 'skip' the test, i.e. we exit for a disabled test. WDYT?

Sonicadvance1:
> @Sonicadvance1 I did something here that makes total sense - in my head - but maybe I am missing context. I added the racy test to `Disabled_Tests` file and removed it from `Known_Failures`. Then I noticed it still fails. Looking at `guest_test_runner.py`, it seems that if a test is Disabled, it just gets the ResultCode -73 (what random number is this) and keeps going, i.e. since we expect a 0 ReturnCode by default. I added a commit to just 'skip' the test, i.e. we exit for a disabled test. WDYT?

Yea, legacy weirdness that a disabled test just turns in to a "failing" test. Don't think about it too hard, your change makes sense, the original behaviour does not.

pmatos:
> > @Sonicadvance1 I did something here that makes total sense - in my head - but maybe I am missing context. I added the racy test to `Disabled_Tests` file and removed it from `Known_Failures`. Then I noticed it still fails. Looking at `guest_test_runner.py`, it seems that if a test is Disabled, it just gets the ResultCode -73 (what random number is this) and keeps going, i.e. since we expect a 0 ReturnCode by default. I added a commit to just 'skip' the test, i.e. we exit for a disabled test. WDYT?
> 
> Yea, legacy weirdness that a disabled test just turns in to a "failing" test. Don't think about it too hard, your change makes sense, the original behaviour does not.

Thanks - ready to review then.

Sonicadvance1 (approved):

PR #3869, opened 2024-07-12
IR: gather x87-ness automatically

alyssarosenzweig:
#3547 manually tracks when x87 is used. This seems verbose and error prone. Better to autogenerate, I think.

@pmatos What do you think?

### DIFF
--- a/FEXCore/Scripts/json_ir_generator.py
+++ b/FEXCore/Scripts/json_ir_generator.py
@@ -53,6 +53,7 @@ class OpDefinition:
     SSAArgNum: int
     NonSSAArgNum: int
     DynamicDispatch: bool
+    X87: bool
     JITDispatch: bool
     JITDispatchOverride: str
     TiedSource: int
@@ -76,6 +77,7 @@ class OpDefinition:
         self.SSAArgNum = 0
         self.NonSSAArgNum = 0
         self.DynamicDispatch = False
+        self.X87 = False
         self.JITDispatch = True
         self.JITDispatchOverride = None
         self.TiedSource = -1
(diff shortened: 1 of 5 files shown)


pmatos:
@alyssarosenzweig Thanks - love the idea. But there's a nit here. The flag is meant to mark when we need to lower operands in the optimization pass rather than when we actually use F80* stuff.

So just the operands in the switch in `X87StackOptimization::Run` need to be marked. This is probably my fault naming `HasX87`. It's supposed to mean `HasX87StackOperands`. This is why, for example, `OpDispatchBuilder::X87FSTCW` won't set `HasX87` - it doesn't generate anything that needs to be lowered by the pass itself.

alyssarosenzweig:
> @alyssarosenzweig Thanks - love the idea. But there's a nit here. The flag is meant to mark when we need to lower operands in the optimization pass rather than when we actually use F80* stuff.
> 
> So just the operands in the switch in `X87StackOptimization::Run` need to be marked. This is probably my fault naming `HasX87`. It's supposed to mean `HasX87StackOperands`. This is why, for example, `OpDispatchBuilder::X87FSTCW` won't set `HasX87` - it doesn't generate anything that needs to be lowered by the pass itself.

Sure. Maybe cherry-pick the first 4 commits here (the ones setting up the infrastructure) and leaving the commit that actually changes IR.json, and then in your branch you can set IR.json however you need for your pass? I'll close this in the mean time.

pmatos:
Thanks - will do that. Since this is independent of the x87 branch. I might even create a new PR with these changes.

pmatos:
hummm, no - actually because the ops that are relevant to mark are part of the patch, I might have to add the IR-ops with the patch itself.

alyssarosenzweig:
> hummm, no - actually because the ops that are relevant to mark are part of the patch, I might have to add the IR-ops with the patch itself.

Ya, that's why I closed it & passed the baton back to you 😄

pmatos:
> > hummm, no - actually because the ops that are relevant to mark are part of the patch, I might have to add the IR-ops with the patch itself.
> 
> Ya, that's why I closed it & passed the baton back to you 😄

Or I add them before in preparation for the pass . See #3872

Issue #3797, opened 2024-06-30
AVX128: VPERMIL{PS,PD} completely unoptimized


Sonicadvance1:
Similar to #3796, these are completely decomposed to zero register and inserts.

64-bit would be fairly easy with zips and 32-bit selections should be similar to other 32-bit shuffles/permutes

Issue #3784, opened 2024-06-30
AVX128: vpshufd can be improved


Sonicadvance1:
The implementation is just using the basic VInsElement loop for both 128-bit and 256-bit, this can be improved dramatically.

PR #3871, opened 2024-07-13
AVX128: Improve VPERMILPS/PD and VPSHUFD 

Sonicadvance1:
VPSHUFD and VPERMILPS are aliases of each other.

Reuses the implementation path from the PSHUFD implementation which has
a few swizzles and then a table lookup.

VPERMILPD is a very simple swizzle per 128-bit lane.

Fixes https://github.com/FEX-Emu/FEX/issues/3797
Fixes https://github.com/FEX-Emu/FEX/issues/3784


Sonicadvance1:
Brute force scanned a bunch more permutations of shuffles for vpermilps/vpshufd which helped add quite a few more handlers. It's not an all encompassing brute force method but it adds quite a few more handlers so we don't need to touch memory as often.

Added a unittest to cover the full immediate encoding of the instruction to ensure correct behaviour.

lioncash (approved):
Very nice move reductions all around

PR #3873, opened 2024-07-16
Remove unused function MmapOverride

pmatos:


pmatos:
clang warns about this unused function. Searching for uses finds nothing so, propose to remove.

Sonicadvance1 (approved):

neobrain (approved):
a047ac16995f76e972c0891ed811a860de0d034a should have removed this when reverting to the simplified interface. Thanks!

PR #3874, opened 2024-07-16
X87: save uop in ReconstructFTW

alyssarosenzweig:
noticed while reviewing Paulo's work


External/jemalloc
Sonicadvance1, line 1: oops, jemalloc got in

alyssarosenzweig: fixed, thanks

Sonicadvance1 (approved):

pmatos:
Nice! :)

PR #3876, opened 2024-07-17
Autogenerate LoweredX87() query, misc json_ir_generator cleanup in the area

alyssarosenzweig:
will be used to clean up the x87 pass


Sonicadvance1 (approved):
I was today days old when I learned about textwrap

PR #3881, opened 2024-07-19
FixedSizePooledAllocation: Fix a race when unclaiming disowned buffers

bylaws:
A disowned buffer could be unclaimed or claimed by a different thread in the time between the !IsFree check and locking the allocation mutex. Fix this and prevent such errors in the future by always checking ownership with the allocator locked before attempting to unclaim buffers.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
+++ b/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
@@ -120,16 +120,16 @@ public:
   }
 
   /**
-   * @brief Immediately release the buffer back to the allocator
+   * @brief Immediately release the buffer back to the allocator, given it has not been reclaimed
    *
    * @param Buffer - The iterator that was previously given with ClaimBuffer
-   *
-   * Once this is called on a buffer then the pool allocator has full ownership of the buffer
    */
-  void UnclaimBuffer(ContainerType::iterator Buffer) {
-    std::unique_lock lk {AllocationMutex};
-    (*Buffer)->CurrentClientOwnedFlag->store(ClientFlags::FLAG_FREE);
-    UnclaimBufferImpl(Buffer);
+  void UnclaimBuffer(ContainerType::iterator Buffer, BufferOwnedFlag* ClientFlag) {
+    // Transition the buffer to free, unclaiming if it wasn't free prior.
+    if (ClientFlag->exchange(ClientFlags::FLAG_FREE) != ClientFlags::FLAG_FREE) {
+      std::unique_lock lk {AllocationMutex};
+      UnclaimBufferImpl(Buffer);
+    }
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Spicy fix!

FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -120,16 +120,16 @@ class IntrusivePooledAllocator {
...
-   * Once this is called on a buffer then the pool allocator has full ownership of the buffer
    */
-  void UnclaimBuffer(ContainerType::iterator Buffer) {
-    std::unique_lock lk {AllocationMutex};
-    (*Buffer)->CurrentClientOwnedFlag->store(ClientFlags::FLAG_FREE);
-    UnclaimBufferImpl(Buffer);
+  void UnclaimBuffer(ContainerType::iterator Buffer, BufferOwnedFlag* ClientFlag) {
+    // Transition the buffer to free, unclaiming if it wasn't free prior.
+    if (ClientFlag->exchange(ClientFlags::FLAG_FREE) != ClientFlags::FLAG_FREE) {
+      std::unique_lock lk {AllocationMutex};
```
Sonicadvance1, line 130: Can't this now still end up in the situation that the flag goes from {DISOWNED,OWNED} to FREE, try and acquire the lock, failing, another thread scoops up the buffer, and then this executes freeing the buffer that another thread now owns?

I think the previous implementation where it checks the owned status, acquires the lock, and then exchanges the owned status is the most bug free.

bylaws: Perhaps I'm missing something, but it shouldn't be possible for another thread to claim the buffer until it's in UnclaimedBuffers, and that only happens after the lock is acquired.

Sonicadvance1:
```
T0.UnclaimBuffer(T0Buffer, Flag) -> T0 unclaiming buffer
T0.exchange(Flag, Free) == Owned; -> Free's buffer

=== This can weasel its way in here ===
T1.ClaimBuffer() returns T0Buffer -> Claims buffer that T0 just freed.

T0.Lock(AllocationMutex) -> Locks the allocation mutex, only gains ownership when T1.ClaimBuffer() is done
T0.UnclaimBuffer(T0Buffer) -> T0 free's the buffer that T1 just claimed, assuming it still owns it.
```

Sonicadvance1: It's a little tricky since `ClaimBuffer` walks both claimed buffers and freed buffer lists looking for disowned buffers inside of the ClaimedBuffers array, and `UnclaimedBuffers` is always free buffers.

bylaws: T1.ClaimBuffer() could never return the buffer though, as it isn't DISOWNED and in the ClaimedBuffers list, nor is it in the UnclaimedBuffers list

Sonicadvance1: Ah, you are probably right, safe since it only checks for disowned, not free.

Sonicadvance1 (approved):
After spending a bit of time to refresh my memory of the implementation, this does look correct

FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
bylaws: The joys of lock free code :)

Issue #3856, opened 2024-07-11
Implement DAZ using AFP.FIZ


Sonicadvance1:
Currently unsupported in FEX but will seemingly fix issues at least in warhammer 40k Gladius where it is doing a vector normalize pass and doing an equality check with zero. It likely generates a denormal with its math and DAZ flattens that to zero.

![Image_2024-07-09_16-30-17](https://github.com/FEX-Emu/FEX/assets/1018829/067bafa4-0604-44d7-bc14-414a2e081fb0)

PR #3883, opened 2024-07-20
Arm64: Implements support for DAZ using AFP.FIZ 

Sonicadvance1:
When AFP is supported then we can actually support DAZ. This might also
fix the audio corruption in Animal Well but I can't test it until Steam
is running on Oryon. Requires a bit of plumbing for MXCSR which we were
hacking around before but now we actually want to store the value.

Fixes https://github.com/FEX-Emu/FEX/issues/3856


alyssarosenzweig (approved):

Sonicadvance1:
DAZ did not infact fix the audio in animal well.

PR #3884, opened 2024-07-21
Telemetry: Remove VEX flag

Sonicadvance1:
This is no longer necessary and it also no longer provides us any useful information. Since we expose the AVX CPUID flag, basically everything uses VEX encoding now, so it is basically always set.


alyssarosenzweig (approved):
i like deleting telemetry

PR #3887, opened 2024-07-22
json_ir_generator: stop prefixing arguments

alyssarosenzweig:
stop prefixing the arguments when we generate allocate ops (in particular), this is more convenient and simpler. in exchange we need to prefix Op to avoid a collision on fcmpscalarinsert which has an argument named Op, but that's a local change at least.

came up when experimenting with new IR, but I think this is probably a win by itself.


Sonicadvance1 (approved):
Nice QOL improvement

Issue #3796, opened 2024-06-30
AVX128: Blends with immediate selector completely unoptimized


Sonicadvance1:
This includes the whole family of blends:
16-bit: VPBLENDW
32-bit: VPBLENDD, VBLENDPS
64-bit: VBLENDPD

64-bit blend is definitely the easiest where we should be able to move or zip depending.

32-bit and 16-bit get a bit more spicy.

These are all currently implemented with just a loop of inserts in to a zero register, so they are quite low performance atm.

PR #3888, opened 2024-07-24
AVX128: Optimize blends 

Sonicadvance1:
Optimizes the AVX128 blends by reusing the prior SSE4.1 implementation.
Only difference is the destination register isn't reused as a source
register.

One confusing thing is that Felix Cloutier's documentation has a typo on
the 256-bit VPBLENDW instruction where it had the top 128-bit lane
reusing the destination instead of sources. So I wrote a unittest to
ensure correctness.

Fixes https://github.com/FEX-Emu/FEX/issues/3796


Sonicadvance1:
Added a minor optimization to {v,}pblendw to hit 16 new implementations. My brute force solver found 240 selectors but they all involved more VInsElement operations which interact badly with our SRA code path and was significantly worse than the VTBX implementation because of it.

lioncash (approved):

Issue #3795, opened 2024-06-30
AVX128: VPERM{Q,PD} can be more optimized


Sonicadvance1:
We only hand optimize four selectors out of the total 256. Anything that falls out of those four turns in to a zero-register plus four inserts which is pretty bad. We don't have these sitting in instcountci at all right now.

An example of the bad output.
```json
    "vpermpd ymm0, ymm1, 01101011b": {
      "ExpectedInstructionCount": 9,
      "Comment": [
        "Map 3 0b01 0x01 256-bit"
      ],
      "ExpectedArm64ASM": [
        "ldr q2, [x28, #32]",
        "movi v3.2d, #0x0",
        "mov v4.16b, v3.16b",
        "mov v4.d[0], v2.d[1]",
        "mov v16.16b, v4.16b",
        "mov v16.d[1], v2.d[0]",
        "mov v3.d[0], v2.d[0]",
        "mov v3.d[1], v17.d[1]",
        "str q3, [x28, #16]"
      ]
    },
``` 

Theoretically worst case should devolve in to a TBL2 instruction which would be better than that mess of inserts, and of course we should support solving more selectors with handwritten optimizations.

VPERMQ and VPERMPD are aliases of each other so they behave the same.

Sonicadvance1:
Turns out a naive approach that uses TBL2 is actually worse than the insert path since we can't guarantee contiguous registers.
```diff
     "vpermpd ymm0, ymm1, 01101011b": {
-      "ExpectedInstructionCount": 9,
+      "ExpectedInstructionCount": 12,
       "Comment": [
         "Map 3 0b01 0x01 256-bit"
       ],
       "ExpectedArm64ASM": [
         "ldr q2, [x28, #32]",
-        "movi v3.2d, #0x0",
-        "mov v4.16b, v3.16b",
-        "mov v4.d[0], v2.d[1]",
-        "mov v16.16b, v4.16b",
-        "mov v16.d[1], v2.d[0]",
-        "mov v3.d[0], v2.d[0]",
-        "mov v3.d[1], v17.d[1]",
-        "str q3, [x28, #16]"
+        "ldr x0, [x28, #2032]",
+        "ldr q3, [x0, #176]",
+        "ldr x0, [x28, #2032]",
+        "ldr q4, [x0, #96]",
+        "mov v0.16b, v17.16b",
+        "mov v1.16b, v2.16b",
+        "tbl v16.16b, {v0.16b, v1.16b}, v3.16b",
[...]

PR #3892, opened 2024-07-26
AVX128: Optimize all cases of vpermq 

Sonicadvance1:
Started by cherry-picking some cases from the variants that appeared when running
Steam, games, AV1 convolve tests, openssl, ffmpeg, libjpeg-turbo,
openh264, libvpx, gemmlowp, libyuv, and dav1d.

Then turned it around and optimized them all since all variants end up
needing to be split in to two halves, that effectively means we need to
have 16 implementations, plus a couple of special cases for duplicated
results.

Fixes https://github.com/FEX-Emu/FEX/issues/3795


alyssarosenzweig:
instcountci coverage?

FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
```diff
@@ -1853,38 +1853,39 @@ void OpDispatchBuilder::AVX128_VPERMQ(OpcodeArgs) {
...
+    }
+  };
 
-    SrcIndex = (Selector >> (2 * 2)) & 0b11;
-    Result.High = _VInsElement(OpSize::i128Bit, OpSize::i64Bit, 0, SrcIndex & 1, ZeroRegister, Selections[SrcIndex]);
-    SrcIndex = (Selector >> (3 * 2)) & 0b11;
-    Result.High = _VInsElement(OpSize::i128Bit, OpSize::i64Bit, 1, SrcIndex & 1, Result.High, Selections[SrcIndex]);
-  }
+  Result.Low = SelectLane(SelectorLow, Src, 0, nullptr);
+  Result.High = SelectLane(SelectorHigh, Src, SelectorLow, Result.Low);
```
alyssarosenzweig, line 1888: I think it would be less confusing if this changed to something like

`Result.High = SelectorLow == SelectorHigh ? Result.Low : SelectLane(SelectorHigh, Src, SelectorLow);`

And then the special case inside SelectLane goes away.

Sonicadvance1: Removed special case and switched to a ternary.

Sonicadvance1:
> instcountci coverage?

Added coverage

alyssarosenzweig (approved):

PR #3893, opened 2024-07-26
Fix call to FNINITF64 and refactor

pmatos:


pmatos:
The instcountci pessimization is a bug fix. FRSTORF64 was not calling FINITF64, which meant it was not calling `_SetRoundingMode`.

alyssarosenzweig (approved):

Issue #3685, opened 2024-06-04
Tohou: Luna Nights tile x87 precision issue


Sonicadvance1:
For some reason when we run with x87 this game has tile rendering issues. When turning on x87 reduced precision the problem goes away.

This could be an issue with us not handling x87 precision modes correctly?

![Screenshot 2024-06-04 04-36-50](https://github.com/FEX-Emu/FEX/assets/1018829/abac1d79-4141-41a7-9473-d9f08f6dca19)

pmatos:
Another for my list to check. cc #3547

Sonicadvance1:
Retested because of recent vector instruction fixes. Behaviour is unchanged.

pmatos:
I can still reproduce this. I will try to understand what's going on.

Sonicadvance1:
Fixed by #4101 woo

PR #3894, opened 2024-07-26
ASM Tests: X87 Rounding modes

pmatos:


pmatos:
I was thinking initially the Luna Nights issue #3685 was related to rounding modes but it doesn't seem likely. Still investigating that one. In the meantime I wrote a bunch of tests that might be useful to have in-tree.

Sonicadvance1:
vixl sim doesn't support any rounding mode other than ties to even, so this will need to fall down the disabled test path for that

pmatos:
> vixl sim doesn't support any rounding mode other than ties to even, so this will need to fall down the disabled test path for that

Makes sense.

pmatos:
@Sonicadvance1 Is there no way to disable this just for the vixl sim path?

Sonicadvance1:
> @Sonicadvance1 Is there no way to disable this just for the vixl sim path?

Just like in the 64-bit ASM tests path, create a `Disabled_Tests_Simulator` file and add the test.
Similar to https://github.com/FEX-Emu/FEX/blob/main/unittests/ASM/Disabled_Tests_Simulator but in the 32Bit_ASM folder.

pmatos:
hummm, interesting, are disabled tests shown as passing? Or are the tests somehow 32bit and 64bit tests actually running and passing in vixl?

pmatos:
> hummm, interesting, are disabled tests shown as passing? Or are the tests somehow 32bit and 64bit tests actually running and passing in vixl?

I wonder if this is a result of us returning zero in `guest_test_runner.py` for disabled tests in:
```
if (disabled_tests.get(test_name)):
    print(f"Test {test_name} is disabled")
    sys.exit(0)
```

What's confusing me is that the print line is not showing up in the logs.

pmatos:
> > hummm, interesting, are disabled tests shown as passing? Or are the tests somehow 32bit and 64bit tests actually running and passing in vixl?
> 
> I wonder if this is a result of us returning zero in `guest_test_runner.py` for disabled tests in:
> 
> ```
> if (disabled_tests.get(test_name)):
>     print(f"Test {test_name} is disabled")
>     sys.exit(0)
> ```
> 
> What's confusing me is that the print line is not showing up in the logs.

Humm, ok, ASM tests use `testharness_runner.py`, not `guest_test_runner.py`.

pmatos:
**Please do not merge - I want to understand what's going on here because something doesn't make sense to me yet.**

pmatos:
OK - now I got it. Which is why I created #3896 . Ready to review.

Sonicadvance1 (approved):

PR #3899, opened 2024-07-27
Arm64: Handle backpatching in a thread-safe manner 

Sonicadvance1:
When code buffers are shared between threads, FEX needs to be careful
around backpatching its code buffers, since one thread might have
backpatched the code that another thread was also planning on
backpatching.

To handle this case, when the handler fails to find a backpatchable
instruction, check if it was already backpatched. This can be determined
by atomically reading the instructions back and seeing if they have
turned in to the non-atomic variants.

In most cases we can just return saying that it has been handled, in the
case of a store we need to back the PC up 4 bytes to ensure the DMB is
executed before the non-atomic store.


neobrain:
Great to have this out of the way :)

Did we have any documentation of how TSO support is implemented? I have a vague idea of what I'm looking at, but I don't have enough context for proper review.

Sonicadvance1:
> Great to have this out of the way :)
> 
> Did we have any documentation of how TSO support is implemented? I have a vague idea of what I'm looking at, but I don't have enough context for proper review.

Nope.
- Loads and stores get emitted as atomic load and store with nop padding on one side
   - Non-natively aligned, or crossing 16-byte granularity (with LSE2) will cause a SIGBUS
   - We then backpatch the NOP in to a DMB, and the loadstore in to a non-atomic loadstore
   - This "half-barrier" approach matches x86 TSO behaviour, with the painful cost of a DMB per loadstore
- Atomic Mops get emitted as an atomic ARM instruction, or a load exclusive+store exclusive pair with ALU op in the middle
  - On non-natively aligned, or crossing 16-byte granularity (with LSE2) will cause a SIGBUS
  - No backpatching occurs, we do two CAS loops inside the signal handler to handle both aligned halves of the operation
  - We update register state and increment the PC past the instruction

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+      LogMan::Msg::EFmt("Unhandled JIT SIGBUS Atomic mem op 0x{:02x}: PC: 0x{:x} Instruction: 0x{:08x}\n", Op, ProgramCounter, PC[0]);
+      return NotHandled;
+    }
+  } else if ((Instr & ArchHelpers::Arm64::LDAXR_MASK) == ArchHelpers::Arm64::LDAXR_INST) { // LDAXR*
+    uint64_t BytesToSkip = ArchHelpers::Arm64::HandleAtomicLoadstoreExclusive(ProgramCounter, GPRs);
+    if (BytesToSkip) {
+      // Skip this instruction now
+      return std::make_pair(true, BytesToSkip);
+    } else {
+      LogMan::Msg::EFmt("Unhandled JIT SIGBUS LDAXR: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
```
Sonicadvance1, line 1926: Oop, just noticed this else statement needs to be allowed to fall through to the LDAXR handler below. Otherwise ARMv8.0-a paranoid TSO mode breaks.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+      LogMan::Msg::EFmt("Unhandled JIT SIGBUS LDAXP: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
+      return NotHandled;
+    }
+  } else if ((Instr & ArchHelpers::Arm64::STLXP_MASK) == ArchHelpers::Arm64::STLXP_INST) { // STLXP
+    // Should not trigger - middle of an LDAXP/STAXP pair.
+    LogMan::Msg::EFmt("Unhandled JIT SIGBUS STLXP: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
+    return NotHandled;
+  }
+
+  ///< Check to see if another thread managed to backpatch this instruction before this thread got here.
```
neobrain, line 2024: From the three possible scenarios (no backpatching, fully backpatched, partially backpatched), which of them apply to the code after this comment and how are they handled? Seems outlining that here.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+    uint32_t LDUR = LDURBase;
+    LDUR |= Size << 30;
+    LDUR |= AddrReg << 5;
+    LDUR |= DataReg;
+    LDUR |= Instr & (0b1'1111'1111 << 9);
+    if (HandleType != UnalignedHandlerType::NonAtomic) {
+      ///< Ordering matters with cross-thread visibility!
+      std::atomic_ref<uint32_t>(PC[1]).store(DMB_LD, std::memory_order_relaxed); // Back-patch the half-barrier.
+    }
+    std::atomic_ref<uint32_t>(PC[0]).store(LDUR, std::memory_order_relaxed);
```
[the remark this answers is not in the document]
Sonicadvance1: Switched to release

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+  constexpr uint32_t LDRBase = 0b0011'1000'0111'1111'0110'1000'0000'0000;
+  constexpr uint32_t STRBase = 0b0011'1000'0011'1111'0110'1000'0000'0000;
+  constexpr uint32_t LDSTRegisterMask = 0b0011'1011'0010'0000'0000'1100'0000'0000;
+
+  constexpr uint32_t LDURBase = 0b0011'1000'0100'0000'0000'0000'0000'0000;
+  constexpr uint32_t STURBase = 0b0011'1000'0000'0000'0000'0000'0000'0000;
+  constexpr uint32_t LDSTUnscaledMask = 0b0011'1011'0010'0000'0000'1100'0000'0000;
+
+  constexpr uint32_t STPBase = 0b0010'1001'0000'0000'0000'0000'0000'0000;
+  constexpr uint32_t LDSTPMask = 0b0011'1011'1000'0000'0000'0000'0000'0000;
```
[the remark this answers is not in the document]
Sonicadvance1: Good shout, legacy that these were exposed in the FEXCore header. Moved the rest internal, and these over to those definitions instead.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+    uint32_t LDR = LDRBase;
+    LDR |= Size << 30;
+    LDR |= AddrReg << 5;
+    LDR |= DataReg;
+    if (HandleType != UnalignedHandlerType::NonAtomic) {
+      ///< Ordering matters with cross-thread visibility!
+      std::atomic_ref<uint32_t>(PC[1]).store(DMB_LD, std::memory_order_relaxed); // Back-patch the half-barrier.
+    }
+    std::atomic_ref<uint32_t>(PC[0]).store(LDR, std::memory_order_relaxed);
+    ClearICache(&PC[0], 16);
```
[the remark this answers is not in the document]
Sonicadvance1: Updated to pass in the number of bytes modified instead of hardcoded to 16. Since it's less than a cacheline the icache clear will still only do a single iteration of a loop so behaviour doesn't change.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2005,10 +2016,48 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
     // Should not trigger - middle of an LDAXP/STAXP pair.
     LogMan::Msg::EFmt("Unhandled JIT SIGBUS STLXP: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
     return NotHandled;
-  } else {
-    LogMan::Msg::EFmt("Unhandled JIT SIGBUS: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
-    return NotHandled;
   }
+
+  ///< Check to see if another thread managed to backpatch this instruction before this thread got here.
```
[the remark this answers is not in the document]
Sonicadvance1: unfluffed

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1948,57 +1992,19 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
     // Back up one instruction and have another go
     return std::make_pair(true, -4);
   } else if ((Instr & ArchHelpers::Arm64::LDAXP_MASK) == ArchHelpers::Arm64::LDAXP_INST) { // LDAXP
-    // Should be compare and swap pair only. LDAXP not used elsewhere
-    uint64_t BytesToSkip = ArchHelpers::Arm64::HandleCASPAL_ARMv8(Instr, ProgramCounter, GPRs);
-    if (BytesToSkip) {
-      // Skip this instruction now
-      return std::make_pair(true, BytesToSkip);
+    /// A little bit quirky this implementation.
```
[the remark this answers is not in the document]
Sonicadvance1: Removed the line

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2005,10 +2016,48 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
-    LogMan::Msg::EFmt("Unhandled JIT SIGBUS: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
-    return NotHandled;
   }
+
+  ///< Check to see if another thread managed to backpatch this instruction before this thread got here.
+  auto AtomicInst = std::atomic_ref<uint32_t>(PC[0]).load(std::memory_order_acquire);
+  if ((AtomicInst & LDSTRegisterMask) == LDRBase || (AtomicInst & LDSTUnscaledMask) == LDURBase) {
+    ///< This atomic instruction likely was backpatched to a load.
+    if (HandleType != UnalignedHandlerType::NonAtomic) {
+      ///< Assume was backpatched in this case
```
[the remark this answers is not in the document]
Sonicadvance1: Removed the comment and restructured this slightly. Found a bug that I was inverting when the DMB needed to be checked. DMB exists when `HandleType != UnalignedHandlerType::NonAtomic` not the other way around.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
Sonicadvance1: Added comment

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1830,237 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+    // Should be compare and swap pair only. LDAXP not used elsewhere
+    uint64_t BytesToSkip = ArchHelpers::Arm64::HandleCASPAL_ARMv8(Instr, ProgramCounter, GPRs);
+    if (BytesToSkip) {
+      // Skip this instruction now
+      return std::make_pair(true, BytesToSkip);
+    }
   }
+
+  // Lock code mutex during any SIGBUS handling that potentially changes code.
+  // Need to be careful to not read any code part-way through modification.
```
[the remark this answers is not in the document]
Sonicadvance1: Added comment

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2005,10 +2016,51 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
     LogMan::Msg::EFmt("Unhandled JIT SIGBUS STLXP: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
     return NotHandled;
-  } else {
-    LogMan::Msg::EFmt("Unhandled JIT SIGBUS: PC: 0x{:x} Instruction: 0x{:08x}\n", ProgramCounter, PC[0]);
-    return NotHandled;
   }
+
+  // Check if another thread backpatched this instruction before this thread got here.
+  // Since we got here, this can happen in a couple situations:
+  // - Unhandled instruction (Shouldn't occur, programmer error)
```
neobrain, line 2117: How would an unhandled instruction trigger SIGBUS / HandleUnalignedAccess?

Also, when talking about "programmer errors", it can be useful to stress which programmer you're talking about (FEX's or the guest application's?).

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1934,227 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+    return NotHandled;
+  }
+
+  ///< Check if another thread backpatched this instruction before this thread got here
+  // Since we got here, this can happen in a couple situations:
+  // - Unhandled instruction (Shouldn't occur, programmer error)
+  // - Another thread backpatched an atomic access to be a non-atomic access
+  auto AtomicInst = std::atomic_ref<uint32_t>(PC[0]).load(std::memory_order_acquire);
+  if ((AtomicInst & LDSTREGISTER_MASK) == LDR_INST || (AtomicInst & LDSTUNSCALED_MASK) == LDUR_INST) {
+    ///< This atomic instruction likely was backpatched to a load.
```
neobrain, line 2121: Why "likely"? When would it not be backpatched? What happens if we run this code  but it wasn't actually backpatched?

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1934,227 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+    if (HandleType != UnalignedHandlerType::NonAtomic) {
+      ///< Check the next instruction to see if it is a DMB.
```
neobrain, line 2123:
```suggestion
      ///< Check if the next instruction is a DMB.
```

neobrain (commented):
Heavy stuff, but I think I got the gist of it. Looks correct to the best I can tell :)

Docstrings need some polish, good to go otherwise.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
Sonicadvance1: Removed likely word. It will have been backpatched if we got there.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
Sonicadvance1: Updated comment to state that it would be a FEX programmer error.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2018,7 +2021,7 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
     return NotHandled;
   }
 
-  // Check if another thread backpatched this instruction before this thread got here.
+  ///< Check if another thread backpatched this instruction before this thread got here
```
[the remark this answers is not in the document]
Sonicadvance1: Removed. Just me being a derp.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1830,178 +1934,227 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+      if (DMBInst == DMB_LD) {
+        return std::make_pair(true, 0);
+      }
+    } else {
+      // No DMB instruction with this HandleType.
+      return std::make_pair(true, 0);
+    }
+  } else if ((AtomicInst & LDSTREGISTER_MASK) == STR_INST || (AtomicInst & LDSTUNSCALED_MASK) == STUR_INST) {
+    if (HandleType != UnalignedHandlerType::NonAtomic) {
+      // Check the previous instruction to see if it is a DMB.
```
neobrain, line 2134:
```suggestion
      // Check if the previous instruction is a DMB.
```

neobrain (approved):

PR #3901, opened 2024-07-30
Reuse Top in ReconstructFSW_Helper

pmatos:


alyssarosenzweig:
Can you explain why? This seems worse in instcountci: the same # of instructions but higher register pressure.

pmatos:
> Can you explain why? This seems worse in instcountci: the same # of instructions but higher register pressure.

For the slow path, I can't explain why... yet!
For the fast path, it's as I expected (see the newly added test); we saved a ldrb. 

I am going to investigate the slow path, which should stay the same.

pmatos:
> > Can you explain why? This seems worse in instcountci: the same # of instructions but higher register pressure.
> 
> For the slow path, I can't explain why... yet! For the fast path, it's as I expected (see the newly added test); we saved a ldrb.
> 
> I am going to investigate the slow path, which should stay the same.

OK - so the issue is that SyncStackSlow always returns a top value when called, not nullptr, so in fnstsw we'll always have a value coming in to the ReconstructFSW_Helper. This value needs to live through the other registers holding the flags increasing pressure. I tried reshuffling things around and moving the insertion of Top into FTW first in trying to improve things.

pmatos:
Might help reg pressure further if we intersperse the C0 retrieval and insertion. Will try that.

pmatos:
@alyssarosenzweig wdyt

alyssarosenzweig (approved):

PR #3902, opened 2024-07-30
man: Fixes newline issue with strenum

Sonicadvance1:
In the environment section this was causing the next environment variable line to be merged with the strenum options

Also makes it so strenum options doesn't have a spurious comma at the end of the list.

Before:
Image_2024-07-30_16-55-57
After:
Image_2024-07-30_16-56-09

### DIFF
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -148,9 +148,8 @@ def print_man_options(options):
             if (value_type == "strenum"):
                 Enums = op_vals["Enums"]
                 output_man.write("\\fBAvailable Options:\\fR\n")
-                for enum_op_key, enum_op_vals in Enums.items():
-                    output_man.write("{}, ".format(enum_op_vals))
-                output_man.write("\n")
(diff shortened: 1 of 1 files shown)


FEXCore/Scripts/config_generator.py
```diff
@@ -148,9 +148,15 @@ def print_man_options(options):
...
                 output_man.write("\\fBAvailable Options:\\fR\n")
+                item_count = len(Enums.items())
+                item_num = 0
                 for enum_op_key, enum_op_vals in Enums.items():
-                    output_man.write("{}, ".format(enum_op_vals))
-                output_man.write("\n")
+                    output_man.write("{}".format(enum_op_vals))
+                    item_num = item_num + 1
+                    if item_num != item_count:
+                        output_man.write(", ")
```
neobrain, line 157: Combining f-strings with `join()` expresses this much more succinctly without manual state tracking:
```suggestion
                output_man.write(", ".join(f"{enum_op_val}" for [_, enum_op_val] in Enums.items()))
```

Sonicadvance1: Neat, switched over to that.

neobrain (approved):

PR #3904, opened 2024-07-31
Scripts: Workaround deprecated parse_version

Sonicadvance1:
Different approach from #3579

Instead of completely dropping the deprecated path, support the new path and the old path using python try-except import exceptions.

This allows us to continue using old packages in CI, while supporting the future API once pkg_resources gets deprecated and removed. Best of both worlds.

### DIFF
--- a/Scripts/aarch64_fit_native.py
+++ b/Scripts/aarch64_fit_native.py
@@ -2,7 +2,10 @@
 import re
 import sys
 import subprocess
-from pkg_resources import parse_version
+try:
+    from packaging import version as version_check
+except:
+    from pkg_resources import parse_version as version_check
 
 # Order this list from oldest to newest
 # try not to list something newer than our minimum compiler supported version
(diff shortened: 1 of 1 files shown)


pmatos:
Thanks for this.

neobrain:
This breaks the build for me:

```
Traceback (most recent call last):
  File "/home/tony/fexsrc/Scripts/aarch64_fit_native.py", line 95, in <module>
    if version_check(clang_version) >= version_check(ID[1]):
       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'module' object is not callable
CMake Error at CMakeLists.txt:373 (string):
  string sub-command STRIP requires two arguments.
```

Presumably this could be fixed by installing some package, but judging by the PR description I'm assuming this was not an intended effect?

Sonicadvance1:
> This breaks the build for me:
> 
> ```
> Traceback (most recent call last):
>   File "/home/tony/fexsrc/Scripts/aarch64_fit_native.py", line 95, in <module>
>     if version_check(clang_version) >= version_check(ID[1]):
>        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
> TypeError: 'module' object is not callable
> CMake Error at CMakeLists.txt:373 (string):
>   string sub-command STRIP requires two arguments.
> ```
> 
> Presumably this could be fixed by installing some package, but judging by the PR description I'm assuming this was not an intended effect?

Not intended, Maybe try changing `from packaging import version as version_check` to `from packaging.version import Version as version_check`
I thought I tested this but maybe I munged the testing.

neobrain:
Yup, that fixes it. `packaging.version` being a submodule rather than a function would explain the error message previously seen :)

PR #3906, opened 2024-07-31
FEXCore: Add a generic spill/fill-all syscall ABI and use for Windows

bylaws:
Also drop the legacy hangover ABI as it has no users.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -51,15 +51,6 @@ void OpDispatchBuilder::SyscallOp(OpcodeArgs, bool IsSyscallInst) {
     FEXCore::X86State::REG_RSI, FEXCore::X86State::REG_RDI, FEXCore::X86State::REG_RBP,
   };
 
-  static constexpr SyscallArray GPRIndexes_Hangover = {
-    FEXCore::X86State::REG_RCX,
-  };
-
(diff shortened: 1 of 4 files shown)


alyssarosenzweig (approved):
is this what XTA does?

bylaws:
> is this what XTA does?
For 32 bit that would be the case since a full context needs to be available within syscalls, for arm64ec it uses exceptions in some way or another.

For context, here is what I have for ARM64EC syscall handling locally:
```
  uint64_t HandleSyscall(FEXCore::Core::CpuStateFrame* Frame, FEXCore::HLE::SyscallArguments* Args) override {
    ProcessPendingCrossProcessEmulatorWork();
    Frame->State.rip = NtDllBase + NtDllSyscallLUT[Frame->State.gregs[FEXCore::X86State::REG_RAX]];
    Frame->State.gregs[FEXCore::X86State::REG_RCX] = Frame->State.gregs[FEXCore::X86State::REG_R10];
    return 0;
  }
```
This just polls any inval events and redirects execution to the right arm exit thunk, the latter part could reasonably be implemented in ASM but I'd rather keep things simple for now. Though if it were to be implemented in asm - ProcessPendingCrossProcessEmulatorWork needs to be called by the JIT fairly regularly to handle e.g. code invalidation by a debugger so that would need to be called somewhere else.

alyssarosenzweig:
Interesting, OK. I'm always a fan of keeping things simple to start out 😄

Sonicadvance1 (approved):

PR #3908, opened 2024-07-31
CMake: Add option to build PDB debug info instead of DWARF

bylaws:


alyssarosenzweig (approved):

CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(COMPILE_VIXL_DISASSEMBLER "Compiles the vixl disassembler in to vixl" FAL
 option(ENABLE_FEXCORE_PROFILER "Enables use of the FEXCore timeline profiling capabilities" FALSE)
 set (FEXCORE_PROFILER_BACKEND "gpuvis" CACHE STRING "Set which backend you want to use for the FEXCore profiler")
 option(ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT "Enables glibc memory allocation hooking with fault for CI testing")
+option(USE_PDB_DEBUGINFO "Builds debug info in PDB format" FALSE)
```
neobrain, line 35: Can we stick to `ENABLE_` prefixes as done for (most of) the other binary options?

bylaws: In this case pdb and dwarf are mutually exclusive, so I think `USE` fits better than `ENABLE`

neobrain: Fair point, let's stick to USE for this one then.

PR #3909, opened 2024-07-31
Directly use the EC code bitmap for determining page arch

bylaws:
The prior approach using the L2 cache was flawed as it assumed L2 page entries had a 1-1 correspondence with actual pages. While the L2 cache could be extended to handle aliases with EC, this could lead to thrashing etc. The cost of a lookup in the actual EC code bitmap is cheap enough to perform every time considering the infrequency of calls to ARM64EC code when compare to X86 L2 hits.


Sonicadvance1 (approved):
Nice

PR #3910, opened 2024-07-31
F80: Drop dependency on state stored in TLS

bylaws:
Windows cannot support the implicit TLS as was used prior, so introduce a state structure and pass it in to functions where necessary.


alyssarosenzweig:
What happened to all the LoadDeferredFCW calls dropped from the f64 routines? Were these dead code before? Is this a behaviour change?

bylaws:
> What happened to all the LoadDeferredFCW calls dropped from the f64 routines? Were these dead code before? Is this a behaviour change?

Just dead code

alyssarosenzweig (approved):

Sonicadvance1 (approved):
Nice! I was wondering what it would take to remove the TLS state from the soft float library. Good job!

PR #3911, opened 2024-07-31
x87StackOptimizationPass: Default initialise StackMemberInfo members

bylaws:
Not doing so is UB.

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
@@ -195,19 +195,17 @@ private:
   struct StackMemberInfo {
     StackMemberInfo() {}
     StackMemberInfo(Ref Data)
-      : StackDataNode(Data)
-      , Source(std::nullopt)
-      , InterpretAsFloat(false) {}
+      : StackDataNode(Data) {}
     StackMemberInfo(Ref Data, Ref Source, OpSize Size, bool Float)
       : StackDataNode(Data)
       , Source({Size, Source})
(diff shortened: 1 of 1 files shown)


neobrain (approved):
`std::optional` doesn't need the initializer since it does have a default constructor, but LGTM.

Sonicadvance1 (approved):

pmatos:
This PR looks like a nonfunctional change. Although you refactored the constructor, the behavior didn't change. You moved the initial values to the member declaration, but they were being initialized anyway in the constructor.

Unless the point here is that default initializing `InterpretAsFloat` to false applies to other constructors, the only member from our Discord chat lacking a default constructor. Another strange thing that I should point out is that UBSan doesn't flag this as UB.

pmatos:
> This PR looks like a nonfunctional change. Although you refactored the constructor, the behavior didn't change. You moved the initial values to the member declaration, but they were being initialized anyway in the constructor.
> 
> Unless the point here is that default initializing `InterpretAsFloat` to false applies to other constructors, the only member from our Discord chat lacking a default constructor. Another strange thing that I should point out is that UBSan doesn't flag this as UB.

OK - after some further look, the impact here is on `StackMemberInfo() {},` and initializing `InterpretAsFloat,` which, as a fundamental type, doesn't have a default constructor. Also, UBsan probably won't show this because it would be a read on an uninitialized member, so that it would be MSan instead. Sorry, this comment is just meant to clarify my understanding. The PR looks good to me. 

I am still curious about how often this default constructor is called, but I will do a more thorough investigation after this lands.

bylaws:
Ref is also just a pointer so that also needs initializing, it's called in clear()

PR #3912, opened 2024-07-31
Don't apply the address-size flag to segment addresses

bylaws:
The address-size flag only applies to the offset from the segment base, rather than the segment address itself.


Sonicadvance1 (approved):
Spicy bug

PR #3913, opened 2024-08-01
Commonise logging and fallback to a log file for debug output on Windows

bylaws:


Source/Windows/Common/Logging.cpp
```diff
@@ -0,0 +1,44 @@
...
+    fwrite(Output.c_str(), 1, Output.size(), LogFile);
+  }
+}
+} // namespace
+
+namespace FEX::Windows::Logging {
+void Init() {
+  WineDbgOut = reinterpret_cast<decltype(WineDbgOut)>(GetProcAddress(GetModuleHandleA("ntdll.dll"), "__wine_dbg_output"));
+  if (!WineDbgOut) {
+    const auto Path = fextl::fmt::format("{}\\fex-{}.log", getenv("LOCALAPPDATA"), GetCurrentProcessId());
```
Sonicadvance1, line 38: Does this have the potential of dumping a ton of logs in to `LOCALAPPDATA` since there is no handling of the SilentLog FEX option?

bylaws: ah good point, I forget not everyone wants log spew :P

Sonicadvance1: That's basically what FEXLoader does. Since the OutputFD is set to -1 before config is loaded

Sonicadvance1 (approved):

PR #3915, opened 2024-08-01
ARM64EC: Support the JIT API as is used by Windows

bylaws:
Ontop of #3913 

This updates the exposed ARM64EC API to match that of Windows and upstream wine, though it is currently usable on neither due to FEX's reliance on msvcrt.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -588,6 +588,10 @@ void NotifyMemoryProtect(void* Address, SIZE_T Size, ULONG NewProt, BOOL After, 
   InvalidationTracker->HandleMemoryProtectionNotification(reinterpret_cast<uint64_t>(Address), static_cast<uint64_t>(Size), NewProt);
 }
 
+NTSTATUS NotifyMapViewOfSection(void* Unk1, void* Address, void* Unk2, SIZE_T Size, ULONG AllocType, ULONG Prot) {
+  return STATUS_SUCCESS;
+}
+
 void NotifyUnmapViewOfSection(void* Address, BOOL After, NTSTATUS Status) {
   if (!InvalidationTracker || !GetCPUArea().ThreadState()) {
     return;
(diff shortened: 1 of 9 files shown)


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -45,21 +46,29 @@ desc: Implements the ARM64EC BT module API using FEXCore
 #include <wine/debug.h>
 
 class ECSyscallHandler;
+
+extern "C" {
 void* X64ReturnInstr; // See Module.S
 extern void* ExitFunctionEC;
 
+// Wine doesn't support issuing direct system calls with SVC, and unlike Windows it doesn't have a 'stable' syscall number for NtContinue
+void* WineSyscallDispatcher;
```
Sonicadvance1, line 55: You probably want to `alignas(4096)` on this so that `adrp` in the Module.S is always guaranteed to have correct alignment when loading this address.

Sonicadvance1: Er wait, misread, you do lo12 offset. nevermind.

Source/Windows/ARM64EC/Module.cpp
```diff
@@ -364,6 +385,7 @@ static void RethrowGuestException(const EXCEPTION_RECORD& Rec, ARM64_NT_CONTEXT&
   uint64_t GuestSp = Context.X[Config.SRAGPRMapping[static_cast<size_t>(FEXCore::X86State::REG_RSP)]];
   struct DispatchArgs {
     ARM64_NT_CONTEXT Context;
+    uint64_t Pad[4]; // Only present on newer Windows versions, likely for SVE.
```
Sonicadvance1, line 388: How new are we talking here? Should a version check be somewhere, and handle the different flavours of arguments? Or a version check at startup that early exits without trying? Does WINE versus Windows care about the difference here?

bylaws: I have no idea really, I have only tested on a single windows version (latest insider)  but heard from the wine devs that it wasn't previously present. This is also quite awkward to test at runtime either way. Perhaps it's worth testing on your oryon laptop if you're still running the stock version of windows? I can send something to test

Upstream wine currently doesn't have the padding but I do locally and will PR it - it also doesn't support the exception handling callback that this ends up being called from so it doesn't really end up mattering.

Sonicadvance1: I'm not too concerned about it. At least not enough to jump in to Windows for any testing.
If it's on Windows and WINE is going to add the padding to match then it's reasonable enough for me. Even easier if WINE doesn't support hitting this code path yet.

Sonicadvance1 (approved):

PR #3916, opened 2024-08-04
FEX: Moves HostFeatures querying to the frontend

Sonicadvance1:
This moves the CPU feature querying to the frontend. The primary purpose here is for the wow64 frontend to not require linux-isms for querying these features. This is required since non-Linux environments don't have the "CPUID" feature for reading EL1 MSRs in EL0.

Wiring up the remaining wow64 registry querying is left for a future exercise.

This also technically removes an xbyak requirement from FEXCore for when building the x86 Test harness runner, but that doesn't really matter for regular use cases.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -145,6 +145,39 @@ if (CMAKE_SYSTEM_PROCESSOR MATCHES "^arm64ec")
   set(ENABLE_JEMALLOC TRUE)
 endif()
 
+include(CheckCXXSourceCompiles)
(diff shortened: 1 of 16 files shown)


bylaws (approved):
Per CPU midr info will need to be here as well, but probably makes sense to do that separately

bylaws:
Could OverrideFeatures be exposed individually aswell? Perhaps it makes sense to handle that on the fexcore side

lioncash (approved):

lioncash:
Needs a rebase

Issue #2996, opened 2023-08-24
State of externals (Updating / Upstreaming / Unforking)


JayFoxRox:
**Take anything below with a grain of salt; it's only based on a quick skim over the externals.**

---

Background: I wrote this list a couple of days ago after briefly discussing fex on the Asahi Linux Fedora matrix channel.
Basically for Asahi Fedora it would be beneficial if more of the dependencies could come from the system.
So I went and checked the externals and wrote down some thoughts.
Because I don't have a lot of time right now I'm posting it in the current state, so it doesn't go to waste on my local disk.
However, it's not polished and might be incorrect.

My hope with this issue is:
- Someone updates the libs which come from upstream.
- Discussion / feedback which libs can not be integrated and why (which could be an argument for packagers).
- Discussion how system libs could be used (e.g. by moving the custom CMake files out of forks or merging them upstream).
- Split this meta-issue into smaller, more manageable, issues (e.g. "Update to catch 3.x") which can be closed once a task is finished.

---

For packaging FEX, it might be beneficial to use system libraries.

I'm listing the state of the externals here:

## [Catch2 @ c4e3767](https://github.com/catchorg/Catch2/tree/c4e3767e265808590986d5db6ca1b5532a7f3d13)

This is v2.13.7
Newest v2: v2.13.10
Newest v3: v3.4.0

## [Vulkan-Headers @ 98f440c](https://github.com/KhronosGroup/Vulkan-Headers/tree/98f440ce6868c94f5ec6e198cc1adda4760e8849)

This is v1.3.231
Newest: v1.3.261

## [fmt @ a0b8a92](https://github.com/fmtlib/fmt/tree/a0b8a92e3d1532361c2f7feb63babc5c18d00ef2)

This is 10.0.0
Newest: 10.1.0

## [imgui @ 4c986ec](https://github.com/Sonicadvance1/imgui/tree/4c986ecb8d2807087fd8e34894d1e7a138bc2f1d)

This is a fork starting at 0bdc1453433ab0bb4d889c72ef9b0353ba3998aa (August 2019)

Changes:
- Don't autodetect OpenGL ES with custom loader
    - ?
- Disable glPolygonMode with ES. We aren't using it anyway.
    - Not sure why this commit exists
- Fixes KP enter on imgui SDL impl
    - This feels like something for upstream
- Add CMake file
    - Upstream rejects this / has trouble finding a working solution
- Add imgui addons
[...]

PR #3918, opened 2024-08-04
Config: Converts two LUT maps over linear scan arrays

Sonicadvance1:
These two maps used for environment lookup translations were getting
globally initialized and then registers with atexit handlers.

Switch over to a constexpr array and just do linear scans. This plus
short-circuiting the environment loader so it skips all entries that
don't start with `FEX_` has the side benefit of cutting the CPU time to
1/10th the time.

This plus https://github.com/FEX-Emu/FEX/pull/3917 removes the global static initializers entirely from
this file.

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -16,7 +16,6 @@
 #include <linux/limits.h>
 #include <pwd.h>
 #endif
-#include <list>
 #include <utility>
 #include <json-maker.h>
 #include <tiny-json.h>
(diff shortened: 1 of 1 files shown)


alyssarosenzweig:
I've never heard of this library. How will this affect #2996?

Sonicadvance1:
> I've never heard of this library. How will this affect #2996?

It's a header only library so it shouldn't cause issues. If they have problems with it then we'll need to find another solution for them, since there is no functional equivalent inside of C/C++.

neobrain:
Why aren't you converting the type to `std::pair<const char*, FEXCore::Config::ConfigOption>>[]` instead of pulling in a new library? The code isn't performance critical as far as I can see, so a theoretical degradation from O(1) to O(log n) shouldn't matter. Since `n` is small here, flat memory may actually perform better.

As an aside, since these PRs keep coming up, could we document the rationale for avoiding global static initializers somewhere?

Sonicadvance1:
Dropped the constexpr maps and switched to linear scan of an array. Also short-circuited the logic in the environment config loader so we stop loading all environment variables in to the temporary map, which accelerates execution time to 1/10th of what it was before.

Source/Common/CMakeLists.txt
```diff
@@ -17,4 +17,5 @@ endif()
 add_library(${NAME} STATIC ${SRCS})
 target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json json-maker FEXHeaderUtils)
 target_include_directories(${NAME} PRIVATE ${CMAKE_CURRENT_SOURCE_DIR}/External/cpp-optparse/)
+target_include_directories(${NAME} PRIVATE ${CMAKE_SOURCE_DIR}/External/eternal/include/)
```
alyssarosenzweig, line 20: please drop this

Sonicadvance1: Oops, dropped.

alyssarosenzweig (approved):

PR #3925, opened 2024-08-06
ARM64EC: Introduce FEX-side CRT and Windows API replacements

bylaws:
Ontop of #3922 

It is dangerous for FEX to rely on the system CRT as calls can have side effects that are also visible to the running application, and if patches are applied to any CRT exports used during compilation the call checker would trigger a reentry into the JIT to compile the patch and hence deadlock. Only functions that FEX actively uses are implemented, with the rest triggering an abort. In addition to this, both windows and upstream wine initialise kernel32 (provides the windows API) after loading the JIT, so dropping the dependency allows for compatibility with them.


Sonicadvance1 (approved):
Spicy, but since it is all isolated this is easy

PR #3926, opened 2024-08-06
FEXCore: Drop deferred signal handling on Windows

bylaws:
The async signal issues this handles do not exist on Windows.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
+++ b/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
@@ -93,22 +93,6 @@ public:
 private:
   pthread_rwlock_t Mutex;
 };
-#else
-// Windows doesn't support forking, so these can be standard mutexes.
-class ForkableUniqueMutex final : public std::mutex {
-public:
-  void StealAndDropActiveLocks() {
-    LogMan::Msg::AFmt("{} is unsupported on WIN32 builds!", __func__);
-  }
-};
-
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):
quick question, when do we use WIN32 and when do we use ARM64EC for ifdeffing?

FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
```diff
@@ -202,7 +202,6 @@ static auto MaskSignalsAndLockMutex(MutexType& mutex, uint64_t Mask = ~0ULL) {
   } scope_guard {ScopedSignalMasker {Mask}, LockType<MutexType> {mutex}};
   return scope_guard;
 #else
-  // TODO: Doesn't block signals which may or may not cause issues.
```
neobrain, line 205: Since Windows indeed doesn't need this, could you drop the interleaved `#ifdef`s and instead implement WIN32-specific variants of the functions at the bottom of the file?

The logic here is already dense enough as-is, no need to make it even more complicated by branching on platforms in each function. This would also make for a good place to comment that the names are misleading on WIN32.

bylaws:
> quick question, when do we use WIN32 and when do we use ARM64EC for ifdeffing?

WIN32 applies to both wow64 (32 bit x86 on 64 bit arm) and arm64ec FEX builds

Sonicadvance1 (approved):

PR #3927, opened 2024-08-06
ARM64EC: Set appropriate AFP and SVE256 state on JIT entry/exit

bylaws:


Source/Windows/ARM64EC/Module.S
```diff
@@ -53,6 +53,11 @@ BeginSimulation:
   // Expects the target code address in x9
 .global ExitFunctionEC
 ExitFunctionEC:
+  // Clear any AFP bits set in FPCR as native code won't expect their behaviour.
+  mrs x17, fpcr
+  and x17, x17, #-5 // NEP + AH + FIZ
```
alyssarosenzweig, line 58: Two questions.

1. For the bits that we preserve, will they ever vary? can we just hardcode the fpcr to restore here and save an instruction?
2. should this be predicated on the hw supporting AFP? (current Apple does not, for ex)

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -106,6 +106,30 @@ void Dispatcher::EmitDispatcher() {
     ptrue(ARMEmitter::SubRegSize::i8Bit, PRED_TMP_16B, ARMEmitter::PredicatePattern::SVE_VL16);
```
alyssarosenzweig, line 106: This would seem redundant in the SVE256 case no?

Source/Windows/ARM64EC/Module.S
bylaws: Since it's just clearing bits HW support shouldn't matter, and we want the rest of the bits to be preserved in arm64ec code

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
bylaws: Oh I was just matching fill here, will commonise the afp+sve SRA filling and fix there

Source/Windows/ARM64EC/Module.S
alyssarosenzweig: Sounds good.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
bylaws: these regs are different :)

Sonicadvance1 (approved):

PR #3928, opened 2024-08-07
InvalidationTracker: Better match Windows code invalidation behaviour

bylaws:
When given a NULL base address, Windows invalidation callbacks will ignore the given size and invalidate all code.

### DIFF
--- a/Source/Windows/Common/InvalidationTracker.cpp
+++ b/Source/Windows/Common/InvalidationTracker.cpp
@@ -63,8 +63,13 @@ void InvalidationTracker::InvalidateContainingSection(uint64_t Address, bool Fre
 }
 
 void InvalidationTracker::InvalidateAlignedInterval(uint64_t Address, uint64_t Size, bool Free) {
+  if (!Address) {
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Maybe a comment there to allow future code spelunkers to read that this is an explicit behaviour that nullptr address is matching Windows invalidation behaviour

Source/Windows/Common/InvalidationTracker.cpp
```diff
@@ -63,6 +63,10 @@ void InvalidationTracker::InvalidateContainingSection(uint64_t Address, bool Fre
 }
 
 void InvalidationTracker::InvalidateAlignedInterval(uint64_t Address, uint64_t Size, bool Free) {
+  if (!Address) {
+    Size = std::numeric_limits<uint64_t>::max();
```
neobrain, line 67: This will overflow when adding `FEX_PAGE_SIZE` to it, causing no pages to be invalidated instead.

bylaws: Oh good catch, thanks.

PR #3929, opened 2024-08-08
FEXInterpreter: Support portable installs

Sonicadvance1:
A couple things to note here.

This expects multiple environment variables to be set
- `FEX_PORTABLE=1`
- `FEX_THUNKHOSTLIBS=<Local install path for thunks>`
- `FEX_THUNKHOSTLIBS32=<Local install path for thunks>`
- `FEX_THUNKGUESTLIBS=<Local install path for thunks>`
- `FEX_THUNKGUESTLIBS32=<Local install path for thunks>`

Some optional environment variables that can be set
- FEX_APP_CONFIG_LOCATION
   - With FEX_PORTABLE set,  this will default to a path at `<FEXInterpreter>/fex-emu/`
   - This overrides the default $HOME behaviour
   - This path is used by the `$HOME/.fex-emu/Config.json` and `$HOME/.fex-emu/AppConfig/<Files>.json`
- FEX_APP_DATA_LOCATION
  - With FEX_PORTABLE set,  this will default to a path at `<FEXInterpreter>/fex-emu/`
  - This overrides the default $HOME behaviour
  - This path is used by the `$HOME/.fex-emu/Telemetry/` and various other data bits like IR cache, code cache and probably in the future more things

This also disables FEX trying to use the binfmt_misc handler even if it is installed, so it stays disjoint from the system install.
It also disables the three global config layers since it doesn't make sense in this config.

This isn't the default and shouldn't be expected as the default use case. It also likely breaks namespaces and chroots so shouldn't be used in those cases most of the time.

### DIFF
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -217,6 +217,14 @@ def print_man_environment_tail():
     ],
     "''", True)
 
+    print_man_env_option(
+    "FEX_PORTABLE",
+    [
+    "Allows FEX to run without installation. Global locations for configuration and binfmt_misc are ignored. These files are instead read from <FEXInterpreterPath>/fex-emu/ by default.",
+    "For further customization, see FEX_APP_CONFIG_LOCATION and FEX_APP_DATA_LOCATION."
+    ],
+    "''", True)
+
 def print_man_header():
     header ='''.Dd {0}
 .Dt FEX
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -338,10 +338,13 @@ fextl::string RecoverGuestProgramFilename(fextl::string Program, bool ExecFDInte
 }
 
 ApplicationNames LoadConfig(fextl::unique_ptr<FEX::ArgLoader::ArgLoader> ArgsLoader, bool LoadProgramConfig, char** const envp,
-                            bool ExecFDInterp, int ProgramFDFromEnv) {
-  FEX::Config::InitializeConfigs();
+                            bool ExecFDInterp, int ProgramFDFromEnv, const PortableInformation& PortableInfo) {
+  const bool IsPortable = PortableInfo.IsPortable;
+  FEX::Config::InitializeConfigs(PortableInfo);
   FEXCore::Config::Initialize();
-  FEXCore::Config::AddLayer(CreateGlobalMainLayer());
+  if (!IsPortable) {
+    FEXCore::Config::AddLayer(CreateGlobalMainLayer());
+  }
   FEXCore::Config::AddLayer(CreateMainLayer());
 
   auto Args = ArgsLoader->Get();
@@ -391,7 +394,9 @@ ApplicationNames LoadConfig(fextl::unique_ptr<FEX::ArgLoader::ArgLoader> ArgsLoa
       }
     }
 
-    FEXCore::Config::AddLayer(CreateAppLayer(ProgramName, FEXCore::Config::LayerType::LAYER_GLOBAL_APP));
+    if (!IsPortable) {
+      FEXCore::Config::AddLayer(CreateAppLayer(ProgramName, FEXCore::Config::LayerType::LAYER_GLOBAL_APP));
+    }
     FEXCore::Config::AddLayer(CreateAppLayer(ProgramName, FEXCore::Config::LayerType::LAYER_LOCAL_APP));
 
     auto SteamID = getenv("SteamAppId");
@@ -399,7 +404,9 @@ ApplicationNames LoadConfig(fextl::unique_ptr<FEX::ArgLoader::ArgLoader> ArgsLoa
       // If a SteamID exists then let's search for Steam application configs as well.
       // We want to key off both the SteamAppId number /and/ the executable since we may not want to thunk all binaries.
       fextl::string SteamAppName = fextl::fmt::format("Steam_{}_{}", SteamID, ProgramName);
-      FEXCore::Config::AddLayer(CreateAppLayer(SteamAppName, FEXCore::Config::LayerType::LAYER_GLOBAL_STEAM_APP));
+      if (!IsPortable) {
+        FEXCore::Config::AddLayer(CreateAppLayer(SteamAppName, FEXCore::Config::LayerType::LAYER_GLOBAL_STEAM_APP));
+      }
       FEXCore::Config::AddLayer(CreateAppLayer(SteamAppName, FEXCore::Config::LayerType::LAYER_LOCAL_STEAM_APP));
     }
   }
@@ -473,12 +480,16 @@ const char* GetHomeDirectory() {
 }
 #endif
 
-fextl::string GetDataDirectory() {
-  fextl::string DataDir {};
+fextl::string GetDataDirectory(const PortableInformation& PortableInfo) {
+  const char* DataOverride = getenv("FEX_APP_DATA_LOCATION");
 
+  if (PortableInfo.IsPortable && !DataOverride) {
+    return fextl::fmt::format("{}fex-emu/", PortableInfo.InterpreterPath);
+  }
+
+  fextl::string DataDir {};
   const char* HomeDir = GetHomeDirectory();
   const char* DataXDG = getenv("XDG_DATA_HOME");
-  const char* DataOverride = getenv("FEX_APP_DATA_LOCATION");
   if (DataOverride) {
     // Data override will override the complete directory
     DataDir = DataOverride;
@@ -489,14 +500,18 @@ fextl::string GetDataDirectory() {
   return DataDir;
 }
 
-fextl::string GetConfigDirectory(bool Global) {
+fextl::string GetConfigDirectory(bool Global, const PortableInformation& PortableInfo) {
+  const char* ConfigOverride = getenv("FEX_APP_CONFIG_LOCATION");
+  if (PortableInfo.IsPortable && (Global || !ConfigOverride)) {
+    return fextl::fmt::format("{}fex-emu/", PortableInfo.InterpreterPath);
+  }
+
   fextl::string ConfigDir;
   if (Global) {
     ConfigDir = GLOBAL_DATA_DIRECTORY;
   } else {
     const char* HomeDir = GetHomeDirectory();
     const char* ConfigXDG = getenv("XDG_CONFIG_HOME");
-    const char* ConfigOverride = getenv("FEX_APP_CONFIG_LOCATION");
     if (ConfigOverride) {
       // Config override completely overrides the config directory
       ConfigDir = ConfigOverride;
@@ -515,15 +530,15 @@ fextl::string GetConfigDirectory(bool Global) {
   return ConfigDir;
 }
 
-fextl::string GetConfigFileLocation(bool Global) {
-  return GetConfigDirectory(Global) + "Config.json";
+fextl::string GetConfigFileLocation(bool Global, const PortableInformation& PortableInfo) {
+  return GetConfigDirectory(Global, PortableInfo) + "Config.json";
 }
 
-void InitializeConfigs() {
-  FEXCore::Config::SetDataDirectory(GetDataDirectory());
-  FEXCore::Config::SetConfigDirectory(GetConfigDirectory(false), false);
-  FEXCore::Config::SetConfigDirectory(GetConfigDirectory(true), true);
-  FEXCore::Config::SetConfigFileLocation(GetConfigFileLocation(false), false);
-  FEXCore::Config::SetConfigFileLocation(GetConfigFileLocation(true), true);
+void InitializeConfigs(const PortableInformation& PortableInfo) {
+  FEXCore::Config::SetDataDirectory(GetDataDirectory(PortableInfo));
+  FEXCore::Config::SetConfigDirectory(GetConfigDirectory(false, PortableInfo), false);
+  FEXCore::Config::SetConfigDirectory(GetConfigDirectory(true, PortableInfo), true);
+  FEXCore::Config::SetConfigFileLocation(GetConfigFileLocation(false, PortableInfo), false);
+  FEXCore::Config::SetConfigFileLocation(GetConfigFileLocation(true, PortableInfo), true);
 }
 } // namespace FEX::Config
--- a/Source/Common/Config.h
+++ b/Source/Common/Config.h
@@ -28,6 +28,12 @@ struct ApplicationNames {
   fextl::string ProgramName;
 };
 
+struct PortableInformation {
+  bool IsPortable;
+  // Path of folder containing FEXInterpreter (including / at the end)
+  fextl::string InterpreterPath;
+};
+
 /**
  * @brief Loads the FEX and application configurations for the application that is getting ready to run.
  *
@@ -40,15 +46,15 @@ struct ApplicationNames {
  * @return The application name and path structure
  */
 ApplicationNames LoadConfig(fextl::unique_ptr<FEX::ArgLoader::ArgLoader> ArgLoader, bool LoadProgramConfig, char** const envp,
-                            bool ExecFDInterp, int ProgramFDFromEnv);
+                            bool ExecFDInterp, int ProgramFDFromEnv, const PortableInformation& PortableInfo);
 
 const char* GetHomeDirectory();
 
-fextl::string GetDataDirectory();
-fextl::string GetConfigDirectory(bool Global);
-fextl::string GetConfigFileLocation(bool Global);
+fextl::string GetDataDirectory(const PortableInformation& PortableInfo);
+fextl::string GetConfigDirectory(bool Global, const PortableInformation& PortableInfo);
+fextl::string GetConfigFileLocation(bool Global, const PortableInformation& PortableInfo);
 
-void InitializeConfigs();
+void InitializeConfigs(const PortableInformation& PortableInfo);
 
 /**
  * @brief Loads the global FEX config
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -64,7 +64,6 @@ $end_info$
 namespace {
 static bool SilentLog;
 static int OutputFD {-1};
-static bool ExecutedWithFD {false};
 
 void MsgHandler(LogMan::DebugLevels Level, const char* Message) {
   if (SilentLog) {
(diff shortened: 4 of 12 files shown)


Sonicadvance1:
This is working as intended and can now be reviewed/merged.

FEXCore/Scripts/config_generator.py
```diff
@@ -217,6 +217,20 @@ def print_man_environment_tail():
...
+    "FEX_PORTABLE",
+    [
+    "Forces FEX in to a more portable installation",
+    "This does a couple of things:",
+    "- Skips executing binfmt_misc handlers through execve",
+    "- global config (typically installed in /usr/share/fex-emu/) disabled",
+    "- FEX_APP_CONFIG_LOCATION defaults to FEXInterpreter relative paths instead of {$HOME, $XDG_DATA_HOME}/.fex-emu/",
+    "  - <FEXInterpreterPath>/fex-emu/"
+    "- FEX_APP_DATA_LOCATION defaults to FEXInterpreter relative path instead of {$HOME, $XDG_DATA_HOME}/.fex-emu/",
+    "  - <FEXInterpreterPath>/fex-emu/"
```
neobrain, line 230: What does "more portable" mean and how would the user know the limitations? Wouldn't it be more sensible to just call it "portable" and instead outline possible *exceptions*?

Something like this might be a starting point for improvement, though it doesn't list the limitations since I don't know what they are:
```suggestion
Allows FEX to run without installation. Global locations for configuration and binfmt_misc are ignored. These files are instead read from <FEXInterpreterPath>/fex-emu/ by default.

For further customization, the environment variables FEX_APP_CONFIG_LOCATION and FEX_APP_DATA_LOCATION can be used.
```

Source/Common/Config.cpp
```diff
@@ -338,10 +338,13 @@ fextl::string RecoverGuestProgramFilename(fextl::string Program, bool ExecFDInte
 }
 
 ApplicationNames LoadConfig(fextl::unique_ptr<FEX::ArgLoader::ArgLoader> ArgsLoader, bool LoadProgramConfig, char** const envp,
-                            bool ExecFDInterp, int ProgramFDFromEnv) {
-  FEX::Config::InitializeConfigs();
+                            bool ExecFDInterp, int ProgramFDFromEnv, const PortableInformation* PortableInfo) {
```
neobrain, line 341: We don't need pointer semantics here. `PortableInformation` is cheap enough to zero-construct (and is already non-null in most places anyway), so this can be a reference instead.

Source/Common/Config.cpp
```diff
@@ -473,12 +480,16 @@ const char* GetHomeDirectory() {
 }
 #endif
 
-fextl::string GetDataDirectory() {
-  fextl::string DataDir {};
+fextl::string GetDataDirectory(const PortableInformation* PortableInfo) {
```
neobrain, line 483:
```suggestion
fextl::string GetDataDirectory(const PortableInformation& PortableInfo) {
```

Source/Common/Config.cpp
```diff
@@ -489,14 +500,18 @@ fextl::string GetDataDirectory() {
   return DataDir;
 }
 
-fextl::string GetConfigDirectory(bool Global) {
+fextl::string GetConfigDirectory(bool Global, const PortableInformation* PortableInfo) {
+  const char* ConfigOverride = getenv("FEX_APP_CONFIG_LOCATION");
```
neobrain, line 504: We should avoid querying this in non-portable mode.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+
+  if (std::from_chars(PortableView.data(), PortableView.data() + PortableView.size(), Value).ec != std::errc {} || Value == 0) {
+    return BadResult;
+  }
+
+  char SelfPath[PATH_MAX];
+  auto Result = readlink("/proc/self/exe", SelfPath, PATH_MAX);
+  if (Result == -1) {
+    return BadResult;
+  }
```
neobrain, line 215: What's going on here? Needs a comment.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+
+  std::string_view SelfPathView {SelfPath, std::min<size_t>(PATH_MAX, Result)};
+
+  ///< Extract the absolute path from the FEXInterpret path
+  return {true, SelfPathView.substr(0, SelfPathView.find_last_of("/") + 1)};
```
neobrain, line 220: Doesn't this return a reference to the stack memory in `SelfPath`? Seems like PortableInformation should hold a `fextl::string` instead.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+}
+
 bool RanAsInterpreter(const char* Program) {
   return ExecutedWithFD || FEXLOADER_AS_INTERPRETER;
 }
 
+static bool InterpreterInstalled {};
+
 bool IsInterpreterInstalled() {
```
neobrain, line 229: The semantics of this have now become sufficiently complex that it needs a docstring to describe what "interpreter installed" means. A variant of the comment below would suffice, though as-is it's not clear why the various points mentioned are combined under the umbrella "interpreter installed".

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
-  // The interpreter is installed if both the binfmt_misc handlers are available
-  // Or if we were originally executed with FD. Which means the interpreter is installed
+  return InterpreterInstalled;
+}
 
-  return ExecutedWithFD || (access("/proc/sys/fs/binfmt_misc/FEX-x86", F_OK) == 0 && access("/proc/sys/fs/binfmt_misc/FEX-x86_64", F_OK) == 0);
+void SetInterpreterInstalled(bool Portable) {
```
neobrain, line 233: This might as well take `PortableInformation` directly, which avoids boolean blindness and doesn't cost us anything:
```suggestion
void SetInterpreterInstalled(const PortableInformation& Portable) {
```

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+  // The interpreter is installed if both the binfmt_misc handlers are available.
+  // Or if we were originally executed with FD. Which means the interpreter is installed.
```
neobrain, line 235: >// Or if we were originally executed with FD. Which means the interpreter is installed.

How are the two related?

FEXCore/Scripts/config_generator.py
```diff
@@ -217,6 +217,20 @@ def print_man_environment_tail():
...
     "''", True)
 
+    print_man_env_option(
+    "FEX_PORTABLE",
+    [
+    "Forces FEX in to a more portable installation",
+    "This does a couple of things:",
+    "- Skips executing binfmt_misc handlers through execve",
+    "- global config (typically installed in /usr/share/fex-emu/) disabled",
+    "- FEX_APP_CONFIG_LOCATION defaults to FEXInterpreter relative paths instead of {$HOME, $XDG_DATA_HOME}/.fex-emu/",
```
neobrain, line 227: Should `FEX_APP_CONFIG_LOCATION` and `FEX_APP_DATA_LOCATION` get their own sections too? Not sure what the typical conventions are for man pages here.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+
+  ///< Extract the absolute path from the FEXInterpret path
+  return {true, SelfPathView.substr(0, SelfPathView.find_last_of("/") + 1)};
+}
+
 bool RanAsInterpreter(const char* Program) {
   return ExecutedWithFD || FEXLOADER_AS_INTERPRETER;
 }
 
+static bool InterpreterInstalled {};
```
neobrain, line 227: Global state should be in a more visible place than in the middle of the file.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+
+  char SelfPath[PATH_MAX];
+  auto Result = readlink("/proc/self/exe", SelfPath, PATH_MAX);
+  if (Result == -1) {
+    return BadResult;
+  }
+
+  std::string_view SelfPathView {SelfPath, std::min<size_t>(PATH_MAX, Result)};
+
+  ///< Extract the absolute path from the FEXInterpret path
```
neobrain, line 219:
```suggestion
  // Extract the absolute path from the FEXInterpreter path
```

Source/Common/Config.h
```diff
@@ -28,6 +28,11 @@ struct ApplicationNames {
   fextl::string ProgramName;
 };
 
+struct PortableInformation {
+  bool IsPortable;
+  std::string_view InterpreterPath;
```
neobrain, line 33:
```suggestion
  // Path of folder containing FEXInterpreter (including / at the end)
  std::string_view InterpreterPath;
```

Source/Tools/FEXLoader/FEXLoader.cpp
neobrain, line 220: FWIW `find_last_of` has a char overload:
```suggestion
  return {true, SelfPathView.substr(0, SelfPathView.find_last_of('/') + 1)};
```

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +194,49 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
   }
 }
 
+FEX::Config::PortableInformation GetIsPortable() {
```
neobrain, line 197: This sounds cheaper than it is. Something like `ReadPortabilityInformation` reflects better that this reads the environment and filesystem.

Source/Common/Config.cpp
Sonicadvance1: With portable mode we are still allowing `FEX_APP_CONFIG_LOCATION` to override the path if it is set.

FEXCore/Scripts/config_generator.py
Sonicadvance1: Those already have their own sections above this one.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Added a comment. It's fetching the full application path from the Linux construct that is the `/proc/self/exe` symlink.

FEXCore/Scripts/config_generator.py
Sonicadvance1: Updated text

Source/Common/Config.cpp
Sonicadvance1: Switched over to a reference.

Source/Common/Config.cpp
Sonicadvance1: Switched over to a reference.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Switched over to `fextl::string` instead of a string_view

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Switched over

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Removed the global state since it was unnecessary to be global anyway.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Updated

Source/Common/Config.h
Sonicadvance1: Added comment

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Updated name of the function.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Added a docstring.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +193,53 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
-  return ExecutedWithFD || (access("/proc/sys/fs/binfmt_misc/FEX-x86", F_OK) == 0 && access("/proc/sys/fs/binfmt_misc/FEX-x86_64", F_OK) == 0);
+/**
+ * @brief Queries if FEX is installed as a binfmt_misc interpreter
+ *
+ * @param ExecutedWithFD If FEXInterpreter was executed using a binfmt_misc FD handle from the kernel
+ * @param Portable Portability information about FEX being run in portable mode
+ *
+ * @return true if the binfmt_misc handlers are installed and being used
+ */
+bool QueryInterpreterInstalled(const bool ExecutedWithFD, const FEX::Config::PortableInformation& Portable) {
```
neobrain, line 236: `const` on value parameters has no meaningful effect for short functions. It's only useful to avoid accidental modification in the implementation of large functions, but it's ignored externally either way.

```suggestion
bool QueryInterpreterInstalled(bool ExecutedWithFD, const FEX::Config::PortableInformation& Portable) {
```

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +193,53 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+  // We know the FEX interpreter is installed through a couple paths.
```
neobrain, line 237: Unclear what this line is trying to say. ~~Are we *checking* any paths here to figure out if the FEX interpreter is installed? Which paths? None of the bullet points after mention any paths.~~

On third read, it seems you're using "path" in a figurative way, which is confusing in code that also does filesystem operations... Is the intent here that we have multiple *methods* of checking whether FEXInterpreter is installed?

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +193,53 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+  // - If either of the binfmt_misc handlers are installed
```
neobrain, line 238:
```suggestion
  // - If both of the binfmt_misc handlers are installed
```

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -194,15 +193,53 @@ void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
...
+  // - Or if we were executed with an FD (Which can only happen from the kernel executing through binfmt_misc)
+  // If portable is set then FEX ignores anything installed through binfmt_misc.
+  return !Portable.IsPortable &&
```
neobrain, line 241: This becomes somewhat more readable by moving the portable check to an early-return statement:
```suggestion
  if (Portable.IsPortable) {
    // Don't use binfmt interpreter even if it's installed
    return false;
  }
  
  // <Longer comment goes here>
  return (ExecutedWithFD || (access("/proc/sys/fs/binfmt_misc/FEX-x86", F_OK) == 0 && access("/proc/sys/fs/binfmt_misc/FEX-x86_64", F_OK) == 0));
```

FEXCore/Scripts/config_generator.py
```diff
@@ -217,6 +217,14 @@ def print_man_environment_tail():
     ],
     "''", True)
 
+    print_man_env_option(
+    "FEX_PORTABLE",
+    [
+    "Allows FEX to run without installation. Global locations for configuration and binfmt_misc are ignored. These files are instead read from <FEXInterpreterPath>/fex-emu/ by default."
+    "For further customization, the environment variables FEX_APP_CONFIG_LOCATION and FEX_APP_DATA_LOCATION can be used."
```
neobrain, line 224: I missed that these aren't new in this PR. Since they have their own sections, this can be even shorter:

```suggestion
    "For further customization, see FEX_APP_CONFIG_LOCATION and FEX_APP_DATA_LOCATION."
```

Source/Common/Config.cpp
neobrain: Ah, I missed that the variables already existed before. Looks fine then.

Source/Common/Config.cpp
```diff
@@ -489,14 +500,18 @@ fextl::string GetDataDirectory() {
   return DataDir;
 }
 
-fextl::string GetConfigDirectory(bool Global) {
+fextl::string GetConfigDirectory(bool Global, const PortableInformation& PortableInfo) {
+  const char* ConfigOverride = getenv("FEX_APP_CONFIG_LOCATION");
+  if (PortableInfo.IsPortable && (Global || (!Global && !ConfigOverride))) {
```
neobrain, line 505: `!Global` is implied by... `Global` being false:

```suggestion
  if (PortableInfo.IsPortable && (Global || !ConfigOverride)) {
```

neobrain (commented):
Looking much better, thanks!

PR title still says "portable-ish". What's "-ish" about it?

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Yes, if the paths exist, or if `ExecutedWithFD` then we know that the Linux kernel binfmt_misc interpreter is installed. `/proc` doesn't need to be mounted inside of a chroot for example, so those two paths checks could not exist, but binfmt_misc interpreter is still installed.

neobrain:
PR title doesn't say "portable-ish" anymore, why did it say so before?

Sonicadvance1:
> PR title doesn't say "portable-ish" anymore, why did it say so before?

Oh right, It's effectively portable as long as you don't care about the quirks that can occur by not having binfmt_misc handlers installed. Which is why I said "portable-ish" before, but it is a non-issue for most things we care about.

Source/Tools/FEXLoader/FEXLoader.cpp
neobrain: This is probably clearer:

```
  // Check if FEX's binfmt_misc handlers are both installed.
  // The explicit check can be omitted if FEX was executed from an FD,
  // since this only happens if the kernel launched FEX through binfmt_misc
```

neobrain: Actually, scratch that, let me revisit this.

neobrain: I guess my proposal captures the essence of it, though it would be nice if the nuance of /proc possibly not being visible could be factored in as well. Then again `/proc` is *typically* mounted in a chroot, so this may be too much of an edge case to bother going into the detail.

Anyway, seems good enough.

neobrain (approved):
> Oh right, It's effectively portable as long as you don't care about the quirks that can occur by not having binfmt_misc handlers installed. Which is why I said "portable-ish" before, but it is a non-issue for most things we care about.

Ah, that makes sense. Sounds fine to advertise it without the "-ish" qualifier indeed then, since it's more of a side effect of binfmt_misc not being used than a limitation of portable mode.

PR #3932, opened 2024-08-09
ARM64EC: Install a custom call checker to bypass NTDLL function patches

bylaws:
Some programs will hook the NTDLL exports that FEX depends on, the regular ARM64EC call checker will detect such patches and invoke the JIT to run them, which leads to infinite recursion if those same exports are used during code compilation. Fix this by resolving all patchable FFSs to their native ARM implementations for all indirect calls performed by FEX, skipping any x86 patches.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -50,9 +50,12 @@ $end_info$
 class ECSyscallHandler;
 
 extern "C" {
-void* X64ReturnInstr; // See Module.S
-extern void* ExitFunctionEC;
+extern IMAGE_DOS_HEADER __ImageBase; // Provided by the linker
 
+extern void* ExitFunctionEC;
+extern void* CheckCall;
+
+void* X64ReturnInstr; // See Module.S
 uintptr_t NtDllBase;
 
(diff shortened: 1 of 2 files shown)


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -183,6 +186,32 @@ void FillNtDllLUTs() {
   }
 }
 
+template<typename T>
+void WriteModuleRVA(HMODULE Module, LONG RVA, T Data) {
+  if (!RVA) {
```
neobrain, line 191: What's "RVA"?

bylaws: relative virtual address - see https://learn.microsoft.com/en-us/windows/win32/debug/pe-format

neobrain: Is it fair to assume this is common knowledge for anyone reading this code, or is it an acronym that could trivially be spelled out somewhere?

bylaws: Yeah, this is a frequently used term within wine

Sonicadvance1 (approved):

PR #3933, opened 2024-08-09
Support cooperative suspend on ARM64EC

bylaws:
ARM64EC doesn't have an explicit callback for suspend, instead this task is delegated to the kernel setting a bool at a JIT-defined memory location, so the faulting based approach as was done for WOW64 cannot be used.


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -577,6 +577,14 @@ bool ResetToConsistentStateImpl(EXCEPTION_RECORD* Exception, CONTEXT* GuestConte
     return true;
   }
 
+  static constexpr uint32_t SuspendTrapMagic {0xD4395FC0}; // brk #0xCAFE
```
neobrain, line 580: Where does the value/instruction come from?

Why is it `constexpr`?
Why is it `static`?

bylaws: It's a constant that's literally just a `brk #0xCAFE` instruction, hence the constexpr and static

neobrain: > It's a constant that's literally just a brk #0xCAFE instruction

Why 0xcafe? I mean I know it matches up with JIT.cpp, but my point is random constants shouldn't be dropped into the source without any explanation.

> hence the constexpr and static

What benefit does this provide over the noise it adds compared to `const`?

bylaws: It's literally just a magic, I don't really think there is much that can be explained about it, I suppose I could comment that they match up?
I think it's generally cleaner to consistently keep all compile-time constants as such, switching between using const and constexpr etc for different types just adds unnecessary mental overhead.

neobrain: > It's literally just a magic that wont be used by arm64 windows programs, I don't really think there is much that can be explained about it, I suppose I could comment that it matches up with what is emitted in JIT.cpp?

That would work, or a shared named constant (once used directly, and once as something like `0xabc0000 | (cafe_constant << 4)`). What's matters is that the link between JIT.cpp and Module.cpp becomes clear. The latter approach has the benefit of being robust against future code moves or file renames.

Please be mindful about this sort of issue, since features subtlely spreading across multiple files without a clear link between them is an easy source of bugs during future refactors.

>  I think it's generally cleaner to consistently keep all compile-time constants as such

We use `static` for global constants, where it makes sense when those constants are used only in a single cpp file. Within a function, `static` has different (often surprising) semantics and hence shouldn't be used for constants.

bylaws: > Within a function, static has different (often surprising) semantics and hence shouldn't be used for constants.

What are these semantics when combined with constexpr? Allowing for a non unique address doesn't seem like it would ever be a downside in the case of constants

neobrain: I don't understand where the insistence is coming from to keep a keyword that doesn't do anything. It seems that you want to use it for consistency, but using static on local constants isn't actually a pattern that we employ much elsewhere, short of arrays perhaps?

If you're blindly applying static wherever possible for convenience, that is a cause for concern. We've had actual bugs coming from code that used static for no reason and regressed after a refactor purely because of the unnecessary `static`.

> Allowing for a non unique address

You're thinking of global variables here. That's not part of the semantics of `static` when applied to a function variable.

> a ctor needs to be called static can allow for better codegen.

Conversely `static` local variables require the compiler to add synchronization code in general due to guaranteed one-time-initialization. "Caching" nontrivial ctor calls using `static` as you suggest is difficult in practice due to state management.

> What are these semantics when combined with constexpr?

For constexpr variables in a function, it just doesn't do anything, assuming a well-behaving compiler.

bylaws: All of the existing windows code defaults to static constexpr so I'd rather be consistent with neighboring code than some other parts of FEX. Like I can change that but it just seems negligible

Perhaps I should make it clear that I'm specifically talking about static with constexpr here, most of your points are just about static in general which for sure should be used very carefully but they  really do not apply to what I'm meaning. Using static constexpr for constants only has none of the downsides you state.

See for codegen benefits: https://stackoverflow.com/a/76696023

> You're thinking of global variables here. That's not part of the semantics of static when applied to a function variable.

Maybe my terminology is wrong here? Static on a local variable will permit the same address across function invocations which can prevent extra copies having to be constructed when the compiler cannot prove e.g. a recursive call.

Sonicadvance1: `static constexpr` and `constexpr static` is definitely an idiom that we have spread throughout the codebase, just about 200 instances in FEXCore/ and Source/. If we don't want this idiom then we should create a tracking issue towards removing it and start working towards that goal rather than blocking this PR for using it.

neobrain: > static constexpr and constexpr static is definitely an idiom that we have spread throughout the codebase

This was mentioned above; the vast majority (all?) of the instances you mention are at global scope.

I can create a spring cleanup PR for this if it helps move the issue forward however, since there's other arguments for removing `static` at global scope, too.

Sonicadvance1: Sounds reasonable to me.
Out of the 183 instances, 63 are in .cpp. So ~34% in local scope.

bylaws: Looking around it seems like it's preferred to do effectively the opposite to what fex currently does: https://youtu.be/IDQ0ng8RIqs
static only serves to benefit for locals and at worst it does nothing.

Perhaps the cleanup could do that if you're doing it?

neobrain: With all respect for Jason and his work, I'm not sure the "constexpr all the things" paradigm lends itself well for deriving guidelines for FEX :)

I've been doing some pondering though and I think I see a good way forward. Will evaluate how viable it is while working on other things, but let's move on with this PR for the time being!

Sonicadvance1 (approved):

Source/Windows/ARM64EC/Module.cpp
neobrain: Opened #3958 to track this.

PR #3935, opened 2024-08-09
Windows: Support CPU feature detection from ID registry keys

bylaws:
Also disables the usage of cache maintenance operations on native Windows. Features can still be disabled in Config.json if need be.


Sonicadvance1:
Needs a rebase

Sonicadvance1 (approved):

bylaws:
rebased

PR #3944, opened 2024-08-11
LinuxSyscalls: Some minor cleanups

Sonicadvance1:
- We can have the SyscallFunctionDefinitions be the correct size out of the gate. Both tables are always 512 entries in size.
- In the RegisterSyscall_{32,64} handlers, just get the reference using operator[]. We always know we will be under the size of the array, add a an assert to check. Removes a bit of vector range checking overhead.
- Namespace 32-bit syscalls like 64-bit syscalls and include in the regular header like 64-bit. This was just an oversight
- Use std::fill for the syscall gap for the invalid syscall, just a minor cleanup.

No functional change.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
@@ -40,6 +40,7 @@ $end_info$
 #endif
 
 #include "LinuxSyscalls/x64/SyscallsEnum.h"
+#include "LinuxSyscalls/x32/SyscallsEnum.h"
 
 #define CONCAT_(a, b) a##b
 #define CONCAT(a, b) CONCAT_(a, b)
@@ -255,7 +256,11 @@ public:
 protected:
   SyscallHandler(FEXCore::Context::Context* _CTX, FEX::HLE::SignalDelegator* _SignalDelegation);
 
-  fextl::vector<SyscallFunctionDefinition> Definitions {};
+  fextl::vector<SyscallFunctionDefinition> Definitions {std::max<std::size_t>(FEX::HLE::x64::SYSCALL_x64_MAX, FEX::HLE::x32::SYSCALL_x86_MAX),
+                                                        {
+                                                          .NumArgs = 255,
+                                                          .Ptr = reinterpret_cast<void*>(&UnimplementedSyscall),
+                                                        }};
   std::mutex MMapMutex;
 
   // BRK management
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Syscalls.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Syscalls.h
@@ -31,7 +31,6 @@ class SignalDelegator;
 }
 
 namespace FEX::HLE::x32 {
-#include "SyscallsEnum.h"
 
 class x32SyscallHandler final : public FEX::HLE::SyscallHandler {
 public:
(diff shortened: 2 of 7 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -255,7 +256,20 @@ class SyscallHandler : public FEXCore::HLE::SyscallHandler, FEXCore::HLE::Source
...
+    } raw;
+    raw.val = in;
+    return raw.raw;
+  };
+
+  fextl::vector<SyscallFunctionDefinition> Definitions {std::max<std::size_t>(FEX::HLE::x64::SYSCALL_x64_MAX, FEX::HLE::x32::SYSCALL_x86_MAX),
+                                                        {
+                                                          .NumArgs = 255,
+                                                          .Ptr = cvt(&UnimplementedSyscall),
+                                                        }};
```
neobrain, line 272: Can we move this to the constructor instead? I'd prefer the funky lambda to remain local state in a function instead of being a visible class member. I don't think `std::vector` allocates any memory on default construction, so I don't think there's a difference in allocation behavior either way.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Syscalls.h
```diff
@@ -48,16 +47,9 @@ class x32SyscallHandler final : public FEX::HLE::SyscallHandler {
                           const fextl::string& TraceFormatString,
 #endif
                           void* SyscallHandler, int ArgumentCount) override {
-    auto& Def = Definitions.at(SyscallNumber);
+    LOGMAN_THROW_A_FMT(SyscallNumber < Definitions.size(), "Trying to add syscall number way beyond extent!");
+    auto& Def = Definitions[SyscallNumber];
```
neobrain, line 51: I'm not convinced the saved cost of a bounds check in Release mode justifies the added code noise here. Was there any particular use case you profiled to come up with this optimization?

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
Sonicadvance1: Moved to constructor.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Syscalls.h
Sonicadvance1: Removed.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -255,7 +256,16 @@ class SyscallHandler : public FEXCore::HLE::SyscallHandler, FEXCore::HLE::Source
 protected:
   SyscallHandler(FEXCore::Context::Context* _CTX, FEX::HLE::SignalDelegator* _SignalDelegation);
 
-  fextl::vector<SyscallFunctionDefinition> Definitions {};
+  static auto constexpr cvt = [](auto in) {
```
neobrain, line 259: This funky lambda is still visible in the header. I'm guessing it's here to avoid duplicating it, but if it stays in the visible interface, it should be converted to a template member function and it should be given a more suitable name + short docstring.

Sonicadvance1: Ah, misinterpreted the comment originally. Will fix.

Sonicadvance1: Just removed the lambda since all it is doing is a reinterpret_cast

neobrain (approved):

PR #3946, opened 2024-08-11
VDSO: Implements v6.11 vdso getrandom

Sonicadvance1:
This is an interesting vdso implementation because it only exists in
x86-64 with kernel v6.11. For Arm64 the implementation is likely to land
in a couple kernel versions from now.

Some differences with vdso_getrandom versus the regular getrandom
syscall
- Has two additional arguments, opaque_state and opaque_len
- Expects the userspace to mmap/munmap this opaque state structure
- Lets userspace query information about this structure upfront
- If the opaque data structure isn't provided then it falls back to
  regular SYS_getrandom

With the "glibc" implementation, we can tell the interface to allocate a
single page that gets unused (otherwise glibc ends up not using the
interface), and fallback to the regular SYS_getrandom.

In the case of the vdso interface, currently the arguments just get
passed through.

~~Keeping this as a WIP until the ARM64 kernel patches land and I can
actually test the things. Running the "glibc" path works today with the
selftest, but the vdso path is currently completely untested.~~

### DIFF
--- a/Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
+++ b/Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
@@ -28,12 +28,14 @@ namespace VDSOHandlers {
   using ClockGetTimeType = decltype(::clock_gettime)*;
   using ClockGetResType = decltype(::clock_getres)*;
   using GetCPUType = decltype(FHU::Syscalls::getcpu)*;
+  using GetRandomType = ssize_t (*)(void*, size_t, uint32_t, void*, size_t);
 
   TimeType TimePtr;
   GetTimeOfDayType GetTimeOfDayPtr;
   ClockGetTimeType ClockGetTimePtr;
   ClockGetResType ClockGetResPtr;
   GetCPUType GetCPUPtr;
+  GetRandomType GetRandomPtr;
 } // namespace VDSOHandlers
 
 using HandlerPtr = void (*)(void*);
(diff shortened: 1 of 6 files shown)


Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -98,6 +100,39 @@ namespace x64 {
...
     }
+
+    static void getrandom(void* ArgsRV) {
+      struct vgetrandom_opaque_params {
+        uint32_t size_of_opaque_state;
+        uint32_t mmap_prot;
+        uint32_t mmap_flags;
+        uint32_t reserved[13];
+      };
+      static_assert(sizeof(vgetrandom_opaque_params) == sizeof(uint32_t[16]), "Oops");
```
neobrain, line 111: `static_assert` can be used without assertion messages, avoiding the need for goofy placeholders.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -553,6 +621,7 @@ void* LoadVDSOThunks(bool Is64Bit, FEX::HLE::SyscallHandler* const Handler) {
     VDSODefinitions[3].ThunkFunction = FEX::VDSO::x32::Handler_clock_gettime64;
     VDSODefinitions[4].ThunkFunction = FEX::VDSO::x32::Handler_clock_getres;
     VDSODefinitions[5].ThunkFunction = FEX::VDSO::x32::Handler_getcpu;
+    ///< VDSODefinitions[6] which is getrandom doesn't exist on 32-bit.
```
neobrain, line 624: Slightly misleading. This is clearer:

```suggestion
    ///< getrandom doesn't exist on 32-bit, so leave VDSODefinitions[6] unfilled
```

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -595,8 +664,8 @@ uint64_t GetVSyscallEntry(const void* VDSOBase) {
   return 0;
 }
 
-const std::span<FEXCore::IR::ThunkDefinition> GetVDSOThunkDefinitions() {
-  return VDSODefinitions;
+const std::span<FEXCore::IR::ThunkDefinition> GetVDSOThunkDefinitions(bool Is64Bit) {
+  return std::span(VDSODefinitions.begin(), VDSODefinitions.begin() + VDSODefinitions.size() - (Is64Bit ? 0 : 1));
```
neobrain, line 668: Why not `end()`?

```suggestion
  return std::span(VDSODefinitions.begin(), VDSODefinitions.end() - Is64Bit);
```

This could also be useful:
```suggestion
  return std::span { VDSODefinitions }.first(VDSODefinitions.size() - Is64Bit);
```

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -431,10 +480,22 @@ void LoadHostVDSO() {
...
+
+  SymbolPtr = dlsym(vdso, "__kernel_getrandom");
+  if (!SymbolPtr) {
+    SymbolPtr = dlsym(vdso, "__vdso_getrandom");
+  }
+
+  if (SymbolPtr) {
+    VDSOHandlers::GetRandomPtr = reinterpret_cast<VDSOHandlers::GetRandomType>(SymbolPtr);
+    x64::Handler_getrandom = x64::VDSO::getrandom;
+    ///< 32-bit doesn't have getrandom vdso
```
neobrain, line 493: Why does this comment start with `///<`?

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Went with the first example since semantically I don't like the looks of the second.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Removed the weird

Sonicadvance1:
VDSO patches for arm64 have now landed on the ML, allowing me to properly test the implementation on arm64. While the implementation may not be upstream, this allows me to test the uAPI regardless, which won't change.

When testing this I realized that we hadn't been actually using the host VDSO implementations and always falling back to non-VDSO. This is due to us setting up the VDSO definitions before fetching and setting the VDSO host symbols, resulting in the desync. Noticed this because running the vdso getrandom microbench didn't show any improvement initially.

With everything fixed and working the microbench shows some good uplift.
```
   vdso: 25000000 times in 1.496027236 seconds
   libc: 25000000 times in 7.924904096 seconds
syscall: 25000000 times in 7.872115351 seconds
```

This is now ready for review since I have confirmed it to work.

alyssarosenzweig:
Looks reasonable but will leave this for @neobrain

Sonicadvance1:
Rebased and merge conflict resolved.

Sonicadvance1:
Rebased and merge conflict resolved.

lioncash (approved):

PR #3951, opened 2024-08-13
Add a hack for multiple destinations & make good use of it

alyssarosenzweig:
Add a really dumb effectively zero line of code hack to bolt on multiple destinations to the IR. This lets us get rid of the far far far worse hack of register pairs - just switching from pairs to multidest removes a pile of moves from affected instructions. It then lets us implement pop properly for a nice perf win.


FEXCore/Source/Interface/IR/IR.json
```diff
@@ -316,42 +324,22 @@
         "HasSideEffects": true
       },
 
-      "GPRPair = CPUID GPR:$Function, GPR:$Leaf": {
+      "CPUID GPR:$Function, GPR:$Leaf, GPR:$OutEAX, GPR:$OutEBX, GPR:$OutECX, GPR:$OutEDX": {
```
Sonicadvance1, line 327: The main thing I want in this PR is a way to read our IR.json and know which arguments are input, output, or rmw. Even if it is purely a textual representation that nothing parses today.

Maybe something in each IR definition that uses this new output, or rmw members like:
```json
"ArgumentTypes": {
  "OutEAX": "Output",
  "OutEBX": "Output",
  "OutECX": "Output",
  "OutEDX": "Output",
},
```

Then in the case of like Pop where an input is RMW
```json
"ArgumentTypes": {
  "InoutAddr": "RMW",
},
```

And something like CASPair where there is a mixture:
GPR:$ExpectedLo, GPR:$ExpectedHi, GPR:$DesiredLo, GPR:$DesiredHi, GPR:$Addr, GPR:$OutLo, GPR:$OutHi
```json
"ArgumentTypes": {
  "ExpectedLo": "Input",
  "ExpectedHi": "Input",
  "DesiredLo": "Input",
  "DesiredHi": "Input",
  "Addr": "Input",
  "OutLo": "Output",
  "OutHigh": "Output",
},
```

Even though we don't parse any of this, I want to be able to come to the JSON in a few months time and not have to wrack my brain trying to remember which SSA arguments are input, output, and RMW, especially if the name of the argument doesn't always imply which one.

Then maybe in the future we can also add some validation to this, which will make reparsing this less of a nightmare as well.

Sonicadvance1 (commented):
Looks good. I'm mostly okay with this style. Just one feature that I want added

FEXCore/Source/Interface/IR/IR.json
alyssarosenzweig: That`s very fair, yeah! I will noodle on this. I don`t love that syntax but will do it if I can`t come up with anything better.

alyssarosenzweig: Reworked.

FEXCore/Source/Interface/Core/JIT/Arm64/BranchOps.cpp
```diff
@@ -427,10 +427,11 @@ DEF_OP(CPUID) {
...
 
   // Results are in x0, x1
-  // Results want to be in a i64v2 vector
-  auto Dst = GetRegPair(Node);
-  mov(ARMEmitter::Size::i64Bit, Dst.first, TMP1);
-  mov(ARMEmitter::Size::i64Bit, Dst.second, TMP2);
+  // Results want to be 4xi32 scalars
+  mov(ARMEmitter::Size::i32Bit, GetReg(Op->OutEAX.ID()), TMP1);
+  mov(ARMEmitter::Size::i32Bit, GetReg(Op->OutECX.ID()), TMP2);
+  lsr(ARMEmitter::Size::i64Bit, GetReg(Op->OutEBX.ID()), TMP1, 32);
```
Sonicadvance1, line 433: You'll want to change these `lsr` instructions in CPUID and XGETBV to `bfe` instructions instead. lsr has lower throughput than bfe.

alyssarosenzweig: Done. Performance should be the same on firestorm, I'm guessing this is a cortex quirk?

Sonicadvance1: Indeed. lsr > 4 goes down the M pipeline, bfe goes down all the integer pipelines.

alyssarosenzweig: Wacky wacky. Anyway, apparently I wasn't pushing the right branch :laughing: should be there now

Sonicadvance1 (approved):
Nice nice

PR #3952, opened 2024-08-14
Unwind again

Sonicadvance1:
Let's do it again. This time without a service. No ELF parsing, no symbol parsing. Let's make it as simple as possible k?

```bash
$ FEX_COREDUMP=1 FEXInterpreter sigsegv_test
RIP: 000055b5af6b51f4
RAX: 000055b5af6cb2b8 RBX: 000055b5af6cb2b0 RCX: 00007fffe3c9c84e RDX 0000000000000000
RSI: 00007fffe3a00fb0 RDI: 0000000000000001 RSP: 00007fffe39ffdf8 RBP 00007fffe39ffe00
R8:  0000000000000000 R9:  00007fffe3a006c0 R10: 0000000000000008 R11 0000000000000246
R12: 00007fffe3a006c0 R13: ffffffffffffff60 R14: 0000000000000000 R15 00007fffffdfeb40

                Stack trace of thread 219449:
                #0  0x000055b5af6b51f4 <sigsegv_test + 0x11f4>          sp=(0x00007fffe39ffdf8)
                #1  0x000055b5af6b56f2 <sigsegv_test + 0x16f2>          sp=(0x00007fffe39ffe10)
                #2  0x000055b5af6b56b5 <sigsegv_test + 0x16b5>          sp=(0x00007fffe39ffe30)
                #3  0x000055b5af6b568d <sigsegv_test + 0x168d>          sp=(0x00007fffe39ffe50)
                #4  0x000055b5af6b5665 <sigsegv_test + 0x1665>          sp=(0x00007fffe39ffe70)
                #5  0x000055b5af6b55b9 <sigsegv_test + 0x15b9>          sp=(0x00007fffe39ffe90)
                #6  0x00007fffe40eabb4 <libstdc++.so.6.0.33 + 0xeabb4>  sp=(0x00007fffe39ffeb0)
                #7  0x00007fffe3d29c3c <libc.so.6 + 0x129c3c>           sp=(0x00007fffe39fff80)
```

PR #3953, opened 2024-08-14
Breakout signal struct generators

Sonicadvance1:
Peeled off of #3952 to ensure these don't break CI.


Sonicadvance1:
Alright, found the silly bug in this, so now it passes CI again.

PR #3961, opened 2024-08-15
FEXCore: Disable vixl linking if vixl disasm or simulator is disabled

Sonicadvance1:
This was mostly there, just needed to remove some extraneous headers and only insert vixl in to the library list if the options were enabled.

### DIFF
--- a/Source/Common/CMakeLists.txt
+++ b/Source/Common/CMakeLists.txt
@@ -16,7 +16,7 @@ if (NOT MINGW_BUILD)
 endif()
 
 add_library(${NAME} STATIC ${SRCS})
-target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json FEXHeaderUtils)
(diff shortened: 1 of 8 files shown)


alyssarosenzweig (approved):

Source/Common/CMakeLists.txt
```diff
@@ -16,7 +16,7 @@ if (NOT MINGW_BUILD)
 endif()
 
 add_library(${NAME} STATIC ${SRCS})
-target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json json-maker FEXHeaderUtils)
+target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json json-maker FEXHeaderUtils vixl)
```
neobrain, line 19: Is this intended?

> only insert vixl in to the library list if the options were enabled.

PR #3962, opened 2024-08-16
HostFeatures: Removes vixl usage 

Sonicadvance1:
If we query the CPU flags ourselves then vixl is no longer a
compile-time or runtime dependency unless the vixl disassembler or
simulator is built.

A bit spicy from all the feature bits we need to load up.

### DIFF
--- a/Source/Windows/Common/CPUFeatures.cpp
+++ b/Source/Windows/Common/CPUFeatures.cpp
@@ -5,7 +5,6 @@
 #include <FEXCore/fextl/fmt.h>
 
 #include <windows.h>
-#include <aarch64/cpu-aarch64.h>
 
 #include "CPUFeatures.h"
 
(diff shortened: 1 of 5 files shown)


Source/Windows/Common/CPUFeatures.cpp
```diff
@@ -26,30 +25,32 @@ uint64_t ReadRegU64(HKEY Key, const char* Name) {
...
-template<typename RegType>
-void AddRegFeatures(vixl::CPUFeatures& Features, HKEY Key, const char* Name) {
-  return Features.Combine(RegType(ReadRegU64(Key, Name)).GetCPUFeatures());
-}
 } // namespace
 
 namespace FEX::Windows {
+class CPUFeaturesFromRegistry final : public FEX::CPUFeatures {
+public:
+  CPUFeaturesFromRegistry(HKEY Key) {
```
lioncash, line 33:
```suggestion
  explicit CPUFeaturesFromRegistry(HKEY Key) {
```

lioncash (approved):
One minor nit, but very nice overall

PR #3963, opened 2024-08-16
Clean up our JSON dependencies

alyssarosenzweig:
The upstreams are abandoned, these are trivial enough that it's easier to just bring them in tree. This does that, merging them into a tiny single tiny-json library that only supports what we care about and makes the problem disappear for packaging.


Sonicadvance1 (approved):
Whoa, somehow this managed to break our CI for PRs backing out to main. I'm just going to push this so it stops breaking us.

PR #3964, opened 2024-08-16
Install libraries in the correct location

davide125:
Some distributions (e.g. Fedora) use `lib64` instead of `lib` for libraries, this ensures we always install in the correct location instead of hardcoding `lib`.


Sonicadvance1 (approved):
Oh cool! Wasn't aware of this, thanks!

Sonicadvance1:
CI flake is unrelated to this PR. Merging!

PR #3968, opened 2024-08-17
binfmt_misc: Support systemd binfmt_misc

### LINKED ISSUE #2417: use binfmt.d rather than binfmts for better compatibility
binfmts is only available on debian based distro. which isn't friendly to other distro. i prefer switch to [binfmt.d](https://www.freedesktop.org/software/systemd/man/binfmt.d.html) (which is a part of systemd, available on most of distro) and keep binfmts for debian only

Sonicadvance1:
Adds support for binfmt_misc through systemd configuration paths. Their configuration files are basically the raw kernel interface description in a .conf file, quite a bit more simple than the legacy debian path.

Default enable this path since systemd is the expected default arrangement these days.

Fixes #2417

Might want to view this PR without whitespace, since there was a large indent.


alyssarosenzweig (approved):
looks good, I've sent this to the fedora folk as well for review

Conan-Kudo (approved):

Data/binfmts/CMakeLists.txt
```diff
@@ -13,5 +13,25 @@ function(GenBinFmt Name)
...
+
+  # Then install the configured binfmt
+  install(
+    FILES ${CMAKE_BINARY_DIR}/Data/binfmts/${FMT_NAME}.conf
+    DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/binfmt.d/)
+endfunction()
+
+if (ENABLE_SYSTEMD_BINFMTMISC)
+  GenBinFmt_Systemd(FEX-x86.conf.in)
+  GenBinFmt_Systemd(FEX-x86_64.conf.in)
```
neobrain, line 33: Inlining the function seems more readable overall since it's much shorter and doesn't require you to mentally substitute the function argument:
```suggestion
if (ENABLE_SYSTEMD_BINFMTMISC)
  configure_file(FEX-x86.conf.in ${CMAKE_BINARY_DIR}/Data/binfmts/FEX-x86.conf)
  configure_file(FEX-x86_64.conf.in ${CMAKE_BINARY_DIR}/Data/binfmts/FEX-x86_64.conf)

  install(
    FILES ${CMAKE_BINARY_DIR}/Data/binfmts/FEX-x86.conf ${CMAKE_BINARY_DIR}/Data/binfmts/FEX-x86_64.conf
    DESTINATION ${CMAKE_INSTALL_PREFIX}/lib/binfmt.d/)
```

This actually applies to the legacy path as well already. Maybe we can additionally set a `BINFMT_CONFIG_PREFIX` variable conditionally to `.conf`/`` and use the same code for both?

CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ option(ENABLE_LIBCXX "Enables LLVM libc++" FALSE)
 option(ENABLE_CCACHE "Enables ccache for compile caching" TRUE)
 option(ENABLE_VIXL_SIMULATOR "Forces the FEX JIT to use the VIXL simulator" FALSE)
 option(ENABLE_VIXL_DISASSEMBLER "Enables debug disassembler output with VIXL" FALSE)
+option(ENABLE_SYSTEMD_BINFMTMISC "Enables the systemd binfmt_misc install targets" TRUE)
```
neobrain, line 31: For better or worse, #3908 kicked off a route where options that trigger mutually exclusive paths should use a `USE_` prefix instead of `ENABLE_`.

Subjectively these options should be named such that they are FALSE by default as well (to better indicate what a "normal" setup looks like), so in this case it would be something like:

```suggestion
option(USE_LEGACY_BINFMTMISC "Uses legacy method of setting up binfmt_misc" FALSE)
```

Sonicadvance1: Sure, renamed.

Data/binfmts/CMakeLists.txt
Sonicadvance1: Sure. inlined.

Issue #3431, opened 2024-02-20
GDB stub non-functional due to regression


neobrain:
Merging #3321 regressed the GDB stub functionality: FEXLoader will now crash once gdb connects to it (or gdb itself hangs due to protocol errors).

To reproduce, run `FEXLoader -G <any x86_64 executable>` and then run the gdb command printed in the console (e.g. `gdb-multiarch -ex "target extended-remote /run/user/1000/FEX_gdbserver/650259-gdb"`).

Sonicadvance1:
I have some WIP code that fixes how this works, basically a couple of bad assumptions that were broken.
Also fixed the issue that a faulting thread wouldn't reconstruct the RIP and it would only be the start of a block, so slightly improved.
Will clean this up and open a PR once the remaining PRs for code reformatting land.

PR #3969, opened 2024-08-17
Removes GDBSymbol integration

Sonicadvance1:
This was only ever wired up to the AOT interface, so was mostly broken.

I've never used this, its integration wasn't ever fully fleshed out, just get rid of it.


alyssarosenzweig (approved):

Conan-Kudo (approved):

neobrain (changes requested):
This gets a -1 from me. Our debugging infrastructure has been continuously regressing (complete breakage of the already limited GDB stub (#3431), removal of interpreter (#3128), removal of x86 support (#3134), ...). Meanwhile, improvements that could balance this out have been progressing slowly or stalled entirely.

The GDB JIT integration is one of the few remaining tools we have to debug complex guest code, so I'd rather keep it until our debugging story meaningfully improves.

neobrain:
> This was only ever wired up to the AOT interface, so was mostly broken.

The feature was working fine in the past, has anything changed? Off the top of my head, I think the GDB JIT integration code just generates source map files whenever the guest loads libraries (using the same interfaces as AOT uses), but it shouldn't be a no-op unless we broke something recently.

PR #3976, opened 2024-08-19
Improvements from bytemark "huffman"

alyssarosenzweig:
score for the subtest improved 26% across the series, which isn't too bad .. this test has the biggest gap to native for me. bit of everything to get there, most of this should benefit other workloads too


Sonicadvance1 (approved):
Nice optimizations!

PR #3980, opened 2024-08-19
Optimize AVX load/store with ldp/stp

alyssarosenzweig:
Big wins for glibc memcpy(). Didn't help the bytemark case I was looking at but surely helps *something*


Sonicadvance1:
Needs a rebase and the submodule changes removed

Sonicadvance1 (approved):
nice nice nice

PR #3982, opened 2024-08-20
Add Qt-based config editor

neobrain:
## Background

[(Screenshots)](#Screenshots)

Dear ImGui can be a great tool for quick prototyping and overlay GUIs. Conversely, it's also an unsuitable option for creating GUIs of moderate complexity without sacrificing usability and accessibility. This is acknowledged by the author of Dear ImGui himself, and will likely never change:

>  I do know that Dear ImGui unfortunately does not align with requirements of accessible software. It was designed as a technical/debug tooling software to be used as an overlay over 3d/graphics applications which themselves tends to have little to no accessibility features. It's not really my fault that people have started using it for desktop-ey applications. If you do use it for a tool please be mindful the tool won't be accessible to most screen readers

Historically Dear ImGui may have been a suitable tool for FEX when there was still a [built-in debugger](https://github.com/FEX-Emu/FEX/pull/627). Nowdays we only use it for FEXConfig, for which it provides no benefit over proper desktop toolkits, while having all of the downsides of an immediate-mode framework that isn't designed for regular desktop applications.

This PR adds a rewrite of FEXConfig based on Qt, a widely available framework with decades of development progress. During the rewrite, I've also cleaned up and restructured some of the settings to make things easier to configure overall.

## Improvements over current FEXConfig

* Look and Feel matching the system:
    * Font rendering: Anti-aliasing, font type, size, ...
    * Keyboard navigation (tabs, file menu, ...)
    * Styling (color theme, native controls, ...)
* Architectural support for accessibility (screen reader, keyboard navigation, ...)
* Native filesystem browser (file picker available in more places than before)
    * All path settings can be configured with this dialog now
* More robust widgets:
    * E.g. dialogs are somehow really glitchy with Dear ImGui (might be a bug in how FEX uses them?)
* Can be built without a custom ImGui fork, opening up the way for [dropping another External](https://github.com/FEX-Emu/FEX/issues/2996)
* AppConfig support made more straightforward
    * Not sure if anyone uses this, since current FEXConfig behaves very quirky
    * Qt provides "tri-state" checkboxes that clearly highlight which settings aren't set in an AppConfig
    * It's easier to delete individual settings in the "Advanced" tab now
    * Overall, this makes it much easier to create minimal config files in a GUI

## Screenshots

2024_08_20_fexqonfic1
2024_08_20_fexqonfic2
2024_08_20_fexqonfic3
2024_08_20_fexqonfic4


Conan-Kudo:
Thank you so much for this! 🥇

Source/Tools/CMakeLists.txt
```diff
@@ -3,6 +3,11 @@ add_subdirectory(CommonTools)
 if (NOT MINGW_BUILD)
   if (BUILD_FEXCONFIG)
     add_subdirectory(FEXConfig/)
+
+    find_package(Qt5 COMPONENTS Qml Quick Widgets)
+    if (Qt5_FOUND)
+      add_subdirectory(FEXQonfig/)
+    endif()
```
Conan-Kudo, line 10: Can we please use Qt 6 instead here? Qt 5 OSS is effectively EOL.

Conan-Kudo (changes requested):

Source/Tools/FEXQonfig/main.qml
```diff
@@ -0,0 +1,669 @@
...
+
+    function refreshUI() {
+        refreshCache = !refreshCache
+    }
+
+    FileDialog {
+        id: openFileDialog
+        title: qsTr("Open FEX configuration")
+        nameFilters: [ "Config files(*.json)", "All files(*)" ]
+        // TODO: Set default folder to ~/.fex-emu?
```
jvoisin, line 37: Or even better, `~/.config/fex-emu/`

Source/Tools/FEXQonfig/main.qml
```diff
@@ -0,0 +1,669 @@
...
+                }
+
+                ConfigCheckBox {
+                    text: qsTr("Disable JIT optimization passes")
+                    config: "O0"
+                }
+            }
+        }
+
+        // Advanced settings
```
jvoisin, line 615: It would also be nice to add a small help/description for them I think.

Source/Tools/CMakeLists.txt
neobrain: Sadly Kubuntu 24.04 LTS ships Qt 5 by default and figuring out what Qt 6 packages are needed to compile (and to not make it use the fallback theme!) seems to be nontrivial.

Supporting both simultaneously (and preferring Qt 6 over 5) should hopefully be possible without too much extra work though. The main breaking changes are in the Dialog APIs, which we should be able to handle via abstraction however.

Source/Tools/FEXQonfig/main.qml
neobrain: You're not wrong. I'm not actually sure why we aren't yet using `XDG_CONFIG_HOME` for the config.

bylaws: ^ currently FEX uses it if it's explicitly set, but defaults to ~/.fex-emu

Source/Tools/CMakeLists.txt
neobrain: I created a compatibility module to deal with the dialogs (surprisingly tricky to make Qt conditionally import the right one). Both Qt 5 and Qt 6 are supported now!

My testing with Qt 6 was limited due to what Ubuntu currently packages, so feel free to give it a spin if you want :)

Source/Tools/CMakeLists.txt
```diff
@@ -3,6 +3,14 @@ add_subdirectory(CommonTools)
 if (NOT MINGW_BUILD)
   if (BUILD_FEXCONFIG)
     add_subdirectory(FEXConfig/)
```
Conan-Kudo, line 5: Can we have a conditional so that we can only build this variant and not the other?

Conan-Kudo (changes requested):

Source/Tools/CMakeLists.txt
neobrain: Sounds reasonable. `BUILD_FEXCONFIG` is now called `USE_FEXCONFIG_TOOLKIT` and can be set to either ""/"qt"/"imgui". Since it's an exclusive choice, FEXQonfig has been renamed to FEXConfig.

neobrain:
This is on feature-parity with the existing FEXConfig now.

We could make it the default already but the CI runners don't have Qt libraries installed (qtdeclarative5-dev).

Sonicadvance1:
Image_2024-08-27_22-49-33
The rootfs list should scroll, since it consumes the whole window otherwise and it's hard to tell that there is more available in the window.

Sonicadvance1 (approved):

neobrain:
>  The rootfs list should scroll, since it consumes the whole window otherwise and it's hard to tell that there is more available in the window.

Ah yes, I limited the list's maximum height to around 4-6 items now (depending on the system theme). After that, it will display a scroll bar while leaving the two buttons visible.

Sonicadvance1 (approved):

CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ CHECK_INCLUDE_FILES ("gdb/jit-reader.h" HAVE_GDB_JIT_READER_H)
 option(BUILD_TESTS "Build unit tests to ensure sanity" TRUE)
 option(BUILD_FEX_LINUX_TESTS "Build FEXLinuxTests, requires x86 compiler" FALSE)
 option(BUILD_THUNKS "Build thunks" FALSE)
-option(BUILD_FEXCONFIG "Build FEXConfig, requires SDL2 and X11" TRUE)
+set(USE_FEXCONFIG_TOOLKIT "imgui" CACHE STRING "If set, build FEXConfig (qt or imgui)")
```
Conan-Kudo, line 10: Do we want to make `qt` the default before merging it?

Sonicadvance1: We'll need to update our CI machines before it can be switched to default. Setting the cmake option until then should be fine.

neobrain: Also probably doesn't hurt to ship this as opt-in for one release, just in case edge case bugs come up :)

Source/Tools/FEXQonfig/main.qml
neobrain: Yup, though for now I'm just aiming to rebuild functionality that already exists (plus some things that were easy to add).

We already have descriptions written down in the Config.json file, but they're not readily available in C++ code yet, so this would need more work.

neobrain:
Gave this a final spin with a proper Qt 6 + Plasma 6 installation on openSUSE, working all fine. Let's do this!

Screenshot_20240829_115411

PR #3955, opened 2024-08-14
Rearrange SRA to let us coalesce cmpxchg moves

alyssarosenzweig:
The biggest smallest MR

PR #3983, opened 2024-08-20
Arm64: Allow directly correlating an ARM register back to an x86 register

Sonicadvance1:
Fixes a bug that is getting introduced in to #3955 when it rearranged register allocation.


alyssarosenzweig (approved):
+1, this is a nice refactor. After this is merged, do we still need the other diff you posted?

Sonicadvance1:
> +1, this is a nice refactor. After this is merged, do we still need the other diff you posted?

Nah, this is the diff in PR form.

PR #3984, opened 2024-08-20
FEXCore: Splits up atomic enablement checks

Sonicadvance1:
PR #3980 is adding a feature to merge loadstores in to paired loadstores, but it was using the incorrect atomic check to determine if it can safely merge them or not. It was using the GPR atomic check instead of the vector atomic check. While this would improve performance on Apple Silicon with its hardware TSO implementation, it would have had zero impact on Cortex and Oryon.

Instead split out the three config options to live as a boolean check in the ContextImpl similar to how we disable "AtomicTSOEmulation". Removing the various configs in the JIT and CPUID so that it queries from the same context. This makes it clearer that if you are wanting the current active configuration for memcpy, vector, or general atomic TSO emulation, you should query one of those three getters.

This also fixes a weird edge case bug in the arm64 JIT where you could have TSO emulation disable, but still have vector TSO enabled partially. Just because half a config wasn't checked in {Load,Store}MemTSO for vectors. If the global "TSOEnabled" option is disabled then TSO should always be disabled.

Alyssa will be able to pull this in to #3980 once merged and get the performance uplift on Cortex and Oryon, since our default configuration is to have vector and memcpy TSO emulation disabled.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -216,6 +216,8 @@ public:
     FEX_CONFIG_OPT(Is64BitMode, IS64BIT_MODE);
     FEX_CONFIG_OPT(TSOEnabled, TSOENABLED);
     FEX_CONFIG_OPT(TSOAutoMigration, TSOAUTOMIGRATION);
+    FEX_CONFIG_OPT(VectorTSOEnabled, VECTORTSOENABLED);
+    FEX_CONFIG_OPT(MemcpySetTSOEnabled, MEMCPYSETTSOENABLED);
     FEX_CONFIG_OPT(ABILocalFlags, ABILOCALFLAGS);
     FEX_CONFIG_OPT(AOTIRCapture, AOTIRCAPTURE);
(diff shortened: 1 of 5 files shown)


alyssarosenzweig (approved):

neobrain:
Please let's avoid insta-merging changes like these so we don't need to do fixup-PRs mere 2 hours later.

For reference, to fetch PR patches for local testing, GitHub let's you do so via `git fetch origin pull/3984/head`.

alyssarosenzweig:
> Please let's avoid insta-merging changes like these so we don't need to do fixup-PRs mere 2 hours later (#3985).
> 
> For reference, to fetch PR patches for local testing, GitHub let's you do so via `git fetch origin pull/3984/head`.

FWIW this PR was fine, the bug was in https://github.com/FEX-Emu/FEX/pull/3980 and the bug was found by inspection, the only symptom was benchmarks that weren't as good as one would like

neobrain:
> FWIW this PR was fine, the bug was in https://github.com/FEX-Emu/FEX/pull/3980 and the bug was found by inspection, the only symptom was benchmarks that weren't as good as one would like

Noted, thanks for the correction!

PR #3986, opened 2024-08-21
Optimize test

alyssarosenzweig:


Sonicadvance1 (approved):
Nice little optimization! This pattern shows up all over the place in x86 code. So it should remove little paper-cuts all over the place.

PR #3991, opened 2024-08-22
Arm64: Fixes SIGBUS handler for FEAT_LRCPC

Sonicadvance1:
When https://github.com/FEX-Emu/FEX/pull/3899 refactored some of this code, it had split an if-else chain
in to two independent if-else chains. Turns out there was a minor
dependency between them. Add back the few checks necessary to ensure
that FEAT_LRCPC falls through before hitting the FEAT_LSE checks.

This is necessary because their instructions share an encoding so the
masks between the two instruction classes overlap slightly.


neobrain (approved):
Oof. Good find!

Issue #3987, opened 2024-08-21
FEXRootFSFetcher does not work in a docker image build step.


Witixin1512:
Currently, I am unable to run `yes 1 | FEXRootFSFetcher` within a docker build step, as the output is not a TTY. In such a case, the tool attempts to open a graphical interface and fails to run.

A possible fix would be introducing an environment variable or argument to the program to force it to run in terminal mode.

Sonicadvance1:
Until a specific command line option exists to force the terminal mode, I might recommend the `-y` option. Unless we're missing something for that.

Witixin1512:
I want to say that for my usecase `yes 1` is what I need, as otherwise the prompt to extract the squashfs file goes unanswered.
I'm not too well versed in the internals of the tool to be able to tell if `-y` should make that choice for you or not.

Sonicadvance1:
> I want to say that for my usecase `yes 1` is what I need, as otherwise the prompt to extract the squashfs file goes unanswered. I'm not too well versed in the internals of the tool to be able to tell if `-y` should make that choice for you or not.

You might want the `-x` option in addition then

Sonicadvance1:
A UI path can now be forced with the `--force-ui`option

PR #3992, opened 2024-08-22
FEXRootFSFetcher: Allow UI override through options

Sonicadvance1:
For #3987

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -44,6 +44,14 @@ fextl::vector<fextl::string> RemainingArgs;
 std::string DistroName {};
 std::string DistroVersion {};
 
+enum class UIOverrideOption {
+  Default,
+  TTY,
+  Zenity,
+};
+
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -85,6 +95,15 @@ void ParseArguments(int argc, char** argv) {
     DistroVersion = Options["distro_version"];
   }
 
+  if (Options.is_set_by_user("force_ui")) {
+    auto Option = Options["force_ui"];
+    if (Option == "tty") {
+      UIOption = UIOverrideOption::TTY;
+    } else if (Option == "zenity") {
+      UIOption = UIOverrideOption::Zenity;
```
neobrain, line 103: How does zenity decide whether or not to spawn a graphical dialog?

FexRootFSFetcher always prompts via CLI for me, but when invoking `zenity` on its own I'm getting a graphical dialog.

neobrain: Ah right, makes sense. Thanks!

neobrain (approved):

PR #3997, opened 2024-08-22
SignalDelegator: Refactor how thread local data is stored

Sonicadvance1:
Two primary things here:
- Remove the static `GlobalDelegator`
- Move the thread_local SignalDelegator::ThreadState information directly in to ThreadStateObject

Having the ThreadStateObject and the SignalDelegator information disjoint was confusing but was required when we didn't have any object in the frontend that could have its own independent data. Since we fixed this with the `ThreadStateObject` type we can now move this over.

The `GlobalDelegator` object is now instead stored in `ThreadStateObject` instead.

Instead of using a thread_local variable, we now just consume 8-bytes of the signal alt-stack since the kernel gives us that information about where it lives. This then converts all the thread_local usage to use either the passed in CPU state if it exists, or fetching it from the alt-stack offset.

Very minor changes in behaviour here, will help when trying to improve FEX's behaviour around signals.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
@@ -49,30 +49,17 @@ __attribute__((naked)) static void sigrestore() {
 
 constexpr static uint32_t X86_MINSIGSTKSZ = 0x2000U;
 
-// We can only have one delegator per process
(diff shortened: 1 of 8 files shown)


alyssarosenzweig (approved):

alyssarosenzweig:
conflicts, needs a rebase

Sonicadvance1:
rebossed, conflict resalved

PR #3998, opened 2024-08-23
HostFeatures: Moves MIDR querying to the frontend

Sonicadvance1:
The MIDR querying is inherently OS specific and needs a bit of special casing. Instead let the frontend inform FEXCore how many CPU cores there are and their MIDRs instead.

This lets us keep the Linux specific code in the frontend.

### DIFF
--- a/Source/Windows/Common/CPUFeatures.cpp
+++ b/Source/Windows/Common/CPUFeatures.cpp
@@ -2,6 +2,7 @@
 
 #include <FEXCore/Core/Context.h>
 #include <FEXCore/Core/HostFeatures.h>
+#include <FEXCore/Utils/CPUInfo.h>
 #include <FEXCore/fextl/fmt.h>
 
 #include <windows.h>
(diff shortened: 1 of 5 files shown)


lioncash (approved):

Source/Windows/Common/CPUFeatures.cpp
```diff
@@ -57,7 +65,13 @@ FEXCore::HostFeatures CPUFeatures::FetchHostFeatures(bool IsWine) {
 
   RegCloseKey(Key);
 
-  return FEX::FetchHostFeatures(Features, !IsWine, CTR, MIDR);
+  auto HostFeatures = FEX::FetchHostFeatures(Features, !IsWine, CTR, MIDR);
+  if (IsWine) {
+    FEX::FillMIDRInformationViaLinux(&HostFeatures);
```
bylaws, line 70: You can drop this path, will implement a generic way as a follow up

Sonicadvance1: Sure thing, removed that path for you

bylaws (approved):

PR #3999, opened 2024-08-23
SignalDelegator: Make sure to defer a signal if the guest signal mask desires

Sonicadvance1:
Fairly trivial because we already supported deferring these signals. We
had just failed the final step of blocking the signal if we can't block
the signal (like with SIGSEGV, SIGBUS, etc).

Once the guest unblocks the signal mask with sigprocmask, the signal
will fire again.

Apparently older glibc relied on this behaviour for signal raising which
this fixes.

### DIFF
--- a/unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
+++ b/unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
@@ -1,0 +1,55 @@
+// A test where a signal is masked, a value is set that the signal handler will overwrite, and then the signal is unmasked to allow it to
+// fire. FEX-Emu had a bug where it wasn't properly deferring signals from sigprocmask if one of the signals was masked by the guest application.
+// In older glibc versions (glibc-2.26), the `raise` implementation would block signals, tgkill, and then unblock signals,
+// expecting the signal to fire once the signals were unblocked.
+
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <catch2/generators/catch_generators_range.hpp>
+
+#include <signal.h>
+#include <unistd.h>
+
+static uint32_t CheckValue {};
+void sig_handler(int signum, siginfo_t* info, void* context) {
+  REQUIRE(CheckValue == 2);
+  CheckValue = 0x1;
+}
+
+static void RaiseSignal(int Signal) {
+  sigset_t Prev {};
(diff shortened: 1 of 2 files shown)


unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
```diff
@@ -0,0 +1,58 @@
+// A test where a signal is masked, a value is set that the signal handler will overwrite, and then the signal is unmasked to allow it to
+// fire. FEX-Emu had a bug where it wasn't properly deferring signals from sigprocmask if one of the signals was the few that FEX must never
+// block. In older glibc versions (glibc-2.26), the `raise` implementation would block signals, tgkill, and then unblocked signals,
```
neobrain, line 3: This sentence is a bit awkward, not sure I understand it correctly but maybe this is better?

```suggestion
// fire. FEX-Emu had a bug where it wasn't properly deferring signals from sigprocmask if one of the signals was masked by the guest application. 
// In older glibc versions (glibc-2.26), the `raise` implementation would block signals, tgkill, and then unblock signals,
```

unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
```diff
@@ -0,0 +1,58 @@
...
+  sigfillset(&New);
+  int Ret = sigprocmask(SIG_BLOCK, &New, &Prev);
+
+  REQUIRE(Ret != -1);
+
+  pid_t Pid = ::getpid();
+  pid_t Tid = ::gettid();
+
+  // Try to raise the signal, even though it is blocked
+  Ret = tgkill(Pid, Tid, Signal);
```
neobrain, line 33: Reduce some noise:
```suggestion
  // Try to raise the signal, even though it is blocked
  Ret = tgkill(::getpid(), ::gettid(), Signal);
```

unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
Sonicadvance1: Switched over to the the comment

unittests/FEXLinuxTests/tests/signal/sigtest_defer.cpp
Sonicadvance1: changed

neobrain (approved):

PR #4002, opened 2024-08-23
OpcodeDispatcher: Allow x86 code to read CNTVCT on ARM64EC

bylaws:
Required by newer insider preview versions, I noticed many crashes with this exception number and QueryPeformanceCounter in the backtrace, testing with XTA found it to not be passed through and instead write the host CNTVCT (unscaled) into RAX. No other registers seem to be affected.


Sonicadvance1 (approved):
Weird. Seems like a lot of overhead to eat for rdtsc scaling emulation on their side.

Sonicadvance1:
Oh, I inverted the comment while reading it. Okay, this being the unscaled path makes way more sense.

PR #4003, opened 2024-08-24
Frontend: short-circuit code generation on invalid instructions with multiblock

Sonicadvance1:
A source of overhead with multiblock is hitting instructions through a conditional branch that can never be executed. Usually AVX512 instructions in glibc. This causes us to emit partial blocks for a ton of targets that will never get executed.

Instead, when we have multiblock enabled, if a block hits an instruction encoding we don't support, then remove all the decoded instructions from the block and early terminate it if it isn't the entry block. This resolves the issue of emitting a bunch of IR and code for blocks never executed.

If the block of code has an invalid instruction in the entry block for decoding then it'll still emit code up to the invalid instruction and raise a SIGILL. This has the potential for generating some additional blocks of code if a game is abusing SIGILL, but since that's unlikely it's a good trade-off.

Also removes a few log instructions that don't really provide anything anymore and just show up as confusing messages when multiblock is enabled.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -598,6 +598,11 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
       uint64_t InstsInBlock = Block.NumInstructions;
 
+      if (InstsInBlock == 0) {
+        // Special case for an empty instruction block.
+        Thread->OpDispatcher->ExitFunction(Thread->OpDispatcher->_EntrypointOffset(IR::SizeToOpSize(GPRSize), Block.Entry - GuestRIP));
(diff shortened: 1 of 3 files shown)


bylaws:
Might also want to drop the `ERROR_AND_DIE_FMT("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");` log message, at least when multiblock is enabled

Sonicadvance1:
> Might also want to drop the `ERROR_AND_DIE_FMT("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");` log message, at least when multiblock is enabled

Will need to do some refactoring for that. Since we would still want it to be fatal, just not in the first entry block. That can come afterwards, I've not run in to anything hitting it quite yet (I'm not testing that Wine version that does this yet)

PR #4004, opened 2024-08-25
Adds more syscall memory access checks

Sonicadvance1:
Does a decent number of syscalls, but not all of them. There's only so many I can do at a time before I start becoming mindnumb.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Msg.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Msg.cpp
@@ -27,6 +27,7 @@ void RegisterMsg(FEX::HLE::SyscallHandler* Handler) {
                               struct timespec tp64 {};
                               struct timespec* timed_ptr {};
                               if (abs_timeout) {
+                                FaultSafeUserMemAccess::VerifyIsReadable(abs_timeout, sizeof(*abs_timeout));
                                 tp64 = *abs_timeout;
                                 timed_ptr = &tp64;
                               }
@@ -41,6 +42,7 @@ void RegisterMsg(FEX::HLE::SyscallHandler* Handler) {
                               struct timespec tp64 {};
                               struct timespec* timed_ptr {};
                               if (abs_timeout) {
+                                FaultSafeUserMemAccess::VerifyIsReadable(abs_timeout, sizeof(*abs_timeout));
                                 tp64 = *abs_timeout;
                                 timed_ptr = &tp64;
                               }
@@ -54,6 +56,7 @@ void RegisterMsg(FEX::HLE::SyscallHandler* Handler) {
       mq_attr HostAttr {};
       mq_attr* HostAttr_p {};
       if ((oflag & O_CREAT) && attr) {
+        FaultSafeUserMemAccess::VerifyIsReadable(attr, sizeof(*attr));
         // attr is optional unless O_CREAT is set
         // Then attr can be valid or nullptr
         HostAttr = *attr;
@@ -65,6 +68,7 @@ void RegisterMsg(FEX::HLE::SyscallHandler* Handler) {
 
   REGISTER_SYSCALL_IMPL_X32(
     mq_notify, [](FEXCore::Core::CpuStateFrame* Frame, FEX::HLE::mqd_t mqdes, const compat_ptr<FEX::HLE::x32::sigevent32> sevp) -> uint64_t {
+      FaultSafeUserMemAccess::VerifyIsReadable(sevp, sizeof(*sevp));
       sigevent Host = *sevp;
       uint64_t Result = ::syscall(SYSCALL_DEF(mq_notify), mqdes, &Host);
       SYSCALL_ERRNO();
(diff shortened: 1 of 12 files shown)


neobrain:
Happy to wave this through since it's easy to refactor later, but here's an idea: Instead of having a two-step "verify"+"execute" that's potentially expensive, could we instead just toggle a thread-local flag that tells SignalDelegator to assume any SIGSEGV is triggered by invalid syscall arguments?

This would turn most code into something like this:
```
      if (Result != -1) {
        ExpectNoWriteFaults guard; // RAII wrapper around thread-local boolean

        // Any SIGSEGV will now be handled as if FaultSafeUserMemAccess::VerifyIsWritable failed.
        // It's important to keep the scope of the wrapper minimal hence

        for (size_t i = 0; i < Result; ++i) {
          events[i] = Events[i];
        }
        // ExpectNoWriteFaults destructor runs here and will unset the thread-local boolean
      }
```

Sonicadvance1:
> Happy to wave this through since it's easy to refactor later, but here's an idea: Instead of having a two-step "verify"+"execute" that's potentially expensive, could we instead just toggle a thread-local flag that tells SignalDelegator to assume any SIGSEGV is triggered by invalid syscall arguments?
> 
> This would turn most code into something like this:
> 
> ```c++
>       if (Result != -1) {
>         ExpectNoWriteFaults guard; // RAII wrapper around thread-local boolean
> 
>         // Any SIGSEGV will now be handled as if FaultSafeUserMemAccess::VerifyIsWritable failed.
>         // It's important to keep the scope of the wrapper minimal hence
> 
>         for (size_t i = 0; i < Result; ++i) {
>           events[i] = Events[i];
>         }
>         // ExpectNoWriteFaults destructor runs here and will unset the thread-local boolean
>       }
> ```

We will also need to return EFAULT in the case that the fault does occur in the future. Not sure how that would operate but something to think about.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Msg.cpp
```diff
@@ -80,17 +84,20 @@ void RegisterMsg(FEX::HLE::SyscallHandler* Handler) {
                               mq_attr* HostOld_p {};
 
                               if (newattr) {
+                                FaultSafeUserMemAccess::VerifyIsReadable(newattr, sizeof(*newattr));
                                 HostNew = *newattr;
                                 HostNew_p = &HostNew;
                               }
 
                               if (oldattr) {
+                                FaultSafeUserMemAccess::VerifyIsReadable(oldattr, sizeof(*oldattr));
```
neobrain, line 93: Does it need to be readable, though? We already check for writability below, and the variable isn't used before that.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Signals.cpp
```diff
@@ -239,6 +250,8 @@ void RegisterSignals(FEX::HLE::SyscallHandler* Handler) {
       siginfo_t* info64_p {};
 
       if (info) {
+        FaultSafeUserMemAccess::VerifyIsReadable(info, sizeof(*info));
+        info64 = *info;
```
neobrain, line 254: What's this added line doing here?

Source/Tools/LinuxEmulation/LinuxSyscalls/x64/Semaphore.cpp
```diff
@@ -29,8 +29,10 @@ void RegisterSemaphore(FEX::HLE::SyscallHandler* Handler) {
     case IPC_SET: {
       struct semid64_ds buf {};
       buf = *semun.buf;
+      FaultSafeUserMemAccess::VerifyIsReadable(semun.buf, sizeof(*semun.buf));
```
neobrain, line 32: Check should be moved before the previous line.

neobrain:
> We will also need to return EFAULT in the case that the fault does occur in the future. Not sure how that would operate but something to think about.

Ah yes, good point.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Signals.cpp
Sonicadvance1: This one was actually a bug, we were forgetting to copy over the data when provided

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Msg.cpp
Sonicadvance1: oop, removed

Source/Tools/LinuxEmulation/LinuxSyscalls/x64/Semaphore.cpp
Sonicadvance1: Moved before

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Signals.cpp
neobrain: Could you move functional changes to a different PR (or at least patch) for visibility? Hiding these in a NFC-sounding change is a bit surprising.

Sonicadvance1: Opened #4015 for the fix

neobrain (approved):

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Signals.cpp
Sonicadvance1: Rebased

PR #4011, opened 2024-08-26
Fix multiblock on ARM64EC

bylaws:


FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -967,8 +967,14 @@ void Decoder::BranchTargetInMultiblockRange() {
     TargetRIP &= 0xFFFFFFFFU;
   }
 
-  // If the target RIP is within the symbol ranges then we are golden
-  if (TargetRIP >= SymbolMinAddress && TargetRIP < SymbolMaxAddress) {
+  // If the target RIP is x86 code within the symbol ranges then we are golden
+  bool ValidMultiblockMember = TargetRIP >= SymbolMinAddress && TargetRIP < SymbolMaxAddress;
+
+#ifdef _M_ARM_64EC
+  ValidMultiblockMember = ValidMultiblockMember && !RtlIsEcCode(TargetRIP);
```
alyssarosenzweig, line 974: Nit: `ValidMultiblockMember &= !RtlIsEcCode(TargetRIP);`

alyssarosenzweig (approved):
+1 with the nit fixed

FEXCore/Source/Interface/Core/Frontend.cpp
bylaws: ah sure, suppose the ec check is cheap enough the lack of short circuiting doesn't really matter

alyssarosenzweig: Oh.. right.. short circuiting might help perf here... mm, I guess this is ok as is then.

alyssarosenzweig (approved):

PR #4013, opened 2024-08-26
Draft: global flag optimization

alyssarosenzweig:
depends on multiblock. few % bump on bytemark, draft because the impl is crap and bloating compile time by a ton


FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -32,47 +35,57 @@ desc: This is not used right now, possibly broken
...
 
+using Flags = uint8_t;
+
 struct FlagInfo {
-  // If set, all following fields are zero, used for a quick exit.
-  bool Trivial;
+  union {
+    uint64_t _Raw;
+
+    struct {
```
neobrain, line 44: Quick heads-up, Clang and GCC both fire a warning with `-Wpedantic` for this, since anonymous structs aren't valid C++. (Only anonymous unions are allowed!)

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
```diff
@@ -385,83 +390,160 @@ bool DeadFlagCalculationEliminination::EliminateDeadCode(IREmitter* IREmit, Ref
...
+    }
+
+    break;
+  }
+  case IR::OP_JUMP: FlagsRead = FlagsIn[ExitOp->Args[0].ID().Value]; break;
+  default: FlagsRead = ~0; break;
+  }
+
+  // Iterate the block in reverse
+  while (1) {
```
neobrain, line 460:
```suggestion
  while (true) {
```

FEXCore/Source/Interface/IR/Passes/RedundantFlagCalculationElimination.cpp
alyssarosenzweig: Entertaining. I'll probably drop that commit, clang generated stupid assembly for it anyway.

PR #4015, opened 2024-08-27
x32/Signals: Fixes bug in the sigqueue syscalls

Sonicadvance1:
In rt_sigqueueinfo and rt_tgsigqueueinfo we were failiny to pass along the provided siginfo_t information to the syscall.

Fix that


neobrain (approved):

neobrain:
Fast-tracking this one since it's innocent enough and already sat around as part of #4004 for two days.

PR #4018, opened 2024-08-28
LinuxEmulation: Moves guest signal frame generation to its own file

Sonicadvance1:
No functional change, just moving the code.

This is entirely freestanding from the rest of the signal delegator handling and mostly gets in the way when I'm working with the rest of the signal handling. Separate it out to improve readability.


alyssarosenzweig (approved):
ack, didn't review, I trust you to not have done a silly

PR #4023, opened 2024-09-01
CPUID: add missing Apple core part numbers

chadmed:
The Ultra-class SoCs are two Max-class SoCs connected via Apple's fabric, and thus use the same core revisions as the Max-class SoCs for both big and LITTLE cores.


Sonicadvance1 (approved):
Weird that Apple uses the Part number to represent different SoC's cores but sure.

Thanks for the patch!

chadmed:
The core variants in each SoC seem to be subtly different designs with silicon bug fixes and such. The M1 Pro Firestorm and Icestorm need different chicken bits set compared to the base M1, for example.

PR #4024, opened 2024-09-02
VDSO: Fixes a pretty nasty bug where we were never using the host VDSO

Sonicadvance1:
This must have happened during a refactor or something, but since we're making a copy of the function pointers, it would have only gotten the version /prior/ to loading host VDSO symbols.

Moves the VDSO thunk definition setting to the end after VDSO symbol definitions in order to get host VDSO symbols working again.

Cherry-picked out of #3946


neobrain:
What's the actual issue?

Sonicadvance1:
When setting the pointers in the `VDSODefinitions` array, we are copying the pointers provided eg, `FEX::VDSO::x64::Handler_time`.

These pointers default to the glibc handlers, but the `LoadHostVDSO` function parses out the host vdso symbols and updates those pointers to the VDSO handlers, eg `FEX::VDSO::x64::Handler_time` default points to `x64::glibc::time` but now points to `x64::VDSO::time`.

So the `VDSODefinitions` array needs to late initialize its pointers.

neobrain:
Ah, I was reading the description the wrong way, looking copies of `VDSODefinitions[n].ThunkFunction` rather than their assigned values. Thanks for the clarification!

neobrain (approved):

PR #4025, opened 2024-09-02
FEXQConfig: Add strict split-lock option

Sonicadvance1:
This was missed in the initial FEXQConfig implementation since it was implemented at the same time.

### DIFF
--- a/Source/Tools/FEXQonfig/main.qml
+++ b/Source/Tools/FEXQonfig/main.qml
@@ -618,6 +618,11 @@ ApplicationWindow {
                             onToggled: tsoButtonGroup.onClickedButton(2)
                         }
                     }
+
+                    ConfigCheckBox {
+                        text: qsTr("Enable non-tearing split-lock atomics")
+                        config: "StrictInProcessSplitLocks"
+                    }
                 }
             }
 


alyssarosenzweig (approved):

Source/Tools/FEXQonfig/main.qml
```diff
@@ -603,6 +603,12 @@ ApplicationWindow {
                                     tooltip: qsTr("Controls TSO emulation on memcpy/memset instructions")
                                     config: "MemcpySetTSOEnabled"
                                 }
+                                ConfigCheckBox {
+                                    leftPadding: 24
+                                    text: qsTr("Strict in-process split-lock atomics")
+                                    tooltip: qsTr("Controls split-lock emulation strictness")
```
neobrain, line 609: The tooltip doesn't add much. Either it should meaningfully elaborate on the feature, or it should be dropped.

Source/Tools/FEXQonfig/main.qml
```diff
@@ -603,6 +603,12 @@ ApplicationWindow {
                                     tooltip: qsTr("Controls TSO emulation on memcpy/memset instructions")
                                     config: "MemcpySetTSOEnabled"
                                 }
+                                ConfigCheckBox {
+                                    leftPadding: 24
+                                    text: qsTr("Strict in-process split-lock atomics")
```
neobrain, line 608: Was this intentionally positioned in the middle of the three other options? It looks rather awkward in practice. It should probably be below the other three options.

Also, could you confirm this is only used for normal TSO operation and that it is ignored for paranoid TSO?

neobrain (commented):
Looks good minus my remarks on the UI layout, but do we anticipate this being used often enough to expose it in the main tabs rather than solely under Advanced?

Source/Tools/FEXQonfig/main.qml
Sonicadvance1: Actually it is independent of the these TSO toggles. Moved it below the three radio buttons since it is still part of the memory model emulation.

Source/Tools/FEXQonfig/main.qml
Sonicadvance1: Changed the tooltip that it controls if split-locks can tear or not.

Sonicadvance1:
> Looks good minus my remarks on the UI layout, but do we anticipate this being used often enough to expose it in the main tabs rather than solely under Advanced?

I'm already using this a lot for testing purposes, very useful to look at the telemetry to inform if split-locks should be strict and that a crash should be verified with this option toggled.

Source/Tools/FEXQonfig/main.qml
neobrain: Does it still need a tooltip if the whole text is changed to "Enable non-tearing split-lock atomics"? Not sure how critical it is to mention them being in-process, since there's no other option that it could be confused with anyway.

Sonicadvance1: Sure, removed the tooltip and changed the option name

neobrain (approved):

PR #4027, opened 2024-09-03
Global flag optimizations

alyssarosenzweig:
Soup up our flag opt pass to optimize globally for a speed up with multiblock. Then use that framework to implement peephole passes to fuse comparisons & axflag into branches when the flags don't otherwise escape.

Perf #s aren't as impressive as i'd hoped, but Billy found a 0.9% improvement on geekbench on x13s on an early version of this series. Hopefully better now. At this point would like to cut my losses and get this in since it's no worse and XTA does something similar so I was going to get to it eventually anyways.


Sonicadvance1 (approved):

Sonicadvance1:
Nice little opts

PR #4030, opened 2024-09-03
Windows: Load per-application configs

bylaws:
Rather than duplicating a bunch of things, this splits up the preexisting config init code that is used for linux-side FEX consumers so that most can be reused for windows FEX.


Sonicadvance1 (approved):
Nice separation

PR #4031, opened 2024-09-03
Windows: Fill in per-core MIDR information

bylaws:


Sonicadvance1 (approved):

Source/Windows/Common/CPUFeatures.cpp
```diff
@@ -63,10 +57,14 @@ FEXCore::HostFeatures CPUFeatures::FetchHostFeatures(bool IsWine) {
   uint64_t CTR = ReadRegU64(Key, "CP 5801");
   uint64_t MIDR = ReadRegU64(Key, "CP 4000");
 
-  RegCloseKey(Key);
 
   auto HostFeatures = FEX::FetchHostFeatures(Features, !IsWine, CTR, MIDR);
-  FillMIDRInformation(&HostFeatures, MIDR);
+
+  for (uint32_t Idx = 0; Key; Key = OpenProcessorKey(++Idx)) {
```
neobrain, line 63: Is this intended to start at `Idx==1` ? All documentation I could find (which wasn't a lot) suggests the first processor is at 0.

It's a bit of a weird way to express the iteration pattern either way; initializing to 1 and using post-increment would have been clearer.

bylaws: It starts at zero because the first iteration of the loop is for core zero which has it's reg key already open from prior

neobrain: Ah... not the worst thing to do but rather subtle.

PR #4034, opened 2024-09-04
IR: fix scalar FMA tied sources

alyssarosenzweig:
needs to be modelled explicitly or else we lose information when translating


Sonicadvance1 (approved):
Very nice catch for my messup. Very wacky that the scalar FMA4 inserts in to the addend instead of dest still.
Screenshot 2024-09-04 at 04 53 38

PR #4036, opened 2024-09-05
ConstProp: stop pooling inline constants

alyssarosenzweig:
despite conventional wisdom, the robin map is slower than emitting extra nodes.

without multiblock, shaves around 2% off node. with multiblock, a bit less than 1% but still a win.


Sonicadvance1 (approved):
Very fancy

PR #4041, opened 2024-09-05
LinuxSyscalls: With poll syscall, ensure fds is writable only if nfds is not zero

Sonicadvance1:
The pointer is allowed to be null or garbage if the number of nfds passed in is zero.
Unify the x86 and x86-64 implementations to ensure consistency.

Fixes Warhammer 40k: Relics of War


alyssarosenzweig (approved):
nice!

PR #4042, opened 2024-09-06
CMake: Compile with system libraries for xxhash, Catch2, and fmt, if available

neobrain:
Instead of compiling these from bundled sources, check if the corresponding system libraries are installed first. This should help FEX packaging for Linux distributions (see https://github.com/FEX-Emu/FEX/issues/2996).


alyssarosenzweig (approved):
as a follow on it would be nice to make these not submodules anymore so our forks don't get pulled unless you actually need them. meson has "subproject wraps" for this, I'm not sure the cmake way.

neobrain:
> meson has "subproject wraps" for this, I'm not sure the cmake way.

CMake has FetchContent for this. I played around with it today and it works in principle, but I'm not sure if the change would come with unexpected side effects (by sole virtue of being *a change*, as opposed to some specific weakness I've observed in FetchContent).

Sonicadvance1 (approved):
Doesn't seem to break anything directly that I could see.

PR #4047, opened 2024-09-07
ARM64EC: Switch to exception-based native syscall dispatch

bylaws:
The previous approach assumed that ntdll wouldn't have been patched before FEX was loaded, but that isn't necessarily the case thanks to CEF. This takes the same approach used by xtajit which wine recently gained support for wine. The overhead to this isn't ideal but most syscalls aren't directly done through x86 code anyway and in the future FEX could forward unpatched thunks to their ARM variants at JIT time.

Tested with steam on windows (sandbox disabled), not entirely sure I like how KiUserExceptionDispatcher is called but this way seems the simplest

### DIFF
--- a/Source/Windows/ARM64EC/Module.S
+++ b/Source/Windows/ARM64EC/Module.S
@@ -127,3 +127,9 @@ direct_syscall:
   add x11, x11, x9
 end:
   ret
+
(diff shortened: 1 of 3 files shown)


Sonicadvance1 (approved):
Very interesting

PR #4048, opened 2024-09-07
FEXCore: Convert Base tables over to constexpr

Sonicadvance1:
Only doing the single table for review purposes. Once reviewed I will hammer out the remaining tables.

Similar to #3320, most of the OpcodeDispatcher tables can be consteval and made to be a compile time constant. This just requires shuffling the code slightly. The idea is to get almost all of the table setup out of the `InstallOpcodeHandlers` function and instead only install the handlers that change based on 32-bit or 64-bit, just like the x86 tables we also did.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
@@ -1,0 +1,107 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include "Interface/Core/OpcodeDispatcher.h"
+
+namespace FEXCore::IR {
+constexpr inline std::tuple<uint8_t, uint8_t, X86Tables::OpDispatchPtr> OpDispatch_BaseOpTable[] = {
+  // Instructions
+  {0x00, 6, &OpDispatchBuilder::Bind<&OpDispatchBuilder::ALUOp, FEXCore::IR::IROps::OP_ADD, FEXCore::IR::IROps::OP_ATOMICFETCHADD, 0>},
+
+  {0x08, 6, &OpDispatchBuilder::Bind<&OpDispatchBuilder::ALUOp, FEXCore::IR::IROps::OP_OR, FEXCore::IR::IROps::OP_ATOMICFETCHOR, 0>},
+
+  {0x10, 6, &OpDispatchBuilder::ADCOp<0>},
+
+  {0x18, 6, &OpDispatchBuilder::SBBOp<0>},
+
+  {0x20, 6, &OpDispatchBuilder::Bind<&OpDispatchBuilder::ALUOp, FEXCore::IR::IROps::OP_ANDWITHFLAGS, FEXCore::IR::IROps::OP_ATOMICFETCHAND, 0>},
+
+  {0x28, 6, &OpDispatchBuilder::Bind<&OpDispatchBuilder::ALUOp, FEXCore::IR::IROps::OP_SUB, FEXCore::IR::IROps::OP_ATOMICFETCHSUB, 0>},
+
+  {0x30, 6, &OpDispatchBuilder::Bind<&OpDispatchBuilder::ALUOp, FEXCore::IR::IROps::OP_XOR, FEXCore::IR::IROps::OP_ATOMICFETCHXOR, 0>},
+
+  {0x38, 6, &OpDispatchBuilder::CMPOp<0>},
+  {0x50, 8, &OpDispatchBuilder::PUSHREGOp},
+  {0x58, 8, &OpDispatchBuilder::POPOp},
+  {0x68, 1, &OpDispatchBuilder::PUSHOp},
+  {0x69, 1, &OpDispatchBuilder::IMUL2SrcOp},
+  {0x6A, 1, &OpDispatchBuilder::PUSHOp},
+  {0x6B, 1, &OpDispatchBuilder::IMUL2SrcOp},
+  {0x6C, 4, &OpDispatchBuilder::PermissionRestrictedOp},
+
+  {0x70, 16, &OpDispatchBuilder::CondJUMPOp},
+  {0x84, 2, &OpDispatchBuilder::TESTOp<0>},
+  {0x86, 2, &OpDispatchBuilder::XCHGOp},
+  {0x88, 4, &OpDispatchBuilder::MOVGPROp<0>},
+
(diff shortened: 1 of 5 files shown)


Sonicadvance1:
> This is neat, but is there measurable and significant benefit to either binary size or startup time to this?
I had some minor stats, it removed a couple hundred instructions from initialization, but there were also memcpy's there so it would be hard to count. It would be a startup time reduction when all tables are converted over to constexpr, but this singular table alone was impossible to see through the noise. Binary size change wasn't much, couple hundred bytes removed, again because it's such a small table.

> If not I think I'd rather keep the more compact code. This will also be a lot of code to review otherwise.
More compact code? It's the same code just moved.
- `BaseOpTable` moved to the constexpr table lambda constructor
- `BaseOpTable_32` and `BaseOpTable_64` moved to the `InitializeBaseTables` function

> Also not sure about the compile-time/-memory impact of moving _all_ of the tables into a single .cpp file.

It won't be moved in to a single file. Although one could argue that today all the OpcodeDispatcher tables are already in a single file (OpcodeDispatcher.cpp) so it won't be a significant difference regardless.  Just like how x86Tables are split between 10 files (https://github.com/FEX-Emu/FEX/tree/main/FEXCore/Source/Interface/Core/X86Tables) these will also be doing the same to match semantically. So in any case it'll be even MORE split up than it is today, thus improving the compile time.

Sonicadvance1:
Got a WIP which converts the rest of the tables over and compile time is within standard deviation of main, about 17.5 to 18.5 seconds compiling libFEXCore average. Showing that this seemingly doesn't affect compile time dramatically, as something else is likely to be blocking an significant difference.

As for binary size differences with all tables converted. FEXInterpreter sizes:
- Before: 4,475,544 bytes
- After: 4,457,096 bytes
- 18,448 bytes removed from the binary

And of course confirmed that the OpcodeDispatcher pointers are correctly populated in the binary without any further fixups, which will make initialization time slightly faster.

FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include "Interface/Core/OpcodeDispatcher.h"
+
+namespace FEXCore::IR {
+constexpr inline void InstallToTable(auto& FinalTable, auto& LocalTable) {
+  for (auto Op : LocalTable) {
```
neobrain, line 7:
```suggestion
  for (const auto& Op : LocalTable) {
```

FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
```diff
@@ -0,0 +1,134 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include "Interface/Core/OpcodeDispatcher.h"
+
+namespace FEXCore::IR {
+constexpr inline void InstallToTable(auto& FinalTable, auto& LocalTable) {
```
neobrain, line 6:
```suggestion
constexpr inline void InstallToTable(auto& FinalTable, const auto& LocalTable) {
```

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -533,7 +536,7 @@ constexpr static inline void GenerateTableWithCopy(X86InstInfo *FinalTable, X86T
     X86InstInfo const &Info = Op.Info;
     for (uint32_t i = 0; i < Op.second; ++i) {
       if (FinalTable[OpNum + i].Type != TYPE_UNKNOWN) {
-        ERROR_AND_DIE_FMT("Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
+        LOGMAN_MSG_A_FMT("Duplicate Entry {}->{}", FinalTable[OpNum + i].Name, Info.Name);
```
neobrain, line 539: Why change these lines to `LOGMAN_MSG_A_FMT`? IIRC that will prevent this constexpr-evaluated function from failing at compile-time in Release builds.

FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
```diff
@@ -0,0 +1,134 @@
...
+    {0xF5, 1, &OpDispatchBuilder::FLAGControlOp},
+    {0xF8, 2, &OpDispatchBuilder::FLAGControlOp},
+    {0xFA, 2, &OpDispatchBuilder::PermissionRestrictedOp},
+    {0xFC, 2, &OpDispatchBuilder::FLAGControlOp},
+  };
+
+  InstallToTable(FinalTable, BaseOpTable);
+}
+
+inline void BaseTables_Install64(auto& FinalTable) {
```
neobrain, line 98: These functions don't really do much. Why not drop them and instead just expose the `constexpr` tables globally? At call-site, we can use `InstallToTable(FinalTable, BaseOpTable_64)` directly then.

As a heads-up, when defined in headers, global constexpr variables must be `inline constexpr`, though I doubt it matters much here since they are only included in one header.

neobrain (commented):
Adding some minor comments, good to bring these up now before all the other tables are converted.

FEXCore/Source/Interface/Core/OpcodeDispatcher/BaseTables.h
Sonicadvance1: Exposed directly.

Sonicadvance1:
Another fun metric that I wasn't expecting. `InstallOpcodeHandlers` in main today uses 22kb of stack space to setup the tables.
Once all tables are converted over it drops to 6k of stack usage. Definitely still some improvements to figure out why it is using so much stack space, when theoretically it should just need to assign to the destination tables?

Sonicadvance1:
And I found what's causing it.
`constexpr std::tuple<uint16_t, uint8_t, FEXCore::X86Tables::OpDispatchPtr> X87OpTable`
versus
`constexpr static std::tuple<uint16_t, uint8_t, FEXCore::X86Tables::OpDispatchPtr> X87OpTable`

Once converted it drops to 64bytes of stack usage in my WIP branch.

neobrain (approved):
Cutting 22.000 bytes of stack use down to 64 sure is nice :)

Sonicadvance1:
I'll get the follow-up to convert the remaining tables over later today.

PR #4050, opened 2024-09-08
JIT: always use 64-bit moves for SRA

alyssarosenzweig:
I see no reason not to. Simpler code and it's slightly faster on Firestorm.


Sonicadvance1 (approved):
Sure

PR #4052, opened 2024-09-09
Bunch of little things reported by coverity

Sonicadvance1:
Mostly NFC, but it actually did find a few bugs, which is kind of unexpected.

Functional changes:
- (f51832ca6bae65c1ae692a0c0db81541106aa706) Fixes a bug in VPSRLDQ where larger than 16-byte shifts weren't zeroing the upper 16-bytes of a YMM register.
- (c559445b62a8f738fb096ed15da1c1ee5405e79c, 2d91c5441e4ef99d691fe8f1d022f241da777ef0, 9141322666eb7381b93ab3bf4605a288ad61900c, c4a0fb66092a2df82114bff9a001473777aabf99) IoctlEmulation: Fixes for a few structs that were copying the wrong direction for Host->Guest and missing a few member copies (Unlikely to affect anything as it was legacy interfaces)
- (98a91d242d77b555e78ce31263dbfa8b58295641) Fix resource leak on XXHash error
- (f6a8e595df2f82eb1dc454228beb2a1f632d4545) Add a TODO for futimesat because the syscall doesn't exist on ARM64
- (a308b9edbeb15c6b0d090e8820759ea1ac2ed7bf) Linux/Generate map: Fixes an FD leak
- (22ab16b2179d94b1b93c68dfecbdcce71c7f8d71) Fixes a "version" check in x32/ipc/msgrcv. The upper 16-bits of the call determine version and we weren't masking off the top bits correctly. Also ensures the shmat handler errors with a higher version to match Linux behaviour.
- (fa293e5e8f4a08efa2ce167f56c5b21438faa299) Fixes a buffer overrun in gdbserver when copying x87 registers

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
@@ -197,6 +197,7 @@ namespace DRM {
       CPYT(size);
       CPYT(low_mark);
       CPYT(high_mark);
+      memcpy(&val.flags, &flags, sizeof(val.flags));
       CPYT(agp_start);
       return val;
     }
@@ -206,6 +207,7 @@ namespace DRM {
       CPYF(size);
       CPYF(low_mark);
       CPYF(high_mark);
+      memcpy(&flags, &val.flags, sizeof(val.flags));
       CPYF(agp_start);
     }
   };
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
@@ -665,7 +665,7 @@ namespace RADEON {
   };
 
   struct FEX_PACKED FEX_ANNOTATE("alias-x86_32-drm_radeon_init") FEX_ANNOTATE("fex-match") fex_drm_radeon_init_t {
-    enum {} func;
+    uint32_t func;
 
     compat_ulong_t sarea_priv_offset;
     int32_t is_pci;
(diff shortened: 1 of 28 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
```diff
@@ -1115,10 +1115,10 @@ namespace I915 {
     fex_drm_i915_cmdbuffer_t(drm_i915_cmdbuffer_t val)
       : buf {auto_compat_ptr {val.buf}}
       , cliprects {auto_compat_ptr {val.cliprects}} {
-      CPYT(sz);
-      CPYT(DR1);
-      CPYT(DR4);
-      CPYT(num_cliprects);
+      CPYF(sz);
```
neobrain, line 1123: Personally I find the CPYT/CPYF macros actually make it harder to verify the code is doing the correct thing.

Looks like the true risk here is to have the `val` parameter not be const, so that assigning to its members would be a compile-error. The other case (using CPYF instead of CPYT) would be covered if we used brace-initialization instead of member-wise init.

Not sure it's worth going through the entire header to fix this though. Arguably the framework driving `thunkgen` could automate all of it.

neobrain:
"Mostly NFC" is not useful information for review. Could you move the functional changes to a dedicated PR, or at least clearly list the patches with functional change in the description?

Either way, great to see latent bugs being uncovered!

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/drm.h
Sonicadvance1: Should be solved in a follow-up PR. The idiom is already in place from before this PR, this just extends it to all the structs. With it covering all the structs we will more easily be able to sed replace them, knowing their correct.

Sonicadvance1:
> "Mostly NFC" is not useful information for review. Could you move the functional changes to a dedicated PR, or at least clearly list the patches with functional change in the description?
> 
> Either way, great to see latent bugs being uncovered!

Updated the description with the commits that have functionality change.

neobrain (approved):

PR #4055, opened 2024-09-10
FEXCore: Remove Context header include when unneeded

Sonicadvance1:
This is one of the most expensive headers in FEX, averaging 1.2 seconds of compile time per include.
We include this header 31 times around the project. Remove the five instances where it is unused to help this issue. Next step would be to make the header lighter if possible.


neobrain (approved):
Compile-time improvements are always appreciated :)

PR #4058, opened 2024-09-10
ARM64EC: Patch the call checker before calling any syscall exports

bylaws:
CEF on Windows patches these before FEX is even loaded, so this needs to be setup as early as possible using some direct syscalls which skip any such patches.


bylaws:
Currently CEF is still broken on wine with this - I need to implement something similar on the wine side

Sonicadvance1 (approved):

PR #4060, opened 2024-09-10
Windows: Switch supported MinGW CRT to UCRT

bylaws:
MSVCRT is deprecated and the reasons FEX used it no longer apply, so support UCRT and UCRT only.

Draft as not tested particularly well and this needs new CI binaries 
cc: @AndreRH

### DIFF
--- a/Source/Windows/ARM64EC/CMakeLists.txt
+++ b/Source/Windows/ARM64EC/CMakeLists.txt
@@ -3,7 +3,6 @@ include(GNUInstallDirs)
 add_library(arm64ecfex SHARED
   Module.cpp
   Module.S
-  ARM64ECSymbols.S
   libarm64ecfex.def
   $<TARGET_OBJECTS:FEXCore_object>
 )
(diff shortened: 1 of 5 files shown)


AndreRH:
I applied it on top of 2408, needed to remove a const and had issues linking cause of control flow guards

Sonicadvance1 (approved):
CI updated with new toolchain, good to go.

AndreRH:
Is somewhere documented what toolchains are on the CI runners? As I understand you use your own runners instead of GitHub provided ones.

Sonicadvance1:
> Is somewhere documented what toolchains are on the CI runners? As I understand you use your own runners instead of GitHub provided ones.

https://wiki.fex-emu.com/index.php/Development:ARM64EC has a bit of documentation

Currently using the toolchain from Bylaw's fork https://github.com/bylaws/llvm-mingw/releases/tag/20240929

Personally no idea what the difference is from upstream llvm-mingw

PR #4063, opened 2024-09-12
FEXRootFSFetcher: Fallback to TTY if zenity isn't installed

Sonicadvance1:
If packaged improperly or if someone has just did a source install then they may not have zenity. Fallback to TTY path if zenity isn't installed.

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -220,7 +220,7 @@ static bool Has_EroFSFsck {false};
 
 void CheckCurl() {
   // Check if curl exists on the host
-  std::vector<const char*> ExecveArgs = {
+  const std::array<const char*, 3> ExecveArgs = {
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -284,6 +285,17 @@ void CheckUnsquashfs() {
   }
   close(fd);
 }
+void CheckZenity() {
+  // Check if zenity exists on the host
+  std::vector<const char*> ExecveArgs = {
```
lioncash, line 290: Perhaps use a std::array here, which would get rid of a trivial allocation

Sonicadvance1: Converted this over to array, then converted the other copy and paste sections of code using vector, which also converts a couple of vector passing arguments to span.

lioncash (approved):
nice nice nice

PR #4066, opened 2024-09-13
Tools: Delete imgui FEXConfig

Sonicadvance1:


neobrain (approved):
Nice!

It's worth mentioning the commit structure here will make `git log FEXConfig/Main.cpp` show the old ImGui history instead of FEXQonfig's. That is probably the better choice since FEXQonfig's git history is fairly uninteresting.

However, if we wanted to preserve the FEXQonfig history, we could split the commit in two: One that removes FEXConfig, and one that renames FEXQonfig to FEXConfig.

Sonicadvance1:
Split the commit to retain history.

PR #4067, opened 2024-09-14
FEXQonfig: Allow system-provided RootFS images

WhatAmISupposedToPutHere:
Add a secondary RootFS directory that can be managed by the system package manager.

### DIFF
--- a/Source/Tools/FEXQonfig/Main.cpp
+++ b/Source/Tools/FEXQonfig/Main.cpp
@@ -189,9 +189,12 @@ static void ConfigInit(fextl::string ConfigFilename) {
 RootFSModel::RootFSModel() {
   INotifyFD = inotify_init1(IN_NONBLOCK | IN_CLOEXEC);
 
-  fextl::string RootFS = FEXCore::Config::GetDataDirectory() + "RootFS/";
-  FolderFD = inotify_add_watch(INotifyFD, RootFS.c_str(), IN_CREATE | IN_DELETE);
-  if (FolderFD != -1) {
+  fextl::string RootFSUser = FEXCore::Config::GetDataDirectory() + "RootFS/";
+  fextl::string RootFSSystem = "/usr/share/fex-emu/RootFS/";
+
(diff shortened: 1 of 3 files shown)


alyssarosenzweig:
Could you explain a bit on the background for this?

For Asahi specifically, the system rootfs will end up living in /run/share/fex-emu, I think, due to krun magic.

WhatAmISupposedToPutHere:
For gentoo, i do not want to do krun magic inside the vm, and instead create the squashfs and overlay mounts on the host. Since mounting it inside the per-user rootfs directory would be quite annoying to implement, I instead want to add a secondary "default" rootfs search directory.

WhatAmISupposedToPutHere:
And while the current gentoo work is asahi-specific, i want to also run it on my rk3588 boards, which will not use krun.

alyssarosenzweig:
OK, makes sense. Not too different from the fedora asahi use case then..

Sonicadvance1:
I had a couple of concerns that I didn't have time to articulate before today. So I spent a bit of time to implement this closer to what FEX would want.

- Hardcoded global paths, we would want FEX to follow what was provided with `$CMAKE_INSTALL_PREFIX`
  - Likely being `/usr` or `/usr/local` typically, depending on what the user wants.
- We want FEXCore to track the global data path so it can tie in to our configuration system, even if only used for this rootfs implementation today.
- There was some extraneous prefixing thing in this PR which wasn't very necessary, we can just show the full system path instead.
  - This was likely a confusion for how FEX's "named" rootfs behaviour works, which isn't necessary for this global path, just supply the full path.  
- I wanted to reimplement the inotify interface to use ppoll instead of select.
  - This was just my bad implementation from the start 
- The watch descriptors didn't need to be saved in the global object so I wanted to change that (We don't support removing a WD from an inotify on full path deletion)

Thank you for bringing up that you wanted this as a feature in FEXConfig!

Sonicadvance1:
Handled by #4074. If there's any more features you want then make sure to let us know.

PR #4068, opened 2024-09-14
CPUID: Update to something a little more modern

Sonicadvance1:
Previously reported as some old CPU without AVX and SSE4 and other things.
Start advertising as something more modern that actually shipped with AVX2 and other features. Should help some modern libraries that do bad family and model checks rather that CPUID features checks.

Also removes the silly `(ES)` tag from CPU-Z.

Also moves generation in to a constexpr function that can actually range check these 4-bit and 8-bit values.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -96,20 +96,39 @@ static uint32_t GetCPUID() {
   return CPU;
 }
 
+struct CPUFamily {
+  uint32_t Stepping         : 4;
+  uint32_t Model            : 4;
+  uint32_t ExtendedModel    : 4;
+  uint32_t FamilyID         : 4;
+  uint32_t ExtendedFamilyID : 8;
+  uint32_t ProcessorType    : 4;
+};
+
+constexpr static uint32_t GenerateFamily(const CPUFamily Family) {
+  return Family.Stepping | (Family.Model << 4) | (Family.FamilyID << 8) | (Family.ProcessorType << 12) | (Family.ExtendedModel << 16) |
+         (Family.ExtendedFamilyID << 20);
+}
+
 #ifdef CPUID_AMD
-constexpr uint32_t FAMILY_IDENTIFIER = 0 |          // Stepping
-                                       (0xA << 4) | // Model
-                                       (0xF << 8) | // Family ID
-                                       (0 << 12) |  // Processor type
-                                       (0 << 16) |  // Extended model ID
-                                       (1 << 20);   // Extended family ID
+constexpr uint32_t FAMILY_IDENTIFIER = GenerateFamily(CPUFamily {
+  .Stepping = 0,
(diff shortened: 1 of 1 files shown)


lioncash (approved):

FEXCore/include/FEXCore/Utils/LogManager.h
```diff
@@ -171,5 +171,12 @@ namespace Msg {
     FEX_TRAP_EXECUTION;             \
   } while (0)
 
+#define ERROR_AND_DIE_THROW_FMT(pred, ...) \
```
neobrain, line 174: Why does it need this need macro?

We already have like 4 other ways of asserting/dieing, and none of them have distinguishable names or are documented.

Sonicadvance1: Similar reason to the comment from #4048: `Why change these lines to LOGMAN_MSG_A_FMT? IIRC that will prevent this constexpr-evaluated function from failing at compile-time in Release builds.`

I can't use `LOGMAN_MSG_A_FMT` because this would prevent constexpr-evaluated function from failing at compile-time in Release builds.
So instead of wrapping all six lines in `GenerateFamily` with if statements, I added a throwable version of `ERROR_AND_DIE_FMT` that can be predicated.

Meanwhile we can't use a basic `static_assert` to handle the constexpr-evaluated expression because clang complains about `Static assertion expression is not an integral constant expression` due to wrapping the arguments in the CPUFamily struct to have named initializers. Which makes it easier to reason about what is getting passed to which argument.

neobrain: We can get something more equivalent to `static_assert` here by using consteval:

```cpp
// Assertion helper for code evaluated at compile-time
inline consteval void consteval_assert(bool assertion) {
    if (!assertion) {
        // Trigger compile error
        throw "consteval_assert fail";
    }
}
```

We could also add proper message support by wrapping this in a macro, but at least for the assertions in this PR we're not losing that much information as-is.

FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -96,20 +96,46 @@ static uint32_t GetCPUID() {
...
+  uint32_t Stepping;
+  uint32_t Model;
+  uint32_t ExtendedModel;
+  uint32_t FamilyID;
+  uint32_t ExtendedFamilyID;
+  uint32_t ProcessorType;
+};
+
+constexpr static uint32_t GenerateFamily(const CPUFamily Family) {
+  ERROR_AND_DIE_THROW_FMT((Family.Stepping & ~0xF) == 0, "Stepping too large");
```
neobrain, line 109: Why not just test for `Family.Stepping < 0x10`?

Sonicadvance1:
Switched over to struct bitfields instead to enforce the size limits.

neobrain (approved):

PR #4069, opened 2024-09-14
CMake: Adds option for clang stack size tracing

Sonicadvance1:
This lets clang generate a symbol section to show how much stack a function uses for later analysis. Very useful for sussing out bad code smells.

To read the resulting section one can just use `llvm-readelf --stack-sizes <ELF>` Then a little bit of manual sed fixup to stick in to a spreadsheet and sort by size.

Let me find an issue where FEXCore is generating its x87 tables on the stack then copying them to the result, which is wacky.


neobrain:
This seems rather specific. Is this indeed better served as a standalone option? How often are people realistically going to use this when they could manually run `-DCMAKE_CXX_FLAGS=-fstack-size-section`?

Conversely adding more options to CMakeLists.txt makes discovery of existing settings more difficult, particularly when it gets to the point that half of them are never used in practice.

Sonicadvance1:
Conversely, if it doesn't exist as an option, then I'll never use it again since I'll forget about it. I look at the options in ccmake pretty much daily.

neobrain:
Not sure the top-level CMakeLists.txt is the appropriate place for personal notes, but I'll let others weigh in on this.

Sonicadvance1:
We should also scrutinize all the other options as well that could be just as easily enabled using cmake flags
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L14 xray
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L16 asan
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L17 tsan
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L18 coverage
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L21-L22 werror and strict werror
- https://github.com/FEX-Emu/FEX/blob/main/CMakeLists.txt#L36 pdb debuginfo

neobrain:
> We should also scrutinize all the other options as well that could be just as easily enabled using cmake flags

There's merit to most of these. Most are more commonly useful (ASAN/TSAN/COVERAGE) and require multiple flags in combination. PDB debuginfo corresponds to a cryptic set of settings that would be difficult to search for.

Only a single one of them corresponds directly to a single flag (`-fxray-instrument). Indeed this is a similarly niche option: It's only useful to those who already know what they're looking for, at which point you can just manually specify the flag.

`ENABLE_WERROR` and `ENABLE_STRICT_WERROR` are maybe edge-cases. They could be justified if we were more selective about the flags, but as-is having two options here is really overkill.

Sonicadvance1:
Updating this with a script that does the sed parsing to generate a semi-colon delimited file that can be slapped in to a spreadsheet.

```sh
#!/bin/sh
llvm-readelf-19 -C --stack-sizes $1 > $2
sed -i 1,3d $2
sed -i 's/^[ \t]*//' $2
sed -i 's/ \{1,\}/;/' $2
sed -i '1s/^/Size;Function\n/' $2
```

`../Scripts/stack_sizes.sh ./Bin/FEXInterpreter out.txt`

PR #4070, opened 2024-09-15
Common: Remove dependency on cpp-optparse (Only for FEXLoader)

Sonicadvance1:
Doesn't remove it from the full project since three tools still use it.

FEXLoader argument loader is fairly special since we have a generator tied to it. This generator lets us have some niceties where everything ends up being a string literal without any sort of dynamic requirements.

cpp-optparse has a systemic issue where it makes deep copies of /everything/ all the time. This means it has huge runtime costs and huge stack usage (4992 bytes). The new implementation uses 608 bytes of stack space (plus 640 if it actually parses a strenum).

When profiling cpp-optparse with FEXLoader's argument it takes ~2,039,307 ns to parse the arguments! As a direct comparison this new implementation takes ~56,885 ns. Both sides not getting passed /any/ arguments, really shows how spicy this library is.

### DIFF
--- a/Source/Common/ArgumentLoader.cpp
+++ b/Source/Common/ArgumentLoader.cpp
@@ -150,16 +150,34 @@ std::pair<fextl::vector<fextl::string>, fextl::vector<const char*>> ArgParser::P
     if (IsShort) {
       ArgFirst = Arg;
 
-      OptionDetails = FindOption(ArgFirst);
+      for (size_t i = 1; i < ArgFirst.size(); ++i) {
+        const auto ArgChar = ArgFirst.substr(i, 1);
+        OptionDetails = FindOption(ArgChar);
 
-      if (OptionDetails == nullptr) [[unlikely]] {
-        ExitWithError(fextl::fmt::format("Unsupported argument: {}\nUse --help for more information.", Arg));
+        if (OptionDetails == nullptr) [[unlikely]] {
+          ExitWithError(fextl::fmt::format("Unsupported argument: {}\nUse --help for more information.", Arg));
+        }
+
+        if (NeedsArg(OptionDetails)) {
+          if (i > 1) {
+            ExitWithError(fextl::fmt::format("Can't combine short arguments that require argument: {}\n", Arg));
+          }
+
+          // Save the FEX argument.
+          ProgramArguments.emplace_back(argv[ArgParsed]);
+
+          ++ArgParsed;
+          ArgSecond = argv[ArgParsed];
+        }
+
+        // Save the FEX argument.
+        ProgramArguments.emplace_back(argv[ArgParsed]);
+
+        // Now parse the argument.
+        ParseArgument(ArgChar, ArgSecond, OptionDetails);
       }
 
-      if (NeedsArg(OptionDetails)) {
-        ++ArgParsed;
-        ArgSecond = argv[ArgParsed];
-      }
+      continue;
     } else if (IsLong) {
       const auto Split = Arg.find_first_of('=');
       bool NeedsSplitArg {};
--- a/Source/Common/ArgumentLoader.cpp
+++ b/Source/Common/ArgumentLoader.cpp
@@ -5,47 +5,242 @@
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
 
-#include "cpp-optparse/OptionParser.h"
 #include "git_version.h"
 
-#include <stdint.h>
-
 namespace FEX::ArgLoader {
+enum class ValueType : uint32_t {
+  String,
+  BoolTrue,
+  BoolFalse,
+  StringEnum,
+  StringArray,
+  Max,
+};
+
+struct OptionDetails {
+  std::string_view Arg;
+  FEXCore::Config::ConfigOption Option;
+  ValueType ParseType;
+};
+
+constexpr static OptionDetails ArgToOption[] = {
+#define ARG_TO_CONFIG
+#include <FEXCore/Config/ConfigOptions.inl>
+};
+
+const OptionDetails* FindOption(std::string_view Argument) {
+  for (auto& Arg : ArgToOption) {
+    if (Arg.Arg == Argument) {
+      return &Arg;
+    }
+  }
+
+  return nullptr;
+}
+
+void PrintHelp() {
+  const char* Arguments[3] {};
+  Arguments[0] = "man";
+  Arguments[1] = "FEX";
+  Arguments[2] = nullptr;
+
+  execvp("man", (char* const*)Arguments);
+}
+
+void ExitWithError(std::string_view Error) {
+  fextl::fmt::print("Error: {}\n", Error);
+  _exit(1);
+}
+
+class ArgParser final {
+public:
+  ArgParser(FEX::ArgLoader::ArgLoader* Loader)
+    : Loader {Loader} {}
+
+  std::pair<fextl::vector<fextl::string>, fextl::vector<const char*>> Parse(int argc, char** argv);
+  void Version(std::string_view version) {
+    _Version = version;
+  }
+
+private:
+  FEX::ArgLoader::ArgLoader* Loader;
+  std::string_view _Version {};
+  using ParseArgHandler = void (ArgParser::*)(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details);
+
+  void ParseArgument_String(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    switch (Details->Option) {
+#define STR_CONVERT_PARSE
+#include <FEXCore/Config/ConfigOptions.inl>
+    [[likely]] default:
+      Loader->SetArg(Details->Option, SecondArg);
+      break;
+    }
+  }
+
+  void ParseArgument_BoolTrue(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    Loader->SetArg(Details->Option, "1");
+  }
+
+  void ParseArgument_BoolFalse(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    Loader->SetArg(Details->Option, "0");
+  }
+
+  void ParseArgument_StringEnum(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    switch (Details->Option) {
+#define STR_ENUM_PARSE
+#include <FEXCore/Config/ConfigOptions.inl>
+    [[unlikely]] default:
+      ExitWithError(fextl::fmt::format("Unknown strenum argument: {}", Arg));
+      break;
+    }
+  }
+
+  constexpr static std::array<ParseArgHandler, FEXCore::ToUnderlying(ValueType::Max)> Parser = {
+    &ArgParser::ParseArgument_String,
+    &ArgParser::ParseArgument_BoolTrue,
+    &ArgParser::ParseArgument_BoolFalse,
+    &ArgParser::ParseArgument_StringEnum,
+    // Behaves the same as string but appends multiple.
+    &ArgParser::ParseArgument_String,
+  };
+
+  void ParseArgument(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    (this->*Parser[FEXCore::ToUnderlying(Details->ParseType)])(Arg, SecondArg, Details);
+  }
+};
+
+static bool NeedsArg(const OptionDetails* Details) {
+  return Details->ParseType == ValueType::String || Details->ParseType == ValueType::StringEnum || Details->ParseType == ValueType::StringArray;
+}
+
+std::pair<fextl::vector<fextl::string>, fextl::vector<const char*>> ArgParser::Parse(int argc, char** argv) {
+  fextl::vector<fextl::string> RemainingArgs {};
+  fextl::vector<const char*> ProgramArguments {};
+
+  // Skip argv[0]
+  int ArgParsed = 1;
+  for (; ArgParsed < argc; ++ArgParsed) {
+    std::string_view Arg = argv[ArgParsed];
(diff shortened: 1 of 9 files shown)


lioncash (approved):

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
+  void Version(std::string_view version) {
+    _Version = version;
+  }
+
+  fextl::vector<fextl::string> GetGuestArgs() {
+    return std::move(RemainingArgs);
+  }
+  fextl::vector<const char*> GetParsedArgs() {
+    return std::move(ProgramArguments);
+  }
```
neobrain, line 70: If we need a way to pull all state underneath this class, the way to do it is to enforce move semantics:
```cpp
auto [Guest, Parsed] = std::move(arg_parser).GetArgs();
```

This is achieved by merging the two functions and making it callable with rvalues, only:
```cpp
std::pair<vector, vector> GetArgs() && {
    return make_pair(std::move(Args1), std::move(Args2));
}
```

Equivalently, the parsed arguments don't need to be state in ArgParser but could instead be returned from Parse() directly.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
 
-  Parser.add_option_group(CPUGroup);
-  Parser.add_option_group(EmulationGroup);
-  Parser.add_option_group(DebugGroup);
-  Parser.add_option_group(HacksGroup);
-  Parser.add_option_group(MiscGroup);
-  Parser.add_option_group(LoggingGroup);
+  void ParseArgument_BoolTrue(std::string_view Arg, std::string_view SecondArg, const OptionDetails* Details) {
+    using namespace std::literals;
+    Loader->SetArg(Details->Option, "1"sv);
```
neobrain, line 91: This is easier to read without using string_view literals, and for a single-char string the compiler produces the same code either way:
```suggestion
    Loader->SetArg(Details->Option, "1");
```

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
+    if (ProgramArguments.empty() && OptionDetails == nullptr) {
+      // In the special case that we hit a parse error and we haven't parsed any arguments, pass everything.
+      // This handles the typical case eg: `FEXLoader /usr/bin/ls /`.
+      // Some would claim that `--` should be used to split FEX arguments from sub-application arguments.
+      break;
+    }
+
+    // Unsupported FEX argument. Error.
+    if (ProgramArguments.empty() && OptionDetails == nullptr) [[unlikely]] {
+      ExitWithError(fextl::fmt::format("Unsupported argument: {}", Arg));
```
neobrain, line 220:
```suggestion
      ExitWithError(fextl::fmt::format("Unsupported argument: {}\nUse --help for more information.", Arg));
```

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
+  return nullptr;
+}
+
+void PrintHelp() {
+  const char* Arguments[3] {};
+  Arguments[0] = "man";
+  Arguments[1] = "FEX";
+  Arguments[2] = nullptr;
+
+  execvp("man", (char* const*)Arguments);
```
neobrain, line 47: Why does it not print a built-in help for FEXLoader like it does for other tools in 4072?

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
-#include <stdint.h>
-
 namespace FEX::ArgLoader {
-void FEX::ArgLoader::ArgLoader::Load() {
-  RemainingArgs.clear();
-  ProgramArguments.clear();
-  if (Type == LoadType::WITHOUT_FEXLOADER_PARSER) {
-    LoadWithoutArguments();
-    return;
+enum class ParserTypes : uint32_t {
```
neobrain, line 11: Since this is already in the ArgLoader namespace, why not call this "ValueType"?

neobrain (commented):
Adding some comments here and some others in 4072.

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Decided to redirect to the man page because the options are complex enough with enough formatting issues that it would be a pain. It's becoming a common thing in Linux space that passing `-h` or `--help` in to Linux applications just load the man page anyway.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
+        ArgSecond = Arg.substr(Split + 1, Arg.size());
+
+        if (ArgSecond.empty()) [[unlikely]] {
+          ExitWithError(fextl::fmt::format("{} needs argument", Arg));
+        }
+      }
+    }
+
+    if (ProgramArguments.empty() && OptionDetails == nullptr) {
+      // In the special case that we hit a parse error and we haven't parsed any arguments, pass everything.
```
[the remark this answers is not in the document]
Sonicadvance1: This already happens.
```bash
$ ./Bin/FEXLoader --non-existing-option `which ls`
Error: Unsupported argument: --non-existing-option
```

This edge case specifically only happens if the argument getting parsed was not prefixed with `-` or `--`, so passing in a command or something. Like the comment said, something like `FEXLoader /usr/bin/ls` would hit this case.

Other cases with invalid argument passing is above in the `IsLong` or `IsShort` checks.

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Resolved.

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Resolved.

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Resolved.

Source/Common/ArgumentLoader.cpp
Sonicadvance1: Resolved.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,250 @@
...
-  optparse::OptionGroup DebugGroup(Parser, "Debug options");
-  optparse::OptionGroup HacksGroup(Parser, "Hacks options");
-  optparse::OptionGroup MiscGroup(Parser, "Miscellaneous options");
-  optparse::OptionGroup LoggingGroup(Parser, "Logging options");
+  return nullptr;
+}
+
+void PrintHelp() {
+  const char* Arguments[3] {};
+  Arguments[0] = "man";
```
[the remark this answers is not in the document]
Sonicadvance1: Resolved.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+        // Now parse the argument.
+        ParseArgument(ArgChar, ArgSecond, OptionDetails);
+      }
+
+      continue;
+    } else if (IsLong) {
+      const auto Split = Arg.find_first_of('=');
+      bool NeedsSplitArg {};
+      if (Split == Arg.npos) {
+        ArgFirst = Arg;
```
neobrain, line 185: This line has no effect.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+    if (Arg == "--") {
+      // Special case break. Remaining arguments get passed to guest.
+      ++ArgParsed;
+      break;
+    }
+
+    const bool IsShort = Arg.find("--", 0, 2) == Arg.npos && Arg.find("-", 0, 1) == 0;
+    const bool IsLong = Arg.find("--", 0, 2) == 0;
+
+    std::string_view ArgFirst {};
```
neobrain, line 145: This variable doesn't need to be declared before proper initialization, since the two branches don't converge in a path that still uses the variable.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+    std::string_view ArgSecond {};
+
+    const OptionDetails* OptionDetails {};
+
+    if (IsShort) {
+      ArgFirst = Arg;
+
+      for (size_t i = 1; i < ArgFirst.size(); ++i) {
+        const auto ArgChar = ArgFirst.substr(i, 1);
```
neobrain, line 154: Is there a reason not to use `Arg` directly here?

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+    (this->*Parser[FEXCore::ToUnderlying(Details->ParseType)])(Arg, SecondArg, Details);
+  }
+};
+
+static bool NeedsArg(const OptionDetails* Details) {
+  return Details->ParseType == ValueType::String || Details->ParseType == ValueType::StringEnum || Details->ParseType == ValueType::StringArray;
+}
+
+std::pair<fextl::vector<fextl::string>, fextl::vector<const char*>> ArgParser::Parse(int argc, char** argv) {
+  fextl::vector<fextl::string> RemainingArgs {};
```
neobrain, line 117: Declaration of this can be moved all the way down.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+
+  std::pair<fextl::vector<fextl::string>, fextl::vector<const char*>> Parse(int argc, char** argv);
+  void Version(std::string_view version) {
+    _Version = version;
+  }
 
-#define BEFORE_PARSE
+private:
+  FEX::ArgLoader::ArgLoader* Loader;
+  std::string_view _Version {};
```
neobrain, line 67: Starting variable names with an underscore is weird style, and this particular name seems common enough that we might actually hit a collision from reserved names at some point.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+
+  execvp("man", (char* const*)Arguments);
+}
+
+void ExitWithError(std::string_view Error) {
+  fextl::fmt::print("Error: {}\n", Error);
+  _exit(1);
+}
+
+class ArgParser final {
```
neobrain, line 55: This needs some documentation to explain why it exists and what's the difference from the confusingly similarly named ArgLoader.

Source/Common/ArgumentLoader.cpp
```diff
@@ -5,47 +5,260 @@
...
+void PrintHelp() {
+  const char* Arguments[3] {};
+  Arguments[0] = "man";
+  Arguments[1] = "FEX";
+  Arguments[2] = nullptr;
+
+  execvp("man", (char* const*)Arguments);
+}
+
+void ExitWithError(std::string_view Error) {
```
neobrain, line 50: Global functions used only locally should be `static`.

unittests/ThunkFunctionalTests/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ function(AddThunksTest Bin ThunksFile)
   set (ARGS
     "-t" "${CMAKE_INSTALL_PREFIX}/lib/fex-emu/HostThunks"
     "-j" "${CMAKE_INSTALL_PREFIX}/share/fex-emu/GuestThunks"
-    "-o" "stderr" "--no-silentlog" "-n" "500"
+    "-o" "stderr" "--no-silentlog" "-n" "500" "--"
```
neobrain, line 7: "--" seems to be listed in one of the uses below already.

unittests/ASM/CMakeLists.txt
```diff
@@ -90,6 +90,8 @@ foreach(ASM_SRC ${ASM_SOURCES})
       list(APPEND ARGS_LIST "--smcchecks=full")
     endif()
 
+    list(APPEND ARGS_LIST "--")
```
neobrain, line 93: Why is this needed now when it wasn't before?

Source/Common/ArgumentLoader.cpp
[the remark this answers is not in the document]
neobrain: I stumbled upon something during testing but can't remember what exactly, might have been a silly mistake on my end though. It's true that this specific example works fine after all.

unittests/ASM/CMakeLists.txt
Sonicadvance1: We had a bug in our arguments getting passed to unittests for years (#4071) and it was undiscovered. Due to a quirk with cpp-optparse where any unsupported arguments were gathered and passed to the guest application. This didn't necessarily break anything in our unittests, just made them run silently, and for all the harness things the additional argument was ignored anyway.

The commit these changes are from should probably be broken out to a freestanding commit before this PR is merged anyway.

Sonicadvance1: and since I realize I didn't actually answer the question. We're being more strict than cpp-optparse splitting out FEXLoader arguments from guest arguments because of that mentioned problem. So the unittests need the split split pattern now, where cpp-optparse didn't need this.

neobrain: > We're being more strict than cpp-optparse splitting out FEXLoader arguments from guest arguments because of that mentioned problem. So the unittests need the split pattern now, where cpp-optparse didn't need this.

Sounds reasonable 👍

Sonicadvance1:
I don't have the energy to push this over the finish line. With a bit of effort the FEXLoader implementation and "everything else" implementation could be merged in to a single class with some inheritance or something.

The modified cpp-optparse still works and the 35x speed-up while nice, is only like a 5% removal of execution time in short running applications (If it even happens, FEXInterpreter early exits this impl).

PR #4072, opened 2024-09-16
Remove cpp-optparse from the rest of our project

Sonicadvance1:
Based on #4070

First step: add an argument parser that handles /only/ what FEX needs for its tools.
Second step: Move everything over to that.
Third step: Delete cpp-optparse.

FEXLoader using the generated loader still.

### DIFF
--- a/Source/Common/MicroArgParser.cpp
+++ b/Source/Common/MicroArgParser.cpp
@@ -1,0 +1,303 @@
+#include "Common/MicroArgParser.h"
+
+#include <FEXCore/fextl/fmt.h>
+#include <algorithm>
+#include <charconv>
+#include <ranges>
+
+namespace FEX {
+[[noreturn]]
+inline void ExitWithError(std::string_view Error) {
+  fextl::fmt::print(stderr, "Error: {}\n", Error);
+  std::exit(1);
+}
+
+template<typename T>
+T MicroArgParser::Value::Get(std::string_view Default) const {
+  auto ParseValue = HasValue ? _Value.front() : Default;
+
+  if constexpr (std::is_same_v<T, bool>) {
+    using namespace std::literals;
+    return ParseValue == "0"sv ? false : true;
+  } else if constexpr (std::is_same_v<T, std::string_view>) {
+    return ParseValue;
+  } else if constexpr (std::is_integral_v<T>) {
+    T Result {};
+    auto Results = std::from_chars(ParseValue.begin(), ParseValue.end(), Result, 10);
+    if (Results.ec == std::errc()) {
+      return Result;
+    }
+    ExitWithError(fextl::fmt::format("Failed to convert '{}' to integer\n", ParseValue));
+  } else if constexpr (std::is_same_v<T, std::string_view>) {
+    return ParseValue;
+  } else {
+    ExitWithError("Unknown type");
+  }
+}
+
+template bool MicroArgParser::Value::Get<bool>(std::string_view Default) const;
+template uint32_t MicroArgParser::Value::Get<uint32_t>(std::string_view Default) const;
+template std::string_view MicroArgParser::Value::Get<std::string_view>(std::string_view Default) const;
+
+void MicroArgParser::PrintVersion() {
+  fextl::fmt::print("{}\n", _Version);
+  std::exit(0);
+}
+
+void MicroArgParser::PrintHelp(char** argv, std::span<ParseMember> ParseList) {
+  fextl::fmt::print("Usage: {} [options]\n", argv[0]);
+
+  if (!_Version.empty()) {
+    fextl::fmt::print("{}\n", _Version);
+  }
+
+  if (!_Desc.empty()) {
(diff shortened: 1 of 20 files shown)


Source/Tools/FEXGetConfig/Main.cpp
```diff
@@ -114,24 +114,26 @@ int main(int argc, char** argv, char** envp) {
...
-  Parser.add_option("--current-rootfs").action("store_true").help("Print the directory that contains the FEX rootfs. Mounted in the case of squashfs");
-
-  Parser.add_option("--tso-emulation-info").action("store_true").help("Print how FEX is emulating the x86-TSO memory model.");
+  static std::array<FEX::MicroArgParser::ParseMember, 6> Args = {{
+    {FEX::MicroArgParser::ParseMember::Type::Bool, {}, "--install-prefix", "Print the FEX install prefix", "0"},
+    {FEX::MicroArgParser::ParseMember::Type::String, {}, "--app", "Load an application profile for this application if it exists", {}},
+    {FEX::MicroArgParser::ParseMember::Type::Bool, {}, "--current-rootfs", "Print the directory that contains the FEX rootfs. Mounted in the case of squashfs", "0"},
+    {FEX::MicroArgParser::ParseMember::Type::Bool, {}, "--tso-emulation-info", "Print how FEX is emulating the x86-TSO memory model.", "0"},
+    // Defaults
+    {FEX::MicroArgParser::ParseMember::Type::Bool, "-v", "--version", "show program's verison number and exit", "0"},
```
neobrain, line 123: verison is spelt wrong here and in a couple of other places.

Source/Tools/pidof/pidof.cpp
```diff
@@ -14,35 +15,34 @@ namespace Config {
...
-    std::istringstream ss {Omit};
+  static std::array<FEX::MicroArgParser::ParseMember, 7> Args = {{
+    {FEX::MicroArgParser::ParseMember::Type::Bool, "-s", {}, "Single shot - Only returns one pid", "0"},
+    {FEX::MicroArgParser::ParseMember::Type::Bool, "-q", {}, "Do not display matched PIDs to stdout. Simply exit with status of true or false if a PID was found", "0"},
+    {FEX::MicroArgParser::ParseMember::Type::BoolInvert, "-z", {}, "Try to detect zombie processes - Usually zombie processes are skipped", "1"},
+    {FEX::MicroArgParser::ParseMember::Type::String, "-d", {}, "Use a different separator if more than one pid is show - Default is space", " "},
+    {FEX::MicroArgParser::ParseMember::Type::StringArray, "-o", {}, "Ignore processes with matched pids", {}},
+
+    // Defaults
+    {FEX::MicroArgParser::ParseMember::Type::Bool, "-v", "--version", "show program's verison number and exit", "0"},
```
neobrain, line 31: `show` should be spelt `Show` for consistency with other descriptions.

Source/Common/MicroArgParser.cpp
```diff
@@ -0,0 +1,303 @@
...
+
+    if (Begin != ArgEnd) {
+      const auto View = std::string_view(Begin, ArgEnd - Begin);
+      if (View == Arg) {
+        return;
+      }
+    }
+  }
+
+  ExitWithError(fextl::fmt::format("'{}' Not in list of choices '{}'\n", Arg, Member->Choices));
```
neobrain, line 188:
```suggestion
  ExitWithError(fextl::fmt::format("Invalid argument '{}'. Valid choices are '{}'\n", Arg, Member->Choices));
```
or better yet:
```suggestion
  ExitWithError(fextl::fmt::format("Invalid argument '{}' for --TODO. Valid choices are '{}'\n", Arg, Member->Choices));
```

Source/Common/MicroArgParser.cpp
```diff
@@ -0,0 +1,303 @@
...
+}
+
+void MicroArgParser::Parse(int argc, char** argv) {
```
neobrain, line 191: Invoking multiple short options at once seems broken now:

```
./Bin/FEXpidof -sq
Error: Unknown argument: -sq
```

Source/Common/MicroArgParser.h
```diff
@@ -0,0 +1,103 @@
...
+    template<typename T>
+    T Get(std::string_view Default) const;
+
+
+    const fextl::vector<std::string_view>& GetVector() const {
+      return _Value;
+    }
+
+  private:
+    fextl::vector<std::string_view> _Value;
```
neobrain, line 35: Identifiers starting with an underscore look out of place here.

Also, this can be an `std::optional<vector<string_view>>` instead to avoid having two variables to manage the same state.

Source/Tools/pidof/pidof.cpp
neobrain, line 31: The apostrophe s isn't needed here (and actually sounds slightly wrong to me due to lack of preceding article):
program's version -> program version

neobrain (commented):
Adding more comments here, most of which I think already apply to 4070.

PR #4074, opened 2024-09-18
FEXConfig: Add the ability to watch and configure global rootfs

Sonicadvance1:
Global rootfs doesn't support a named rootfs (only local supports that).
So when it is a global rootfs option just give the full path to the
config path.

Just needs some special handling to ensure the full path is retained for
global paths.

A reimplementation of https://github.com/FEX-Emu/FEX/pull/4067 that doesn't hardcode paths and routes the
global data path through FEXCore.


Sonicadvance1:
Rebased and merge conflict fixed.

PR #4075, opened 2024-09-20
Thunks: Move to the frontend

Sonicadvance1:
FEX's thunks are fundamentally a Linux specific feature, but also with some interaction with the FEXCore in order to handle callback trampolines and passing the thunk information to the backend.

This takes the time to decouple these two halves and move it to the frontend FEXLoader. A little bit of restructuring necessary to be done in order to do this, but nothing too scary.

Walking the commits from earliest to latest will show some of the work necessary for decoupling.


Sonicadvance1:
Rebased and merge conflict resolved.

Source/Tools/LinuxEmulation/Thunks.h
```diff
@@ -0,0 +1,22 @@
...
+#include <FEXCore/Core/Thunks.h>
+#include <FEXCore/fextl/memory.h>
+
+#include <span>
+
+namespace FEX::HLE {
+struct ThreadStateObject;
+
+class ThunkHandler : public FEXCore::ThunkHandler {
+public:
```
lioncash, line 11: Should this interface have a virtual destructor?

lioncash (approved):

PR #4077, opened 2024-09-22
SignalDelegator: Let the frontend inform AVX support

Sonicadvance1:
Since the frontend has changed to informing the backend if AVX is supported, there is no reason to feed that configuration back in to SignalDelegator from the backend.

Instead inform the SignalDelegator directly in the frontend instead of this now weird round-about path.


alyssarosenzweig:
Why don't we support AVX unconditionally?

Sonicadvance1:
> Why don't we support AVX unconditionally?

It's disabled on 32-bit out of concern for stack usage breaking some games

lioncash (approved):

PR #4078, opened 2024-09-22
Config: Fixes stack usage in EnvLoader

Sonicadvance1:
EnvLoader::Load was generating strings on the stack and then copying them to the FEXCore::Config state. FEXCore::Config::Set supports string_views directly which will emplace in to the map directly.

This improves this functions stack usage from ~4464 bytes to ~288 bytes.

I believe this was missed when we were converting from std::string to fextl::string.


lioncash (approved):
Nice!

PR #4079, opened 2024-09-22
Fix (v)maxss on non-AFP platforms

alyssarosenzweig:
Fixes miscompiles in the vulkan cts and probably more.


unittests/ASM/FEX_bugs/MinMaxNaN.asm
```diff
@@ -0,0 +1,118 @@
...
+%endmacro
+
+%macro max 3
+  single_case max, %1, %2, %3
+%endmacro
+
+zero equ 0x00000000
+negzero equ 0x80000000
+qnan equ 0x7fc00000
+snan equ 0x7fc00000
```
Sonicadvance1, line 61: Is snan and qnan expected to be the same here?

alyssarosenzweig: doh. No, of course not, copypaste fail

alyssarosenzweig: fixed, thanks!

Sonicadvance1 (approved):
nice catch!

PR #4081, opened 2024-09-22
Config: Be a bit smarter in ReloadMetaLayer

Sonicadvance1:
Reduces stack usage from ~1024 bytes to ~704 bytes and stops making a bunch of string copies when they can just be pointers instead.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.cpp
+++ b/FEXCore/Source/Interface/Config/Config.cpp
@@ -230,27 +230,26 @@ void Load() {
   }
 }
 
-fextl::string ExpandPath(const fextl::string& ContainerPrefix, fextl::string PathName) {
+fextl::string ExpandPath(const fextl::string& ContainerPrefix, const fextl::string& PathName) {
   if (PathName.empty()) {
     return {};
   }
-
 
   // Expand home if it exists
   if (FHU::Filesystem::IsRelative(PathName)) {
     fextl::string Home = getenv("HOME") ?: "";
     // Home expansion only works if it is the first character
     // This matches bash behaviour
-    if (PathName.at(0) == '~') {
-      PathName.replace(0, 1, Home);
-      return PathName;
+    if (PathName.starts_with("~/")) {
+      Home.append(PathName.begin() + 1, PathName.end());
+      return Home;
     }
 
     // Expand relative path to absolute
     char ExistsTempPath[PATH_MAX];
     char* RealPath = FHU::Filesystem::Absolute(PathName.c_str(), ExistsTempPath);
-    if (RealPath) {
-      PathName = RealPath;
+    if (RealPath && FHU::Filesystem::Exists(RealPath)) {
+      return RealPath;
     }
 
     // Only return if it exists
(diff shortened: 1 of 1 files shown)


lioncash (approved):

FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -318,65 +316,61 @@ fextl::string FindContainerPrefix() {
...
 
-  if (FEXCore::Config::Exists(FEXCore::Config::CONFIG_CACHEOBJECTCODECOMPILATION)) {
-    FEX_CONFIG_OPT(CacheObjectCodeCompilation, CACHEOBJECTCODECOMPILATION);
-  }
-
-  fextl::string ContainerPrefix {FindContainerPrefix()};
-  auto ExpandPathIfExists = [&ContainerPrefix](FEXCore::Config::ConfigOption Config, fextl::string PathName) {
-    auto NewPath = ExpandPath(ContainerPrefix, PathName);
+  const fextl::string ContainerPrefix {FindContainerPrefix()};
+  auto ExpandPathIfExists = [&ContainerPrefix](FEXCore::Config::ConfigOption Config, const fextl::string* PathName) {
```
neobrain, line 320: Turning `PathName` into a pointer doesn't help much here AFAICT. We never reused the external storage, and passing in rvalues (as we've done before) will avoid unnecessary copies already.

Sonicadvance1: This isn't where the copy was occuring, the copy was occurring below in `FEX_CONFIG_OPT`, this is just to change the interface to match what `Meta->Get` returns which is a `std::optional<std::string*>` which is a pointer to the internal string tracked by the config system.
This just lets us pass that as a const pointer all the way through to `ExpandPathIfExists`->`ExpandPath` so it's matching the interface.

neobrain: Oof, I see. Can we at least use a const reference to highlight that these pointers won't randomly be null? This seems to be the assumption anyway since they're immediately dereferenced in the function.

FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -230,27 +230,25 @@ void Load() {
...
-
   // Expand home if it exists
   if (FHU::Filesystem::IsRelative(PathName)) {
     fextl::string Home = getenv("HOME") ?: "";
     // Home expansion only works if it is the first character
     // This matches bash behaviour
     if (PathName.at(0) == '~') {
-      PathName.replace(0, 1, Home);
-      return PathName;
+      return fextl::fmt::format("{}{}", Home, std::string_view(PathName).substr(1));
```
neobrain, line 244: Allow reuse of memory allocated by `Home`:
```suggestion
      Home.append(PathName.begin() + 1, PathName.end());
      return Home;
```

FEXCore/Source/Interface/Config/Config.cpp
Sonicadvance1: Sure, switched over to a const ref

neobrain (approved):

FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -230,27 +230,26 @@ void Load() {
...
     return {};
   }
 
-
   // Expand home if it exists
   if (FHU::Filesystem::IsRelative(PathName)) {
     fextl::string Home = getenv("HOME") ?: "";
     // Home expansion only works if it is the first character
     // This matches bash behaviour
     if (PathName.at(0) == '~') {
```
neobrain, line 243: Side note: Filenames can start with a tilde  😩 Should we check for `PathName.starts_with("~/")` instead hence?

neobrain: Err, nevermind. Since this is just Config code we can assume the filename is something reasonable.

Sonicadvance1: Doesn't really matter yea. if the user footgunned themselves with `~Config.json` or whatever then it's going to fail regardless.

PR #4082, opened 2024-09-22
FEXLoader: Fixes newer wine versions and Fedora

Sonicadvance1:
This was brought up by #3831 but I finally got the courage to look at the hard problem.

Although I'm only tackling half of the problem with this PR, which is that FEXLoader needs to strip the rootfs path from the executed path if it begins with the rootfs, plus some changes to the surrounding code.

The primary concern here is that when an application has been executed under FEX, specifically through binfmt_misc, then FEX needs to prepend the full rootfs path otherwise Linux can't find the program. Additionally execveat with an FD will resolve a full path to the rootfs.

So past FEX's initial setup, we need to strip off the rootfs path to provide an "absolute" path that is visible to the guest application later. Which is kind of funny since we have a `RootFSRedirect` function which did the exact opposite. This was due to legacy problems in the original ELFLoader that couldn't handle symlinks correctly, which has since been resolved, so that no longer needs to exist.

There was also some weirdness in `GetApplicationNames` where the passed in argument list was modifying Args[0] and then saving the Program as well. Which I just got rid of. Also stopped passing in the arguments by value because....why did I write it like that?

In InterpreterHandler we now need to check if we can open the path inside the rootfs or fallback without it. Plus I had to change the shebang handling so it stopped prefixing the rootfs AGAIN. Took the time to change the shebang handling there so it stops creating string copies and instead just generates views.

Overall this fixes a fairly major flaw with how we were representing `/proc/self` to the application, which was breaking wine since it would prefix the rootfs multiple times, which was weird.

It doesn't address the remaining problem in #3831, which is that applications can still see some of the leaky abstractions with symlinks through the rootfs, but I want to get at least this step in.

### DIFF
--- a/Source/Common/Config.h
+++ b/Source/Common/Config.h
@@ -40,7 +40,7 @@ struct PortableInformation {
  *
  * @return The application name and path structure
  */
-ApplicationNames GetApplicationNames(fextl::vector<fextl::string> Args, bool ExecFDInterp, int ProgramFDFromEnv);
+ApplicationNames GetApplicationNames(const fextl::vector<fextl::string>& Args, bool ExecFDInterp, int ProgramFDFromEnv);
 
 /**
  * @brief Loads the FEX and application configurations for the application that is getting ready to run.
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -38,6 +38,7 @@ $end_info$
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
 #include <FEXHeaderUtils/Filesystem.h>
+#include <FEXHeaderUtils/StringArgumentParser.h>
 
 #include <atomic>
 #include <cerrno>
@@ -135,63 +136,44 @@ private:
 };
 } // namespace AOTIR
 
-void InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fextl::vector<fextl::string>* args) {
-  // Open the Filename to determine if it is a shebang file.
-  int FD = open(Filename->c_str(), O_RDONLY | O_CLOEXEC);
+bool InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fextl::vector<fextl::string>* args) {
+  int FD {-1};
+
+  // Attempt to open the filename from the rootfs first.
+  FD = open(fextl::fmt::format("{}{}", RootFS, *Filename).c_str(), O_RDONLY | O_CLOEXEC);
   if (FD == -1) {
-    return;
+    // Failing that, attempt to open the filename directly.
+    FD = open(Filename->c_str(), O_RDONLY | O_CLOEXEC);
+    if (FD == -1) {
+      return false;
+    }
   }
 
   std::array<char, 257> Header;
   const auto ChunkSize = 257l;
   const auto ReadSize = pread(FD, &Header.at(0), ChunkSize, 0);
+  close(FD);
 
   const auto Data = std::span<char>(Header.data(), ReadSize);
 
   // Is the file large enough for shebang
   if (ReadSize <= 2) {
-    close(FD);
-    return;
+    return false;
   }
 
   // Handle shebang files
   if (Data[0] == '#' && Data[1] == '!') {
-    fextl::string InterpreterLine {Data.begin() + 2, // strip off "#!" prefix
-                                   std::find(Data.begin(), Data.end(), '\n')};
-    fextl::vector<fextl::string> ShebangArguments {};
-
-    // Shebang line can have a single argument
-    fextl::istringstream InterpreterSS(InterpreterLine);
-    fextl::string Argument;
-    while (std::getline(InterpreterSS, Argument, ' ')) {
-      if (Argument.empty()) {
-        continue;
-      }
-      ShebangArguments.push_back(std::move(Argument));
-    }
+    std::string_view InterpreterLine {Data.begin() + 2, // strip off "#!" prefix
+                                      std::find(Data.begin(), Data.end(), '\n')};
+    const auto ShebangArguments = FHU::ParseArgumentsFromString(InterpreterLine);
 
     // Executable argument
-    fextl::string& ShebangProgram = ShebangArguments[0];
-
-    // If the filename is absolute then prepend the rootfs
-    // If it is relative then don't append the rootfs
-    if (ShebangProgram[0] == '/') {
-      ShebangProgram = RootFS + ShebangProgram;
-    }
-    *Filename = ShebangProgram;
+    *Filename = ShebangArguments.at(0);
 
     // Insert all the arguments at the start
     args->insert(args->begin(), ShebangArguments.begin(), ShebangArguments.end());
   }
-  close(FD);
-}
-
-void RootFSRedirect(fextl::string* Filename, const fextl::string& RootFS) {
-  auto RootFSLink = ELFCodeLoader::ResolveRootfsFile(*Filename, RootFS);
-
-  if (FHU::Filesystem::Exists(RootFSLink)) {
-    *Filename = RootFSLink;
-  }
+  return true;
 }
 
 FEX::Config::PortableInformation ReadPortabilityInformation() {
@@ -435,10 +417,21 @@ int main(int argc, char** argv, char** const envp) {
   FEXCore::Profiler::Init();
   FEXCore::Telemetry::Initialize();
 
-  RootFSRedirect(&Program.ProgramPath, LDPath());
-  InterpreterHandler(&Program.ProgramPath, LDPath(), &Args);
+  if (Program.ProgramPath.starts_with(LDPath())) {
+    // From this point on, ProgramPath needs to not have the LDPath prefixed on to it.
+    auto RootFSLength = LDPath().size();
+    if (Program.ProgramPath.at(RootFSLength) != '/') {
+      // Ensure the modified path starts as an absolute path.
+      // This edge case can occur when ROOTFS ends with '/' and passed a path like `<ROOTFS>usr/bin/true`.
+      --RootFSLength;
+    }
 
-  if (!ExecutedWithFD && FEXFD == -1 && !FHU::Filesystem::Exists(Program.ProgramPath)) {
+    Program.ProgramPath.erase(0, RootFSLength);
+  }
+
+  bool ProgramExists = InterpreterHandler(&Program.ProgramPath, LDPath(), &Args);
+
+  if (!ExecutedWithFD && FEXFD == -1 && !ProgramExists) {
     // Early exit if the program passed in doesn't exist
     // Will prevent a crash later
     fextl::fmt::print(stderr, "{}: command not found\n", Program.ProgramPath);
--- a/FEXHeaderUtils/FEXHeaderUtils/StringArgumentParser.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/StringArgumentParser.h
@@ -1,0 +1,39 @@
+#pragma once
+#include <FEXCore/fextl/vector.h>
+#include <FEXCore/fextl/fmt.h>
+
+#include <algorithm>
+#include <string_view>
+
+namespace FHU {
+
+/**
+ * @brief Parses a string of arguments, returning a vector of string_views.
+ *
+ * @param ArgumentString The string of arguments to parse
+ *
+ * @return The array of parsed arguments
+ */
+static inline fextl::vector<std::string_view> ParseArgumentsFromString(const std::string_view ArgumentString) {
+  fextl::vector<std::string_view> Arguments;
+
+  auto Begin = ArgumentString.begin();
+  auto ArgEnd = Begin;
+  const auto End = ArgumentString.end();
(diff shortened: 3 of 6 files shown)


Source/Common/Config.h
```diff
@@ -40,7 +40,7 @@ struct PortableInformation {
  *
  * @return The application name and path structure
  */
-ApplicationNames GetApplicationNames(fextl::vector<fextl::string> Args, bool ExecFDInterp, int ProgramFDFromEnv);
+ApplicationNames GetApplicationNames(const fextl::vector<fextl::string>& Args, bool ExecFDInterp, int ProgramFDFromEnv);
```
neobrain, line 43: This `const` is misleading: The container itself isn't modified, but the data is being moved from it.

neobrain:
Could you add some concrete examples of what this PR is aiming to fix? From the description it's a bit difficult to reconstruct what exactly the problematic cases here are. (As a general suggestion, please avoid mixing in rants about `split_view` into high-level overviews since that gets in the way of communicating the line of thought.)

Source/Common/Config.h
Sonicadvance1: Oop, forgot to remove the move, we actually want the copy since `RecoverGuestProgramFilename` mutates it.
Resolved.

Sonicadvance1:
> Could you add some concrete examples of what this PR is aiming to fix? From the description it's a bit difficult to reconstruct what exactly the problematic cases here are. 

The concrete example is that `FEXBash winecfg` doesn't work; It exits out when wine-preloader tries to execute wine32. This is due to us accidentally providing a fully resolved path of the application all the way to the guest. This causes wine to try and execute a path that is `<Rootfs_Prefix>/<RootFS_Prefix>/usr/bin/wine32` and then execve returns ENOENT.

FEX initially needs the path fully resolved to actually load the application, but then we need to strip off the rootfs prefix (And NOT reapply it!). This fixes the logic inside of FileManagement with proc/self and syscalls so Wine can properly execute its subprocess applications.

> (As a general suggestion, please avoid mixing in rants about `split_view` into high-level overviews since that gets in the way of communicating the line of thought.)

Removed the sentence.

neobrain:
As discussed externally, this is what's going on:

## Overview

When launching programs contained in the RootFS, FEX could previously report invalid program names to the guest (via /proc/self/exe and argv[0]) by double-prepending the RootFS path. This could produce paths like /home/ryanh/.fex-emu/RootFS/Ubuntu_24_04/home/ryanh/.fex-emu/RootFS/Ubuntu_24_04/usr/bin/winecfg.

This could break programs like Wine that read /proc/self/exe to determine their installation path.

To resolve the issue, FEX now consistently strips away the RootFS prefix from executable paths reported to the guest through /proc/self/exe and argv[0].

## Implementation details

Fixes come from multiple places.

* Config.cpp: RecoverGuestProgramFilename: Was modifying Args[0] in-place. This would later get reused in...
* FEXLoader.cpp: InterpreterHandler & RootFSRedirect: Was reading Args[0] & ProgramName respectively with rootfs prepended
* Returned values from Config.cpp was stored in CONFIG_APP_FILENAME which was then read by...
* FileManagement.cpp: FileManager::GetSelf: Which results in symlinkat emulation to return the fully qualified <ROOTFS>/<inside rootfs path> which would feed in to the first two values via...
* Syscalls.cpp: ExecveHandler - Which would check to see if the path exists and pass it directly to the newly executed interpreter.

asahilina:
FWIW I rebased #3831 on top of this (which replaces one of the two commits there) and it works as intended, plus Proton is now working fine again too (that was due to a regression in #3831, not this PR).

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -150,47 +156,40 @@ void InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fe
...
     // Executable argument
-    fextl::string& ShebangProgram = ShebangArguments[0];
-
-    // If the filename is absolute then prepend the rootfs
-    // If it is relative then don't append the rootfs
-    if (ShebangProgram[0] == '/') {
-      ShebangProgram = RootFS + ShebangProgram;
-    }
-    *Filename = ShebangProgram;
+    *Filename = ShebangArguments[0];
```
neobrain, line 186: Looks like this was already an issue before, but: This should use `.at(0)` instead so that malformed shebang lines are handled in any other way than UB.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -150,47 +156,40 @@ void InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fe
   // Is the file large enough for shebang
   if (ReadSize <= 2) {
     close(FD);
-    return;
+    return true;
```
neobrain, line 159: Why does this return success?

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -404,10 +403,14 @@ int main(int argc, char** argv, char** const envp) {
   FEXCore::Profiler::Init();
   FEXCore::Telemetry::Initialize();
 
-  RootFSRedirect(&Program.ProgramPath, LDPath());
-  InterpreterHandler(&Program.ProgramPath, LDPath(), &Args);
+  // Program.ProgramPath is highly likely to be prefixed with FEX's rootfs. Get rid of that.
```
neobrain, line 406: > // Program.ProgramPath is highly likely to be prefixed with FEX's rootfs

Why?

neobrain, line 406: This comment explains the obvious by repeating what this simple code does. It should instead explain *why* this is done.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -404,10 +403,14 @@ int main(int argc, char** argv, char** const envp) {
...
+  if (Program.ProgramPath.starts_with(LDPath())) {
+    Program.ProgramPath = Program.ProgramPath.substr(LDPath().size());
+  }
```
neobrain, line 409: Is there any chance this will result in a relative path? If not, how is it prevented? It would probably be undesirable behavior.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -150,47 +156,40 @@ void InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fe
...
-    while (std::getline(InterpreterSS, Argument, ' ')) {
-      if (Argument.empty()) {
-        continue;
+    using namespace std::literals;
+    auto Begin = InterpreterLine.begin();
+    auto ArgEnd = Begin;
+    auto End = InterpreterLine.end();
+    const auto Delim = " "sv;
+    for (; ArgEnd != End && Begin != End; Begin = ArgEnd + 1) {
+      // Find the end
```
neobrain, line 175: This comment was somewhat helpful with the long `find_first_of` expression after, but once find is used this is actually more confusing than helpful. (Which end? End of the line or of the current word?)

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -404,10 +403,14 @@ int main(int argc, char** argv, char** const envp) {
   FEXCore::Profiler::Init();
   FEXCore::Telemetry::Initialize();
 
-  RootFSRedirect(&Program.ProgramPath, LDPath());
-  InterpreterHandler(&Program.ProgramPath, LDPath(), &Args);
+  // Program.ProgramPath is highly likely to be prefixed with FEX's rootfs. Get rid of that.
+  if (Program.ProgramPath.starts_with(LDPath())) {
+    Program.ProgramPath = Program.ProgramPath.substr(LDPath().size());
```
neobrain, line 408: Avoid the extra memory allocation:
```suggestion
    Program.ProgramPath.erase(0, LDPath().size());
```

neobrain (commented):
Reasonable change overall, just some comments to polish the implementation a little.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Do I need to explain again in the comment that execve will always have the full host qualified path to the executable inside the rootfs?

Source/Tools/FEXLoader/FEXLoader.cpp
neobrain: Is there a chance we accidentally strip one '/' off to many, though? For example, does it make a difference if the user configures the RootFS as "/home/user/rootfsfolder/" vs "/home/user/rootfsfolder"?

Another possible failure path: Can the program paths end up containing "..", escaping the RootFS?

neobrain: To clarify, I'm not concerned about LDPath being relative, but about the result of calling `substr()`.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Function is returning if the file exists. Doesn't really matter, trying to execute a malformed file that is 2 bytes or less will error out anyway. Switched it to false.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: > Another possible failure path: Can the program paths end up containing "..", escaping the RootFS?

Probably, but this doesn't change any behaviour around that anyway. It'll be broken before and it'll be broken afterwards.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -151,47 +157,42 @@ void InterpreterHandler(fextl::string* Filename, const fextl::string& RootFS, fe
...
-    while (std::getline(InterpreterSS, Argument, ' ')) {
-      if (Argument.empty()) {
-        continue;
+    auto Begin = InterpreterLine.begin();
+    auto ArgEnd = Begin;
+    auto End = InterpreterLine.end();
+    while (ArgEnd != End && Begin != End) {
+      // The end of an argument ends with a space or the end of the interpreter line.
+      ArgEnd = std::find(Begin, End, ' ');
+      if (Begin == ArgEnd) {
```
neobrain, line 176: Won't this erroneously return immediately for inputs like `#! /usr/bin/env sh`?

I think changing to `ArgEnd == End` and moving it in front of the `Begin = ArgEnd + 1;` is correct here. The `ArgEnd != End` can then be removed from the loop condition, and `ArgEnd` doesn't need to be declared outside of the loop.

Source/Tools/FEXLoader/FEXLoader.cpp
```diff
@@ -405,10 +406,14 @@ int main(int argc, char** argv, char** const envp) {
   FEXCore::Profiler::Init();
   FEXCore::Telemetry::Initialize();
 
-  RootFSRedirect(&Program.ProgramPath, LDPath());
-  InterpreterHandler(&Program.ProgramPath, LDPath(), &Args);
+  if (Program.ProgramPath.starts_with(LDPath())) {
+    // From this point on, ProgramPath needs to not have the LDPath prefixed on to it.
```
neobrain, line 410: It's probably better to move towards RootFS long-term since that's what every other place referring to that setting calls it.

BTW, is ProgramPath used for anything else (other than `/proc/self/exe` and `argv[0]`) after this point? Is removing the prefix the right choice for all other uses as well?

Sonicadvance1: > It's probably better to move towards RootFS long-term since that's what every other place referring to that setting calls it.

A follow-up PR can be done to rename the variable names from `LDPath` to `RootFS`, there's a couple of places that can be hit at once.
Additional followup PR should be to rename `APP_FILENAME` to `APP_FILEPATH` because it's not the filename but the fully qualified guest path. But that should be independent.

> BTW, is ProgramPath used for anything else (other than `/proc/self/exe` and `argv[0]`) after this point? Is removing the prefix the right choice for all other uses as well?

* It also gets passed to the ELFCodeLoader, but it correctly handles the rootfs prefixing as necessary.
* It gets passed to seccomp, but only for a log.
* It gets passed to gdbserver, but only for removing the  executable from the segment map for some reason?
  * Probably something to check on later, won't really affect this.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Slightly changed the code doing some more thorough testing, to ensure the beginning `/` isn't stripped off.

Source/Tools/FEXLoader/FEXLoader.cpp
Sonicadvance1: Changed this around to live in FHU so we can unittest the thing. It now gets properly unittested and works as expected.

Sonicadvance1:
Reworked a bit of this and now it's less frenetic. Adding unittests for the argument parsing and ensuring the ProgramPath is always absolute if prefixed with the rootfs path. Will be good to get this in before release.

Sonicadvance1:
Making noise.

alyssarosenzweig (approved):

PR #4085, opened 2024-09-26
Fix FScale'ing zero that should return zero

pmatos:
Added tests will fail without the current patch.

### DIFF
--- a/FEXCore/Source/Common/SoftFloat.h
+++ b/FEXCore/Source/Common/SoftFloat.h
@@ -262,6 +262,10 @@ struct FEX_PACKED X80SoftFloat {
 
     return Result;
 #else
+    extFloat80_t Zero {0, 0};
+    if (extF80_eq(state, lhs, Zero)) {
+      return lhs;
+    }
     X80SoftFloat Int = FRNDINT(state, rhs, softfloat_round_minMag);
     LIBRARY_PRECISION Src2_d = Int.ToFMax(state);
     Src2_d = exp2l(Src2_d);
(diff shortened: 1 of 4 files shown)


pmatos:
This pcem issue https://github.com/sarah-walker-pcem/pcem/commit/2e037b50760c4ee35b86104e5102bbc1cb1881a5#diff-56c35b563dbfc08142a6ac0ee8bd3c626b5556065b41f39732fdb3625982064cR677 made me look into this in FEX.

The table in https://www.felixcloutier.com/x86/fscale clearly shows that `fscale` should return zero unless st0 is either NaN or +inf.

Screenshot From 2024-09-26 13-20-17

Sonicadvance1 (approved):
Nice catch!

Sonicadvance1:
Had a thought, does this return the correct sign of zero?

pmatos:
> Had a thought, does this return the correct sign of zero?

Good point! It doesn't. I assumed it didn't matter. I am still unsure how it deals with the exceptional cases of st1 being +inf or Nan, which return NaN, or when st1 is -inf, in which case it returns -/+ 0. There's an odd asymmetry here. Unfortunately, I cannot run MDK2 yet.

pmatos:
> Had a thought, does this return the correct sign of zero?

It does now! :) 
Updated code and tests.

pmatos:
> > Had a thought, does this return the correct sign of zero?
> 
> It does now! :) Updated code and tests.

ugh - clearly something is not working 100%.

pmatos:
I am slightly dumbfounded that after much searching, the crashes were due to an include of `softfloat_types.h` into `SoftFloat.h`. To make matters worse, my editor keeps wanting to add it to the changeset without letting me know. Really need to understand which feature this is and disable it.

lioncash (approved):

PR #4088, opened 2024-09-27
ARM64EC: Dynamically determine syscall numbers under wine.

bylaws:
Takes advantage of the consistent alphabetical syscall ordering that wine follows. This finally allows the CEF sandbox to work with arm64ec FEX under wine.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -69,10 +69,9 @@ uint32_t NtDllRedirectionLUTSize;
 
 // Wine doesn't support issuing direct system calls with SVC, and unlike Windows it doesn't have a 'stable' syscall number for NtContinue
 void* WineSyscallDispatcher;
-// TODO: this really shouldn't be hardcoded, once wine gains proper syscall thunks this can be dropped.
-uint64_t WineNtContinueSyscallId = 0x1a;
-uint64_t WineNtAllocateVirtualMemorySyscallId = 0xb;
-uint64_t WineNtProtectVirtualMemorySyscallId = 0x76;
+uint64_t WineNtContinueSyscallId;
+uint64_t WineNtAllocateVirtualMemorySyscallId;
+uint64_t WineNtProtectVirtualMemorySyscallId;
 
 NTSTATUS NtContinueNative(ARM64_NT_CONTEXT* NativeContext, BOOLEAN Alert);
 NTSTATUS NtAllocateVirtualMemoryNative(HANDLE, PVOID*, ULONG_PTR, SIZE_T*, ULONG, ULONG);
@@ -151,7 +150,6 @@ bool IsDispatcherAddress(uint64_t Address) {
 
 
 void FillNtDllLUTs(HMODULE NtDll) {
-  NtDllBase = reinterpret_cast<uintptr_t>(NtDll);
   ULONG Size;
   const auto* LoadConfig =
     reinterpret_cast<_IMAGE_LOAD_CONFIG_DIRECTORY64*>(RtlImageDirectoryEntryToData(NtDll, true, IMAGE_DIRECTORY_ENTRY_LOAD_CONFIG, &Size));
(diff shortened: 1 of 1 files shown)


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -196,6 +194,31 @@ void PatchCallChecker() {
...
+  ULONG Size;
+  const auto* Exports = reinterpret_cast<IMAGE_EXPORT_DIRECTORY*>(RtlImageDirectoryEntryToData(NtDll, true, IMAGE_DIRECTORY_ENTRY_EXPORT, &Size));
+  const auto* NameTable = reinterpret_cast<uint32_t*>(NtDllBase + Exports->AddressOfNames);
+
+  // Wine orders syscalls alphabetically, take advantage of that to find the syscall indices for those which we need to
+  // manually issue. Note that all functions starting with Nt besides NtGetTickCount are syscalls.
+  uint32_t CurSyscallId = 0;
+  for (uint32_t Idx = 0; Idx < Exports->NumberOfNames; Idx++) {
+    const char* Name = reinterpret_cast<const char*>(NtDllBase + NameTable[Idx]);
+    if (Name[0] == 'N' && Name[1] == 't' && strcmp(Name, "NtGetTickCount")) {
```
Sonicadvance1, line 208: Does this strcmp against `NtGetTickCount` actually do anything since you're going to immediately follow it up with an exact name check against the three syscalls we care about?
I can understand checking that the first two characters are `Nt` at least.

bylaws: It's done so that CurSyscallId isn't incremented for it

Sonicadvance1: ah! I see. I misread the code. Good to know

Sonicadvance1 (approved):
Hopefully at some point we can get wine to export a syscall entry table so we don't have to do this. But until then this is fine.

bylaws:
> Hopefully at some point we can get wine to export a syscall entry table so we don't have to do this. But until then this is fine.

Yeah that won't happen :) There's no reason for wine to deviate from windows here when native apps can get syscall exports just fine through a bunch of different ways. The way done here is wine specific but if wine were to change their layout to match windows it would then be the case that hardcoding as is done for windows would work fine.

Source/Windows/ARM64EC/Module.cpp
```diff
@@ -196,6 +194,31 @@ void PatchCallChecker() {
...
+        WineNtContinueSyscallId = CurSyscallId;
+      } else if (!strcmp(Name, "NtAllocateVirtualMemory")) {
+        WineNtAllocateVirtualMemorySyscallId = CurSyscallId;
+      } else if (!strcmp(Name, "NtProtectVirtualMemory")) {
+        WineNtProtectVirtualMemorySyscallId = CurSyscallId;
+      }
+
+      CurSyscallId++;
+    }
+  }
```
neobrain, line 219: Checking for string inequality by casting strcmp to bool is unnecessarily clever and seems to have already been misinterpreted once.

Using string_view avoids this issue:
```suggestion
    std::string_view Name = reinterpret_cast<const char*>(NtDllBase + NameTable[Idx]);
    if (Name.starts_with("Nt") && Name != "NtGetTickCount") {
      if (Name == "NtContinue") {
        WineNtContinueSyscallId = CurSyscallId;
      }
      // ...

      CurSyscallId++;
    }
  }
```

bylaws: right, was avoiding string view out of caution here so it was very clear exactly what ntdll library calls were made (as is said in comments, until our call checker is initialised we have to be incredibly cautious of what we call as who-knows-what exports could be patched), there's a small chance that some called by InitSyscalls will need to be reimplemented purely within FEX as more drm schemes are tested. That being said checking the resulting binary it seems fine in terms of export usage, so will let you judge if the readability benefit outweighs the above (will also  namespace such functions together so it's harder to miss the context on their limitations)

neobrain: *not* using `string_view` is also fine, but if we need to compromise for weaker abstractions due to this extra level of caution, we should put an equal effort into compensating the cost of this.

Using an early continue and an explicit comparison, along with a comment to express intent, would achieve this:
```
// Skip non-Nt syscalls and NtGetTickCount
if (Name[0] != 'N' || Name[1] != 't' || strcmp(Name, "NtGetTickCount") == 0){
  continue;
}
```

bylaws: made the != 0 explicit and added more info to comments

Sonicadvance1 (approved):

PR #4089, opened 2024-09-27
ARM64EC: AVX register save/restore support under wine

bylaws:
Requires corresponding a wine patch to function:
https://github.com/bylaws/wine/commit/7d411bd74bd03ba90cf202623560733cb229cc7d
^^ I am unsure if it makes sense to link this somewhere in the FEX repo itself, but I will on the wiki page either way

Since AVX still works in almost all cases without this, I chose not to e.g. detect the patch and disable AVX support when it is missing.


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -291,7 +291,13 @@ static void LoadStateFromECContext(FEXCore::Core::InternalThreadState* Thread, C
 
   if ((Context.ContextFlags & CONTEXT_FLOATING_POINT) == CONTEXT_FLOATING_POINT) {
     // Floating-point register state
-    CTX->SetXMMRegistersFromState(Thread, reinterpret_cast<const __uint128_t*>(Context.FltSave.XmmRegisters), nullptr);
+    if ((Context.ContextFlags & CONTEXT_XSTATE) == CONTEXT_XSTATE) {
```
Sonicadvance1, line 294: Is this equality operator expecting `CONTEXT_XSTATE` or is it meant to be checking for `CONTEXT_ARM64_FEX_YMMSTATE` which is set during the store state side of things?

bylaws: The context here is an x64 context which stores avx state as part of xstate (this is the same as on regular x64 wine), the CONTEXT_ARM64_FEX_YMMSTATE flag is for indicating that an arm64ec context packed into an arm64 context contains avx state in the higher 16 vector registers. See the corresponding wine patch for the translation between the two

Sonicadvance1: Ah, okay I see what you mean after looking closelier at the wine patch.

Sonicadvance1 (dismissed):

bylaws:
DNM, this will break on main

bylaws:
Should be fine to merge now

Sonicadvance1 (approved):

PR #4091, opened 2024-09-30
Allow NZCVSelect to work on FPR register generating fcsel

pmatos:
@alyssarosenzweig This is causing regalloc to fail. I wonder if it's easier, or even the only way to do it, to have a  `NZCVSelectV` for `fcsel`.


pmatos:
Closing - since #4092 approach works.

PR #4092, opened 2024-09-30
Implement NZCVSelectV for selection on FPRs

pmatos:
Behaves like NZCVSelect for FPRs.

PR #4093, opened 2024-09-30
Fix FXTRACT for 0.0 and -0.0

pmatos:
Fixes fxtract by returning the correct values for 0.0 and -0.0. We moved the split of fxtract into _sig and _exp, to the opcode dispatcher, to ease some comparisons.

Also removed the IR node F80XTRACTStack which is not needed anymore.

Depends on landing #4092 (commit d62dd50d7f174cbd2ac3ec1f4161facdf565c300 included here but to be merged separately)


pmatos:
For context the spec mentions that for fscale:

> The FSCALE instruction can also be used to reverse the action of the FXTRACT instruction, as shown in the following example:
> 
> FXTRACT;
> FSCALE;
> FSTP ST(1);

When testing this, through `FScaleFXtract.asm` unit test, I noticed that for 0.0 and -0.0, fxtract was generating incorrect values. I fixed these there.

pmatos:
All good here now - instcounci failure coming from #4092 that needs to land **before** this one.

pmatos:
pinging here for review! 
One thing I would like another pair of eyes on is the IR definition of `NZCVSelectV`. Do we actually need to specify a ResultSize? Can't we just say that the ResultSize is the same size as the TrueVal (or FalseVal). Also maybe we need to validate that these have the same size?

Sonicadvance1 (approved):
Seems reasonable to me, but I'll leave it to Alyssa for the IR change.

FEXCore/Source/Common/SoftFloat.h
```diff
@@ -256,6 +260,12 @@ struct FEX_PACKED X80SoftFloat {
 
     return Result;
 #else
+    // Zero is a special case, the exponent is always -inf
+    if (lhs.Exponent == 0x0 && lhs.Significand == 0x0) {
+      X80SoftFloat Result(1, 0x7FFFUL, 0x8000000000000000UL);
```
neobrain, line 265: Hardcoded constants with lots of zeroes like these are prone to typos that aren't easily caught in code review either. There's two ways to avoid this issue:
* Compute the constant symbolically: `1UL << 63`
* Use digit separators to make double-checking easier: `0x8000'0000'0000'0000UL`

pmatos:
ping @alyssarosenzweig

alyssarosenzweig:
instead of And+SubNZCV, just use TestNZ to save an instruction

alyssarosenzweig:
`CondClassType {.Val = FEXCore::IR::COND_EQ}` can maybe just be `{COND_EQ}`

alyssarosenzweig:
looks ok save the things i mentioned

alyssarosenzweig:
```
+  Ref ExpNZ = _And(OpSize::i64Bit, Gpr, _Constant(0x7ff0000000000000LL));
+  ExpNZ = _Lshr(OpSize::i64Bit, ExpNZ, _Constant(52));
``` 

can you combine these two into a _Bfe instruction?

alyssarosenzweig:
```
+  Ref SigNZ = _And(OpSize::i64Bit, Gpr, _Constant(0x800fffffffffffffLL));
+  SigNZ = _Or(OpSize::i64Bit, SigNZ, _Constant(0x3ff0000000000000LL));
```

reexpress as `x |( ~(1 << 63))` followed by `x & ~(1 << 63)`. This lets you use constants that will inline, whereas these ones don't, thereby saving a few moves

JosJuice:
0x800f'ffff'ffff'ffff and 0x3ff0'0000'0000'0000 do successfully inline for logical op instructions. You can see it in the InstCountCI output.

alyssarosenzweig:
> 0x800f'ffff'ffff'ffff and 0x3ff0'0000'0000'0000 do successfully inline for logical op instructions. You can see it in the InstCountCI output.

Oops, right. disregard that part!

alyssarosenzweig (changes requested):
as commented

pmatos:
> as commented

Thanks for the comments - I think I have addressed them now.

Sonicadvance1 (approved):

alyssarosenzweig (approved):

PR #4094, opened 2024-09-30
unittests: Be more strict with -- separator

Sonicadvance1:
> We had a bug in our arguments getting passed to unittests for years (https://github.com/FEX-Emu/FEX/pull/4071) and it was undiscovered. Due to a quirk with cpp-optparse where any unsupported arguments were gathered and passed to the guest application. This didn't necessarily break anything in our unittests, just made them run silently, and for all the harness things the additional argument was ignored anyway.

cherry-picked from #4070 with the review comments resolved.

Only the ASM and 32-bit ASM tests were missing this.

Breaking it out in to an independent PR since it can be freestanding.

### DIFF
--- a/unittests/32Bit_ASM/CMakeLists.txt
+++ b/unittests/32Bit_ASM/CMakeLists.txt
@@ -92,7 +92,9 @@ foreach(ASM_SRC ${ASM_SOURCES})
       "${CMAKE_SOURCE_DIR}/unittests/32Bit_ASM/Disabled_Tests_${CPU_CLASS}"
       "Test_32Bit_${REL_TEST_ASM}"
       ${LAUNCH_PROGRAM}
-      ${ARGS_LIST} "${OUTPUT_NAME}" "${OUTPUT_CONFIG_NAME}")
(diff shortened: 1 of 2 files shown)


unittests/32Bit_ASM/CMakeLists.txt
```diff
@@ -83,6 +83,9 @@ foreach(ASM_SRC ${ASM_SOURCES})
 
     set(TEST_NAME "${TEST_DESC}/Test_32Bit_${REL_TEST_ASM}")
     string(REPLACE " " ";" ARGS_LIST ${ARGS})
+
+    list(APPEND ARGS_LIST "--")
```
neobrain, line 87: The logic `ARGS_LIST` is involved in is complex enough already. Can we change the below to this instead?

```
${LAUNCH_PROGRAM}
${ARGS_LIST}
"--"
"{OUTPUT_NAME}..."
```
This makes the structure of the arguments much clearer than hiding the separator in a variable.

neobrain (approved):

PR #4095, opened 2024-10-02
speed up the JIT

alyssarosenzweig:
9% decrease in node time across the series (both with and without multiblock). bit of everything. unfortunatley, no perceivable improvement in steam startup time :crying_cat_face:

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
@@ -139,7 +139,10 @@ private:
 // Constants are pooled per block.
 void ConstProp::HandleConstantPools(IREmitter* IREmit, const IRListView& CurrentIR) {
   const uint32_t SSACount = CurrentIR.GetSSACount();
-  fextl::vector<Ref> Remap(SSACount, NULL);
+
+  // Allocation/initialization deferred until first use, since many multiblocks
+  // don't have constants leftover after all inlining.
+  fextl::vector<Ref> Remap {};
 
   for (auto [BlockNode, BlockIROp] : CurrentIR.GetBlocks()) {
     for (auto [CodeNode, IROp] : CurrentIR.GetCode(BlockNode)) {
@@ -151,11 +154,15 @@ void ConstProp::HandleConstantPools(IREmitter* IREmit, const IRListView& Current
           uint32_t Value = CurrentIR.GetID(CodeNode).Value;
           LOGMAN_THROW_A_FMT(Value < SSACount, "def not yet remapped");
 
+          if (Remap.empty()) {
+            Remap.resize(SSACount, NULL);
+          }
+
           Remap[Value] = it->second;
         } else {
           ConstPool[Op->Constant] = CodeNode;
         }
-      } else {
+      } else if (!Remap.empty()) {
         const uint8_t NumArgs = IR::GetArgs(IROp->Op);
         for (uint8_t i = 0; i < NumArgs; ++i) {
           if (IROp->Args[i].IsInvalid()) {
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -144,23 +142,39 @@ void ConstProp::HandleConstantPools(IREmitter* IREmit, const IRListView& Current
...
+          if (K.Value == Op->Constant) {
+            uint32_t Value = CurrentIR.GetID(CodeNode).Value;
+            LOGMAN_THROW_A_FMT(Value < SSACount, "def not yet remapped");
+
+            if (Remap.empty()) {
+              Remap.resize(SSACount, NULL);
+            }
+
+            Remap[Value] = K.R;
+            Found = false;
```
Sonicadvance1, line 171: Shouldn't this be `Found = true;` ?

FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -139,23 +137,46 @@ class ConstProp final : public FEXCore::IR::Pass {
...
+  // don't have constants leftover after all inlining.
+  fextl::vector<Ref> Remap {};
+
+  struct Entry {
+    int64_t Value;
+    Ref R;
+  };
 
   for (auto [BlockNode, BlockIROp] : CurrentIR.GetBlocks()) {
+    fextl::vector<Entry> Pool {};
```
neobrain, line 151: This will clear and re-allocate memory in each loop iteration. Instead, declaring the vector outside of the loop and just `.clear()`ing it in each iteration allows to minimize redundant allocations.

FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
alyssarosenzweig: Oops, I thought I fixed that. Fixed

FEXCore/Source/Interface/IR/Passes/ConstProp.cpp
```diff
@@ -139,23 +137,46 @@ class ConstProp final : public FEXCore::IR::Pass {
...
+        // Search for the constant. This is O(n^2) but n is small since it's
+        // local and most constants are inlined. In practice, it ends up much
+        // faster than a hash table.
+        for (auto K : Pool) {
+          if (K.Value == Op->Constant) {
+            uint32_t Value = CurrentIR.GetID(CodeNode).Value;
+            LOGMAN_THROW_A_FMT(Value < SSACount, "def not yet remapped");
+
+            if (Remap.empty()) {
+              Remap.resize(SSACount, NULL);
```
neobrain, line 167: `NULL` is so 2010, let's use nullptr for consistent code style.

Sonicadvance1 (approved):

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -461,9 +466,11 @@ void ConstrainedRAPass::Run(IREmitter* IREmit_) {
   auto IR_ = IREmit->ViewIR();
   IR = &IR_;
 
+  // SSAToNewSSA, NewSSAToSSA allocated on first-use
+  SSAToNewSSA.clear();
+  NewSSAToSSA.clear();
```
neobrain, line 471: Wouldn't it make more sense to clear these at the end of `Run()` rather than at the beginning of its next invocation?

The underlying storage won't be freed either way, but it's easier to reason about code when there are clear invariants like "SSAToNewSSA is empty outside of Run()".

alyssarosenzweig:
addressed Tony's comments, and ripped out some code from constprop

Sonicadvance1 (approved):

PR #4106, opened 2024-10-08
FEXCore: adds support for compat input prctl

slp:
The size of the input_event struct differs between 32 bits applications and 64 bits applications. To deal with this, the kernel implements a compat variant for the input syscalls, but it's only enabled for 32 bit processes.

In libkrunfw we're introducing a prctl that enables a 64 bit process to request the kernel to enable the compat variant for the input syscalls. This commit makes use of that interface for enabling/disabling the compat input variant as required.

The visible effect is that input devices such as gamepads work properly on emulated 32 bit applications.

### DIFF
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -287,6 +287,36 @@ void SetupTSOEmulation(FEXCore::Context::Context* CTX) {
 }
 } // namespace FEX::TSO
 
+namespace FEX::CompatInput {
+void SetupCompatInput(bool enable) {
+  // We need to check if these are defined or not. This is a very fresh feature.
+#ifndef PR_GET_COMPAT_INPUT
+#define PR_GET_COMPAT_INPUT 0x63494e50
+#endif
+#ifndef PR_SET_COMPAT_INPUT
+#define PR_SET_COMPAT_INPUT 0x43494e50
+#endif
+#ifndef PR_SET_COMPAT_INPUT_DISABLE
+#define PR_SET_COMPAT_INPUT_DISABLE 0
+#endif
+#ifndef PR_SET_COMPAT_INPUT_ENABLE
+#define PR_SET_COMPAT_INPUT_ENABLE 1
+#endif
+  // Check to see if this is supported.
+  auto Result = prctl(PR_GET_COMPAT_INPUT, 0, 0, 0, 0);
+  if (Result == -1) {
+    // Unsupported, early exit.
+    return;
+  }
+
+  if (enable) {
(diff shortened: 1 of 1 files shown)


slp:
This one needs to be paired with: https://github.com/slp/linux/commit/897560fd00e1add63a9973c31fc437b006855680

Does this kind of feature require a `FEXConfig` tunable too?

Sonicadvance1:
Short term this is probably a fine workaround. Long term I think it'll have problems when we do more 32-bit thunking and then 32-bit and 64-bit libraries are both reading gamepad inputs, but since that's a non-issue for today, it's not a real concern yet.

> Does this kind of feature require a FEXConfig tunable too?

FEXConfig shouldn't matter here, it's a runtime detectable option and would always want it currently.

I assume your kernel code is having processes inherit the prctl value being set, which is why it happens in all processes. Seems reasonable to me.

alyssarosenzweig (approved):
I agree that long-term FEX needs to learn how to do this emulation or we need to get something mainlined, for the benefit of platforms other than Asahi. 

As a hot fix I am so so ok with this. +1.

alyssarosenzweig:
> Long term I think it'll have problems when we do more 32-bit thunking and then 32-bit and 64-bit libraries are both reading gamepad inputs, but since that's a non-issue for today, it's not a real concern yet.

I don't expect Linux FEX to ever thunk anything that would read gamepads on 32-bit. Even thunking things like SDL on 64-bit doesn't seem to vibe with how we've approached thunking in FEX. I know box does it but they make compatibility tradeoffs in doing so that I'd rather FEX not.

slp:
> Short term this is probably a fine workaround. Long term I think it'll have problems when we do more 32-bit thunking and then 32-bit and 64-bit libraries are both reading gamepad inputs, but since that's a non-issue for today, it's not a real concern yet.

Yes, I also consider this a temporary fix.
 
> > Does this kind of feature require a FEXConfig tunable too?
> 
> FEXConfig shouldn't matter here, it's a runtime detectable option and would always want it currently.
> 
> I assume your kernel code is having processes inherit the prctl value being set, which is why it happens in all processes. Seems reasonable to me.

Originally I was clearing the value on fork(), but I've noticed trouble in some 32 bits apps. I didn't dig deep, but my guess is that 32 bits apps may fork without going to through the FEXLoader's main (which does makes sense). Keeping the value around and disabling it for 64 bit apps makes everything work as expected.

Sonicadvance1 (approved):

neobrain:
> I don't expect Linux FEX to ever thunk anything that would read gamepads on 32-bit. Even thunking things like SDL on 64-bit doesn't seem to vibe with how we've approached thunking in FEX. I know box does it but they make compatibility tradeoffs in doing so that I'd rather FEX not.

FTR libSDL is definitely one of the better-behaving APIs. This should be straightforward to forward to host now that we have a better approach to forwarding windowing system use.

PR #4107, opened 2024-10-08
Remove split restore/save from mm7

pmatos:


pmatos:
@Sonicadvance1 I have a feeling that the split in saving/restoring mm7 is something due to past representations of the registers. I can't find a reason for these not. While trying to understand why, I go back a few years to when you split a bunch of code into several files, one of them x87.cpp. 

Removing this split improves instcountci and passes all the tests, so it doesn't feel like it's necessary. Can you take a look? If it's indeed necessary, we should probably add a test that fails when the split in save and restore is removed.

Sonicadvance1:
> @Sonicadvance1 I have a feeling that the split in saving/restoring mm7 is something due to past representations of the registers. I can't find a reason for these now. While trying to understand why, I go back a few years to when you split a bunch of code into several files, one of them x87.cpp.
> 
> Removing this split improves instcountci and passes all the tests, so it doesn't feel like it's necessary. Can you take a look? If it's indeed necessary, we should probably add a test that fails when the split in save and restore is removed.

The split is due to the instruction only saving 10-bytes rather than 16-bytes per register. We're taking advantage of overlapping writes to simulate the same behaviour. The final store still needs to be 10-bytes in order to ensure we don't accidentally write past the end of a page, and also ensure we don't overwrite some other data after the the tail.

This change will break that edge case, we just don't have ASM unittests in place for it.

PR #4108, opened 2024-10-08
unittests/ASM: Ensures FNSAVE and FRSTOR only store as much data as required

Sonicadvance1:
The instruction definition only allows these instructions to load/store 94 or 108 bytes, not affecting any bytes afterwards. This is a bit awkward because 80-bit x87 registers are stored at the end.

FEX has an optimization today where it uses overlapping loads and stores for the first seven x87 registers, and a split loadstore for the final register. This ensures that we get the correct data while reducing the number of loadstores.

We didn't have a unittest in place to ensure we only ever write the correct amount of data, so changes like in #4107 which look correct from an initial glance, would have resulted in broken behaviour.

This unittest ensures both that the instructions don't try to access beyond the end of the page, and also ensures that they don't overwrite subsequent data. Making sure that potentially broken behaviour doesn't make its way in.

### DIFF
--- a/unittests/ASM/FEX_bugs/fnsave_fnrstor_size.asm
+++ b/unittests/ASM/FEX_bugs/fnsave_fnrstor_size.asm
@@ -1,0 +1,54 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RBX": "0x4142434445464748",
+    "RCX": "0x5152535455565758"
+  },
+  "MemoryRegions": {
+    "0x100000000": "4096"
+  }
+}
+%endif
+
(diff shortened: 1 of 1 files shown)


pmatos:
Ah nice - I was going to look into writing this today but you got there before me. Thanks!

pmatos:
LGTM, I just ran this with my *broken* branch and it does fail this test.

Sonicadvance1:
Oh right, didn't say. I did test this by reproducing the changes your branch did. The new test *correctly* caught that it was accessing over the page and also overwriting bytes after the save frame (when end of page test was disabled).

lioncash (approved):

Issue #1119, opened 2021-06-25
Walk through personality flags on 32-bit and 64-bit and ensure correct behaviour


Sonicadvance1:
Not sure exactly how much this matters but we should double check it.

PR #4116, opened 2024-10-11
LinuxEmulation: Personality handling

Sonicadvance1:
Comes from #1119 

Previously we were just passing the personality to the host kernel, which worked for most cases but not all.
- Adds per-thread tracking of personality
- Handles PER_LINUX32 which AArch64-only hardware would have returned error on
- Supports the uname changes that PER_LINUX32 and UNAME26 changes

Flags still not supported after this PR:
- ADDR_COMPAT_LAYOUT
- READ_IMPLIES_EXEC
- ADDR_LIMIT_3GB

Adds a unit test to ensure personality is set and returned as expected.


alyssarosenzweig (approved):
nice!

PR #4125, opened 2024-10-21
OpcodeDispatcher: Unify PSHUF{L,H}W implementations

### LINKED ISSUE #4123: {V,}PSHUF{L,H}W implementation can be unified
Currently the AVX128 implementation does a InsElement dance, the SSE and MMX versions are slightly more efficient in that it understands broadcast and a TBL implementation.

These three versions can be unified, which will immediately improve the perf of the AVX128 implementation.

Related to #3785

Sonicadvance1:
MMX, SSE, and AVX were all using different implementations. Unify the three of them.
This is a nop for the MMX and SSE implementation, but this does some minor optimization for the AVX implementation since the SSE path was slightly more optimal.

Fixes #4123


alyssarosenzweig:
nice!

alyssarosenzweig (approved):

lioncash (approved):

PR #4127, opened 2024-10-22
Optimize bsf, bsr, register cmpxchg, pcmpistri

alyssarosenzweig:
Just some random little things from looking at Secondary.json

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/Arm64/ALUOps.cpp
@@ -1282,15 +1282,12 @@ DEF_OP(FindLSB) {
 
   if (IROp->Size != 8) {
     ubfx(EmitSize, TMP1, Src, 0, IROp->Size * 8);
-    cmp(EmitSize, TMP1, 0);
     rbit(EmitSize, TMP1, TMP1);
   } else {
     rbit(EmitSize, TMP1, Src);
-    cmp(EmitSize, Src, 0);
   }
 
   clz(EmitSize, Dst, TMP1);
-  csinv(EmitSize, Dst, Dst, ARMEmitter::Reg::zr, ARMEmitter::Condition::CC_NE);
 }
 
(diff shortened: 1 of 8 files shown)


unittests/ASM/TwoByte/0F_BC.asm
```diff
@@ -41,6 +38,8 @@ bsf r12w, word  [rdx + 8 * 1]
 bsf r11d, dword [rdx + 8 * 1]
 bsf r10,  qword [rdx + 8 * 1]
 
+; bsf of zero has an undefined result in x86, so we don't check these results.
```
Sonicadvance1, line 41: This isn't entirely true for bsf and bsr, we provide the stricter guarantees from AMD's platform. Intel is the only one with undefined (Although they also don't modify the destination)
> If the second operand contains 0, the instruction sets ZF to 1 and does not change the contents of the destination register

alyssarosenzweig: I see. Is there a good reason to provide the AMD semantics when the documented behaviour is undefined? This undefined behaviour lets us save 2 instructions in the case that flags are discarded (1 instruction when flags are read).

Sonicadvance1: I think you're misunderstanding. The documented behaviour is both, the quote I posted was from AMD's documentation, the quote you posted was from Intel's. Both are correct.

I believe I implemented the "correct" AMD documented behaviour once I saw an application relying on it years ago. Which very well could be the same reason why Intel hardware has the same behaviour but their documentation claims undefined.

alyssarosenzweig: I see. I'll add a comment.

alyssarosenzweig: Updated to keep the old semantics (but implemented more efficiently), dropped the unit test change, and added a comment.

Sonicadvance1 (approved):
nice nice nice

PR #4128, opened 2024-10-22
add Ubuntu 24.10 and remove unsupported releases

ahoneybun:
This adds that release since it is built in the PPA:

https://launchpad.net/~fex-emu/+archive/ubuntu/fex?field.series_filter=oracular

It may be recommended to remove non-LTS releases or at least ones that are no longer supported by Ubuntu perhaps.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -7,7 +7,7 @@ FEX is very much work in progress, so expect things to change.
 
 
 ## Quick start guide
-### For Ubuntu 20.04, 21.04, 21.10, 22.04
+### For Ubuntu 22.04, 24.04 and 24.10
 Execute the following command in the terminal to install FEX through a PPA.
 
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Thanks for the patch, I had forgot to update this script with the last release.
Definitely a good idea to remove the series we no longer support.
https://github.com/FEX-Emu/FEX-ppa/blob/main/create_packages.py#L17-L19 Has the last of series we support.

ahoneybun:
I just made that change now, let me know if you want me to rebase to keep it in one commit.

Sonicadvance1:
That looks great, thanks!

ahoneybun:
Oh I should update the README as well!

Sonicadvance1 (approved):

PR #4129, opened 2024-10-23
Disable RPRES in instcounci files

pmatos:
This was giving false changes on RPRES enabled HW.


Sonicadvance1 (approved):
hah, oops. That makes sense.

PR #4130, opened 2024-10-24
X87 Code Simplification

pmatos:
The instcountci difference is due to a duplicate store we were doing on the F64 side.


Sonicadvance1 (approved):

alyssarosenzweig:
nice!

PR #4132, opened 2024-10-25
Fix IR operation usage to use OpSize when possible

Sonicadvance1:
No functional change.

This is a continued stride towards using OpSize enums instead of constant numbers to reduce confusion and misunderstandings.

It gets very confusing at times when an argument to an IR operation is a size versus an index, or scale or another constant. We had already cleaned up a decent amount of this early this year, late last year but this hits all the OpcodeDispatcher files in the folder. Once this is merged we can start enforcing OpSize passing to IR operations without implicit conversions. Likely will require a bit more mopping up on the edges, but should make it easier.

Just a little cleanup while I'm distracted.


lioncash:
Needs a rebase

Sonicadvance1:
Rebased

lioncash (approved):

PR #4134, opened 2024-10-25
JIT: Moves Arm64 JIT up one folder

Sonicadvance1:
We have only one JIT, there is no reason to subfolder this. Move it up one folder.

NFC


alyssarosenzweig:
appreciate it.

alyssarosenzweig (approved):

lioncash (approved):

PR #4135, opened 2024-10-25
X86Tables: Removes XOP tables

Sonicadvance1:
These weren't even wired up to the frontend. We aren't going to support XOP, so just remove the tables.

NFC


alyssarosenzweig (approved):
memes

lioncash (approved):

PR #4137, opened 2024-10-25
OpcodeDispatcher: Minor optimization to small pushf 

Sonicadvance1:
The push operation already truncates the result, there's no need to bfe
it. Noticed this while cleaning up in https://github.com/FEX-Emu/FEX/pull/4134. Removes one instruction for
16-bit and 32-bit pushf instructions.


alyssarosenzweig:
Sure

alyssarosenzweig (approved):

lioncash (approved):

PR #4138, opened 2024-10-26
FileManagement: Hide the FEX RootFS fd from /proc/self/fd

asahilina:
Chromium/CEF has code that iterates through all open FDs and bails if any are directories (apparently a sandboxing sanity check). To avoid this check, we need to hide the RootFS FD. This requires hooking all the getdents variants to skip that entry.

To keep the runtime cost low, we keep track of the inode of /proc/self/fd/<rootfs fd> (note: not the RootFS inode, the inode of the magic symlink in /proc), and first do a quick check on that. If it matches, then we stat the dirfd we are reading and check against the procfs device, to complete the inode equality check.

As an extra benefit, this also fixes code that tries to iterate and close all/extra FDs and ends up closing the RootFS fd.


Sonicadvance1 (approved):
Seems reasonable to me.

PR #4139, opened 2024-10-26
Linux/EmulatedFiles: Check return for memfd_create & work around hack

asahilina:
xshmwrap is a hack to make libxshmfence use shared memory files for futexes instead of memfds. To do this, it bans memfd_create(*, MFD_ALLOW_SEALING | MFD_CLOEXEC) using seccomp-bpf. This happens to hit the FEX GenTmpFD() code.

If this failure happens, work around it by setting the FD_CLOEXEC flag separately.1


alyssarosenzweig:
upstream fix merged

PR #4140, opened 2024-10-28
Convert all of the IR operations to use OpSize

Sonicadvance1:
No functional change. A couple of minor helpers added that will get removed later, just for type conversion.

Walked through the IR.json and converted all of the IR operations over to use OpSize.


FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp
```diff
@@ -388,7 +388,7 @@ void OpDispatchBuilder::VAESDecLastOp(OpcodeArgs) {
   Ref Key = LoadSource(FPRClass, Op, Op->Src[1], Op->Flags);
   Ref Result = _VAESDecLast(DstSize, State, Key, LoadZeroVector(DstSize));
 
-  StoreResult(FPRClass, Op, Result, -1);
+  StoreResult(FPRClass, Op, Result, OpSize::iInvalid);
```
lioncash, line 391: Perhaps we should have a three operand version of StoreResult, so that iInvalid doesn't need to be super prevalent all over the place

Sonicadvance1: I think that's reasonable, but I would want to leave that refactoring until after this change. Due to the large number of line changes I don't even want to change argument counts yet

Sonicadvance1:
Slammed through the remaining IR operations and now all of them are using OpSize.

FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp
lioncash: Makes sense

lioncash (approved):

saeed6944 (approved):

PR #4147, opened 2024-10-29
IR: Converts base IR operations to store OpSize sizes

Sonicadvance1:
NFC

Finally converts the IR operations themselves to store the OpSize for the IR operation size and element sizes.

This also finally, FINALLY, converts that remaining `_Constant` helper to stop using a size field that is specified in bits rather than bytes like all the other IR op handlers. That thing was so confusing and now it's gone.

### DIFF
--- a/FEXCore/Source/Interface/IR/IR.h
+++ b/FEXCore/Source/Interface/IR/IR.h
@@ -549,6 +549,7 @@ protected:
 // This must directly match bytes to the named opsize.
 // Implicit sized IR operations does math to get between sizes.
 enum OpSize : uint8_t {
+  iUnsized = 0,
   i8Bit = 1,
   i16Bit = 2,
   i32Bit = 4,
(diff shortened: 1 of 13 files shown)


FEXCore/Source/Interface/IR/IR.h
```diff
@@ -607,14 +610,36 @@ static inline uint8_t OpSizeToSize(IR::OpSize Size) {
...
+  LOGMAN_THROW_A_FMT(Size != IR::OpSize::iInvalid, "Invalid Size");
+  return IR::OpSizeToSize(Size) * 8u;
+}
+
 static inline OpSize MultiplyOpSize(IR::OpSize Size, uint8_t Multiplier) {
+  LOGMAN_THROW_A_FMT(Size != IR::OpSize::iInvalid, "Invalid Size");
   return IR::SizeToOpSize(IR::OpSizeToSize(Size) * Multiplier);
 }
 
 static inline OpSize DivideOpSize(IR::OpSize Size, uint8_t Multiplier) {
```
lioncash, line 623: `Multiplier` should probably be `Divisor` here. Though this function seems like it can be merged into the operator variant down below

Sonicadvance1: Ah right, renamed the variable.
Sadly can't yet be folded in to the division operator overload yet since that causes ambiguous compile errors until we can convert the enum to an enum class.

lioncash (approved):

PR #4149, opened 2024-10-29
IR: Convert OpSize over to enum class

Sonicadvance1:
NFC

Do the final mopping up to convert the OpSize enum to an enum class!


alyssarosenzweig (approved):
ack, only skimmed though

lioncash (approved):
did the big read through and everything seems all good

PR #4153, opened 2024-11-04
Add some more tests as unsupported by vixl

pmatos:
It seems a form of `mrs` is unsupported as well as the hint `wfe` used for `pause`.

Remove skipping Rounding(Neg|Pos).asm as they are passing.


pmatos:
humm, in 32 bits the test is not being skipped... converting to draft so it's not merged.

pmatos:
Apparently the two tests in that file that were supposed to be skipped, actually pass and they were not being skipped because they missed the Test_32Bit_ prefix. So removed the useless lines.

Sonicadvance1 (approved):

saeed6944 (approved):

PR #4154, opened 2024-11-06
Check if a candidate home directory exists before using it

Liamolucko:
This allows FEX to be used in situations where `HOME` is set to something invalid, e.g. inside Nix builds.


Sonicadvance1 (approved):
Weird situation but sure

PR #4157, opened 2024-11-12
Support CLONE_FS and CLONE_FILES with fork() semantics

asahilina:
Needed by Discord, part of the Chromium sandbox code. The warning still triggers because Chromium asks for CLONE_VM on x86_64, but that can be safely ignored (CLONE_FS is the one that matters).

I'm not sure if this is exactly the right semantics, but it seems like an improvement at least?

With this and a hack to allow the FEXServer connection to fail (which happens because the sandbox puts the process in a new netns so the UNIX socket doesn't work, not sure what to do about that...), Discord gets as far as the main window and then crashes (without these changes it dies very early).


neobrain:
What's the difference between fork and clone3 as used here? Are they equivalent as long as these new flags aren't set?

asahilina:
> What's the difference between fork and clone3 as used here? Are they equivalent as long as these new flags aren't set?

I *think* so but I haven't found an exact reference for what `clone3()` inputs are strictly identical to `fork()`, so this is just what I gather from skimming the manpage...

Sonicadvance1:
Needs code formatted.

asahilina:
Should be fixed I think!

saeed6944 (approved):

PR #4159, opened 2024-11-12
FEXServer: Listen on both abstract & named sockets

asahilina:
Abstract sockets have one limitation: they are bound to a network namespace. Chromium/CEF sandboxes using a new netns, which breaks connecting to the FEXServer.

To work around this, use and try *both* abstract and named sockets. As long as either the filesystem or the network is unsandboxed, things will work. If both are sandboxed, there isn't much we can do... but at that point we shouldn't be reinitializing the FEXServer connection anyway since the FS should be available on FEXInterpreter startup.

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -96,14 +96,21 @@ fextl::string GetServerRootFSLockFile() {
 }
 
 fextl::string GetTempFolder() {
-  auto XDGRuntimeEnv = getenv("XDG_RUNTIME_DIR");
-  if (XDGRuntimeEnv) {
-    // If the XDG runtime directory works then use that.
-    return XDGRuntimeEnv;
+  const std::array<const char*, 5> Vars = {
+    "XDG_RUNTIME_DIR", "TMPDIR", "TMP", "TEMP", "TEMPDIR",
+  };
+
+  for (auto& Var : Vars) {
+    auto Path = getenv(Var);
+    if (Path) {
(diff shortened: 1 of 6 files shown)


Sonicadvance1:
Looks like you got clipped by the glibc fault test path. Will need a backtrace to see where glibc is attempting to do an allocation in this mode.

asahilina:
I *think* I fixed it? I don't really know how to run the test, but the only code I can think of that would do allocations behind my back is `std::filesystem::temp_directory_path()`... (this was actually already used here in the past when the socket was not abstract, but I think that glibc test didn't exist back then).

Sonicadvance1:
Build FEX with `-DENABLE_GLIBC_ALLOCATOR_HOOK_FAULT=True -DENABLE_JEMALLOC_GLIBC_ALLOC=False` to enable the glibc fault checking. and then for the gcc tests specifically that are failing would be `ninja gcc_target_tests_64`

asahilina:
Oof, it was *both* `std::filesystem::temp_directory_path()` as I suspected *and* `strerror()`. Which means this was always broken, it's just that the messages never got printed during the tests until now... and I had a missing errno check that was causing one of them to log unnecessarily.

Sonicadvance1:
Looks like we should add `std::filesystem::temp_directory_path` to the [Programming Concerns](https://github.com/FEX-Emu/FEX/blob/main/docs/ProgrammingConcerns.md) document.

asahilina:
Added to the docs.

Sonicadvance1 (approved):

PR #4160, opened 2024-11-12
FEXLoader: Align stack base

asahilina:
This ensures that __libc_stack_end is aligned, the same way it is on native.


asahilina:
I think this is actually what was making Steam crash (the V8 GC code expects the stack base to be aligned, and for the renderer process presumably the main thread is relevant for V8, which is why that kept crashing).

Sonicadvance1 (approved):

PR #4161, opened 2024-11-13
FEXCore: Emulate EFLAGS.TF

bylaws:
When set - either via POPF or a thread context operation - the trap flag raises a single step exception after the execution of each instruction.As e.g. a JUMP instruction with TF set will raise an exception at the jump target. Handle this on the FEX side by storing both the flag itself (in bit 0) and a 'block exceptions' flag (in bit 1, inverted). Each generated block when TF is set is then forced to a single instruction with logic to raise the exception at the start. Initially after setting TF exceptions are blocked, then at the start of the block they are unblocked so that after the instruction executes an exception is raised at the start of the next block.

This needs some linux-side testing


FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -711,6 +712,46 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, const FEXCore
   adr(TMP1, &JITCodeHeaderLabel);
   str(TMP1, STATE, offsetof(FEXCore::Core::CPUState, InlineJITBlockHeader));
 
+  if (CheckTF) {
+    ARMEmitter::SingleUseForwardLabel l_TFUnset;
```
Sonicadvance1, line 716: Could the implementation details of this be put in some function like `EmitTFCheck` or something?

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -879,9 +889,11 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
     return (uintptr_t)CodePtr;
   }
 
-  // Insert to lookup cache
-  // Pages containing this block are added via AddBlockExecutableRange before each page gets accessed in the frontend
-  AddBlockMapping(Thread, GuestRIP, CodePtr);
+  if (!SkipCache) {
+    // Insert to lookup cache if TF is unset, otherwise the compiled block is a single instruction ephemeral block which shouldn't be cached
+    // Pages containing this block are added via AddBlockExecutableRange before each page gets accessed in the frontend
+    AddBlockMapping(Thread, GuestRIP, CodePtr);
```
neobrain, line 895: What does skipping this line imply in terms of ownership of compiled data? Will the instruction just permanently linger in the CodeBuffer unless the CodeBuffer gets freed for unrelated reasons?

That would probably be fine but as a personal reminder I'd need to do some chewing if this could cause any problems with code caching due to the 128 MiB limit for relative jumps.

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -810,17 +818,19 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   };
 }
 
-uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_t GuestRIP, uint64_t MaxInst) {
+uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_t GuestRIP, uint64_t MaxInst, bool SkipCache) {
```
neobrain, line 821: Does executing the BlockJITNaming branch make any sense with SkipCache? Similarly, does it make sense to include single-instruction compiles in profiles?

Truly the EFLAGS.TF path only needs to acquire the lock and call CompileCode directly. Semantically, a "SkipCache" parameter doesn't model this: The intent is never to "skip the cache" but to enable single-stepping, so implementing one via the other is a leaky abstraction.

Instead of hiding 90% of this function in an ad-hoc branch, I'd opt for a dedicated interface that's just called something like "PrepareSingleStep". Just inlining the required logic where needed would also work.

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -783,7 +787,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
 
   if (!IR) {
     // Generate IR + Meta Info
-    auto [IRCopy, TotalInstructions, TotalInstructionsLength, _StartAddr, _Length] = GenerateIR(Thread, GuestRIP, Config.GDBSymbols(), MaxInst);
+    auto [IRCopy, _TotalInstructions, TotalInstructionsLength, _StartAddr, _Length] = GenerateIR(Thread, GuestRIP, Config.GDBSymbols(), MaxInst);
```
neobrain, line 790: Unused rename. Let's avoid weird underscore prefixes in identifier names.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -484,11 +484,12 @@ static void IndirectBlockDelinker(FEXCore::Core::CpuStateFrame* Frame, FEXCore::
...
 static uint64_t Arm64JITCore_ExitFunctionLink(FEXCore::Core::CpuStateFrame* Frame, FEXCore::Context::ExitFunctionLinkData* Record) {
   auto Thread = Frame->Thread;
+  bool TFSet = Thread->CurrentFrame->State.flags[X86State::RFLAG_TF_RAW_LOC];
+  uintptr_t HostCode {};
   auto GuestRip = Record->GuestRIP;
 
-  auto HostCode = Thread->LookupCache->FindBlock(GuestRip);
-
-  if (!HostCode) {
+  if (TFSet || !(HostCode = Thread->LookupCache->FindBlock(GuestRip))) {
```
neobrain, line 491: I get that this is convenient when writing code, but for the sake of readability let's not combine control flow with short-circuiting behavior and variable initializations.

FEXCore/Source/Interface/Core/Core.cpp
bylaws: Yeah it'll linger until it's cleared for some reason, it's really not a problem though considering the frequency of single steps/inline SMC and the size of generated code. If it is an issue I think it's better to solve when we get to it

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: Fair enough :)

bylaws:
Should be good for review now

alyssarosenzweig:
This looks really solid. Great work, Billy!

alyssarosenzweig (approved):

Sonicadvance1 (approved):

PR #4162, opened 2024-11-13
Avoid warning on assertionless builds

pmatos:
This was causing unused variable warning due
to the variable only being used in an assertion.

### DIFF
--- a/FEXCore/Scripts/json_ir_generator.py
+++ b/FEXCore/Scripts/json_ir_generator.py
@@ -641,8 +641,7 @@ def print_ir_allocator_helpers():
     output_file.write("\t}\n\n")
 
     output_file.write("\tuint8_t GetOpElements(const OrderedNode *Op) const {\n")
-    output_file.write("\t\tauto HeaderOp = Op->Header.Value.GetNode(DualListData.DataBegin());\n")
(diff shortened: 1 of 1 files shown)


FEXCore/Scripts/json_ir_generator.py
```diff
@@ -641,7 +641,7 @@ def print_ir_allocator_helpers():
     output_file.write("\t}\n\n")
 
     output_file.write("\tuint8_t GetOpElements(const OrderedNode *Op) const {\n")
-    output_file.write("\t\tauto HeaderOp = Op->Header.Value.GetNode(DualListData.DataBegin());\n")
+    output_file.write("\t\t[[maybe_unused]] auto HeaderOp = Op->Header.Value.GetNode(DualListData.DataBegin());\n")
     output_file.write("\t\tLOGMAN_THROW_A_FMT(OpHasDest(Op), \"Op {} has no dest\\n\", GetName(HeaderOp->Op));\n")
```
Sonicadvance1, line 645: Could also delete the `HeaderOp` definition in the previous line and use `GetOpName(Op)` inside the assertion.

pmatos: Good point - done.

neobrain (approved):
Thanks!

Sonicadvance1 (approved):

PR #4163, opened 2024-11-17
OpcodeDispatcher: Mark CF as being possibly set after inverting

bylaws:
Required as if for example CF is non-inverted, code zeroing it doesn't need to mark it as being possibly set, however after inverting it will now be set. To be safe, always mark it as possibly set after inversion.

Fixes crashes with MaxInst > 2 in Uplay

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
@@ -1730,6 +1730,7 @@ private:
       }
 
       CFInverted ^= true;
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
I can only assume the ASM test is breaking since the register state isn't setup coming in to the test.

alyssarosenzweig:
this is "correct" but I'd like to see if there's a better fix possible, now that CI is up and I can test things again

alyssarosenzweig:
#4172

PR #4165, opened 2024-11-18
Support inline self modifying code

bylaws:
When an SMC trap happens: reconstruct the context before the SMC write then compile the write as a single instruction block to reduce it to regular SMC. SMC where the writing instruction is the instruction being patched will hit the signal handler at most twice: the 1st will trigger the write to be compiled as a single instuction block, the 2nd will detect inline SMC of a single instruction block and then just take the usual invalidate+reprotect+continue step, avoiding a potential infinite loop of recompilation.

Ontop of TF support


neobrain:
> Ontop of TF support

For reference, this is referring to #4161.

Sonicadvance1:
Needs a rebase.

alyssarosenzweig (approved):

Sonicadvance1 (approved):

PR #4166, opened 2024-11-20
Generate SVE for 80bit load/stores when possible

### LINKED ISSUE #4126: 80-bit x87 Loadstores can use SVE masked loadstores
Currently we need to split these in to a 64-bit and 16-bit loadstore operations.
With SVE and masking loadstores this can be changed in to one.
First step would probably be to convert current operations that need to do 80-bit loadstores to synthesize the required predicate and then to the loadstore.
Step after that would be to add predicate registers to our RA allocator so we can cache the 80-bit predicate across multiple loadstores (like fnsave, frstor would want).

Purely a performance optimization around x87 80-bit loadstores.

pmatos:
Fixes #4126


pmatos:
Probably not a big win in practice. A single 80-bit store required 2 stores (64 + 16). Now, we require three instructions: mov + whilelt + st1b. You need three 80-bit stores in a block to get to a draw instruction-wise. The next step would be not to assemble the predicate register every time, which I will do next, but we'll still require in practice at least three stores per block for it to "win" instruction-wise.

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
```diff
@@ -1536,6 +1536,17 @@ DEF_OP(StoreMem) {
...
       break;
     }
+    case IR::OpSize::f80Bit: {
+      LOGMAN_THROW_A_FMT(HostSupportsSVE128, "80-bit stores are only supported with SVE128 support");
+      const auto MemSrc = GenerateSVEMemOperand(OpSize, MemReg, Op->Offset, Op->OffsetType, Op->OffsetScale);
+      // We need to generate the predicate register to copy the 10-byte value
+      // FIXME: don't hardcode p5
+      ARMEmitter::PRegister preg = ARMEmitter::PReg::p5;
+      mov(TMP1, 10);
+      whilelt(ARMEmitter::SubRegSize::i8Bit, preg, ARMEmitter::XReg::zr, TMP1);
```
Sonicadvance1, line 1546: Can't generating the predicate use something like `ptrue(ARMEmitter::SubRegSize::i16Bit, preg, ARMEmitter::PredicatePattern::SVE_VL5` and then use `st1h<ARMEmitter::SubRegSize::i8Bit>(Src.Z(), preg, MemSrc);` ?
If it works then the predicate generation is 1 instruction (which is zero-cycle latency on A720), and then a single store.

pmatos: Maybe I am misunderstanding but VL5 is `0b00101`. The predicate needs to be setup with 0b1010 == 0xa, which wouldn't work.

pmatos: oh wait, but there is a pattern that would work, `  SVE_VL32 = 0b01010,`. OK - let me give it a go.

Sonicadvance1: Wrote a quick test, looks like my idea would work?
```cpp
#include <cstdio>
#include <cstdint>
#include <string.h>

__attribute__((naked))
void DoSVEStore(const void *in, void *out) {
	__asm volatile(R"(
ptrue p0.h, vl5;
ldr q0, [x0];
st1h z0.h, p0, [x1];
ret;
	)"
	::: "v0", "memory", "cc");
}

static const uint64_t Data[2] = {
	0x4142434445464748,
	0x5152535455565758,
};

static uint8_t DataOutSVE[16]{};
static uint8_t DataOutMemcpy[16]{};
void Do80BitStore(const void *in, void *out) {
	memcpy(out, in, 10);
}

int main() {
	DoSVEStore(Data, DataOutSVE);
	Do80BitStore(Data, DataOutMemcpy);

	fprintf(stderr, "      SVE | CPY  | Matches?\n"); 
	for (size_t i = 0; i < 16; ++i) {
	fprintf(stderr, "%02zd:  0x%02x | 0x%02x | %s\n", i, DataOutSVE[i], DataOutMemcpy[i], DataOutSVE[i] == DataOutMemcpy[i] ? "Yes" : "No");
	}
	return 0;
}
```

Output:
```bash
      SVE | CPY  | Matches?
00:  0x48 | 0x48 | Yes
01:  0x47 | 0x47 | Yes
02:  0x46 | 0x46 | Yes
03:  0x45 | 0x45 | Yes
04:  0x44 | 0x44 | Yes
05:  0x43 | 0x43 | Yes
06:  0x42 | 0x42 | Yes
07:  0x41 | 0x41 | Yes
08:  0x58 | 0x58 | Yes
09:  0x57 | 0x57 | Yes
10:  0x00 | 0x00 | Yes
11:  0x00 | 0x00 | Yes
12:  0x00 | 0x00 | Yes
13:  0x00 | 0x00 | Yes
14:  0x00 | 0x00 | Yes
15:  0x00 | 0x00 | Yes
```

Also tested putting the edge of the store ten bytes before the end of a page and confirmed it is only storing 10-bytes

pmatos: Thanks - I see in ptrue you're using p0.h rather than p0.b and then vl5. Maybe I haven't truly understood how these patterns work but your example helps.

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -57,6 +57,12 @@ namespace x64 {
     ARMEmitter::Reg::r24, ARMEmitter::Reg::r25, ARMEmitter::Reg::r30, ARMEmitter::Reg::r18,
   };
 
+  // p6 and p7 registers are used as temporaries no not added here for RA
+  // See PREF_TMP_16B and PREF_TMP_32B
+  // Also p8-p15 cannot be used can only encode p0-p7, so we're left with p0-p5.
+  constexpr std::array<ARMEmitter::PRegister, 6> PR = {ARMEmitter::PReg::p0, ARMEmitter::PReg::p1, ARMEmitter::PReg::p2,
```
Sonicadvance1, line 63: A thing of note, p0 and p1 are used as temporaries in the JIT for various things. So this actually leaves p2-p5 for allocation. Very few, but should be fine for our use cases still.

pmatos: Right - maybe this could be cleaned up and use RA? What do you think?

Sonicadvance1: They are internal implementation details, I think RAing the temps would be more annoying than it is worth.
Similar to the GPRs and FPRs having temporaries in the backend, I think we just need to label them `PTMP1` and `PTMP2`, to ensure the backend doesn't overstep and use RA allocated ones.

pmatos: Ok - make sense.

pmatos:
This is almost ready - the bit missing is really that the predicate register is not yet being cached. I thought RA would take care of this but... apparently there's still some magic missing.

pmatos:
@Sonicadvance1 pushed changes with the predicate register caching. It's more general than what we need now but if we generate predicate registers from patterns for other purposes we can use the cache as for that as well.

FEXCore/Scripts/json_ir_generator.py
```diff
@@ -221,7 +222,8 @@ def parse_ops(ops):
                     if (OpArg.IsSSA and
                         (OpArg.Type == "GPR" or
                         OpArg.Type == "GPRPair" or
-                        OpArg.Type == "FPR")):
+                        OpArg.Type == "FPR" or
+                        OpArg.Type == "PR")):
```
Sonicadvance1, line 226: Should this `PR` be `PRED`?

pmatos:
I will add the generation of SVE instructions to load/store x87 environment.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -4439,11 +4445,15 @@ void OpDispatchBuilder::StoreResult_WithOpSize(FEXCore::IR::RegisterClassType Cl
 
   if (OpSize == OpSize::f80Bit) {
     Ref MemStoreDst = LoadEffectiveAddress(A, true);
-
-    // For X87 extended doubles, split before storing
-    _StoreMem(FPRClass, OpSize::i64Bit, MemStoreDst, Src, Align);
-    auto Upper = _VExtractToGPR(OpSize::i128Bit, OpSize::i64Bit, Src, 1);
-    _StoreMem(GPRClass, OpSize::i16Bit, Upper, MemStoreDst, _Constant(8), std::min(Align, OpSize::i64Bit), MEM_OFFSET_SXTX, 1);
+    if (CTX->HostFeatures.SupportsSVE128 || CTX->HostFeatures.SupportsSVE256) {
+      auto PReg = _InitPredicate(OpSize::i16Bit, FEXCore::ToUnderlying(ARMEmitter::PredicatePattern::SVE_VL5));
```
Sonicadvance1, line 4449: Looks like this should switch over to `InitPredicateCached`

FEXCore/Scripts/json_ir_generator.py
pmatos: Indeed.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
pmatos: Well spotted. Thanks.

Sonicadvance1 (approved):
Nice!

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
```diff
@@ -1551,6 +1552,75 @@ DEF_OP(StoreMem) {
...
+  }
+}
+
+DEF_OP(LoadMemPredicate) {
+  const auto Op = IROp->C<IR::IROp_LoadMemPredicate>();
+  const auto Dst = GetVReg(Node);
+  const auto Predicate = GetPReg(Op->Mask.ID());
+  const auto MemReg = GetReg(Op->Addr.ID());
+
+  LOGMAN_THROW_A_FMT(HostSupportsSVE128 || HostSupportsSVE256, "StoreMemPredicate needs SVE support");
```
lioncash, line 1599:
```suggestion
  LOGMAN_THROW_A_FMT(HostSupportsSVE128 || HostSupportsSVE256, "LoadMemPredicate needs SVE support");
```

lioncash (commented):
Just one super minor nit, but looks good overall

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
pmatos: Thanks - fixed.

pmatos:
> I will add the generation of SVE instructions to load/store x87 environment.

Actually this unfortunately is not worth it due to the fact that predicated SVE ldst seem to be more restrictive than 128bit ldst memory addressing. This means that we lose in address computation if we try to use this in fnsave and fnrstor. 

I have some work locally to add offsets to Load/Store MemPredicate and then fold Constants into InlineConstants in ConstProp into these and generate these loads/stores from FNSAVE and FNRSTOR but it turns out not to be worth it so I think this work is now complete. 

Thanks for the reviews.

PR #4169, opened 2024-11-24
unittests/ASM: Fixes x87 80-bit loads on the edge of page boundaries.

Sonicadvance1:
These tests ensure that FEX's x87 80-bit loadstores don't read/write past the end of the page.

Noticed we were missing testing this edge case a few days ago.


Sonicadvance1:
Nice, this immediately uncovered a bug in our implementation on the load side. Which is typically a non-problem since long double alignment is 16-byte, but the instructions for loads themselves don't care about alignment.

pmatos:
LGTM. That's an interesting patch - so afaiu the problem was loading 128bit at page boundary. Well, splitting the 128bit into 64+16 bit loads creates a stronger case for 
#4166 . Thanks.

Sonicadvance1:
https://www.youtube.com/watch?v=fC_q9KPczAg

lioncash (approved):

PR #4171, opened 2024-11-26
Utils/StringUtil: Fixes ltrim and adds a unittest

Sonicadvance1:
ltrim had the issue that it would always consume the left-most character even if it wasn't whitespace. So `FEXLoader` would turn in to `EXLoader`, even without any whitespace in the string.

Adds a test to ensure this doesn't occur again.


unittests/APITests/StringUtils.cpp
```diff
@@ -0,0 +1,91 @@
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
+#include <Common/StringUtil.h>
+
+#define trim_test_type(trim_type, in, out) \
+  do {                                     \
+    fextl::string str = in;                \
+    FEX::StringUtil::trim_type(str);       \
+    CHECK(str == out);                     \
+  } while (0)
```
neobrain, line 10: Don't use macros when a function would do. As usual macros don't combine well with things, in this case they break proper test reporting.

neobrain:
```cpp
std::string get_Ntrimmed_string(std::string str) \
    Ntrimmed(str);
    return str;
}
```

Sonicadvance1: didit

pmatos:
Nice - lgtm. :) 
We don't likely use this in any critical place; otherwise, I would wonder how anything works.

pmatos:
Is there a reason we are not using `std::erase_if` from c++20, instead of erase+find_if?
https://en.cppreference.com/w/cpp/string/basic_string/erase2

Or the equivalent using `std::remove_if`, if we need to use iterators.

Sonicadvance1:
> Is there a reason we are not using `std::erase_if` from c++20, instead of erase+find_if? https://en.cppreference.com/w/cpp/string/basic_string/erase2
> 
> Or the equivalent using `std::remove_if`, if we need to use iterators.

Haven't really looked at it, but I assume the predication would have different behaviour. I'm obviously not a very big C++ buff, using things for convenience rather than knowing every detail of it.

pmatos:
> > Is there a reason we are not using `std::erase_if` from c++20, instead of erase+find_if? https://en.cppreference.com/w/cpp/string/basic_string/erase2
> > Or the equivalent using `std::remove_if`, if we need to use iterators.
> 
> Haven't really looked at it, but I assume the predication would have different behaviour. I'm obviously not a very big C++ buff, using things for convenience rather than knowing every detail of it.

Ah, actually I have tried it and it wouldn't work because erase_if would remove all white space instead of just trimming it, so it seems like your approach is the best unless you want to get into views and ranges etc. But it's simpler as you have it. Sorry for the noise.

unittests/APITests/StringUtils.cpp
```diff
@@ -0,0 +1,96 @@
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/generators/catch_generators.hpp>
```
neobrain, line 2: this include is probably not needed

neobrain (approved):

PR #4172, opened 2024-11-26
OpcodeDispatcher: drop PossiblySetNZCV

alyssarosenzweig:
this is a pain to track and, it turns out, buys us virtually nothing on flagm
    systems. rip it out.
    
    this fixes a bug with failing to set in all the right places.
    
    on non-flagm systems there's a slight instcountci impact, but that is mostly
    mitigated by the earlier patches in the series. so overall a wash there but
    worth it for making the codebase easier
    to reason about.

replaces #4163


Sonicadvance1 (approved):
Neato

PR #4179, opened 2024-11-28
FEXCore: Moves ThreadWaiting to the frontend 

Sonicadvance1:
Only in one location does the frontend actually care about this, the
backend doesn't care at all.


FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
```diff
@@ -585,3 +577,15 @@ fextl::unique_ptr<Alloc::HostAllocator> Create64BitAllocator() {
   return fextl::make_unique<OSAllocator_64Bit>();
 }
 } // namespace Alloc::OSAllocator
+
+namespace FEXCore::Allocator {
+#ifndef _WIN32
```
bylaws, line 582: This file is not built for windows, you can drop the ifdef here. Perhaps it makes sense to move this implementation to the linux side and pass in the allocator to SetupHooks? Or ideally go even further and move all of https://github.com/FEX-Emu/FEX/blob/ee592ba03ceabccb22d6f6f02cd2846377fadb0b/FEXCore/Source/Utils/Allocator.cpp#L35 out of FEXCore.

lioncash (approved):

PR #4181, opened 2024-11-29
FEXCore: Removes ExitReason from InternalThreadState

Sonicadvance1:
FEXCore hasn't been returning anything other than EXIT_SHUTDOWN for a long time, so this ended up just moving data around for no reason.

This isn't going to be used for further GdbServer work anyway, so just completely remove it.


lioncash (approved):
nice

Issue #4184, opened 2024-11-29
FEXServer auto-shutdown regression


neobrain:
FEXServer is supposed to exit automatically (with a 10s delay) if no more FEXInterpreter processes are running. Since d761fc44f4a694b205519e0f1fe36eea7de789f3, this doesn't happen anymore: FEXServer permanently stays open.

@asahilina Any idea what's happening?

PR #4185, opened 2024-11-29
FEXServer: Fix auto-shutdown regression

asahilina:
Fixes: #4184


neobrain (approved):
Thanks for the quick fix, glad to see it just needed this simple change!

Sonicadvance1 (approved):

PR #4188, opened 2024-12-01
FEXCore: Removes ExitHandler and RunUntilExit

Sonicadvance1:
Now that all the threading behaviour has been correctly separated/moved to the frontend, these functions serve no purpose.

- Instead of using RunUntilExit, all threads can use `ExecuteThread` directly, since there's nothing special about the primary thread now.
  - This also removes the public function definition of `ExecutionThread` since that was only used for threading logic.
- Instead of using an exit handler, just do the same cleanup after `ExecuteThread` has returned.
  - Just make gdbserver is cleaned up early if it exists since it may want to send some things to the connected gdb instance before threads are exited.


lioncash (commented):
needs a rebase

Sonicadvance1:
based

lioncash (approved):

PR #4190, opened 2024-12-01
FEXCore: Removes GetGPRSize and convert all uses to GetGPROpSize

Sonicadvance1:
Only a few remaining uses left, easy enough to convert. This finally switches the final few uses over.

NFC


lioncash:
needs a rebase

Sonicadvance1:
Rebased

lioncash (approved):

PR #4195, opened 2024-12-05
LinuxEmulation: Don't use clone3 for fork

Sonicadvance1:
clone3 was added in Linux 5.3 but our minimum spec is 5.0. Additionally the Raspberry Pi 5 kernel seems to complain about clone3 for some reason?

Just use clone instead of clone3


lioncash (approved):

neobrain:
This change broke cefsimple (`cef_binary_131.2.7+g9a14dc9+chromium-131.0.6778.86_linux64_client`), which now crashes early during startup.

PR #4202, opened 2024-12-09
Avoid dangling string_view in GetVar

pmatos:
Also makes sure we don't look up EnvMap twice.

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -224,21 +224,23 @@ void EnvLoader::Load() {
     }
   }
 
-  auto GetVar = [](EnvMapType& EnvMap, const std::string_view id) -> std::optional<std::string_view> {
-    if (EnvMap.find(id) != EnvMap.end()) {
-      return EnvMap.at(id);
+  auto GetVar = [](EnvMapType& EnvMap, const std::string_view id) -> std::optional<fextl::string> {
+    auto it = EnvMap.find(id);
+    if (it != EnvMap.end()) {
+      return fextl::string(it->second);
     }
 
     // If envp[] was empty, search using std::getenv()
-    const char* vs = std::getenv(id.data());
+    fextl::string id_str(id);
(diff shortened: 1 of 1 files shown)


neobrain:
Is this a bug fix? If yes, what's the bug?

pmatos:
> Is this a bug fix? If yes, what's the bug?

Yep - I noticed that sometimes the FEX_ROOTFS value will get mangled when returned as a `std::string_view` from `GetVar` from `EnvMap`. You can see this if you print the result of `GetVar`. Then I sorted out two other things: 
1. Removed `.at()` and just used the result of `.find()` instead. 
2. And to make sure that we don't have another dangling `std::string_view` from the result of `getenv` I changed the call and return in that case.

neobrain:
> I noticed that sometimes the FEX_ROOTFS value will get mangled when returned as a std::string_view from GetVar from EnvMap

Can you explain more where exactly the `string_view`s are referencing memory that is no longer valid? I'm probably just missing it because the the messy X-macros, but I don't see anything that's wrong here.

pmatos:
> > I noticed that sometimes the FEX_ROOTFS value will get mangled when returned as a std::string_view from GetVar from EnvMap
> 
> Can you explain more where exactly the `string_view`s are referencing memory that is no longer valid? I'm probably just missing it because of the messy X-macros, but I don't see anything that's wrong here.

Sorry - I should have been more explicit. 
Let me reproduce what I saw on Friday and get back to you.

pmatos:
> > I noticed that sometimes the FEX_ROOTFS value will get mangled when returned as a std::string_view from GetVar from EnvMap
> 
> Can you explain more where exactly the `string_view`s are referencing memory that is no longer valid? I'm probably just missing it because of the messy X-macros, but I don't see anything that's wrong here.

Thank you for bringing this up. I have been looking at the code and trying to repro what I saw locally on Friday but can't. Lets close this for now and if I get to repro it in the future, I will reopen. Just by looking at the code, everything does look ok.

PR #4205, opened 2024-12-10
GdbServer: Support 32-bit context definitions

Sonicadvance1:
Requires restructuring a couple of things, but nothing too crazy here.

First commit is from #4203

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
@@ -109,19 +109,20 @@ GdbServer::GdbServer(FEXCore::Context::Context* ctx, FEX::HLE::SignalDelegator* 
       }
 
       auto ThreadObject = FEX::HLE::ThreadManager::GetStateObjectFromFEXCoreThread(Thread);
-      ThreadObject->GdbInfo.Signal = Signal;
+      ThreadObject->GdbInfo = {};
+      ThreadObject->GdbInfo->Signal = Signal;
 
-      ThreadObject->GdbInfo.SignalPC = ArchHelpers::Context::GetPc(ucontext);
-      uint32_t IgnoreMask = Thread->CurrentFrame->InSyscallInfo & 0xFFFF;
-      this->SignalDelegation->SpillSRA(Thread, ucontext, IgnoreMask);
+      ThreadObject->GdbInfo->SignalPC = ArchHelpers::Context::GetPc(ucontext);
+      this->SignalDelegation->SpillSRA(Thread, ucontext, Thread->CurrentFrame->InSyscallInfo);
 
-      memcpy(ThreadObject->GdbInfo.GPRs, ArchHelpers::Context::GetArmGPRs(ucontext), sizeof(uint64_t) * 32);
-      ThreadObject->GdbInfo.PState = ArchHelpers::Context::GetArmPState(ucontext);
+      memcpy(ThreadObject->GdbInfo->GPRs, ArchHelpers::Context::GetArmGPRs(ucontext), sizeof(ThreadObject->GdbInfo->GPRs));
+      ThreadObject->GdbInfo->PState = ArchHelpers::Context::GetArmPState(ucontext);
 
       // Let GDB know that we have a signal
       this->Break(Thread, Signal);
 
       WaitForThreadWakeup();
+      ThreadObject->GdbInfo.reset();
 
       return true;
       },
@@ -154,6 +155,10 @@ static fextl::string hexstring(fextl::istringstream& ss, int delm) {
   }
 
   return ret;
+}
+
+static fextl::string appendHex(const char* data, size_t length) {
+  return fextl::fmt::format("{:#02x}", fmt::join(data, data + length, ""));
 }
 
 static fextl::string encodeHex(const unsigned char* data, size_t length) {
@@ -282,7 +287,6 @@ const FEX::HLE::ThreadStateObject* GdbServer::FindThreadByTID(uint32_t TID) {
   return Threads->at(0);
 }
 
-
 GdbServer::GDBContextDefinition GdbServer::GenerateContextDefinition(const FEX::HLE::ThreadStateObject* ThreadObject) {
   GDBContextDefinition GDB {};
   FEXCore::Core::CPUState state {};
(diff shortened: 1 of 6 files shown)


lioncash (approved):

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
```diff
@@ -128,23 +128,30 @@ class GdbServer {
     uint8_t Data[10];
   };
 
-  struct FEX_PACKED GDBContextDefinition {
-    uint64_t gregs[FEXCore::Core::CPUState::NUM_GPRS];
-    uint64_t rip;
+  template<size_t GPRCount, typename GPRType, size_t XMMCount, size_t XMMArray>
+  struct FEX_PACKED GDBContextDefinitionBase {
+    GPRType gregs[GPRCount];
+    GPRType rip;
```
neobrain, line 134: Making all this compile-time-based increases the code complexity here a lot. Have you considered using a runtime-based approach via `std::span` instead?

```cpp
bool is_64_bit;
GPRType gregs_storage[NUM_GPRS];
std::span gregs(gregs_storage, is_64_bit ? NUM_GPRS : NUM_GPRS / 2);
```

AFAICS we only need binary serialization in one place. We can do a little more work manually there, at the benefit of saving all the template ugliness this change currently adds (most notably `GenerateAndFetchContextDefinition.template operator()<GDBContextDefinition>`...).

Sonicadvance1: I don't see what you are intending here? We give the connected gdb instance a layout that is defined via `BuildTargetXML` and we need to ensure that the resulting binary that we encode and send over the pipe matches that exact struct layout, which differs between x86-64 and i386.

This is probably just me being daft since I don't quite understand what the intended change does.

Sonicadvance1: The struct definition is technically used in both `CommandReadRegisters` and `CommandReadReg` in different ways.
But I've removed any amount of being "clever".

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -84,6 +84,14 @@ struct ThreadStateObject : public FEXCore::Allocator::FEXAllocOperators {
   std::atomic_bool ThreadSleeping {false};
   FEXCore::InterruptableConditionVariable ThreadPaused;
 
+  // GDB signal information
+  struct GdbInfoStruct {
+    int Signal {};
+    uint64_t SignalPC {};
+    uint64_t GPRs[32];
+    uint64_t PState {};
+  } GdbInfo;
```
neobrain, line 93: When is the data here supposed to be valid / up to date? Does it make sense to turn it into an `std::optional` so we can assert against "overwritten while still in use" or "not initialized" errors?

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -107,6 +108,16 @@ GdbServer::GdbServer(FEXCore::Context::Context* ctx, FEX::HLE::SignalDelegator*
         return false;
       }
 
+      auto ThreadObject = FEX::HLE::ThreadManager::GetStateObjectFromFEXCoreThread(Thread);
+      ThreadObject->GdbInfo.Signal = Signal;
+
+      ThreadObject->GdbInfo.SignalPC = ArchHelpers::Context::GetPc(ucontext);
+      uint32_t IgnoreMask = Thread->CurrentFrame->InSyscallInfo & 0xFFFF;
```
neobrain, line 115: What does the magic 0xffff value do here?

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -145,6 +156,12 @@ static fextl::string hexstring(fextl::istringstream& ss, int delm) {
   return ret;
 }
 
+static void appendHex(fextl::ostringstream* ss, const unsigned char* data, size_t length) {
```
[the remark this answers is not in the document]
Sonicadvance1: But it isn't `Candidate function not viable: no known conversion from 'uint64_t *' (aka 'unsigned long *') to 'char *' for 1st argument  GdbServer.cpp:665:12: error: no matching function for call to 'appendHex'`

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
Sonicadvance1: Switched to an optional.

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: Nothing, I removed it.

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -145,6 +156,12 @@ static fextl::string hexstring(fextl::istringstream& ss, int delm) {
...
+  for (size_t i = 0; i < length; i++) {
+    *ss << std::setfill('0') << std::setw(2) << std::hex << int(data[i]);
+  }
+}
+
```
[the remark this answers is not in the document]
Sonicadvance1: Switched to fmt

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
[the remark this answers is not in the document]
neobrain: Right, but at least you could turn the mouthful that's `reinterpret_cast<const unsigned char*>` into `reinterpret_cast<const char*>`, couldn't you?

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -639,8 +661,41 @@ GdbServer::HandledPacketType GdbServer::CommandReadRegisters(const fextl::string
...
+  str += appendHex(reinterpret_cast<const unsigned char*>(&GDB.gs), sizeof(uint32_t));
+  for (auto& mm : GDB.mm) {
+    str += appendHex(reinterpret_cast<const unsigned char*>(&mm), sizeof(X80Float));
+  }
+
+  str += appendHex(reinterpret_cast<const unsigned char*>(&GDB.fctrl), sizeof(uint32_t));
+  str += appendHex(reinterpret_cast<const unsigned char*>(&GDB.fstat), sizeof(uint32_t));
+  for (auto& dummy : GDB.dummies) {
+    str += appendHex(reinterpret_cast<const unsigned char*>(&dummy), sizeof(uint32_t));
+  }
```
neobrain, line 690: Actually can't this just be `str += appendHex(reinterpret_cast<const unsigned char*>(GDB.dummies), sizeof(GDB.dummies));`?

Sonicadvance1: Switched.

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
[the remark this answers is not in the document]
Sonicadvance1: Sure, still looks like the same noise to me, but switched it.

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -281,9 +296,16 @@ GdbServer::GDBContextDefinition GdbServer::GenerateContextDefinition(const FEX::
...
-  memcpy(&GDB.rip, &state.rip, sizeof(GDB.rip));
+  if (ThreadObject->GdbInfo.has_value()) {
+    GDB.rip = CTX->RestoreRIPFromHostPC(ThreadObject->Thread, ThreadObject->GdbInfo->SignalPC);
 
-  GDB.eflags = CTX->ReconstructCompactedEFLAGS(ThreadObject->Thread, false, nullptr, 0);
+    const bool WasInJIT = CTX->IsAddressInCodeBuffer(ThreadObject->Thread, ThreadObject->GdbInfo->SignalPC);
+    GDB.eflags = CTX->ReconstructCompactedEFLAGS(ThreadObject->Thread, WasInJIT, const_cast<uint64_t*>(ThreadObject->GdbInfo->GPRs),
+                                                 ThreadObject->GdbInfo->PState);
+  } else {
+    GDB.rip = ThreadObject->Thread->CurrentFrame->State.rip;
```
neobrain, line 306: Should this happen in normal operation?

Sonicadvance1: Yes, inspecting thread state for a different thread that hasn't faulted.

neobrain (approved):

PR #4206, opened 2024-12-10
Windows: Track RWX regions in mapped images

bylaws:
As section permissions are set on the unix side we don't get a
protection callback for them, workaround this by iterating over
the sections of all executables after mapping and tracking the RWX
ones.


Sonicadvance1 (approved):
Nice!

PR #4207, opened 2024-12-10
unittests: Add a test for constprop size bugs

bylaws:
Note that and and neg don't hit issues as they don't directly use their underlying IR ops (which are still broken tho)


alyssarosenzweig:
cherry-picked to #4215, thanks for the handy test case!

PR #4208, opened 2024-12-10
GdbServer: Rework sleeping

Sonicadvance1:
This is enough to get GdbServer working for my simple test case again.

Some things of note:
- There is now a callback which signals to GdbServer when a thread is
  created.
- Thread sleeping is now a single uint32_t futex word for knowing when a
  thread is sleeping and if we need to signal it to wake up.
- GdbServer thread sleeping versus ThreadManager thread sleeping
  distinction has been removed. GdbServer now uses ThreadManager to put
  a thread to sleep.

This still only gets the attach at startup scheme of gdbserver working.
I haven't yet dived back in to getting arbitrary attach working.
Primarily the work necessary here is actually getting thread creation at
the start to stop and wait for gdbserver to attach. Then actually
sleeping correctly while gdbserver is communicating with gdb.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
@@ -601,7 +601,7 @@ GdbServer::HandledPacketType GdbServer::ThreadAction(char action, uint32_t tid) 
   switch (action) {
   case 'c': {
     {
-      std::lock_guard lk(*SyscallHandler->TM.GetThreadsCreationMutex());
+      std::lock_guard lk(SyscallHandler->TM.GetThreadsCreationMutex());
       auto Threads = SyscallHandler->TM.GetThreads();
       for (auto& Thread : *Threads) {
         Thread->ThreadSleeping.NotifyOne();
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
@@ -62,13 +62,12 @@ $end_info$
 namespace FEX {
 
 #ifndef _WIN32
-void GdbServer::Break(FEXCore::Core::InternalThreadState* Thread, int signal) {
+void GdbServer::Break(FEX::HLE::ThreadStateObject* ThreadObject, int signal) {
   std::lock_guard lk(sendMutex);
   if (!CommsStream) {
     return;
   }
 
-  auto ThreadObject = FEX::HLE::ThreadManager::GetStateObjectFromFEXCoreThread(Thread);
   // Current debugging thread switches to the thread that is breaking.
   CurrentDebuggingThread = ThreadObject->ThreadInfo.TID.load();
 
@@ -119,7 +118,7 @@ GdbServer::GdbServer(FEXCore::Context::Context* ctx, FEX::HLE::SignalDelegator* 
       ThreadObject->GdbInfo->PState = ArchHelpers::Context::GetArmPState(ucontext);
 
       // Let GDB know that we have a signal
-      this->Break(Thread, Signal);
+      this->Break(ThreadObject, Signal);
 
       WaitForThreadWakeup();
       ThreadObject->GdbInfo.reset();
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
@@ -10,6 +10,7 @@ $end_info$
 #include "GdbServer/Info.h"
 
 #include "LinuxSyscalls/NetStream.h"
+#include "LinuxSyscalls/SignalDelegator.h"
 
 #include <cstdlib>
 #include <cstdio>
@@ -75,11 +76,6 @@ void GdbServer::Break(FEX::HLE::ThreadStateObject* ThreadObject, int signal) {
   SendPacket(*CommsStream, str);
 }
 
-void GdbServer::WaitForThreadWakeup() {
-  // Wait for gdbserver to tell us to wake up
-  ThreadBreakEvent.Wait();
-}
-
 GdbServer::~GdbServer() {
   CloseListenSocket();
   CoreShuttingDown = true;
@@ -120,7 +116,7 @@ GdbServer::GdbServer(FEXCore::Context::Context* ctx, FEX::HLE::SignalDelegator* 
       // Let GDB know that we have a signal
       this->Break(ThreadObject, Signal);
 
-      WaitForThreadWakeup();
+      this->SyscallHandler->TM.SleepThread(this->CTX, ThreadObject);
       ThreadObject->GdbInfo.reset();
 
       return true;
@@ -604,8 +600,14 @@ GdbServer::HandledPacketType GdbServer::handleProgramOffsets() {
 GdbServer::HandledPacketType GdbServer::ThreadAction(char action, uint32_t tid) {
   switch (action) {
   case 'c': {
+    {
+      std::lock_guard lk(*SyscallHandler->TM.GetThreadsCreationMutex());
+      auto Threads = SyscallHandler->TM.GetThreads();
+      for (auto& Thread : *Threads) {
+        Thread->ThreadSleeping.NotifyOne();
+      }
+    }
     SyscallHandler->TM.Run();
-    ThreadBreakEvent.NotifyAll();
     SyscallHandler->TM.WaitForThreadsToRun();
     return {"", HandledPacketType::TYPE_ONLYACK};
   }
@@ -1448,6 +1450,13 @@ void GdbServer::StartThread() {
   uint64_t OldMask = FEXCore::Threads::SetSignalMask(~0ULL);
   gdbServerThread = FEXCore::Threads::Thread::Create(ThreadHandler, this);
   FEXCore::Threads::SetSignalMask(OldMask);
+}
+
+void GdbServer::OnThreadCreated(FEX::HLE::ThreadStateObject* ThreadObject) {
+  if (SyscallHandler->TM.GetThreadCount() == 1) {
+    // Sleep the first thread created. This is because FEX only supports attaching at startup currently.
+    SyscallHandler->TM.SleepThread(CTX, ThreadObject);
+  }
 }
 
 void GdbServer::OpenListenSocket() {
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
@@ -36,7 +36,7 @@ public:
   }
 
 private:
-  void Break(FEXCore::Core::InternalThreadState* Thread, int signal);
+  void Break(FEX::HLE::ThreadStateObject* ThreadObject, int signal);
 
   void OpenListenSocket();
   void CloseListenSocket();
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
@@ -35,6 +35,8 @@ public:
     LibraryMapChanged = true;
   }
 
+  void OnThreadCreated(FEX::HLE::ThreadStateObject* ThreadObject);
+
 private:
   void Break(FEX::HLE::ThreadStateObject* ThreadObject, int signal);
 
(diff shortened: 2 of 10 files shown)


Sonicadvance1:
Removed the dependency on #4208 for this PR since this is independent.

pmatos:
> Removed the dependency on #4208 for this PR since this is independent.

You probably meant to reference some other PR here, since #4208 is this PR.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,80 @@ namespace FEX::HLE {
...
+  std::atomic<uint32_t> Mutex {};
+  constexpr static uint32_t HAS_NO_WAITER = 0;
+  constexpr static uint32_t HAS_WAITER = 1;
+  constexpr static uint32_t HAS_SIGNALED = 2;
+
+  void DoNotify(int Waiters) {
+    uint32_t Expected = HAS_WAITER;
+    const uint32_t Desired = HAS_SIGNALED;
+
+    // If the mutex is in a waiting state and we have CAS exchanged it to HAS_SIGNALED, then futex.
```
neobrain, line 85: futex isn't a verb.

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -604,8 +600,11 @@ GdbServer::HandledPacketType GdbServer::handleProgramOffsets() {
 GdbServer::HandledPacketType GdbServer::ThreadAction(char action, uint32_t tid) {
   switch (action) {
   case 'c': {
+    auto Threads = SyscallHandler->TM.GetThreads();
+    for (auto& Thread : *Threads) {
+      Thread->ThreadSleeping.NotifyOne();
+    }
```
neobrain, line 606: Is it safe to iterate over all threads like this? Can the threads list change while this code is running?

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
```diff
@@ -1451,6 +1449,13 @@ void GdbServer::StartThread() {
   FEXCore::Threads::SetSignalMask(OldMask);
 }
 
+void GdbServer::CreateThreadCallback(FEX::HLE::ThreadStateObject* ThreadObject) {
+  if (SyscallHandler->TM.GetThreadCount() == 1) {
+    // Sleep the first thread created. This is because FEX only currently supports attaching at startup.
```
neobrain, line 1454: Wrong word order. I think this is what you mean?

```suggestion
    // Sleep the first thread created. This is because FEX only supports attaching at startup currently.
```

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.h
```diff
@@ -35,8 +35,10 @@ class GdbServer {
     LibraryMapChanged = true;
   }
 
+  void CreateThreadCallback(FEX::HLE::ThreadStateObject* ThreadObject);
```
neobrain, line 38:
```suggestion
  void OnThreadCreated(FEX::HLE::ThreadStateObject* ThreadObject);
```

Avoids confusion as to what is being created.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -71,7 +84,10 @@ void ThreadManager::NotifyPause() {
   // Tell all the threads that they should pause
   std::lock_guard lk(ThreadCreationMutex);
   for (auto& Thread : Threads) {
-    SignalDelegation->SignalThread(Thread->Thread, SignalEvent::Pause);
+    if (Thread->ThreadSleeping.HasWaiter() == false) {
```
neobrain, line 87:
```suggestion
    if (!Thread->ThreadSleeping.HasWaiter()) {
```

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,80 @@ namespace FEX::HLE {
 class SyscallHandler;
 class SignalDelegator;
 
+// This is similar to FEXCore::InterruptableConditionVariable with the exception that notifying only occurs if there is a waiter and can check if
```
neobrain, line 25: Comparison to other synchronization primitives should be a side note, not the main introduction to a class. The docstring should describe the purpose/use of the class directly rather than approximating it indirectly (as it currently does).

Source/Tools/LinuxEmulation/LinuxSyscalls/GdbServer.cpp
Sonicadvance1: Technically unsafe. Grabbed the thread creation mutex while doing this.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,79 @@ namespace FEX::HLE {
 class SyscallHandler;
 class SignalDelegator;
 
+// A latch that can be inspected to see if there is a waiter currently active.
+// This allows us to remove the race condition between a thread trying to go asleep and something else telling it to go to sleep or wake up.
```
neobrain, line 26: This race condition really comes out of nowhere, making the entire line very vague. I think by trying to be very specific (i.e. mentioning the concrete example), it actually gets in the way of conveying the point.

What you're really trying to say here is that entering waiting state and signaling aren't required to happen in any specific order:

```
In contrast to a condition variable, an InspectableLatch may be signaled *before* being waited on without change of effect.
```

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,79 @@ namespace FEX::HLE {
...
+// Only one thread can ever wait on a latch, while another thread signals it.
+class InspectableLatch final {
+public:
+  bool Wait(struct timespec* Timeout = nullptr) {
+    while (true) {
+      uint32_t Expected = HAS_NO_WAITER;
+      const uint32_t Desired = HAS_WAITER;
+
+      if (Mutex.compare_exchange_strong(Expected, Desired)) {
+        // We have latched, now futex.
```
neobrain, line 37: Another instance of using futex as a verb. Code comments should describe semantics instead of just repeating the code literally.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,79 @@ namespace FEX::HLE {
...
+                               0);      // val3
+
+        if (Timeout && Result == -1 && errno == ETIMEDOUT) {
+          return false;
+        }
+      } else if (Expected == HAS_SIGNALED) {
+        // Reset the latch once signaled
+        Mutex.store(HAS_NO_WAITER);
+        return true;
+      }
```
neobrain, line 53: Should there be an assert against `Expected != HAS_NO_WAITER && Expected != HAS_SIGNALED` here to cheaply detect if multiple waiters are active?

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -169,12 +243,22 @@ class ThreadManager final {
     }
   }
 
+  FEXCore::ForkableUniqueMutex* GetThreadsCreationMutex() {
```
neobrain, line 246: This is never null, so it should return a reference instead.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,79 @@ namespace FEX::HLE {
...
+  bool WaitFor(const std::chrono::duration<Rep, Period>& time) {
+    struct timespec Timeout {};
+    auto SecondsDuration = std::chrono::duration_cast<std::chrono::seconds>(time);
+    Timeout.tv_sec = SecondsDuration.count();
+    Timeout.tv_nsec = std::chrono::duration_cast<std::chrono::nanoseconds>(time - SecondsDuration).count();
+    return Wait(&Timeout);
+  }
+
+  void NotifyOne() {
+    DoNotify(1);
```
neobrain, line 67: Similarly, there should probably be an assert here to ensure the latch isn't already in signaled state.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
Sonicadvance1: Sure, updated comment.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,83 @@ namespace FEX::HLE {
 class SyscallHandler;
 class SignalDelegator;
 
+// This is basically a std::condition_variable with only one waiter and can be inspected if there is a waiter.
```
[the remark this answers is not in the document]
neobrain: Looking over the other changes (resolved but unchanged), this seems unintended. Did you accidentally push an old version of the branch?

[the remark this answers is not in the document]
Sonicadvance1: Used your comment.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,83 @@ namespace FEX::HLE {
...
+class InspectableLatch final {
+public:
+  bool Wait(struct timespec* Timeout = nullptr) {
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
```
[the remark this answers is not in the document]
Sonicadvance1: Removed the assertion check wrapper.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,82 @@ namespace FEX::HLE {
 class SyscallHandler;
 class SignalDelegator;
 
+// A latch that can be inspected to see if there is a waiter currently active.
+// In contrast to std::condition_variable, an InspectableLatch may be signaled *before* being waited on without change of effect.
+class InspectableLatch final {
+public:
+  bool Wait(struct timespec* Timeout = nullptr) {
+    LOGMAN_THROW_A_FMT(Mutex.load() != HAS_SIGNALED, "Stale signal mutex!");
```
neobrain, line 30: This line breaks pre-signaled latches, doesn't it?

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -22,6 +22,82 @@ namespace FEX::HLE {
...
+  bool WaitFor(const std::chrono::duration<Rep, Period>& time) {
+    struct timespec Timeout {};
+    auto SecondsDuration = std::chrono::duration_cast<std::chrono::seconds>(time);
+    Timeout.tv_sec = SecondsDuration.count();
+    Timeout.tv_nsec = std::chrono::duration_cast<std::chrono::nanoseconds>(time - SecondsDuration).count();
+    return Wait(&Timeout);
+  }
+
+  void NotifyOne() {
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
```
neobrain, line 67: Another unnecessary preprocessor guard.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
Sonicadvance1: It specifically doesn't support pre-signaled latches. That's one of the two reasons why I implemented this.

neobrain: Why did you copy my docstring proposal if pre-signaling is not supported? The two are in direct contradiction as-is.

> That's one of the two reasons why I implemented this.

You implemented InspectableLatch in order to be able to... not support something? Why?

(FTR I'd suggest not live-iterating the docstring any further until we find a suitable alternative, since I feel like I'm understanding increasingly less what this class is trying to achieve but GitHub makes going back to your older revisions a pain.)

PR #4209, opened 2024-12-11
InstCountCI: Implement support for TSO and LRCPC and add hot block that could be optimized

Sonicadvance1:
Small immediate offsets that can fit in loads is getting missed for some reason.
Add this to instcountCI.

This block was showing up hot in libvulkan_freedreno


alyssarosenzweig (approved):
thanks, will poke at this locally.

PR #4215, opened 2024-12-13
ConstProp: fix 32-bit masking behaviour

alyssarosenzweig:
if we want to replace a node with one of its sources, we need to zero extend if
    the source is 64-bit and the destination is 32-bit.


Sonicadvance1 (approved):
Nice nice

PR #4216, opened 2024-12-13
Optimize SIMM9 with TSO

alyssarosenzweig:
per @Sonicadvance1's nerdsnipe

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -4176,10 +4176,18 @@ AddressMode OpDispatchBuilder::SelectAddressMode(AddressMode A, bool AtomicTSO, 
   // addresses are reserved and therefore wrap around is invalid.
   //
   // TODO: Also handle GPR TSO if we can guarantee the constant inlines.
-  if (SupportsRegIndex && (A.Base || A.Segment) && A.Offset) {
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Looks like the CI failures are not spurious but actually a result of the change.

FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
```diff
@@ -4176,35 +4176,41 @@ AddressMode OpDispatchBuilder::SelectAddressMode(AddressMode A, bool AtomicTSO,
...
-
-        return {
-          .Base = LoadEffectiveAddress(B, true /* AddSegmentBase */, false),
-          .Index = _Constant(A.Offset),
-          .IndexType = MEM_OFFSET_SXTX,
-          .IndexScale = 1,
-        };
-      }
+  if ((A.Base || A.Segment) && A.Offset) {
+    const bool SupportsSmallConst = A.AddrSize == OpSize::i32Bit && GPRSize == OpSize::i32Bit;
```
Sonicadvance1, line 4180: I think this needs to be `const bool SupportsSmallConst = A.AddrSize ==  GPRSize;` Otherwise we lose small const support for LRCPC2 on 64-bit processes with 64-bit addresses?

alyssarosenzweig: I think the small const logic is wacked as it is on 64-bit. I'll investigate this a bit more.

PR #4218, opened 2024-12-14
Utils/FileLoading: Fix LoadFileImpl

Sonicadvance1:
It is not an error that pread returns /less/ than what was requested. In fact it's very common for the Linux kernel to return less than the data requested from procfs.

procfs keeps coming back to bite this function, previously it was fstat returning size of 0 which it hit. Now it only feeds data as much as it wants per loop. In particular /proc/self/maps would only read ~3k bytes on my system, but not be complete.

To fully fix the issue, always make sure to keep reading until there is either an error OR zero is reached!


lioncash (approved):
oof, nice catch!

PR #4221, opened 2024-12-17
ThreadManager: Add some sanity asserts

Sonicadvance1:
These couple of functions have some footguns that I'm encountering while rewriting gdbserver. Ensure that assertion builds capture the problems


neobrain (approved):

pmatos:
LGTM. Nice - I personally thing this kind of checks are very useful in an Debug/Assertions environment, so the more checks the better. :)

PR #4224, opened 2024-12-19
FHU: Fix shebang line parsing

asahilina:
This function is used to parse shebang lines, and Linux does not parse shebang lines like shell command lines. Instead, it only splits it into an interpreter and an optional single argument, with leading/trailing whitespace stripped.

### DIFF
--- a/FEXHeaderUtils/FEXHeaderUtils/StringArgumentParser.h
+++ b/FEXHeaderUtils/FEXHeaderUtils/StringArgumentParser.h
@@ -8,31 +8,38 @@
 namespace FHU {
 
 /**
- * @brief Parses a string of arguments, returning a vector of string_views.
+ * @brief Parses a shebang string, returning a vector of string_views.
  *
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
Interesting, I guess I looked at the incorrect behaviour when I was implementing this. Good to know

FEXHeaderUtils/FEXHeaderUtils/StringArgumentParser.h
```diff
@@ -8,32 +8,39 @@
 namespace FHU {
 
 /**
- * @brief Parses a string of arguments, returning a vector of string_views.
+ * @brief Parses a shebang string, returning a vector of string_views.
```
neobrain, line 11: Sadly this heavy change in semantics makes ParseArgumentsFromString a misnomer. It's no longer a generic argument parser, so the function/header file/unit test should be renamed.

PR #4225, opened 2024-12-19
Support a merged RootFS (and a bunch of related fixes)

asahilina:
This is a new RootFS approach where the entire filesystem hierarchy is bind-mounted under the RootFS directory, with the RootFS itself overlaid. This, in principle, lets FEX avoid all the RootFS merge logic and lets the kernel do it, which should be both faster and more correct. In particular, symlinks between user directories and the RootFS work, and listing a directory correctly returns the merged entries from the RootFS and the host FS, instead of only the RootFS ones.

This PR includes fixes to make a merged-RootFS configuration work (as well as some misc adjacent issues), but does not add an explicit configuration option to disable the filesystem overlay logic in this scenario. That may come later, but it should only be an optimization at that point and may or may not be worth it.

The merged RootFS scenario stress-tests the FEX path handling, since in principle everything is opened through the RootFS and therefore there are more opportunities for the RootFS path to leak to the guest or end up in the wrong place.

Unlike #3831, this PR should *improve* performance, since it removes system calls in a hot path. `open()`/`openat()`/`openat2()` go from (in the common case):

  - `readlink()` * O(directory components) (from `realpath()` in emufd)
  - `fstatat()` (from `GetEmulatedFDPath()`, and possibly more if it's a symlink)
  - The `openat2()` itself

To:
  - One `openat2()` for the RootFS
  - If not found in RootFS, one `openat2()` for the host
  - One single `readlink()` to check for emu files

Tested with Steam and some games under Proton and Portal 2, both with and without the merged-RootFS setup.

When used with a [merged RootFS](https://github.com/AsahiLinux/muvm/pull/130), this finally fixes Wine. This supersedes #3831 *only* in the merged-rootfs case, so the question remains whether the non-merged resolution should be fixed to handle the problem or not...

Note that the fixes here are not comprehensive (proper symlink resolution behavior is limited to `open()` family syscalls, not any others), but it's enough to fix Wine and shouldn't regress any cases.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -333,6 +333,8 @@ FileManager::FileManager(FEXCore::Context::Context* ctx)
     ProcFSDev = Buffer.st_dev;
   }
 
+  uint32_t KernelVersion = FEX::HLE::SyscallHandler::CalculateHostKernelVersion();
+  HasOpenat2 = KernelVersion >= FEX::HLE::SyscallHandler::KernelVersion(5, 8, 0);
   UpdatePID(::getpid());
 }
 
@@ -646,15 +648,17 @@ uint64_t FileManager::Open(const char* pathname, int flags, uint32_t mode) {
 
   if (!ShouldSkipOpenInEmu(flags)) {
     FDPathTmpData TmpFilename;
-    auto Path = GetEmulatedFDPath(AT_FDCWD, SelfPath, false, TmpFilename);
+    auto Path = GetEmulatedFDPath(AT_FDCWD, SelfPath, !HasOpenat2, TmpFilename);
     if (Path.first != -1) {
       FEX::HLE::open_how how = {
         .flags = (uint64_t)flags,
         .mode = (flags & (O_CREAT | O_TMPFILE)) ? mode & 07777 : 0, // openat2() is stricter about this
         .resolve = (Path.first == AT_FDCWD) ? 0u : RESOLVE_IN_ROOT, // AT_FDCWD means it's a thunk and not via RootFS
       };
-      fd = ::syscall(SYSCALL_DEF(openat2), Path.first, Path.second, &how, sizeof(how));
-      if (fd == -1 && errno == EXDEV) {
+      if (HasOpenat2) {
+        fd = ::syscall(SYSCALL_DEF(openat2), Path.first, Path.second, &how, sizeof(how));
+      }
+      if (fd == -1 && (!HasOpenat2 || errno == EXDEV)) {
         // This means a magic symlink (/proc/foo) was involved. In this case we
         // just punt and do the access without RESOLVE_IN_ROOT.
         fd = ::syscall(SYSCALL_DEF(openat), Path.first, Path.second, flags, mode);
@@ -899,15 +903,17 @@ uint64_t FileManager::Openat([[maybe_unused]] int dirfs, const char* pathname, i
 
   if (!ShouldSkipOpenInEmu(flags)) {
     FDPathTmpData TmpFilename;
-    auto Path = GetEmulatedFDPath(dirfs, SelfPath, false, TmpFilename);
+    auto Path = GetEmulatedFDPath(dirfs, SelfPath, !HasOpenat2, TmpFilename);
     if (Path.first != -1) {
       FEX::HLE::open_how how = {
         .flags = (uint64_t)flags,
         .mode = (flags & (O_CREAT | O_TMPFILE)) ? mode & 07777 : 0, // openat2() is stricter about this,
         .resolve = (Path.first == AT_FDCWD) ? 0u : RESOLVE_IN_ROOT, // AT_FDCWD means it's a thunk and not via RootFS
       };
-      fd = ::syscall(SYSCALL_DEF(openat2), Path.first, Path.second, &how, sizeof(how));
-      if (fd == -1 && errno == EXDEV) {
+      if (HasOpenat2) {
+        fd = ::syscall(SYSCALL_DEF(openat2), Path.first, Path.second, &how, sizeof(how));
+      }
+      if (fd == -1 && (!HasOpenat2 || errno == EXDEV)) {
         // This means a magic symlink (/proc/foo) was involved. In this case we
         // just punt and do the access without RESOLVE_IN_ROOT.
         fd = ::syscall(SYSCALL_DEF(openat), Path.first, Path.second, flags, mode);
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -787,7 +787,7 @@ uint64_t FileManager::Readlink(const char* pathname, char* buf, size_t bufsiz) {
     if (Result == -1 && errno == EINVAL) {
       // This means that the file wasn't a symlink
       // This is expected behaviour
-      return -errno;
+      return -1;
     }
   }
   if (Result == -1) {
@@ -865,7 +865,7 @@ uint64_t FileManager::Readlinkat(int dirfd, const char* pathname, char* buf, siz
     if (Result == -1 && errno == EINVAL) {
       // This means that the file wasn't a symlink
       // This is expected behaviour
-      return -errno;
+      return -1;
     }
   }
 
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -339,6 +339,85 @@ FileManager::~FileManager() {
   close(RootFSFD);
 }
 
+size_t FileManager::GetRootFSPrefixLen(const char* pathname, size_t len, bool AliasedOnly) {
+  if (len < 2 ||            // If no pathname or root
+      pathname[0] != '/') { // If we are getting root
+    return 0;
+  }
+
+  const auto& RootFSPath = LDPath();
+  if (RootFSPath.empty()) { // If RootFS doesn't exist
+    return 0;
+  }
+
+  auto RootFSLen = RootFSPath.length();
+  if (RootFSPath.ends_with("/")) {
+    RootFSLen -= 1;
+  }
+
+  if (RootFSLen > len) {
+    return 0;
+  }
+
+  if (memcmp(pathname, RootFSPath.c_str(), RootFSLen) || (len > RootFSLen && pathname[RootFSLen] != '/')) {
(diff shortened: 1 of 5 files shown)


neobrain:
Awesome, thanks for having continued to look into this!

> When used with a https://github.com/AsahiLinux/muvm/pull/130, this finally fixes Wine

Just curious, can an equivalent RootFS setup be achieved with vanilla OverlayFS or are there muvm-specific bits to this approach? My understanding was that normal OverlayFS has restrictions that make it infeasible for FEX RootFS management in general, though I don't remember if this was just an issue of lack of multiarch/lib in most Linux distributions or something more fundamental.

asahilina:
There's nothing special about muvm, the same setup can be achieved on any system. The main limitation is that overlayfs is fundamentally read-only the way we intend to use it here, so you have to partition the file hierarchy into subtrees that are overlaid (and therefore cannot be written through) and subtrees that are bind-mounted straight through. overlayfs also can't handle sub-mounts in its layers.

`muvm` achieves this by populating the RootFS directory on a top-level-directory basis. In the end, since our RootFS only has /etc and /usr, those two end up as OverlayFS mounts and everything else is a bind mount.

In addition, overlayfs nominally requires that the layers not mutate out from under it, so technically making any changes to the overlaid subtrees of the real host filesystem is also "UB" (constrained, won't crash your kernel). Realistically I don't think this will cause any major issues other than "maybe if you add files to the host root you won't see them in FEX until a remount", which is probably fine.

In our setup, as long as /etc and /usr don't have any submounts (they don't in any default Fedora installs), and as long as they don't mutate while `muvm` is running (they generally don't, certainly nothing that FEX cares about). If anything I would expect `/etc` to be the first place to become problematic, but that could probably be fixed by moving the overlay behavior to one level deeper there.

For reference, this is the mount tree that `muvm` ends up with:

```
tmpfs on /run/fex-emu type tmpfs (rw,nosuid,noexec,relatime)
/dev/vda on /run/fex-emu/vda type erofs (ro,relatime,user_xattr,acl,cache_strategy=readaround)
/dev/vdb on /run/fex-emu/vdb type erofs (ro,relatime,user_xattr,acl,cache_strategy=readaround)
/dev/vdc on /run/fex-emu/vdc type erofs (ro,relatime,user_xattr,acl,cache_strategy=readaround)
/dev/vda on /run/fex-emu/rootfs/.profile type erofs (ro,relatime,user_xattr,acl,cache_strategy=readaround)
/dev/root on /run/fex-emu/rootfs/.viminfo type virtiofs (rw,relatime)
/dev/root on /run/fex-emu/rootfs/media type virtiofs (rw,relatime)
/dev/root on /run/fex-emu/rootfs/mnt type virtiofs (rw,relatime)
/dev/root on /run/fex-emu/rootfs/opt type virtiofs (rw,relatime)
overlay on /run/fex-emu/rootfs/etc type overlay (ro,relatime,lowerdir=/run/fex-emu/vdc/etc:/run/fex-emu/vdb/etc:/run/fex-emu/vda/etc:/run/muvm-host/etc,redirect_dir=nofollow,userxattr)
none on /run/fex-emu/rootfs/tmp type virtiofs (rw,relatime)
tmpfs on /run/fex-emu/rootfs/tmp/.X11-unix type tmpfs (rw,nosuid,noexec,relatime)
overlay on /run/fex-emu/rootfs/usr type overlay (ro,relatime,lowerdir=/run/fex-emu/vdc/usr:/run/fex-emu/vdb/usr:/run/fex-emu/vda/usr:/usr,redirect_dir=nofollow,userxattr)
none on /boot type virtiofs (rw,relatime)
none on /run/fex-emu/rootfs/boot type virtiofs (rw,relatime)
sysfs on /run/fex-emu/rootfs/sys type sysfs (rw,nosuid,nodev,noexec,relatime)
cgroup2 on /run/fex-emu/rootfs/sys/fs/cgroup type cgroup2 (rw,nosuid,nodev,noexec,relatime)
/dev/root on /run/fex-emu/rootfs/afs type virtiofs (rw,relatime)
/dev/root on /run/fex-emu/rootfs/root type virtiofs (rw,relatime)
/dev/root on /run/fex-emu/rootfs/srv type virtiofs (rw,relatime)
none on /run/fex-emu/rootfs/home type virtiofs (rw,relatime)
devtmpfs on /run/fex-emu/rootfs/dev type devtmpfs (rw,relatime,size=25559024k,nr_inodes=6389756,mode=755)
devpts on /run/fex-emu/rootfs/dev/pts type devpts (rw,nosuid,noexec,relatime,mode=600,ptmxmode=000)
devshm on /run/fex-emu/rootfs/dev/shm type virtiofs (rw,nosuid,noexec,relatime,dax=always)
tmpfs on /run/fex-emu/rootfs/run type tmpfs (rw,nosuid,noexec,relatime)
/dev/root on /run/fex-emu/rootfs/run/muvm-host type virtiofs (rw,relatime)
proc on /run/fex-emu/rootfs/proc type proc (rw,nosuid,nodev,noexec,relatime)
binfmt_misc on /run/fex-emu/rootfs/proc/sys/fs/binfmt_misc type binfmt_misc (rw,nosuid,noexec,relatime)
/dev/root on /run/fex-emu/rootfs/var type virtiofs (rw,relatime)
none on /run/fex-emu/rootfs/etc/resolv.conf type virtiofs (rw,relatime)
```
All the sub-mounts (e.g. /dev/shm) are automatically handled since we bind-mount with recursion. `/etc/resolv.conf` is a special case since `muvm` overlays that on the real / filesystem too. `/.profile` is an artifact of our RootFS build and `/.viminfo` is junk on my system (I could delete those but I left them around as test cases for file, not dir, mounts).

asahilina:
Interesting that all those tests failed... they pass on my machine. Any idea how I can replicate the testing environment? Different RootFS, I guess?

Sonicadvance1:
> Interesting that all those tests failed... they pass on my machine. Any idea how I can replicate the testing environment? Different RootFS, I guess?

The rootfs that the builders use is an Ubuntu 22.04 image that has been used for about two years.

asahilina:
I mean, can I grab it somewhere?

asahilina:
Wait, this isn't just an ancient kernel issue right? openat2() requires 5.6.

asahilina:
Oof... then I need to gate all the changes on kernel version and revert to the old more broken behavior if it's too old... ;;

Sonicadvance1:
> I mean, can I grab it somewhere?

https://drive.google.com/file/d/18vz4L2yMkGrI6ACiui-K2-MKHxYstCjs/view?usp=sharing For the rootfs image that I built for the builders.

alyssarosenzweig:
> > Wait, this isn't just an ancient kernel issue right? openat2() requires 5.6.
> 
> Some of the devices are running ancient kernels. 5.0 is our minspec.

Sounds like now's a good time to bump our minspec then, 5.0 predates panfrost.ko getting merged!

asahilina:
I added a quick gate (turned out to be not too horrible). That does also mean that the CI isn't testing the new approach, though...

asahilina:
One thing worth considering: Now that we have a merged RootFS, would it actually be practical to pivot_root into it? Obviously FEX still needs to be able to access files outside the RootFS, but that could be done by keeping an fd around (and pivoting back before execing a native binary or FEXInterpreter again). With some care FEX might even be able to claim actual robust isolation, as long as it can prevent that fd from leaking into the guest. Perhaps this is the right direction to go in for stuff like Flatpak to work?

asahilina:
Okay, that makes more sense. So the ARMv8 runner was on an old kernel, and it looks like I broke thunks on the new ones with the new code.

Sonicadvance1:
> Okay, that makes more sense. So the ARMv8.0 runner was on an old kernel, and it looks like I broke thunks on the new ones with the new code. Now I need to figure out how to build & test with thunks... ^^;;

I'll see what we can do about raising our kernel version minspec. Just need to check our CI system to see what needs updated, or how hard it will be to update them.

Sonicadvance1:
Took a look at our CI systems with their kernel use.
- 3 units at 6.x (various)
- 2 units at > 5.6, can easily be updated to 6.x (armv8.4)
- 1 unit at > 5.6, can be updated to 6.x fairly easily but means I need to take it offline for a bit (armv8.2)
- 1 unit at < 5.6, will be kind of a pain to update to 6.x, but is possible (armv8.0)

So we could update our minspec to 6.0 /relatively/ easily, I just need to do it. I could fit that in to this weekend or next week.

pmatos:
> > Wait, this isn't just an ancient kernel issue right? openat2() requires 5.6.
> 
> Some of the devices are running ancient kernels. 5.0 is our minspec.

Here it mentions 4.17 though:
https://github.com/FEX-Emu/FEX/blob/e44d1f136b5d508cc314d27df8d6d0aa213e6e56/Source/Tools/FEXLoader/FEXLoader.cpp#L444

Sonicadvance1:
> > > Wait, this isn't just an ancient kernel issue right? openat2() requires 5.6.
> > 
> > 
> > Some of the devices are running ancient kernels. 5.0 is our minspec.
> 
> Here it mentions 4.17 though:
> 
> https://github.com/FEX-Emu/FEX/blob/e44d1f136b5d508cc314d27df8d6d0aa213e6e56/Source/Tools/FEXLoader/FEXLoader.cpp#L444

Yea, 4.17 was where we exactly knew problems were going to happen because of old Jetsons, so se decided to put a spotlight on it. If anything when we raise the minspec for real, we can just change 4.17 to 6.0 or whatever.

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -515,23 +615,63 @@ static bool ShouldSkipOpenInEmu(int flags) {
...
+  Tmp[PathLength] = '\0';
+
+  // And try to open via EmuFD
+  auto EmuFd = EmuFD.Open(Tmp, flags, mode);
+  if (EmuFd == -1) {
+    return false;
+  }
+
+  // If we succeeded, swap out the fd
+  ::dup2(EmuFd, fd);
```
Sonicadvance1, line 639: Is there a reason to do this? We open a tmpfd for every emulatedfd. afaik there shouldn't be a need to dup the fd returned.

Sonicadvance1: Oh no, I see what you're doing here. You're duplicated over the previously opened FD. never mind.

Sonicadvance1 (approved):

neobrain:
Since it seems likely it would be affected: Has library forwarding functionality been verified to still work after this change?

asahilina:
Yes, that's what I fixed after CI caught it (it was a missed case, look for the `AT_FDCWD` tests in the last commit).

neobrain:
> Yes, that's what I fixed after CI caught it (it was a missed case, look for the `AT_FDCWD` tests in the last commit).

Great, thanks! Wasn't sure from memory if this had just been a build failure or not, but good to know it's taken care of.

Also, happy new year of course :)

PR #4226, opened 2024-12-19
InstructionCountCI: add some hot blocks from Factorio

alyssarosenzweig:
Factorio hammers its drawSprite() function and ends up cpu bound under FEX. Unfortunately, its hot blocks seem to be translated pretty optimally :-/


alyssarosenzweig:
I'm working on the 8-bit cmp, but not much we can do for the rest. Still useful as "documentation" I think

lioncash (approved):

PR #4227, opened 2024-12-20
ArchHelpers/Arm64: Fixes LDAPUR and STLUR backpatching

Sonicadvance1:
The immediate offset masking was at the completely wrong offset when I wrote these handlers. No idea how I managed to mess those up so badly.

Should fix at least some of the issues with #4216


alyssarosenzweig (approved):
Oops!

Issue #1996, opened 2022-09-16
Build thunks GuestLibs on other arm64 distros (say, Arch Linux)


phire:
Currently we rely on debian's `gcc-x86-64-linux-gnu` package to supply a working x86_64 cross-compiler on Debian/Ubuntu distros. But other distros (like Arch Linux) are missing such handy packages.

I've done a bit of preliminary research, and it seems like these are our rough options:

### Option One - Use clang instead:

Clang helpfully includes compilers for all triples by default, and cross-compiling is simply a matter of passing the correct triple into clang/clang++. This seems like an elegant solution, we already require clang for thunk generation, why not use it for building thunks too.

However... This solution *only* gets you a compiler.
The `gcc-x86-64-linux-gnu` package (and the `g++`/`multilib-i686` variants) actually pull in a bunch of other packages to supply a full libc and libc++ and other support libraries that our thunks need to build. If we go with the clang option, then we need to supply the missing include from somewhere else. 

But from where? We have a two potential options:

 * Most of the include files can be found in the host, though we will need to supply the missing x86 bits of linux headers, glibc and libstdc++. 
   * This option runs the risk that other random libraries will change their include files slightly depending on host arch, and might end up doing the wrong thing
   * We will be missing the .so files, can we create our thunks without them?
 * We just supply all the includes, either:
    * a rootfs style download (generated along-side our rootfs? That would keep them in-sync)
    * as a submodule

### Option Two - Generate equivalent packages for Arch

These packages are clearly useful for Debian, would be nice to have equivalents in Arch Linux ARM.

There are some downsides to this approach:

 1. these packages will be in AUR and will take a while to build from source (especially if people are building them on lower-end ARM SBCs, maybe we could supply equivalent binary packages?)
 2. We would need to maintain them
 3. It only fixes the issue for Arch Linux ARM and derivatives; Other distros still won't be able to build thunks.

### Idea three - Just use a native x86 toolchain inside our rootfs

This is an ugly and lazy option. Requires a large download (and probably unpacking the rootfs to install a compiler)

------

Whatever solution we go with, we should standardise across all distros, so we aren't using widely different approaches on different distros. I'm currently leaning towards Option One, with us packaging headers, not just for libc/libstdc++, but all libraries we thunk.

neobrain:
Is the issue on Arch that there is no x86 glibc/libstdc++ package available, or not even the x86 cross-compiler itself? I'm guessing chances are slim someone else already wrote an AUR for this?

[...]

PR #4230, opened 2024-12-20
Library Forwarding: Allow reading standard library headers from a development x86 rootfs

asahilina:
This is the last commit from #3597, rebased. It's sufficient to build FEX with thunks on Fedora aarch64, given a suitable x86_64/i686 sysroot with the required includes and glibc/libstdc++ and friends (and some gnarly toochain override files because the cross-compilers Fedora ships on aarch64 aren't actually configured identically to the native compilers on x86/x86_64).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,7 @@ option(USE_PDB_DEBUGINFO "Builds debug info in PDB format" FALSE)
 
 set (X86_32_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_32.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting i686")
 set (X86_64_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_64.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
+set (X86_DEV_ROOTFS "/" CACHE FILEPATH "Path to the sysroot used for cross-compiling for i686 and x86_64")
 set (DATA_DIRECTORY "${CMAKE_INSTALL_PREFIX}/share/fex-emu" CACHE PATH "global data directory")
 
 string(FIND ${CMAKE_BASE_NAME} mingw CONTAINS_MINGW)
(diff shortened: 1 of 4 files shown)


alyssarosenzweig (approved):

Sonicadvance1 (approved):
Nothing here seems scary to me.

CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ option(USE_PDB_DEBUGINFO "Builds debug info in PDB format" FALSE)
 
 set (X86_32_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_32.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting i686")
 set (X86_64_TOOLCHAIN_FILE "${CMAKE_CURRENT_SOURCE_DIR}/toolchain_x86_64.cmake" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
+set (X86_DEV_ROOTFS "/" CACHE FILEPATH "Toolchain file for the (cross-)compiler targeting x86_64")
```
neobrain, line 40: This still needs a proper description :)

neobrain:
> It's sufficient to build FEX with thunks on Fedora aarch64, given a suitable x86_64/i686 sysroot with the required includes and glibc/libstdc++ and friends (and some gnarly toochain override files because the cross-compilers Fedora ships on aarch64 aren't actually configured identically to the native compilers on x86/x86_64).

Could you provide a vague run-down of the required tweaks for reference (or just post an example toolchain file)?

teohhanhui:
I think that's

https://github.com/FEX-Emu/FEX/issues/1996#issuecomment-2557881892

right?

asahilina:
Our Fedora setup is here: https://src.fedoraproject.org/rpms/fex-emu/pull-request/7

Fedora only ships "generic" cross-compilers with no system headers or glibc, which are intended to be used for freestanding code. The target tuple is different to the native compiler. To make those work for building thunks, we ship a minimal sysroot with just the required system headers (built from x86_64 devel packages), and toolchain files with include overrides and which disable linking with native libs and startup files (since those aren't needed for building shared libraries). I also force disable the cmake compiler checks, since this setup can't build executables (only libs).

I don't think there's much more to do on the FEX side here, since these hacks are Fedora-specific. Distros which ship similar cross-compilers can do something similar, and will probably need their own tweaks anyway since the `-redhat` tuple stuff is obviously very specific to Fedora/RH, and there are probably other differences in how the headers are installed anyway. Distros which ship full cross-compilers with glibc/etc can just use the existing FEX toolchain files (probably).

CMakeLists.txt
asahilina: Done!

asahilina:
I don't know what's up with that failing test... I don't see how it can be related to this PR?

neobrain:
> I don't know what's up with that failing test... I don't see how it can be related to this PR?

Don't worry about it, this is pending on #4232 .

(Happy holidays!)

PR #4231, opened 2024-12-20
OpcodeDispatcher: Minor division improvement 

Sonicadvance1:
No need to extract the subregisters out before operating on them since
the long division and long remainder IR operations correctly zero/sign
extend the incoming sources as necessary. Saves a couple of
instructions.


alyssarosenzweig (approved):
Sure

Issue #4217, opened 2024-12-13
SOMA: Cannot open drawers or kitchen cabinets in the starting room


kuruczgy:
**What Game**
SOMA
https://store.steampowered.com/app/282140/SOMA/

**Describe the bug**
Cannot open drawers or kitchen cabinets in the starting room.

**To Reproduce**
1. Start a new game. (Wait for a few minutes until the cutscenes are over and you wake up in the starting room.)
2. Put your crosshair over a drawer or a kitchen cabinet in the room, and try to drag it by holding down the left mouse button.
3. Observe that the hand symbol that should appear when hovering is missing, and the dragging does not work.

**Expected behavior**
You can open drawers and kitchen cabinet doors.

**Screenshots and Video**
This is how it's supposed to work: https://youtu.be/FD0h-s6NUeM?t=311

**System information:**
 - OS: NixOS unstable
 - CPU/SoC: Snapdragon X Elite (X1E80100)
 - Video driver version: Mesa 25.0.0-devel (commit a2339542f5d71b1c60febff78fb5dbad7ceb9191)
 - RootFS used: none (I use bubblewrap instead)
 - FEX version: 2412 (+ this patch: https://github.com/kuruczgy/nixos-aarch64-gaming/blob/6a8acdb69f2963a3f5f998f3a65c8520dfffa6b8/packages/fex-emu-shebang-absolute-path-fix.patch )
 - Thunks Enabled: No

See this commit for the nix expressions for the bwrap based chroot: https://github.com/kuruczgy/nixos-aarch64-gaming/tree/a747c19f80f64ce566f315ac5c93d1b9345c13d4

**Additional context**
 - Is this an x86 or x86-64 game: Unknown
 - Does this reproduce on AArch64 with Radeon/Intel/Nvidia: Untested
 - Is this a Vulkan game: Unknown

Sonicadvance1:
Reproduced this. I don't recall this being a problem a couple years ago, so smells like a regression.
I'll need to tinker with it to see if I can find how they are doing their raycasting to find interactable objects to determine the cause.

kuruczgy:
I wanted to bisect this, the issue is already present in `2402`, but anything before that wouldn't launch steam for me. (I also cherry-picked #4160, but no luck.)

[...]

Issue #1663, opened 2022-04-28
Emit FJCVTZS on ARMv8.3+


CallumDev:
For float -> 32-bit int overflow, `fjcvtzs` will produce a result of `0x80000000` matching X87 `fist` and SSE `CVTTSD2SI`. Easy accuracy win on ARMv8.3+ to replace `fcvtzs` with `fjcvtzs`

bylaws:
This isn't actually the case

alyssarosenzweig:
Yeah, fjcvtzs doesn't seem to match x86 behaviour despite a lot of people who seem to be confused about either what arm does or what x86 does.

PR #4237, opened 2024-12-28
Fix float->int conversion overflow behaviour

### LINKED ISSUE #2995: cvttpd2dq is probably broken

bylaws:
ARM behaviour here is to saturate on overflow or NaN inputs, whereas X86 returns a sentinel value of 2^(bitsize-1), explicitly emulate this.

Closes #2995


bylaws:
Needs unittests + sve256 fixes, but this should be a better approach than checking fpsr flags. Hopefully this doesn't have too significant an impact on real perf

Sonicadvance1:
Does this handle underflow cases as well?

Sonicadvance1:
This also fixes "The Talos Principle" timing bug with audio cutting out after a few seconds, and the intro scenes playing on superspeed.

Sonicadvance1:
This also fixes #4217

Sonicadvance1:
This also fixes Animal Well's music have chirps and pops in it.

bylaws:
> Does this handle underflow cases as well?

This works with the standard arm behaviour, this is tested now

bylaws:
Opts:
- https://github.com/FEX-Emu/FEX/issues/1663
- Synthesise some of the constants

unittests/InstructionCountCI/FEXOpt/MultiInst_TSO_32bit.json
```diff
@@ -65,66 +65,6 @@
         "subs w26, w4, w20",
         "mov x4, x26"
       ]
-    },
```
Sonicadvance1, line 68: Looks like you accidentally reverted this addition.

unittests/InstructionCountCI/FlagM/HotBlocks.json
```diff
@@ -693,71 +693,6 @@
         "mov x26, x5"
       ]
     },
-    "Factorio drawSprite+0x890": {
```
Sonicadvance1, line 696: Same here

Sonicadvance1 (approved):

Sonicadvance1:
Apparently this also fixes Satisfactory's dedicated server crashing.

neobrain:
Good fix, and happy new year!

For the record though, please avoid opening placeholder PRs without a description and changing them into proper ones after the fact. If people unsubscribe for what looks like a playground PR, these updates will be missed. (Depending on the mail client, all updates will be collapsed under the original PR title as well.)

Opening a new PR instead avoids these problems.

PR #4253, opened 2025-01-03
x87StackOptimizationPass: Minor opt to f80 fchs and fabs 

Sonicadvance1:
It's faster to load the f80 sign mask from our named vector constants
than synthesizing the values. Changes a 4 instruction sequence to
synthesize to be 1 load.


alyssarosenzweig (approved):
Sure, why not? :package:

pmatos:
lgtm here as well. nice little opt! :)

lioncash (approved):
nice

PR #4259, opened 2025-01-09
FEXConfig: Fixes instcount not being editable by keyboard

### LINKED ISSUE #4121: FEXConfig: Block size can't be modified with keyboard
Image_2024-10-17_15-01-02

Means making large changes impossible and I've fallen back to editing Config.json directly.
No idea what the problem is. QT garbage?

Sonicadvance1:
Fixes #4121

### DIFF
--- a/Source/Tools/FEXConfig/main.qml
+++ b/Source/Tools/FEXConfig/main.qml
@@ -222,6 +222,8 @@ ApplicationWindow {
     component ConfigSpinBox: SpinBox {
         property string config
 
+        editable: true
+
         textFromValue: (val) => {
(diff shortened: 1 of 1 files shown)


lioncash (approved):

neobrain:
Have you confirmed this also works on Qt 5 (which is used in our Ubuntu LTS builds)?

Sonicadvance1:
> Have you confirmed this also works on Qt 5 (which is used in our Ubuntu LTS builds)?

Yes it does, this is what I tested with.

neobrain:
Great, thanks!

PR #4260, opened 2025-01-09
Profiler: Setup for usage on Windows

Sonicadvance1:
This will get gpuviz working under Wine.

### DIFF
--- a/FEXCore/Source/Utils/Profiler.cpp
+++ b/FEXCore/Source/Utils/Profiler.cpp
@@ -7,6 +7,7 @@
 #include <linux/magic.h>
 #include <sys/stat.h>
 #include <sys/vfs.h>
+#include <time.h>
 #endif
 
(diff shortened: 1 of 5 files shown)


lioncash (approved):

bylaws:
(you'll need to impl QueryPerformanceCounter in the WinAPI layer as well here for this to work)

bylaws (approved):

Sonicadvance1:
Had to do some changes to this to get it fully working. Had to fix a issue in CRT/IO that files opened for writing should be shared with other writers, otherwise the trace file would get ENOENT.
Also had to fix the win32 performance query version of `GetTime`. We always want nanoseconds but Wine divides the result by 100. So we need to query the frequency and calculate the multiplier to get back to nanoseconds rather than ticks.

Also had to actually initialize the profiler interface.

lioncash (approved):

PR #4265, opened 2025-01-10
Revert pred cache

pmatos:


Sonicadvance1 (approved):

Sonicadvance1:
Hopefully we can get a machine with SVE in CI sometime soon so we can actually catch regressions like these.

Issue #4252, opened 2025-01-03
Crysis 2 Maximum edition: Audio thread completely saturated with x87, breaks audio 


Sonicadvance1:
The first level of the game completely saturates its audio thread, resulting in it dropping samples. Even in reduced precision mode. Pretty much all the CPU time is spent in its fmodex.dll. and most of the blocks are just a ton of x87.

```
  19.30%  [JIT] tid 494822     [.] JIT_0x2e7a85c_0x413e867fc
  11.39%  [JIT] tid 494822     [.] JIT_0x2ea4521_0x413e642a4
  10.69%  [JIT] tid 494822     [.] JIT_0x2ea46e9_0x413e6d978
   7.19%  [JIT] tid 494822     [.] JIT_0x2e8e7af_0x413e94d28
   4.86%  [JIT] tid 494822     [.] JIT_0x2ea3931_0x413e39858
   4.58%  [JIT] tid 494822     [.] JIT_0x2e8e6e1_0x413ea3658
   3.84%  [JIT] tid 494822     [.] JIT_0x2e8e761_0x413ea3ce8
   2.84%  [JIT] tid 494822     [.] JIT_0x2ea4085_0x413e5119c
   2.74%  [JIT] tid 494822     [.] JIT_0x2e8eb0d_0x413e9a00c
   2.12%  [JIT] tid 494822     [.] JIT_0x2e735d4_0x413e9e090
   2.05%  [JIT] tid 494822     [.] JIT_0x2e8e659_0x413ea3330
   1.62%  [JIT] tid 494822     [.] JIT_0x2e8e373_0x413ea1cf0
   1.39%  FEXInterpreter       [.] 0x0000000000185480
   0.99%  [JIT] tid 494822     [.] JIT_0x2e8eb51_0x413e9a79c
   0.90%  [JIT] tid 494822     [.] JIT_0x2e8eb04_0x413e99a64
   0.83%  [JIT] tid 494822     [.] JIT_0x2e8e4dd_0x413e90e94
   0.78%  [JIT] tid 494822     [.] JIT_0x2ea46cf_0x413e69d04
   0.78%  [JIT] tid 494822     [.] JIT_0x2e8e4df_0x413e905ec
   0.70%  [JIT] tid 494822     [.] JIT_0x2ea44ed_0x413e605dc
   ```
fmodex.dll base address at 0x2e20000 in this capture.

First block:
```asm
02e7a85c  fld     st0, dword [edx]
[...]

PR #4266, opened 2025-01-10
x87 fst/fld optimization for different addrmodes

pmatos:
Includes tests and instcountci files and tests.
When the x87 optimizations were implement, we missed optimizing different addressing modes. This commit addresses this issue.

Discussed in #4252 .


unittests/InstructionCountCI/x87.json
```diff
@@ -17815,6 +17815,40 @@
         "bic w20, w21, w20",
         "strb w20, [x28, #1298]"
       ]
+    },
+    "Multiple fld/fst": {
+      "x86InstructionCount": 4,
+      "ExpectedInstructionCount": 22,
+      "x86Insts": [
+        "fld     dword [ebp+16380]",
```
Sonicadvance1, line 17823: For the two 64-bit addressing variants of this in x87.json and x86_f64.json would you what the address modes to be 64-bit so we're testing this instead of the 32-bit zero-extend implementation? I think that would matter more
```
        "fld     dword [rbp+16380]",
        "fstp    dword [rax-0x4]",
        "fld     dword [rbp-0x8]",
        "fstp    dword [rax+16370]"
```

pmatos: I have changed this as suggested. Thanks.

Sonicadvance1 (approved):

Issue #4267, opened 2025-01-11
Don't depend on any binary blobs


kuruczgy:
Currently the tests depend a bunch of prebuilt binaries through submodules, e.g. `External/fex-gcc-target-tests-bins` and some others.

Binary blobs are bad from an auditability perspective. One could say that "those are just for tests so they don't count", but I am not aware of any distro packaging system that lets one (easily) selectively clone submodules, so in practice they are all present during builds.

Building the test binaries from source would be an obvious idea, but e.g. pulling in the whole gcc repo is probably even less desirable than the current solution, so I see why it's not done.

Ideas for alternatives:
- Instead of using submodules, manually `git clone --depth 1` the bin repos in the CI.
- Maybe there is some way to configure `.gitmodules` to not clone these submodules by default with `git clone --recursive`? E.g. not sure if `submodule.<name>.fetchRecurseSubmodules` does that.

In either case, it should be documented which submodules are needed for building FEX and which ones only for the tests (and which tests), currently it's a bit of a guessing game.

Also honorable mention to `Source/Windows/wine_builtin.bin`, but that's small enough to be audited by hand with a hex editor, so it's not a priority. (It's just text with some null bytes afterwards.)

Sonicadvance1:
Our previous Arch package file would skip the submodule checkout for the test binaries, but that's just an adjacent topic.

As you thought, building these on the CI systems aren't a viable solution, especially as some of them are a pain to compile, take a long time to compile, and/or compiled with -march=native on some AVX supporting system. Additionally they would need cross compiling on any ARM CI machine which adds its own layer of pain. This is why we chose to have them as just binary submodules.

For our Ubuntu PPA, we also just remove a bunch of folders that are unnecessary for reducing the size of the source package.
```
rm -Rf $TEMP_SOURCE/External/{fex-gcc-target-tests-bins,fex-gvisor-tests-bins,fex-posixtest-bins}
rm -Rf $TEMP_SOURCE/.git
rm -Rf $TEMP_SOURCE/unittests
rm -Rf $TEMP_SOURCE/External/vixl/src/aarch32/
rm -Rf $TEMP_SOURCE/External/vixl/test
```

So it can be seen that basically just the bins folders are unnecessary, the other folders are just tertiary things that don't really matter. There can probably be some documentation somewhere about these three folders being unnecessary.

kuruczgy:
Thanks, that list is useful, I agree that it would be useful to document them somewhere properly.

[...]

PR #4268, opened 2025-01-12
Windows: Removes opaque wine_builtin binary

Sonicadvance1:
Instead generate it with an echo command which is easier to see what it is doing. It's a trivial file that is just a string and 16 zeroes.

Brought up by #4267


neobrain:
-1 on this. Generating this file at build-time doesn't really add anything but increases build script complexity, and the issue reporter mentioned it's not a real issue for auditability:

> wine_builtin.bin [is] small enough to be audited by hand with a hex editor, so it's not a priority. (It's just text with some null bytes afterwards.)

Instead of moving the file contents to CMake, auditability would probably benefit more from a comment explaining what the file is actually used for.

PR #4269, opened 2025-01-12
JIT: Avoid OOB EC bitmap checks in ExitFunction

bylaws:
Multiblock can generate such not-taken ExitFunction calls when enabled


Sonicadvance1 (approved):
Interesting that `RtlIsEcCode` doesn't handle non-canonical addresses for you, but sure.

alyssarosenzweig (approved):

PR #4270, opened 2025-01-12
Frontend: Disallow cross-page branches in multiblock

bylaws:
This avoids both the generation of multiblocks that cover massive spans of guest code, which causes issues for both SMC and context reconstruction and attempting to decode branch targets in unmapped memory regions.

Once support for querying mappings from the FEX frontend is in place this limit could be increased if necessary, but this seems fine for now.

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -929,6 +929,7 @@ void Decoder::BranchTargetInMultiblockRange() {
   uint64_t TargetRIP = 0;
   const auto GPRSize = CTX->GetGPROpSize();
   bool Conditional = true;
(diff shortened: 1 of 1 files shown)


neobrain:
> which causes issues for both SMC and context reconstruction and attempting to decode branch targets in unmapped memory regions.

What are these issues?

> Once support for querying mappings from the FEX frontend is in place

Why is this needed for increasing the limit? Couldn't we equally well constrain to a neighborhood of N pages around the multiblock entrypoint or prevent following non-local jumps?

Sonicadvance1 (approved):

neobrain (approved):
Thanks for the additional background!

Further context from Discord is that these Ubisoft titles apparently map one single page of a larger block of code, hence "cutting off" the control flow graph.

PR #4271, opened 2025-01-12
CPUBackend: Make guest RIP reconstruction offsets signed

bylaws:
With multiblock enabled, host code generated from guest code with a lower address may be placed after host code generated from guest code with a higher address in a multiblock. As each guest RIP reconstruction entry is always relative to the one before it the offset needs to be signed to allow this.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -115,7 +115,7 @@ namespace CPU {
       uint16_t HostPCOffset;
 
       // How much to offset the RIP from the previous entry.
-      uint16_t GuestRIPOffset;
+      int16_t GuestRIPOffset;
     };
 
     /**
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -35,6 +35,7 @@ $end_info$
 #include <stdio.h>
 #include <unistd.h>
 #include <string.h>
+#include <limits>
 
 static constexpr size_t INITIAL_CODE_SIZE = 1024 * 1024 * 16;
 // We don't want to move above 128MB atm because that means we will have to encode longer jumps
@@ -875,6 +876,11 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
     for (size_t i = 0; i < DebugData->GuestOpcodes.size(); i++) {
       const auto& GuestOpcode = DebugData->GuestOpcodes[i];
       auto& RIPEntry = JITRIPEntries[i];
+      uint64_t HostPCOffset = GuestOpcode.HostEntryOffset - CurrentPCOffset;
+      int64_t GuestRIPOffset = GuestOpcode.GuestEntryOffset - CurrentRIPOffset;
+      LOGMAN_THROW_AA_FMT(HostPCOffset <= std::numeric_limits<uint16_t>::max(), "PC offset too large");
+      LOGMAN_THROW_AA_FMT(GuestRIPOffset >= std::numeric_limits<int16_t>::min(), "RIP offset too small");
+      LOGMAN_THROW_AA_FMT(GuestRIPOffset <= std::numeric_limits<int16_t>::max(), "RIP offset too large");
       RIPEntry.HostPCOffset = GuestOpcode.HostEntryOffset - CurrentPCOffset;
       RIPEntry.GuestRIPOffset = GuestOpcode.GuestEntryOffset - CurrentRIPOffset;
       CurrentPCOffset = GuestOpcode.HostEntryOffset;


FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -115,7 +115,7 @@ namespace CPU {
       uint16_t HostPCOffset;
 
       // How much to offset the RIP from the previous entry.
-      uint16_t GuestRIPOffset;
+      int16_t GuestRIPOffset;
```
neobrain, line 118: Should we add an assert in JIT.cpp to ensure this never overflows? Seems a bit sketchy to have this 16-bit limitation with seemingly no safeguards in place to ensure we consistently adhere to it.

Sonicadvance1 (approved):

Sonicadvance1 (approved):

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -875,6 +876,13 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
     for (size_t i = 0; i < DebugData->GuestOpcodes.size(); i++) {
       const auto& GuestOpcode = DebugData->GuestOpcodes[i];
       auto& RIPEntry = JITRIPEntries[i];
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
```
neobrain, line 879: The checks here are trivial (particularly compared to the two memory stores right after), so there's no need to restrict them to assertion-builds.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -875,6 +876,13 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
...
+      uint64_t HostPCOffset = GuestOpcode.HostEntryOffset - CurrentPCOffset;
+      int64_t GuestRIPOffset = GuestOpcode.GuestEntryOffset - CurrentRIPOffset;
+      LOGMAN_THROW_AA_FMT(HostPCOffset <= std::numeric_limits<uint16_t>::max(), "PC offset too large");
```
neobrain, line 882: Did you measure a performance benefit from using the `AA` variant here? Otherwise, `A` should be used by default to avoid communicating ambiguous intent.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: This seems to be a code pattern used throughout FEX, unsure why

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: https://github.com/FEX-Emu/FEX/blob/48c03d747abc0357cfdc45ebcbf0995fce39a152/FEXCore/include/FEXCore/Utils/LogManager.h#L49

Perhaps I misunderstood this comment? It seems to state the opposite

FEXCore/Source/Interface/Core/JIT/JIT.cpp
neobrain: I don't see a reason either, but a bad pattern being widespread doesn't make a compelling reason to stick to it either way.

bylaws: Yeah I'll clean it up in general

FEXCore/Source/Interface/Core/JIT/JIT.cpp
neobrain: That comment is [not backed by data](https://github.com/FEX-Emu/FEX/pull/1840#issuecomment-1179511914) and it's certainly not a good default... I hadn't realized (or forgot?) that the PR that introduced this feature mass-converted existing assertions, so this seems like something to address separately.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
neobrain: Besides, there's a certain irony in using `__builtin_assume` annotations only if the macro to enable expensive assertions is active.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: I'm confused though why assume would be harmful? It seems like you would want to assume the condition holds true? At worst the downside is negligible

FEXCore/Source/Interface/Core/JIT/JIT.cpp
Sonicadvance1: The pattern here of wrapping `LOGMAN_THROW_{A,AA}_FMT` usage in additional `#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED` wrappers only matters in non-trivial places.

Like in ThreadManager.cpp this happens because it ensures correctness by locking a mutex and walking an array. So it does both, but we don't want that mutex locking in release builds.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
neobrain: > I'm confused though why assume would be harmful?

In general, effectively placed compiler hints communicate where hotspots are located and what invariants their efficient execution relies on. If you start placing them arbitrarily, identifying the deliberately placed hints becomes impossible.

More pragmatically, pointless `assume` directives will interfere with debugging of non-assertion builds. Since the compiler generates code as-if the condition were true, a debugger will print wrong variable values or have erratic source stepping behavior. Even [debug printfs](https://godbolt.org/z/5oox8jn6h) will print misleading information.

bylaws: Right, I just am uncertain that any such hot paths where this would realistically have an observable impact exist in fex, and as such I think it might be preferable to remove this all together without anything to the contrary?

neobrain: > it might be preferable to remove this all together without anything to the contrary?

That matches my opinion, but I didn't want to bother this PR with that conversation. Sorry if I was unclear that it's not crucial right now, since I only realized it was already in widespread use after my initial comment.

neobrain (approved):
Let's pick up the discussion about `__builtin_assume` somewhere else. The unnecessary `ASSERTIONS_ENABLED` noise should probably go, but other than that this looks good!

Sonicadvance1 (approved):

PR #4273, opened 2025-01-12
Frontend: End multiblocks early after hitting 2 consecutive null bytes

bylaws:
'add [rax], al' is almost never seen in actual code so the assumption can be made that we are most likely trying to explore garbage code and that this will never be hit. If it is then code will be generated at that point (where Entrypoint == true).


Sonicadvance1 (approved):

alyssarosenzweig (approved):
x86 moment.

alyssarosenzweig:
clang format blocking the merge

bylaws:
fixed

Issue #4264, opened 2025-01-10
Cached predicate register being clobbered by softfp lib call


pmatos:
Unfortunately, the code to cache predicate registers for SVE x87 ldst optimization and the precision tests that do a bunch of:

```
load
op
store/pop
```

were developed at a similar time and landed more of less at a similar time. We might not have SVE hw in CI and I didn't test this until now. The culprit commit is 72a40636515aaa8986f500ce7f4c5cb546d6127f . What's happening is that in this specific case we are caching p2, calling softfp fcos impl which clobbers p2 and its use in fstp causes the bug - i.e. the test to fail.

PR #4274, opened 2025-01-13
Ensure predicate cache is reset when control flow leaves block

pmatos:
Whenever the control float leaves the block, it might clobber the
predicate register so we reset the cache whenever that happens.

The difficulty here is that the cache is valid only during IR generation
so we need to make sure we catch all the cases during this pass where
the execution might leave the block.

Fixes https://github.com/FEX-Emu/FEX/issues/4264


pmatos:
To be fair, this is not my favourite fix because it risks missing places where the cache should be reset but isn't and it will be a pain to debug cases that fail if you don't keep this issue in the back of your mind.

But the predicate cache is only valid during IR generation so we cannot use any of the backend stuff like Spill functions.

Sonicadvance1 (approved):
Interesting way to fix it, but I'm fine with it.

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87.cpp
```diff
@@ -164,6 +168,9 @@ void OpDispatchBuilder::FADD(OpcodeArgs, IR::OpSize Width, bool Integer, OpDispa
   if (Op->Src[0].IsNone()) { // Implicit argument case
     auto Offset = Op->OP & 7;
     auto St0 = 0;
+    if (!ReducedPrecisionMode) {
```
bylaws, line 171: I'm trying to understand why this check is here but I don't quite get it, won't this code only run for full precision anyway? I would have thought reduced would take https://github.com/FEX-Emu/FEX/blob/8cfc016b3f8b8dfba98cacdd4f18886661c95674/FEXCore/Source/Interface/Core/OpcodeDispatcher/X87F64.cpp#L131

Also just thinking through perhaps it would be cleaner to do this in FlushRegisterCache (!SRAOnly case) and to save the predicate when spilling/filling regs?

pmatos: Thanks - that's a great point. My intention was to merge (`FADD` and `FADDF64`) those and I didn't but thought I did.

I did attempt to do this through `FlushRegisterCache` but this one is called at `_Break()`. This seems to be called by `INTOp` and we don't want to reset the cache in these cases - when I attempted this, it was basically resetting the cache too often, which is why the commit calls `ResetInitPredicateCache()` alongside `FlushRegisterCache` mostly everywhere but not on `Break()`.

I am happy to hear suggestions on how to improve this if you have any ideas. :)

pmatos: Also, you mention saving the predicate reg, but if we do managed to reset the predicate register in `FlushRegisterCache` then there was no need to save the predicate in spilling/filling regs. The reason for not spilling/filling the predicate register is because afaiu doing this for predicate registers takes a few instructions and would most likely void the optimization that we are attempting to do, so it's better to simply reset the cache and then regenerate the predicate with a single `ptrue`.

bylaws: > Also, you mention saving the predicate reg, but if we do managed to reset the predicate register in `FlushRegisterCache` then there was no need to save the predicate in spilling/filling regs.

Hehe, yeah this is true :)

> The reason for not spilling/filling the predicate register is because afaiu doing this for predicate registers takes a few instructions and would most likely void the optimization that we are attempting to do, so it's better to simply reset the cache and then regenerate the predicate with a single `ptrue`.

Right yeah, I missed that we'd have to save all the predicate registers since there's no way to know which have changed - though looking through only one predicate type is ever cached, so the cache could be replaced with statically allocating a single extra predicate register.  Then that is always spilled, even if a few such registers are needed in the future I think the cost would be negligible.

> I did attempt to do this through FlushRegisterCache but this one is called at _Break(). This seems to be called by INTOp and we don't want to reset the cache in these cases - when I attempted this, it was basically resetting the cache too often, which is why the commit calls ResetInitPredicateCache() alongside FlushRegisterCache mostly everywhere but not on Break().

I'm quite surprised break/INT is being called that often in x87 code, what ends up causing this? If that can be figured then clearing the predicate cache for all !JITDispatch ops in the IR json (I think this is what you are trying to achieve by manually annotating x87 ops? Correct me if I'm wrong) could work, there's also cases like cpuid, ludiv etc that I think would need to clear the cache but don't currently.  I personally prefer the former solution I mentioned since it's nice to have the guarantee that spilling and filling registers in a handler won't cause issues but will leave the choice up to you

pmatos: So I was looking at what's causing INTOp to be called and I am a little bit mistified myself about this. Take the test:
```
%ifdef CONFIG
{
  "RegData": {
    "MM7":  ["0x8000000000000000", "0x4000"]
  }
}
%endif

lea rdx, [rel data]
fld tword [rdx + 8 * 0]

lea rdx, [rel data2]
lea rax, [rdx + 8 * 0]
fstp tword [rax]
fld tword [rdx + 8 * 0]

hlt

align 8
data:
  dt 2.0
  dq 0
data2:
  dt 0.0
  dq 0
```

This calls fstp twice, you should only need to reset the predicate cache at the beginning of the block. However, here if you call the reset function from FlushRegisterCache, you end up resetting it several times due to INTOp calling it. I have not yet figured out why it's being called but it's related to threading. It comes from `Core.cpp:668`:
```
          std::invoke(Fn, Thread->OpDispatcher, DecodedInfo);
```

The backtrace from ResetInitPredicateCache shows:
```
Breakpoint 1, FEXCore::IR::IREmitter::ResetInitPredicateCache (this=0x7ff6ef7000)
    at /home/steamvr/dev/FEX/FEXCore/Source/Interface/IR/IREmitter.h:77
77	    InitPredicateCache.clear();
(gdb) bt
#0  FEXCore::IR::IREmitter::ResetInitPredicateCache (this=0x7ff6ef7000)
    at /home/steamvr/dev/FEX/FEXCore/Source/Interface/IR/IREmitter.h:77
#1  0x0000005555723324 in FEXCore::IR::OpDispatchBuilder::FlushRegisterCache (this=0x7ff6ef7000, SRAOnly=false)
    at /home/steamvr/dev/FEX/FEXCore/Source/Interface/Core/OpcodeDispatcher.h:1179
#2  0x0000005555794a34 in FEXCore::IR::OpDispatchBuilder::INTOp (this=0x7ff6ef7000, Op=0x7fe2e00300)
    at /home/steamvr/dev/FEX/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp:4693
#3  0x000000555573eb6c in std::__invoke_impl<void, void (FEXCore::IR::OpDispatchBuilder::*&)(FEXCore::X86Tables::DecodedInst const*), FEXCore::Core::NonMovableUniquePtr<FEXCore::IR::OpDispatchBuilder>&, FEXCore::X86Tables::DecodedInst const*&> (
    __f=@0x7fffffdb50: (void (FEXCore::IR::OpDispatchBuilder::*)(FEXCore::IR::OpDispatchBuilder * const, const FEXCore::X86Tables::DecodedInst *)) 0x5555794860 <FEXCore::IR::OpDispatchBuilder::INTOp(FEXCore::X86Tables::DecodedInst const*)>, __t=..., 
    __args=@0x7fffffdc20: 0x7fe2e00300) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:74
#4  0x000000555573eacc in std::__invoke<void (FEXCore::IR::OpDispatchBuilder::*&)(FEXCore::X86Tables::DecodedInst const*), FEXCore::Core::NonMovableUniquePtr<FEXCore::IR::OpDispatchBuilder>&, FEXCore::X86Tables::DecodedInst const*&> (
    __fn=@0x7fffffdb50: (void (FEXCore::IR::OpDispatchBuilder::*)(FEXCore::IR::OpDispatchBuilder * const, const FEXCore::X86Tables::DecodedInst *)) 0x5555794860 <FEXCore::IR::OpDispatchBuilder::INTOp(FEXCore::X86Tables::DecodedInst const*)>, 
    __args=@0x7fffffdc20: 0x7fe2e00300, __args=@0x7fffffdc20: 0x7fe2e00300)
    at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:96
#5  0x0000005555723eb8 in std::invoke<void (FEXCore::IR::OpDispatchBuilder::*&)(FEXCore::X86Tables::DecodedInst const*), FEXCore::Core::NonMovableUniquePtr<FEXCore::IR::OpDispatchBuilder>&, FEXCore::X86Tables::DecodedInst const*&> (
    __fn=@0x7fffffdb50: (void (FEXCore::IR::OpDispatchBuilder::*)(FEXCore::IR::OpDispatchBuilder * const, const FEXCore::X86Tables::DecodedInst *)) 0x5555794860 <FEXCore::IR::OpDispatchBuilder::INTOp(FEXCore::X86Tables::DecodedInst const*)>, 
    __args=@0x7fffffdc20: 0x7fe2e00300, __args=@0x7fffffdc20: 0x7fe2e00300)
    at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/functional:120
#6  0x000000555572247c in FEXCore::Context::ContextImpl::GenerateIR (this=0x7ff6e57000, Thread=0x7ff6ee3000, GuestRIP=65536, 
    ExtendedDebugInfo=false, MaxInst=0) at /home/steamvr/dev/FEX/FEXCore/Source/Interface/Core/Core.cpp:668
#7  0x00000055557248f4 in FEXCore::Context::ContextImpl::CompileCode (this=0x7ff6e57000, Thread=0x7ff6ee3000, GuestRIP=65536, 
    MaxInst=0) at /home/steamvr/dev/FEX/FEXCore/Source/Interface/Core/Core.cpp:801
#8  0x0000005555724e84 in FEXCore::Context::ContextImpl::CompileBlock (this=0x7ff6e57000, Frame=0x7ff6ee3090, GuestRIP=65536, 
    MaxInst=0) at /home/steamvr/dev/FEX/FEXCore/Source/Interface/Core/Core.cpp:846
```

Do you understand what's happening? I will keep trying to understand this in any case.

pmatos: hummm... maybe this INTOp is not actually happening in between the instructions that are relevant so it is probably a red-herring.

pmatos: OK, actually leaning on FlushRegisterCache works once I moved all the predicate cache resets onto IR generation time. I was initially doing some of the resets during IR generation time and others inside the x87 stack opt. pass and this didn't work. I think it's much better now. WDYT?

bylaws: This seems good - it's worth double checking this works for cpuid/ludiv, I'm not entirely sure if flushregistercache ends up being called in such cases?

bylaws: (also fallback inst handlers for x87 ops), I might be misunderstanding something here though

pmatos: CPUID is fine (added test). What's ludiv?
Yeah, the x87 handlers still need the call and I don't know what I was thinking when I removed them.

Back in place.

bylaws: I think in your test cpuid is being constant folded so the block exit doesn't occur if it's still needed around x87 handler's then I'm certain that it would also be there - what I would do then is rather than manually annotating x87 ops is to add a field to IR.json to indicate that pred regs are clobbered and set it for all interpreted ops (x87, some SSE) and those that call PushDynamicRegsAndLR (just grep) and those that reference https://github.com/FEX-Emu/FEX/blob/d01db8f2932235325f40cd4dc01acdc6c5099b2a/FEXCore/include/FEXCore/Core/CoreState.h#L302
But this should be everything then, also perhaps add a comment to pushdynamicregsandlr to be very careful

pmatos: Actually it's not being constant folded, after some investigation, it's also not `CPUID` properly resetting the stack and it works in this specific case by chance. *sigh*. 

The reason things work in the test is because the test does this:
1. Calls LoadFromSource.
1.1. Calls InitPredicate to generate the predicate.
2. Calls _PushStack.
2.1. Forces a state change from mmx to x87 which calls FlushRegisterCache, and Resets the predicate. This is actually unnecessary I must say but it does trick this test into passing.
3. CPUID does its thing.
4. FST calls InitPredicate because the cache has been reset, but not by CPUID...

Also, I am not sure changing IR.json will make things simpler.
All of the predicate register initialization / cache reset needs to be done at IR generation time. However, a lot of X87 IR is generated at x87 pass time. For example, just for F80 addition we have: `F80AddStack`, `F80AddValue`, `F80Add`. The last is only generated by the x87 stack pass. The first two generate the third in different ways. If we were to annotate IR.json, the correct way would be to annotate the 3 instructions, but calling it from F80Add is useless. We could just annotate the ones that we know are called at IR generation time but in that case, why not annotate directly in code in X87.cpp?

pmatos: Slightly conflicted about this but I will try the IR.json annotation solution and see how far I get.

pmatos: Another disadvantage of annotating the IR.json is when we generate multiple x87 instructions to convert to values and then perform calculations, each of those will reset the cache, when you would have only needed to do it once... it might be pretty fast to reset an empty cache but it's still time take in code that sometimes needs to be as fast as possible. On the positive side if someone in the future adds a call to one of this functions, it doesn't need to remember to call the cache reset.

bylaws: Ah I see what you mean with the pass - you're right there, it really does seem to some degree that just statically allocating a single predicate register is the cleanest way to do this

pmatos: Won't I have the same problem of having to save / restore the static predicate register anyway if I want to cache it?

bylaws: Yeah you have to save and restore it but rather than the 16 registers prior it's just 1 so can just go in spillsra

pmatos: @bylaws I finished implementing your earlier suggestion, i.e. property in IR.json rather than static register. It seems to be fine and I think I got all the instructions that you mentioned that are either cpuid/ludiv or call PushDynamicRegsAndLR. Tests and instcountci look good.

pmatos: Ufff, but we probably shouldn't do it for all !JITDispatch instructions right? Some of them could potentially just lower to other instructions that don't need to reset the cache. 

Whole this work is to enable a relatively small optimization (see https://github.com/FEX-Emu/FEX/pull/4166#issuecomment-2488115966) and I am wondering if at some point we might have to say that it's better to go without it.

pmatos: Explicitly marked VPCMPESTRX for now.

bylaws: > Ufff, but we probably shouldn't do it for all !JITDispatch instructions right? Some of them could potentially just lower to other instructions that don't need to reset the cache
Yeah right this is just an issue in general, I think in some sense if an instruction is always lowered into something else it should be marked differently to one that uses a fallback. 
> 
> Whole this work is to enable a relatively small optimization (see [#4166 (comment)](https://github.com/FEX-Emu/FEX/pull/4166#issuecomment-2488115966)) and I am wondering if at some point we might have to say that it's better to go without it.

Yeah - this is why I suggested just doing it statically, it's so awkward with all the edge cases when statically allocating the pred reg would be a 10 or so line change and avoid having to deal with any of this and solve the original issue

pmatos: > Yeah - this is why I suggested just doing it statically, it's so awkward with all the edge cases when statically allocating the pred reg would be a 10 or so line change and avoid having to deal with any of this and solve the original issue

Yep! OK, lets take a step back here and let me refactor this for a statically allocated predicate register solution. 

Thanks for your patience and help on this one.

bylaws: For reference do see loadpf, you basically want to copy that exactly  (Inc ra changes) but taking in predicate init params

pmatos: Thanks! Will do that.

pmatos:
Closing in favour of #4292

PR #4284, opened 2025-01-20
CodeEmitter: Auto-format .inl headers

neobrain:
External tooling couldn't process these files before since there was no context, but with a little preprocessor hack this can be supported. I'm not too happy about needing this, but we only need it for these 8 files.

As a nice bonus, other tooling (IDEs in particular) can process these files as well now, allowing "go to definition", error highlighting, and other features to work out of the box.

Best reviewed per patch, since the second patch reformats 18.000 lines of code that hadn't been auto-formatted before.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -3,10 +3,13 @@
 # Ignore all files in the External directory
 External/*
 
-# SoftFloat-3e code doesn't belong to us 
+# SoftFloat-3e code doesn't belong to us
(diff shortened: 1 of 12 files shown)


Sonicadvance1 (approved):

neobrain:
(Only change is that I dropped misplaced `{`s from some comments. Will merge at EOD)

Sonicadvance1:
Needs rebase.

PR #4286, opened 2025-01-21
Drop assume-asserting logging macros

neobrain:
Replacing debug-only assertions with `__builtin_assume` in optimizing builds seems like a free performance-win at first, but it turns out to be a footgun with no practical upside: The binary size reduction I measured is 0.03% (a factor of 0.0003).

One big problem is that these hints interfere with debugging of non-assertion builds. Since the compiler generates code as-if the condition were true, a debugger will print wrong variable values or have erratic source stepping behavior. This can even apply to Debug builds, since FEX uses a dedicated `ASSERTIONS_ENABLED` macro.

Even [debug printfs](https://godbolt.org/z/5oox8jn6h) will print misleading information, which isn't obvious when your assumption is hidden in an innocent-looking macro that's barely distinguished from the non-hinting one.

As a future guideline, optimization hints should instead be placed *intentionally* after identifying bottlenecks in a profiler to communicate where hotspots are located and what invariants their efficient execution relies on.

### DIFF
--- a/CodeEmitter/CodeEmitter/ALUOps.inl
+++ b/CodeEmitter/CodeEmitter/ALUOps.inl
@@ -666,12 +666,12 @@ public:
   }
 
   void add(ARMEmitter::Size s, ARMEmitter::Register rd, ARMEmitter::Register rn, ARMEmitter::Register rm, ARMEmitter::ShiftType Shift = ARMEmitter::ShiftType::LSL, uint32_t amt = 0) {
-    LOGMAN_THROW_AA_FMT(Shift != ARMEmitter::ShiftType::ROR, "Doesn't support ROR");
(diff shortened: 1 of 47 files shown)


neobrain:
(For context, this came up during review of https://github.com/FEX-Emu/FEX/pull/4271#discussion_r1916863896, where I noticed how widely these macros had been applied.)

alyssarosenzweig:
+1 to this, I've never liked this distinction and miss just regular old asserts.

(I'd also like to rename this macro to something short and lowercase, probably `fex_assert`. But that's a follow-on discussion.)

Sonicadvance1 (approved):

pmatos:
> +1 to this, I've never liked this distinction and miss just regular old asserts.
> 
> (I'd also like to rename this macro to something short and lowercase, probably `fex_assert`. But that's a follow-on discussion.)

`fex_assert` - I like the sound of that.

PR #4288, opened 2025-01-21
Fix slight inaccuracy in test 3_F7_05_2

OFFTKP:
Needs cwd/cdq/cqo to sign extend into RDX/EAX/AX for imul to do what is described, for example in
```
; Negative * Negative
mov ax, -128
imul word [r8 + 8 * 0 + 0]
ofcfmerge
```

dx:ax is not negative, unless if we do `cwd`


Sonicadvance1 (approved):
Nice catch! Thanks for the improvement!

PR #4289, opened 2025-01-21
Pass through FPRs argument

pmatos:
At the moment this seems to be NFC since all calls to this function end up passing `true` but it seems that it's a bug waiting to happen if someone ever calls the function with `false`.


Sonicadvance1 (approved):
Oops

PR #4290, opened 2025-01-22
LinuxEmulation: Ensure syscall wrapper declaration has CpuStateFrame as the first argument

Sonicadvance1:
Otherwise crashes occur.


Sonicadvance1:
Added more words to the commit message with more meaningful meanings.

neobrain (approved):
Fixing an oversight in #4283 .

PR #4291, opened 2025-01-22
FEX: Implements new sampling based stats

Sonicadvance1:
The ftrace/gpuviz path has an issue with some usability. It can't be put in high-frequency locations like the SIGBUS handler, and it is hard to correlate between stutters in the game with a gpuviz trace after the fact.

With the combination of these two problems I have written this little SHM implementation of statistics that are /virtually/ free on the FEX side and requires the reader to accumulate the data correctly.

Some design decisions
- All atomics are weak-memory ordered, requiring the use of memory_barriers and mutexes in appropriate locations
  - This improves performance since it's avoid atomic overheads entirely (Which FEX already has issues with)
  - This means all stats are per-thread and the accumulation has to happen on the reader side
- The stats SHM region can never shrink, which would result in problems.
- The stats are in a singly linked-list because the expectation is low thread counts
- The stats are in a single allocation region which allows us to expand the stats if the number of threads grow

This is implemented for Linux and WINE, so we can use it everywhere.

This gets us some nice stat graphics like what follows:
Screenshot 2025-01-21 at 19 28 37

We only have four stats currently, but there can definitely be more stats as time progresses.
The Mangohud patches are forthcoming while I clean them up.


bylaws:
I'll make a fixup to replace the wine support with something simpler in a bit

Source/Common/Profiler.h
```diff
@@ -0,0 +1,67 @@
...
+    return (Offset - sizeof(FEXCore::Profiler::ThreadStatsHeader)) / sizeof(FEXCore::Profiler::ThreadStats);
+  }
+
+  void SaveHeader(FEXCore::Profiler::AppType AppType);
+
+  void* Base;
+  size_t CurrentSize {};
+  FEXCore::Profiler::ThreadStatsHeader* Head {};
+  FEXCore::Profiler::ThreadStats* Stats;
+  FEXCore::Profiler::ThreadStats* StatTail {};
```
neobrain, line 51: Why are you preferring a custom-implemented linked list over flat memory? This seems neither more convenient for the implementation nor like it would improve performance.

Source/Common/Profiler.cpp
```diff
@@ -0,0 +1,120 @@
...
+
+FEXCore::Profiler::ThreadStats* StatAllocBase::AllocateBaseSlot(uint32_t TID) {
+  if (!RemainingSlots) {
+    if (!AllocateMoreSlots()) {
+      return nullptr;
+    }
+  }
+
+  // Find a free slot
+  memory_barrier();
```
neobrain, line 50: Compared to the cost of the profiled events (SIGBUS/SMC), is it really worth dealing with the added complexity here instead of just using stronger atomics? Also, isn't thread creation/destruction in particular rare enough that we could use strong memory ordering on slot management variables while leaving the stats themselves weakly ordered?

neobrain (commented):
Gave this a first look - heavy stuff, but gathering and visualizing these metrics indeed is useful functionality to have.

Will mangohud need to maintain a fixed list of FEX-specific metrics (and update them if we add any) or will it automatically pick up new metrics added in FEX?

Is this designed to support accumulation of metrics across processes as well or is it solely for the process running mangohud?

The current implementation leaves me uncomfortable, since there's only so much faith one can have in reliability when interleaving weakly-ordered atomics, shared memory, and a custom linked list implementation. Well-placed and tested abstractions and more explaining docs would help support that faith.

More generally, can I ask what's the benefit of a MangoHud-specific approach compared to an established nanosecond-resolution profiler like Tracy? Obviously MangoHud is great for general stats like FPS and CPU load, but FEX's metrics are very specific and I'm not sure how valuable having a short window of these is. Most metrics could probably even be reported via ftrace despite its overhead (i.e. less rapidly updated metrics, or those that relate to costly SIGBUS/SMC events anyway), which could give us free multi-process awareness without bespoke accumulation logic in external projects.

FEXCore/include/FEXCore/Utils/Profiler.h
```diff
@@ -34,6 +93,36 @@ class ProfilerBlock final {
...
+    const auto Duration = GetCycleCounter() - Begin + FlatOffset;
+    if (Stat) {
+      auto ref = std::atomic_ref<T>(*Stat);
+      ref.fetch_add(Duration, std::memory_order_relaxed);
+    }
+  }
+
+private:
+  uint64_t Begin;
+  T* Stat;
```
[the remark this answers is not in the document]
Sonicadvance1: It's nullptr when profile stats are disabled.

Source/Common/CMakeLists.txt
```diff
@@ -7,7 +7,8 @@ set(SRCS
   EnvironmentLoader.cpp
   HostFeatures.cpp
   JSONPool.cpp
-  StringUtil.cpp)
+  StringUtil.cpp
+  Profiler.cpp)
```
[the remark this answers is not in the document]
Sonicadvance1: This is frontend code that is shared between FEXInterpreter, wow64, and arm64ec.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -4,21 +4,174 @@
...
+  if (CurrentSize == 0) {
+    CurrentSize = 4096;
+  }
+
+  if (ftruncate(fd, CurrentSize) == -1) {
+    LogMan::Msg::EFmt("[StatAlloc] ftruncate failed");
+    goto err;
+  }
+
+  // 128MB ought to be enough for anyone.
```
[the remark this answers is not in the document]
Sonicadvance1: Updated.

Source/Common/Profiler.h
```diff
@@ -0,0 +1,67 @@
...
+#else
+static inline void memory_barrier() {
+  // Intentionally empty.
+}
+#endif
+
+namespace FEX::Profiler {
+class StatAllocBase {
+public:
+  virtual ~StatAllocBase() = default;
```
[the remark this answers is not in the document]
Sonicadvance1: Removed.

FEXCore/include/FEXCore/Debug/InternalThreadState.h
```diff
@@ -95,6 +99,8 @@ struct InternalThreadState : public FEXCore::Allocator::FEXAllocOperators {
 
   std::shared_mutex ObjectCacheRefCounter {};
 
+  FEXCore::Profiler::ThreadStats* ThreadStats {};
```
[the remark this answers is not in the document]
Sonicadvance1: Added comment.

FEXCore/include/FEXCore/Utils/Profiler.h
```diff
@@ -1,13 +1,72 @@
...
+  // Accumulated event counts
+  uint64_t AccumulatedSIGBUSCount;
+  uint64_t AccumulatedSMCCount;
+};
+
 #ifdef ENABLE_FEXCORE_PROFILER
 
+#ifdef _M_ARM_64
+/**
+ * @brief Get the raw cycle counter which is synchronizing.
```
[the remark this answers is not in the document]
Sonicadvance1: Updated comment.

Source/Common/Profiler.h
```diff
@@ -0,0 +1,67 @@
...
+  // I (Sonicadvance1) wanted to reserve 128MB of VA space because it's cheap, but ran in to a bug when running WINE.
+  // WINE allocates [0x7fff'fe00'0000, 0x7fff'ffff'0000) which /consistently/ overlaps with FEX's sigaltstack.
+  // This only occurs when this stat allocation size is large as the top-down allocation pushes the alt-stack further.
+  // Additionally, only occurs on 48-bit VA systems, as mmap on lesser VA will fail regardless.
+  // TODO: Bump allocation size up once FEXCore's allocator can first use the 128TB of blocked VA space on 48-bit systems.
+  constexpr static size_t MAX_STATS_SIZE = 4 * 1024 * 1024;
+
+private:
+  virtual uint64_t AllocateMoreSlots(uint64_t NewSize) = 0;
+  bool AllocateMoreSlots();
```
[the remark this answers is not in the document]
Sonicadvance1: Changed name,

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -4,21 +4,174 @@
...
+  }
+
+err:
+  close(fd);
+}
+
+uint64_t ThreadManager::StatAlloc::AllocateMoreSlots(uint64_t NewSize) {
+  if (CurrentSize == MAX_STATS_SIZE) {
+    // Nope.
+    return CurrentSize;
```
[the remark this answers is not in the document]
Sonicadvance1: Updated comment.
When the frontend can't allocate more slots then new threads will be given nullptr for stats, which means they don't get tracked, which is safe.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -4,21 +4,174 @@
...
+  }
+
+  {
+    auto SharedBase = ::mmap(Base, NewSize, PROT_READ | PROT_WRITE, MAP_SHARED | MAP_FIXED, fd, 0);
+    if (SharedBase == MAP_FAILED) {
+      LogMan::Msg::EFmt("[StatAlloc] allocate more mmap shm failed");
+      goto err;
+    }
+
+    // TODO: Just a sanity check.
```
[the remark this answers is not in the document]
Sonicadvance1: Removed.

Source/Common/Profiler.cpp
Sonicadvance1: False dependency sharing in the hot path is a significant concern, so I'm keeping them as weak atomics.

Source/Common/Profiler.h
Sonicadvance1: Sparsity in the allocation buffer is a concern, so this is how I'm keeping the implementation.

Source/Common/Profiler.cpp
neobrain: Where's the hot path? The currently traced events are orders of magnitude more expensive to process than a potential cache miss.

This applies even more to thread creation/destruction.

Source/Common/Profiler.h
neobrain: Could you elaborate what exactly the problem is?

Source/Common/CMakeLists.txt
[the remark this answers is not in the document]
neobrain: Why is *this* frontend code but the existing Profiler infrastructure in FEXCore not, despite also being Linux specific?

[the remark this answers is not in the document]
neobrain: I guess my problem is that the 6 files (if I'm counting correctly) are placed in seemingly random locations: The existing Linux-specific code probably shouldn't be in FEXCore since the Windows-specific equivalent isn't there either.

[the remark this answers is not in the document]
Sonicadvance1: That code should also get moved to the frontend at some point as long as the `TraceObject` handlers can still get called from FEXCore.

Source/Common/Profiler.h
neobrain: What scenario exactly would trigger this? I don't know what your reader side looks like, but it sounds like a bitmask and/or active slot count would easily resolve this.

Sonicadvance1: I'm just going to leave it as a linked list. We can rev the version number and change it to a bitmask in the future if someone is feeling up to writing that code in the future.

neobrain: If the choice between linked lists and flat memory doesn't seem to be so critical, why is the focus on weakly-ordered atomics considered crucial?

Sonicadvance1: For the atomics I don't want the writer side to be affected by cacheline sharing problems at all. Semantically these are two independent things. The containing data which is `FEXCore::Profiler::ThreadStats` and the way to walk the list of data.

neobrain: I understand they are different things, but it seems arbitrary to want this (with no explicit motivation) while using a data structure with poor performance characteristics in the same module.

Sonicadvance1: Indeed, which is why we leave some of the "poor performance" problems on the reader side which has something like 500ms to sample the data, even though this is going to be hundreds of nanoseconds regardless.

Sonicadvance1:
https://github.com/Sonicadvance1/MangoHud/tree/fex
For the WIP FEX support in mangohud.

Source/Common/Profiler.h
```diff
@@ -0,0 +1,69 @@
...
+#pragma once
+#include <FEXCore/Utils/Profiler.h>
+
+namespace FEXCore::Core {
+struct InternalThreadState;
+}
+
+#ifdef _M_ARM_64
+static inline void memory_barrier() {
+  asm volatile("dmb ishst;" ::: "memory");
```
bylaws, line 17: can atomic_thread_fence not be used here? instead of inline asm

Sonicadvance1: That helper can't actually generate a `dmb ishst`, it can only generate `dmb ish` or `dmb ishld`. Sadly.

bylaws: ah, this should ideally be called something more explicit then

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -4,21 +4,171 @@
...
+  SaveHeader(Is64BitMode() ? FEXCore::Profiler::AppType::LINUX_64 : FEXCore::Profiler::AppType::LINUX_32);
+}
+
+void ThreadManager::StatAlloc::Initialize() {
+  if (!ProfileStats()) {
+    return;
+  }
+
+  int fd = shm_open(fextl::fmt::format("fex-{}-stats", ::getpid()).c_str(), O_CREAT | O_TRUNC | O_RDWR, USER_PERMS);
+  if (!fd) {
```
bylaws, line 28: fd == -1

Source/Windows/WOW64/Module.cpp
```diff
@@ -499,9 +501,17 @@ void BTCpuProcessInit() {
...
+  FEX_CONFIG_OPT(ProfileStats, PROFILESTATS);
+
+  if (IsWine && ProfileStats()) {
+    StatAllocHandler = fextl::make_unique<FEX::Windows::StatAlloc>(FEXCore::Profiler::AppType::WIN_WOW64);
+  }
 }
 
-void BTCpuProcessTerm(HANDLE Handle, BOOL After, ULONG Status) {}
+void BTCpuProcessTerm(HANDLE Handle, BOOL After, ULONG Status) {
+  StatAllocHandler.reset();
```
bylaws, line 513: Can be dropped

Source/Windows/ARM64EC/Module.cpp
```diff
@@ -567,10 +571,17 @@ NTSTATUS ProcessInit() {
...
+
+  if (IsWine && ProfileStats()) {
+    StatAllocHandler = fextl::make_unique<FEX::Windows::StatAlloc>(FEXCore::Profiler::AppType::WIN_ARM64EC);
+  }
   return STATUS_SUCCESS;
 }
 
-void ProcessTerm(HANDLE Handle, BOOL After, NTSTATUS Status) {}
+void ProcessTerm(HANDLE Handle, BOOL After, NTSTATUS Status) {
+  StatAllocHandler.reset();
```
bylaws, line 583: Can be dropped

Source/Windows/WOW64/Module.cpp
```diff
@@ -519,6 +533,8 @@ void BTCpuThreadTerm(HANDLE Thread, LONG ExitCode) {
     return;
   }
 
+  auto* OldThreadState = TLS.ThreadState();
```
bylaws, line 536:
```suggestion
  auto* ThreadState = TLS.ThreadState();
```

Source/Common/Profiler.h
Sonicadvance1: renamed to `store_memory_barrier`

Source/Windows/ARM64EC/Module.cpp
bylaws: I think it can be done on the wine side on exit with my patch - but there's a lot of crashing cases where a process can exit without any exit handlers being called in general so it's probably best if mangohud can clean these up or something? It would be very nice to avoid having sigkill leak memory

bylaws: I'm more just concerned that this could be enabled by people that don't have something in place to cleanup afterwards, nothing in the option description suggests it's really quite likely to leak resources (a lot of things will launch subprocesses that end up crashing but work regardless) seems like something fexserver could do actually?

Sonicadvance1: Potentially FEXServer could do it on the Linux side, but since arm64ec/wow64 fex doesn't use FEXServer, it wouldn't work there.

bylaws: I think in the case you already have a Linux fex installation it might be fine? Though yeah it would have to poll shm files

Source/Common/Profiler.cpp
Sonicadvance1: Yes, we have some leniency on thread creation and teardown, which is already slow and causes global locks to be taken, but this is the way I have decided to implement it.

bylaws:
I would err towards only supporting the wine side with the fex patch without a solid usecase

Sonicadvance1:
> I would err towards only supporting the wine side with the fex patch without a solid usecase for the questionable path



> I would err towards only supporting the wine side with the fex patch without a solid usecase for the questionable path

Sure, ripped it out.

Source/Common/Profiler.h
```diff
@@ -0,0 +1,69 @@
...
+    return reinterpret_cast<uint64_t>(Stat) - reinterpret_cast<uint64_t>(Base);
+  }
+  uint32_t TotalSlotsFromSize() const {
+    return (CurrentSize - sizeof(FEXCore::Profiler::ThreadStatsHeader)) / sizeof(FEXCore::Profiler::ThreadStats) - 1;
+  }
+  uint32_t TotalSlotsFromSize(uint32_t Size) const {
+    return (Size - sizeof(FEXCore::Profiler::ThreadStatsHeader)) / sizeof(FEXCore::Profiler::ThreadStats) - 1;
+  }
+
+  uint32_t SlotIndexFromOffset(uint32_t Offset) {
```
lioncash, line 43: Should this be const or static, like the above function?

lioncash (commented):
Just one super small nit

Source/Common/Profiler.h
Sonicadvance1: Good catch, both this and the above function can be static. Switched over.

lioncash (approved):

PR #4292, opened 2025-01-22
Predicate cache alternative implementation

### LINKED ISSUE #4264: Cached predicate register being clobbered by softfp lib call

pmatos:
Alternative implementation to #4274 

Fixes #4264

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -1,7 +1,6 @@
 // SPDX-License-Identifier: MIT
 #include "Interface/Core/ArchHelpers/Arm64Emitter.h"
 #include "FEXCore/Core/X86Enums.h"
-#include "FEXCore/Utils/AllocatorHooks.h"
 #include "Interface/Core/Dispatcher/Dispatcher.h"
 #include "Interface/Context/Context.h"
 
@@ -60,8 +59,8 @@ namespace x64 {
   // p6 and p7 registers are used as temporaries no not added here for RA
   // See PREF_TMP_16B and PREF_TMP_32B
   // p0-p1 are also used in the jit as temps.
-  // Also p8-p15 cannot be used can only encode p0-p7, so we're left with p2-p5.
-  constexpr std::array<ARMEmitter::PRegister, 4> PR = {ARMEmitter::PReg::p2, ARMEmitter::PReg::p3, ARMEmitter::PReg::p4, ARMEmitter::PReg::p5};
+  // Also p8-p15 cannot be used can only encode p0-p7, p2 is a special register, so we're left with p3-p5.
+  constexpr std::array<ARMEmitter::PRegister, 3> PR = {ARMEmitter::PReg::p3, ARMEmitter::PReg::p4, ARMEmitter::PReg::p5};
 
   constexpr unsigned RAPairs = 6;
 
@@ -100,6 +99,7 @@ namespace x64 {
     ARMEmitter::Reg::r20,
     ARMEmitter::Reg::r21,
     ARMEmitter::Reg::r22,
+    // PF/AF must be last.
     REG_PF,
     REG_AF,
   };
@@ -112,8 +112,8 @@ namespace x64 {
   // p6 and p7 registers are used as temporaries no not added here for RA
   // See PREF_TMP_16B and PREF_TMP_32B
   // p0-p1 are also used in the jit as temps.
-  // Also p8-p15 cannot be used can only encode p0-p7, so we're left with p2-p5.
-  constexpr std::array<ARMEmitter::PRegister, 4> PR = {ARMEmitter::PReg::p2, ARMEmitter::PReg::p3, ARMEmitter::PReg::p4, ARMEmitter::PReg::p5};
+  // Also p8-p15 cannot be used can only encode p0-p7, p2 is a special register, so we're left with p3-p5.
+  constexpr std::array<ARMEmitter::PRegister, 3> PR = {ARMEmitter::PReg::p3, ARMEmitter::PReg::p4, ARMEmitter::PReg::p5};
 
   constexpr unsigned RAPairs = 6;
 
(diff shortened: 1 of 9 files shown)


FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -822,6 +824,10 @@ void Arm64Emitter::FillStaticRegs(bool FPRs, uint32_t GPRFillMask, uint32_t FPRF
     }
   }
 
+  if (FillP2) {
```
bylaws, line 827: Perhaps you can drop this and change to a simple sve supported check, as is it'll not work for the FillSRA in dispatcher and a single ptrue is cheap enough not to matter

pmatos: But if we always fill `p2` when SVE is supported this whole optimization makes no longer sense from a size pov. If we do it without actually needing to use `p2` for the SVE store optimization then we are definitely pessimizing things.

bylaws: Spills are rare enough that a ptrue certainly isn't going to change anything, the opt still works within code blocks which is where it matters as they can be much botter

pmatos: Humm, I still didn't manage to get SRA to work. The RA algo is still allocating the wrong register to the predicate register, i.e. p5. It should be p2, which is the SRA register I selected to use in DecodeSRAReg but it's not working. First time I am looking at this code so slightly confused about this.

At the same time, i am wondering if this is necessary. What's the advantage of using SRA over just hardcoding p2 usage in the instructions, like done with `PRED_TMP_16B` for example?

bylaws: > Humm, I still didn't manage to get SRA to work. The RA algo is still allocating the wrong register to the predicate register, i.e. p5. It should be p2, which is the SRA register I selected to use in DecodeSRAReg but it's not working. First time I am looking at this code so slightly confused about this.
> 
> At the same time, i am wondering if this is necessary. What's the advantage of using SRA over just hardcoding p2 usage in the instructions, like done with `PRED_TMP_16B` for example?

I don't think there are any, that would work fine too you'd just need to modify the store op to have some way to indicate that and you can drop LoadSVEOptPredicate and rely on it being set in FillSpecialRegs as PRED_TMP_16B is (you probably want to keep InitPredicate for other future users)

bylaws: (can drop this as it's called in FillSpecialRegs)

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
```diff
@@ -1631,9 +1631,13 @@ DEF_OP(StoreMemPredicate) {
 DEF_OP(LoadMemPredicate) {
```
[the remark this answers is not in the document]
pmatos: Right ... done. To be honest, I am thinking that the work done to enable RA of predicate registers can be reverted. I see no reason to keep code in FEX that doesn't have any users and this is this was the only user of that code.

@Sonicadvance1 what do you think about reverting the predicate register RA?

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
pmatos: Right - thanks.

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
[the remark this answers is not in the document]
Sonicadvance1: Sure, sounds reasonable to me.

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -644,6 +644,10 @@ void Arm64Emitter::FillSpecialRegs(ARMEmitter::Register TmpReg, ARMEmitter::Regi
       ptrue(ARMEmitter::SubRegSize::i8Bit, PRED_TMP_16B, ARMEmitter::PredicatePattern::SVE_VL16);
     }
   }
+
+  if (FillP2) {
+    ptrue(ARMEmitter::SubRegSize::i16Bit, PRED_X87_SVEOPT, ARMEmitter::PredicatePattern::SVE_VL5);
+  }
```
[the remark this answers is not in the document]
pmatos: SetPreds and FillP2 are not really equivalent. We only need to fill P2 if we really need the optimization enabled. This is not true in non-SVE cabable systems or if we don't have X87 ldst instructions in the block.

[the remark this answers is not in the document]
pmatos: Do I understand correctly that suggest I move it into the `SetPredRegs` block? I have done that. I haven't dropped the FillP2 conditional. That is necessary to only generate this when strictly necessary.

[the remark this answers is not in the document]
bylaws: So for the conditional right, the issue is that some helper routines are emitted as part of the dispatcher such as LUDiv. These won't be emitted with FillP2 set to true but they may still be called from a block that has it set as such and clobber the registers. For simplicities sake I think filling it unconditionally is the best solution to this, spills are rare in general outside of x87 code (which will likely have FillP2 as true anyway) so I don't see always setting it having a measurable performance impact. 

Also occurred to me that some control flow patterns with multiblock could break this, so that's another reason for just always filling:
loop:
cpuid
dec r0
jz r0 end
x87 write
jmp loop
end: x87 write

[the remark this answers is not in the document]
pmatos: I have now updated the patch to drop `FillP2` and always fill.

FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
```diff
@@ -1628,13 +1623,14 @@ DEF_OP(StoreMemPredicate) {
...
+DEF_OP(LoadMemX87SVEOptPredicate) {
+  const auto Op = IROp->C<IR::IROp_LoadMemX87SVEOptPredicate>();
   const auto Dst = GetVReg(Node);
-  const auto Predicate = GetPReg(Op->Mask.ID());
+  const auto Predicate = PRED_X87_SVEOPT;
   const auto MemReg = GetReg(Op->Addr.ID());
 
-  LOGMAN_THROW_A_FMT(HostSupportsSVE128 || HostSupportsSVE256, "LoadMemPredicate needs SVE support");
+  LOGMAN_THROW_A_FMT(HostSupportsSVE128 || HostSupportsSVE256, "LoadMemX87SVEOptPredicate needs SVE support");
+  ptrue(ARMEmitter::SubRegSize::i16Bit, Predicate, ARMEmitter::PredicatePattern::SVE_VL5);
```
bylaws, line 1633: I think you need to remove this otherwise the patch will do nothing :) (which is fine as it'll always be set in FillSRA)

pmatos: *sigh* of course.

Sonicadvance1 (approved):
Looks fine to me, even if this PR doesn't remove the Predicate register RA.

PR #4297, opened 2025-01-23
Update upload-artifact action to v4

pmatos:

PR #4298, opened 2025-01-23
Library Forwarding/wayland: Fix regression caused by erroneous format

neobrain:
`clang-format` tried a bit too hard to help here.

Before:
```cpp
LOAD_LIB_INIT(libwayland - client, OnInit)
```

After:
```cpp
LOAD_LIB_INIT(libwayland-client, OnInit)
```

This caused FEX to search for a file named `libwayland - client-host.so` instead of `libwayland-client-host.so`, hence failing to map the library to its host equivalent.

### DIFF
--- a/ThunkLibs/libwayland-client/Guest.cpp
+++ b/ThunkLibs/libwayland-client/Guest.cpp
@@ -337,4 +337,5 @@ void OnInit() {
   fex_wl_exchange_interface_pointer(const_cast<wl_interface*>(&wl_callback_interface), "wl_callback_interface");
 }
 
-LOAD_LIB_INIT(libwayland - client, OnInit)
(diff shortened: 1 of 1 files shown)


pmatos:
lgtm - wonder how the latest versions of clang-format do.

pmatos:
the failures are just because we need to merge https://github.com/FEX-Emu/FEX/pull/4297 first.

neobrain:
> lgtm - wonder how the latest versions of clang-format do.

19.1.1 is no better, sadly. I don't blame it though, since in a macro there's no way to distinguish the string-ified `wayland-client` symbol from the difference of two variables named `wayland` and `client`.

Sonicadvance1 (approved):

PR #4299, opened 2025-01-23
FEX: Allocate a VMA allocator when running on a 48-bit VA

Sonicadvance1:
When running on a system with a 48-bit VA, if FEX does any allocations between us reserving the upper 128TB and the application running, then /technically/ we are intersecting with the application's memory region in the lower 47-bits.

This didn't typically result in any problems due to how ASLR works, but if we did any large allocations (like #4291 wants with 128MB VMA region) then these typically get pushed higher in the VA space.

Again not usually a problem, but if you happen to be running an application that is using MAP_FIXED with hardcoded addresses then this can stomp over FEX-Emu memory causing problems.

This is what happens with Wine, it reserves the upper-32MB of its 47-bit VA space, which is /highly/ likely to stomp on FEX memory. In-fact it likely occurs all the time, we just got lucky with whatever it was clobbering wasn't used at the time.

On 39-bit VA systems this isn't a problem because the mmap fails outright with a warning message from WINE.

Because we are already reserving the upper 128TB of VA space, instead just always enable our allocator and use the regions that were reserved. We need to be a little bit careful to ensure we don't accidentally allocate more memory post-reservation but that just requires a small adjustment to our unique_ptr and constructor for the 64BitAllocator.

This means /all/ FEX-Emu allocations will be in the upper 128TB VA space when running 64-bit applications on a 48-bit VA system. Which is kind of nice.

Fixes WINE in #4291 when the allocator stats are bumped to 128MB per process.

### DIFF
--- a/FEXCore/Source/Utils/Allocator/FlexBitSet.h
+++ b/FEXCore/Source/Utils/Allocator/FlexBitSet.h
@@ -145,8 +145,14 @@ struct FlexBitSet final {
     return Get(Element);
   }
 
-  static size_t Size(uint64_t Elements) {
-    return FEXCore::AlignUp(Elements / MinimumSizeBits, MinimumSizeBits);
+  // Returns the number of bits required to hold the number of elements.
+  // Just rounds up to the MinimumSizeInBits.
+  constexpr static size_t SizeInBits(uint64_t Elements) {
+    return FEXCore::AlignUp(Elements, MinimumSizeBits);
+  }
+  // Returns the number of bytes required to hold the number of elements.
+  constexpr static size_t SizeInBytes(uint64_t Elements) {
+    return SizeInBits(Elements) / 8;
   }
 };
 
--- a/FEXCore/Source/Utils/Allocator.cpp
+++ b/FEXCore/Source/Utils/Allocator.cpp
@@ -112,12 +112,16 @@ void ReenableSBRKAllocations(void* Ptr) {
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-void SetupHooks() {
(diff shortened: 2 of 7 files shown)


Sonicadvance1:
Had to do a tertiary bug fix in our Allocator that we hadn't hit before. This never occurred on the 32-bit side where we reserve the upper 32-bits because the first VMA region would be hundreds of gigabytes. Since we're using the regions in the top 128TB on 48-bit systems now, due to how ASLR is laid out (roughly starts in the middle of the VA) there was a pretty high chance of the /first/ VMA region being small enough to hit the bug.

FEXCore/Source/Utils/Allocator/FlexBitSet.h
```diff
@@ -145,8 +145,14 @@ struct FlexBitSet final {
...
 
-  static size_t Size(uint64_t Elements) {
-    return FEXCore::AlignUp(Elements / MinimumSizeBits, MinimumSizeBits);
+  // Returns the number of bits required to hold the number of elements.
+  // Just rounds up to the MinimumSizeInBits.
+  constexpr static size_t SizeInBits(uint64_t Elements) {
+    return FEXCore::AlignUp(Elements, MinimumSizeBits);
+  }
+  // Returns the number of bytes required to hold the number of elements.
+  static size_t SizeInBytes(uint64_t Elements) {
```
lioncash, line 154: Shouldn't this one also be constexpr, since the one above is?

lioncash (commented):
Just one minor nit

FEXCore/Source/Utils/Allocator/FlexBitSet.h
Sonicadvance1: Fixed

lioncash (approved):

PR #4303, opened 2025-01-27
CodeEmitter: Fixes clang_format

Sonicadvance1:


lioncash (approved):

neobrain:
How did this slip through CI before?

Sonicadvance1:
> How did this slip through CI before?

As far as I can tell, slightly different versions of clang-16 giving different results that are still compatible with each other.

PR #4309, opened 2025-01-28
Skip 3DNow tests with precision issues

### LINKED ISSUE #4280: Some 3DNow tests are broken on Oryon
Specifically, the following:

```
Test_3DNow/86.asm
Test_3DNow/87.asm
Test_3DNow/96.asm
Test_3DNow/97.asm
```

According to @Sonicadvance1:
> Yea, this test failure is expected on FEAT_RPRES supporting devices. We need to add some ability to our ASM tests to know how to they are safe to be off by 1-bit of precision. Or just run with `FEX_HOSTFEATURES=disableafp,disablerpres`

pmatos:
Fixes #4280

### DIFF
--- a/unittests/ASM/Disabled_Tests
+++ b/unittests/ASM/Disabled_Tests
@@ -17,7 +17,6 @@ Test_Primary/Primary_E9.asm
 
 # Relies on undefined behaviour
 Test_X87/D9_F9.asm
-
 Test_X87/D9_F2.asm
 
 # Relies on rounding correctness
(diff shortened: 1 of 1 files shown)


unittests/ASM/Disabled_Tests
```diff
@@ -17,11 +17,16 @@ Test_Primary/Primary_E9.asm
...
-
 Test_X87/D9_F2.asm
 
 # Relies on rounding correctness
 Test_X87/D9_F8.asm
 
 # This is basically a benchmark.
 FEX_bugs/XeSS_quadratic.asm
+
+# 3DNow reciprocal not providing enough precision
```
Sonicadvance1, line 28: This implies we aren't emulating enough precision. Should probably be a comment that the level of precision is different between rpres hardware and not, both are correct but one fails CI.

pmatos: Comment clarified.

Sonicadvance1 (approved):

PR #4310, opened 2025-01-29
Wine: Properly Store and Restore TF, DF, and AF before clear

MelodicAlbuild:
In it's current implementation, FEX discards TF, DF, and AF when throwing guest exceptions. These bits are restored by Windows after an exception, and therefore needs to be saved by FEX before clear.

This PR is a Draft while I work on the Wine Patch and testing the implementation with Wine.
_This is not the final version of these changes as I plan on making it a bit cleaner and more performant_

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -118,6 +118,11 @@ struct ThreadCPUArea {
   }
 };
 
+/* Reserved CPSR bits for emulated x86 flags */
+#define CPSR_X86_TF (1u << 25) /* Trap flag */
+#define CPSR_X86_DF (1u << 26) /* Direction flag */
+#define CPSR_X86_AF (1u << 27) /* Auxiliary carry flag */
+
 namespace {
 fextl::unique_ptr<FEXCore::Context::Context> CTX;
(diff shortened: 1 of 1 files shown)


bylaws:
TF is already handled here so that part can be dropped. Is the an application you have ran into depending on af/df being restored?

bylaws:
I.e. this patch should really only be a thrre line change, setting https://github.com/FEX-Emu/FEX/blob/fe8d9b046989965e398a1d269682bdcfeb6badf3/Source/Windows/ARM64EC/Module.cpp#L429 with a commit this is specific behaviour for a downstream wine patch and adding to https://github.com/FEX-Emu/FEX/blob/fe8d9b046989965e398a1d269682bdcfeb6badf3/Source/Windows/ARM64EC/Module.cpp#L532

MelodicAlbuild:
Didn't realize TF had already been handled, I saw the TODO and figured there hadn't been any implementation yet. 

Currently running into the issue when running Photoshop on top of Wine, specifically when using long chain actions and an action fails. The AF flag specifically is used here _Unsure why_ and is cleared when the action failed, then the application crashes after the fail is caught because it tries to resume, but AF is now 0, and it fails out.

neobrain:
What's the way forward here? Is the 3 line change suggested by ByLaws worth integrating?

If it is, let's do it, otherwise let's not keep this PR lingering around without a clear path forward.

PR #4313, opened 2025-01-30
Add option StartupSleepProcName

pmatos:
Sleeps only if current process matches this name. Leave empty to sleep StartupSleep seconds on all processes.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -472,6 +472,13 @@
           "Sleeps the process at startup for a duration of seconds.",
           "Useful if an application crashes too quickly to attach a debugger."
         ]
+      },
+      "StartupSleepProcName": {
+        "Type": "str",
+        "Default": "",
+        "Desc": [
+          "Contrains the startup sleep to only apply to processes that match this name."
+        ]
       }
     },
     "Misc": {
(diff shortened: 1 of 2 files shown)


pmatos:
When starting steam, many processes tend to start. If I just want to target a specific process by name, I couldn't find a way to do it. This has been helpful to sleep only on a process with the given name.

Sonicadvance1 (approved):
I usually use an app profile for this, but sure.
```
ryanh@ubuntu:~/.fex-emu/AppConfig$ cat sekiro.exe.json
{
  "ThunksDB": {
    "GL": 1,
    "Vulkan": 1
  },
  "Config":{
    "StartupSleep":"15"
  }
}
```

Sonicadvance1:
> > I usually use an app profile for this, but sure.
> > ```
> > ryanh@ubuntu:~/.fex-emu/AppConfig$ cat sekiro.exe.json
> > {
> >   "ThunksDB": {
> >     "GL": 1,
> >     "Vulkan": 1
> >   },
> >   "Config":{
> >     "StartupSleep":"15"
> >   }
> > }
> > ```
> 
> So what connects this config file to the process is that the prefix of the filename is the procname? So the configuration file above will take precedence over all other options for `sekiro.exe` process?

Yes, the filename of the json is the procname. So this config file ends up in a higher priority layer as per https://wiki.fex-emu.com/index.php/Config#Loadable_layers

PR #4314, opened 2025-01-30
Arm64: Fix bitmask used to match load/store instructions

neobrain:
When multiple threads simultaneously SIGBUS on the same address, one of them will perform the backpatching while the other will detect the backpatched instruction sequence and hence report the SIGBUS as "handled".

This typo broke the instruction detection logic: The second thread would assume the source of the SIGBUS was unrelated to TSO emulation and hence report the signal as unhandled (generally triggering program abortion).

In practice, this problem did not manifest as FEX does not currently share CodeBuffers between threads. With CodeBuffer sharing, God of War crashed early during startup due to spawning worker threads in quick succession that each SIGBUS on the same address.

References:
* Encoding for [LDUR (load register unscaled)](https://developer.arm.com/documentation/ddi0602/2024-12/Base-Instructions/LDUR--Load-register--unscaled--)
* Encoding for [STUR (store register unscaled)](https://developer.arm.com/documentation/ddi0602/2024-12/Base-Instructions/STUR--Store-register--unscaled--)


Sonicadvance1 (approved):
oops

PR #4324, opened 2025-02-03
FEXCore/JIT: Pass Softfloat arguments as vector registers 

Sonicadvance1:
This is preparation work to allow passing the corestate to the x87 soft
float handlers directly for some profile stats.

Performance-wise, this change falls within noise because it basically
moves the GPR->Vector moves from the JIT in to C code, my microbench saw
the largest excursion of 5% but that's still within noise in the current
design of my bench.

A more tangible win from this change alone is less codegen on the JIT
side.


FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -377,17 +368,14 @@ void Arm64JITCore::Op_Unhandled(const IR::IROp_Header* IROp, IR::NodeID Node) {
       const auto Src2 = GetVReg(IROp->Args[1].ID());
 
       ldrh(ARMEmitter::WReg::w0, STATE, offsetof(FEXCore::Core::CPUState, FCW));
-      umov<ARMEmitter::SubRegSize::i64Bit>(ARMEmitter::Reg::r1, Src1, 0);
-      umov<ARMEmitter::SubRegSize::i16Bit>(ARMEmitter::Reg::r2, Src1, 4);
+      mov(ARMEmitter::VReg::v0.Q(), Src1.Q());
+      mov(ARMEmitter::VReg::v1.Q(), Src2.Q());
```
bylaws, line 372: This isn't valid on arm64ec, Src2 could be allocated to v0 - you'll need to copy to vtmp before hand as is done for some others with tmp

bylaws (changes requested):

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -85,17 +86,16 @@ void Arm64JITCore::Op_Unhandled(const IR::IROp_Header* IROp, IR::NodeID Node) {
 #endif
   } else {
     auto FillF80Result = [&]() {
+      auto ResultReg = ARMEmitter::VReg::v0;
```
bylaws, line 89: You can just use VTMP1 always here, the point of TMP_ABIARGS is to explicitly indicate that such is valid

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -331,17 +326,14 @@ void Arm64JITCore::Op_Unhandled(const IR::IROp_Header* IROp, IR::NodeID Node) {
       const auto Src2 = GetVReg(IROp->Args[1].ID());
 
       ldrh(ARMEmitter::WReg::w0, STATE, offsetof(FEXCore::Core::CPUState, FCW));
-      umov<ARMEmitter::SubRegSize::i64Bit>(ARMEmitter::Reg::r1, Src1, 0);
-      umov<ARMEmitter::SubRegSize::i16Bit>(ARMEmitter::Reg::r2, Src1, 4);
-
-      umov<ARMEmitter::SubRegSize::i64Bit>(ARMEmitter::Reg::r3, Src2, 0);
-      umov<ARMEmitter::SubRegSize::i16Bit>(ARMEmitter::Reg::r4, Src2, 4);
+      mov(ARMEmitter::VReg::v0.Q(), Src1.Q());
```
bylaws, line 329: See below review comment

Sonicadvance1:
The arm64ec side is now fixed. pretty sure we could be smarter to check intersections and only does tmp moves when necessary, but since the original code didn't try to save a couple moves, I also decided not to.

Sonicadvance1:
Ping for review.

bylaws (approved):

lioncash (approved):

PR #4325, opened 2025-02-04
FEXCore: Don't copy IR after compilation

bylaws:
This wastes a significant amount of compilation time (30%!!) when the copies IR is promptly thrown away after compilation anyway.

Breaks AOTIR - this is already broken and removal is planned so shouldn't be an issue.

Stats (mean over 3*8 runs of nodejs):
With: 2419065553ns
Without:  3307522698ns


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -700,7 +666,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
         if (HadDispatchError && TotalInstructions == 0) {
           // Couldn't handle any instruction in op dispatcher
           Thread->OpDispatcher->ResetWorkingList();
-          return {nullptr, 0, 0, 0, 0};
+          return {{}, nullptr, 0, 0, 0, 0};
```
Sonicadvance1, line 669: Should the `{}` be `std::nullopt` instead?

Sonicadvance1 (approved):

PR #4326, opened 2025-02-04
Frontend: Split blocks at jump target boundaries

bylaws:
With the prior approach, backwards jumps into existing blocks would explore the overlapping part rather than splitting the block, generating needless code and wasting time decoding. Similarly, the current block wouldn't be split when it is extended to overlap with a pending jump target.

Solve this by tracking blocks in a sorted vector and splitting existing blocks on jumps when appropriate, in order to avoid any possibility of overlapping blocks, which would break the lookup, misaligned and zero instruction blocks are disallowed.

When exploring, keep track of the next block/jmp target following the current to know when to split or end the block (note the special case of A rejoining a block B but misaligned, A is dropped at such point but added to VisitedBlocks so future jumps there don't bother trying to explore it)

Stats (mean of 3*4 nodejs runs):
Before: 3790969649ns
After: 3579219306ns

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -1100,10 +1100,10 @@ void Decoder::DecodeInstructionsAtEntry(const uint8_t* _InstStream, uint64_t PC,
 
     while (1) {
       // MAX_INST_SIZE assumes worst case
-      auto OpMinAddress = RIPToDecode + PCOffset;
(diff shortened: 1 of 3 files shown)


bylaws:
DNM - after fixing some other bugs this breaks some games

bylaws:
Denuvo games have magically started working - perhaps leave this a bit for testing but I suppose merging makes sense

Sonicadvance1 (approved):
Did some testing. Let's do it.

PR #4327, opened 2025-02-04
Config: Enable multiblock by default

bylaws:
Dep #4325 #4326 

Before these:
MB on: 3790969649ns
MB off: 3169002331ns
After:
MB on: 2594371996ns
MB off: 2386604655ns

Seems reasonable to take the 10% hit here


bylaws:
> Supposedly Factorio breaks under Multiblock. I haven't tried to reproduce yet, but might not be safe to enable by default yet.

Quite likely this was tested with fex before my fixes

Sonicadvance1 (approved):
Alright, let's get this in for a month of dogfooding.

PR #4332, opened 2025-02-06
ARM64EC: Set EC_ENTRY_CPUAREA_REG at inline SMC dispatcher entry

bylaws:
Was set by pure luck before


Sonicadvance1 (commented):
Spicy!

Sonicadvance1 (approved):

PR #4333, opened 2025-02-06
Async: Add framework for multiplexing IO on network sockets and other file descriptors

neobrain:
## Overview

This PR adds a general framework for event loop based asynchronous programming. This comes up in various different places (ProcessPipe, Logger, FEXConfig, GdbServer), where FEX's previous approach was to copy-paste a variation of the same `ppoll`-based event loop. The new framework has a number of benefits over this:
* Code duplication is reduced
* Code is easier to follow due to separating the generic event loop logic from the module-specific event handlers
* Edge cases (e.g. disconnect events) are handled consistently
* It's easier to add new event loops

The design leans heavily on [Boost.Asio](https://www.boost.org/doc/libs/1_87_0/doc/html/boost_asio/overview/basics.html), which has stood the test of time and is the basis for C++26/29's upcoming network support. FEX doesn't need most of Asio's features, but using it as inspiration allowed me to quickly come up with a clean API that covers everything we do need.

Here is an example of a simple server to communicate over a local socket with multiple clients:
```cpp
poll_reactor Reactor;
std::vector<std::byte> Data;

// Listen on local socket /tmp/FEXSocket
auto Acceptor = tcp_acceptor::create(Reactor, false, "/tmp/FEXSocket");

// Setup connection handler
Acceptor.async_accept([this](error ec, std::optional<tcp_socket> Socket) {
  // Setup read handler for new connection
  async_read(*Socket, { Data }, [](error ec, size_t NumBytes) {
    fmt::println("Received {} bytes of data!", NumBytes);
  });
});

// Enter event loop
Reactor.run();
```

This would generate the following event loop:
```mermaid
graph TD
    A[poll_reactor] --> B["async_accept(FD, ...)"]
    A --> C["async_read(FD, 16)"]
    
    B -->|incoming connection| B1[callback]
    B1 -->|add new client socket| B2["async_accept(FD, ...)"]
    B2 -->|incoming connection| B3[...]

    C -->|16 bytes received| C1[callback]
    C1 -->|call user-provided callback| C2[...]
    
    subgraph "Event Loop"
        A
    end

    subgraph "client socket"
        C
        C1
        C2
    end

    subgraph "listen() socket FD"
        B
        B1
        B2
        B3
    end
```

The framework isn't limited to sockets, but any object backed by a file descriptor that can be passed to `ppoll` can be used with the new API.

## Implementation

To facilitate callback handling, a variant of `std::move_only_function` was added to fextl. The added benefit of this `fextl` class over `std::function` is that it supports custom allocators, which is [critical for FEX](docs/ProgrammingConcerns.md).


neobrain:
Fixed the last few blockers I was aware of, so this is now be ready for review.

My failed attempt at cleaning up FEXServer socket pathname setup in #4334 has been superseded by [this code in AsyncNet.h](https://github.com/FEX-Emu/FEX/blob/43dcc84c07bc0fd5e6c6c8637ac6d56a2f99a5fd/Source/Common/AsyncNet.h#L208-L224). This will error out if the name is found to be too large now, which isn't necessary but seems like an unlikely enough scenario anyway.

Sonicadvance1:
Looks like this PR makes it so the socket communications are no longer robust against spurious EINTR.
Easiest way to test is to just start FEXServer and slam it with signals that cause EINTR (Testing spurious EINTR by not being spurious).

```bash
while true ; do kill -s SIGUSR1 `pidof FEXServer` ; done
```

neobrain:
Interesting issue, I hadn't thought of this before.

> Looks like this PR makes it so the socket communications are no longer robust against spurious EINTR.

Is the correct solution to just check for EINTR in all poll/recvmsg/sendmsg calls and repeat them if needed?

Just curious, were the modules I touched robust against EINTR before, and are we currently doing anything that requires them to be? Current main doesn't check for EINTR on any of these as far as I can see.

Sonicadvance1:
> Interesting issue, I hadn't thought of this before.
> 
> > Looks like this PR makes it so the socket communications are no longer robust against spurious EINTR.
> 
> Is the correct solution to just check for EINTR in all poll/recvmsg/sendmsg calls and repeat them if needed?
Probably. I don't remember the semantics for each syscall.

> 
> Just curious, were the modules I touched robust against EINTR before, and are we currently doing anything that requires them to be? Current main doesn't check for EINTR on any of these as far as I can see.

At least gdbserver and FEXServer were. Most of the code was just written to not assume that results of -1 were fatal instead of checking for EINTR.

neobrain:
> Probably. I don't remember the semantics for each syscall.

I went through the man pages and found that poll/sendmsg/recvmsg/accept document EINTR and EAGAIN as return values (while socket/listen don't), so I changed `fasio` to retry in those cases now.

> At least gdbserver and FEXServer were. Most of the code was just written to not assume that results of -1 were fatal instead of checking for EINTR.

Ah yes, that explains the difference. With EINTR/EAGAIN properly considered this is no longer an issue from my testing (monitored `FEXServer -f` during Steam startup while that SIGUSR1 spam is in progress, and GdbServer also withstands it).

neobrain:
Ported one FEXServerClient function over as well, but the rest don't really benefit since there's no async IO. In doing so I realized (and fixed) that `tcp_socket::recv_some` didn't handle FD receiving properly.

Sonicadvance1:
With this API, what is the ability to interrupt a blocking `::recvmsg` now?
An example would be nice to see.

Sonicadvance1:
> With this API, what is the ability to interrupt a blocking `::recvmsg` now? An example would be nice to see.

Or a ppoll, whatever the blocking call is when not receiving data on the socket.

neobrain:
(extra comment for visibility: I just edited in some additional use cases in the comment above)

Sonicadvance1 (approved):

PR #4334, opened 2025-02-06
FEXServer: Clean up socket pathname setup

neobrain:
The character counting logic isn't actually needed, since bind() doesn't need the exact byte length of the input data. Dropping the manual bookkeeping cleans up this code considerably.

Stumbled on this while working on #4333.

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -187,13 +187,11 @@ int ConnectToServer(ConnectionOption ConnectionOption) {
   // The entirety of the name is used as a path to a socket that doesn't have any filesystem backing.
   struct sockaddr_un addr {};
   addr.sun_family = AF_UNIX;
-  size_t SizeOfSocketString = std::min(ServerSocketName.size() + 1, sizeof(addr.sun_path) - 1);
-  addr.sun_path[0] = 0; // Abstract AF_UNIX sockets start with \0
-  strncpy(addr.sun_path + 1, ServerSocketName.data(), SizeOfSocketString);
-  // Include final null character.
-  size_t SizeOfAddr = sizeof(addr.sun_family) + SizeOfSocketString;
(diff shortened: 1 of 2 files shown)


neobrain:
Fixed a bug in the initial patch: Passing `sizeof(addr)` to `bind` indeed works, but it must be done on the client side as well. Should be all good now!

Sonicadvance1:
This is now incorrect as you're listening on Socket `@1000.FEXServer.Socket@@@@@@@@...` maxing out the 108 bytes of the socket name.
```
ryanh@ryanh-TR7000:~/work/FEXNew$ lsof -U | ag "FEXServer"
FEXServer 120534 ryanh    4u  unix 0x0000000000000000      0t0 1143956 @1000.FEXServer.Socket@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@@ type=STREAM (LISTEN)
FEXServer 120534 ryanh    5u  unix 0x0000000000000000      0t0 1143957 /run/user/1000/1000.FEXServer.Socket type=STREAM (LISTEN)
```

Previously it was exactly the size as expected.
```
ryanh@ryanh-TR7000:~/work/FEXNew$ lsof -U | ag "FEXServer"
FEXServer 122242 ryanh    4u  unix 0x0000000000000000      0t0 1212470 @1000.FEXServer.Socket type=STREAM (LISTEN)
FEXServer 122242 ryanh    5u  unix 0x0000000000000000      0t0 1212471 /run/user/1000/1000.FEXServer.Socket type=STREAM (LISTEN)
```

neobrain:
Ah, so these abstract sockets don't actually use null-termination :(

I'll still need to change the code slightly for PR 4333, but it doesn't make much sense in isolation then. Thanks for the explanation!

PR #4335, opened 2025-02-06
WINE: Fixes FEX_PORTABLE usage

Sonicadvance1:
Completely didn't listen to FEX_PORTABLE. Necessary otherwise it can read configs from some random locations when portable is enabled.

### DIFF
--- a/Source/Windows/Common/PortabilityInfo.h
+++ b/Source/Windows/Common/PortabilityInfo.h
@@ -1,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include "Common/Config.h"
+
+namespace FEX {
+static inline FEX::Config::PortableInformation ReadPortabilityInformation() {
+  const FEX::Config::PortableInformation BadResult {false, {}};
+  const char* PortableConfig = getenv("FEX_PORTABLE");
+  if (!PortableConfig || strtol(PortableConfig, nullptr, 0) == 0) {
+    return BadResult;
+  }
+
+  return {true, getenv("LOCALAPPDATA")};
+}
+} // namespace FEX
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -220,7 +220,11 @@ def print_man_environment_tail():
     print_man_env_option(
     "FEX_PORTABLE",
     [
-    "Allows FEX to run without installation. Global locations for configuration and binfmt_misc are ignored. These files are instead read from <FEXInterpreterPath>/fex-emu/ by default.",
+    "Allows FEX to run without installation. Global locations for configuration and binfmt_misc are ignored.",
+    "For FEXInterpreter on Linux:",
+    "These files are instead read from <FEXInterpreterPath>/fex-emu/ by default.",
+    "For Arm64ec/Wow64 WINE builds:",
+    "These files are instead read from $LOCALAPPDATA/fex-emu/ by default.",
     "For further customization, see FEX_APP_CONFIG_LOCATION and FEX_APP_DATA_LOCATION."
     ],
     "''", True)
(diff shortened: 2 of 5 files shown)


bylaws:
this isn't at all what's intended with FEX_PORTABLE? the semantics of the current executable path differs entirely to the fex interpreter path - it just isn't applicable to windows. I think setting FEX_APP_CONFIG_LOCATION can be used to achieve any desired behaviour here

Sonicadvance1:
> this isn't at all what's intended with FEX_PORTABLE? the semantics of the current executable path differs entirely to the fex interpreter path - it just isn't applicable to windows which by its very nature is always a portable install in some sense. I think setting FEX_APP_CONFIG_LOCATION can be used to achieve any desired behaviour here and this can be ignored

Yes, it is currently semantically different because the meaning of returning where libfexcorearm64ec.dll lives doesn't make much sense, but also just returning root also doesn't make sense.

Semantically we are relying on FEX_PORTABLE to remove global and local paths, along with also FEX_APP_CONFIG_LOCATION, they work hand-in-hand here.

So it seems to me the least abrasive path is to just use the path of the executable in the case of portable.

bylaws:
I don't think its helpful to use the current executable directory, a reasonable approach would just be `getenv("LOCALAPPDATA")/fex-emu/` as is used for debug output here https://github.com/FEX-Emu/FEX/blob/main/Source/Windows/Common/Logging.cpp#L44C60-L44C82

This gives portable reasonable semantics of meaning per-prefix for wine, and for real windows I guess it's whatever

Also update this as appropriate: https://github.com/FEX-Emu/FEX/blob/20b00ecc9b70451a6c21ef09773330329d85d22a/FEXCore/Scripts/config_generator.py#L223

Source/Windows/Common/PortabilityInfo.h
```diff
@@ -0,0 +1,23 @@
...
+
+namespace FEX {
+static inline FEX::Config::PortableInformation ReadPortabilityInformation(const fextl::string& ExecutablePath) {
+  const FEX::Config::PortableInformation BadResult {false, {}};
+  const char* PortableConfig = getenv("FEX_PORTABLE");
+  if (!PortableConfig) {
+    return BadResult;
+  }
+
+  uint32_t Value {};
```
bylaws, line 13: this would be simpler:
```
const char* PortableConfig = getenv("FEX_PORTABLE");
if (!PortableConfig || strtol(PortableConfig, nullptr, 0) == 0) {
  return {false, {}};
} else {
...
}
```

Sonicadvance1:
> I don't think its helpful to use the current executable directory, a reasonable approach would just be `getenv("LOCALAPPDATA")/fex-emu/` as is used for debug output here https://github.com/FEX-Emu/FEX/blob/main/Source/Windows/Common/Logging.cpp#L44C60-L44C82
> 
> This gives portable reasonable semantics of meaning per-prefix for wine, and for real windows I guess it's whatever
> 
> Also update this as appropriate:
> 
> https://github.com/FEX-Emu/FEX/blob/20b00ecc9b70451a6c21ef09773330329d85d22a/FEXCore/Scripts/config_generator.py#L223

Sounds reasonable to me, changed it.

Source/Windows/Common/PortabilityInfo.h
Sonicadvance1: Sure thing, changed over.

Source/Windows/Common/PortabilityInfo.h
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include "Common/Config.h"
+
+namespace FEX {
+static inline FEX::Config::PortableInformation ReadPortabilityInformation(const fextl::string& ExecutablePath) {
```
bylaws, line 6: Drop arg and the GetExecutableFileName method, also comment below is outdated

Sonicadvance1: Right, fixed.

bylaws (approved):

PR #4336, opened 2025-02-07
FEXCore/Profiler: Implement support for JIT float fallbacks 

Sonicadvance1:
Based on https://github.com/FEX-Emu/FEX/pull/4291 and https://github.com/FEX-Emu/FEX/pull/4324. Ideally this gets merged at the same time so
we can have Mangohud be on version 2 before giving them an upstream
patch.

Performance-wise this change falls within noise of my x87 microbench.

This just lets us track the number of float fallbacks FEX does, letting
us detect things like x87 fallbacks and how frequent they are, so we can
detect if a game might be slow or stuttering because of these fallbacks.


lioncash (approved):

Sonicadvance1:
https://github.com/Sonicadvance1/MangoHud/tree/fex For the current mangohud implementation

PR #4337, opened 2025-02-07
Extend maxps test

OFFTKP:
Extended the maxps test to cover these cases
image


OFFTKP:
I realized this case is covered in `MinMaxNaN.asm`

PR #4338, opened 2025-02-07
FEXCore/vl64: Fixes int16 encoding

Sonicadvance1:
For some reason when I was writing the tests I got the byte order
incorrect. The type header needs to be in the first byte, not the second
byte.

### DIFF
--- a/FEXCore/Source/Utils/variable_length_integer.h
+++ b/FEXCore/Source/Utils/variable_length_integer.h
@@ -55,8 +55,8 @@ struct vl64 final {
 
     if (vl8_type->Type == vl8_type_header) {
       return {vl8_type->Integer, sizeof(vl8_enc)};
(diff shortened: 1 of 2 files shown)


lioncash (approved):

bylaws (approved):
I wonder if it would be preferable to drop the bitfield use entirely and just mask and shift as appropriate for all lengths? The current approach is still reasonable though

Sonicadvance1:
> I wonder if it would be preferable to drop the bitfield use entirely and just mask and shift as appropriate for all lengths? The current approach is still reasonable though

The main optimization here is keeping as many values as 8-bit as possible since ~97% fit within the signed 7-bit range that fits within that encoding. The remaining 3% fitting within the 16-bit encoding (14-bit signed value encoded) seems reasonable enough, which is why the 32-bit and 64-bit encoding doesn't try to be any smarter.

PR #4341, opened 2025-02-09
OpcodeDispatcher: Use offset for LRCPC2 more frequently 

Sonicadvance1:
We were missing small offset immediate encoded LRCPC2 pretty much
always.
This fixes that. Finishes up what https://github.com/FEX-Emu/FEX/pull/4216 started.


Sonicadvance1:
Rebased to fix instcountci conflict

PR #4349, opened 2025-02-11
Config: Correctly handle relative paths with portable 

Sonicadvance1:
It is desired that FEX_APP_CONFIG and FEX_APP_CONFIG_LOCATION support
relative paths when portable is used. Support this.

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -108,7 +108,7 @@ class MainLoader final : public OptionMapper {
 public:
   explicit MainLoader(FEXCore::Config::LayerType Type);
   explicit MainLoader(fextl::string ConfigFile);
-  explicit MainLoader(FEXCore::Config::LayerType Type, const char* ConfigFile);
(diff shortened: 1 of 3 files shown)


Source/Common/Config.cpp
```diff
@@ -108,7 +108,7 @@ class MainLoader final : public OptionMapper {
 public:
   explicit MainLoader(FEXCore::Config::LayerType Type);
   explicit MainLoader(fextl::string ConfigFile);
-  explicit MainLoader(FEXCore::Config::LayerType Type, const char* ConfigFile);
+  explicit MainLoader(FEXCore::Config::LayerType Type, const std::string_view ConfigFile);
```
lioncash, line 111:
```suggestion
  explicit MainLoader(FEXCore::Config::LayerType Type, std::string_view ConfigFile);
```

Since this is just passed by value

lioncash (commented):
Just one minor nit

Source/Common/Config.cpp
Sonicadvance1: Good catch, fixed.

lioncash (approved):

PR #4355, opened 2025-02-13
Fixes a couple locations where is variable is copied when it could be moved

Sonicadvance1:


neobrain (approved):

Source/Tools/CommonTools/Linux/Utils/ELFSymbolDatabase.cpp
```diff
@@ -55,7 +55,7 @@ bool ELFSymbolDatabase::FindLibraryFile(fextl::string* Result, const char* Libra
   for (auto& Path : LibrarySearchPaths) {
     const fextl::string TmpPath = fextl::fmt::format("{}/{}", Path, Library);
```
lioncash, line 56: This would need to be made non-const for the move to actually occur

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
```diff
@@ -1061,7 +1061,7 @@ SyscallHandler::GenerateMap(const std::string_view& GuestBinaryFile, const std::
 
     auto rv = fextl::make_unique<FEXCore::HLE::SourcecodeMap>();
 
-    rv->SourceFile = GuestSourceFile;
+    rv->SourceFile = std::move(GuestSourceFile);
```
lioncash, line 1064: Ditto for `GuestSourceFile`

Source/Tools/CommonTools/Linux/Utils/ELFSymbolDatabase.cpp
Sonicadvance1: Interesting, TIL. resolved.

PR #4357, opened 2025-02-13
FileManagement: Throw a warning if `/proc` can't be opened

Sonicadvance1:
We use this for ProcFD collision checking. Give a warning if it can't be opened, also not doing the additional work when it fails.

This isn't likely to occur unless someone messes up their rootfs mounts.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -325,12 +325,15 @@ FileManager::FileManager(FEXCore::Context::Context* ctx)
 
   // Keep an fd open for /proc, to bypass chroot-style sandboxes
   ProcFD = open("/proc", O_RDONLY | O_CLOEXEC);
-
-  // Track the st_dev of /proc, to check for inode equality
-  struct stat Buffer;
-  auto Result = fstat(ProcFD, &Buffer);
-  if (Result >= 0) {
-    ProcFSDev = Buffer.st_dev;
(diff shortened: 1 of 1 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -325,12 +325,15 @@ FileManager::FileManager(FEXCore::Context::Context* ctx)
...
-    ProcFSDev = Buffer.st_dev;
+  if (ProcFD != -1) {
+    // Track the st_dev of /proc, to check for inode equality
+    struct stat Buffer;
+    auto Result = fstat(ProcFD, &Buffer);
+    if (Result >= 0) {
+      ProcFSDev = Buffer.st_dev;
+    }
+  } else {
+    LogMan::Msg::EFmt("Couldn't open `/proc`. Is ProcFS mounted? FEX won't be able to track FD conflicts");
```
neobrain, line 336: If this indicates invalid setup, wouldn't a fatal error be more appropriate?

Sonicadvance1: It's not necessarily fatal if people know what they're doing. The user is already in a semi-broken Linux state if procfs isn't mounted, so they're already "colouring outside of the lines".

neobrain: Fair, sounds reasonable.

neobrain (approved):

PR #4358, opened 2025-02-13
ArchHelpers/Arm64: Fix loadstore mask

Sonicadvance1:
This would become an issue when multiple threads are contending with the SIGBUS handler on the same code.

We were failing to mask the VR, OPC, Rm, and Option bits, resulting in a comparison below always resulting in a false result if another thread managed to backpatch.

This was just unlikely to be seen on LRCPC2 supporting hardware and since we fixed `LDSTUNSCALED_MASK` before, this wasn't really getting seen.


lioncash (approved):

rostomandjj (commented):
Cc

PR #4371, opened 2025-02-23
Update vixl to ff82b3328c59fa4cf2fe36697b44eae15a650371

Sonicadvance1:
A few minor changes that affected us since our last update.
- pmullb/pmullt with 16-bit element now decodes as unallocated. Because it isn't a real instruction.
- It now supports dczid_el0 with an hardcoded 64-byte size.
- It now supports sha, aes, crc, pcmul. So update test expectations


lioncash (approved):
it's about time

PR #4377, opened 2025-02-27
Implement SIDT/LSL

bylaws:
These are used by recent Denuvo. LSL behaviour doesn't match Windows or Linux exactly but I don't see this mattering, and the Windows behaviour is to encode the cpu index in the segment size anyway so exposing e.g. a way for the frontend to set sizes wouldn't cut it.
Unsure what to do about tests for these since they're not deterministic across different OSs.


Sonicadvance1 (approved):

Sonicadvance1:
Looks like some unit tests are failing. Probably describing instructions too broadly or something.

bylaws:
Don't entirely trust the decoder changes but test pasts so I think good?

Sonicadvance1 (approved):
Okay, yea. Looked over how these tables are encoded again and I see that group 7 is a little weird which you uncovered. So yes, for group 7 decoding specifically, when modrm.mod = 11b it chooses the `SecondaryModRMExtensionOpTable` table, when it is anything else it'll choose the `SecondaryExtensionOpTable` table.

We could probably encode the tables slightly differently but that's a fairly significant refactor. This works just as well.

PR #4378, opened 2025-02-27
Implement PE volatile metadata support

bylaws:
Needs some further testing but seems to work as is.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -186,6 +186,10 @@ public:
 
   void AddThunkTrampolineIRHandler(uintptr_t Entrypoint, uintptr_t GuestThunkEntrypoint) override;
 
+  void AddForceTSOInformation(const IntervalList<uint64_t>& ValidRanges, fextl::set<uint64_t>&& Instructions) override;
+
+  void RemoveForceTSOInformation(uint64_t Address, uint64_t Size) override;
+
 public:
   friend class FEXCore::HLE::SyscallHandler;
 #ifdef JIT_ARM64
@@ -373,5 +377,7 @@ private:
   std::shared_mutex CustomIRMutex;
   std::atomic<bool> HasCustomIRHandlers {};
   fextl::unordered_map<uint64_t, std::tuple<CustomIREntrypointHandler, void*, void*>> CustomIRHandlers;
+  IntervalList<uint64_t> ForceTSOValidRanges; // The ranges for which ForceTSOInstructions has populated data
+  fextl::set<uint64_t> ForceTSOInstructions;
 };
 } // namespace FEXCore::Context
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -565,6 +565,16 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
     for (size_t j = 0; j < CodeBlocks->size(); ++j) {
       const FEXCore::Frontend::Decoder::DecodedBlocks& Block = CodeBlocks->at(j);
+
+      bool BlockInForceTSOValidRange = false;
+      auto InstForceTSOIt = ForceTSOInstructions.end();
+      if (ForceTSOValidRanges.Contains({Block.Entry, Block.Entry + Block.Size})) {
+        if (auto It = ForceTSOInstructions.lower_bound(Block.Entry); *It < Block.Entry + Block.Size) {
+          InstForceTSOIt = It;
+          BlockInForceTSOValidRange = true;
+        }
+      }
+
       // Set the block entry point
       Thread->OpDispatcher->SetNewBlockIfChanged(Block.Entry);
 
@@ -581,6 +591,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
       }
 
       for (size_t i = 0; i < InstsInBlock; ++i) {
+        uint64_t InstAddress = Block.Entry + BlockInstructionsLength;
         const FEXCore::X86Tables::X86InstInfo* TableInfo {nullptr};
         const FEXCore::X86Tables::DecodedInst* DecodedInfo {nullptr};
 
@@ -603,7 +614,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
         Thread->OpDispatcher->FlushRegisterCache(true);
 
         if (ExtendedDebugInfo || Thread->OpDispatcher->CanHaveSideEffects(TableInfo, DecodedInfo)) {
-          Thread->OpDispatcher->_GuestOpcode(Block.Entry + BlockInstructionsLength - GuestRIP);
+          Thread->OpDispatcher->_GuestOpcode(InstAddress - GuestRIP);
         }
 
         if (Config.SMCChecks == FEXCore::Config::CONFIG_SMC_FULL) {
@@ -620,7 +631,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
           Thread->OpDispatcher->SetCurrentCodeBlock(CodeWasChangedBlock);
           Thread->OpDispatcher->_ThreadRemoveCodeEntry();
-          Thread->OpDispatcher->ExitFunction(Thread->OpDispatcher->_EntrypointOffset(GPRSize, Block.Entry + BlockInstructionsLength - GuestRIP));
+          Thread->OpDispatcher->ExitFunction(Thread->OpDispatcher->_EntrypointOffset(GPRSize, InstAddress - GuestRIP));
 
           auto NextOpBlock = Thread->OpDispatcher->CreateNewCodeBlockAfter(CurrentBlock);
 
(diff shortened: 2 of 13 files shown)


FEXCore/include/FEXCore/Utils/IntervalList.h
```diff
@@ -142,4 +149,14 @@ class IntervalList {
 
     return It != Intervals.end() && It->Offset < Entry.End;
   }
+
+  bool Contains(Interval Entry) {
```
lioncash, line 153: seems like this could be const qualified, no?

lioncash (commented):
just one super minor nit

Source/Tools/FEXConfig/main.qml
```diff
@@ -601,6 +601,11 @@ ApplicationWindow {
                         text: qsTr("Enable non-tearing split-lock atomics")
                         config: "StrictInProcessSplitLocks"
                     }
+                    ConfigCheckBox {
+                        topPadding: 4
+                        text: qsTr("Enable PE volatile metadata")
+                        config: "VolatileMetadata"
+                    }
```
neobrain, line 608: This won't do anything for vanilla FEX, will it? FEXConfig isn't typically used with WoA FEX, so I think we can do without a dedicated option (the Advanced tab will still list it anyway).

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -565,6 +565,16 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
     for (size_t j = 0; j < CodeBlocks->size(); ++j) {
       const FEXCore::Frontend::Decoder::DecodedBlocks& Block = CodeBlocks->at(j);
+
+      bool BlockInForceTSOValidRange = false;
```
neobrain, line 569: There's some redundancy between this variable and the `!= end()` checks further below. We can either drop this variable or replace the iterator checks with it.

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -680,6 +697,10 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
         if (Thread->OpDispatcher->FinishOp(DecodedInfo->PC + DecodedInfo->InstSize, i + 1 == InstsInBlock)) {
           break;
         }
+
+        while (InstForceTSOIt != ForceTSOInstructions.end() && *InstForceTSOIt < Block.Entry + BlockInstructionsLength) {
+          InstForceTSOIt++;
+        }
```
neobrain, line 703: Not entirely happy with how the `InstForceTSOIt` variable is used in just three isolated places that are however spread across 150 lines of code. One minor improvement to this is to at least replace this loop-in-a-loop-in-a-loop:

```suggestion
        InstForceTSOIt = std::find_if(InstForceTSOIt, ForceTSOInstructions.end(), [&](auto Val) { return Val >= Block.Entry + BlockInstructionsLength; });
```

Perhaps it would be better to move this (and the `ForceTSO` declaration) to the top of the inner loop though? That would keep the uses of that variable much more local.

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -971,6 +992,19 @@ void ContextImpl::AddThunkTrampolineIRHandler(uintptr_t Entrypoint, uintptr_t Gu
...
 }
 
+void ContextImpl::AddForceTSOInformation(const IntervalList<uint64_t>& ValidRanges, fextl::set<uint64_t>&& Instructions) {
+  LogMan::Throw::AFmt(CodeInvalidationMutex.try_lock() == false, "CodeInvalidationMutex needs to be unique_locked here");
+  ForceTSOValidRanges.Insert(ValidRanges);
+  ForceTSOInstructions.merge(Instructions);
+}
+
+void ContextImpl::RemoveForceTSOInformation(uint64_t Address, uint64_t Size) {
+  LogMan::Throw::AFmt(CodeInvalidationMutex.try_lock() == false, "CodeInvalidationMutex needs to be unique_locked here");
```
neobrain, line 1002: This will run `try_lock` even in release / non-assert configurations. Is that intended?

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -46,6 +46,12 @@ enum class BTAction {
   BTComplement,
 };
 
+enum class ForceTSOMode {
+  Default,
```
neobrain, line 50: This seems to be a good place to mention what the default is.

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -373,5 +377,7 @@ class ContextImpl final : public FEXCore::Context::Context {
   std::shared_mutex CustomIRMutex;
   std::atomic<bool> HasCustomIRHandlers {};
   fextl::unordered_map<uint64_t, std::tuple<CustomIREntrypointHandler, void*, void*>> CustomIRHandlers;
+  IntervalList<uint64_t> ForceTSOValidRanges; // The ranges for which ForceTSOInstructions has populated data
```
neobrain, line 380: Why does it need this ranges list in addition to ForceTSOInstructions? Since you're already using an ordered set, it would seem that algorithmic complexity shouldn't be a problem, but I haven't fully looked into the implementation details.

Source/Windows/Common/InvalidationTracker.h
```diff
@@ -22,13 +22,17 @@ class InvalidationTracker {
   InvalidationTracker(FEXCore::Context::Context& CTX, const std::unordered_map<DWORD, FEXCore::Core::InternalThreadState*>& Threads);
   void HandleMemoryProtectionNotification(uint64_t Address, uint64_t Size, ULONG Prot);
   void HandleImageMap(uint64_t Address);
-  void InvalidateContainingSection(uint64_t Address, bool Free);
+  struct InvalidateContainingSectionResult {
```
neobrain, line 25: Wouldn't it make sense to use `IntervalList<uint64_t>::Interval` here instead of defining a new type that's only used once?

Source/Tools/FEXConfig/main.qml
bylaws: FEXconfig just works the same with windows fex, not sure why it shouldn't be here

FEXCore/Source/Interface/Context/Context.h
bylaws: Not all regions will have TSO metadata, the default tso setting needs to be used in such cases

Source/Windows/Common/InvalidationTracker.h
bylaws: I had it as that but thought to avoid confusion between size/end a struct made more sense

FEXCore/Source/Interface/Core/Core.cpp
bylaws: it's what's done in other places and is practically free so probably fine?

FEXCore/Source/Interface/Core/Core.cpp
bylaws: They are both necessary - a memory access past the last volatile metadata entry but within a range will be a no-tso access. A memory access outside of the range will use the default setting

Source/Tools/FEXConfig/main.qml
neobrain: The setting doesn't have any effect on other (currently more commonly used) configurations. Options that don't do anything aren't good UX, but if you can find a way to disable it just for those more common configurations, I'd be fine with that as well.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
neobrain: I think this value needs renaming, since I actually misunderstood what "Default" means here. IIUC, this value is actually used for unannotated regions, for which should FEX's normal TSO configuration should be used. If that's correct, maybe something like "NoOverride" is a clearer name?

Source/Tools/FEXConfig/main.qml
bylaws: The same config file is used for both - it's not really something that can be distinguished. I can add a note that it applies to wine fex only? Considering it's now distributed via officials repos I don't think it makes sense to just hide it overall when FEXConfig really isn't that user facing anyway

FEXCore/Source/Interface/Core/Core.cpp
bylaws: This can't really be moved to the top of the loop, it's intended this happens at the end after BlockInstructionsLength is updated and after any breaks that could occur that would make the work redundant

Sonicadvance1 (approved):

Source/Tools/FEXConfig/main.qml
neobrain: Fair, you've convinced me :)

FTR I also argued against adding a GUI for non-tearing split lock emulation. But it's a good point that the same config file may be shared for either type of FEX build, and it's definitely a useful setting for WoA configurations (even more so if distros are already shipping that widely).

Source/Tools/FEXConfig/main.qml
```diff
@@ -601,6 +601,11 @@ ApplicationWindow {
                         text: qsTr("Enable non-tearing split-lock atomics")
                         config: "StrictInProcessSplitLocks"
                     }
+                    ConfigCheckBox {
+                        topPadding: 4
+                        text: qsTr("Enable PE volatile metadata (ARM64EC FEX only)")
```
neobrain, line 606: Minor suggestion to make this somewhat shorter:

```suggestion
                        text: qsTr("Use PE volatile metadata for ARM64EC")
```

FEXCore/Source/Interface/Core/Core.cpp
Sonicadvance1: I'm fine with it walking the interval list like this

Sonicadvance1 (approved):

FEXCore/Source/Interface/Core/Core.cpp
bylaws: Should I revert my change?

neobrain (changes requested):
Not seeing the rationale for keeping the code unnecessarily non-local (see https://github.com/FEX-Emu/FEX/pull/4378#discussion_r1976408461). Seems simple enough to fix though?

bylaws:
> Not seeing the rationale for keeping the code unnecessarily non-local (see [#4378 (comment)](https://github.com/FEX-Emu/FEX/pull/4378#discussion_r1976408461)). Seems simple enough to fix though?

(I just haven't had time to actually do it)

neobrain (approved):
Looking good, thanks! Also appreciated the clarifications about how this feature works during review :)

PR #4390, opened 2025-03-06
Softfloat: Define INLINE

Sonicadvance1:
This is embarassing. We were throwing away performance by failing to use the softfloat library's inline helpers.

Turns out we needed to define `INLINE` to something in order for them to work.

Feels bad.


lioncash (approved):
oops

Sonicadvance1:
Alright, updated with the `SOFTFLOAT_FAST_*` defines as well. Should improve a couple of handlers.

lioncash (approved):

PR #4392, opened 2025-03-07
JIT: Fixe scalar reciprocal when AFP is supported

Sonicadvance1:
Somehow I had completely missed this and recent reciprocal tests have exposed it as a problem. When AFP is supported but not RPRES then we were hitting this code path.

We were failing to insert in to the destination correctly, which because the reciprocal is calculated using fdiv using a synthesized constant, this would just zero the remaining portion of the register.


lioncash (approved):

pmatos:
Nice - how did you notice this without a test failing? Should we add a test as well to make sure these things don't regress?

PR #4399, opened 2025-03-10
Remove some brackets from expected results in tests

OFFTKP:
Almost all tests don't surround the GPR results with [ ] so I think it's better for consistency if these don't do it too, unless if there's some reason I'm unaware of.


Sonicadvance1 (approved):
Thanks!

PR #4401, opened 2025-03-11
Add tests for F(U)COMI(P)

OFFTKP:
Untested

### DIFF
--- a/unittests/ASM/X87/DB_E8.asm
+++ b/unittests/ASM/X87/DB_E8.asm
@@ -82,5 +82,5 @@ minus_one:
 
 align 8
 qnan:
-  dw 0x7FFF
   dq 0x8000000000000000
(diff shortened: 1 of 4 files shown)


OFFTKP:
Since it passes on hostrunner I assume the test themselves are fine, yes?

Also, unsure what the X87_F64 equivalent should be.

Sonicadvance1:
> Since it passes on hostrunner I assume the test themselves are fine, yes?

Interesting, sounds like it might be a legitimate bug in our JIT backend. Will need some investigation to see what's broken.

> Also, unsure what the X87_F64 equivalent should be.

It can probably be the same tests with the `"Env": { "FEX_X87REDUCEDPRECISION" : "1" }` option set like the rest of the tests.

Sonicadvance1 (approved):

PR #4404, opened 2025-03-12
LibraryForwarding/vulkan: Add missing vkCmdPushDescriptorSetWithTemplate declaration

neobrain:
This was missed in #4402 since `thunkgen` doesn't error out when annotating parameters of an undeclared entry point.


Sonicadvance1 (approved):
Oops

Issue #4301, opened 2025-01-23
We can not dynamically link anything


Sonicadvance1:
#5476 uncovered this issue using the glibc fault CI test. A runner happened to have libxxhash-dev installed, which causes symbol lookup to happen dynamically, which causes glibc allocations.
```
Program terminated with signal SIGILL, Illegal instruction.
#0  FEXCore::Assert::ForcedAssert () at /home/ryanh/FEX/FEXCore/Source/Utils/ForcedAssert.cpp:9
9         asm volatile("hlt #1");
(gdb) y
Undefined command: "y".  Try "help".
(gdb) bt
#0  FEXCore::Assert::ForcedAssert () at /home/ryanh/FEX/FEXCore/Source/Utils/ForcedAssert.cpp:9
#1  0x0000aaaac2ac23f0 in FEXCore::Allocator::EvaluateReturnAddress (Return=0xffffbb8c0ed4 <__GI__dl_exception_create_format+244>) at /home/ryanh/FEX/FEXCore/Source/Utils/AllocatorOverride.cpp:114
#2  0x0000aaaac2ac20e4 in fault_malloc (size=56) at /home/ryanh/FEX/FEXCore/Source/Utils/AllocatorOverride.cpp:130
#3  0x0000ffffbb8c0ed4 in malloc (size=56) at ../include/rtld-malloc.h:56
#4  __GI__dl_exception_create_format (exception=exception@entry=0xffffeda66d68, objname=0xffffeda6c721 "./Bin/FEXLoader", fmt=fmt@entry=0xffffbb8df4d0 "undefined symbol: %s%s%s") at ./elf/dl-exception.c:157
#5  0x0000ffffbb8c74a8 in _dl_lookup_symbol_x (undef_name=0xaaaac2508f06 "XXH3_64bits", undef_map=undef_map@entry=0xffffbb8fa370, ref=ref@entry=0xffffeda66dd8, symbol_scope=<optimized out>, version=0xffffbb8bb018, type_class=type_class@entry=1,
    flags=1, skip_map=skip_map@entry=0x0) at ./elf/dl-lookup.c:877
#6  0x0000ffffbb8cd080 in _dl_fixup (l=0xffffbb8fa370, reloc_arg=3504) at ./elf/dl-runtime.c:95
#7  0x0000ffffbb8cf1ac in _dl_runtime_resolve () at ../sysdeps/aarch64/dl-trampoline.S:100
[...]

PR #4410, opened 2025-03-15
CMake: Prefer static libraries over shared libraries

Sonicadvance1:
This option allows us to give cmake a priority when searching for libraries. Invert the order of library suffixes to be static first then shared.

This doesn't fully remove the problems of shared library linking but at least mitigates the problem that #4301 brought up.

As a test I have the system libxxhash installed on my system and cmake now prefers the static library from that package instead of the shared library.

Still more work to go, but this at least removes a bit of the headache.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,15 @@ if (NOT MINGW_BUILD)
   endif()
 endif()
 
+# Setup library search paths to prefer static libraries if available.
+# Helps for when package maintainers try to replace our static libraries with system libraries.
+# See https://github.com/FEX-Emu/FEX/issues/4301 why we don't want dynamic libraries.
+if (MINGW_BUILD)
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
check code formatting workflow is now disabled. The repository that held the action has been deleted off of github.

neobrain:
(FTR this is pending on discussion in https://github.com/FEX-Emu/FEX/pull/4411#issuecomment-2731948655)

neobrain:
Cleaning up stale PRs as per https://github.com/FEX-Emu/FEX/pull/4411#issuecomment-2804812593 , feel free to reopen if desired:

> To summarize the discussion so far, dynamic linking doesn't per se cause any memory allocations in normal operation. We have no evidence that it causes problems for FEX, so we don't need to bend over backwards to force static linking everywhere.

PR #4411, opened 2025-03-15
FEXCore: Use musl libm statically

Sonicadvance1:
This removes our dependency on libm (although it is still included with
libstdc++ and libgcc). The only place libm is required is actually the
x87 softfloat where we fallback to 128-bit softfloat functions. These
are the couple of transcendental operations that the Berkely Softfloat
library doesn't handle.

This doesn't quite solve the issue that https://github.com/FEX-Emu/FEX/issues/4301 brought up, but it's
getting closer.


neobrain (changes requested):
How can the underlying issue with dynamic linking be reproduced/verified?

Quoting my original comment from https://github.com/FEX-Emu/FEX/issues/4301#issuecomment-2621132832 :

> Before we outright ban any dynamic linking whatsoever, it would be nice to understand what specific circumstances actually trigger this.

Sonicadvance1:
> How can the underlying issue with dynamic linking be reproduced/verified?
> 
> Quoting my original comment from [#4301 (comment)](https://github.com/FEX-Emu/FEX/issues/4301#issuecomment-2621132832) :
> 
> > Before we outright ban any dynamic linking whatsoever, it would be nice to understand what specific circumstances actually trigger this.

Reproduced is easy, just install your distro's packages for libfmt-dev and libxxhash-dev. FEX will then pick them up automatically. Run ldd on the executable and see that it is linked against those system libraries.

```bash
$ ldd FEXInterpreter
    linux-vdso.so.1 (0x0000ffffba4a4000)
    libfmt.so.11 => not found
    libxxhash.so.0 => /lib/aarch64-linux-gnu/libxxhash.so.0 (0x0000ffffb9c10000)
    libstdc++.so.6 => /lib/aarch64-linux-gnu/libstdc++.so.6 (0x0000ffffb99f0000)
    libm.so.6 => /lib/aarch64-linux-gnu/libm.so.6 (0x0000ffffb9950000)
    libgcc_s.so.1 => /lib/aarch64-linux-gnu/libgcc_s.so.1 (0x0000ffffb9910000)
    libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffb9760000)
    /lib/ld-linux-aarch64.so.1 (0x0000ffffba460000)
```

The specific circumstances are still the few packages we try to do `find_package` or whatever in cmake and when they're found they typically only have a shared library version globally installed instead of also a static library.
fmt has shared and static, xxhash has only shared at least on Ubuntu.

Validating this in our local CI would be trivial, just count the number of lines returned when doing ldd.

neobrain:
> Reproduced is easy, just install your distro's packages for libfmt-dev and libxxhash-dev. FEX will then pick them up automatically. Run ldd on the executable and see that it is linked against those system libraries.

I'm not asking about how to check for dynamic linking. How can the spurious allocations triggered by library linking be reproduced? What's causing these allocations?

Sonicadvance1:
> > Reproduced is easy, just install your distro's packages for libfmt-dev and libxxhash-dev. FEX will then pick them up automatically. Run ldd on the executable and see that it is linked against those system libraries.
> 
> I'm not asking about how to check for dynamic linking. How can the spurious allocations triggered by library linking be reproduced? What's causing these allocations?

Oh, enable the glibc faulting jemalloc and just run an application under FEX. If you have xxhash dynamically linked in then glibc will allocate memory on the first block emitted once an xxhash function is called.

I would guess that fmt is the same and would hit a crash as soon as the first log message is emitted.

It's a systemic issue so it's not something special about these two libraries.

Sonicadvance1:
> > Oh, enable the glibc faulting jemalloc and just run an application under FEX. If you have xxhash dynamically linked in then glibc will allocate memory on the first block emitted once an xxhash function is called.
> 
> This doesn't reproduce for me on either openSUSE or Ubuntu with `ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT=ON` and `ENABLE_JEMALLOC_GLIBC_ALLOC=OFF` despite `ldd FEXInterpreter` listing /lib/aarch64-linux-gnu/libxxhash.so.0. I hacked in a call to `XXH3_64bits` into main() to ensure the library actually gets used.
> 
> Maybe there's some weird interaction with other settings in my CMakeCache.txt. Can you give me a full cmake command line that would reproduce the problem on a fresh build folder with predictable configuration?

Ah yes, I see why it's hard to reproduce. The problem is that this is also susceptible to glibc's internal allocations which change depending on many things. We're just getting lucky since I've cut out so many global static initializers at this point that it's hard to reproduce, since our allocations are /relatively/ consistent run to run now.

Which makes this even more terrible/terrifying, since that means any change in the development environment could introduce allocations without us knowing.

neobrain:
Going by the hard data we have: The allocation that triggered the backtrace posted in https://github.com/FEX-Emu/FEX/issues/4301 seems to be caused by an `undefined symbol` error during symbol lookup. Why is it throwing that error in the first place?

bylaws:
Arm64ec is already effectively doing this so it can definitely be cased behind !MINGW, just including some non-libm musl parts also.

I really don't think it matters but the sources could be unified:
https://github.com/FEX-Emu/FEX/tree/main/Source/Windows/Common/CRT/Musl

Sonicadvance1:
> Going by the hard data we have: The allocation that triggered the backtrace posted in #4301 seems to be caused by an `undefined symbol` error during symbol lookup. Why is it throwing that error in the first place?

As far as I can tell, that error was just a byproduct of something else, but there are other allocations that happen in that codepath that would hit the same problem. Just harder to reproduce. Spooky scary allocations.

neobrain:
> there are other allocations that happen in that codepath that would hit the same problem. Just harder to reproduce. Spooky scary allocations.

Could you give an example?

Sonicadvance1:
> > there are other allocations that happen in that codepath that would hit the same problem. Just harder to reproduce. Spooky scary allocations.
> 
> Could you give an example?

The easiest one is that the lazy got.plt lookup has a little array for lookups. Which can cause allocations as it grows, We just don't happen to hit it currently.

Sonicadvance1:
> > The easiest one is that the lazy got.plt lookup has a little array for lookups. Which can cause allocations as it grows, We just don't happen to hit it currently.
> 
> Can you point me to the source location? What exactly would need to happen for that array to be exhausted? Wouldn't eager binding resolve that specific problem?
> 
> (EDIT: Accidentally hit a keycombo that apparently comments and auto-closes the PR, thanks GitHub 😓 )

Sure. https://github.com/bminor/glibc/blob/master/elf/dl-lookup.c#L688

neobrain:
> Sure. https://github.com/bminor/glibc/blob/master/elf/dl-lookup.c#L688

This `add_dependency` function isn't called for libraries that are dynamically linked at compile-time but only for those that are `dlopen`ed: https://github.com/bminor/glibc/blob/release/2.41/master/elf/dl-lookup.c#L843-L848

I think this would only apply to our library forwarding code, but if I'm reading the dl code correctly even those are fine since they use explicit lookups.

neobrain:
Are you planning to pick this up again?

To summarize the discussion so far, dynamic linking doesn't per se cause any memory allocations in normal operation. We have no evidence that it causes any problems for FEX, so there's no reason for us to bend over backwards to force static linking everywhere.

neobrain:
Cleaning up stale PRs as per previous comment, feel free to reopen if desired.

PR #4412, opened 2025-03-15
actions: Update to step-security change-files action

Sonicadvance1:
This repo is provided by them with the commit prior to when it was compromised. LLVM has switched to the same repo for now.

We probably want to find some better solution to this at some point.


neobrain (approved):
Matches what LLVM did for the time being: https://github.com/llvm/llvm-project/commit/6616acd80cd91a0075e3cd481bb9a6d82fd4ea9e

PR #4413, opened 2025-03-16
FEXCore: Statically link libgcc and libstdc++

Sonicadvance1:
This PR along with the previous ones gets FEXCore and FEXInterpreter only linking in libc.

```bash
$ ldd ./Bin/FEXInterpreter
        linux-vdso.so.1 (0x0000ffffa206f000)
        libc.so.6 => /lib/aarch64-linux-gnu/libc.so.6 (0x0000ffffa1460000)
        /lib/ld-linux-aarch64.so.1 (0x0000ffffa2020000)
```

The only remaining issue is if the builder has libfmt, then it is unlikely to have static versions of those libraries available which can ruin this.

Looks like we'll need to go back to #3709 and #3677 to make that problem go away.

Additionally xxhash can be installed on the host, but on Ubuntu at least they provide a statically library as well, which #4410 at least resolves.


neobrain (changes requested):
How can the underlying issue with dynamic linking be reproduced/verified?

Quoting my original comment from https://github.com/FEX-Emu/FEX/issues/4301#issuecomment-2621132832 : 
> Before we outright ban any dynamic linking whatsoever, it would be nice to understand what specific circumstances actually trigger this.

neobrain:
Cleaning up stale PRs as per https://github.com/FEX-Emu/FEX/pull/4411#issuecomment-2804812593, feel free to reopen if desired:

> To summarize the discussion so far, dynamic linking doesn't per se cause any memory allocations in normal operation. We have no evidence that it causes problems for FEX, so we don't need to bend over backwards to force static linking everywhere.

PR #4417, opened 2025-03-18
OpcodeDispatcher: fix BT flags

alyssarosenzweig:
ZF needs to be preserved.
    
    the new code is the same instr count on flagm although probably an extra uop.
    the inst count regression is on flagm, but we can't tolerate broken behaviour.


Sonicadvance1:
Interesting. Looks like the ZF behaviour of the BT operations is an Intel specific behaviour, while AMD claims the instruction has it undefined. I'm fine with the stricter behaviour of the Intel behaviour although.

alyssarosenzweig:
> Interesting. Looks like the ZF behaviour of the BT operations is an Intel specific behaviour, while AMD claims the instruction has it undefined. I'm fine with the stricter behaviour of the Intel behaviour although.

Entertaining. I got this from Felix's docs.

The stricter behaviour isn't much harder to do when we have flagm so, :shrug:

Sonicadvance1:
> > Interesting. Looks like the ZF behaviour of the BT operations is an Intel specific behaviour, while AMD claims the instruction has it undefined. I'm fine with the stricter behaviour of the Intel behaviour although.
> 
> Entertaining. I got this from Felix's docs.
> 
> The stricter behaviour isn't much harder to do when we have flagm so, 🤷

Indeed, those docs are extracted from Intel's documentation.

Sonicadvance1 (approved):

PR #4421, opened 2025-03-20
Thunks/Vulkan: New generation based description json

Sonicadvance1:
Based on #4420 which should be merged first.

This describes everything about the Vulkan interface that we desire and is flexible enough for our needs.
This trivializes updating the Vulkan headers now, as I just point the DefinitionExtract script at the new folder and it generates a new descriptions, then just fixing up any new function or types required.

This also gives me a really good look as to what is missing in our Vulkan thunking implementation as we have explicitly lists of types and functions that are disabled outright.

This basically matches the workflow that I want when updating vulkan every few months.


neobrain:
Many of the goals here could be achieved differently to fit into existing infrastructure:
* A "64_bit_only" annotation can be added to skip functions that don't work on 32-bit
* Updating definitions with the existing DefinitionExtract.py can be done with short grep oneliners (which is how I double-checked the update in #4402 and which barely takes a minute)
* Certain `fex_gen_param` annotations could imply `custom_host_impl` to facilitate diffing

Does it *really* need effectively yet another layer of code generation in this subsystem? Each of these layers makes understanding final code output more difficult and makes iterating on the codebase more involved. The existing layers have very tangible value propositions to justify this:
* vulkan_interface.cpp+thunkgen generate boilerplate that would be infeasible to write manually without mistakes, adding safety checks along the way (to warn you about APIs that require custom handling)
* Templates in Host.h express type-generic code more easily than code-generation logic in thunkgen would (again adding safety checks by using C++'s type system)

Does this new JSON layer have a similarly strong value proposition? We barely update these definitions twice a year, and the current process to do so takes me a few minutes at worst. There's also still parts this PR doesn't automate, so a conscious effort is needed regardless.

I could probably live with this added JSON-based generator if the existing libvulkan_interface.cpp output didn't change (to allow dropping the generator again if we need the flexibility later). Reordering things alphabetically makes feature-wise bringup on 32-bit more tricky, and blowing up the interface file to 5000 lines makes code navigation cumbersome. Documenting comments shouldn't be duplicated across two files (one which is a JSON blob where lack of context makes the comment useless, and one autogenerated which became too large to navigate).

I like the general idea of making the updates easier, but I don't like the ripple effects of this particular approach. Splitting libvulkan_interface.cpp into different, individually updatable files seems reasonable. We'll probably need to pull in vk.xml at some point, so it makes sense to design with this in mind.

Sonicadvance1:
> Many of the goals here could be achieved differently to fit into existing infrastructure:

To note, this script is already existing infrastructure

> 
> * A "64_bit_only" annotation can be added to skip functions that don't work on 32-bit
> * Updating definitions with the existing DefinitionExtract.py can be done with short grep oneliners (which is how I double-checked the update in [LibraryForwarding/vulkan: Update to 1.4.310 #4402](https://github.com/FEX-Emu/FEX/pull/4402) and which barely takes a minute)
> * Certain `fex_gen_param` annotations could imply `custom_host_impl` to facilitate diffing
> 
> Does it _really_ need effectively yet another layer of code generation in this subsystem? Each of these layers makes understanding final code output more difficult and makes iterating on the codebase more involved. The existing layers have very tangible value propositions to justify this:

This is already part of the loop of code generation that is in use for updating the Vulkan interface file, it just moves the manual labor from manually twiddling `libvulkan_interface.cpp` , to adjusting json which describes all the information already.

> 
> * vulkan_interface.cpp+thunkgen generate boilerplate that would be infeasible to write manually without mistakes, adding safety checks along the way (to warn you about APIs that require custom handling)

This isn't removing or changing any of those thunkgen boilerplate safety checks?

> * Templates in Host.h express type-generic code more easily than code-generation logic in thunkgen would (again adding safety checks by using C++'s type system)
> 
> Does this new JSON layer have a similarly strong value proposition? We barely update these definitions twice a year, and the current process to do so takes me a few minutes at worst. There's also still parts this PR doesn't automate, so a conscious effort is needed regardless.

The value proposition here is that updating Vulkan becomes a heck of a lot easier and less mistake prone. Manually auditing the code generated and ensuring all the decorations are on the existing functions is time consuming. Having some file format that describes all these decorations, and disabling of functions and all the other gubbins makes this easier to reason about when four months from now I'm updating Vulkan and I lose a random `vkCreateComputePipelines` in the middle of 3000 line generated interface file change.

> 
> I could probably live with this added JSON-based generator if the existing libvulkan_interface.cpp output didn't change (to allow dropping the generator again if we need the flexibility later). Reordering things alphabetically makes feature-wise bringup on 32-bit more tricky, and blowing up the interface file to 6000 lines makes code navigation cumbersome. Documenting comments shouldn't be duplicated across two files (one which is a JSON blob where lack of context makes the comment useless, and one autogenerated which became too large to navigate).

If you don't like the comments duplicated then it is easy enough to say that you want it disabled and keep it in the json.
It's a generated file (just like the current libvulkan_interface.cpp practically is), navigating it is a secondary concern compared to parsing the ground truth for all the minutiae which is now getting described in json.

> 
> I like the general idea of making the updates easier, but I don't like the ripple effects of this particular approach. Splitting libvulkan_interface.cpp into different, individually updatable files seems reasonable. We'll probably need to pull in vk.xml at some point, so it makes sense to design with this in mind.

Switching over to parsing vk.xml could definitely be a future goal, it would remove all the clang parsing shenanigans that this does, but that's not today. I'll end up reusing this script for other libraries in the future if/when we desire it.

neobrain:
> This isn't removing or changing any of those thunkgen boilerplate safety checks?

I wasn't saying it did. I was illustrating that the existing layers of code generation have a strong value proposition, whereas this JSON layer is automating a 2-minute task we run twice a year. It's automation for its own sake.

If it bothers you to run this task I can take care of it for future updates (as I've done [before](https://github.com/FEX-Emu/FEX/pull/3477) and [before](https://github.com/FEX-Emu/FEX/pull/3159)). If this ends up taking more than 2 minutes at some point we can revisit the idea of adding more automation.

PR #4422, opened 2025-03-20
WINE: Support sleeping a process

Sonicadvance1:
Support using an app profile for putting a process to sleep at the start. Useful for getting a debugger attached early.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -530,7 +530,8 @@ NTSTATUS ProcessInit() {
   InitSyscalls();
 
   FEX::Windows::InitCRTProcess();
-  FEX::Config::LoadConfig(nullptr, FEX::Windows::GetExecutableFilePath(), nullptr, FEX::ReadPortabilityInformation());
+  const auto ExecutablePath = FEX::Windows::GetExecutableFilePath();
+  FEX::Config::LoadConfig(nullptr, ExecutablePath, nullptr, FEX::ReadPortabilityInformation());
   FEXCore::Config::ReloadMetaLayer();
   FEX::Windows::Logging::Init();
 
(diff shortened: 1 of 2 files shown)


bylaws:
Not super sure this makes sense for wine since you can just run it under gdb directly and there's no particular reason for this to be the JITs responsibility

Sonicadvance1:
> Not super sure this makes sense for wine since you can just run it under gdb directly and there's no particular reason for this to be the JITs responsibility Vs wine itself (which has a wait for debugger option)

I get that there is some way to do it, but figuring out how to punch through all the proton layers and also getting it to only stop on the particular app I care about with launchers in the way gets difficult.

It's much easier to create a `rdr2.exe.json` file and gdb attach once it is sleeping.

neobrain:
> (which has a wait for debugger option)

For personal reference, which option is that? It doesn't seem to be something that comes up in web search easily.

bylaws:
> > Not super sure this makes sense for wine since you can just run it under gdb directly and there's no particular reason for this to be the JITs responsibility Vs wine itself (which has a wait for debugger option)
> 
> I get that there is some way to do it, but figuring out how to punch through all the proton layers and also getting it to only stop on the particular app I care about with launchers in the way gets difficult.
> 
> It's much easier to create a `rdr2.exe.json` file and gdb attach once it is sleeping.

Right, wine's wait for debugger is just for the specific executable being launched, with launchers you're right this would be better. (I assume you're not actually trying to debug RDR2 btw, the issue on that is known)

Source/Windows/ARM64EC/Module.cpp
```diff
@@ -573,10 +574,18 @@ NTSTATUS ProcessInit() {
...
   FEX_CONFIG_OPT(ProfileStats, PROFILESTATS);
+  FEX_CONFIG_OPT(StartupSleep, STARTUPSLEEP);
+  FEX_CONFIG_OPT(StartupSleepProcName, STARTUPSLEEPPROCNAME);
 
   if (IsWine && ProfileStats()) {
     StatAllocHandler = fextl::make_unique<FEX::Windows::StatAlloc>(FEXCore::Profiler::AppType::WIN_ARM64EC);
   }
+
+  if (StartupSleep() && (StartupSleepProcName().empty() || ExecutablePath == StartupSleepProcName())) {
+    LogMan::Msg::IFmt("[{}][{}] Sleeping for {} seconds", ::getpid(), ExecutablePath, StartupSleep());
```
bylaws, line 585: GetCurrentProcessId

Source/Windows/WOW64/Module.cpp
```diff
@@ -504,10 +505,17 @@ void BTCpuProcessInit() {
...
   FEX_CONFIG_OPT(ProfileStats, PROFILESTATS);
+  FEX_CONFIG_OPT(StartupSleep, STARTUPSLEEP);
+  FEX_CONFIG_OPT(StartupSleepProcName, STARTUPSLEEPPROCNAME);
 
   if (IsWine && ProfileStats()) {
     StatAllocHandler = fextl::make_unique<FEX::Windows::StatAlloc>(FEXCore::Profiler::AppType::WIN_WOW64);
   }
+
+  if (StartupSleep() && (StartupSleepProcName().empty() || ExecutablePath == StartupSleepProcName())) {
+    LogMan::Msg::IFmt("[{}][{}] Sleeping for {} seconds", ::getpid(), ExecutablePath, StartupSleep());
```
bylaws, line 516: GetCurrentProcessId

bylaws (changes requested):

bylaws (approved):

PR #4424, opened 2025-03-22
Convert config options once

Sonicadvance1:
Instead of keeping the vlaue as a string array in the MetaLayer, convert the value to its final type once.

Improves performance in some hotpaths that were doing config based string conversion in a relatively high frequency.

### DIFF
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -423,7 +423,7 @@ def print_parse_argloader_options(options):
                 # these need a bit more help
                 output_argloader.write("\tauto Array = Options.all(\"{0}\");\n".format(op_key))
                 output_argloader.write("\tfor (auto iter = Array.begin(); iter != Array.end(); ++iter) {\n")
-                output_argloader.write("\t\tSet(FEXCore::Config::ConfigOption::CONFIG_{0}, *iter);\n".format(op_key.upper()))
+                output_argloader.write("\t\tAppendStrArrayValue(FEXCore::Config::ConfigOption::CONFIG_{0}, *iter);\n".format(op_key.upper()))
                 output_argloader.write("\t}\n")
             else:
                 if (NeedsString):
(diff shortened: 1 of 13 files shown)


Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -447,7 +447,18 @@ class TestEnvLoader final : public FEXCore::Config::Layer {
...
     for (auto& it : EnvConfigLookup) {
       if (auto Value = GetVar(it.first); Value) {
-        Set(it.second, *Value);
+#define OPT_BASE(type, group, enum, json, default) // Nothing
+#define OPT_STRARRAY(group, enum, json, default)                   \
+  if (it.second == FEXCore::Config::ConfigOption::CONFIG_##enum) { \
+    AppendStrArrayValue(it.second, *Value);                        \
+  }
+
+        if (false) {
```
lioncash, line 456: what's with the `if (false)` here?

FEXCore/Scripts/config_generator.py
```diff
@@ -468,6 +468,10 @@ def print_parse_jsonloader_options(options):
                 output_argloader.write("else if (KeyName == \"{0}\") {{\n".format(op_key))
                 output_argloader.write("Set(KeyOption, FEXCore::Config::EnumParser<FEXCore::Config::{}ConfigPair>(FEXCore::Config::{}_EnumPairs, Value_View));\n".format(op_key, op_key, op_key))
                 output_argloader.write("}\n")
+            elif (value_type == "strarray"):
+                output_argloader.write("else if (KeyName == \"{0}\") {{\n".format(op_key))
+                output_argloader.write("AppendStrArrayValue(KeyOption, ConfigString);\n")
```
lioncash, line 473: should this have a `\t` before the append to indent it? It's generated code so it doesn't matter either way, but

Sonicadvance1: Sure, I added a \t there to match some of the other generated behaviour.

Source/Tools/CodeSizeValidation/Main.cpp
Sonicadvance1: Oops, the define should have used an `else if` so the `if (false)` is the start of the chain of generated code hitting `OPT_STRARRAY` and everything else falls down the final `else` path.

Fixed the define.

lioncash (approved):

PR #4429, opened 2025-03-25
RFC: Use round-robin regalloc

alyssarosenzweig:
This helps a bit with instcountci for reasons I haven't fully digested, and it will make post-RA optimizations more effective (what I'm actually looking at.)


alyssarosenzweig:
I cannot reproduce the asm test fails locally.

pmatos:
> I cannot reproduce the asm test fails locally.

I tried it on both my devices and I cannot reproduce the failures here either. :( Slightly surprising...

Sonicadvance1:
Looks like this is only failing if the host compiler doesn't support preserve_all

alyssarosenzweig:
> Looks like this is only failing if the host compiler doesn't support preserve_all

Can we bump our CI machines and require preserve_all and drop the slow path? ..

Sonicadvance1:
> > Looks like this is only failing if the host compiler doesn't support preserve_all
> 
> Can we bump our CI machines and require preserve_all and drop the slow path? ..

preserve_all requires clang-17 as a minimum, which our minspec is currently clang-13. Not sure exactly when we'll bump that.
Bigger issue is that wine/mingw/arm64ec doesn't support preserve_all so we'll need to support the fallback regardless.

alyssarosenzweig:
We were failing to spill/fill x18 across calls, even though it was a dynamic. Ugh, this is a mess.

alyssarosenzweig:
OK, passing CI now. Still trying to figure out how much of a /good/ idea this is.

Sonicadvance1:
> OK, passing CI now. Still trying to figure out how much of a /good/ idea this is.

Could this potentially break the `AllocateAfter` thing we have for sequential register allocation, if it wraps around?

Other than that, I don't really see a downside to doing this.

PR #4433, opened 2025-03-26
Improve JSON file validation and error reporting

pmatos:
Turn invalid JSON files into fatal errors.

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -555,8 +555,8 @@ std::optional<std::vector<FileTargets>> GetRootFSLinks() {
   const json_t* json = FEX::JSON::CreateJSON(Data, Pool);
 
   if (!json) {
-    fprintf(stderr, "Couldn't create json");
-    return {};
+    fmt::print(stderr, "Failed to parse JSON from RootFSLinks file '{}' - invalid JSON format", Data);
+    std::abort();
   }
 
   const json_t* RootList = json_getProperty(json, "v1");
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -31,8 +31,7 @@ namespace JSON {
     const json_t* json = FEX::JSON::CreateJSON(Data, Pool);
 
     if (!json) {
-      LogMan::Msg::EFmt("Couldn't create json");
-      return;
+      ERROR_AND_DIE_FMT("Failed to parse JSON from file '{}' - invalid JSON format", Config);
     }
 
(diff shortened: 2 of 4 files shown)


Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -555,8 +555,8 @@ std::optional<std::vector<FileTargets>> GetRootFSLinks() {
   const json_t* json = FEX::JSON::CreateJSON(Data, Pool);
 
   if (!json) {
-    fprintf(stderr, "Couldn't create json");
-    return {};
+    fmt::print(stderr, "Failed to parse JSON from RootFSLinks file '{}' - invalid JSON format", Data);
+    __builtin_unreachable();
```
neobrain, line 559: The unreachable annotation makes these branches undefined behavior as an optimization hint to the compiler (ironically allowing the print to be removed entirely), which isn't what we're looking for here.

Sadly FEX's error handling isn't particularly versatile, but either `std::abort` or `ERROR_AND_DIE_FMT` are the closest in behavior here.

pmatos: Good point. Hummm, then I guess I shouldn't be using FEX_UNREACHABLE either.

pmatos: Ah, no... the others are fine because EFmt is really meant to abort after printing the message so the unreachable is fine.

Now fixed.

neobrain: > Ah, no... the others are fine because EFmt is really meant to abort after printing the message so the unreachable is fine.

This is really misleading and fragile though. Why not just clearly abort in these cases too, where there's no need to bring UB into play?

Sonicadvance1: `AFmt` does

Sonicadvance1: But also only in builds with `ASSERTIONS_ENABLED` enabled. Needs `ERROR_AND_DIE_FMT` to actually assert in release builds.

pmatos: yeah - i misread this:
```
#define ERROR_AND_DIE_FMT(...)      \
  do {                              \
    LogMan::Msg::EFmt(__VA_ARGS__); \
    FEX_TRAP_EXECUTION;             \
  } while (0)
```

I skimmed this code and assumed EFmt would call FEX_TRAP_EXECUTION (which in turn calls ForcedAssert). Thanks for pointing this out.

Sonicadvance1 (approved):

PR #4460, opened 2025-03-29
FEXCore/Frontend: Remove logically dead code

Sonicadvance1:
This code can't get hit.

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -912,12 +912,6 @@ bool Decoder::DecodeInstruction(uint64_t PC) {
       }
     }
   }
-
-  if (DecodeInst->Dest.IsGPR()) {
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -913,11 +913,7 @@ bool Decoder::DecodeInstruction(uint64_t PC) {
     }
   }
 
-  if (DecodeInst->Dest.IsGPR()) {
-    LOGMAN_THROW_A_FMT(DecodeInst->Dest.Data.GPR.GPR != FEXCore::X86State::REG_INVALID, "Destination GPR was invalid");
-  }
-
-  return true;
+  FEX_UNREACHABLE;
```
neobrain, line 916: We can remove even the `FEX_UNREACHABLE`, can't we? The `function does not return in all control paths` warning typically doesn't fire when all control flow graph routes can easily be checked (like here). Conversely, removing `FEX_UNREACHABLE` will ensure that warning does fire in an erroneous future change.

Sonicadvance1: Sure, removed.

PR #4463, opened 2025-03-29
Telemetry: Removes unnecessary indirection

Sonicadvance1:
Telemetry value address generation was forcing an indirection at all times which was unnecessary. These values live in the BSS, zero initialized at process start and is unnecessary.

Instead change the wrapper defines to directly operate on the enum passed in which saves an indirection on all of these telemetry operations (except for the ones in the JIT which are required to be PIC compliant).

This also fixes an annoying warning about
`FEXCORE_TELEMETRY_STATIC_INIT` causing initialization and destruction order being unspecified, so two wins.


lioncash (approved):
Nice

PR #4468, opened 2025-03-30
FEXCore/Softfloat: Wire up cephes math library for transcendental operations

Sonicadvance1:
This is solving a different problem than what https://github.com/FEX-Emu/FEX/pull/4411 is specifically
trying to solve.

For our transcendental operations, we can't currently guarantee that
these functions will actually operate at the 128-bit softfloat
precision. While this is true with glibc, this is /not/ true for musl
and likely more libraries.

Instead of relying on our libc implementation to implement these,
instead include the cephes math library directly which is what most
people use for this. Including musl even, but not for all operations.

With this we are no longer beholden to the standard libraries for
providing a correct implementation.


Sonicadvance1:
> Is FEX without glibc even a supported build config?

We technically support musl from the few users that asked for it. We also support win32 which makes `long double` only 64-bit there. More work required to get win32 to use the full precision.

bylaws:
> > Is FEX without glibc even a supported build config?
> 
> We technically support musl from the few users that asked for it. We also support win32 which makes `long double` only 64-bit there. More work required to get win32 to use the full precision.

Is there anything known to depend on this? Can have a look into supporting it in llvm if desired

Sonicadvance1:
> > > Is FEX without glibc even a supported build config?
> > 
> > 
> > We technically support musl from the few users that asked for it. We also support win32 which makes `long double` only 64-bit there. More work required to get win32 to use the full precision.
> 
> Is there anything known to depend on this? Can have a look into supporting it in llvm if desired

Issue is just ABI problems. `long double` on 64-bit Windows API is defined to be identical to `double`, which means there is no native support for 128-bit softfloat there.

We will need to more heavily wire in the berkely softfloat library so we have a 128-bit softfloat library for regular math operations, and munge the cephes library to use 128-bit floats rather than `long double`.

With those features in place we should get full precision x87 under Win32 ABI without compiler changes.

Sonicadvance1:
> > > > Is FEX without glibc even a supported build config?
> > > 
> > > 
> > > We technically support musl from the few users that asked for it. We also support win32 which makes `long double` only 64-bit there. More work required to get win32 to use the full precision.
> > 
> > 
> > Is there anything known to depend on this? Can have a look into supporting it in llvm if desired
> 
> Issue is just ABI problems. `long double` on 64-bit Windows API is defined to be identical to `double`, which means there is no native support for 128-bit softfloat there.
> 
> We will need to more heavily wire in the berkely softfloat library so we have a 128-bit softfloat library for regular math operations, and munge the cephes library to use 128-bit floats rather than `long double`.
> 
> With those features in place we should get full precision x87 under Win32 ABI without compiler changes.

Alternative take, we use the 80-bit cephes library and /heavily/ invasively change it to the the Berkely 80-bit softfloat library everywhere and then there is no 80bit->128bit->80bit dance, but that requires some very heavy surgery.

This intermediate step is a good first step without ripping up all the carpet.

Sonicadvance1:
> > > > > Is FEX without glibc even a supported build config?
> > > > 
> > > > 
> > > > We technically support musl from the few users that asked for it. We also support win32 which makes `long double` only 64-bit there. More work required to get win32 to use the full precision.
> > > 
> > > 
> > > Is there anything known to depend on this? Can have a look into supporting it in llvm if desired
> > 
> > 
> > Issue is just ABI problems. `long double` on 64-bit Windows API is defined to be identical to `double`, which means there is no native support for 128-bit softfloat there.
> 
> I mean changing the abi shouldn't be a problem here, imagine it's much easier to solve there than a fex side solution if it's already done for other abis.

ABI isn't the only problem, win32 would also need to provide all the 128-bit float math operations. add, sub, div, mul, comparisons, integer conversions. Would need to provide all those as well, which is already stepping in to doing a `%s/long double/__float128_t` in the cephes math library and having that `float128_t` type use the Berkely soft float math operations for those. Right now we're leaning on libc to provide those math operations with the expectation that it'll be full 128-bit.

I don't think compiler changes are really worth diving in to for it.

lioncash:
Is there anything stopping cephes from being in the External folder to maintain the separation from core files? Not a huge deal either way, but it would be cool if it was doable

Sonicadvance1:
> Is there anything stopping cephes from being in the External folder to maintain the separation from core files? Not a huge deal either way, but it would be cool if it was doable

Sure thing, moved it over.

.clang-format-ignore
```diff
@@ -13,3 +13,6 @@ FEXCore/Source/Interface/Core/X86Tables/*
 # Inline headers with list-like content that can't be processed individually
 Source/Tools/LinuxEmulation/LinuxSyscalls/x*/SyscallsNames.inl
 Source/Tools/LinuxEmulation/LinuxSyscalls/x*/Ioctl/*.inl
+
+# cephes math library
+FEXCore/Source/Common/cephes/*
```
lioncash, line 18: Only nit seems to be that this can be removed

Sonicadvance1: Oop, removed.

lioncash (approved):

PR #4469, opened 2025-03-31
Initialize OutputFS to -1

pmatos:
Avoids the non-fatal error: "[ERROR] Close closing FEX FD 2" that happens when the guest program executes a syscall to close fd 2, and it's not in fex tracked set.

### DIFF
--- a/Source/Tools/FEXLoader/FEXLoader.cpp
+++ b/Source/Tools/FEXLoader/FEXLoader.cpp
@@ -529,7 +529,7 @@ int main(int argc, char** argv, char** const envp) {
   auto VDSOMapping = FEX::VDSO::LoadVDSOThunks(Loader.Is64BitMode(), SyscallHandler.get());
 
   // Now that we have the syscall handler. Track some FDs that are FEX owned.
-  if (OutputFD != -1) {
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Can we instead change the check that adds the FD to tracking to only add it if the FD is `> 2`?
Down here: https://github.com/FEX-Emu/FEX/blob/3c11aba4378effd2f91993ec1d966c8bd5eeb484/Source/Tools/FEXLoader/FEXLoader.cpp#L532

I change the defaultFD from -1 a couple months ago so our early logs gets dumped in to stderr by default and this would break that.

pmatos:
> Can we instead change the check that adds the FD to tracking to only add it if the FD is `> 2`? Down here:
> 
> https://github.com/FEX-Emu/FEX/blob/3c11aba4378effd2f91993ec1d966c8bd5eeb484/Source/Tools/FEXLoader/FEXLoader.cpp#L532
> 
> I change the defaultFD from -1 a couple months ago so our early logs gets dumped in to stderr by default and this would break that.

Sure - done.

Sonicadvance1 (approved):

PR #4470, opened 2025-03-31
Opt/cvtss2si smash

alyssarosenzweig:


unittests/InstructionCountCI/VEX_map1_FRRINTS.json
```diff
@@ -0,0 +1,249 @@
+{
```
Sonicadvance1, line 1: You probably want this file to be the one in the AVX128 directory path instead of the SVE256 path. Since SVE256 is low-priority.

alyssarosenzweig: The AVX128 one is FRRINTS, we have three copies of VEX_map1 for the convert instrs after. (We'd need a fourth for full coverage, the one combo we miss is "AVX128 and no VEX_map1" but w/e).

PR #4471, opened 2025-03-31
Optimize float->integer conversions with Feat_FRINTTS

alyssarosenzweig:
https://github.com/FEX-Emu/FEX/pull/4237 fixed our float->int conversions to match x86 semantics. That fixed a bunch of games, but unfortunately significantly regressed performance. Fortunately, when frintts is supported, we can optimize the sequence using the appropriate arm instruction. This should recover most of the perf lost by #4237.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -59,6 +59,8 @@
           "DISABLEFLAGM": "disableflagm",
           "ENABLEFLAGM2": "enableflagm2",
           "DISABLEFLAGM2": "disableflagm2",
+          "ENABLEFRINTTS": "enablefrintts",
+          "DISABLEFRINTTS": "disablefrintts",
           "ENABLECRYPTO": "enablecrypto",
           "DISABLECRYPTO": "disablecrypto",
           "ENABLERPRES": "enablerpres",
--- a/FEXCore/Source/Interface/Core/JIT/ConversionOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/ConversionOps.cpp
@@ -6,6 +6,7 @@ $end_info$
 */
 
 #include "Interface/Core/JIT/JITClass.h"
+#include "Interface/Context/Context.h"
 
 namespace FEXCore::CPU {
 #define DEF_OP(x) void Arm64JITCore::Op_##x(IR::IROp_Header const* IROp, IR::NodeID Node)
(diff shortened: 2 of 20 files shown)


alyssarosenzweig:
Not sure what to do about the zeroupperbits vs !zeroupperbits path but this passes CI so we evidently need more unit tests to expose whatever bug is there... Cc @bylaws do you remember the details here? git blames to you

Sonicadvance1 (approved):
\o/ :tada: 

I had taken a peek at these instructions and noticed it would save some instructions, I didn't realize how much.

Also somewhat annoyingly, it doesn't look like SVE has equivalent instructions? So SVE-256bit implementation would be splitting the 128-bit halves and doing the regular ASIMD path. It's really looking like if SVE-256bit chips ever land that we are going to need to have SVE2.1 as the minspec, since it adds a bunch of helpers for this behaviour.

alyssarosenzweig:
> 🎉 \o/
> 
> I had taken a peek at these instructions and noticed it would save some instructions, I didn't realize how much.
> 
> Also somewhat annoyingly, it doesn't look like SVE has equivalent instructions? So SVE-256bit implementation would be splitting the 128-bit halves and doing the regular ASIMD path. It's really looking like if SVE-256bit chips ever land that we are going to need to have SVE2.1 as the minspec, since it adds a bunch of helpers for this behaviour.

The important instruction here (frint32z) should exist for SVE too, at least it seemed to when i skimmed the arm docs for it yesterday. The i64->i32 truncate I don't know how to do with sve-256 but I'm sure there's an equivalent, it's a very basic operation.

alyssarosenzweig:
> > 🎉 \o/
> > I had taken a peek at these instructions and noticed it would save some instructions, I didn't realize how much.
> > Also somewhat annoyingly, it doesn't look like SVE has equivalent instructions? So SVE-256bit implementation would be splitting the 128-bit halves and doing the regular ASIMD path. It's really looking like if SVE-256bit chips ever land that we are going to need to have SVE2.1 as the minspec, since it adds a bunch of helpers for this behaviour.
> 
> The important instruction here (frint32z) should exist for SVE too, at least it seemed to when i skimmed the arm docs for it yesterday. The i64->i32 truncate I don't know how to do with sve-256 but I'm sure there's an equivalent, it's a very basic operation.

(I didn't actually wire up the sve-256 path because it would've required plumbing a bunch of stuff for no real benefit right now but it should be theoretically possible if hw ever lands for it.)

bylaws:
> Not sure what to do about the zeroupperbits vs !zeroupperbits path but this passes CI so we evidently need more unit tests to expose whatever bug is there... Cc @bylaws do you remember the details here? git blames to you

I think it's an opt, I carried forward from what was there before so not particularly sure

alyssarosenzweig:
> > Not sure what to do about the zeroupperbits vs !zeroupperbits path but this passes CI so we evidently need more unit tests to expose whatever bug is there... Cc @bylaws do you remember the details here? git blames to you
> 
> I think it's an opt, I carried forward from what was there before so not particularly sure

In that case I guess this should be good as is then.

PR #4474, opened 2025-03-31
Support multiple entrypoints into a multiblock and executable permission tracking

bylaws:
This allows a single JIT call to translate the entire bodies of most functions not using indirect jumps, previously multiblocks would end at calls which often led to regions being translated multiple times as in e.g.:

```
if (a)
  do_b()
a = false; // <-- this would be translated twice prior to this if ran with both a=false then a=true
...
```

This paves the way for call-ret optimisations and decreases MB-enabled nodejs startup JIT time by ~8%.

Executable permission tracking is necessary to avoid following a non-taken return location into unmapped memory - this occurs in many unity games.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -557,9 +557,9 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
     bool HadDispatchError {false};
     bool HadInvalidInst {false};
 
-    Thread->FrontendDecoder->DecodeInstructionsAtEntry(GuestCode, GuestRIP, MaxInst,
-                                                       [Thread](uint64_t BlockEntry, uint64_t Start, uint64_t Length) {
-      if (Thread->LookupCache->AddBlockExecutableRange(BlockEntry, Start, Length)) {
+    Thread->FrontendDecoder->DecodeInstructionsAtEntry(
+      GuestCode, GuestRIP, MaxInst, [Thread](const fextl::set<uint64_t>& BlockEntryPoints, uint64_t Start, uint64_t Length) {
+      if (Thread->LookupCache->AddBlockExecutableRange(BlockEntryPoints, Start, Length)) {
         static_cast<ContextImpl*>(Thread->CTX)->SyscallHandler->MarkGuestExecutableRange(Thread, Start, Length);
       }
     });
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -412,7 +412,7 @@ void ContextImpl::InitializeCompiler(FEXCore::Core::InternalThreadState* Thread)
   Thread->OpDispatcher = fextl::make_unique<FEXCore::IR::OpDispatchBuilder>(this);
   Thread->OpDispatcher->SetMultiblock(Config.Multiblock);
   Thread->LookupCache = fextl::make_unique<FEXCore::LookupCache>(this);
-  Thread->FrontendDecoder = fextl::make_unique<FEXCore::Frontend::Decoder>(this);
+  Thread->FrontendDecoder = fextl::make_unique<FEXCore::Frontend::Decoder>(Thread);
   Thread->PassManager = fextl::make_unique<FEXCore::IR::PassManager>();
 
   Thread->CurrentFrame->Pointers.Common.L1Pointer = Thread->LookupCache->GetL1Pointer();
@@ -557,12 +557,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
     bool HadDispatchError {false};
     bool HadInvalidInst {false};
 
-    Thread->FrontendDecoder->DecodeInstructionsAtEntry(
-      GuestCode, GuestRIP, MaxInst, [Thread](const fextl::set<uint64_t>& BlockEntryPoints, uint64_t Start, uint64_t Length) {
-      if (Thread->LookupCache->AddBlockExecutableRange(BlockEntryPoints, Start, Length)) {
-        static_cast<ContextImpl*>(Thread->CTX)->SyscallHandler->MarkGuestExecutableRange(Thread, Start, Length);
-      }
-    });
+    Thread->FrontendDecoder->DecodeInstructionsAtEntry(GuestCode, GuestRIP, MaxInst);
 
     auto BlockInfo = Thread->FrontendDecoder->GetDecodedBlockInfo();
     auto CodeBlocks = &BlockInfo->Blocks;
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -750,7 +750,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
       auto CompiledCode = Thread->CPUBackend->RelocateJITObjectCode(GuestRIP, CodeCacheEntry);
       if (CompiledCode) {
         return {
-          .CompiledCode = CompiledCode,
+          .CompiledCode = {},
           .DebugData = nullptr, // nullptr here ensures that code serialization doesn't occur on from cache read
           .StartAddr = 0,       // Unused
           .Length = 0,          // Unused
@@ -769,7 +769,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   // Generate IR + Meta Info
   auto [IRView, TotalInstructions, TotalInstructionsLength, StartAddr, Length] = GenerateIR(Thread, GuestRIP, Config.GDBSymbols(), MaxInst);
   if (!IRView) {
-    return {nullptr, nullptr, 0, 0};
+    return {{}, nullptr, 0, 0};
   }
 
   // Attempt to get the CPU backend to compile this code
@@ -780,7 +780,10 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   if (MaxInst != 1) {
     if (auto Block = Thread->LookupCache->FindBlock(GuestRIP)) {
       Thread->OpDispatcher->DelayedDisownBuffer();
-      return {.CompiledCode = reinterpret_cast<uint8_t*>(Block), .DebugData = nullptr, .StartAddr = 0, .Length = 0};
+      return {.CompiledCode = {.BlockBegin = reinterpret_cast<uint8_t*>(Block), .EntryPoints = {{GuestRIP, reinterpret_cast<uint8_t*>(Block)}}},
+              .DebugData = nullptr,
+              .StartAddr = 0,
+              .Length = 0};
     }
   }
 
@@ -795,10 +798,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   Thread->OpDispatcher->DelayedDisownBuffer();
 
   return {
-    // FEX currently throws away the CPUBackend::CompiledCode object other than the entrypoint
-    // In the future with code caching getting wired up, we will pass the rest of the data forward.
-    // TODO: Pass the data forward when code caching is wired up to this.
-    .CompiledCode = CompiledCode.BlockEntry,
+    .CompiledCode = std::move(CompiledCode),
     .DebugData = std::move(DebugData),
     .StartAddr = StartAddr,
     .Length = Length,
@@ -821,7 +821,8 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
     return HostCode;
   }
 
-  auto [CodePtr, DebugData, StartAddr, Length] = CompileCode(Thread, GuestRIP, MaxInst);
+  auto [CompiledCode, DebugData, StartAddr, Length] = CompileCode(Thread, GuestRIP, MaxInst);
+  auto CodePtr = CompiledCode.EntryPoints[GuestRIP];
   if (CodePtr == nullptr) {
     return 0;
   } else if (!DebugData) {
@@ -831,7 +832,7 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
 
   // The core managed to compile the code.
   if (Config.BlockJITNaming()) {
-    auto FragmentBasePtr = reinterpret_cast<uint8_t*>(CodePtr);
+    auto FragmentBasePtr = CompiledCode.BlockBegin;
 
     if (DebugData) {
       auto GuestRIPLookup = SyscallHandler->LookupAOTIRCacheEntry(Thread, GuestRIP);
@@ -840,7 +841,7 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
         for (auto& Subblock : DebugData->Subblocks) {
           auto BlockBasePtr = FragmentBasePtr + Subblock.HostCodeOffset;
           if (GuestRIPLookup.Entry) {
-            Symbols.Register(Thread->SymbolBuffer.get(), BlockBasePtr, DebugData->HostCodeSize, GuestRIPLookup.Entry->Filename,
+            Symbols.Register(Thread->SymbolBuffer.get(), BlockBasePtr, CompiledCode.Size, GuestRIPLookup.Entry->Filename,
                              GuestRIP - GuestRIPLookup.VAFileStart);
           } else {
             Symbols.Register(Thread->SymbolBuffer.get(), BlockBasePtr, GuestRIP, Subblock.HostCodeSize);
@@ -848,10 +849,10 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
         }
       } else {
         if (GuestRIPLookup.Entry) {
-          Symbols.Register(Thread->SymbolBuffer.get(), FragmentBasePtr, DebugData->HostCodeSize, GuestRIPLookup.Entry->Filename,
+          Symbols.Register(Thread->SymbolBuffer.get(), FragmentBasePtr, CompiledCode.Size, GuestRIPLookup.Entry->Filename,
                            GuestRIP - GuestRIPLookup.VAFileStart);
         } else {
-          Symbols.Register(Thread->SymbolBuffer.get(), FragmentBasePtr, GuestRIP, DebugData->HostCodeSize);
+          Symbols.Register(Thread->SymbolBuffer.get(), FragmentBasePtr, GuestRIP, CompiledCode.Size);
         }
       }
     }
@@ -864,8 +865,8 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
         .GuestRIP = GuestRIP,
         .GuestCodeLength = Length,
         .GuestCodeHash = 0,
-        .HostCodeBegin = CodePtr,
-        .HostCodeLength = DebugData->HostCodeSize,
+        .HostCodeBegin = CompiledCode.BlockBegin,
+        .HostCodeLength = CompiledCode.Size,
         .HostCodeHash = 0,
         .ThreadJobRefCount = &Thread->ObjectCacheRefCounter,
         .Relocations = std::move(*DebugData->Relocations),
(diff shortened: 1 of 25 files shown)


bylaws:
DNM - block permission tracking is necessary here for unity games to work

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -1092,10 +1094,12 @@ const uint8_t* Decoder::AdjustAddrForSpecialRegion(const uint8_t* _InstStream, u
   return _InstStream - EntryPoint + RIP;
 }
 
-void Decoder::DecodeInstructionsAtEntry(const uint8_t* _InstStream, uint64_t PC, uint64_t MaxInst,
-                                        std::function<void(uint64_t BlockEntry, uint64_t Start, uint64_t Length)> AddContainedCodePage) {
+void Decoder::DecodeInstructionsAtEntry(
+  const uint8_t* _InstStream, uint64_t PC, uint64_t MaxInst,
+  std::function<void(fextl::set<uint64_t> BlockEntryPoints, uint64_t Start, uint64_t Length)> AddContainedCodePage) {
```
neobrain, line 1099: `BlockEntryPoints` should be a reference parameter, shouldn't it? I'm mildly surprised this compiles, since the lambda you pass for `AddContainedCodePage` below does use a reference.

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -341,6 +341,8 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
 
   void EmitInterruptChecks(bool CheckTF);
 
+  void EmitEntryPoint(ARMEmitter::BackwardLabel* HeaderLabel, bool CheckTF);
```
neobrain, line 344: `HeaderLabel` can never be nullptr, so it should be a reference parameter for clarity.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -89,8 +89,8 @@ class LookupCache {
   void AddBlockMapping(uint64_t Address, void* HostCode) {
     std::lock_guard<std::recursive_mutex> lk(WriteLock);
 
-    [[maybe_unused]] auto Inserted = BlockList.emplace(Address, (uintptr_t)HostCode).second;
-    LOGMAN_THROW_A_FMT(Inserted, "Duplicate block mapping added");
+    // This may replace an existing mapping
+    BlockList.emplace(Address, (uintptr_t)HostCode);
```
neobrain, line 93: Judging by the comment, is the intent here to *override* the previous HostCode value? `emplace` won't do this, so you have to use `operator[]` instead.

bylaws: Yep - thanks for spotting

bylaws:
I can split this up slightly if preferred but keeping executable range handling after multiple entrypoint handling saves a bunch of conflicts

bylaws:
Should be good for review now from e0f2ec2de2e2c7716209395e6ba422ef51351a7b onwards, may still have issues with thunks though which aren't tested in CI - I need to retest.

May regress a few 32-bit programs with WOW64 - needs some wine/fex patches and I didn't run into issues with anything I tested so I think thats fine to do after

bylaws:
should work with thunks now - note https://github.com/FEX-Emu/FEX/commit/adcefe62cfc159e5e9d4885b6cfdf59d9b811180 means instcountci wont display code after CALL instructions anymore but perhaps that is desirable and a different PR could just remove the CALLs from the x87 tests?

neobrain:
This now breaks Celeste startup for me (likely among some other issues, like sporadic failure of Steam startup). 435250a37052e975a3161c1101eac1703edbfa3d works fine (launching the game in <20 seconds), whereas the game gets stuck before launching a window in f404139f292ce5a7337844873e2787aec7c621e7 (couldn't test the commits between due to build failures).

bylaws:
> This now breaks Celeste startup for me (likely among some other issues, like sporadic failure of Steam startup). [a378f93](https://github.com/FEX-Emu/FEX/commit/a378f9313bc9ff4480b74c97e86fd3389490237f) works fine (launching the game in <20 seconds), whereas the game gets stuck before launching a window in [c05a2f1](https://github.com/FEX-Emu/FEX/commit/c05a2f1dccc2f9a1025d1bc74e66cc0b4eae4817) (couldn't test the commits between due to build failures).

Fixed

Source/Windows/Common/InvalidationTracker.cpp
```diff
@@ -177,4 +177,18 @@ bool InvalidationTracker::HandleRWXAccessViolation(uint64_t FaultAddress) {
...
+  std::shared_lock Lock(IntervalsLock);
+  const auto XResult = XIntervals.Query(Address);
+  if (!XResult.Enclosed) {
+    return {};
+  }
+  const auto RWXResult = RWXIntervals.Query(Address);
+  if (RWXResult.Enclosed || (RWXResult.Size && RWXResult.Size < XResult.Size)) {
+    return {RWXResult.Size, RWXResult.Enclosed};
+  }
+  return {XResult.Size, false};
```
Sonicadvance1, line 191: Am I misreading this or is the size returned between the Linux SMC tracking and this tracking different?
It looks like this WIN32 implementation returns the full size of the region that contains the address, while the Linux SMC side is returning the size of the region minus the offset of the address queried? So full size on Win32 versus Remaining size on Linux?

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -64,33 +67,75 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
 
+bool Decoder::CheckRangeExecutable(uint64_t Address, uint64_t Size) {
+  // Treat FEX-internal X86 callbacks as always executable
+  if (EntryPoint == CTX->X86CodeGen.CallbackReturn) {
+    return true;
+  }
+
+  while (Address < ExecutableRangeBase || Address + Size > ExecutableRangeEnd) {
+    auto RangeInfo = CTX->SyscallHandler->QueryGuestExecutableRange(Thread, Address);
+    ExecutableRangeBase = Address;
```
Sonicadvance1, line 88: Depending on the implementation details in the frontend here (with the potential for one returning size of the range versus remaining range), should `QueryGuestExecutableRange` return the full VMA range for the queried address? This way ExecutableRangeBase can be set to the VMA range instead of `Address`, reducing the number of frontend queries if code is jumping locally within a single mapped VMA?

Source/Windows/Common/InvalidationTracker.cpp
bylaws: Query.Size is the size from the address queried, not the size of the region

FEXCore/Source/Interface/Core/Frontend.cpp
bylaws: This seems reasonable, shall do

Source/Windows/Common/InvalidationTracker.cpp
Sonicadvance1: Alright cool, that clears up the confusion.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -749,6 +769,7 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
   this->RAData = RAData;
   this->DebugData = DebugData;
   this->IR = IR;
+  CodeData.EntryPoints.clear();
```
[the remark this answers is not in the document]
bylaws: I think I prefer it being explicit here - much like how in the frontend how state is reset prior

FEXCore/Source/Interface/Core/Frontend.h
```diff
@@ -112,6 +120,7 @@ class Decoder final {
   fextl::set<uint64_t> BlocksToDecode;
   fextl::set<uint64_t> VisitedBlocks;
   fextl::set<uint64_t>* ExternalBranches {nullptr};
+  fextl::set<uint64_t> BlockEntryPoints;
```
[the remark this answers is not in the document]
bylaws: When I tried using unordered map for some things here prior I saw a measurable perf decrease - realistically I'm not sure this case will be measurable, will check in nodejs but otherwise keep as is for consistency.

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -64,33 +67,75 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
   PoolObject.UnclaimBuffer();
 }
 
+bool Decoder::CheckRangeExecutable(uint64_t Address, uint64_t Size) {
+  // Treat FEX-internal X86 callbacks as always executable
+  if (EntryPoint == CTX->X86CodeGen.CallbackReturn) {
+    return true;
+  }
+
+  while (Address < ExecutableRangeBase || Address + Size > ExecutableRangeEnd) {
```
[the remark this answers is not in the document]
bylaws: End is exclusive here, Reading 0xfff of size 1 shouldn't perform a new query

neobrain:
How difficult would it be to add a central registry for blocks pending compilation that other threads can wait on?

While rebasing #4479 on this, I realized the huge number of combined blocks compiled at once now will bottleneck parallel compilation in scenarios where many threads attempt to compile overlapping areas of code. With each individual compilation requiring much more time to complete than before (with smaller multiblocks), there's a greater chance that other threads will reach an overlapping part of the program and hence compile an almost matching multiblock in parallel (whereas previously it might have seen a LookupCache hit). 

In #4479, even when minimizing the time the CodeBufferWriteMutex is held (by compiling to a temporary CodeBuffer), the remaining mutex-protected work (relocating temporary CodeBuffer and appending entry points to LookupCache) adds up to critical amount of lock contention.

If we could have each thread register the set of entry points it discovered in the frontend phase before kicking off the backend, this problem could be avoided by having other threads wait for that first thread instead.

neobrain:
I found a way to reduce the lock contention in 4479 even further, so my previous comment isn't critical as originally observed. Once the two PRs are merged, a central registry still seems beneficial though to minimize redundant work in general.

neobrain:
To adapt this PR for the now-merged #4479, two changes are needed:

JIT.cpp:
```diff
     // Adjust host addresses
     const auto Delta = GetCursorAddress<uint8_t*>() - CodeData.BlockBegin;
     CodeData.BlockBegin += Delta;
-    CodeData.BlockEntry += Delta;
+    for (auto& EntryPoint : CodeData.EntryPoints) {
+      EntryPoint.second += Delta;
+    }
```

Core.cpp:
```diff
@@ -785,7 +785,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   if (MaxInst != 1) {
     if (auto Block = Thread->LookupCache->FindBlock(GuestRIP)) {
       Thread->OpDispatcher->DelayedDisownBuffer();
-      return {.CompiledCode = reinterpret_cast<uint8_t*>(Block), .DebugData = nullptr, .StartAddr = 0, .Length = 0};
+      return {.CompiledCode = {.BlockBegin = reinterpret_cast<uint8_t*>(Block), .EntryPoints = {{GuestRIP, reinterpret_cast<uint8_t*>(Block)}} }, .DebugData = nullptr, .StartAddr = 0, .Length = 0};
     }
   }
 ```

File: FEXCore/Source/Interface/Core/Frontend.cpp [hunk 9]
```diff
@@ -64,33 +67,75 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
-uint8_t Decoder::PeekByte(uint8_t Offset) const {
-  uint8_t Byte = InstStream[InstructionSize + Offset];
+uint8_t Decoder::PeekByte(uint8_t Offset) {
+  uint8_t Byte = 0;
+  uint64_t ByteAddress = reinterpret_cast<uint64_t>(InstStream + InstructionSize + Offset);
+  if (CheckRangeExecutable(ByteAddress, 1)) {
+    Byte = InstStream[InstructionSize + Offset];
+  } else {
+    HitNonExecutableRange = true;
+    // Pretend we read 0, the main decode loop will see HitNonExecutableRange and rollback the instruction.
```
neobrain: In my code caching branch (which integrates this PR), I keep running into different issues that lead me back to this line. The bugs are on the code caching branch, but they are difficult to debug because returning 0 here silently swallows potential errors and defers them until an unspecific later point in time. For example:

* I had implemented QueryExecutableRange implemented slightly incorrectly, causing some instructions to be replaced with null bytes
* I compiled valid x86 code that simply hadn't be marked as `PROT_EXEC` yet, causing FEX to just emit a panic block

Are there any safeguards we can put in place around this code to detect invalid scenarios more eagerly?

bylaws: I think your issue itself is effectively desired behaviour if multiblock is enabled, we don't really want a warning in the log when this case happens so often and is perfectly fine. Perhaps if multiblock is disabled and the block hits a decode failure we could print?

neobrain: > warning in the log

I'm not too interested in adding likely false-positives to the log either, but more in situations where we definitely know something went wrong and hence can loudly fail. This could well be a heuristic approach, i.e. any combination of trivial `if` statements that doesn't hinder normal multiblock operation but eases debugging.

> I think your issue itself is effectively desired behaviour if multiblock is enabled

Is this unconditionally true? What about the case where we haven't successfully decoded any instructions yet - is there a valid use case for allowing this?

> Perhaps if multiblock is disabled and the block hits a decode failure we could print?

Is there a reason not to panic in this case?

bylaws: > 
> Is this unconditionally true? What about the case where we haven't successfully decoded any instructions yet - is there a valid use case for allowing this?

Hmm yeah, if it's a decode failure in the entrypoint block that is another case we could log at.

> 
> > Perhaps if multiblock is disabled and the block hits a decode failure we could print?
> 
> Is there a reason not to panic in this case?

Yeah - the game could catch the fault.

bylaws:
Ready for testing now - I *think* I have UD cases that would cause FEX crashes handled

Sonicadvance1:
Looks like this needs a rebase and a look in to why instcountci isn't happy with the changes.

bylaws:
should be ready now

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -64,29 +67,72 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
+  , OSABI {CTX->SyscallHandler ? CTX->SyscallHandler->GetOSABI() : FEXCore::HLE::SyscallOSABI::OS_UNKNOWN}
+  , PoolObject {CTX->FrontendAllocator, sizeof(FEXCore::X86Tables::DecodedInst) * DefaultDecodedBufferSize} {}
+
+bool Decoder::CheckRangeExecutable(uint64_t Address, uint64_t Size) {
+  // Treat FEX-internal X86 callbacks as always executable
+  if (EntryPoint == CTX->X86CodeGen.CallbackReturn) {
+    return true;
+  }
+
+  while (Address < ExecutableRangeBase || Address + Size > ExecutableRangeEnd) {
```
neobrain, line 82: IIUC the way ExecutableRangeBase/End are used as a sort of cache here can result in incorrect results immediately after unloading a library. Perhaps not the most likely thing to happen in practice, but it's easy to fix by resetting ExecutableRangeEnd... somewhere.

neobrain: Alternatively, maybe it's cleaner design to move the caching of the last result into the SyscallHandler (which can also invalidate the cached result easily). We can then unconditionally call QueryGuestExecutableRange at least once here to avoid stale caches.

neobrain:
The updated version breaks Celeste startup under Linux again for me.

bylaws:
> The updated version breaks Celeste startup under Linux again for me.

Rip, could you grab a log

FEXCore/Source/Interface/Core/Frontend.cpp
bylaws: I thought about this prior but decided to leave it since it mattering in practice is so unlikely. Having this in syscall handler would be suboptimal though, I would expect it to have a measurable perf impact. I think I can quite easily tie this to invalidate range calls so will do that

neobrain: > Having this in syscall handler would be suboptimal though, I would expect it to have a measurable perf impact

If you cache the last query (like done here already), why would there be a performance impact?

bylaws: > If you cache the last query (like done here already), why would there be a performance impact?

PeekByte is very hot - it would prevent any inlining and add a virtual call to it - in addition to duplicating the logic in two places. The approach I mentioned with invalidating this in invalidate though should work fine and I think is preferable

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -64,29 +67,72 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
+    ExecutableRangeEnd = RangeInfo.Base + RangeInfo.Size;
+
+    if (RangeInfo.Size == 0) {
+      return false;
+    }
+
+    if (Size > RangeInfo.Size) {
+      Size -= RangeInfo.Size;
+      Address += RangeInfo.Size;
+    }
```
neobrain: This seems completely off now, perhaps because of a QueryGuestExecutableRange semantics change?

Consider this:
```
Base  Address End
0x1000 -- 0x1ffc -- 0x2000
```
With Size = 8, `Address` will never actually be changed, freezing us in an infinite loop.

Further, the increment/decrement shouldn't be `RangeInfo.Size` but the actual distance of Address to the end.

neobrain:
> Rip, could you grab a log

No useful information in the logs, but gdb says it gets stuck in a CheckRangeExecutable loop:
```
ExecutableRangeBase = 0x7ffbc000
ExecutableRangeEnd  = 0x7ffcc000
Address             = 0x7ffcbffb
Size = 8
```
Looks like the increment at the end of the while loop is wrong, added a comment about it.

FEXCore/Source/Interface/Core/Frontend.cpp
bylaws: Ah yeah the semantics change broke this

FEXCore/Source/Interface/Core/Frontend.cpp
neobrain: > PeekByte is very hot - it would prevent any inlining and add a virtual call to it - in addition to duplicating the logic in two places

Did you make measurements to back up this impact? The virtual call is trivially predictable even if devirtualization didn't apply, and I'm not sure the function gets inlined as-is. The logic is only needed for the main SyscallHandler implementation (since DummySyscallHandler is trivial).

Managing the state in Frontend.cpp can work too, I'm just not a fan of this argument. Arguably doing so in Frontend.cpp has the upside that a future refactor could avoid byte-wise reading entirely instead of adding band-aid optimizations in the lower layers.

bylaws: > Did you make measurements to back up this impact? The virtual call is trivially predictable even if devirtualization didn't apply, and I'm not sure the function gets inlined as-is. The logic is only needed for the main SyscallHandler implementation (since DummySyscallHandler is trivial).

Adding these checks alone adds a measurable few % to decode time, and with how simple they are to cause that yeah I'd assume extra indirection here would likely matter. Though of course I've not benchmarked. There are two syscall handler implementations, one for windows one for Linux hence this would be duplicated code

neobrain: > There are two syscall handler implementations, one for windows one for Linux hence this would be duplicated code

My apologies, I was grepping from the wrong directory before 🙈 

I would assume the overhead you saw would be mainly from cache misses (i.e. unavoidable full lookups via QueryGuestExecutableRange), but alright - let's go with the implementation in Frontend.cpp.

bylaws:
> > Rip, could you grab a log
> 
> No useful information in the logs, but gdb says it gets stuck in a CheckRangeExecutable loop:
> 
> ```
> ExecutableRangeBase = 0x7ffbc000
> ExecutableRangeEnd  = 0x7ffcc000
> Address             = 0x7ffcbffb
> Size = 8
> ```
> 
> Looks like the increment at the end of the while loop is wrong, added a comment about it.

Should be fixed now

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -908,6 +907,10 @@ uintptr_t ContextImpl::CompileSingleStep(FEXCore::Core::CpuStateFrame* Frame, ui
 }
 
 static void InvalidateGuestThreadCodeRange(FEXCore::Core::InternalThreadState* Thread, uint64_t Start, uint64_t Length) {
+  // Safe as the thread's code invalidation mutex must be locked here, ensures now-modified mappings aren't cached as
+  // being in their previous state.
```
neobrain: This is difficult to understand due to using fragments instead of full sentences. Also you're explaining why this is safe but it's unclear what the concern is to begin with.

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -64,29 +67,73 @@ static uint32_t MapVEXToReg(uint8_t vvvv, bool HasXMM) {
...
+  while (Address < ExecutableRangeBase || Address + Size > ExecutableRangeEnd) {
+    auto RangeInfo = CTX->SyscallHandler->QueryGuestExecutableRange(Thread, Address);
+    ExecutableRangeBase = RangeInfo.Base;
+    ExecutableRangeEnd = RangeInfo.Base + RangeInfo.Size;
+
+    if (RangeInfo.Size == 0) {
+      return false;
+    }
+
+    uint64_t RangeRemainingSize = RangeInfo.Size - (Address - ExecutableRangeBase);
```
neobrain:
```suggestion
    uint64_t RangeRemainingSize = ExecutableRangeEnd - Address;
```

Sonicadvance1:
I'll do some smoketesting today for this.

PR #4475, opened 2025-04-01
x87OptimizationPass: Fixes {Inc,Dec}StackPop

Sonicadvance1:
On the slow path these were pushing and popping in the wrong direction.
Switch them around to ensure the unittests work.

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
@@ -1044,7 +1044,7 @@ void X87StackOptimization::Run(IREmitter* Emit) {
 
       case OP_INCSTACKTOP: {
         if (SlowPath) {
-          UpdateTopForPush_Slow();
+          UpdateTopForPop_Slow();
         } else {
           StackData.rotate(false);
         }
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
This might also fix Far Cry? But I haven't tested it in a while so it could have been fixed before this.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -1044,7 +1044,7 @@ void X87StackOptimization::Run(IREmitter* Emit) {
 
       case OP_INCSTACKTOP: {
         if (SlowPath) {
-          UpdateTopForPush_Slow();
+          UpdateTopForPop_Slow();
```
pmatos, line 1047: Good catch... 
Not only I can't believe I missed this, but also that it was not caught by other tests.

pmatos:
LGTM.

pmatos:
I was very confused about why D9_F7.asm was not failing before or after the patch but I just noticed that we are not actually testing fincstp there, although that's the whole purpose of the test. :facepalm: 

I will patch this (and related) tests next.

Sonicadvance1:
There we go. Updated the fix commit message. This is good to go.

pmatos:
> There we go. Updated the fix commit message. This is good to go.

Yes, yes, and yes. Drop it like it's hot!

lioncash (approved):

PR #4479, opened 2025-04-02
Reduce JIT time by 25% by sharing code buffers between threads

neobrain:
## Overview

To ease state management, FEX currently dedicates a separate CodeBuffer to each thread for storing the output of JIT binary recompilation. Effectively, this means lots of code must be compiled multiple times, particularly during startup and loadscreens of an application. This PR fixes this by sharing CodeBuffers between threads wherever possible, drastically reducing JIT time and mitigating microstutters (see below for benchmarks).

The broader goal here is that CodeBuffer memory should never be discarded unless strictly necessary, establishing the invariant "every guest block is compiled exactly once". This will also prepare us for effective on-disk code caching (which will further reduce startup times).

## Implementation: Partially persistent data structures

> ... partially what now?

This is an [idea from functional programming](https://en.wikipedia.org/wiki/Persistent_data_structure) to ensure thread-safe data access: Trying to modify an object will branch its state into a new version for the active thread but preserving the old version as a read-only copy in other threads. This can be implemented without CPU overhead as long as write accesses are mutex-protected.

This PR turns CodeBuffer into such a partially persistent data structure. In practice this means:
* Exactly one CodeBuffer is now designated as "active", which means data can be *appended* to it
* Lossy modifications to the active CodeBuffer will not invalidate any data in use by other threads (which is what enables save CodeBuffer sharing across threads)
* Instead, such lossy modifications trigger a new "version" of the data in the modifying thread. Old versions of the CodeBuffer persist as read-only data for use by the other threads.
* The other threads can update their version of the CodeBuffer. This will decrease the reference count and eventually trigger deallocation of the old version

With the code in this PR, starting a new CodeBuffer version will wipe its entire contents (similar to today's semantics), but that's an implementation detail. In the future we can carry over old data to the new CodeBuffer using relocations, which would still be in line with the persistence model.

## Measurements

Preliminary testing yielded the following numbers (gathered from Tracy):

|| Mirror's Edge[^1] | God of War[^3] | Yooka-Laylee[^4] | Hollow Knight[^5] |
|-|------------------|----------------|-------------------|-----|
| JIT invocations (million) | 6.9⇨3.1(⇨2.9[^2]) | 0.97⇨0.72(⇨0.56) | 0.80⇨0.72(⇨0.53) | 0.71⇨6.8(⇨4.1)|
| Total JIT time (sec) | 20.9⇨15.5(⇨13.0)| 10.3⇨8.4(⇨5.1) | 6.3⇨5.1(⇨3.9) | 5.0⇨4.4(⇨2.7)|
| JIT time savings (%) | 26% (38%) | 19% (50%) | 19% (38%) | 11% (47%) |

[^1]: Tested first 100s of the game, trying to gain control of the character as quickly as possible
[^2]: Potential for follow-up change to carry over CodeBuffer data on resize
[^3]: Tested first 60s of the game without entering any inputs
[^4]: Tested first 45s of the game, trying to enter one level in my save file as quickly as possible and waiting for the camera to stop moving
[^5]: Tested first 36s of the game, loading my last save file and jumping once

I also measured the Steam startup time (up to the completed first render of the Library view). Since we can't profile JIT time across multiple processes, I had to measure the total startup time. The results are perhaps weakly favorable towards this PR (best measurement 35s⇨33s), but the fluctuation is too strong to say with confidence.

### Events that trigger a new CodeBuffer version

* Running out of CodeBuffer space
* TSO auto-migration (i.e. a one-time event upon starting the second application thread)
* Single-stepping

Notably, switching between threads that compile new code does *not* trigger a new version. However, modifying accesses must be serialized, so a global lock is used when running the JIT backend (see next point).

### Global JIT lock

This approach to sharing code buffers requires all JIT (backend) compilation to be logically single-threaded, so a global `CodeBufferWriteMutex` mutex is used. This incurs some contention during startup of heavily multithreaded applications. In Mirror's Edge, I measured a lock contention of 9% (5%[^2]) of total time spent in JIT (worst case was 14% in smaller time windows). That overhead is *included* in the net improvements above, so this indicates practical room for further improvement.

It seems unlikely that even strongly multithreaded applications would suffer from this: Whereas now threads are merely fighting for a global lock, the app threads would've previously wasted time recompiling the same code over and over again.

In the future, the lock contention can be minimized by compiling to a short-lived thread-local buffer and relocating the resulting code to the actual CodeBuffer. Only the relocation would need to be mutex-protected then.

### Interaction with L3 LookupCache

We must rethink the semantics of the L3 LookupCache: It's not really a cache but rather the source for mapping guest addresses to host code pointers. Subsequently, the PR decouples it from the L1/L2 LookupCaches, renames it to GuestToHostMap, and embeds it into CodeBuffer. This means the same versioning effect applies to all data contained in GuestToHostMap. Overall, this makes sense since without sharing this mapping, we'd use the same CodeBuffer but we wouldn't know where to find blocks already compiled on other threads.

Since this data was assumed to be thread-local previously, care must be taken when invalidating it (e.g. for self-modifying code). I checked existing invalidation code paths and verified they are either idempotent (i.e. they run the same action multiple times without order-dependence or other change in effect) or already unused/broken today. ~~The SMC-full mode is an exception that needs to be fixed.~~ The latter group includes SMC-full mode, which has been non-functional on main for an unknown amount of time.

### Signal handlers

Currently, FEX has dedicated CodeBuffer code to cope with signal handlers that must extend the lifetime of a CodeBuffer. This mechanism is now based on persistence as well, which makes it much more straightforward. Signal handlers will now simply increase the refcount of a CodeBuffer so that it doesn't get deallocated too early.

### Stale CodeBuffer versions

Old CodeBuffer versions only get deallocated when all threads that reference it cooperatively *update* to a newer CodeBuffer. This may never happen in some cases (imagine e.g. a background thread spawned at start that enters a syscall for an event that's never raised until program end). Due to geometric growth, the *combined* worst-case memory overhead of all such CodeBuffers is the size of the active CodeBuffer.

It's unclear if this can be fully resolved. However, we can largely mitigate this by minimizing the number of events that trigger version branching in the first place:
* Use a larger initial size of the first CodeBuffer (possibly using a heuristic based on the program size itself)
* Maybe more?

## TODO
* [x] Clean up
* [x] ~~Fix SMC-full mode~~ (actually broken on main)
* [x] Revisit single-stepping code (`ThreadManager::Step`)
* [x] ~~WoA support?~~ (nothing special to do)
* [x] Verify invalidation logic for thread-local LookupCaches is functionally equivalent
* [x] ~~Deallocate previously compiled code on TSO auto-migration~~ (will just carry around dead code until a CodeBuffer clear for now)

## Future

See #4514 for ideas to mitigate single-threaded compilation and other follow-up optimizations.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -9,6 +9,7 @@ $end_info$
 #pragma once
 
 #include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/fextl/memory.h>
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
 
@@ -32,12 +33,21 @@ namespace CodeSerialize {
 }
 
 namespace CPU {
+  struct CodeBuffer {
+    uint8_t* Ptr;
+    size_t Size;
+
+    CodeBuffer(size_t Size);
+    CodeBuffer(const CodeBuffer&) = delete;
+    CodeBuffer& operator=(const CodeBuffer&) = delete;
+    CodeBuffer(CodeBuffer&& oth) = delete;
+    CodeBuffer& operator=(CodeBuffer&&) = delete;
+
+    ~CodeBuffer();
+  };
+
   class CPUBackend {
   public:
-    struct CodeBuffer {
-      uint8_t* Ptr;
-      size_t Size;
-    };
 
     /**
      * @param InitialCodeSize - Initial size for the code buffers
@@ -154,19 +164,16 @@ namespace CPU {
     CodeBuffer* GetEmptyCodeBuffer();
 
     // This is the current code buffer that we are tracking
-    CodeBuffer* CurrentCodeBuffer {};
+    std::shared_ptr<CodeBuffer> CurrentCodeBuffer;
 
   private:
-    CodeBuffer AllocateNewCodeBuffer(size_t Size);
-    void FreeCodeBuffer(CodeBuffer Buffer);
+    fextl::shared_ptr<CodeBuffer> AllocateNewCodeBuffer(size_t Size);
 
-    void EmplaceNewCodeBuffer(CodeBuffer Buffer) {
-      CurrentCodeBuffer = &CodeBuffers.emplace_back(Buffer);
-    }
+    void EmplaceNewCodeBuffer(fextl::shared_ptr<CodeBuffer> Buffer);
 
     // This is the array of code buffers. Unless signals force us to keep more than
     // buffer, there will be only one entry here
-    fextl::vector<CodeBuffer> CodeBuffers {};
+    fextl::vector<fextl::shared_ptr<CodeBuffer>> CodeBuffers;
   };
 
 } // namespace CPU
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -73,7 +73,7 @@ namespace CPU {
     // Write offset into the latest CodeBuffer
     std::size_t LatestOffset;
 
-    // Protects writes to the latest CodeBuffer
+    // Protects writes to the latest CodeBuffer and changes to LatestOffset
     FEXCore::ForkableUniqueMutex CodeBufferWriteMutex;
 
     virtual void OnCodeBufferAllocated(CodeBuffer&) {};
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -46,6 +46,11 @@ namespace CPU {
     ~CodeBuffer();
   };
 
+  class CodeBufferManager {
+  public:
+    fextl::shared_ptr<CodeBuffer> AllocateNew(size_t Size);
+  };
+
   class CPUBackend {
   public:
 
@@ -53,7 +58,7 @@ namespace CPU {
      * @param InitialCodeSize - Initial size for the code buffers
      * @param MaxCodeSize - Max size for the code buffers
      */
-    CPUBackend(FEXCore::Core::InternalThreadState* ThreadState, size_t InitialCodeSize, size_t MaxCodeSize);
+    CPUBackend(FEXCore::Core::InternalThreadState*, size_t InitialCodeSize, size_t MaxCodeSize);
 
     virtual ~CPUBackend();
 
@@ -163,12 +168,12 @@ namespace CPU {
     [[nodiscard]]
     CodeBuffer* GetEmptyCodeBuffer();
 
-    // This is the current code buffer that we are tracking
-    std::shared_ptr<CodeBuffer> CurrentCodeBuffer;
+    // This is the size of the last code buffer we allocated
+    size_t CurrentCodeBufferSize = 0;
+
+    CodeBufferManager manager; // TODO: Rename
 
   private:
-    fextl::shared_ptr<CodeBuffer> AllocateNewCodeBuffer(size_t Size);
-
     void EmplaceNewCodeBuffer(fextl::shared_ptr<CodeBuffer> Buffer);
 
     // This is the array of code buffers. Unless signals force us to keep more than
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -9,6 +9,7 @@ $end_info$
 #pragma once
 
 #include <FEXCore/Utils/CompilerDefs.h>
+#include <FEXCore/Utils/SignalScopeGuards.h>
 #include <FEXCore/fextl/memory.h>
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
@@ -32,10 +33,14 @@ namespace CodeSerialize {
   struct CodeObjectFileSection;
 }
 
+struct GuestToHostMap;
+
 namespace CPU {
   struct CodeBuffer {
     uint8_t* Ptr;
     size_t Size;
+
+    fextl::unique_ptr<GuestToHostMap> LookupCache;
 
     CodeBuffer(size_t Size);
     CodeBuffer(const CodeBuffer&) = delete;
@@ -46,8 +51,36 @@ namespace CPU {
     ~CodeBuffer();
   };
 
+  /**
+   * A manager that coordinates access to the CodeBuffer used for compiling new code across threads.
+   *
+   * The CodeBuffer is managed as a partially persistent data structure:
+   * - Exactly one CodeBuffer is now designated as "active", which means data can be appended to it
+   * - Lossy modifications to the active CodeBuffer will not invalidate any data in use by other threads (which is what enables save CodeBuffer sharing across threads)
+   * - Instead, such lossy modifications trigger a new "version" of the data in the modifying thread. Old versions of the CodeBuffer persist as read-only data for use by the other threads.
+   * - The other threads can update their version of the CodeBuffer. This will decrease the reference count and eventually trigger deallocation of the old version
+   */
   class CodeBufferManager {
   public:
+    // Get the CodeBuffer that was most recently allocated.
+    // This is the only CodeBuffer that data may be written to.
+    fextl::shared_ptr<CodeBuffer> GetLatest();
+
+    // Allocate a new CodeBuffer with geometric growth.
+    // Subsequent calls to GetLatest will point to the returned buffer.
+    fextl::shared_ptr<CodeBuffer> StartLargerCodeBuffer(size_t MaxCodeSize);
+
+    // Write offset into the latest CodeBuffer
+    std::size_t LatestOffset;
+
+    // Protects writes to the latest CodeBuffer
+    FEXCore::ForkableUniqueMutex CodeBufferWriteMutex;
+
+    virtual void OnCodeBufferAllocated(CodeBuffer&) {};
+
+  private:
+    fextl::shared_ptr<CodeBuffer> Latest;
+
     fextl::shared_ptr<CodeBuffer> AllocateNew(size_t Size);
   };
 
@@ -55,10 +88,9 @@ namespace CPU {
   public:
 
     /**
-     * @param InitialCodeSize - Initial size for the code buffers
      * @param MaxCodeSize - Max size for the code buffers
      */
-    CPUBackend(FEXCore::Core::InternalThreadState*, size_t InitialCodeSize, size_t MaxCodeSize);
+    CPUBackend(CodeBufferManager&, FEXCore::Core::InternalThreadState*, size_t MaxCodeSize);
 
     virtual ~CPUBackend();
 
@@ -157,6 +189,11 @@ namespace CPU {
 
     bool IsAddressInCodeBuffer(uintptr_t Address) const;
 
+    // Updates the CodeBuffer if needed and returns a reference to the old one.
+    // The returned reference should be kept alive carefully to avoid early deletion of resources.
+    [[nodiscard]]
+    fextl::shared_ptr<CodeBuffer> CheckCodeBufferUpdate();
+
   protected:
     // Max spill slot size in bytes. We need at most 32 bytes
     // to be able to handle a 256-bit vector store to a slot.
@@ -164,21 +201,21 @@ namespace CPU {
 
     FEXCore::Core::InternalThreadState* ThreadState;
 
-    size_t InitialCodeSize, MaxCodeSize;
+    size_t MaxCodeSize;
     [[nodiscard]]
     CodeBuffer* GetEmptyCodeBuffer();
 
-    // This is the size of the last code buffer we allocated
-    size_t CurrentCodeBufferSize = 0;
+    // This is the code buffer containing the main code under execution by this thread.
+    // CheckCodeBufferUpdate must be used before compiling new code.
+    fextl::shared_ptr<CodeBuffer> CurrentCodeBuffer;
 
-    CodeBufferManager manager; // TODO: Rename
+    // Old CodeBuffer generations required to be valid until returning from signal handlers
+    fextl::vector<fextl::shared_ptr<CodeBuffer>> SignalHandlerCodeBuffers;
+
+    CodeBufferManager& manager; // TODO: Rename
 
   private:
-    void EmplaceNewCodeBuffer(fextl::shared_ptr<CodeBuffer> Buffer);
-
-    // This is the array of code buffers. Unless signals force us to keep more than
-    // buffer, there will be only one entry here
-    fextl::vector<fextl::shared_ptr<CodeBuffer>> CodeBuffers;
+    void RegisterForSignalHandler(fextl::shared_ptr<CodeBuffer>);
   };
 
 } // namespace CPU
--- a/FEXCore/Source/Interface/Core/CPUBackend.h
+++ b/FEXCore/Source/Interface/Core/CPUBackend.h
@@ -66,9 +66,9 @@ namespace CPU {
     // This is the only CodeBuffer that data may be written to.
     fextl::shared_ptr<CodeBuffer> GetLatest();
 
-    // Allocate a new CodeBuffer with geometric growth.
+    // Allocate a new CodeBuffer with geometric growth up to an internal maximum.
     // Subsequent calls to GetLatest will point to the returned buffer.
-    fextl::shared_ptr<CodeBuffer> StartLargerCodeBuffer(size_t MaxCodeSize);
+    fextl::shared_ptr<CodeBuffer> StartLargerCodeBuffer();
 
     // Write offset into the latest CodeBuffer
     std::size_t LatestOffset;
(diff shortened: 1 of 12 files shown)


FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -16,6 +16,88 @@
...
+  fextl::robin_map<uint64_t, uint64_t> BlockList;
+
+  GuestToHostMap();
+  ~GuestToHostMap();
+
+  // Adds to Guest -> Host code mapping
+  void AddBlockMapping(uint64_t Address, void* HostCode, const LockToken&) {
+    [[maybe_unused]] auto Inserted = BlockList.emplace(Address, (uintptr_t)HostCode).second;
+    // NOTE: If this was inserted twice, we've probably raced against another thread to compile this block. Just ignore this one
+    // TODO: Should reset CodeBuffer cursor in that case...
```
bylaws, line 67: Not sure how ordering will work with this pr and multiple entrypoint merges, but this will need to be done for the latter

neobrain: What sort of ordering is needed? This is still protected by the global lock, so there should be no conflict I think. (Taking the lock is a bit subtle - it needs to happen midway through CompileCode but the acquired `unique_lock` is returned to the caller!)

bylaws:
Have you considered allowing for some holes in the codebuffer during multithreaded compilation? Perhaps I'm missing a reason for it not to work but something like
```
LocalCodeBufferOffset= 0
locked{
if (CurThreadWriterMaxSize) {
   CodeBufferOffset += CurThreadWriterMaxSize;
}
LocalCodeBufferOffset= CodeBufferOffset
CurThreadWriterMaxSize = SSANodes * 12

}

<Compile from LocalCodeBufferOffset>

locked{
if (LocalCodeBufferOffset == CodeBufferOffset)  CurThreadWriterMaxSize=0
}
```

Invalidation I suppose would need a barrier now but doesn't intuitively seem complex.

FEXCore/Source/Interface/Core/LookupCache.h
bylaws: Apologies I meant merge ordering :)

neobrain:
> Have you considered allowing for some holes in the codebuffer during multithreaded compilation?

The size estimates would be pretty rough - not sure how large the gaps we'd leave would be on average (wasting committed memory and possibly impacting host icache efficiency). When implementing the idea of relocation post-compile, this approach sounds better than using temporary storage though.

bylaws:
> The size estimates would be pretty rough - not sure how large the gaps we'd leave would be on average (wasting committed memory and possibly impacting host icache efficiency). When implementing the idea of relocation post-compile, this approach sounds better than using temporary storage though.

Right - I wonder if this would be worth instrumenting just to get some numbers

Sonicadvance1:
I haven't yet looked at this since we had the discussion about block linking and unlinking being a race. What was decided on that?

bylaws:
> I haven't yet looked at this since we had the discussion about block linking and unlinking being a race. What was decided on that?

I had this issue on master, not this pr

Sonicadvance1:
> > I haven't yet looked at this since we had the discussion about block linking and unlinking being a race. What was decided on that?
> 
> I had this issue on master, not this pr

So this PR...doesn't have a race between multiple threads trying to execute the same code and link/unlink at the same time?

> This approach to sharing code buffers requires all JIT (backend) compilation to be logically single-threaded

This also spooks me a bit. I'll need to try some games like Cyberpunk, UE5 games, and RUINER to see if this harms performance on those highly threaded games.

bylaws:
> So this PR...doesn't have a race between multiple threads trying to execute the same code and link/unlink at the same time?
The race with delinking happens even on master, though actually you're right that a linking one would be new - hadn't considered that at all but you're right it probably warrants fixing within this

FEXCore/Source/Interface/Core/LookupCache.h
neobrain: The latest version of this PR also changes the insert-if-new to an entry override (though it only happens in exceptional cases here), so this works out now.

Sonicadvance1:
> > I haven't yet looked at this since we had the discussion about block linking and unlinking being a race. What was decided on that?
> 
> Yeah we all talked past each other that meeting, I think. I hadn't seen any issues (otherwise I would've noted it down here) and ByLaws was referring to an issue on main.
> 
> > This also spooks me a bit. I'll need to try some games like Cyberpunk, UE5 games, and RUINER to see if this harms performance on those highly threaded games.
> 
> Would've been nice to bring this up without letting this PR labeled as RFC sit for 2 weeks, but sure let me know if you find anything different from my measurements on this. It's plausible that single-threaded compilation is just as effective as the previous behavior of parallel-but-redundant compilation.

So this introduces a known race in block linking that can result in a tear in visibility when one thread is patching the code and another thread is executing it.
I would very much want this to be fixed before this is merged. Make sure the matrix of four race conditions (link-direct, link-indirect, unlink-direct, unlink indirect) has its data that is getting modified is aligned to 16-byte granule and do a 128-bit atomic store on it.

Also with some testing shows the global JIT lock does make stutters worse in games that hammer compilation at the same time like RUINER-Linux, and Cyberpunk 2077. RUINER even just lost 10FPS of perf, dropping from 70FPS to 60FPS. So having the threads JIT in a transient buffer and then copy to share would be preferred. A little bit of additional work that gets thrown away if multiple threads are compiling the same code is preferred to stuttering due to job queue systems hitting the same code and stalling.

Still need to throw some more UE5 games at this to see how they behave.

FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -144,31 +177,37 @@ namespace CPU {
 
     bool IsAddressInCodeBuffer(uintptr_t Address) const;
 
+    fextl::shared_ptr<CodeBuffer> GetThreadLocalCodeBuffer() const;
```
bylaws, line 180: With other comments I think this could be dropped entirely (and in ctor it can populate the initial shared member)

FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -144,31 +177,37 @@ namespace CPU {
...
 
-    size_t InitialCodeSize, MaxCodeSize;
+    size_t MaxCodeSize;
     [[nodiscard]]
     CodeBuffer* GetEmptyCodeBuffer();
 
-    // This is the current code buffer that we are tracking
-    CodeBuffer* CurrentCodeBuffer {};
+    // This is the code buffer actively used by this thread
+    fextl::shared_ptr<CodeBuffer> CurrentCodeBuffer;
```
bylaws, line 202: Maybe rename to ActiveCodeBuffer

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -497,18 +498,33 @@ void ContextImpl::LockBeforeFork(FEXCore::Core::InternalThreadState* Thread) {
...
 
-  Thread->LookupCache->ClearCache();
-  Thread->CPUBackend->ClearCache();
+  if (NewCodeBuffer) {
+    // NOTE: Holding on to the reference here is required to ensure validity of the WriteLock mutex
+    std::shared_ptr PrevCodeBuffer = Thread->CPUBackend->GetThreadLocalCodeBuffer();
+    std::lock_guard lk(PrevCodeBuffer->LookupCache->WriteLock);
+
+    // Allocate new CodeBuffer + L3 LookupCache, then clear L1+L2 caches
+    Thread->CPUBackend->ClearCache();
```
bylaws, line 522: Could CPUBackend handle the mapping change itself? It already does on code compilation and it would allow CPUBackend->ClearCache to be called without worrying about updating the mapping exteernally

bylaws (changes requested):

neobrain:
Thanks for testing!

> So this introduces a known race in block linking that can result in a tear in visibility when one thread is patching the code and another thread is executing it. I would very much want this to be fixed before this is merged. Make sure the matrix of four race conditions (link-direct, link-indirect, unlink-direct, unlink indirect) has its data that is getting modified is aligned to 16-byte granule and do a 128-bit atomic store on it.

This has been addressed in #4528 now.

> Also with some testing shows the global JIT lock does make stutters worse in games that hammer compilation at the same time like RUINER-Linux, and Cyberpunk 2077. RUINER even just lost 10FPS of perf, dropping from 70FPS to 60FPS. So having the threads JIT in a transient buffer and then copy to share would be preferred. A little bit of additional work that gets thrown away if multiple threads are compiling the same code is preferred to stuttering due to job queue systems hitting the same code and stalling.

RUINER with its >700 thread spawns per second is a rather pathological case, but making stutters in Cyberpunk noticeably worse would indeed be an undesirable side effect.

I'll look into re-enabling parallel compilation but will work on advancing the code caching work a bit more first. Putting this back to draft mode for now.

FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -144,31 +177,37 @@ namespace CPU {
 
     bool IsAddressInCodeBuffer(uintptr_t Address) const;
 
+    fextl::shared_ptr<CodeBuffer> GetThreadLocalCodeBuffer() const;
+
+    // Returns true if the active CodeBuffer is shared with any other threads
+    bool UsesSharedCodeBuffer() const;
```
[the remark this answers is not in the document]
neobrain: Nothing anymore. Good catch, will drop it in the next update.

neobrain:
I realized we can hold the CodeBufferWriteMutex for a much shorter time, as only the actual code emission and offset update must be protected by it. This reduced lock contention in Yooka Laylee by 66%. Compiling to a temporary buffer can yield an additional 11% reduction (from the baseline). I implemented a prototype for this (see b6af251c4eadeee10611e81f1c03182df11a8d34) but the design decisions are nontrivial so I would rather do this in a follow-up PR.

@Sonicadvance1 Could you retest Cyberpunk with the new changes? If the stuttering persists, can you quantify how much worse it is? Did you find any other reasonable games that were impacted negatively?

Ultimately there is a trade-off between different types of workloads: Stuttering might be reduced in only some games while battery drain should be reduced across the board. Hopefully this latest update shifts the trade-off closer towards the optimum, though, and on-disk code caching should more than make up the few regressions eventually.

bylaws:
Quite interesting how slow ICache flush takes if it meaningfully increases contention more than the actual emission being under the lock. I wonder if on some other occasion after this PR its a good idea to profile that.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -16,6 +16,86 @@
...
+    if (HostCode == BlockList.end()) {
+      return std::nullopt;
+    }
+    return HostCode->second;
+  }
+
+  void Erase(FEXCore::Core::CpuStateFrame* Frame, uint64_t Address, const LockToken&) {
+    // Sever any links to this block
+    auto lower = BlockLinks->lower_bound({Address, nullptr});
+    auto upper = BlockLinks->upper_bound({Address, reinterpret_cast<FEXCore::Context::ExitFunctionLinkData*>(UINTPTR_MAX)});
```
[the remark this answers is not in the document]
neobrain: This code just moved around, but I also don't think `numeric_limits` works for pointers (since this `UINTPTR_MAX` cast is technically UB anyway).

neobrain:
> Compiling to a temporary buffer can yield an additional 11% reduction (from the baseline). I implemented a prototype for this (see https://github.com/FEX-Emu/FEX/commit/b6af251c4eadeee10611e81f1c03182df11a8d34) but the design decisions are nontrivial so I would rather do this in a follow-up PR.

Implementing the same approach but using ThreadPoolAllocator to manage the temporary buffer improved the reduction to around 20% (from the baseline). It also takes care of most of the design decisions, so it is implemented as part of this PR now after all. Every thread will compile to a temporary buffer now, and the global lock is only taken to move the temporary output to the current CodeBuffer. (I still need to hunt down the source for the one unit test failure, but it's all working fine from real-world testing.)

Unfortunately we have to take a hit on space overhead for this when multiple threads compile the same block: Preventing this would require checking the LookupCache while CodeBufferWriteMutex is still locked, however that in turn requires waiting for the LookupCache mutex to become available (which increases overall lock contention again). We should probably look into maintaining some "in-progress" queue instead that threads can check before even attempting to compile in the first place, but this is out of scope for this PR.

During testing this also revealed that our codegen size estimate has been too conservative, which was a latent source of bugs/crashes.

neobrain:
Now that the JIT compiles to a temporary buffer, its code output must not depend on the host address. The only such dependency was an optimization in Arm64Emitter's `LoadConstant`. This optimization is almost exclusively applied when translating `THUNK_OP`, so I've disabled it for now. Once we fully integrate and make use of code relocation patching, this optimization can be re-enabled.

FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -33,19 +35,50 @@ namespace CodeSerialize {
...
+  public:
+    fextl::shared_ptr<CodeBuffer> AllocateNewCodeBuffer(size_t Size);
+
+    size_t GetCurrentCodeBufferSize() {
+      return GetCurrentCodeBuffer()->Size;
+    }
+
+    fextl::shared_ptr<CodeBuffer> GetCurrentCodeBuffer();
+
+    fextl::shared_ptr<CodeBuffer> Latest;
```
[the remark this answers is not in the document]
neobrain: Actually going the opposite way here, I think. Since Core.cpp doesn't use this interface anymore, we can just make all state public and access variables directly. `CodeBuffers.Latest` isn't quite the mouthful that `CodeBuffers.GetCurrentCodeBuffer()` was either.

Will drop GetCurrentCodeBufferSize however.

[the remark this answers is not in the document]
neobrain: Err, nevermind, I wasn't thinking. GetCurrentCodeBuffer isn't trivial, so we can't just drop it. Will follow your suggestion after all then.

FEXCore/Source/Interface/Core/CPUBackend.h
neobrain: I think CurrentCodeBuffer is more fitting. `Active` has a ring of "being dynamic" to it, which would be misleading here: The current CodeBuffer may be read-only and hence will need to be updated when new code is compiled.

This is a weak opinion though, if you're convinced "active" is better I can change it.

neobrain: I'll update the comment though: `// This is the code buffer containing the main code under execution by this thread`

Art-Chen:
testing with latest head of this mr on Wine ARM64EC, i meet hung when :
- cpu-z benchmark
- exiting MiSide(which is an Windows Unity Game)
- Starting Windows Steam and hung after login screen

fps drop on GTA 5

even if rebase this mr to the main branch, still the same.

maybe need some check on Windows ARM64EC?

Art-Chen:
> though cpu-z is crashing for me with upstream Wine (git) and upstream FEX (2505) already

CPU-Z crash is known issue and need an workaround patch for wine  to get it to run.

You can just try running MiSide to test it. 
https://store.steampowered.com/app/2527500/_MiSide/

bylaws:
> > > testing with latest head of this mr on Wine ARM64EC, i meet hung when :
> > > cpu-z benchmark
> > > exiting MiSide(which is an Windows Unity Game)
> > > Starting Windows Steam and hung after login screen
> 
> > How often does this happen? Does it happen every time or only sometimes?
> 
> Every time.
> 
> > How much did fps drop?
> 
> about 15-20fps. (testing with wine client sync patches, ie. esync / fsync and so on.)
> 
> > Are you using upstream Wine or ByLaws's fork?
> 
> Upstream Wine and picked some patch by ByLaws (AVX / SVE / some hack)

There's a known bug here Ifrom a couple weeks back 'm not sure I pushed a fix toy upstream-arm64ec branch for. Might also be missing suspend patches. Could you use the arm64ec-10 branch please?

Art-Chen:
> Might also be missing suspend patches

I tried to pick up the suspend related patches from your branch, but it will cause cef based application hang when using client sync(esync/fsync). so i reverted it.

> Could you use the arm64ec-10 branch please?

i'll try later. and could you telling more detail about this issue? and which patch addressed this issue? Thanks!

Sonicadvance1:
Needs a rebase but I'm now looking at this.

Sonicadvance1:
CyberPunk 2077 is worse off with this PR because of its consistent work threads code invalidation and jitting causing significant per-frame contention, but I don't think it is a blocker.

UE5 games so far don't seem worse off, might be slightly better off even since their worker threads would end up randomly jitting new code that a different thread would have executed prior.

Sonicadvance1:
Recent changes with this seems to improve RUINER on native Linux. Around 65FPS with this PR versus around 30 on main due to all the VLC jitting overhead.

Sonicadvance1 (approved):

neobrain:
Forgot replying to this before, but thanks for re-testing! I'll note down Cyberpunk for future benchmarking since there's some more low-hanging fruit for optimization here.

PR #4481, opened 2025-04-04
InstcountCI: Ensure RIP of blocks is consistent 

Sonicadvance1:
This changes the instcountCI code to consistently load test data in to
RIP 0x1'0000 so we don't have any spurious changes due to json changes.

This has been a minor annoyance where if a test was added, it had the
potential to shift the rest of the data in the tests. This now ensures
it is consistent.


Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -39,7 +74,7 @@ class CodeSizeValidation final {
   void CalculateBaseStats(FEXCore::Context::Context* CTX, FEXCore::Core::InternalThreadState* Thread);
 private:
   void ClearStats() {
-    RIPToStats.clear();
+    memset(&CurrentStats, 0, sizeof(CurrentStats));
```
lioncash, line 77: Seems like this could just be `CurrentStats = {};`

lioncash (commented):
One minor nit

Source/Tools/CodeSizeValidation/Main.cpp
Sonicadvance1: Oop, yep, That was before I changed it away from a std::pair so it was complaining about a default constructor not existing. Changed.

lioncash (approved):

PR #4485, opened 2025-04-07
cephes: Rewrite to use softfloat-3e 128-bit 

Sonicadvance1:
And also use it at the same time, since the function signatures changed.

Instead of relying on the host libc math libraries for `long double` ALU
operations, rewrite the entire thing to use softfloat-3e fixed width
float128_t types.

This is a very invasive change in cephes but is a necessary requirement
for getting the precision we require in environments that map `long
  double` to be the same as `double`, like Win32 and MacOS.

This fixes the precision issue in transcendental operations when running
under WINE.


lioncash (approved):
Seems ok, with a very load-bearing "seems" since this is quite the reworking of an external

Sonicadvance1:
Ran through a bunch of games with x87reduced precision disabled to double check that they aren't broken.
Nothing broken that wasn't already, like MGRR crashing on the title screen and Mirror's Edge having flickering graphics with reduced precision.

Then finally running Bioshock Infinite hard locked up my board from the Radeon GPU crashing.

So nothing unexpected from these changes really.

PR #4487, opened 2025-04-08
Run gcc target tests with block size 1

### LINKED ISSUE #4436: gcc test fails on HDK8650
Commit 2f6f8b93e (PR #4424) breaks `pr88240.c.gcc-target-test-64.jit.gcc-target-64` on the HDK. Passes on the Lenovo T14s, which makes me think this is potentially SVE or RPRES related.

Commit d48413cbe is the last passing commit.

pmatos:
Fixes #4436


pmatos:
Interesting that `pr88240.c` which is what #4436 is about should have been already fixed. I am surprised this is failing in CI.

pmatos:
Stabilizing this PR will take a week as I am away over Easter, so marking it as draft.

pmatos:
@Sonicadvance1 ping - any comments here?

Sonicadvance1 (approved):
Oh, I missed that this was changed.

PR #4488, opened 2025-04-08
Add option to enable UBSAN

pmatos:


CMakeLists.txt
```diff
@@ -13,6 +13,7 @@ option(ENABLE_IWYU "Enables include what you use program" FALSE)
 option(ENABLE_LTO "Enable LTO with compilation" TRUE)
 option(ENABLE_XRAY "Enable building with LLVM X-Ray" FALSE)
 set(USE_LINKER "" CACHE STRING "Allow overriding the linker path directly")
+option(ENABLE_ASAN "Enables Clang UBSAN" FALSE)
```
lioncash, line 16:
```suggestion
option(ENABLE_UBSAN "Enables Clang UBSAN" FALSE)
```

pmatos: Thanks for catching this. It's so strange that even with this typo - `-DENABLE_UBSAN` works. :-/

Sonicadvance1: cmake is wacky like that. These option lines aren't strictly necessary and only really needed if you're wanting to use nice tools like `ccmake`.

pmatos:
To build this with `-DENABLE_VIXL_DISASSEMBLER=True` needs https://github.com/FEX-Emu/vixl/pull/5

Sonicadvance1 (approved):

PR #4489, opened 2025-04-09
FEXCore: Reduce stack usage in CalculateNumberOfCPUs

Sonicadvance1:
Nothing crazy, just recalculate the maximum string length rather than
use PATH_MAX.

### DIFF
--- a/Source/Common/CPUInfo.cpp
+++ b/Source/Common/CPUInfo.cpp
@@ -12,7 +12,7 @@
 #include <linux/limits.h>
 #endif
 
-namespace FEXCore::CPUInfo {
+namespace FEX::CPUInfo {
 #ifndef _WIN32
 uint32_t CalculateNumberOfCPUs() {
   constexpr auto parse_string = FMT_COMPILE("/sys/devices/system/cpu/cpu{}");
@@ -36,4 +36,4 @@ uint32_t CalculateNumberOfCPUs() {
   return std::thread::hardware_concurrency();
 }
 #endif
-} // namespace FEXCore::CPUInfo
+} // namespace FEX::CPUInfo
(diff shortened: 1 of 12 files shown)


Source/Common/CPUInfo.cpp
```diff
@@ -11,14 +12,16 @@
 #include <linux/limits.h>
 #endif
 
-namespace FEXCore::CPUInfo {
+namespace FEX::CPUInfo {
 #ifndef _WIN32
 uint32_t CalculateNumberOfCPUs() {
-  char Tmp[PATH_MAX];
+  constexpr auto parse_string = FMT_COMPILE("/sys/devices/system/cpu/cpu{}");
+  constexpr auto max_parse_size = ::fmt::formatted_size(parse_string, std::numeric_limits<uint32_t>::max());
```
lioncash, line 19: Since it's a non-generic context, this could use UINT32_MAX just to be a tad less verbose

```suggestion
  constexpr auto max_parse_size = ::fmt::formatted_size(parse_string, UINT32_MAX);
```

lioncash (approved):

Source/Tools/FEXLoader/AOT/AOTGenerator.cpp
```diff
@@ -103,7 +104,7 @@ void AOTGenSection(FEXCore::Context::Context* CTX, ELFCodeLoader::LoadedSection&
 
   // This code is tricky to refactor so it doesn't allocate memory through glibc.
   FEXCore::Allocator::YesIKnowImNotSupposedToUseTheGlibcAllocator glibc;
-  for (int i = 0; i < FEXCore::CPUInfo::CalculateNumberOfCPUs(); i++) {
+  for (int i = 0; i < FEX::CPUInfo::CalculateNumberOfCPUs(); i++) {
```
lioncash, line 107: Should have caught this in the previous review, but perhaps the retrieval of the upper limit should be stored into a variable and then used in the limit, so that it's not doing all of the string processing over and over for every iteration

Sonicadvance1: Good shout, moved it to a temp variable.

lioncash (approved):

PR #4491, opened 2025-04-09
FEXCore/Emitter: Stop creating a vector on the heap

Sonicadvance1:
In the Push/Pop CalleeSavedRegisters these vectors were getting created on the heap, allocating memory and then just iterating them.

Just use a std::array which makes it stop allocating memory and saves the number of instructions.

Before:
Image_2025-04-08_18-38-56

After:
Image_2025-04-08_18-39-02


lioncash (approved):
Nice

PR #4492, opened 2025-04-09
Frontend: Be more tolerant of bad instruction encodings

bylaws:


FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -615,7 +627,6 @@ bool Decoder::NormalOpHeader(const FEXCore::X86Tables::X86InstInfo* Info, uint16
         255, 0, 1, 2, 255, 255, 255, 3,
       };
       uint8_t Field = RegToField[ModRM.reg];
-      LOGMAN_THROW_A_FMT(Field != 255, "Invalid field selected!");
```
bylaws, line 618: (will fix this)

Sonicadvance1 (approved):

PR #4497, opened 2025-04-10
Cleanup info test files for 32bits

pmatos:
No point in duplicating test info between Disabled and Known Failures.

Leaving tests known to fail in Known_Failures. Flakes / Unreliable tests go into Disabled_Tests.

### DIFF
--- a/unittests/gcc-target-tests-32/Disabled_Tests
+++ b/unittests/gcc-target-tests-32/Disabled_Tests
@@ -4,18 +4,6 @@ pr57275.c.gcc-target-test-64
 # Consistently crashes on Solidrun only in CI
 20080723-1.c.gcc-target-test-32
 
-# These tests fail because of things unrelated to the sse4.1 instructions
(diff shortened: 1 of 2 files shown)


pmatos:
@Sonicadvance1 the `sse4_1-...` tests seem to pass in all the configs I tried. OK to remove from the skip files?

Sonicadvance1:
> @Sonicadvance1 the `sse4_1-...` tests seem to pass in all the configs I tried. OK to remove from the skip files?

I think we even removed these skips from the 64-bit side a month or two ago because we fixed the rounding bugs, just missed the 32-bit side.

pmatos:
> > @Sonicadvance1 the `sse4_1-...` tests seem to pass in all the configs I tried. OK to remove from the skip files?
> 
> I think we even removed these skips from the 64-bit side a month or two ago because we fixed the rounding bugs, just missed the 32-bit side.

Good - it's done then on the 32bit side as well now.

Sonicadvance1 (approved):

PR #4499, opened 2025-04-11
OpcodeDispatcher: Mark NZCV as dirty when clobbering in FCOMIF64

bylaws:


bylaws:
I think making NZCVToX87 use NZCV ops directly makes the most sense here to fix the problems? Round tripping through eflags to restore it (unset dirty then reset after) feels unintuitive

alyssarosenzweig:
Sorry, can you explain what's happening here? Is this supposed to be a bugfix or an optimization or what?

bylaws:
> Sorry, can you explain what's happening here? Is this supposed to be a bugfix or an optimization or what?

It's intended to be a bugfix, GetNZCV doesn't mark NZCV as dirty and only saves it, so if you had the instruction afterwards call GetRFlag then it would take https://github.com/FEX-Emu/FEX/blob/870f03b810d0108b0f61f343779deef4b108a247/FEXCore/Source/Interface/Core/OpcodeDispatcher.h#L2004 this path and read the clobbered value when FCOMIS should only modify x87 flags and leave eflags untouched.

alyssarosenzweig:
Rrright OK, yeah, I think SaveNZCV is indeed the right thing here although I only faintly remember the details here.

But CI seems to disagree pretty hard :cactus:

bylaws:
Yeah - NZCVToX87 is using getrflags rn which is using the saved NZCV rather than the just written one, should be easy to fix

bylaws:
Think this should be good now

alyssarosenzweig:
Awkward. Thanks!

alyssarosenzweig (approved):

Sonicadvance1 (approved):

Issue #4131, opened 2024-10-24
Bear's Restaurant has race condition with deferred signal handling


Sonicadvance1:
[Bear's Restaurant](https://store.steampowered.com/app/1687550/Bears_Restaurant/) has a race condition with FEX's deferred signal handling. Its [game engine](https://ebitengine.org/) is based on golang, which sends SIGURG /very/ quickly for doing preemption to ensure thread fairness.

A couple of things to fix in FEX's deferred signal handling.
- If its not a realtime signal, then these shouldn't be queued and instead just set in a mask of available signals.
  - Linux doesn't queue non-RT signals, expecting many non-RT signals to be lost rather than queued
- FEX should block the `sa_mask` in the signal handler of the signal getting deferred, to ensure it doesn't receive anymore of that signal.
  - This would ensure FEX doesn't overrun its small signal queue, which can typically result in crashes.
  - Make sure to restore signal mask once actually handling the signal

I don't have time to implement these fixes right now but I needed a tracking issue to remember it.

I originally wanted to test this engine to see it break under Asahi Linux, but apparently it uses a new enough version of golang that it no longer hits the VDSO bug, might need to try an older game using this engine to see the Asahi breakage.

Sonicadvance1:
Sadly it looks like #4501 didn't fix this, so there's more problems to fix.
For now I have added some workarounds to the [wiki](https://wiki.fex-emu.com/index.php/Bear%27s_Restaurant) until it gets fixed.

PR #4501, opened 2025-04-12
LinuxSyscalls: Update signal mask at deferring time

JunChi1022:
In regular x86 programs, when a signal occurs, the signal will not be handled within the signal handler. However, under FEX's defer signal mechanism, the signal is not immediately masked when it is deferred. When returning to the location that receives the signal and continues processing, the signal might be received again, causing inconsistency between the emulation and the actual program.

Here is an unit test for this patch from ltp:
https://github.com/linux-test-project/ltp/blob/master/testcases/kernel/syscalls/timer_settime/timer_settime03.c

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
@@ -18,6 +18,7 @@ $end_info$
 #include <FEXCore/Utils/Profiler.h>
 #include <FEXCore/Utils/SignalScopeGuards.h>
 
+#include <bits/types/sigset_t.h>
 #include <cstdint>
 #include <linux/seccomp.h>
 
@@ -37,6 +38,7 @@ struct ThreadStateObject : public FEXCore::Allocator::FEXAllocOperators {
   struct DeferredSignalState {
     siginfo_t Info;
     int Signal;
+    uint64_t SigMask;
   };
 
   FEXCore::Core::InternalThreadState* Thread;
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp
@@ -536,6 +536,27 @@ static bool IsAsyncSignal(const siginfo_t* Info, int Signal) {
   return true;
 }
 
+uint64_t SignalDelegator::GetNewSigMask(int Signal) const {
+  const SignalHandler& Handler = HostHandlers[Signal];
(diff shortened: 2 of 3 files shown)


FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
```diff
@@ -127,7 +127,11 @@ class DeferredSignalRefCountGuard final {
 #else
       Thread->CurrentFrame->State.DeferredSignalRefCount.Decrement(1);
       auto InterruptFaultPage = reinterpret_cast<Core::NonAtomicRefCounter<uint64_t>*>(&Thread->InterruptFaultPage);
-      InterruptFaultPage->Store(0);
+      // In nested signal-deferring sections, trigger deffered signals at the
+      // outermost section to avoid unnecessary segmentation handling
+      if (Thread->CurrentFrame->State.DeferredSignalRefCount.Load() == 0) {
```
Sonicadvance1, line 132: There's no reason to change this. We are explicitly eating the cost of some additional SIGSEGV on the interruptFaultPage because our deferred signal sections don't nest deeply.

Additionally our JIT also doesn't emit a branch check on this and still faults multiple times regardless.

Also this makes this half of the code behave the same as the x86 side, except it's doing two atomic operations instead of one, so if we wanted to go this path we would just use the x86 implementation on arm64.

JunChi1022: I have remove these code, please help to review the other code, thank you

Sonicadvance1: I take a closer look at this in a bit, no worries.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -37,6 +38,7 @@ struct ThreadStateObject : public FEXCore::Allocator::FEXAllocOperators {
   struct DeferredSignalState {
     siginfo_t Info;
     int Signal;
+    sigset_t SigMask;
```
Sonicadvance1, line 41: Just make this a uint64_t and memcpy for copying. No need to burn 128-bytes per deferred signal state just because glibc wanted to future proof having support for 1024 signals or whatever.

JunChi1022: Done.

Sonicadvance1 (approved):
Nice! Thanks for the contribution!
I'll check later today if this fixes #4131 as well, I had a similar solution as a WIP but I never finished it.

PR #4508, opened 2025-04-15
Emitter: Add missing ASIMD crypto operations

lioncash:
Like with the SVE side of things, we can just drop these into place, now that vixl supports them


Sonicadvance1 (approved):
Nice.

PR #4522, opened 2025-04-19
LinuxEmulation: Copy the host persona on thread startup

bylaws:


Sonicadvance1:
This is probably fine, but also should probably research which personalities conflict between arm64 and x86/x86-64.

bylaws:
> This is probably fine, but also should probably research which personalities conflict between arm64 and x86/x86-64.

A quick look shows no personalities supported on arm that aren't supported on x86, and these two are supported on x86 but not arm:
[ADDR_LIMIT_3GB](https://elixir.bootlin.com/linux/v6.14.2/C/ident/ADDR_LIMIT_3GB))
[ADDR_COMPAT_LAYOUT](https://elixir.bootlin.com/linux/v6.14.2/C/ident/ADDR_COMPAT_LAYOUT)

Sonicadvance1 (approved):

PR #4523, opened 2025-04-19
Improve tracking of executable mappings

bylaws:
Prerequisite for NX emulation, split apart since there's nothing tying them to being merged together.

### DIFF
--- a/Source/Tools/CommonTools/HarnessHelpers.h
+++ b/Source/Tools/CommonTools/HarnessHelpers.h
@@ -419,14 +419,8 @@ public:
     return RIP;
   }
 
-  bool MapMemory() {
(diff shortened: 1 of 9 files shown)


Source/Tools/CommonTools/HarnessHelpers.h
```diff
@@ -419,14 +419,8 @@ class HarnessCodeLoader final : public FEX::CodeLoader {
     return RIP;
   }
 
-  bool MapMemory() {
+  bool MapMemory(std::function<void*(uint64_t, size_t)> DoMMap) {
```
lioncash, line 422: This should probably taken by const reference, since it's only ever called, so the copy isn't really necessary

Sonicadvance1 (approved):

PR #4531, opened 2025-04-23
InstcountCI: Changes how code size is calculated 

Sonicadvance1:
Due to how jit block tail padding is working, there's no real good way
to determine the true "implementation size" of an instruction without
the backend being aware of wanting to investigate it.

Trying to inject another instruction, or another IR operation actually
subtly changes codegen in a way that gives invalid results. The only
real way to get around this is to inject a known token in to the
instruction stream as we `ExitFunction`.

So inject a `udf #0x420f`, and change the scanning behaviour to find the
first one and cut everything else off afterwards.

This already scoops out some code in some game blocks that were
accidentally landing ExitFunction code in the json.

This also has been tested to work with https://github.com/FEX-Emu/FEX/pull/4528 with its InstCountCI
specific changes reverted.

This means we don't need to play subtle padding tricks in the JIT to get
the information we want in InstcountCI.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
@@ -49,6 +49,11 @@ DEF_OP(ExitFunction) {
 
   ResetStack();
 
+  if (CTX->HostFeatures.IsInstCountCI) [[unlikely]] {
+    // Emit function end marker
+    udf(0x420F);
+  }
+
   uint64_t NewRIP;
 
   if (IsInlineConstant(Op->NewRIP, &NewRIP) || IsInlineEntrypointOffset(Op->NewRIP, &NewRIP)) {
--- a/FEXCore/include/FEXCore/Core/HostFeatures.h
+++ b/FEXCore/include/FEXCore/Core/HostFeatures.h
@@ -43,6 +43,9 @@ struct HostFeatures {
   bool SupportsAFP {};
   bool SupportsFloatExceptions {};
 
+  // Flag if this is InstCountCI
+  bool IsInstCountCI {};
+
   // MIDR information
   // Also used for determining number of CPU cores for CPUID
   fextl::vector<uint32_t> CPUMIDRs;
--- a/FEXCore/unittests/Emitter/System_Tests.cpp
+++ b/FEXCore/unittests/Emitter/System_Tests.cpp
@@ -6,6 +6,10 @@
 
 using namespace ARMEmitter;
 
+TEST_CASE_METHOD(TestDisassembler, "Emitter: Reserved") {
+  TEST_SINGLE(udf(0), "udf #0x0");
+  TEST_SINGLE(udf(0xFFFF), "udf #0xffff");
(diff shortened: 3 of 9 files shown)


Source/Tools/CodeSizeValidation/Main.cpp
```diff
@@ -144,12 +131,18 @@ bool CodeSizeValidation::ParseMessage(const char* Message) {
...
+      CurrentStats.second.erase(CurrentStats.second.begin(), CurrentStats.second.begin() + HeaderSize);
     }
-    if (BaseStats.TailSize) {
-      CurrentStats.second.erase(CurrentStats.second.end() - BaseStats.TailSize, CurrentStats.second.end());
+    // Find the first `udf #0x420f` and remove everything from that point onward.
+    for (auto it = CurrentStats.second.begin(); it != CurrentStats.second.end(); ++it) {
+      if (*it == "udf #0x420f") {
+        CurrentStats.second.erase(it, CurrentStats.second.end());
+        break;
+      }
```
neobrain, line 142:
```suggestion
    auto EraseBegin = std::find(CurrentStats.second.begin(), CurrentStats.second.end(), "udf #420f");
    CurrentStats.second.erase(EraseBegin, CurrentStats.second.end());
```

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -49,6 +49,10 @@ DEF_OP(ExitFunction) {
 
   ResetStack();
 
+  if (CTX->HostFeatures.IsInstcountCI) [[unlikely]] {
+    udf(0x420F);
```
neobrain, line 53:
```suggestion
    // Emit function end marker
    udf(0x420F);
```

FEXCore/include/FEXCore/Core/HostFeatures.h
```diff
@@ -43,6 +43,9 @@ struct HostFeatures {
   bool SupportsAFP {};
   bool SupportsFloatExceptions {};
 
+  // Flag if this is InstCountCI
+  bool IsInstcountCI {};
```
neobrain, line 47: Nit:
```suggestion
  bool IsInstCountCI {};
```

FEXCore/unittests/Emitter/System_Tests.cpp
```diff
@@ -6,6 +6,9 @@
 
 using namespace ARMEmitter;
 
+TEST_CASE_METHOD(TestDisassembler, "Emitter: Reserved") {
+  TEST_SINGLE(udf(0), "udf #0x0");
```
neobrain, line 10: Does it make sense to add a test case for the maximum reserved value as well?

neobrain (approved):
I'll take this any day over the workaround in my PR - thanks!

FEXCore/unittests/Emitter/System_Tests.cpp
Sonicadvance1: Sure, added one.

PR #4532, opened 2025-04-24
X87 state block reset

pmatos:


FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -112,7 +112,15 @@ class OpDispatchBuilder final : public IREmitter {
     // Changes get stored out by CalculateDeferredFlags.
     CachedNZCV = nullptr;
     CFInverted = CFInvertedABI;
-    FlushRegisterCache();
```
bylaws, line 115: I think keeping this then setting MMXState directly without using ChgStateMMX_X87 might be cleaner (can also drop the force param)

bylaws: It seems good to me to differentiate just a cache reset to an actual state transition here, and as a counterpoint, if some change to x87 transition stopped it from flushing the register cache then everything would be broken here.

pmatos:
The instcountci adds two instructions:

```
+        "strb wzr, [x28, #1019]",
         "mov w20, #0xffff",
+        "strb w20, [x28, #1298]",
```

These are from fixing the default state at the start of a block to X87. An MMX instruction will set the top to zero, and then set the top bits of the MMX reg being set to 1s. Looks ok.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
pmatos: Yeah, I think it goes both ways. Not totally convinced about this tbh but I don't think it's that relevant. I can change it if you prefer.

bylaws (approved):

Sonicadvance1 (approved):

PR #4537, opened 2025-04-24
updated Dockerfile to reflect latest FEX-Emu releases

sdpoueme:
Add multi-stage Dockerfile for FEX emulator build

- Stage 1 (Builder): Sets up build environment with Ubuntu 22.04
  - Installs development tools and dependencies
  - Builds FEX using clang-13 with optimized settings
  - Configures cmake with LTO enabled and tests disabled

- Stage 2 (Runner): Creates minimal runtime image
  - Includes only necessary runtime libraries
  - Copies built binaries from builder stage

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,31 +1,31 @@
 # --- Stage 1: Builder ---
-FROM ubuntu:20.04 as builder
+FROM ubuntu:22.04 as builder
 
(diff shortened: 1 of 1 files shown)


sdpoueme:
> Looks reasonable enough to me.

Thank you. We have more contributions around this and will submit them in the upcoming weeks.

Issue #4535, opened 2025-04-24
Work towards reducing codegen for Interpreter fallbacks


Sonicadvance1:
Remove the ABI spilling and everything from the JIT blocks themselves and move it to the Dispatcher. Will alleviate some of the problems that #4486 is hitting but not expected to remove it.

A bit counter-intuitively, generating a bunch of ABI handling code inline in the JIT is actually /worse/ for performance than one might expect. It's actually better to generate all this code in the dispatcher, and then eat an additional branch to jump to it instead. This is something that I found out years ago in the Dolphin JIT.
In a multithreaded environment (without code sharing) it also means that we only have this ABI handling consuming cachelines from one memory region, making it more likely to already be in L2 for cores.

I already started working towards this a couple months ago when I was keeping FPRs in vector registers for ABI calls, which reduces the burden of putting things in GPRs (especially because of win32 ABI).

Basic gist is to move all the ABI handling for the 16 different fallback ABIs to the dispatcher, and then the JIT only needs to do some /very/ minor work instead. This'll cut the number of instructions that the JIT needs to emit from 34-60ish to something like less than a dozen? Haven't implemented it yet so I don't have hard numbers yet.

alyssarosenzweig:
> It's actually better to generate all this code in the dispatcher, and then eat an additional branch to jump to it instead.

i-cache?

PR #4538, opened 2025-04-24
JIT: Move interpreter ABI handlers in to the dispatcher 

Sonicadvance1:
Fixes https://github.com/FEX-Emu/FEX/issues/4535

A handful of improvements on this.
* Reduces codegen around interpreter fallbacks
* Keeps ABI handling code in common Dispatcher code
* Improves I$ hitrate by most of the heavy code staying in Dispatcher

This cuts the amount of codegen inside the JIT for most interpreter
fallbacks by 1/2 or 1/3, by only doing the minimal amount of work in the
code blocks and doing most things in the dispatcher. The cost of which
is an additional branch per operation.

This should bring marginal performance improvements, but it should also
basically fall within noise. The bigger thing to care about here is a
smaller amount of code being generated for x87 blocks.


FEXCore/Source/Interface/Core/Interpreter/Fallbacks/InterpreterFallbacks.cpp
```diff
@@ -88,11 +124,11 @@ bool InterpreterOps::GetFallbackHandler(bool SupportsPreserveAllABI, const IR::I
 
     switch (Op->SrcSize) {
     case IR::OpSize::i32Bit: {
-      *Info = {FABI_F80_I16_F32_PTR, (void*)&FEXCore::CPU::OpHandlers<IR::OP_F80CVTTO>::handle4, Core::OPINDEX_F80CVTTO_4, SupportsPreserveAllABI};
+      *Info = {FABI_F80_I16_F32_PTR, Core::OPINDEX_F80CVTTO_4, SupportsPreserveAllABI};
```
bylaws, line 127: A shared Core::OPINDEX to FABI map would be nice since this is effectively duplicated above for indexing ABIHandlers

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -590,6 +617,427 @@ void Dispatcher::ExecuteJITCallback(FEXCore::Core::CpuStateFrame* Frame, uint64_
...
+  case FABI_F80_I16_I32_PTR: {
+    // Linux Reg/Win32 Reg:
+    // tmp4 (x4/x13): FallbackHandler
+    // x30: return
+    // tmp2 (x1/x11): source
+    SpillForABICall(CTX->HostFeatures.SupportsPreserveAllABI, TMP3, true);
+
+    if (TMP2 != ABI2) {
+      if (ABI == FABI_F80_I16_I16_PTR) {
+        sxth(ARMEmitter::Size::i32Bit, ARMEmitter::Reg::r1, TMP2);
```
bylaws, line 723: The JIT.cpp side already sign extends when writing into TMP2 so I don't think this is necessary

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -590,6 +617,427 @@ void Dispatcher::ExecuteJITCallback(FEXCore::Core::CpuStateFrame* Frame, uint64_
...
+    // tmp4 (x4/x13): FallbackHandler
+    // x30: return
+    // vtmp1 (v0/v16): vector source 1
+    // vtmp2 (v1/v17): vector source 2
+    SpillForABICall(CTX->HostFeatures.SupportsPreserveAllABI, TMP3, true);
+
+    if (VTMP1 != VABI1) {
+      fmov(VABI1.D(), VTMP1.D());
+    }
+    if (VTMP2 != VABI2) {
```
bylaws, line 812: Any reason for not using TMP_ABIARGS for all of these?

bylaws (changes requested):

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: You're right, frontend now doing the sign/zero-extend and the dispatcher only doing a move like the other ops.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: Sure, switched it over.

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/InterpreterFallbacks.cpp
Sonicadvance1: That would require quite a bit more refactoring that I'll think about doing after this PR at some point.

bylaws (approved):

PR #4540, opened 2025-04-25
mprotect last page of CodeBuffer

pmatos:
protects last page of codebuffer. This should cause a SIGSEGV if we try to access it. Until now it was possible to go over and access out of bounds.

In addition, there a couple of clang-tidy fixes which should be NFC.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUBackend.cpp
+++ b/FEXCore/Source/Interface/Core/CPUBackend.cpp
@@ -4,6 +4,7 @@
 #include "Interface/Context/Context.h"
 #include "Interface/Core/CPUBackend.h"
 #include "Interface/Core/Dispatcher/Dispatcher.h"
+#include <cstdint>
 
 #ifndef _WIN32
 #include <sys/prctl.h>
@@ -325,7 +326,7 @@ namespace CPU {
           CodeBuffers.resize(1);
         }
         // Set the current code buffer to the initial
-        CurrentCodeBuffer = &CodeBuffers[0];
+        CurrentCodeBuffer = CodeBuffers.data();
 
         if (CurrentCodeBuffer->Size != MaxCodeSize) {
           FreeCodeBuffer(*CurrentCodeBuffer);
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Interface/Core/CPUBackend.cpp
```diff
@@ -382,6 +382,20 @@ namespace CPU {
...
     }
+
+    // Protect the last page of the allocated buffer to trigger SIGSEGV on write access
+#ifndef _WIN32
+    const long pageSize = sysconf(_SC_PAGESIZE);
+    if (pageSize > 0 && Buffer.Size >= (size_t)pageSize) {
+      uintptr_t lastPageAddr = reinterpret_cast<uintptr_t>(Buffer.Ptr) + Buffer.Size - pageSize;
+      lastPageAddr = lastPageAddr & ~(pageSize - 1);
+
+      if (mprotect(reinterpret_cast<void*>(lastPageAddr), pageSize, PROT_NONE) != 0) {
```
bylaws, line 393: Use VirtualProtect here which is generic - and you'll probably want to adjust the CodeSize check to trigger a resize a page earlier

FEXCore/Source/Interface/Core/CPUBackend.cpp
```diff
@@ -382,6 +382,20 @@ namespace CPU {
     if (static_cast<Context::ContextImpl*>(ThreadState->CTX)->Config.GlobalJITNaming()) {
       static_cast<Context::ContextImpl*>(ThreadState->CTX)->Symbols.RegisterJITSpace(Buffer.Ptr, Buffer.Size);
     }
+
+    // Protect the last page of the allocated buffer to trigger SIGSEGV on write access
+#ifndef _WIN32
+    const long pageSize = sysconf(_SC_PAGESIZE);
+    if (pageSize > 0 && Buffer.Size >= (size_t)pageSize) {
+      uintptr_t lastPageAddr = reinterpret_cast<uintptr_t>(Buffer.Ptr) + Buffer.Size - pageSize;
+      lastPageAddr = lastPageAddr & ~(pageSize - 1);
```
bylaws, line 391: AlignDown(reinterpret_cast<uintptr_t>(Buffer.Ptr) + Buffer.Size - 1, FEXCore::Utils::FEX_PAGE_SIZE)
Might be cleaner

Sonicadvance1 (approved):

bylaws (changes requested):
https://github.com/FEX-Emu/FEX/blob/39e2a8354f1252b0a8dad624e240866c6fa2ced2/FEXCore/Source/Interface/Core/JIT/JIT.cpp#L757 will need to be updated

pmatos:
> https://github.com/FEX-Emu/FEX/blob/39e2a8354f1252b0a8dad624e240866c6fa2ced2/FEXCore/Source/Interface/Core/JIT/JIT.cpp#L757
> 
> will need to be updated

Done.

bylaws (approved):

Sonicadvance1 (approved):

PR #4546, opened 2025-05-01
Make shmdt reset the unmapped pages in the 32-bit allocator

OFFTKP:
in 32-bit mode, shmat allocates similarly to mmap, so shmdt should unmap those pages like munmap

unfortunately untested, as I have no means of testing right now

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/LinuxAllocator.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/LinuxAllocator.cpp
@@ -514,8 +514,20 @@ uint64_t MemAllocator32Bit::Shmdt(const void* shmaddr) {
     return -EINVAL;
   }
 
+  int shmid = it->second;
+  struct shmid_ds buf {};
+  if (shmctl(shmid, IPC_STAT, &buf) == 0) {
+    size_t PagesLength = FEXCore::AlignUp(buf.shm_segsz, FEXCore::Utils::FEX_PAGE_SIZE) >> FEXCore::Utils::FEX_PAGE_SHIFT;
+    SetFreePages(AddrPage, PagesLength);
+  } else {
+    LOGMAN_MSG_A_FMT("Failed to get shm size during shmdt");
+  }
+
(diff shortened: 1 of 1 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/LinuxAllocator.cpp
```diff
@@ -515,7 +515,19 @@ uint64_t MemAllocator32Bit::Shmdt(const void* shmaddr) {
...
 
   uint64_t Result = ::shmdt(shmaddr);
-  PageToShm.erase(it);
+
+  if (Result != -1) {
+    int shmid = it->second;
+    PageToShm.erase(it);
+
+    struct shmid_ds buf {};
+    if (shmctl(shmid, IPC_STAT, &buf) == 0) {
```
Sonicadvance1, line 524: I think this needs to occur before the `shmdt` because if the refcount of the sysv shm region `shm_nattch` reaches zero (which shmdt decrements) then it gets deleted before we can get the information.

Sonicadvance1: Alternative take `PageToShm[LowerPage] = shmid;` in the above `shmat` handler should store ID and size, so we can recover it without doing a shmctl here.

Sonicadvance1:
Hopefully in a real linux kernel, mmap with `MAP_FIXED` on top of a systemv shm range is rejected with EINVAL 😬

Source/Tools/LinuxEmulation/LinuxSyscalls/LinuxAllocator.cpp
OFFTKP: Oh right, fixed. Also moved the `erase` call inside an if, to prevent the rare occasion that it could mess the shmdt errno.

Sonicadvance1:
Seems reasonable to me, going to run it through some testing first though.

Sonicadvance1 (approved):

PR #4547, opened 2025-05-02
Linux/SMCTracking: Stop calling mprotect on a memory region times the number of threads

Sonicadvance1:
I noticed this cascade of mprotects when poking at Crypt of the Necrodancer, since it consistently is invalidating code. I saw us calling mprotect on the same page 32 times in a tight loop and thought surely this isn't FEX doing this.

Turns out we were calling the callback after invalidating each thread. It should instead be done once at the end of invalidating the thread's caches while still holding the locks.

Fixes this weird cascade of mprotects that equal the number of FEX threads.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
@@ -191,7 +191,7 @@ public:
   }
 
   void InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState* CallingThread, uint64_t Start, uint64_t Length,
-                                FEXCore::Context::CodeRangeInvalidationFn callback) {
+                                FEXCore::Context::CodeRangeInvalidationFn after_callback) {
     std::lock_guard lk(ThreadCreationMutex);
 
     // Potential deferred since Thread might not be valid.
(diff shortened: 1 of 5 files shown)


lioncash (approved):
oof, nice fix

neobrain (approved):
I had implemented this as part of #4479, but no reason not to merge it on its own.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h
```diff
@@ -200,8 +200,11 @@ class ThreadManager final {
     auto CodeInvalidationlk = GuardSignalDeferringSectionWithFallback(CTX->GetCodeInvalidationMutex(), CallingThread);
 
     for (auto& Thread : Threads) {
-      CTX->InvalidateGuestCodeRange(Thread->Thread, Start, Length, callback);
+      CTX->InvalidateGuestCodeRange(Thread->Thread, Start, Length);
```
neobrain, line 203: I think we can remove the `callback` parameter from this function's signature now, as it's never used anywhere.

Sonicadvance1: Removed.

PR #4551, opened 2025-05-03
Fix 32-bit fadvise64

OFFTKP:
The 32-bit and 64-bit fadvise64 is not the same, the 32-bit version has a high+low offset and a 32-bit len

https://github.com/torvalds/linux/blob/master/arch/x86/kernel/sys_ia32.c#L112


Sonicadvance1 (approved):
Oop, good shout, I missed that.

Sonicadvance1:
`sh -c "CLANG_FORMAT=clang-format-16 ./Scripts/reformat.sh . > /dev/null 2>&1"` locally to get code formatting CI to pass.
Useful to throw in to a little bash script for executing before committing.

PR #4555, opened 2025-05-05
Win32/CRT: Remove unused math functions

Sonicadvance1:
These are no longer used now that Cephes is builtin.


Sonicadvance1:
Turns out these are still used for the fp64 variants.

PR #4562, opened 2025-05-07
Windows: Fix building with llvm-libcxx

WhatAmISupposedToPutHere:
Libcxx uses GetSystemTimePreciseAsFileTime if _WIN32_WINNT specifies a version new enough to have it.


Sonicadvance1 (commented):
I'll leave this up to @bylaws but it at least looks reasonable to me.

bylaws (approved):
This is fine but be aware this configuration might break in the future.

Sonicadvance1 (approved):

Issue #4557, opened 2025-05-06
`VMA Tracking corruption` is a vengeful spirit


Sonicadvance1:
I have been getting hit by this hard recently. It seems to occur fairly frequently with Steam when it is hammering a big game download.
In my case I'm hitting it with `Spider-Man 2` quite consistently.

Backtrace:
```
#0  FEXCore::Assert::ForcedAssert () at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Utils/ForcedAssert.cpp:9
9         asm volatile("hlt #1");
[Current thread is 1 (Thread 0xc1438cc0 (LWP 72746))]
(gdb) bt
#0  FEXCore::Assert::ForcedAssert () at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Utils/ForcedAssert.cpp:9
#1  0x0000aaaae53d8be0 in LogMan::Throw::MFmt (fmt=<optimized out>, args=...) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Utils/LogManager.cpp:29
#2  0x0000aaaae52dd964 in LogMan::Throw::AFmt<>(bool, char const*) (Value=<optimized out>, fmt=<optimized out>) at /mnt/Work/Work/work/FEXNew/FEXCore/include/FEXCore/Utils/LogManager.h:53
#3  FEX::HLE::SyscallHandler::TrackMremap (this=0xffff94622000, Thread=0x48e88b000, OldAddress=3100446720, OldSize=<optimized out>, NewSize=<optimized out>, flags=1, NewAddress=2937729024) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp:283
#4  0x0000aaaae52ea724 in FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_4::operator()(FEXCore::Core::CpuStateFrame*, void*, unsigned long, unsigned long, int, void*) const (Frame=0x48e88b0a0, old_address=0xb8cd1000, old_size=806912, new_size=1052672, flags=1, new_address=<optimized out>, this=<optimized out>) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp:95
[...]

PR #4561, opened 2025-05-07
LinuxEmulation: Minor cleanup by separating VMA definitions

Sonicadvance1:
NFC, just moving this to its own header. It's already a huge PITA to read. I just want to try and preserve some sanity while attempting to fix #4557

PR #4563, opened 2025-05-07
SyscallsVMATracking: Minor cleanup

Sonicadvance1:
Just some minor changes and rebrands to make this feel less like NIH and help with making it easier to parse what is going on while I work towards fixing #4557

### DIFF
--- a/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
+++ b/FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
@@ -44,6 +44,15 @@ public:
   void StealAndDropActiveLocks() {
     Mutex = PTHREAD_MUTEX_INITIALIZER;
   }
+
+  // Checks and asserts if the mutex isn't owned by the calling thread.
+  void check_lock_owned_by_self() {
+    const auto Result = pthread_mutex_lock(&Mutex);
+    if (Result == 0) [[unlikely]] {
+      ERROR_AND_DIE_FMT("User of unique lock must have already locked mutex as write!");
+    }
+  }
+
 private:
   pthread_mutex_t Mutex;
 };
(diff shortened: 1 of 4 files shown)


FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
```diff
@@ -85,6 +94,15 @@ class ForkableSharedMutex final {
     const auto Result = pthread_rwlock_tryrdlock(&Mutex);
     return Result == 0;
   }
+
+  // Checks and asserts if the rwlock isn't owned by the calling thread with write semantics.
```
neobrain, line 98:
```suggestion
  // Asserts that the rwlock isn't exclusively owned by the calling thread.
```

FEXCore/include/FEXCore/Utils/SignalScopeGuards.h
```diff
@@ -85,6 +94,15 @@ class ForkableSharedMutex final {
...
+  void check_lock_owned_by_self_as_write() {
+    const auto Result = pthread_rwlock_wrlock(&Mutex);
+    if (Result == 0) [[unlikely]] {
+      ERROR_AND_DIE_FMT("User of rwlock must have already locked mutex as write!");
```
neobrain, line 102: Didn't we have an assert function that combines the check with the assert?

neobrain (commented):
FWIW disk caching interacts a lot with these interfaces, so I'd appreciate if the changes that will be made to this land in a single PR (that I can dedicate time checking for conflicts) rather than multiple smaller ones. #4561 is probably fine if it's just moving around code blocks though.

Sonicadvance1:
Review comments addressed

Sonicadvance1:
Rebased.

lioncash (approved):

PR #4566, opened 2025-05-09
ThreadPoolAllocator: Add support for updating the size of managed data

neobrain:
This will be used in #4479 to manage temporary buffers for parallel compilation.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
+++ b/FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
@@ -149,6 +149,30 @@ public:
   }
 
   /**
+   * @brief Try to reown a buffer that was previously disowned
+   *
+   * @param Buffer - The buffer we previously disowned
+   * @param Size - The size of the buffer
+   * @param CurrentClientFlag - The client tracked flag
+   *
+   * Once DisownBuffer has been called, it is unsafe to use the buffer until it has been reowned
+   * Always reown a buffer before use!
+   *
+   * @return The original buffer passed in on successful reown, otherwise std::nullopt
+   */
+  std::optional<ContainerType::iterator> TryToReownBuffer(const ContainerType::iterator& Buffer, size_t Size, BufferOwnedFlag* CurrentClientFlag) {
+    ClientFlags Expected = ClientFlags::FLAG_DISOWNED;
+    if (!CurrentClientFlag->compare_exchange_strong(Expected, ClientFlags::FLAG_OWNED)) {
+      return std::nullopt;
+    }
+
+    // If we managed to change the flag from DISOWNED to OWNED then we have successfully reclaimed
+    // Finish setting up state
+    (*Buffer)->LastUsed.store(ClockType::now(), std::memory_order_relaxed);
+    return Buffer;
+  }
+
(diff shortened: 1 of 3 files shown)


FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
```diff
@@ -427,13 +427,20 @@ class FixedSizePooledAllocation final {
...
    *
-   * @return object of type `Type` allocated with at least the size of `Size` from the constructor
+   * @param NewSize Optional new size for managed data
+   *
+   * @return object of type `Type` allocated within the selected buffer
    */
-  Type ReownOrClaimBuffer() {
+  Type ReownOrClaimBuffer(std::optional<size_t> NewSize = std::nullopt) {
+    Size = NewSize.value_or(Size);
+    if (NewSize && ClientOwnedFlag != FEXCore::Utils::IntrusivePooledAllocator::ClientFlags::FLAG_FREE && (*Info)->Size < NewSize.value()) {
```
neobrain, line 439: @Sonicadvance1 Is reading `(*Info)->Size` thread-safe for claimed-but-not-owned buffers? If not, we would have to move this check past ReownOrClaimBuffer and re-try if we find the size is insufficient.

Sonicadvance1:
Instead of hacking in support for dynamic sized buffer allocations in to the object that is supposed to only do fixed size allocations, hence the `FixedSizePooledAllocation` name. Either create a new wrapper that supports dynamically allocating a sized buffer or use the `IntrusivePooledAllocator` directly which understands handling pooled buffers of multiple different sizes.

neobrain:
> Instead of hacking in support for dynamic sized buffer allocations in to the object that is supposed to only do fixed size allocations, hence the FixedSizePooledAllocation name. Either create a new wrapper that supports dynamically allocating a sized buffer or use the IntrusivePooledAllocator directly which understands handling pooled buffers of multiple different sizes.

What do you find hacky about it? The implementation has no requirement for the size to be completely fixed as far as I can see. Perhaps the interface could use a new name, but an allocator optimized for high-frequency allocations of a buffer with rarely changing size seems like a valid use case to me. Copy-pasting the entire interface to do the same thing line-by-line but with one added feature doesn't sound like an improvement.

If it's the `std::optional` parameter that's bothering you, this could be moved to another overload of ReownOrClaimBuffer instead; or we could drop the `Size` from the constructor (putting responsibility on the caller not to change the value too often).

FEXCore/include/FEXCore/Utils/ThreadPoolAllocator.h
Sonicadvance1: It's not safe.

neobrain: Fixed - we'll now first attempt to reown the buffer and check its size, and then disclaim it and claim a new one if needed.

neobrain:
Fixed the issue about unsafe reading of metadata of a non-owned buffer. I've also renamed the helper class to better reflect that its main feature is the timer-based unclaiming of the managed buffer.

Sonicadvance1 (approved):

PR #4572, opened 2025-05-14
JIT: stop using .ID() pattern

alyssarosenzweig:
clean up for now, should let me refactor more stuff but tbd.


Sonicadvance1 (approved):
sed go brr.

PR #4575, opened 2025-05-15
IR: drop DestSize inference

alyssarosenzweig:
In the near future we'll want to create instructions without knowing what instructions wrote the sources... the dest size inference gets in the way of that. Luckily, it's not load bearing. This PR drops it as a preliminary cleanup.


Sonicadvance1 (approved):
Noice.

PR #4576, opened 2025-05-15
Linux/SMCTracking: Fixes nasty race condition causing invalid memory tracking

Sonicadvance1:
This has been a bug that we have technically lived with ever since SMC
tracking was introduced. The problem boils down to the fact that memory
management syscalls from multiple threads can race our SMC tracking.

This was only uncovered due to recent changes in the Steam client where
downloading games has more aggressively started reallocating memory.
This causes Steam to oversubscribe the CPU by a small margin, causing
threads to context switch more heavily during memory management.

The strace that finally managed to capture this:
```
41574 munmap(0xba84e000, 724992 <unfinished ...>
<...>
41227 mmap(NULL, 540672, PROT_READ|PROT_WRITE, MAP_PRIVATE|MAP_ANONYMOUS, -3, 0 <unfinished ...>
<...>
41574 <... munmap resumed>)             = 0
<...>
41227 <... mmap resumed>)               = 0xba87b000
```

While FEX's tracking linearly was:
```
mmap, 0xba87b000, 0x84000, 0x3, 0x22, 0xfffffffd, 0x0
munmap, 0xba84e000, 0xb1000
```

The way munmap and mmap perfectly interleave while getting context switched meant that the kernel's view of munmap then mmap didn't match our view of mmap completing first then munmap happening afterwards.
The kernel/strace is obviously the correct view in this instance.

This all comes down to how these threads are racing the VMA tracking
mutex after the syscall happens and not guaranteeing sequential
consistency that matches the kernel's view.

The only way to correct this sanely is to extend the locking period to
also encompass the syscalls getting executed. This is a bit tricky since
the VMA tracking needs to ensure that the lock is no longer held once
ThreadManager invalidation occurs so a callback to do the syscall
operation is about the only sane approach here. Luckily we now have
fextl::move_only_function.

Fixes consistent crashes with Steam game downloads (and maybe some
chromium crashes?)

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -154,186 +154,102 @@ FEXCore::HLE::AOTIRCacheEntryLookupResult SyscallHandler::LookupAOTIRCacheEntry(
 }
 
 // MMan Tracking
-void SyscallHandler::TrackMmap(FEXCore::Core::InternalThreadState* Thread, uintptr_t Base, uintptr_t Size, int Prot, int Flags, int fd,
-                               off_t Offset) {
-  Size = FEXCore::AlignUp(Size, FEXCore::Utils::FEX_PAGE_SIZE);
+void SyscallHandler::TrackMmap(FEXCore::Core::InternalThreadState* Thread, uint64_t addr, size_t length, int prot, int flags, int fd, off_t offset) {
+  size_t Size = FEXCore::AlignUp(length, FEXCore::Utils::FEX_PAGE_SIZE);
 
-  if (Flags & MAP_SHARED) {
-    CTX->MarkMemoryShared(Thread);
-  }
+  VMATracking::MappedResource* Resource = nullptr;
 
-  {
-    // NOTE: Frontend calls this with a nullptr Thread during initialization, but
-    //       providing this code with a valid Thread object earlier would allow
-    //       us to be more optimal by using GuardSignalDeferringSection instead
-    auto lk = FEXCore::GuardSignalDeferringSectionWithFallback(VMATracking.Mutex, Thread);
+  if (!(flags & MAP_ANONYMOUS)) {
+    struct stat64 buf;
+    fstat64(fd, &buf);
+    VMATracking::MRID mrid {buf.st_dev, buf.st_ino};
 
-    static uint64_t AnonSharedId = 1;
+    char Tmp[PATH_MAX];
+    auto PathLength = FEX::get_fdpath(fd, Tmp);
 
-    VMATracking::MappedResource* Resource = nullptr;
-
-    if (!(Flags & MAP_ANONYMOUS)) {
-      struct stat64 buf;
-      fstat64(fd, &buf);
-      VMATracking::MRID mrid {buf.st_dev, buf.st_ino};
-
-      char Tmp[PATH_MAX];
-      auto PathLength = FEX::get_fdpath(fd, Tmp);
-
-      if (PathLength != -1) {
-        Tmp[PathLength] = '\0';
-        auto [Iter, Inserted] = VMATracking.EmplaceMappedResource(mrid, VMATracking::MappedResource {nullptr, nullptr, 0});
-        Resource = &Iter->second;
-
-        if (Inserted) {
-          Resource->AOTIRCacheEntry = CTX->LoadAOTIRCacheEntry(fextl::string(Tmp, PathLength));
-          Resource->Iterator = Iter;
-        }
-      }
-    } else if (Flags & MAP_SHARED) {
-      VMATracking::MRID mrid {VMATracking::SpecialDev::Anon, AnonSharedId++};
-
+    if (PathLength != -1) {
+      Tmp[PathLength] = '\0';
       auto [Iter, Inserted] = VMATracking.EmplaceMappedResource(mrid, VMATracking::MappedResource {nullptr, nullptr, 0});
-      LOGMAN_THROW_A_FMT(Inserted == true, "VMA tracking error");
       Resource = &Iter->second;
-      Resource->Iterator = Iter;
-    } else {
-      Resource = nullptr;
+
+      if (Inserted) {
+        Resource->AOTIRCacheEntry = CTX->LoadAOTIRCacheEntry(fextl::string(Tmp, PathLength));
+        Resource->Iterator = Iter;
+      }
     }
+  } else if (flags & MAP_SHARED) {
+    VMATracking::MRID mrid {VMATracking::SpecialDev::Anon, AnonSharedId++};
 
-    VMATracking.TrackVMARange(CTX, Resource, Base, Offset, Size, VMATracking::VMAFlags::fromFlags(Flags), VMATracking::VMAProt::fromProt(Prot));
+    auto [Iter, Inserted] = VMATracking.EmplaceMappedResource(mrid, VMATracking::MappedResource {nullptr, nullptr, 0});
+    LOGMAN_THROW_A_FMT(Inserted == true, "VMA tracking error");
+    Resource = &Iter->second;
+    Resource->Iterator = Iter;
+  } else {
+    Resource = nullptr;
   }
 
-  if (SMCChecks != FEXCore::Config::CONFIG_SMC_NONE) {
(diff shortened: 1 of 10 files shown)


lioncash (approved):
As far as I can tell this looks good. Might be nice to get another set of eyes on it too, though

neobrain (changes requested):
Phew, great work figuring this out!

For the actual change, pushing code *down* the stack to fix the issue isn't ideal: It complicates the control flow in SyscallsSMCTracking.cpp and breaks separation of concerns (SMCTracking was mainly FEX-internal accounting before, now it changes system state as well). The lambdas everywhere just make it complicated to grasp who's doing what anymore.

What if we instead pushed mutex management *up* the stack into the syscall handlers, along with the calls to MarkMemoryShared and InvalidateGuestCodeRange? I think that would be much cleaner than splitting half the logic into lambdas and the other half into SMCTracking itself.

At first I thought this required type-erasing the mutexes, but the interfaces here are non-virtual Linux-specific anyway, so we can just use SignalScopeGuard directly.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -201,54 +206,80 @@ void SyscallHandler::TrackMmap(FEXCore::Core::InternalThreadState* Thread, uintp
...
+      return Result;
+    }
 
-    VMATracking.DeleteVMARange(CTX, Base, Size);
+    VMATracking.DeleteVMARange(CTX, reinterpret_cast<uintptr_t>(addr), Size);
   }
 
   if (SMCChecks != FEXCore::Config::CONFIG_SMC_NONE) {
-    _SyscallHandler->TM.InvalidateGuestCodeRange(Thread, (uintptr_t)Base, Size);
+    _SyscallHandler->TM.InvalidateGuestCodeRange(Thread, reinterpret_cast<uintptr_t>(addr), Size);
```
neobrain, line 239: For personal reference, is this affected by the same race condition? A sequence of munmap followed by mmap could be scheduled in such a way that we call InvalidateGuestCodeRange after having mapped the code.

This is probably fine since it would additionally require code to be compiled before InvalidateGuestCodeRange gets called (leaving a very slim time window), and even then the worst that'll happen is probably a recompile. But with code caching, the effect would be that a cache that we just loaded gets discarded immediately, so I should double-check this interaction.

Sonicadvance1: Currently this could just cause some additional invalidation (that would have already occurred), plus there's some nasty interactions between recursive mutex locking that can't be solved, which is why this mutex juggle needs to occur like this.

neobrain:
> My original approach was actually pushing the mutex management up and pulling the tracking logic towards syscall handlers was actually worse off because it needed to do a bunch of code duplication between 32-bit and 64-bit syscalls. So this is definitely the less painful route to go.

I didn't see any logic that would need duplication beyond the three things I mentioned. Is there any particular syscall where this would be an obvious issue?

Sonicadvance1:
> > My original approach was actually pushing the mutex management up and pulling the tracking logic towards syscall handlers was actually worse off because it needed to do a bunch of code duplication between 32-bit and 64-bit syscalls. So this is definitely the less painful route to go.
> 
> I didn't see any logic that would need duplication beyond the three things I mentioned. Is there any particular syscall where this would be an obvious issue?

mmap, munmap, shmat, and shmdt would require duplication between 32-bit and 64-bit because the implementations call different allocators to handle nuance differences.

Additionally 32-bit shmat/shmdt has two implementations which slightly differ in behaviour which need to be accounted for, which is kind of annoying, but if the implementation is getting duplicated for 32-bit and 64-bit then it would just live as a detail in the 32-bit implementation.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp
```diff
@@ -116,28 +97,23 @@ void RegisterMemory(FEX::HLE::SyscallHandler* Handler) {
...
-      return ResultAddr;
-    } else {
+    return FEX::HLE::_SyscallHandler->EmulateShmat(Frame->Thread, shmid, shmaddr, shmflg, [](int shmid, const void* shmaddr, int shmflg) -> uint64_t {
+      uint32_t ResultAddr {};
+      uint64_t Result = static_cast<FEX::HLE::x32::x32SyscallHandler*>(FEX::HLE::_SyscallHandler)
+                          ->GetAllocator()
+                          ->Shmat(shmid, reinterpret_cast<const void*>(shmaddr), shmflg, &ResultAddr);
+      if (!FEX::HLE::HasSyscallError(Result)) {
+        return ResultAddr;
+      }
```
neobrain, line 107: (Continuing conversation from main thread here)

> Additionally 32-bit shmat/shmdt has two implementations which slightly differ in behaviour which need to be accounted for, which is kind of annoying, but if the implementation is getting duplicated for 32-bit and 64-bit then it would just live as a detail in the 32-bit implementation.

Doesn't this work?

```suggestion
    CTX->MarkMemoryShared(Thread);

    auto Lock = GuardSignalDeferringSection { ... };

    shmid_ds stat;
    auto res = shmctl(shmid, IP_STAT, &stat);
    LOGMAN_THROW_A_FMT(res != -1);

    uint32_t ResultAddr {};
    uint64_t Result = static_cast<FEX::HLE::x32::x32SyscallHandler*>(FEX::HLE::_SyscallHandler)
                        ->GetAllocator()
                        ->Shmat(shmid, reinterpret_cast<const void*>(shmaddr), shmflg, &ResultAddr);

    if (!FEX::HLE::HasSyscallError(Result)) {
      FEX::HLE::_SyscallHandler->TrackShmat(Frame->Thread, shmid, ResultAddr, shmflg);
      FEX::HLE::_SyscallHandler->InvalidateIfNeeded(...); // New interface (checking for SMC_NONE)
      return ResultAddr;
    } else {
```

It only needs a couple of lines to be repeated twice, and it saves the need to pass around function pointers with inversion of control flow. All the system-interfacing bits live in the actual syscalls code then, while our FEX internal VMA tracking is self-contained in SyscallsSMCTracking.cpp.

Sonicadvance1: https://github.com/FEX-Emu/FEX/pull/4581
See if you like that code duplication.

neobrain:
Cheers, I'll take a look to ensure the changes from #4583 extend to the new code here. Might take me a few days though since I'm on a conference starting from today.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp
```diff
@@ -24,32 +26,54 @@ namespace FEX::HLE::x32 {
...
+      return reinterpret_cast<void*>(Result);
+    }
+
+    LOGMAN_THROW_A_FMT((Result >> 32) == 0 || (Result >> 32) == 0xFFFFFFFF, "values must fit to 32 bits");
 
-  if (!FEX::HLE::HasSyscallError(Result)) {
     FEX::HLE::_SyscallHandler->TrackMmap(Thread, Result, length, prot, flags, fd, offset);
-    return (void*)Result;
-  } else {
-    errno = -Result;
```
neobrain, line 35: What happened to this errno write? Not seeing any obvious match in the new version.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp
```diff
@@ -24,32 +26,54 @@ namespace FEX::HLE::x32 {
...
-    return MAP_FAILED;
   }
+
+  FEX::HLE::_SyscallHandler->InvalidateCodeRangeIfNecessary(Thread, Result, Size);
+  return reinterpret_cast<void*>(Result);
 }
 
-int x32SyscallHandler::GuestMunmap(FEXCore::Core::InternalThreadState* Thread, void* addr, uint64_t length) {
-  LOGMAN_THROW_A_FMT((uintptr_t(addr) >> 32) == 0, "values must fit to 32 bits");
-  LOGMAN_THROW_A_FMT((length >> 32) == 0, "values must fit to 32 bits");
```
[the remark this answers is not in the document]
Sonicadvance1: Added back.

neobrain:
The commit message for the first patch needs a quick redo though, since it was written before integrating my proposals (still talks about `move_only_function` etc).

PR #4578, opened 2025-05-15
Use -print-libgcc-file-name to get the compiler-rt file name

cjacek:
Avoid hardcoding the file name. Upstream Clang currently expects the aarch64 variant of compiler-rt. Changing this to use arm64ec in the file name could be problematic in the future, if the MinGW toolchain gains support for ARM64X. Querying Clang for the correct file name is the most forward-compatible approach.

### DIFF
--- a/Source/Windows/ARM64EC/CMakeLists.txt
+++ b/Source/Windows/ARM64EC/CMakeLists.txt
@@ -23,7 +23,8 @@ target_link_libraries(arm64ecfex
   ntdll_ex
 )
 
-target_link_options(arm64ecfex PRIVATE -static -nostdlib -nostartfiles -nodefaultlibs -lc++ -lc++abi -lunwind -lclang_rt.builtins-arm64ec)
+target_link_options(arm64ecfex PRIVATE -static -nostdlib -nostartfiles -nodefaultlibs -lc++ -lc++abi -lunwind)
+target_link_libraries(arm64ecfex PRIVATE ${LIBGCC_PATH})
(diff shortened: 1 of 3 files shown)


Sonicadvance1 (approved):
Reasonable enough for me and I see it correctly pulls `libclang_rt.builtins-arm64ec` as expected.

Source/Windows/ARM64EC/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ target_link_libraries(arm64ecfex
   ntdll_ex
 )
 
-target_link_options(arm64ecfex PRIVATE -static -nostdlib -nostartfiles -nodefaultlibs -lc++ -lc++abi -lunwind -lclang_rt.builtins-arm64ec)
+target_link_options(arm64ecfex PRIVATE -static -nostdlib -nostartfiles -nodefaultlibs -lc++ -lc++abi -lunwind ${libgcc_path})
```
neobrain, line 26: CMake variables should be ALL_UPPERCASE. We don't use lowercase variables anywhere else.

neobrain, line 26: Should this be moved to a separate `target_link_libraries` line? On my local system, clang and gcc both return a plain filename without `-l` prefix; not sure if mingw behaves differently though.

cjacek: Fixed in the new version.

bylaws (approved):

PR #4580, opened 2025-05-16
IR: Inline registers into the IR

alyssarosenzweig:
Rather than keeping SSA all the way thru JIT and putting allocated registers in a sideband RAData data structure, just encode registers inline in the IR. This significantly simplifies the RA, and I measured a small improvement to JIT time (on the order of 1%). It's also a stepping stone to making post-RA optimizations sensible.


FEXCore/Source/Interface/IR/IR.h
```diff
@@ -318,6 +358,7 @@ class OrderedNode final {
   }
 
   void AddUse() {
+    LOGMAN_THROW_A_FMT(NumUses < 65535, "never exceeded in practice");
```
Sonicadvance1, line 361: I wonder if this would get hit with older XeSS and large enough blocks.

alyssarosenzweig: I have a long-standing yak shave to remove NumUses, perhaps I should just do that as a prerequisite for this here yak shave ....

alyssarosenzweig: I've dropped this change. OrderedNode is grown a bit which is unfortunate, but I measured only a slight % difference attributable here, and this series is still a win overall to JIT time. So we should be able to merge it now and then circle back to this later for even more zoomies.

Sonicadvance1 (approved):
Looks sick, no complaints here.

PR #4581, opened 2025-05-16
Vma race the neobrain cut

Sonicadvance1:


neobrain:
Thanks for trying out this path! The duplication is unfortunate, but I think it *is* much more straightforward to read. Something seems to have broken the unit tests (and Steam doesn't start up with `_GLIBCXX_DEBUG` anymore) though.

I've pushed some ideas to cut down the code duplication in #4583, rebasing on that might make debugging a bit easier, otherwise I can apply the cleanup after this PR is fixed and merged. In combination, this should result in something very passable.

Sonicadvance1:
Fixed all the bugs with this quick hacked together version and merged back to #4576

PR #4583, opened 2025-05-18
LinuxSyscalls: Reduce code duplication between 32-bit and 64-bit paths

neobrain:
We can deal with the differences between these much more easily with a dedicated Is64Bit parameter instead of using runtime polymorphism.

Additionally, moving `Get32BitAllocator` to the parent class as a virtual member function makes the 32-bit implementation closer to the 64-bit one.

~~(Marking as draft until we decide whether to merge #4581 first or later.)~~

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -274,6 +274,30 @@ uint64_t SyscallHandler::GuestShmat(bool Is64Bit, FEXCore::Core::InternalThreadS
   return Result;
 }
 
+uint64_t SyscallHandler::GuestShmdt(bool Is64Bit, FEXCore::Core::InternalThreadState* Thread, const void* shmaddr) {
+  uint64_t Result {};
+  uint64_t Length {};
+  {
+    auto lk = FEXCore::GuardSignalDeferringSection(FEX::HLE::_SyscallHandler->VMATracking.Mutex, Thread);
+    if (Is64Bit) {
+      Result = ::shmdt(shmaddr);
+      if (Result == -1) {
+        return Result;
+      }
+    } else {
+      Result = FEX::HLE::_SyscallHandler->Get32BitAllocator()->Shmdt(shmaddr);
+      if (FEX::HLE::HasSyscallError(Result)) {
+        return Result;
+      }
+    }
+
+    Length = FEX::HLE::_SyscallHandler->TrackShmdt(Thread, reinterpret_cast<uintptr_t>(shmaddr));
+  }
+
+  FEX::HLE::_SyscallHandler->InvalidateCodeRangeIfNecessary(Thread, reinterpret_cast<uintptr_t>(shmaddr), Length);
+  return Result;
+}
+
 // MMan Tracking
 void SyscallHandler::TrackMmap(FEXCore::Core::InternalThreadState* Thread, uint64_t addr, size_t length, int prot, int flags, int fd, off_t offset) {
   size_t Size = FEXCore::AlignUp(length, FEXCore::Utils::FEX_PAGE_SIZE);
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -221,6 +221,24 @@ uint64_t SyscallHandler::GuestMunmap(bool Is64Bit, FEXCore::Core::InternalThread
   return Result;
 }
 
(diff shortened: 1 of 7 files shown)


neobrain:
4591 is merged, time to shine.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -153,6 +153,177 @@ FEXCore::HLE::AOTIRCacheEntryLookupResult SyscallHandler::LookupAOTIRCacheEntry(
...
+uint64_t SyscallHandler::GuestMremap(bool Is64Bit, FEXCore::Core::InternalThreadState* Thread, void* old_address, size_t old_size,
+                                     size_t new_size, int flags, void* new_address) {
+  uint64_t Result {};
+
+  {
+    auto lk = FEXCore::GuardSignalDeferringSection(FEX::HLE::_SyscallHandler->VMATracking.Mutex, Thread);
+    if (Is64Bit) {
+      Result = reinterpret_cast<uint64_t>(::mremap(old_address, old_size, new_size, flags, new_address));
+      if (Result == -1) {
+        return Result;
```
Sonicadvance1: This one is a little weird, it should return return `-errno` here so that it always returns combined errno result.
Then on the 64-bit syscall handler it can just return the result directly.
Currently this code is a weird edge of 64-bit returns result plus errno, where 32-bit returns single result.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -153,6 +153,177 @@ FEXCore::HLE::AOTIRCacheEntryLookupResult SyscallHandler::LookupAOTIRCacheEntry(
...
+  uint64_t Result {};
+  uint64_t Length {};
+  CTX->MarkMemoryShared(Thread);
+
+  {
+    auto lk = FEXCore::GuardSignalDeferringSection(FEX::HLE::_SyscallHandler->VMATracking.Mutex, Thread);
+    if (Is64Bit) {
+      Result = reinterpret_cast<uint64_t>(::shmat(shmid, shmaddr, shmflg));
+      if (Result == -1) {
+        return Result;
```
Sonicadvance1: Same situation here as `GuestRemap`

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -153,6 +153,177 @@ FEXCore::HLE::AOTIRCacheEntryLookupResult SyscallHandler::LookupAOTIRCacheEntry(
...
+
+uint64_t SyscallHandler::GuestShmdt(bool Is64Bit, FEXCore::Core::InternalThreadState* Thread, const void* shmaddr) {
+  uint64_t Result {};
+  uint64_t Length {};
+  {
+    auto lk = FEXCore::GuardSignalDeferringSection(FEX::HLE::_SyscallHandler->VMATracking.Mutex, Thread);
+    if (Is64Bit) {
+      Result = ::shmdt(shmaddr);
+      if (Result == -1) {
+        return Result;
```
Sonicadvance1: Same situation here as `GuestRemap`

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -153,6 +153,177 @@ FEXCore::HLE::AOTIRCacheEntryLookupResult SyscallHandler::LookupAOTIRCacheEntry(
...
+      }
+    }
+    FEX::HLE::_SyscallHandler->TrackMremap(Thread, reinterpret_cast<uint64_t>(old_address), old_size, new_size, flags, Result);
+  }
+
+  FEX::HLE::_SyscallHandler->InvalidateCodeRangeIfNecessaryOnRemap(Thread, reinterpret_cast<uint64_t>(old_address), Result, old_size, new_size);
+  return Result;
+}
+
+uint64_t SyscallHandler::GuestMprotect(FEXCore::Core::InternalThreadState* Thread, void* addr, size_t len, int prot) {
```
Sonicadvance1, line 250: `GuestMprotect` is the only function out of these seven `Guest*` that don't return a combined result+errno value instead keeping it separate. This is a weird contrast and instead it should just return -errno on error and remove the `SYSCALL_RET()` in the handlers.

This makes all the "Guest" handlers return a combined result for consistency.

neobrain: Good point, fixed in latest version.

Sonicadvance1 (approved):

PR #4585, opened 2025-05-19
Pair push/pop

alyssarosenzweig:
Implement a simple peephole pass to pair push/pops.

Depends on #4580 - without those changes, this pass would be substantially more complex, messy, and slow.

---

There's a lot more post-RA optimization we can do, this is just the beginning :smile:


unittests/InstructionCountCI/FlagM/Primary_32Bit.json
```diff
@@ -247,67 +247,53 @@
...
-        "str w4, [x20, #-4]!",
-        "str w7, [x20, #-4]!",
-        "str w5, [x20, #-4]!",
-        "str w6, [x20, #-4]!",
-        "str w8, [x20, #-4]!",
-        "str w9, [x20, #-4]!",
-        "str w10, [x20, #-4]!",
-        "mov w8, w20",
-        "str w11, [x8, #-4]!"
+        "mov x20, x8",
```
Sonicadvance1, line 253: This move is actually unsafe and the guest stack will get corrupted on asynchronous signal. Need to ensure not to allow moves like this.

alyssarosenzweig: This isn't a regression from the previous code generated for the same instruction, no?

alyssarosenzweig: I'm redoing the pusha stuff anyway

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -557,9 +563,60 @@ void ConstrainedRAPass::Run(IREmitter* IREmit_) {
...
+      // Delete instructions that only exist for RA
+      if (IsTrivial(CodeNode, IROp)) {
+        IREmit->RemovePostRA(CodeNode);
+      }
     }
 
     LOGMAN_THROW_A_FMT(SourceIndex == 0, "Consistent source count in block");
+
+    // Post-RA peephole optimization.
+    Ref LastNode = nullptr;
```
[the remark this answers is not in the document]
alyssarosenzweig: Separate function definitely. Separate file... *grumbles about fex's pass manager being annoying*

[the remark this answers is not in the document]
alyssarosenzweig: Reworked the whole thing to speed up the JIT but now this doesn't really apply

Sonicadvance1 (approved):
Nice nice nice nice.

PR #4596, opened 2025-05-30
Error when JSON config includes unknown options

pmatos:


pmatos:
When I was trying to figure out what was going on with another issue, I noticed that I had "MultiBlock" in my config file, which clearly was being ignored because the option is "Multiblock". This is worth it to ensure we properly validate the JSON config.

pmatos:
@Sonicadvance1 The option UnifiedMemory is set in the hosted runners but doesn't exist anymore as found by the patch. Can we remove it from the runners in order to get this passing?

```
Error:  Unknown configuration option 'UnifiedMemory' in JSON config file '/home/buildbot/.fex-emu/Config.json'
```

Sonicadvance1:
Maybe a less "nuclear" option is to print a log instead of asserting so it can still be obvious to the user that a config option was incorrectly set?

pmatos:
> Maybe a less "nuclear" option is to print a log instead of asserting so it can still be obvious to the user that a config option was incorrectly set?

Sure - lets just print a log then. That sounds like the most pragmatic thing to do.

pmatos:
Pushed a non-nuclear option.

Sonicadvance1 (approved):

PR #4597, opened 2025-05-30
Constant fold on the fly

alyssarosenzweig:
Because we don't optimize across instructions pre-RA (and this is unlikely to change), ConstProp exists only to clean up the mess left behind by OpcodeDispatcher. That means lots of instruction sequences go thru the JIT like this:

   Generate bad code --> Replace with better code --> Delete the original bad code

It'd be a lot faster to just generate good code in the first place. This PR does just that, updating almost every bit of the frontend to generate good code, and then deleting all the algebraic optimizations & constant folding from constprop. That lets us simplify/speed up constprop quite a bit. nodejs jit time is **2.5%** faster across this series, and there are obvious follow ups for even more zoomies but this is already a large PR.


unittests/InstructionCountCI/Primary.json
```diff
@@ -4349,13 +4333,13 @@
     "mov ah, 0x0": {
       "ExpectedInstructionCount": 1,
       "ExpectedArm64ASM": [
-        "and x4, x4, #0xffffffffffff00ff"
+        "and x4, x4, #0xffffffffffffff00"
```
Sonicadvance1, line 4336: This looks like a bug

unittests/InstructionCountCI/Primary.json
```diff
@@ -4349,13 +4333,13 @@
...
       ]
     },
     "xor ah, ah": {
       "ExpectedInstructionCount": 2,
       "ExpectedArm64ASM": [
-        "and x4, x4, #0xffffffffffff00ff",
+        "and x4, x4, #0xffffffffffffff00",
```
Sonicadvance1, line 4342: Samesies

unittests/InstructionCountCI/Primary.json
```diff
@@ -4490,21 +4474,23 @@
...
-      "ExpectedInstructionCount": 3,
+      "ExpectedInstructionCount": 4,
       "Comment": "0xf8",
       "ExpectedArm64ASM": [
-        "mrs x20, nzcv",
-        "orr w20, w20, #0x20000000",
-        "msr nzcv, x20"
+        "mov w20, #0x1",
+        "mrs x21, nzcv",
+        "bfi w21, w20, #29, #1",
```
Sonicadvance1, line 4482: What happened here to clc/stc?

unittests/InstructionCountCI/VEX_map1.json
```diff
@@ -489,51 +489,42 @@
       ]
     },
     "vmovmskps rax, ymm0": {
-      "ExpectedInstructionCount": 41,
+      "ExpectedInstructionCount": 32,
```
Sonicadvance1, line 492: That's kind of cute. At some point we should do the same optimization that AVX128 does to make it better, but hard to care when nothing implements SVE with 256-bit wide registers.

unittests/InstructionCountCI/Primary.json
alyssarosenzweig: this is the non-flagm path. speeding it up would add overhead to the flagm path and in terms of actual uops this isn't worse. so I accepted the bandage hit on v8.0 in exchange for not needing Even More Code of dubious value.

alyssarosenzweig:
fixed & added a unit test.

Sonicadvance1:
Still reading through the code, but does this delete cpuid and syscall inlining when the cpuid function or syscall number is a known constant?

Also xgetbv

Sonicadvance1:
Also does this delete l{u,}{div,rem} operation narrowing from 128-bit when source is known 64-bit?

alyssarosenzweig:
> Still reading through the code, but does this delete cpuid and syscall inlining when the cpuid function or syscall number is a known constant?
> 
> Also xgetbv



> Also does this delete l{u,}{div,rem} operation narrowing from 128-bit when source is known 64-bit?

Yes to all of the above, but I'm pretty sure none of these have worked for at least a year. If these actually matter - and I haven't seen evidence they do - we need instcountci for it.

alyssarosenzweig:
3.1% speedup across the series:

```
    N           Min           Max        Median           Avg        Stddev
x 100     0.4472467    0.46646308    0.45708424    0.45713057  0.0040838243
+  50    0.43149892    0.45173984    0.44252267    0.44295575  0.0045621814
	-0.0141748 +/- 0.00144218
	-3.10082% +/- 0.315485%
	(Student's t, pooled s = 0.00424817)
```

Sonicadvance1:
> > Still reading through the code, but does this delete cpuid and syscall inlining when the cpuid function or syscall number is a known constant?
> > Also xgetbv
> 
> > Also does this delete l{u,}{div,rem} operation narrowing from 128-bit when source is known 64-bit?
> 
> Yes to all of the above, but I'm pretty sure none of these have worked for at least a year. If these actually matter - and I haven't seen evidence they do - we need instcountci for it.

Geh. You're right, this is definitely a performance regression because 128-bit divide is uber common in 64-bit code. This sucks.

Sonicadvance1 (approved):
DangerZone

Issue #4483, opened 2025-04-06
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111


kaazoo:
After updating fex-emu-armv8.4 from 2502 to 2504 on an Ubuntu 24.04 host, I get the following error when starting a game from Steam via muvm:
```
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111
[ERROR] Couldn't connect to FEXServer socket /run/user/1001/1001.FEXServer.Socket 2
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111
[ERROR] Couldn't connect to FEXServer socket /run/user/1001/1001.FEXServer.Socket 2
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111
[ERROR] Couldn't connect to FEXServer socket /run/user/1001/1001.FEXServer.Socket 2
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111
[ERROR] Couldn't connect to FEXServer socket /run/user/1001/1001.FEXServer.Socket 2
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket 111
[ERROR] Couldn't connect to FEXServer socket /run/user/1001/1001.FEXServer.Socket 2
[ERROR] Couldn't connect to FEXServer socket 1001.FEXServer.Socket after launching the process
[ERROR] FEXServerClient: Failure to setup client
```
Switching back to fex-emu-armv8.4 2502 makes it work again.

RootFS is Ubuntu 25.10 based on https://github.com/kaazoo/RootFS/blob/main/Configs/Ubuntu_24_10_asahi.json

Steam client version is 1743554648 (stable).

Sonicadvance1:
Sounds like FEXServer is failing to start for some reason. Maybe it's crashing which systemd-coredump could potentially report.
Could also try starting a foreground instance of FEXServer in one terminal with `FEXServer -f -p` and then in another running FEXInterpreter and some basic program like ls, `FEXInterpreter /usr/bin/ls`

Sonicadvance1:
This might be fixed with #4545 merged. Can't test it locally.

WhatAmISupposedToPutHere:
[...]

PR #4599, opened 2025-05-31
Fixes FEXServer path search

pmatos:
Is it ok to use `system` here to ensure that FEXServer exists in PATH?

### DIFF
--- a/Source/Common/FEXServerClient.cpp
+++ b/Source/Common/FEXServerClient.cpp
@@ -11,6 +11,7 @@
 #include <FEXCore/fextl/vector.h>
 #include <FEXHeaderUtils/Filesystem.h>
 
+#include <cstdlib>
 #include <fcntl.h>
 #include <linux/limits.h>
 #include <unistd.h>
@@ -23,6 +24,7 @@
 #include <sys/un.h>
 #include <sys/uio.h>
 #include <thread>
+#include <cstring>
 
 namespace FEXServerClient {
 int RequestPIDFDPacket(int ServerSocket, PacketType Type) {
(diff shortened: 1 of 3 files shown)


pmatos:
This might fix #4483 .

Sonicadvance1:
`execvp` already does a `PATH` search so this would just change where the error message occurs if it is failing to launch the FEXServer.

pmatos:
> `execvp` already does a `PATH` search so this would just change where the error message occurs if it is failing to launch the FEXServer.

ah, but those don't show up. Initially I realized that if there's no FEXServer, there's no way for it to report, which is why I used fprintf. But even with FEX_SILENTLOG=0 FEX_OUTPUTLOG=stderr doesn't work. Then I checked the code and noticed that there's code to set this up properly in FEXLoader but not FEXBash, which is why in FEXBash you don't see any error.

I am travelling now but will think of a fix later.

pmatos:
> > > `execvp` already does a `PATH` search so this would just change where the error message occurs if it is failing to launch the FEXServer.
> > 
> > 
> > ah, but those don't show up. Initially I realized that if there's no FEXServer, there's no way for it to report, which is why I used fprintf. But even with FEX_SILENTLOG=0 FEX_OUTPUTLOG=stderr doesn't work. Then I checked the code and noticed that there's code to set this up properly in FEXLoader but not FEXBash, which is why in FEXBash you don't see any error.
> > I am travelling now but will think of a fix later.
> 
> We should probably just stop starting FEXServer inside of FEXBash, there's no need to do that since FEXInterpreter will find the bash applications in the rootfs itself. The shim program can really be lightened up.

Interesting point. After looking at what was going on, I did seem like the FEXBash could be simplified. I have done that and it looks simpler. Now FEXInterpreter will handle the FEXServer connection.

Source/Common/FEXServerClient.cpp
```diff
@@ -222,11 +224,21 @@ int ConnectToAndStartServer(std::string_view InterpreterPath) {
...
+      InterpreterDir = InterpreterDir.substr(0, LastSlash);
+    }
+
+    fextl::string FEXServerPath = fextl::fmt::format("{}/FEXServer", InterpreterDir);
     // Check if a local FEXServer next to FEXInterpreter exists
     // If it does then it takes priority over the installed one
     if (!FHU::Filesystem::Exists(FEXServerPath)) {
-      FEXServerPath = "FEXServer";
+      // Check if FEXServer exists in PATH
+      if (system("which FEXServer > /dev/null 2>&1") == 0) {
```
Sonicadvance1: I still don't think this system usage is required, it looks like the real fix here is `InterpreterPath` change above is doing the real work, since execvp is still going to scan PATH just like this `system` usage would.

pmatos: Yeah, that's right. You had mentioned that and I missed reverting this hunk. Sorry 'bout that.

Sonicadvance1 (approved):

PR #4603, opened 2025-06-02
LibraryForwarding: Use CMAKE_INSTALL_FULL_LIBDIR instead of constructing the library install paths manually

neobrain:
This fixes issues in the nix build, where CMAKE_INSTALL_LIBDIR is an absolute path instead of the typical "lib(64)".


Sonicadvance1 (approved):
Sure.

PR #4614, opened 2025-06-12
Merge moves that are immediately consumed

alyssarosenzweig:
This implements a basic post-RA peephole optimization for the following
    patterns:
    
       mov xA, xB
       op xA, ... xA ..
    
       mov wA, wB
       op wA, ... wA ..
    
       mov xA, xB
       and wA, wA, ..
    
    ..and merges the mov into the second instruction. This is not as powerful as a
    real copyprop would be (even post-RA) but it can be done very cheaply. Doing the
    full thing is going to cost more JIT time for diminishing returns, so we'd need
    to evaluate if it's worth it as a marginal improvement over this dumb version.

With n=100, no JIT time difference at 80% confidence. Minimum/median JIT times hurt 0.3% across the series but again this is within statistical noise.

---

I'm unsure if the cost:benefit is there for full post-RA copy-prop, we'll see I guess. This is an easy place to start though, and it's definitely correct even with synchronous faults.


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -988,7 +988,8 @@ void ContextImpl::AddThunkTrampolineIRHandler(uintptr_t Entrypoint, uintptr_t Gu
     const auto GPRSize = GetGPROpSize();
 
     if (GPRSize == IR::OpSize::i64Bit) {
-      emit->_StoreRegister(emit->_Constant(Entrypoint), X86State::REG_R11, IR::GPRClass, GPRSize);
+      IR::Ref R = emit->_StoreRegister(emit->_Constant(Entrypoint), GPRSize);
+      R->Reg = IR::PhysicalRegister(IR::GPRClass, X86State::REG_R11).Raw;
```
Sonicadvance1: `IR::GPRClass` needs to be `IR::GPRFixedClass` otherwise Thunking is broken.

alyssarosenzweig: fixed, thanks

Sonicadvance1 (approved):

anonymususer04:
@alyssarosenzweig bicha veia do krai

PR #4615, opened 2025-06-13
Improve log message formatting

neobrain:
* Changed "ERROR" strings to single-letters to reduce noise and to ensure message alignment
* Added mild use of colors
* Time stamps:
  * Fixed zero-padding for time stamps (previously `0.1234` would be printed when the actual time was `0.000001234`)
  * Using milliseconds instead of nanoseconds now, which should be more than sufficient and reduces clutter
  * Using time stamps relative to the first log message now
* Process ID and thread ID are separated using `|` instead of `.` to facilitate copy-paste
* Fixed `^C` being printed before the final message when exiting FEXServer running in the foreground

Before:
log_before

After:
log_after

### DIFF
--- a/Source/Windows/Common/Logging.cpp
+++ b/Source/Windows/Common/Logging.cpp
@@ -23,7 +23,7 @@ static void MsgHandler(LogMan::DebugLevels Level, const char* Message) {
 }
 
 static void AssertHandler(const char* Message) {
-  const auto Output = fextl::fmt::format("[ASSERT] {}\n", Message);
+  const auto Output = fextl::fmt::format("[A] {}\n", Message);
   if (WineDbgOut) {
     WineDbgOut(Output.c_str());
   } else if (LogFile) {
--- a/Source/Windows/Common/Logging.cpp
+++ b/Source/Windows/Common/Logging.cpp
@@ -14,7 +14,7 @@ void (*WineDbgOut)(const char* Message);
 FILE* LogFile;
 
 static void MsgHandler(LogMan::DebugLevels Level, const char* Message) {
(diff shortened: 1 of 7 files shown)


bylaws:
Could you do this for windows too? Though can drop the colours in that case

neobrain:
> Could you do this for windows too? Though can drop the colours in that case

Which changes are left that would apply? The single-letter change translated to Windows since it's implemented in `LogMan::DebugLevelStr`. The only difference I can see is the use of `[]`, but not sure we want to drop that when colors aren't used on Windows.

bylaws:
Was just thinking the dropping of [] for separation, I think the lack of colours would be fine? Majority of the case where I've found logs useful for debugging on either are when you've piped them to a file anyway, which would lack on Linux anyhow

Source/Windows/Common/Logging.cpp
```diff
@@ -23,7 +23,7 @@ static void MsgHandler(LogMan::DebugLevels Level, const char* Message) {
 }
 
 static void AssertHandler(const char* Message) {
-  const auto Output = fextl::fmt::format("[ASSERT] {}\n", Message);
+  const auto Output = fextl::fmt::format("[A] {}\n", Message);
```
neobrain: TestHarnessRunner.cpp also needs this.

neobrain:
> Was just thinking the dropping of [] for separation, I think the lack of colours would be fine? Majority of the case where I've found logs useful for debugging on either are when you've piped them to a file anyway, which would lack on Linux anyhow

Fair enough. Applied the removal of `[]` to the other modules as well.

Sonicadvance1 (approved):

PR #4617, opened 2025-06-17
FEXConfig: Infer folders for 32-bit library wrappers automatically

### LINKED ISSUE #4588: FEXConfig doesn't expose 32-bit folders for library forwarding
These use separate config entries, but FEXConfig only sets the 64-bit ones.

neobrain:
There's no need to bother the user to select these paths manually. Just bake in the assumption that it's the same folders with `_32` appended.

Fixes #4588.


neobrain:
While writing the PR description, I realized removing the second set of options entirely is the better way forward. #4618 should be pursued instead.

neobrain:
Obsoleted by #4618.

PR #4619, opened 2025-06-17
Remove code generation build step for install prefix

neobrain:
Having a dedicated build step just for the install prefix is unnecessarily involved. This information can easily be retrieved otherwise, which also has the upside of not triggering a source rebuild when changing the install prefix.

Furthermore, this cuts down the number of files FEX uses that start with "Config" from 10 to 8 and the number of top-level folders from 16 to 15. Baby steps.

Some functional changes to FEXBash are documented via PR comments.

### DIFF
--- a/Source/Tools/FEXBash/FEXBash.cpp
+++ b/Source/Tools/FEXBash/FEXBash.cpp
@@ -6,7 +6,6 @@ desc: Launches bash under FEX and passes arguments via -c to it
 $end_info$
 */
 
-#include "ConfigDefines.h"
 #include "Common/ArgumentLoader.h"
 
 #include <FEXCore/Config/Config.h>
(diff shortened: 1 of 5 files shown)


Source/Tools/FEXBash/FEXBash.cpp
```diff
@@ -25,11 +24,21 @@ int main(int argc, char** argv, char** const envp) {
   // Use /bin/sh for -c commands and /bin/bash for interactive mode
   const char* BashPath = Args.empty() ? "/bin/bash" : "/bin/sh";
 
-  std::string FEXInterpreterPath = std::filesystem::path(argv[0]).parent_path().string() + "FEXInterpreter";
+  std::string FEXInterpreterPath = std::filesystem::path(argv[0]).parent_path().string() + "/FEXInterpreter";
```
neobrain, line 27: This seemingly was broken before, since `parent_path` does not include a trailing `/`. Instead, paths like `/usr/binFEXInterpreter` were checked.

Source/Tools/FEXBash/FEXBash.cpp
```diff
@@ -25,11 +24,21 @@ int main(int argc, char** argv, char** const envp) {
...
+
   // Check if a local FEXInterpreter to FEXBash exists
   // If it does then it takes priority over the installed one
   if (!std::filesystem::exists(FEXInterpreterPath)) {
-    FEXInterpreterPath = FEXCore::Config::FindContainerPrefix() + FEXINTERPRETER_PATH;
```
neobrain, line 32: I couldn't find a scenario where the container prefix needed to be applied, and in fact opening a shell in pressure vessel indicated that the path `/run/host/usr/bin/FEXInterpreter` doesn't actually exist.

PR #4622, opened 2025-06-17
FEXConfig: Add icon

neobrain:
We don't have any desktop menu entry, but the X11/Wayland logo placeholders previously still showed up in the title bar, task bar, and task switcher.

~~Marking as draft since I'm not sure if we have an SVG version of this file anywhere. If not, the PNG logo should at least be centered properly.~~ We don't have an SVG version of the logo, so I took the one from https://wiki.fex-emu.com/index.php/Multimedia:Brand_Assets and centered it.


neobrain:
Centered the logo, good to go.

Sonicadvance1 (approved):
Cool.

PR #4624, opened 2025-06-20
Some static code analysis fixes

Sonicadvance1:
NFC, just clearing out some warnings.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -793,10 +793,8 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
       }
     }
 
-    if (DebugData) {
-      DebugData->Subblocks.push_back({static_cast<uint32_t>(BlockStartHostCode - CodeData.BlockEntry),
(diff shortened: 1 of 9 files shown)


FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -438,14 +438,14 @@ void Arm64JITCore::Op_Unhandled(const IR::IROp_Header* IROp, IR::Ref Node) {
 
 static void DirectBlockDelinker(FEXCore::Core::CpuStateFrame* Frame, FEXCore::Context::ExitFunctionLinkData* Record) {
   // Emit new 16 bytes of code to a temporary patch, then atomically apply it
-  __uint128_t Patch;
-  ARMEmitter::Emitter emit((uint8_t*)&Patch, sizeof(Patch));
+  auto Patch = reinterpret_cast<__uint128_t*>(alloca(sizeof(__uint128_t)));
+  ARMEmitter::Emitter emit((uint8_t*)Patch, sizeof(*Patch));
```
neobrain: This seems overly silly a dance. Surely the static analyzer has a way to just silence the warning so that we don't have to go out of our way to write a simple variable? (You didn't mention which tool you're using, so I can't check myself)

Sonicadvance1: Removed the offending change.

PR #4625, opened 2025-06-20
ARM64EC: Sind Wine 10.10 there're some fixed syscall numbers

AndreRH:


Sonicadvance1:
I'll leave this to @bylaws to review. I don't want to accidentally break older Wine versions.

bylaws:
I don't think hardcoding this when we want older wine compat is ideal, it would be better to lookup the syscall thunk address for each and then order by that (this is valid on both windows and all wine versions). (I appreciate this is somewhat annoying with the constraints at this specific initialization stage, can do myself if you want)

AndreRH:
I'm fine with u doing it. You already digged deeper into the topic

PR #4627, opened 2025-06-24
Optimize long division

alyssarosenzweig:
post-RA

PR #4629, opened 2025-06-25
Optimize some constant cpuid/xgetbv cases

alyssarosenzweig:
On top of #4627 

This is as good as we can do with the merge paradigm. Can't deal with syscalls the way we handle things right now :/ TBD what we do longer term, but this is the easy win we can get with minimal JIT time impact.


Sonicadvance1:
Needs a rebase.

alyssarosenzweig:
rebased

Sonicadvance1 (approved):
Nice nice, thanks for fixing this.

Issue #4630, opened 2025-06-27
Library Forwarding: Find a more reliable way to deal with wl_interface globals


neobrain:
libQt6WaylandClient.so ships its own set of `wl_interface`s, which interferes with our code at: https://github.com/FEX-Emu/FEX/blob/a5fad89e57127e91ca3f6b7d6b020e65f7d7fb5f/ThunkLibs/libwayland-client/Guest.cpp#L10-L27

This can be reproduced simply by starting [Cutter](https://cutter.re/) in FEX, which will hang during startup. Using `WAYLAND_DEBUG=1`, it will print the error `interface '(null)' has no event 0`.

A reliable workaround is to set `LD_PRELOAD=libwayland-client.so`.

PR #4631, opened 2025-06-27
LibraryForwarding: Fix various Wayland issues

neobrain:
This gets [Cutter](https://cutter.re/) working with host libwayland-client:

Screenshot_20250626_223204

Due to https://github.com/FEX-Emu/FEX/issues/4630, `LD_PRELOAD=libwayland-client.so` must be set for now.


neobrain:
Also observed these while having Steam open in the background:
```
/usr/bin/zenity: symbol lookup error: /lib/x86_64-linux-gnu/libgtk-4.so.1: undefined symbol: wl_data_device_interface
Failed to load shared library 'libgdk-3.so.0' referenced by the typelib: /lib/x86_64-linux-gnu/libgdk-3.so.0: undefined symbol: wl_data_device_interface
```

Sonicadvance1 (approved):

PR #4632, opened 2025-06-27
Build: Add nix-based helpers to facilitate cross-compilation

neobrain:
FEX has various subprojects that require cross-compilation with different compilers:
* FEXLinuxTests, requiring 32-bit gcc and 64-bit gcc targeting x86
* Library forwarding, requiring clang targeting x86 (both 32-bit and 64-bit) with a full C++ standard library
* Wine-on-Arm, requiring custom mingw-w64 toolchains

Having to set up a custom toolchain for Wine-on-Arm is merely a minor annoyance. However for the first two points we found that most Linux distributions besides Debian/Ubuntu simply lack the required cross-compilers to reliably enable these features. The situation has improved over time (see #3597 and #4230), however fundamentally the process is still very involved.

**Nix provides a general purpose solution that runs on any distribution.** This PR hence sets up a few convenience scripts to take the problem of setting up toolchains entirely off the user's shoulders.

For advanced uses, Nix also has a feature called `nix-shell` that allows loading up a bash shell with the toolchains directly available in the PATH. This is mainly useful to build VKD3D and vkd3d-proton.

## FEXLinuxTests

In an existing build folder, run `cmake_enable_flt.sh`:
```sh
cd build
../Data/nix/cmake_enable_flt.sh
ninja
```

## Library Forwarding

In an existing build folder, run `cmake_enable_libfwd.sh`:
```sh
cd build
../Data/nix/cmake_enable_libfwd.sh
ninja
```

## Wine-on-Arm

Create a fresh build folder and run `cmake_configure_woa*.sh` for the 64-bit build, then repeat for the 32-bit one:
```sh
mkdir -p build/arm64ec && cd build/arm64ec
../../Data/nix/cmake_configure_woa64.sh ../../ -G Ninja
ninja
mkdir ../wow64 && cd ../wow64
../../Data/nix/cmake_configure_woa32.sh ../../ -G Ninja
ninja
```

## Building VKD3D or vkd3d-proton

Open a `nix-shell` and follow the upstream instructions for the manual build:

```sh
nix-shell Data/nix/WineOnArm/shell.nix
meson setup $FEX_MESON_CROSSFILE
```
(After configuration, you can exit the nix-shell again)

### DIFF
--- a/Data/nix/FEXLinuxTests/shell.nix
+++ b/Data/nix/FEXLinuxTests/shell.nix
@@ -1,0 +1,35 @@
+{ pkgs ? import <nixpkgs> { } }:
+
+let
+  pkgsCross32 = pkgs.pkgsCross.gnu32;
+  pkgsCross64 = pkgs.pkgsCross.gnu64;
+
+  gcc32 = pkgs.writeText "toolchain_nix_gcc_x86_32.txt" ''
+    set(CMAKE_SYSTEM_PROCESSOR i686)
+    set(CMAKE_C_COMPILER ${pkgsCross32.buildPackages.gcc}/bin/i686-unknown-linux-gnu-gcc)
+    set(CMAKE_CXX_COMPILER ${pkgsCross32.buildPackages.gcc}/bin/i686-unknown-linux-gnu-g++)
+  '';
+
+  gcc64 = pkgs.writeText "toolchain_nix_gcc_x86_64.txt" ''
+    set(CMAKE_SYSTEM_PROCESSOR x86_64)
+    set(CMAKE_C_COMPILER ${pkgsCross64.buildPackages.gcc}/bin/x86_64-unknown-linux-gnu-gcc)
+    set(CMAKE_CXX_COMPILER ${pkgsCross64.buildPackages.gcc}/bin/x86_64-unknown-linux-gnu-g++)
+  '';
+in
+pkgs.mkShell {
+  buildInputs = [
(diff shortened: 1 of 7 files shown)


bylaws:
I would just use wine instead on wineonarm tbh, I think it's best not overload the already existing WoA acronym

neobrain:
> I would just use wine instead on wineonarm tbh, I think it's best not overload the already existing WoA acronym

Does a better umbrella term exist? Just titling it Wine carries ambiguity since "x86 Wine in FEX" continues to be a supported use case with a very different technical architecture. I've certainly tripped over this when I wasn't yet familiar with the topic, though historically there was also another configuration (Hangover) that complicated things further.

Regarding overloading WoA, is there any real risk of confusion there? After all, "FEX in Wine-on-Arm" is the same deployment scenario as "XTA in Windows-on-Arm". So even if you abbreviate Wine-on-Arm to WoA, it can easily be inferred what you're referring to (so long as it happens in a general context of x86 emulation).

neobrain:
Latest changes fix issues brought up by Alyssa during testing (thanks!): Added Wayland headers to the dev rootfs, compile with `-lstdc++`, change `CLANG_EXEC_PATH` CMake variable only if it's not already set.

This is confirmed working on Asahi Linux's muvm setup as well now :)

alyssarosenzweig:
thunking & arm64ec scripts both tested (including dxvk build)

Sonicadvance1 (approved):
Sure.

PR #4634, opened 2025-07-02
EmulatedFiles: Emulate `current_clocksource`

Sonicadvance1:
WINE uses this to determine TSC frequency and because it doesn't say `tsc` on ARM devices, it was ignoring TSC and instead using CPU maximum frequency.

This was causing Horizon to think the TSC ran at whatever the max frequency of a core was  (1.8Ghz to 2.6Ghz depending?) This was causing all of Horizon Zero Dawn's physics to run at slower than real time speeds because our 1Ghz (on Orion) TSC is significantly lower than the max clock speeds of the cores.

This is still a bug in Wine that it is using the maximum CPU clock speed in the case of current_clocksource not being TSC, but that's a battle for a different time.


Sonicadvance1:
Rebased.

PR #4637, opened 2025-07-03
RAPass: Partial revert 7e39eb3d

Sonicadvance1:
This change causes Hades to hang on startup inside of a futex forever. Currently unknown why but it is a regression from 2506, so disable it for now while investigations occur.


alyssarosenzweig:
Replaced with #4640

PR #4642, opened 2025-07-04
Implement x87 invalid operation bit on F80 mode

pmatos:


FEXCore/Source/Interface/Core/OpcodeDispatcher/X87F64.cpp
```diff
@@ -107,6 +107,63 @@ void OpDispatchBuilder::FISTF64(OpcodeArgs, bool Truncate) {
   const auto Size = OpSizeFromSrc(Op);
 
   Ref data = _ReadStackValue(0);
+
+  // Need to check for overflow/NaN and set Invalid Operation flag
+  // The _Float_ToGPR_S/ZS operations don't set x87 exception flags in F64 mode
+  // So we need to do it manually here
+
+  // Extract the float value as GPR to check for special values
+  Ref DataGPR = _VExtractToGPR(OpSize::i64Bit, OpSize::i64Bit, data, 0);
```
bylaws: This is gonna be very slow, I *think* https://github.com/FEX-Emu/FEX/blame/fa0a54deb901cf1245357c2d834d9eefeb2083d7/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp#L2103 should match in behaviour though.

bylaws (changes requested):
I would be quite afraid of enabling this by default for f64 as is - seems like a ton of extra overhead if there's little depending on it and those that do could probably get deal with f80?.

This could likely be a lot better off in the f64 case using the IOC FPSR flag and somehow keeping track of when that is clobbered in order to flush it at appropriate times - from a brief check it seems to match the x86 behaviour here and would come for free. It would require basically redoing the bulk of your changes though which isn't the best, even if the result would probably be simpler.

pmatos:
> I would be quite afraid of enabling this by default for f64 as is - seems like a ton of extra overhead if there's little depending on it and those that do could probably get deal with f80?.
> 
> This could likely be a lot better off in the f64 case using the IOC FPSR flag and somehow keeping track of when that is clobbered in order to flush it at appropriate times - from a brief check it seems to match the x86 behaviour here and would come for free. It would require basically redoing the bulk of your changes though which isn't the best, even if the result would probably be simpler.
> 
> Depends what others think, but might make sense to get the f80 side of this split-up and merged and the rest later? Since that's a lot simpler and 'free'

Yeah, I agree with you. I will split this into F80 and F64. I was not aware of the IOC FPSR flag. Will take a look after the split and rework the F64 part. Thanks for the quick turnaround.

Sonicadvance1:
It should be noted that FP exceptions are optional in ARM spec so not all CPU cores implement it. I believe Apple Silicon and Oryon support it, but Cortex doesn't? We do have a check if the feature is supported that is currently unused although.

bylaws:
Even if exceptions aren't implemented, IOC should still be set I think? That would be the main thing here

Sonicadvance1:
Good shout, I didn't check if the exception is disabled (or forced to zero by unsupported), if the sticky bit changes or not.

pmatos:
I have updated this PR to focus on F80 (non reduced precision).

I will add a further PR based on this one to add the work for F64.

FEXCore/Source/Interface/Core/OpcodeDispatcher/X87F64.cpp
pmatos: This was removed from current version.

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/F80Fallbacks.h
```diff
@@ -36,6 +36,14 @@ FEXCORE_PRESERVE_ALL_ATTR static softfloat_state SoftFloatStateFromFCW(uint16_t
   return State;
 }
 
+// Helper function to propagate SoftFloat exceptions to X87 flags
+FEXCORE_PRESERVE_ALL_ATTR static void HandleX87Exception(const softfloat_state& State, FEXCore::Core::CpuStateFrame* Frame) {
```
bylaws, line 39: Thoughts on making an RAII object with this and SoftFloatStateFromFCW? Would allow e.g.:

ScopedSoftFloatState State{FCW, Frame}
return X80SoftFloat(src).ToI32(&State.State);

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -737,12 +752,22 @@ void X87StackOptimization::Run(IREmitter* Emit) {
       }
 
       case OP_F80SINSTACK: {
-        HandleUnop(OP_F64SIN, false, OP_F80SIN);
+        Ref St0 = LoadStackValue();
```
bylaws: Could the below 5 cases be handled simpler in the interpreter fallback itself? Or does this allow constprop to remove checks

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/InterpreterFallbacks.cpp
```diff
@@ -279,6 +293,44 @@ bool InterpreterOps::GetFallbackHandler(const IR::IROp_Header* IROp, FallbackInf
   case IR::OP_VPCMPESTRX: *Info = {FABI_I32_I64_I64_V128_V128_I16, Core::OPINDEX_VPCMPESTRX}; return true;
   case IR::OP_VPCMPISTRX: *Info = {FABI_I32_V128_V128_I16, Core::OPINDEX_VPCMPISTRX}; return true;
 
+  // Float to GPR conversion fallbacks
+  case IR::OP_FLOAT_TOGPR_S: {
```
bylaws: I think you can drop all of these? They can be solved on the dispatcher side for f64 as CVT2SSI etc does rn, and f80 uses _F80CVTInt

bylaws (changes requested):

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/F80Fallbacks.h
pmatos: Thanks - wouldn't have thought of that. I implemented as suggested. I think it encapsulates well the x87 exception handling.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: Tbh, I am not sure what you mean here. SIN and COS are similar but others are just slightly different.

bylaws: Rather than on the IR level, they could be done in c++ in the code for the corresponding fallback

pmatos: Ah yes, but then the IR will lack specificity which aids debugging quite a bit. Unless there's a real reason to move these away from the pass, I would prefer to keep them here.

pmatos: edit: removed a previous comment because I misunderstood your message.

Ah, you mean the infinity check part, right? That's a good point. I will try to analyze the impact of moving this to the fallbacks.

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/InterpreterFallbacks.cpp
pmatos: Thanks.

FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
pmatos: I think I addressed your concerns and things are looking much better. Of course, a lot of the inflation of instcountci is now hidden because the checks run in the fallbacks, but it looks much simpler so that seems like a plus. Thanks for the time helping out on this with your reviews.

pmatos:
ping @bylaws

bylaws (approved):

PR #4644, opened 2025-07-06
Fix: wrong magic value in fpstate.

ChanthMiao:
FEX should only set fpx_sw_bytes.magic1 with FP_XSTATE_MAGIC when avx is enabled. Otherwise it may cause segfault in ntdll::save_context, which requires access to extended xstate info if magic1 equals FP_XSTATE_MAGIC.


Sonicadvance1 (approved):
Good catch, missed this since we typically don't disable AVX for 64-bit processes.

PR #4647, opened 2025-07-08
OpcodeDispatcher: Fix several cases where incorrectly sized loads/stores could be used

bylaws:


Sonicadvance1 (approved):
unittest update incoming to show the `vcvtsd2si` fix.

PR #4650, opened 2025-07-09
Add --changed flag to reformat.sh script

pmatos:
The flag and the optional directory helps target the reformatting changes. In any case, this should be backwards compatible. Running it without flags should do what it did before.

### DIFF
--- a/Scripts/reformat.sh
+++ b/Scripts/reformat.sh
@@ -3,12 +3,48 @@
 # Save current directory
 DIR=$(pwd)
 
+# Get the absolute path to the Scripts directory
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+# Parse arguments
+CHANGED_ONLY=false
+TARGET_DIR=""
+
+while [[ $# -gt 0 ]]; do
+    case $1 in
+        --changed) # Only reformat changed files (staged and unstaged)
+            CHANGED_ONLY=true
+            shift
+            ;;
+        *)
+            TARGET_DIR=$1
+            shift
+            ;;
+    esac
+done
+
 # Change to the directory passed as argument if any
-if [ $# -eq 1 ]; then
-    cd $1
+if [ -n "$TARGET_DIR" ]; then
+    cd "$TARGET_DIR"
 fi
 
-# Reformat whole tree.
-# This is run by the reformat target.
-git ls-files -z '*.cpp' '*.h' '*.inl' ':(exclude)External/*' | xargs -0 -n 1 -P $(nproc) clang-format-19 -i
+# Reformat files
+if [ "$CHANGED_ONLY" = true ]; then
+    # Check for unstaged deletions
+    if git ls-files -d | head -1 | grep -q .; then
+        echo "Error: Unstaged deletions detected. Please stage or discard deletions before formatting."
+        exit 1
+    fi
+
+    CHANGED_FILES=$(git ls-files -m '*.cpp' '*.h' '*.inl')
+    if [ -n "$CHANGED_FILES" ]; then
+        echo "$CHANGED_FILES" | xargs -d '\n' -n 1 -P $(nproc) python3 "$SCRIPT_DIR/clang-format.py" -i
+    else
+        echo "No changed files to format."
+    fi
+else
+    # Reformat whole tree (original behavior)
+    git ls-files -z '*.cpp' '*.h' '*.inl' | xargs -0 -n 1 -P $(nproc) python3 "$SCRIPT_DIR/clang-format.py" -i
+fi
+
 cd $DIR


Scripts/reformat.sh
```diff
@@ -3,12 +3,43 @@
...
 # Change to the directory passed as argument if any
-if [ $# -eq 1 ]; then
-    cd $1
+if [ -n "$TARGET_DIR" ]; then
+    cd "$TARGET_DIR"
+fi
+
+# Reformat files
+if [ "$CHANGED_ONLY" = true ]; then
+    # Only reformat changed files (staged and unstaged)
```
neobrain: Could you move the comment to the `--changed` line above? Since that's where people would typically look to find out what the option does.

Scripts/reformat.sh
```diff
@@ -3,12 +3,43 @@
...
+    CHANGED_FILES=$(git diff --name-only HEAD --diff-filter=AM | grep -E '\.(cpp|h|inl)$')
```
neobrain: Is there a difference between this and the simpler `git ls-files -m '*.cpp' '*.h' '*.inl'`?

Scripts/reformat.sh
```diff
@@ -3,12 +3,43 @@
...
+    if [ -n "$CHANGED_FILES" ]; then
+        echo "$CHANGED_FILES" | tr '\n' '\0' | xargs -0 -n 1 -P $(nproc) python3 "$SCRIPT_DIR/clang-format.py" -i
```
neobrain: Isn't `tr '\n' '\0' | xargs -0` equivalent to `xargs`? Or if the previous suggestion is adopted, using `ls-files -mz` should allow for `xargs -0` without the `tr`.

Scripts/reformat.sh
pmatos: TIL `-m`.

Scripts/reformat.sh
pmatos: Given `-m` is a thing, `-mz` makes a lot more sense. I was just using the `xargs -0` that we already had. Thanks for the great suggestions.

Scripts/reformat.sh
pmatos: Actually, I think we might want to use `git diff`. `git ls-files -m` includes deleted files which we don't want to format.

pmatos:
@neobrain wondering if you have further suggestions. After some tries I ended up getting something that ended up with `\0` in `CHANGED_FILES` and bash would complain with :
```
Scripts/reformat.sh: line 33: warning: command substitution: ignored null byte in input
```

on testing. So I decided to just use newlines instead of the null byte. Afaiu this only breaks if newlines exist in filenames that's not going to happen - hopefully. Do you see any further issues with the current approach?

Scripts/reformat.sh
neobrain: > Actually, I think we might want to use git diff. git ls-files -m includes deleted files which we don't want to format.

After playing around with this in the terminal, I just realized it's a bit more subtle: It only includes *unstaged deletion*, whereas `git rm File.cpp` will remove the file from the `ls-files` output as well. This is analogous to *creating* files, which is an action that also won't be reflected in `ls-files` until the file change is added to the git index.

It seems unlikely someone would remove a file without staging that change while also wanting to run a reformat immediately after, so personally I'd probably keep using `ls-files` to avoid diverging the code branches in this script too much.

pmatos: > > Actually, I think we might want to use git diff. git ls-files -m includes deleted files which we don't want to format.
> 
> After playing around with this in the terminal, I just realized it's a bit more subtle: It only includes _unstaged deletion_, whereas `git rm File.cpp` will remove the file from the `ls-files` output as well. This is analogous to _creating_ files, which is an action that also won't be reflected in `ls-files` until the file change is added to the git index.
> 
> It seems unlikely someone would remove a file without staging that change while also wanting to run a reformat immediately after, so personally I'd probably keep using `ls-files` to avoid diverging the code branches in this script too much.

The problem is if I `rm file.cpp` and then run the script we'll call `clang-format` with a file that doesn't exist. So, if we use `ls-files` then lets at least check there are no unstaged deletions beforehand. Let me check how to achieve that.

neobrain: I understand the problem you're describing, but when are you realistically going to do delete a file that's under version control *without* using `git rm` and then additionally run the reformat script? It's possible on paper but seems like something that's not actually going to happen in practice. (Even if it did, perhaps erroring due to the removed file that's not in the git index is a good thing.)

pmatos: I am happy to use `ls-files` but we should still check if there are files that are unstaged deletions. Isn't it better than an error from clang-format because the file does not exist? I regularly remove files with `rm` instead of `git rm` and stage them later for example.

neobrain (approved):
Seems benign enough either way. Since it's useful to you, let's merge it as-is.

neobrain:
> on testing. So I decided to just use newlines instead of the null byte. Afaiu this only breaks if newlines exist in filenames that's not going to happen - hopefully. Do you see any further issues with the current approach?

Not sure why we went with `-z` initially to begin with, but I don't see a problem with using `\n` as the item separator instead.

neobrain:
Needs a rebase due to #4662.

pmatos:
Thanks for the review, rebased.

PR #4651, opened 2025-07-09
Upgrade to clang-format-19

### LINKED ISSUE #4577: Upgrade clang-format to something other than 16
Debian and Ubuntu in their infinite wisdom struck down clang-16 while keeping the other versions from 14 onwards. We just so happened to be using that version of clang for clang-format.

```
$ apt-cache search clang-format
clang-format-20 - Tool to format C/C++/Obj-C code
clang-format - Tool to format C/C++/Obj-C code
clang-format-14 - Tool to format C/C++/Obj-C code
clang-format-15 - Tool to format C/C++/Obj-C code
clang-format-17 - Tool to format C/C++/Obj-C code
clang-format-18 - Tool to format C/C++/Obj-C code
clang-format-19 - Tool to format C/C++/Obj-C code
```

We should upgrade to clang-format-18.

pmatos:
Fixes #4577


pmatos:
@Sonicadvance1 this might require installing clang-format-19 on the self-hosted x64 runners.

neobrain:
Does the upgrade not require a reformat for you, or is that planned for a followup PR?

Sonicadvance1:
I have updated the runner to have clang-format-19, but we have a step `check for correct clang-format version` which checks if `clang-format` is the correct version instead of `clang-format-19`.
I assume if I update the symlink then hell will break loose, so can these be explicitly versioned?

pmatos:
> I have updated the runner to have clang-format-19, but we have a step `check for correct clang-format version` which checks if `clang-format` is the correct version instead of `clang-format-19`. I assume if I update the symlink then hell will break loose, so can these be explicitly versioned?

Yes! I will use clang-format-19 everywhere.

pmatos:
> Does the upgrade not require a reformat for you, or is that planned for a followup PR?

Let me ensure this works and then I will reformat in a followup PR.

pmatos:
@Sonicadvance1 once you merge this, I will have a PR ready for a whole-tree reformat.

PR #4661, opened 2025-07-11
Whole-tree reformat with clang-format-19

pmatos:


pmatos:
@Sonicadvance1 here it is.

neobrain:
The changes to lambda indentation are unfortunate - does `LambdaBodyIndentation: Signature` produce a less drastic diff by any chance?

CodeEmitter/CodeEmitter/ALUOps.inl
```diff
@@ -185,7 +186,8 @@ public:
 
   void ands(ARMEmitter::Size s, ARMEmitter::Register rd, ARMEmitter::Register rn, uint64_t Imm) {
     uint32_t n, immr, imms;
-    [[maybe_unused]] const auto IsImm = IsImmLogical(Imm, RegSizeInBits(s), &n, &imms, &immr);
+    [[maybe_unused]]
+    const auto IsImm = IsImmLogical(Imm, RegSizeInBits(s), &n, &imms, &immr);
```
neobrain: The line breaks here add a lot of noise, so I'm tempted to just switch to `BreakAfterAttributes: Leave` at the expense of not enforcing breaks in function declarations anymore.

What do you think?

pmatos: Yeah, I will try to see what works to avoid that. Another thing I just noticed is that we are reformatting stuff in `External/`. Does it make sense? My gut feeling is that we shouldn't be doing it.

pmatos: I just checked and we didn't do it for the initial reformatting so lets skip it again, unless you disagree.

pmatos:
Been trying to undo the issue with attribute being broken but playing around with the penalties means making something else worse. Do you have any further comments or suggestions?

CodeEmitter/CodeEmitter/ALUOps.inl
neobrain: > Another thing I just noticed is that we are reformatting stuff in External/.
> I just checked and we didn't do it for the initial reformatting so lets skip it again, unless you disagree.

Sounds good.

> Been trying to undo the issue with attribute being broken but playing around with the penalties means making something else worse. Do you have any further comments or suggestions?

Does `BreakAfterAttributes: Leave` have undesired side-effects too or are you just experimenting to see if something better can be found?

Sonicadvance1:
From a quick scan, nothing here seems terrible. Fine with me.

pmatos:
> From a quick scan, nothing here seems terrible. Fine with me.

OK - rebased. Once it's ok just merge, otherwise I will have to keep updating the sha in `.git-blame-ignore-revs`.

neobrain (changes requested):
Can we try `BreakAfterAttributes: Leave` and `LambdaBodyIndentation: Signature` before merging this to avoid switching formatting back and forth unnecessarily? I suspect that would significantly cut down the number of modified lines.

pmatos:
> Can we try `BreakAfterAttributes: Leave` and `LambdaBodyIndentation: Signature` before merging this to avoid switching formatting back and forth unnecessarily? I suspect that would significantly cut down the number of modified lines.

Argh, I reformatted and forgot those because I resetted the changed .clang-format. Sorry. Added the `.clang-format` change as a commit together wtih the reformat and so on. What do you think?

neobrain:
> Argh, I reformatted and forgot those because I resetted the changed .clang-format. Sorry. Added the .clang-format change as a commit together wtih the reformat and so on. What do you think?

No worries! Going from a 1250 loc diff to 250 sounds like a solid win to me.

Not sure why clang 16 missed so many lambda indentation cases before, but personally I like `LambdaBodyIndentation: Signature` better anyway so if it's closer to existing code I'll gladly take it. The other changes look in line with what I'd expect from a clang-format update.

Thanks for taking care of this; looking forward to finally re-enabling auto-formatting in my IDE :)

neobrain:
CI is failing with
```
clang-format: Unknown command line argument '-list-ignored'.  Try: 'clang-format --help'
```
Any idea what's happening there?

pmatos:
> > Argh, I reformatted and forgot those because I resetted the changed .clang-format. Sorry. Added the .clang-format change as a commit together wtih the reformat and so on. What do you think?
> 
> No worries! Going from a 1250 loc diff to 250 sounds like a solid win to me.
> 
> Not sure why clang 16 missed so many lambda indentation cases before, but personally I like `LambdaBodyIndentation: Signature` better anyway so if it's closer to existing code I'll gladly take it. The other changes look in line with what I'd expect from a clang-format update.
> 
> Thanks for taking care of this; looking forward to finally re-enabling auto-formatting in my IDE :)


yeah, I also prefer Signature. Unsure why we decided initially to use Outer.

pmatos:
> CI is failing with
> 
> ```
> clang-format: Unknown command line argument '-list-ignored'.  Try: 'clang-format --help'
> ```
> 
> Any idea what's happening there?

Looking into it.

pmatos:
> CI is failing with
> 
> ```
> clang-format: Unknown command line argument '-list-ignored'.  Try: 'clang-format --help'
> ```
> 
> Any idea what's happening there?

I assumed git-clang-format-19 would always call clang-format-10 but doesn't. it calls the system clang-format which in the CI runners is clang-format-16 which don't support `-list-ignored`. We need to define the clang-format to call in `clangformat.binary` git config. I will create a PR for this.

pmatos:
See https://github.com/FEX-Emu/FEX/pull/4666

Sonicadvance1:
With that PR merged, this should be trivial once rebased?

pmatos:
> With that PR merged, this should be trivial once rebased?

Good to go, I think.

PR #4663, opened 2025-07-11
External/code-format-helper: Update requirements

Sonicadvance1:
Latest of everything, let's see what happens.
To get rid of dependabot alerts.


pmatos:
Uff, I forgot this was a thing. thanks for sorting it out.

Sonicadvance1:
Added a couple more versions that weren't covered by my initial bump, that way we can hammer all the dependabot alerts.

neobrain (approved):

PR #4666, opened 2025-07-14
Set clang-format-19 as the version git-clang-format should run in wor…

pmatos:
…kflow


pmatos:
Alternatively we would need to specify `--binary` when calling `git-clang-format`.

pmatos:
I pushed the alternative of setting the `--binary` argument instead. I think it's better because it will work for us locally as well.

Sonicadvance1 (approved):

PR #4670, opened 2025-07-14
Implement call-ret stack optimisations

bylaws:
The current approach for handling function calls and returns with multiblock incurs a fairly significant amount of overhead, both due to putting the pressure of every return on the L1 cache and relying on a hard to predict indirect branch after the lookup. 

This is solved by introducing a shadow stack of <rip, hostcode> pairs, this is pushed to on calls, with hostcode pointing to the instruction after the bl/blr. On returns, the top pair on the shadow stack is popped and if the target rip matches the l1 lookup is skipped and a `ret xN` instruction used to return to hostcode, balancing the CPU return address stack. 

Geekbench comparison: https://browser.geekbench.com/v6/cpu/compare/12837458?baseline=12837357

Clang benefits about 10% from using the host return address stack itself, with the remaining benefit being from the decreased L1 cache load.

See the commit message for  [JIT: Rewrite block linking to support direct ExitFunction calls](https://github.com/FEX-Emu/FEX/commit/e3e18e0b06abace21c03652729a8b93cad600542) for a detailed explanation of that block linking changes.

TODO: linux support


Sonicadvance1:
I haven't yet taken a look at this yet, but does this ensure there are protection pages on the ends of the shadow stack so badly behaving applications don't break the emulation?

bylaws:
> I haven't yet taken a look at this yet, but does this ensure there are protection pages on the ends of the shadow stack so badly behaving applications don't break the emulation?

Yeah, that's left up to frontend side code to handle.

bylaws:
Are the vixl simulator failures here something obvious? I'm very unfamiliar with it

Sonicadvance1:
> Are the vixl simulator failures here something obvious? I'm very unfamiliar with it

This is a fairly dense change so I'll need to review to see what's going on. Generally when vixl breaks it is due to accidentally jumping from Arm64 JIT code over to host code on accident.

Sonicadvance1:
Actually looks like the vixl simulator has some logs about this that aren't showing up in the github logs. Not sure what this means yet.
```
GCS mismatch: lr = 0x00007ffff7fb40f0, gcs = 0x00007fffe18001a8
GCS records, most recent first:
 gcs0[4]: 0x00007fffe18001a8
 gcs0[3]: 0x00007fffe1800128
 gcs0[2]: 0x00007fffe1800114
 gcs0[1]: 0x00007fffe1800098
 gcs0[0]: 0x00007fffe1800020
End of GCS records.
GCS failed in /mnt/Work/Work/work/FEXNew/External/vixl/src/aarch64/simulator-aarch64.h, line 5684

Program received signal SIGABRT, Aborted.
```

Sonicadvance1:
ooooh, It's checking for `FEAT_GCS` things. Which means that this'll explode on GCS hardware.

Sonicadvance1:
With the two PRs I opened, this will fix your unittests.

Sonicadvance1:
Looks like there is some race on process shutdown with this at least, doesn't repro on main. Running a 32-bit vulkaninfo showcasing this, 64-bit seemed fine.
```bash
Thread 1 "FEXInterpreter" received signal SIGSEGV, Segmentation fault.
0x0000aaaaab0a78bc in std::__atomic_impl::store<unsigned int> (__ptr=0x3646136d0, __t=335544328, __m=std::memory_order::relaxed) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/atomic_base.h:1093
1093            __atomic_store(__ptr, __atomic_impl::__clear_padding(__t), int(__m));
(gdb) bt
#0  0x0000aaaaab0a78bc in std::__atomic_impl::store<unsigned int> (__ptr=0x3646136d0, __t=335544328, __m=std::memory_order::relaxed) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/atomic_base.h:1093
#1  std::__atomic_ref<unsigned int, true, false>::store (this=0xffffffffaf90, __t=335544328, __m=std::memory_order::relaxed) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/atomic_base.h:1640
#2  0x0000aaaaab0a7b10 in FEXCore::CPU::IndirectBlockDelinker (Frame=0xfffff6ebf0a0, Record=0x3646136e0) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/JIT/JIT.cpp:530
#3  0x0000aaaaaaf34a00 in FEXCore::GuestToHostMap::Erase (this=0xfffff6fed100, Frame=0xfffff6ebf0a0, Address=3884503008) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/LookupCache.h:84
#4  0x0000aaaaaaf17218 in FEXCore::LookupCache::Erase (this=0xfffff6e1c230, Frame=0xfffff6ebf0a0, Address=3884503008) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/LookupCache.h:195
#5  0x0000aaaaaaf170d4 in FEXCore::Context::ContextImpl::ThreadRemoveCodeEntry (Thread=0xfffff6ebf000, GuestRIP=3884503008) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/Core.cpp:962
#6  0x0000aaaaaaf16f70 in FEXCore::Context::InvalidateGuestThreadCodeRange (Thread=0xfffff6ebf000, Start=3881828352, Length=14671872) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/Core.cpp:926
#7  0x0000aaaaaaf16e3c in FEXCore::Context::ContextImpl::InvalidateGuestCodeRange (this=0xfffff6e4b000, Thread=0xfffff6ebf000, Start=3881828352, Length=14671872) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/Core.cpp:938
#8  0x0000aaaaab23459c in FEX::HLE::ThreadManager::InvalidateGuestCodeRange (this=0xfffff6e4b618, CallingThread=0xfffff6ebf000, Start=3881828352, Length=14671872) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.h:202
#9  0x0000aaaaab232310 in FEX::HLE::SyscallHandler::InvalidateCodeRangeIfNecessary (this=0xfffff6e4b600, Thread=0xfffff6ebf000, Base=3881828352, Length=14671872) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h:267
#10 0x0000aaaaab2ae9d0 in FEX::HLE::SyscallHandler::GuestMunmap (this=0xfffff6e4b600, Is64Bit=false, Thread=0xfffff6ebf000, addr=0xe7600000, length=14669276) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp:241
#11 0x0000aaaaab2c4104 in FEX::HLE::x32::x32SyscallHandler::GuestMunmap (this=0xfffff6e4b600, Thread=0xfffff6ebf000, addr=0xe7600000, length=14669276) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Syscalls.h:52
#12 0x0000aaaaab2d2af4 in FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_2::operator()(FEXCore::Core::CpuStateFrame*, void*, unsigned long) const (this=0xffffffffb3e7, Frame=0xfffff6ebf0a0, addr=0xe7600000, length=14669276) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp:46
#13 0x0000aaaaab2d2aa4 in FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_2::__invoke(FEXCore::Core::CpuStateFrame*, void*, unsigned long) (Frame=0xfffff6ebf0a0, addr=0xe7600000, length=14669276) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Memory.cpp:46
#14 0x0000aaaaab293080 in std::__invoke_impl<unsigned long, unsigned long (*&)(FEXCore::Core::CpuStateFrame*, unsigned long, unsigned long), FEXCore::Core::CpuStateFrame*&, unsigned long&, unsigned long&> (__f=@0xfffff6e78890: 0xaaaaab2d2a78 <FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_2::__invoke(FEXCore::Core::CpuStateFrame*, void*, unsigned long)>, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276)
    at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:61
#15 0x0000aaaaab293030 in std::__invoke<unsigned long (*&)(FEXCore::Core::CpuStateFrame*, unsigned long, unsigned long), FEXCore::Core::CpuStateFrame*&, unsigned long&, unsigned long&> (__fn=@0xfffff6e78890: 0xaaaaab2d2a78 <FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_2::__invoke(FEXCore::Core::CpuStateFrame*, void*, unsigned long)>, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276)
    at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/invoke.h:96
#16 0x0000aaaaab29035c in std::invoke<unsigned long (*&)(FEXCore::Core::CpuStateFrame*, unsigned long, unsigned long), FEXCore::Core::CpuStateFrame*&, unsigned long&, unsigned long&> (__fn=@0xfffff6e78890: 0xaaaaab2d2a78 <FEX::HLE::x32::RegisterMemory(FEX::HLE::SyscallHandler*)::$_2::__invoke(FEXCore::Core::CpuStateFrame*, void*, unsigned long)>, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276, __args=@0xffffffffb540: 14669276)
    at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/functional:120
#17 0x0000aaaaab28a3f8 in FEX::HLE::SyscallHandler::HandleSyscall (this=0xfffff6e4b600, Frame=0xfffff6ebf0a0, Args=0xffffffffb530) at /mnt/Work/Work/work/FEXNew/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp:831
#18 0x00000004a7a9773c in ?? ()
Backtrace stopped: previous frame inner to this frame (corrupt stack?)
```
[vulkaninfo_32.tar.gz](https://github.com/user-attachments/files/21304422/vulkaninfo_32.tar.gz)

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
```diff
@@ -73,13 +73,13 @@ namespace x64 {
...
-  constexpr std::array<ARMEmitter::Register, 8> RA = {
+  constexpr std::array<ARMEmitter::Register, 7> RA = {
     // All these callee saved
     ARMEmitter::Reg::r20, ARMEmitter::Reg::r21, ARMEmitter::Reg::r22, ARMEmitter::Reg::r23,
-    ARMEmitter::Reg::r24, ARMEmitter::Reg::r25, ARMEmitter::Reg::r30, ARMEmitter::Reg::r18,
+    ARMEmitter::Reg::r24, ARMEmitter::Reg::r30, ARMEmitter::Reg::r18,
   };
 
-  constexpr unsigned RAPairs = 6;
+  constexpr unsigned RAPairs = 4;
```
Sonicadvance1, line 82: Wait a minute, was this RAPairs thing even correct? Walking it through to where it is used in RAPass it's always wanting the the registers to be sequential pairs? And the count is wrong already? RA Bug? @alyssarosenzweig

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
```diff
@@ -87,7 +87,7 @@ void FileManager::LoadThunkDatabase(fextl::unordered_map<fextl::string, ThunkDBO
       };
 
       // We only need to generate 32-bit or 64-bit depending on the operating mode.
-      const auto ArchPrefix = Is64BitMode() ? "lib64" : "lib";
+      const auto ArchPrefix = Is64BitMode() ? "lib" : "lib32";
```
Sonicadvance1: This change seems unrelated and will heavily depend on differences between Fedora and Arch.

unittests/InstructionCountCI/Primary.json
```diff
@@ -2804,50 +2804,46 @@
...
     },
     "db 0x48, 0xa1; dq 0x00000000e0000008": {
-      "ExpectedInstructionCount": 3,
+      "ExpectedInstructionCount": 2,
       "Comment": [
         "mov rax, [0xe0000008]",
         "0xa1"
       ],
       "ExpectedArm64ASM": [
-        "mov w20, #0x400",
```
Sonicadvance1: Wot?

Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
bylaws: Ah oops I didn't intend for this to sneak in

FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
alyssarosenzweig: The current code is right, just written in the most confusing way I could have possible done it.

RAPairs is "the number of GPR registers that are paired in the array", in other words, "GPRs[0:RAPairs] is a list of `(RAPairs / 2)` pairs of registers". Then in the RA, we only consider registers `GPR[0], GPR[2], GPR[4], ..., GPR[(RAPairs - 1) / 2]` when allocating the first register for the pair.

unittests/InstructionCountCI/Primary.json
Sonicadvance1: #4694 for this

Sonicadvance1:
Dogfooding at this point.

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -58,31 +58,105 @@ DEF_OP(ExitFunction) {
...
       add(ARMEmitter::Size::i64Bit, ARMEmitter::Reg::rsp, StaticRegisters[X86State::REG_RSP], 0);
       LoadConstant(ARMEmitter::Size::i64Bit, EC_CALL_CHECKER_PC_REG, NewRIP);
       ldr(TMP2, STATE_PTR(CpuStateFrame, Pointers.Common.ExitFunctionEC));
       br(TMP2);
     } else {
 #endif
-      // Align to 16 byte to allow atomic patching of the following 16 byte
-      // of code (excluding the RIP data) on platforms that support LSE2
-      Align16B();
+      // This code will be backpatched by Arm64JITCore_ExitFunctionLink, below is an enumeration of all the possible cases.
```
neobrain, line 79: IIUC you're listing all possible variations of the block that will ever be stored in memory at runtime here. Can you clarify which of these we're generating *right here* and which ones result from backpatching? Probably better to reflect this distinction in the structure of the block comment, i.e. first explain what's generated here and then explain what results from backpatching.

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -58,31 +58,105 @@ DEF_OP(ExitFunction) {
...
+      // 00: b 0x8
+      // 04: br TMP1
+      // 08: ldr TMP1, <Shared exit linker>
+      // 0c: blr TMP1
+      // 10: HostCode
+      // 18: GuestRIP
+      // 20: CallerOffset
+      //
+      //
+      // Call with known return block - linked in range
```
neobrain: I'm guessing this refers to the 128MB (?) range limit for near jumps? 

```suggestion
      // Call with known return block - linked in 128 MiB range
```

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -58,31 +58,105 @@ DEF_OP(ExitFunction) {
...
+      // 04: br TMP1
+      // 08: ldr TMP1, <Shared exit linker>
+      // 0c: blr TMP1
+      // 10: HostCode
+      // 18: GuestRIP
+      // 20: CallerOffset
+      //
+      //
+      // Jump - linked in range
+      // 00: b HostCode                                         - MODIFIED
```
neobrain: Some indentation for these code blocks wouldn't hurt, and avoids the need for the out-of-place double-newlines:

```suggestion
      // Jump (unlinked)
      //     00: b JmpThunk00
      //     JmpThunk00:
      //     00: b 0x8
      //     04: br TMP1
      //     08: ldr TMP1, <Shared exit linker>
      //     0c: blr TMP1
      //     10: HostCode
      //     18: GuestRIP
      //     20: CallerOffset
      //
      // Jump (linked in range)
      //     00: b HostCode                                         - MODIFIED
```

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -495,17 +495,42 @@ void Arm64JITCore::Op_Unhandled(const IR::IROp_Header* IROp, IR::Ref Node) {
...
+}
+
+static void DirectBlockCallDelinker(FEXCore::Core::CpuStateFrame* Frame, FEXCore::Context::ExitFunctionLinkData* Record) {
+  DirectBlockDelinker(Frame, Record, true);
+}
+
+static void DirectBlockJumpDelinker(FEXCore::Core::CpuStateFrame* Frame, FEXCore::Context::ExitFunctionLinkData* Record) {
+  DirectBlockDelinker(Frame, Record, false);
+}
+
```
neobrain: What's the purpose of these functions? Unless there's concrete near-term plans to change these to something else, this seems to hinder code clarity more than it helps.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -526,27 +551,50 @@ static uint64_t Arm64JITCore_ExitFunctionLink(FEXCore::Core::CpuStateFrame* Fram
     return Frame->Pointers.Common.DispatcherLoopTop;
   }
 
-  uintptr_t branch = (uintptr_t)(Record)-8;
-  LOGMAN_THROW_A_FMT((branch % 16) == 0, "Incorrect alignment for block linking record");
+  // See BranchOps.cpp, ExitFunction for an assembly level view of the handled cases.
```
neobrain:
```suggestion
  // See ExitFunction in BranchOps.cpp for an assembly level view of the handled cases.
```

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -831,22 +881,34 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
...
       if (PendingTargetLabel && PendingTargetLabel != &IsTarget->second) {
         b(PendingTargetLabel);
+        PendingTargetLabel = nullptr;
       }
 
       if (BlockIROp->EntryPoint) {
         uint64_t BlockStartRIP = Entry + BlockIROp->GuestEntryOffset;
-        if (EmittedEntry) {
+
+        const auto IsReturnTarget = CallReturnTargets.try_emplace(Node).first;
```
neobrain, line 900: Why use `try_emplace` over `emplace` here?

Source/Windows/WOW64/Module.cpp
```diff
@@ -93,6 +95,12 @@ struct TLS {
   FEXCore::Core::InternalThreadState*& ThreadState() const {
     return reinterpret_cast<FEXCore::Core::InternalThreadState*&>(TEB->TlsSlots[FEXCore::ToUnderlying(Slot::THREAD_STATE)]);
   }
+
+  // This is used to work around user callback handling (see Wow64KiUserCallbackDispatcher in wine) unbalancing the stack
```
neobrain: What does "unbalancing the stack" mean?

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -58,31 +58,105 @@ DEF_OP(ExitFunction) {
...
-      // Align to 16 byte to allow atomic patching of the following 16 byte
-      // of code (excluding the RIP data) on platforms that support LSE2
-      Align16B();
+      // This code will be backpatched by Arm64JITCore_ExitFunctionLink, below is an enumeration of all the possible cases.
+      //
+      // Call with known return block - unlinked
+      // 00: adr TMP1, 0xC
+      // 04: stp RetReg, TMP1, [SpReg, -0x10]!
+      // 08: bl JmpThunk00
+      // JmpThunk00:
```
neobrain: Where do these jump thunks get emitted? Should be mentioned in the comment since it doesn't seem to be here.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -154,6 +154,8 @@ void ThreadManager::StatAlloc::UnlockAfterFork(FEXCore::Core::InternalThreadStat
   ThreadObject->Thread->ThreadStats = AllocateSlot(ThreadObject->ThreadInfo.TID);
 }
 
+static constexpr size_t CALLRET_STACK_ALLOC_SIZE = FEXCore::Core::InternalThreadState::CALLRET_STACK_SIZE + 2 * FEXCore::Utils::FEX_PAGE_SIZE;
```
neobrain: `static` is implied for global constexpr objects.

neobrain (commented):
Could you add some more general high-level documentation on how this whole setup works now?

The commit message in e3e18e0b06abace21c03652729a8b93cad600542 is solid information already, but it describes the relative change to the previous system, so it's not immediately helpful without that context. I also like the notes in BranchOps.cpp, so either that file or JIT.cpp could be good places for this.

We didn't have much high-level information before, but the code was (barely) simple enough to figure out. This change is pushing the complexity above a threshold were documentation is needed to understand what's going on without spending hours chasing down all the different modules involved.

Source/Windows/WOW64/Module.cpp
bylaws: Pushing onto the call ret stack without a pop, causing a guard page hit and signal

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
bylaws: Oh - JIT.cpp

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: Oh, use a lambda instead?

neobrain: Ah, apologies - I thought these were just identity wrappers around DirectBlockDelinker, but they actually set up the last parameter. Could use a lambda indeed, or template DirectBlockDelinker on the boolean parameter since it's such a short function anyway.

FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
bylaws: Added a note on which is which, think keeping the ordering is better though as it helps to have everything for each case together

FEXCore/Source/Interface/Core/JIT/JIT.cpp
bylaws: I have no particular attachment to either, feels like try_emplace is the more modern thing to use for this case from some brief reading

bylaws:
Pushed some additional changes to call compileblock directly from the exit function linker. This probably warrants a rough retest

Sonicadvance1:
Looks like the Peggle Deluxe regression with Proton hasn't been resolved yet.

bylaws:
> Looks like the Peggle Deluxe regression with Proton hasn't been resolved yet.

See discord for this

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -253,7 +254,6 @@ class ContextImpl final : public FEXCore::Context::Context, CPU::CodeBufferManag
   template<auto Fn>
   static uint64_t ThreadExitFunctionLink(FEXCore::Core::CpuStateFrame* Frame, ExitFunctionLinkData* Record) {
     auto Thread = Frame->Thread;
```
Sonicadvance1: Unused now, should be removed.

Sonicadvance1 (dismissed):
Some more testing happening with the recent changes.

bylaws:
> Could you add some more general high-level documentation on how this whole setup works now?

Added some more to BranchOps. Should be good for more testing now

Sonicadvance1:
No issues so far with this.

Sonicadvance1:
Potentially hunting a mutex deadlock.

Sonicadvance1:
```
(gdb) bt
#0  futex_wait (futex_word=0xffffab075140, expected=2, private=0) at ../sysdeps/nptl/futex-internal.h:146
#1  __GI___lll_lock_wait (futex=futex@entry=0xffffab075140, private=private@entry=0) at ./nptl/lowlevellock.c:49
#2  0x0000ffffabd89558 [PAC] in lll_mutex_lock_optimized (mutex=0xffffab075140) at ./nptl/pthread_mutex_lock.c:48
#3  ___pthread_mutex_lock (mutex=0xffffab075140) at ./nptl/pthread_mutex_lock.c:128
#4  0x0000aaaaafdc3f88 [PAC] in __gthread_mutex_lock (__mutex=0xffffab075140) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/aarch64-linux-gnu/c++/14/bits/gthr-default.h:762
#5  __gthread_recursive_mutex_lock (__mutex=0xffffab075140) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/aarch64-linux-gnu/c++/14/bits/gthr-default.h:824
#6  std::recursive_mutex::lock (this=0xffffab075140) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/mutex:125
#7  std::lock_guard<std::recursive_mutex>::lock_guard (__m=..., this=<optimized out>) at /usr/lib/gcc/aarch64-linux-gnu/14/../../../../include/c++/14/bits/std_mutex.h:250
#8  FEXCore::GuestToHostMap::AcquireLock (this=0xffffab075140) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/LookupCache.h:28
#9  FEXCore::LookupCache::FindBlock (this=0xffffab07a4c0, Address=140737313412688) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/LookupCache.h:138
#10 0x0000aaaaafe6f204 in FEXCore::CPU::Arm64JITCore::ExitFunctionLink (Frame=0xffffab08b0a0, Record=0xaaaab581b088) at /mnt/Work/Work/work/FEXNew/FEXCore/Source/Interface/Core/JIT/JIT.cpp:539
#11 0x0000aaaaafda51e8 in ?? ()
```
Looks like we have a hang here depending on situation.
```
(gdb) info thread
  Id   Target Id                                            Frame
* 1    Thread 0x7fffffaf8cc0 (LWP 489168) "pressure-vessel" futex_wait (futex_word=0xffffab075140, expected=2, private=0) at ../sysdeps/nptl/futex-internal.h:146
(gdb) p WriteLock
$2 = {<std::__recursive_mutex_base> = {_M_mutex = {__data = {__lock = 2, __count = 2, __owner = 489164, __nusers = 1, __kind = 1, __spins = 0, __list = {__prev = 0x0, __next = 0x0}}, __size = "\002\000\000\000\002\000\000\000\314v\a\000\001\000\000\000\001", '\000' <repeats 30 times>, __align = 8589934594}}, <No data fields>}
```
Looks like it is locked already by a different TID (489164) which is a sibling TID to this process but through what I assume to be a fork.
So that thread acquired the mutex while this thread was forking and then couldn't continue.
This mutex looks like it needs to be a fork safe mutex so it can be reset safely after fork semantics occur.

Sonicadvance1:
Reproduced by just running PV over and over until it hung.
```
SteamLinuxRuntime_sniper$ for i in {0..9999} ; do echo $i; FEXInterpreter ./run -- /usr/bin/true ; done 
```

bylaws:
Would adding back the wrapping codeinvalidation mutex lock fix this? I'm not super familiar with this, but it seems reasonable to do that if it would

Sonicadvance1:
> Would adding back the wrapping codeinvalidation mutex lock fix this? I'm not super familiar with this, but it seems reasonable to do that if it would

Maybe? With the addition of the new mutexes, would need to ensure that there isn't a case where they deadlock each other.

Sonicadvance1:
```diff
diff --git a/FEXCore/Source/Interface/Core/JIT/JIT.cpp b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
index e0e4438bd..3e204be54 100644
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -527,6 +527,7 @@ static void IndirectBlockDelinker(FEXCore::Core::CpuStateFrame* Frame, FEXCore::

 uint64_t Arm64JITCore::ExitFunctionLink(FEXCore::Core::CpuStateFrame* Frame, FEXCore::Context::ExitFunctionLinkData* Record) {
   auto Thread = Frame->Thread;
+  auto lk_inval = GuardSignalDeferringSection<std::shared_lock>(static_cast<Context::ContextImpl*>(Thread->CTX)->CodeInvalidationMutex, Thread);
   bool TFSet = Thread->CurrentFrame->State.flags[X86State::RFLAG_TF_RAW_LOC];
   uintptr_t HostCode {};
   auto GuestRip = Record->GuestRIP;
   ```
   
   With this diff I've been running for a couple of minutes without a hang. Will let it run for a while and see if it is fine.
   Not sure about the performance characteristics since ExitFunctionLink seems to be doing more work now?

Sonicadvance1:
This patch seems to have resolved the fork hang. I'll leave it to you for speculating about shared_lock performance problems here.
I'll leave my loop running for a while but it has already gotten to 300x3 iterations which is significantly higher than the dozen or so it usually takes.

Sonicadvance1:
700x3 iterations later, I think with the diff applied it is safe.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
neobrain: Fair. It mainly caught my attention because we don't use it much anywhere else.

When copying smaller data into the container, `insert` is preferable over both `emplace` and `try_emplace` since the latter two can produce horrendous error messages during refactors (due to being templates). That's a minor concern though of course.

bylaws:
What you suggested would deadlock for nested compiles I think? I pushed something that should avoid that case and still work

neobrain:
> Added some more to BranchOps. Should be good for more testing now

Just got around reading it fully - much clearer how it all connects now, thanks for taking the time to write it up!

PR #4672, opened 2025-07-15
code-format-helper: Fix formatting for the format helper

neobrain:


Sonicadvance1 (approved):
We format so we can format and format forever more.

pmatos:
Thanks. :+1:

PR #4676, opened 2025-07-15
Revert to using clang-format.py helper

pmatos:
Unfortunately clang-format-19 doesn't support globstar or recursive patterns in .clang-format.ignore.

We need those currently so we revert some changes in previous commits.

### DIFF
--- a/Scripts/clang-format.py
+++ b/Scripts/clang-format.py
@@ -1,0 +1,116 @@
+#!/usr/bin/env python3
+# Imported from LLVM
+# It's basically a clang-format wrapper with .clang-format-ignore support.
+# Can be removed once we adopt Clang19, which supports this out of the box.
+#
+# This is called by git-clang-format and it works in two modes.
+# It has a path in the command line to format, and it will output the result post format to stdout.
+# Or it has -assume-filename=<path> in the command line, and the file is passed via stdin.
+# Post format output will be given in stdout.
+
+import subprocess
+import sys
+import os
+import re
+import fnmatch
+
+# Wrapper globals
+project_root = os.path.join(os.path.dirname(os.path.realpath(__file__)), "..")
+ignore_file_path = os.path.join(project_root, ".clang-format-ignore")
+clang_format_command = os.getenv("CLANG_FORMAT") or "clang-format-19"
+
+
+def glob_to_regex(pattern):
+    # Normalize directory separators
+    pattern = pattern.replace("\\", "/")
+    return fnmatch.translate(pattern)
+
+
+def load_ignore_patterns(ignore_file_path):
+    # Check if the file exists
+    if not os.path.exists(ignore_file_path):
+        raise FileNotFoundError(f"No such file: '{ignore_file_path}'")
+
+    with open(ignore_file_path, "r") as file:
+        lines = file.readlines()
+
+    patterns = []
+    for line in lines:
+        line = line.strip()
+        if line and not line.startswith("#"):  # Ignore empty lines and comments
+            pattern = glob_to_regex(line)
+            patterns.append(re.compile(pattern))
+
+    # Print the number of patterns found
+    print(f"Number of patterns found: {len(patterns)}", file=sys.stderr)
+
+    return patterns
+
+
+def normalize_path(file_path):
+    absolute_path = os.path.abspath(file_path)
+    normalized_path = absolute_path.replace("\\", "/")
+    return normalized_path
+
+
+def should_ignore(file_path, ignore_patterns):
+    normalized_path = normalize_path(file_path)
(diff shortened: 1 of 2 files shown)


Scripts/clang-format.py
```diff
@@ -0,0 +1,116 @@
+#!/usr/bin/env python3
+# Imported from LLVM
+# It's basically a clang-format wrapper with .clang-format-ignore support.
+# Can be removed once we adopt Clang19, which supports this out of the box.
```
neobrain, line 4: Needs a refresh since we're waiting for clang-20 now.

neobrain:
I'm fine with this PR, but as an alternative would it be feasible to just use manual glob-ing in our `.clang-format-ignore` file?

I.e.
```
# Ignore all files in the External directory                                                                                                                                                                            
External/*
External/*/*
External/*/*/*
```

It's a little annoying, but the patterns are simple enough that we probably won't need a ton of them.

pmatos:
> I'm fine with this PR, but as an alternative would it be feasible to just use manual glob-ing in our `.clang-format-ignore` file?
> 
> I.e.
> 
> ```
> # Ignore all files in the External directory                                                                                                                                                                            
> External/*
> External/*/*
> External/*/*/*
> ```
> 
> It's a little annoying, but the patterns are simple enough that we probably won't need a ton of them.

That _might_ work but what about these patterns?
```
Source/Tools/LinuxEmulation/LinuxSyscalls/x*/SyscallsNames.inl
```

The way that upstream llvm had in `-19` suggested the above is to add a `.clang-format` to `External/` with `DisableFormat: true`.  That's feasible I think but we are still missing a way to handle the above pattern.

Context: https://github.com/llvm/llvm-project/issues/114969

neobrain:
> That _might_ work but what about these patterns?
> 
> ```
> Source/Tools/LinuxEmulation/LinuxSyscalls/x*/SyscallsNames.inl
> ```

This specific example only covers two files, so we can list them explicitly. For the Ioctl line we could use `Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/*`+`Source/Tools/LinuxEmulation/LinuxSyscalls/x64/Ioctl/*` instead.

This is all assuming that clang-19 at least understands wildcards at the *end* of a path.

EDIT: [clang-19 documentation](https://releases.llvm.org/19.1.0/tools/clang/docs/ClangFormat.html) says this is supported indeed:
> To match all files in a directory, use e.g. foo/bar/*.

pmatos:
> 
> > To match all files in a directory, use e.g. foo/bar/*.

This doesn't work recursively thought but if I add more patterns to `.clang-format-ignore`, it should work.

pmatos:
> > That _might_ work but what about these patterns?
> > ```
> > Source/Tools/LinuxEmulation/LinuxSyscalls/x*/SyscallsNames.inl
> > ```
> 
> This specific example only covers two files, so we can list them explicitly. For the Ioctl line we could use `Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/*`+`Source/Tools/LinuxEmulation/LinuxSyscalls/x64/Ioctl/*` instead.
> 

That would include the *.h files that we currently format. We could instead just list all the *.inl files there.

pmatos:
At this point, I think that #4685 is preferable to landing this PR. @neobrain ?

neobrain:
Sorry for not asking earlier, but how did you check that `Source/Tools/LinuxEmulation/LinuxSyscalls/x*/Ioctl/*.inl` doesn't work? When I run `clang-format-19 -i Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/radeon_drm.inl`, it does skip the formatting unless I comment out the quoted line from `.clang-format-ignore`.

I think the only change in clang-20 is support for `**`, for which I like the elegant solution you've found in #4685 by using `DisableFormat: true` in a separate `.clang-format` file. Seems like that's all we need?

pmatos:
> Sorry for not asking earlier, but how did you check that `Source/Tools/LinuxEmulation/LinuxSyscalls/x*/Ioctl/*.inl` doesn't work? When I run `clang-format-19 -i Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Ioctl/radeon_drm.inl`, it does skip the formatting unless I comment out the quoted line from `.clang-format-ignore`.
> 
> I think the only change in clang-20 is support for `**`, for which I like the elegant solution you've found in #4685 by using `DisableFormat: true` in a separate `.clang-format` file. Seems like that's all we need?

*sigh* sorry, I made a mistake. The mistake was changing the file so that it requires formatting, running clang-format -i and checking if the file was modified. The file was indeed modified because I modified it, not because clang-format did. :(

I am closing this one. Take a look at the changes in #4685

PR #4681, opened 2025-07-15
FEXCore/OpcodeDispatcher: Removes tuple usage for Dispatch tables

Sonicadvance1:
NFC

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.h
@@ -80,6 +80,12 @@ struct LoadSourceOptions {
   bool AllowUpperGarbage = false;
 };
 
+struct DispatchTableEntry {
+  uint16_t Op;
+  uint8_t Count;
+  X86Tables::OpDispatchPtr Ptr;
+};
+
 class OpDispatchBuilder final : public IREmitter {
   friend class FEXCore::IR::Pass;
   friend class FEXCore::IR::PassManager;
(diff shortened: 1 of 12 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -80,6 +80,18 @@ struct LoadSourceOptions {
   bool AllowUpperGarbage = false;
 };
 
+struct DispatchTable_U8_U8 {
+  uint8_t Op;
+  uint8_t Count;
+  X86Tables::OpDispatchPtr Ptr;
+};
+
+struct DispatchTable_U16_U8 {
```
neobrain: These two structs have the same size due to padding data, so is there a benefit over using the u16 variant everywhere?

Sonicadvance1: Sure, converted.

FEXCore/Source/Interface/Core/OpcodeDispatcher.h
```diff
@@ -80,6 +80,12 @@ struct LoadSourceOptions {
   bool AllowUpperGarbage = false;
 };
 
+struct DispatchTable_U16_U8 {
```
neobrain: This no longer needs a cryptic `U16_U8` suffix. Probably should be called DispatchTableEntry though, rather than `DispatchTable`.

Sonicadvance1: Removed the cryptography extension usage.

PR #4682, opened 2025-07-15
Windows: Support enabling hardware TSO on Asahi Linux

bylaws:
Relies on a [https://github.com/bylaws/wine/commit/d41dcbe25dcc707ffd3a110993e480053c5c8a1a](wine-side) patch to expose the prctl to the PE-side


Sonicadvance1 (approved):
Nice

PR #4685, opened 2025-07-16
Ensure .clang-format-ignore is compatible with clang-format-19

pmatos:
Alternative to #4676 
@neobrain This might work instead of the alternative PR of bringing back the clang-format.py script.


pmatos:
> Glad we found such a simple way to make it work in the end :)
> 
> Commit message and PR description contain outdated information since this only needs to take care of recursively ignoring Externals now, otherwise looks good!

Updated. Thanks.

PR #2811, opened 2023-07-24
Arm64: Implements support for CSSC popcount

Sonicadvance1:
Moves singular flags usage over to it

PR #4692, opened 2025-07-17
FEXCore: Reintroduce support for CSSC

Sonicadvance1:
Now that the PF flag isn't using popcount, this is a win across the board if the hardware supports it.

Been a while since I last looked at this, added a new instcountci file to show the improvement.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/ALUOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/ALUOps.cpp
@@ -1038,34 +1038,55 @@ DEF_OP(Popcount) {
   const auto Dst = GetReg(Node);
   const auto Src = GetReg(Op->Src);
 
-  switch (OpSize) {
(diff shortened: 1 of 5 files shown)


alyssarosenzweig:
remind me, why was this a loss when PF used it...?

Sonicadvance1:
> remind me, why was this a loss when PF used it...?

https://github.com/FEX-Emu/FEX/pull/2811#issuecomment-1648663331
For 8-bit operations it unintentionally introduced some additional masking which was unnecessary due to how vector popcount works on arm. Instead of spending effort working around it, I just closed it instead. Since other stuff was in flux at the time, it was easier to keep it unchanged.

This cost is no longer eaten on non-CSSC platforms with the new behaviour about allowing upper garbage or not.

alyssarosenzweig:
Ahh, gotchu. Cool cats.

lioncash (approved):

PR #4699, opened 2025-07-22
Inline SMC fixes and support for WOW64

bylaws:
Fixes peggle deluxe (with an additional wine patch) under wow64. Also moves slightly closer to linux inline SMC support

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
@@ -84,7 +84,8 @@ constexpr uint64_t EC_CODE_BITMAP_MAX_ADDRESS = 1ULL << 47;
 #endif
 
 // Will force one single instruction block to be generated first if set when entering the JIT filling SRA.
-constexpr auto ENTRY_FILL_SRA_SINGLE_INST_REG = TMP1;
+// FillStaticRegs must preserve this
+constexpr auto ENTRY_FILL_SRA_SINGLE_INST_REG = TMP2;
 
 // Predicate to use in the X87 SVE optimization
 constexpr ARMEmitter::PRegister PRED_X87_SVEOPT = ARMEmitter::PReg::p2;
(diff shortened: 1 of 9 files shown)


FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
```diff
@@ -84,7 +84,8 @@ constexpr uint64_t EC_CODE_BITMAP_MAX_ADDRESS = 1ULL << 47;
 #endif
 
 // Will force one single instruction block to be generated first if set when entering the JIT filling SRA.
-constexpr auto ENTRY_FILL_SRA_SINGLE_INST_REG = TMP1;
+// FillStaticRegs must preserve this
+constexpr auto ENTRY_FILL_SRA_SINGLE_INST_REG = TMP2;
```
Sonicadvance1, line 88: https://github.com/FEX-Emu/FEX/blob/fe61aabc523ee5f48ead554ad14c37544f0223ba/Source/Windows/ARM64EC/Module.S#L52
The register here also needs to be changed right?

bylaws: Oh good catch

Sonicadvance1:
While I confirmed this fixes peggle deluxe with wow64, it's basically unplayable for some reason? Looks like a thread is hammering cache clears and memcpy? Maybe single stepping got enabled?

Image_2025-07-22_17-29-35

bylaws:
Ah yeah, I have an xorg issue on my t14s that's causing opengl programs to lag severely so I had assumed it was that. Will see about it

bylaws:
I checked this, not a fex bug. You likely dont have placed mapping patched into your system mesa install (even if you did , this is currently zink only). This wastes years memcpying the to/from uncached memory.

alyssarosenzweig:
> I checked this, not a fex bug. You likely dont have placed mapping patched into your system mesa install (even if you did , this is currently zink only). This wastes years memcpying the to/from uncached memory.

What's this about?

bylaws:
Did some wider testing with denuvo games, thankfully this doesn't impact perf significantly

bylaws:
> > I checked this, not a fex bug. You likely dont have placed mapping patched into your system mesa install (even if you did , this is currently zink only). This wastes years memcpying the to/from uncached memory.
> 
> What's this about?

See here:
https://list.winehq.org/mailman3/hyperkitty/list/wine-devel@winehq.org/thread/4WM25AUL7VE4ZQLAYPX3SR2L246UXWWT/#BTDG4PCWJ3PQUUFO2HUS6O2KF2C6HXCC

Note none of this is upstream, yet. I think there are conflicting views on what to actually do.

PR #4704, opened 2025-07-23
LookupCache: Move CodePages to GuestToHostMap

bylaws:
Prevents invalidations being missed under the following circumstances: Thread A JITs block A into the global codebuffer, adding the guest to host mapping to its CodePages, thread A is then killed. Thread B then performs SMC on block A. An exception will be triggered but as CodePages was stored per-thread, and thread A is now killed when all threads are iterated over by the frontend to perform invalidations it will be missed.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -909,26 +909,32 @@ uintptr_t ContextImpl::CompileSingleStep(FEXCore::Core::CpuStateFrame* Frame, ui
   return (uintptr_t)CodePtr;
 }
 
-static void InvalidateGuestThreadCodeRange(FEXCore::Core::InternalThreadState* Thread, uint64_t Start, uint64_t Length) {
+static void InvalidateGuestThreadCodeRange(FEXCore::Core::InternalThreadState* Thread, InvalidatedEntryAccumulator& Accumulator,
+                                           uint64_t Start, uint64_t Length) {
   // Ensures now-modified mappings aren't cached as being in their previous non-executable state.
   // Accessing FrontendDecoder is safe as the thread's code invalidation mutex must be locked here.
   Thread->FrontendDecoder->ResetExecutableRangeCache();
 
   auto lk = Thread->LookupCache->AcquireLock();
+  auto& CodePages = Thread->LookupCache->Shared->CodePages;
 
-  auto lower = Thread->LookupCache->CodePages.lower_bound(Start >> 12);
-  auto upper = Thread->LookupCache->CodePages.upper_bound((Start + Length - 1) >> 12);
+  auto lower = CodePages.lower_bound(Start >> 12);
+  auto upper = CodePages.upper_bound((Start + Length - 1) >> 12);
 
   for (auto it = lower; it != upper; it++) {
-    for (auto Address : it->second) {
-      ContextImpl::ThreadRemoveCodeEntry(Thread, Address);
+    Accumulator.emplace_back(std::move(it->second));
+  }
+
+  for (const auto& PageEntries : Accumulator) {
+    for (const auto& Entry : PageEntries) {
+      ContextImpl::ThreadRemoveCodeEntry(Thread, Entry);
     }
-    it->second.clear();
   }
 }
 
-void ContextImpl::InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState* Thread, uint64_t Start, uint64_t Length) {
-  InvalidateGuestThreadCodeRange(Thread, Start, Length);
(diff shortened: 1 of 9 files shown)


FEXCore/include/FEXCore/Core/Context.h
```diff
@@ -63,6 +63,7 @@ class AOTIRWriter {
 };
 
 using CodeRangeInvalidationFn = std::function<void(uint64_t start, uint64_t Length)>;
+using InvalidatedEntryAccumulator = fextl::vector<fextl::vector<uint64_t>>;
```
neobrain, line 68: We're juggling around too many of these "collection of collection of numbers"-types, which make it difficult to tell e.g. whether something is a guest address or a JIT block address. Some more comments about the contained elements are needed, though it's also worth turning some of these into structs (since named members serve as self-documentation at use sites).

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -737,6 +737,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
     .TotalInstructionsLength = TotalInstructionsLength,
     .StartAddr = Thread->FrontendDecoder->DecodedMinAddress,
     .Length = Thread->FrontendDecoder->DecodedMaxAddress - Thread->FrontendDecoder->DecodedMinAddress,
+    .NeedsAddGuestCodeRanges = !HasCustomIR,
```
neobrain, line 740: Why does custom IR need to be considered here?

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -1230,7 +1231,7 @@ void Decoder::DecodeInstructionsAtEntry(const uint8_t* _InstStream, uint64_t PC,
 
   uint64_t CurrentCodePage = PC & FEXCore::Utils::FEX_PAGE_MASK;
 
-  fextl::set<uint64_t> CodePages = {CurrentCodePage};
+  BlockInfo.CodePages.insert(CurrentCodePage);
```
neobrain: Can't we just keep using `{CurrentCodePage}` here? That keeps code more local by establishing the full object state, whereas the current version requires scrolling up to see that CurrentCodePage will indeed be the only contained object.

FEXCore/Source/Interface/Core/Frontend.h
```diff
@@ -38,6 +38,8 @@ class Decoder final {
   struct DecodedBlockInformation final {
     uint64_t TotalInstructionCount;
     fextl::vector<DecodedBlocks> Blocks;
+    fextl::set<uint64_t> EntryPoints;
+    fextl::set<uint64_t> CodePages;
```
neobrain: Kinda unrelated to this PR, but could you add a tiny comment like this here so we have explained it *somewhere*?

```suggestion
    fextl::set<uint64_t> EntryPoints
    fextl::set<uint64_t> CodePages; // Page start addresses for code reachable through any entry point
```

FEXCore/Source/Interface/Core/Core.cpp
bylaws: Custom IR handlers have no guest backing and skip the frontend decoder, and shouldn't be removed on smc. This matches the existing behaviour

neobrain: Is this hit in practice though? It's currently only used for library forwarding, where SMC doesn't happen. But perhaps your upcoming patches are different.

If it doesn't happen in practice I'd opt for ignoring this detail to keep the interfaces simpler instead (so we don't need to pass a boolean through three layers).

(Arguably this existing behavior is the wrong behavior anyway - why would SMC *not* also invalidate any custom IR handlers?)

bylaws: I think regardless we'll need to pass something out to indicate this case, if we do want SMC to invalidate handlers we'll need to avoid accessing the stale frontend variables and manually add the block entry as a code page instead.

And no, I don't think this is hit in practice, more just trying to make the edge case have somewhat reasonable behaviour.

neobrain: Sounds like it's safer to outright assert that any SMC-related invalidation does not cover custom IR handlers... This might be an opportunity for a broader cleanup though, so let's defer this to another time.

bylaws: I will note- SMC invalidation won't remove the handler's so that's fine - but unmapping also won't - the function to remove handlers isn't ever called (and is broken, though I have a fix for that pending). Will make an issue to keep track

PR #4711, opened 2025-07-24
InvalidationTracker: Fix queries in the non-intersecting RWX case

bylaws:
This needs to return the base of the non-RWX interval, with a size that when added to the base is the start of the RWX interval.


bylaws:
Fixes a deadlock in steam under arm64ec

PR #4713, opened 2025-07-25
Profiler: Decouple profile stats from the profiler option

Sonicadvance1:
This option is free and only enabled if the config option is set. Enable it always at build time so that users can pick it up without enabling the full gpuviz/tracy paths.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -46,6 +46,7 @@ $end_info$
 #include "FEXCore/Utils/SignalScopeGuards.h"
 #include <FEXCore/Utils/Threads.h>
 #include <FEXCore/Utils/Profiler.h>
+#include <FEXCore/Utils/SHMStats.h>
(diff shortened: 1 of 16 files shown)


neobrain:
Since this has always been a logically exclusive feature anyway, could you move it to an entirely different file instead (called something that's a more distinguished name than "Profiler")?

I was going to bring this up in the initial review for the feature, but then that PR got merged early due to miscoordination.

Source/Common/Profiler.h
```diff
@@ -6,7 +6,7 @@ desc: Frontend profiler common code
 $end_info$
```
neobrain, line 6: SHMStats is much better, but the Common and Windows files are still using the old Profiler naming scheme. This is relevant because the name overloading makes it tricky to quickly open the right file in an IDE.

Sonicadvance1: Rebranded the remaining files.

bylaws (approved):

PR #4717, opened 2025-07-25
ArchHelpers: Remove pair usage in unaligned handler

Sonicadvance1:
It's being treated like optional, where a value means it has been handled, and no value means it hasn't been handled. Stop using pair in this case.


alyssarosenzweig (approved):
Oh nice

lioncash (approved):

PR #4718, opened 2025-07-25
ELFContainer: Remove tuple usage

Sonicadvance1:


alyssarosenzweig (approved):

lioncash:
Seems like the `<tuple>` header can be removed (or moved to the cpp file if there's any usages of it there)

Sonicadvance1:
> Seems like the `<tuple>` header can be removed (or moved to the cpp file if there's any usages of it there)

Nice, deleted.

lioncash (approved):

PR #4719, opened 2025-07-25
Runtime mode switch take 2

Sonicadvance1:
#3303 initially implemented this but it broke things.
It was then reverted in #3400.

A comment pointed out where I had missed the GDT limit being set at: https://github.com/FEX-Emu/FEX/pull/3303#pullrequestreview-2956968748

With that fixed, the breakages are now resolved, so we can bring this back.


bylaws (approved):
Tested with wow64/arm64ec

PR #4721, opened 2025-07-26
Pool constants as we go

alyssarosenzweig:
Rather than spamming out lots of Constant instructions then trying to deduplicate, just deduplicate as we go. This should be faster since it saves us an expensive remapping dance in ConstProp and avoids ever creating Constant instructions that won't be used later. There's a trivial instruction count (but not cycle count) regression for x87 since this relies on emitting code in a single pass for good results. Otherwise should get basically the same results as before. Draft since I haven't benchmarked even though the theory is sound.


alyssarosenzweig:
Total win on nodejs jit time:

```
	-1.82179% +/- 0.424777%
```

Sonicadvance1 (approved):
Just going to double check that this doesn't cause xess to get angry. I assume it won't, but going to double check.

PR #4722, opened 2025-07-27
CMake: Work around QCom disabling SVE in their chips

### LINKED ISSUE #4124: Fix classification script to work around qualcomm disabling SVE on their chips
Causes things compiled through AUR or native with default classification to SIGILL at the start. Need to check for enabled features through cpuinfo or cpuid flags and work around that.

Sonicadvance1:
Fixes #4124

clang feature checking can't check beyond MIDR, so compiling for a specific cortex version means compiling SVE on these CPUs that disabled them.

Just detect the particular situation in-which SVE isn't inside proc/cpuinfo and is one of the snapdragon cores that are supposed to support SVE. Then compile for Cortex-a78 instead.

### DIFF
--- a/Scripts/NeedDisabledSVE.py
+++ b/Scripts/NeedDisabledSVE.py
@@ -1,0 +1,73 @@
+#!/usr/bin/python3
+
+# Qualcomm in their infinite wisdom decided to disable SVE in a handful of SoCs.
+# When compiling for a specific CPU architecture or `-mcpu=native`, we need to ensure
+# that SVE is disabled on these platforms that had the feature disabled.
+# Check for the handful of Cortex CPUs that support SVE in hardware, but are disabled
+# in software.
+import re
+import sys
+
+def GetCPUFeatures():
+    File = open("/proc/cpuinfo", "r")
+    Lines = File.readlines()
+    File.close()
+
+    for Line in Lines:
+        if "Features" in Line:
+            Features = Line.split(":")[1].strip().split(" ")
+            return Features
+
+SnapdragonIDsWithDisabledSVE = {
+    # Snapdragon 8 Gen 3
+    tuple([0x41, 0xd82]): True, # Cortex-X4
+    tuple([0x41, 0xd81]): True, # Cortex-A720
+    tuple([0x41, 0xd80]): True, # Cortex-A520
+
+    # Snapdragon 8 Gen 2
+    tuple([0x41, 0xd4e]): True, # Cortex-X3
+    tuple([0x41, 0xd4d]): True, # Cortex-A715
+    tuple([0x41, 0xd47]): True, # Cortex-A710
+    tuple([0x41, 0xd46]): True, # Cortex-A510
+
+    # Snapdragon 8 Gen 1
+    tuple([0x41, 0xd48]): True, # Cortex-X2
+    # A710
+    # A510
+}
+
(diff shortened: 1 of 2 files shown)


alyssarosenzweig (approved):
garbage-tier cpu engineering :sob:

Scripts/NeedDisabledSVE.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python3
+import os
+import subprocess
+import re
+import sys
+import tempfile
+import platform
+
+def ListContainsRequired(Features, RequiredFeatures):
```
neobrain: What's the benefit over this function over just using `in` directly?

Scripts/NeedDisabledSVE.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python3
+import os
+import subprocess
+import re
+import sys
+import tempfile
+import platform
```
neobrain: Not all of these seem to be used, could you double-check which are actually needed?

Scripts/NeedDisabledSVE.py
```diff
@@ -0,0 +1,75 @@
...
+    Lines = File.readlines()
+    File.close()
+
+    for Line in Lines:
+        if "Features" in Line:
+            Features = Line.split(":")[1].strip().split(" ")
+            return Features
+
+
+SnapdragonIDsThatSupportSVE = {
```
neobrain: Why does the "Need *disabled* SVE" script return true for CPUs that "support SVE"?

Scripts/NeedDisabledSVE.py
```diff
@@ -0,0 +1,75 @@
+#!/usr/bin/python3
```
neobrain, line 1: Could you add a general comment explaining why this script is needed?

Scripts/NeedDisabledSVE.py
```diff
@@ -0,0 +1,75 @@
...
+
+    return False
+
+
+def main():
+    Features = GetCPUFeatures()
+
+    # If SVE is reported from cpuinfo just return.
+    if ListContainsRequired(Features, "sve"):
+        sys.exit(0)
```
lioncash: should these be returning the exit code, rather than calling `sys.exit()` directly? Only bringing it up, since the check for `__main__` down below already feeds the result of main() into a `sys.exit()` call.

Scripts/NeedDisabledSVE.py
Sonicadvance1: Removed and used directly.

Scripts/NeedDisabledSVE.py
Sonicadvance1: Removed unnecessary imports.

Scripts/NeedDisabledSVE.py
Sonicadvance1: Renamed object to be less confusing.

Scripts/NeedDisabledSVE.py
Sonicadvance1: Added a comment.

Scripts/NeedDisabledSVE.py
Sonicadvance1: Switched over to returning the value.

lioncash (approved):

PR #4728, opened 2025-07-28
Implement mono specific hacks

bylaws:
I have tested this across a decent number of more modern unity games. Of note is that older 32-bit MSVC versions don't inline InterlockedExchange, which prevents the first hack from working but doesn't negatively impact behaviour or perf.

This implements two mono specific 'hacks':

#### Calls
Function calls in mono generated code will initially jump to a dispatcher, which will compile the block then backpatch its caller to call the generated code directly. See:
https://github.com/Unity-Technologies/mono/blob/7ad1bda11b055b6d02ed9b1e2469a97934c157c8/mono/mini/tramp-amd64.c#L145
https://github.com/Unity-Technologies/mono/blob/7ad1bda11b055b6d02ed9b1e2469a97934c157c8/mono/mini/tramp-x86.c#L91

While the 64-bit case is large enough to never be inlined, in recent versions of 32-bit mono the backpatcher can be is inlined into the dispatcher. As a result simple function entry replacement isn't viable. In both cases however, all the SMC writes are contained within a single multiblock, and are the only XCHG operations in said block. 

Take advantage of this to allow the following approach:
- Start with the default FEX SMC behaviour
- When mono is loaded, set a variable to mark as such
- If we receive an SMC signal, said variable is set and the writing instruction is an XCHG in the mono dll region. The block entrypoint at this moment will be the address of the multiblock containing all SMC writes
- Inform the backend of this, disable all SMC detection by reprotecting RWX memory as RWX and invalidate the block.
- When compilation of that block is triggered again, the JIT will replace all XCHG operations with a special invalidating-write IROp.
- When the block is ran, this will then write the modified code and invalidate the FEX code cache without triggering signals or unaligned atomics

#### Jumps
NOTE: This code path isn't exercised in any game i could find, but I have validated it against the assembly patterns it checks for.

These are handled in mono by keeping a hash table mapping a block address to all non-backpatched jump instructions to it, and then iterating this on every compile. The code around this has changed several times throughout mono's history and isn't particularly detectable, so instead detect the generated instruction pattern for these jumps and force full SMC checking for them.

X64 new: https://github.com/Unity-Technologies/mono/blob/9af80327e4032121e18992091799f4deda183d4d/mono/mini/mini-amd64.c#L5371
X64 old: https://github.com/Unity-Technologies/mono/blob/unity-5.6/mono/mini/mini-amd64.c#L4051

X86 old: https://github.com/Unity-Technologies/mono/blob/unity-5.6/mono/mini/mini-x86.c#L2812
X86 new: https://github.com/Unity-Technologies/mono/blob/9af80327e4032121e18992091799f4deda183d4d/mono/mini/mini-x86.c#L3140

### Disabling multiblock after-call exploration for JIT code
Mono generates a significant number of noreturn calls, and as described above, each individual call in a block will trigger an invalidation when hit due to backpatching. This effectively causes a quadratic amount of wasted work to be done. For example:
1: call a
2: call b
3: call c
4:return

FEX will compile 1-4 in a single multiblock, and run it, 1 will be backpatched, this will invalidate the block. A multiblock containing 2-4 would then be compiled and ran, causing 2 to be backpatched. etc. 

This behaviour is entirely avoided by disabling after-call exploration for JIT code. At the cost of a slight perf hit (I expect not more than 5%)

TODO: This *requires* multiblock with a high enough maxinst for correctness. I am unsure how to best handle that.


bylaws:
Limbo can be solved separately, and only happens with maxinst >1000. This is due to bad size estimation/far branches

Sonicadvance1 (approved):
Let's hit it.

PR #4730, opened 2025-07-28
unittests: Fixes thunk unittest path

Sonicadvance1:
This hasn't been run in CI for awhile so this was missed when I was testing it. I need to see what I can do to get this going in CI again.

### DIFF
--- a/unittests/ThunkFunctionalTests/CMakeLists.txt
+++ b/unittests/ThunkFunctionalTests/CMakeLists.txt
@@ -2,7 +2,7 @@ set(FUNCTIONAL_DEPENDS "")
 
 function(AddThunksTest Bin ThunksFile)
   set (ARGS
-    "-t" "${CMAKE_INSTALL_PREFIX}/lib/fex-emu/HostThunks"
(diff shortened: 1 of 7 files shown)


unittests/ThunkFunctionalTests/CMakeLists.txt
```diff
@@ -1,8 +1,13 @@
 set(FUNCTIONAL_DEPENDS "")
 
+set (HOSTLIBS_DATA_DIRECTORY "" CACHE PATH "Global data directory (override)")
+if (NOT HOSTLIBS_DATA_DIRECTORY)
+  set(HOSTLIBS_DATA_DIRECTORY "${CMAKE_INSTALL_FULL_LIBDIR}/fex-emu")
+endif()
+
```
neobrain: This variable is already declared in ThunkLibs/HostLibs/CMakeLists.txt, so it's better to move it to the root CMakeLists.txt inside the `BUILD_THUNKS` branch to avoid the duplication.

Sonicadvance1: Moved to the root CMakeLists file.

PR #4731, opened 2025-07-28
github: Upload armtifacts

Sonicadvance1:


Sonicadvance1:
Okay. This will generate build artifacts for the wine DLL files when something is pushed to main.
It won't run on PRs to ensure we don't burn a bunch of artifact space.

neobrain (approved):

PR #4742, opened 2025-07-30
Clear IE flag on fninit

pmatos:
Tests and instcountci changes.


bylaws:
Can you extend all other writes from the frontend to also set this as appropriate? If you grep X87FLAG_C2_LOC

pmatos:
> Can you extend all other writes from the frontend to also set this as appropriate? If you grep X87FLAG_C2_LOC

Oh my - I missed quite a few bits here... thanks for pointing it out.

pmatos:
@bylaws I think I caught all places.

PR #4746, opened 2025-07-30
ASIMDOps: Move base opcode into ASIMDModifiedImm()/ASIMDShiftByImm()

lioncash:
Deduplicates the need to respecify the opcode over and over for every instruction in the respective categories


Sonicadvance1 (approved):
weeeeee

PR #4752, opened 2025-08-01
RegisterAllocationPass: defer next-use analysis

alyssarosenzweig:
This is expensive and only needed for spilling, so only do it for spilling. This
    complicates the RA a bit but speeds us up on average since most blocks
    don't spill. Total results of this change (including the prep commits that
    slowed things down temporarily):
    
            -1.71952% +/- 0.455996%

(This requires us to introduce the notion of kill-bits to replace next-use info for the non-spilling code path)


Sonicadvance1 (approved):
Fancy Fancy.

PR #4758, opened 2025-08-04
OpcodeDispatcher: optimize SetX87FTW

alyssarosenzweig:
In b8dd5d95b ("OpcodeDispatcher: optimize X87FTWTag"), we optimized
    X87FTWTag using an efficient Morton interleave operation. Here, we do the
    inverse, optimizing SetX87FTW using an efficient Morton deinterleave
    operation.


Sonicadvance1 (approved):
Nice nice nice nice.

PR #4759, opened 2025-08-04
RegisterAllocationPass: simplify next-use logic

alyssarosenzweig:
I doubt this will fix the regression but it might make it easier to identify.


Sonicadvance1 (approved):
This fixes Monster Hunter Rise here.

PR #4764, opened 2025-08-06
Drop ConstProp

alyssarosenzweig:
At this point, this pass only exists to inline constants. That's pretty silly. This PR reimplements ConstProp's remaining functionality with "smarter" IREmitter abstractions - capable of inlining constants as we go - allowing us to drop ConstProp. Net deletion of 219 lines... while speeding up the JIT translation time about 4%!


Sonicadvance1 (approved):
Nice, this even increases perf in Cyberpunk slightly because of its consistent SMC.

Sonicadvance1:
Needs a base of re.

PR #4765, opened 2025-08-06
Support (downstream) kernel-side unaligned atomic handling

bylaws:
This enables support for a kernel patch that replicates the current FEX userspace unaligned atomic handler in kernelspace.

Kernel branch: https://github.com/bylaws/linux/commits/fex-unaligned-atomic/
Wine branch: https://github.com/bylaws/wine/commits/upstream-arm64ec/

Passes FEX ASM tests (when I manually hacked each option and kernel enablement in).
This speeds up backpatching/atomic handling by roughly 500x and properly supports cross-process split locks (albeit this isn't tested).


Sonicadvance1:
It looks like your kernel patch has a bug when `PR_ARM64_UNALIGN_ATOMIC_BACKPATCH` is enabled that if multiple threads hit the same SIGBUS location that backpatching is checked and PC is modified to ensure that dmb+str is still executed correctly? I see you're using CAS loops to ensure that competing threads at least won't...directly fight each other.

bylaws:
> It looks like your kernel patch has a bug when `PR_ARM64_UNALIGN_ATOMIC_BACKPATCH` is enabled that if multiple threads hit the same SIGBUS location that backpatching is checked and PC is modified to ensure that dmb+str is still executed correctly? I see you're using CAS loops to ensure that competing threads at least won't...directly fight each other.

In such a case the atomic op will be emulated in the kernel for both threads (see how the CAS falls through), so there's no backwards jumping necessary here. The worst that can happen is an emulated atomic write which hits a DMB after returning, which seems fine. This works since each load and its respective atomic variant share the same instruction operand layout

FEXCore/Source/Interface/Core/JIT/AtomicOps.cpp
```diff
@@ -322,13 +322,26 @@ DEF_OP(AtomicFetchNeg) {
...
   auto MemSrc = GetReg(Op->Addr);
 
-  ARMEmitter::BackwardLabel LoopTop;
-  Bind(&LoopTop);
-  ldaxr(SubEmitSize, TMP2, MemSrc);
-  neg(EmitSize, TMP3, TMP2);
-  stlxr(SubEmitSize, TMP4, TMP3, MemSrc);
-  cbnz(EmitSize, TMP4, &LoopTop);
-  mov(EmitSize, GetReg(Node), TMP2.R());
+  if (CTX->HostFeatures.SupportsAtomics) {
```
Sonicadvance1, line 325: I'm curious, you have a little unaligned atomic bench. Could you measure the performance between these two paths, aligned and unaligned? Completely uncontended is fine.
I believe the CAS loop will be slower, but I don't have facts to back that up.

bylaws: I assume you want this with userspace atomic handling for both?

Sonicadvance1: Yea, since kernel can't handle the LLSC path anyway.

bylaws: LLSC:
Aligned time taken: 7916300 ns
Unaligned time taken: 2924085300 ns
CAS:
Aligned time taken: 9404400 ns
Unaligned time taken: 2914928800 ns

Sonicadvance1: Neat. Thanks for catering to my whims for the random benchmark. Falls in line with my expectations but always good to retest.

bylaws:
I think it might be worth running the kernel side patches for this through someone familiar with things to be sure I'm not introducing any particular security holes before merging

Sonicadvance1:
Should we split this patch set out so we can merge the immediately possible change, and keep the kernel bits in a smaller PR while kernel bits are investigated?

bylaws:
### just the first patch? Sure

Sonicadvance1:
Yea, I guess swapping AtomicNeg to CAS if possible, and unifying TSO/Paranoid. Will at least keep the kernel bits more self-contained.

Sonicadvance1:
This is now merged as part of #4985

PR #4766, opened 2025-08-06
Disable 3DNow by default on WOW64 FEX

bylaws:
Works around a bug in WOW64 FEX that occurs using MS d3d9x when 3DNow is enabled.

### DIFF
--- a/Source/Common/HostFeatures.cpp
+++ b/Source/Common/HostFeatures.cpp
@@ -435,6 +435,7 @@ static void OverrideFeatures(FEXCore::HostFeatures* Features, uint64_t ForceSVEW
   ENABLE_DISABLE_OPTION(SupportsSVEBitPerm, SVEBITPERM, SVEBITPERM);
   ENABLE_DISABLE_OPTION(SupportsPreserveAllABI, PRESERVEALLABI, PRESERVEALLABI);
   ENABLE_DISABLE_OPTION(SupportsWFXT, WFXT, WFXT);
+  ENABLE_DISABLE_OPTION(Supports3DNow, 3DNOW, 3DNOW);
   GET_SINGLE_OPTION(Crypto, CRYPTO);
 
 #undef ENABLE_DISABLE_OPTION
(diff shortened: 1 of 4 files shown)


Sonicadvance1 (approved):
Get your act together Microsoft.

Source/Common/HostFeatures.cpp
```diff
@@ -497,6 +498,14 @@ FEXCore::HostFeatures FetchHostFeatures(FEX::CPUFeatures& Features, bool Support
 #endif
   HostFeatures.SupportsAVX = true;
 
+#ifdef _WIN32
+  // Disable 3DNow! by default to better match the set of extensions exposed on modern CPUs
+  // Works around a bug breaking some uses of 3DNow in WOW64 FEX
```
neobrain, line 503: Could you elaborate the comment on what the actual bug is for future reference? Right now it's not apparent what would need to happen to unconditionally re-enable the feature.

bylaws: Will do once we know what the bug actually is. Potentially it's on the Microsoft side

neobrain: Was there any progress on this?

If not, could you fix the documentation to state how to reproduce the issue so that we can reliably verify if it's still needed in the future?

We can add more details later if more specific information is known, but this is the type of issue that just falls under the radar otherwise.

bylaws: Ryan's ordering an AMD platform supporting it, turns out regular fex is also broken but the feature detection doesn't pick it up there. might be a Microsoft bug

neobrain: Ok. So how exactly does the bug manifest? Is it a crash? Is there an easily testable application that will break?

bylaws: Oh just things using ms d3d9x will run with a black screen, e.g. fallout NV or old EA games

Issue #4493, opened 2025-04-10
[ARM64EC] Steam on Wine has random crash in steamwebhelper


Art-Chen:
**FEX Version:** git head: 09e622d5a03d55e39ff81668ab08a6595ddc5c5b
**Wine Version**: Wine tag 10.5

### Log

`097c:warn:seh:dispatch_exception "src\\common\\asyncfileiohandler.cpp (769) : dwRet != WAIT_FAILED\n"
097c:trace:seh:dispatch_exception code=40010006 (DBG_PRINTEXCEPTION_C) flags=0 addr=000000007A322339
097c:trace:seh:dispatch_exception  info[0]=0000000000000040
097c:trace:seh:dispatch_exception  info[1]=000000000FF9E900
097c:trace:seh:dispatch_exception  pc=0000007fff89b5c8  sp=0000000115ffee30  lr=0000000000000000  fp=000000007ac77000
097c:trace:seh:dispatch_exception  x0=0000000000000000  x1=000000000ff9e428  x2=0000000000000001  x3=00000001038bba90
097c:trace:seh:dispatch_exception  x4=0000000115fff160  x5=0000000000000000  x6=000000000ece2000  x7=000000007a851cec
097c:trace:seh:dispatch_exception  x8=0000000000000090  x9=0000000000005d17 x10=00000000000001e0 x11=0000000100210000
097c:trace:seh:dispatch_exception x12=00000000000000b0 x13=00000000000011d1 x14=000000000ff9e404 x15=000000000ff9e418
097c:trace:seh:dispatch_exception x16=0000007fffa13210 x17=0000000115fff1f0 x18=000000000ece0000 x19=000000000ff9e700
097c:trace:seh:dispatch_exception x20=000000000ff9e428 x21=0000000000000001 x22=00000001038ce3e0 x23=0000000115fffd20
097c:trace:seh:dispatch_exception x24=0000000000000001 x25=000000007ac78000 x26=0000000000000246 x27=000000007ac78002
097c:trace:seh:dispatch_exception x28=0000000115fff2d0 cpsr=20000000 fpcr=00000000 fpsr=00000000
097c:trace:seh:call_seh_handlers calling handler 0000007FFF89B5E0 (rec=0000000115FFED70, frame=115fff1f0 context=0000000115FFE9C0, dispatch=0000000115FFE5F8)
[...]

PR #4768, opened 2025-08-07
Avoid OOB reads for mem sources in some float conversion ops

bylaws:
Fixes the other error reported in #4493


Sonicadvance1 (approved):
Nice! Amazing that this was sitting there for so long.

PR #4771, opened 2025-08-08
Local JIT translation time wins

alyssarosenzweig:
Staring at perf report.

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
@@ -148,7 +148,8 @@ private:
   };
 
   bool HasSource(IROp_Header* I, PhysicalRegister Reg) {
-    for (auto s = 0; s < IR::GetRAArgs(I->Op); ++s) {
+    int NumArgs = IR::GetRAArgs(I->Op);
+    for (int s = 0; s < NumArgs; ++s) {
       if (I->Args[s].IsImmediate()) {
         // When spilling for a destination, we'll see register sources
         if (PhysicalRegister(I->Args[s]) == Reg) {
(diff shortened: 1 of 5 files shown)


FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -148,7 +148,8 @@ class ConstrainedRAPass final : public RegisterAllocationPass {
   };
 
   bool HasSource(IROp_Header* I, PhysicalRegister Reg) {
-    for (auto s = 0; s < IR::GetRAArgs(I->Op); ++s) {
+    unsigned NumArgs = IR::GetRAArgs(I->Op);
+    for (auto s = 0; s < NumArgs; ++s) {
```
lioncash: s is an int here, so this can cause signedness warnings

FEXCore/Source/Interface/IR/Passes/RegisterAllocationPass.cpp
```diff
@@ -548,7 +549,8 @@ bool ConstrainedRAPass::TryPostRAMerge(Ref LastNode, Ref CodeNode, IROp_Header*
   if (PhysicalRegister(LastNode) == PhysicalRegister(CodeNode) && KillMove(LastOp, IROp, LastNode, CodeNode)) {
     LOGMAN_THROW_A_FMT(!PhysicalRegister(CodeNode).IsInvalid(), "invariant");
 
-    for (auto s = 0; s < IR::GetRAArgs(IROp->Op); ++s) {
+    unsigned NumArgs = IR::GetRAArgs(IROp->Op);
+    for (auto s = 0; s < NumArgs; ++s) {
```
lioncash: ditto (and for the two below)

lioncash (commented):
just a few minor nits

alyssarosenzweig:
switched to ints throughout

Sonicadvance1 (approved):

PR #4776, opened 2025-08-12
Windows: Misc race and correctness fixes

bylaws:
I didn't encounter these in any applications, but thought they were worth fixing. Steam/rockstar launcher/unity games all still work so hopefully this hasn't introduced any new issues


Sonicadvance1 (approved):
Fancy fancy

Source/Windows/Common/Handle.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+  explicit ScopedHandle(HANDLE Handle)
+    : Handle(Handle) {}
+
+  // Move-only type
+  ScopedHandle(const ScopedHandle&) = delete;
+  ScopedHandle& operator=(ScopedHandle&) = delete;
+  ScopedHandle(ScopedHandle&& rhs)
+    : Handle(rhs.Handle) {
+    Handle = INVALID_HANDLE_VALUE;
```
lioncash: Is this actually intended to be setting `rhs.Handle` to an invalid handle value, rather than the handle for the object being moved to?

PR #4778, opened 2025-08-12
HostFeatures: Fix insufficient documentation to reproduce 3DNow! bug

neobrain:
This ensures we don't forget about it if bug diagnosis leads nowhere.

### DIFF
--- a/Source/Common/HostFeatures.cpp
+++ b/Source/Common/HostFeatures.cpp
@@ -499,8 +499,9 @@ FEXCore::HostFeatures FetchHostFeatures(FEX::CPUFeatures& Features, bool Support
   HostFeatures.SupportsAVX = true;
 
 #ifdef _WIN32
-  // Disable 3DNow! by default to better match the set of extensions exposed on modern CPUs
-  // Works around a bug breaking some uses of 3DNow in WOW64 FEX
+  // Disable 3DNow! by default to better match the set of extensions exposed on modern CPUs.
+  // This works around a bug that manifests in some games using native d3dx9 DLLs (most easily reproduced in WoW64 builds).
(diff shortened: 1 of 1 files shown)


Source/Common/HostFeatures.cpp
```diff
@@ -499,8 +499,9 @@ FEXCore::HostFeatures FetchHostFeatures(FEX::CPUFeatures& Features, bool Support
   HostFeatures.SupportsAVX = true;
 
 #ifdef _WIN32
-  // Disable 3DNow! by default to better match the set of extensions exposed on modern CPUs
-  // Works around a bug breaking some uses of 3DNow in WOW64 FEX
+  // Disable 3DNow! by default to better match the set of extensions exposed on modern CPUs.
+  // This works around a bug in WOW64 FEX that manifests in some games using d3dx9.
```
bylaws:
```suggestion
  // This works around a bug that manifests in some games using d3dx9.
```
prob more accurate but yeah this is fine

neobrain: Fair. Since you're saying "probably", is this an expectation or has Linux FEX been verified to also break?

bylaws: linux FEX doesn't break as is because wine's 3DNow! detection only enables it with an AuthenticAMD vendor name and we default to intel. but if 3DNow! is forced on then yeah we have the same symptoms

neobrain: Alright - adapted the wording to no longer imply it's WoW64 specific while still recommending WoW64 for testing since it will break out of the box.

bylaws (approved):

PR #4779, opened 2025-08-13
Windows: Support EnvLoader via _environ

asLody:


Source/Windows/Common/CRT/Misc.cpp
```diff
@@ -25,12 +28,35 @@ void InitEnv() {
...
+
   RtlReleasePebLock();
 }
 
 __attribute__((used, section(".CRT$FEXB"))) void (*_InitEnv)(void) = InitEnv;
 } // namespace
 
+char*** __p__environ() {
+  static char** environ_ptr = nullptr;
+  environ_ptr = EnvArray;
```
bylaws: Is return &EnvArray; not enough here?

bylaws (approved):

PR #4784, opened 2025-08-16
FEXCore: Move LDT/GDT management to the frontend

Sonicadvance1:
This will allow the frontend to manage LDT and GDT so that Linux will be
able to install 32-bit code segments.

A slight behaviour change on wow64 and arm64ec where threads are now sharing their gdt entries, but because we don't yet support having the guest modify those it should be a non-issue? I need to some testing with arm64ec/wow64 still to ensure this doesn't break things, but this is the probably the scariest change in this series.

### DIFF
--- a/Source/Windows/WOW64/Module.cpp
+++ b/Source/Windows/WOW64/Module.cpp
@@ -508,6 +508,7 @@ void BTCpuProcessInit() {
   CTX->SetSignalDelegator(SignalDelegator.get());
   CTX->SetSyscallHandler(SyscallHandler.get());
   CTX->InitCore();
+
   InvalidationTracker.emplace(*CTX, Threads);
 
   auto NtDllX86 = reinterpret_cast<SYSTEM_DLL_INIT_BLOCK*>(GetProcAddress(NtDll, "LdrSystemDllInitBlock"))->ntdll_handle;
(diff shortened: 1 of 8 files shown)


Sonicadvance1:
Turns out the arm64ec/wow64 path was a lot easier than expected. Because arm64ec doesn't allow LDT tinkering, I can just allocate the single gdt buffer and mirror it, very similar to how thread creation worked before.

Future work can have arm64ec/wow64 take advantage of the InternalThreadState object's `FrontendPtr` if we want more data without consuming more TLS slots or ChpeV2CpuAreaInfo slots.

Source/Windows/WOW64/Module.cpp
```diff
@@ -565,10 +566,28 @@ void BTCpuThreadInit() {
   FEX::Windows::InitCRTThread();
   auto* Thread = CTX->CreateThread(0, 0);
 
+  // Default segment setup.
+  auto Frame = Thread->CurrentFrame;
+  auto NewSegments = reinterpret_cast<FEXCore::Core::CPUState::gdt_segment*>(calloc(32, sizeof(FEXCore::Core::CPUState::gdt_segment)));
```
bylaws: new/delete is probably better?

Sonicadvance1: Sure, switched over to new[]/delete[].

bylaws (approved):

PR #4794, opened 2025-08-22
Scripts: Fix use of removed helper script

neobrain:
This was accidentally changed back to clang-format.py in 3b322d8f375b32f08ea4c5824af4536ba7fde98a.


pmatos:
Oh - so reformat.sh was not even working. :-/ 
That's weird because I am pretty sure that's what I used to do the whole-tree reformatting. 
But yeah, this looks good. Maybe there was some file version mix-up locally.

PR #4795, opened 2025-08-22
Do not use diff_from_common_commit

pmatos:
It was not complaining on improper formatting for new files.

### DIFF
--- a/.github/workflows/pr-code-format.yml
+++ b/.github/workflows/pr-code-format.yml
@@ -60,7 +60,6 @@ jobs:
           START_REV: ${{ github.event.pull_request.base.sha }}
           END_REV: ${{ github.event.pull_request.head.sha }}
           CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
-        # Using --diff_from_common_commit option available in clang-format-19
         run: |
           python ./External/code-format-helper/code-format-helper.py \
             --repo "FEX-emu/FEX" \
--- a/External/code-format-helper/code-format-helper.py
+++ b/External/code-format-helper/code-format-helper.py
@@ -214,7 +214,6 @@ class ClangFormatHelper(FormatHelper):
             self.clang_fmt_path,
             "--binary=clang-format-19",
             "--diff",
-            "--diff_from_common_commit",
         ]
 
         if args.start_rev and args.end_rev:


pmatos:
@neobrain this should fix the issue. See:

Screenshot From 2025-08-22 15-07-50

I will remove your file from the PR so that this can be merged.

neobrain:
Successfully failing as intended now, thanks! :partying_face:

neobrain:
What's the trade-off for removing this option though?

Since we have a few badly formatted files in the tree now, do we need to run and merge a reformat before/when merging this PR?

pmatos:
> What's the trade-off for removing this option though?
> 

This option was enabled when we moved to clang-format-19 because initially I followed llvm-project suggestion. My feeling is that this was useful to choose the commit to compare the changes against but I think I was misled and llvm-project also removed this through https://github.com/llvm/llvm-project/commit/f11d46c1514d0ef4d3e92a96a9d330a9bc3164e1 .

As it turns out, llvm-project went further and is not comparing HEAD to HEAD~1 (https://github.com/llvm/llvm-project/commit/2898c3e0bbf904c68efb194525bcdfeedb3014c2). I don't understand this. I think we need to compare the changes from HEAD to the merge-base between main and the HEAD (ie. the point at which the branch for the PR was created).

> Since we have a few badly formatted files in the tree now, do we need to run and merge a reformat before/when merging this PR?

It's not strictly necessary but we can run a reformat - sure. We can do it once we land this and are sure it's working.

pmatos:
> > What's the trade-off for removing this option though?
> 
> This option was enabled when we moved to clang-format-19 because initially I followed llvm-project suggestion. My feeling is that this was useful to choose the commit to compare the changes against but I think I was misled and llvm-project also removed this through [llvm/llvm-project@f11d46c](https://github.com/llvm/llvm-project/commit/f11d46c1514d0ef4d3e92a96a9d330a9bc3164e1) .

The idea what `--diff_from_common_commit` was to the use that and instead of using `github.event.pull_request.base.ref` we could just pass to `clang-format` main and HEAD. But we have that value from github so it doesn't really help us. Also, for some reason I didn't dig into new files were being ignored.

neobrain:
> It's not strictly necessary but we can run a reformat - sure. We can do it once we land this and are sure it's working.

Yeah that's fair, I just wasn't sure if pending PRs would fail formatting with this fixed before that reformat. Since this issue just affected newly added files it should be fine though.

> This option was enabled when we moved to clang-format-19 because initially I followed llvm-project suggestion. My feeling is that this was useful to choose the commit to compare the changes against but I think I was misled and llvm-project also removed this through https://github.com/llvm/llvm-project/commit/f11d46c1514d0ef4d3e92a96a9d330a9bc3164e1 .
> 
> As it turns out, llvm-project went further and is not comparing HEAD to HEAD~1 (https://github.com/llvm/llvm-project/commit/2898c3e0bbf904c68efb194525bcdfeedb3014c2). I don't understand this. I think we need to compare the changes from HEAD to the merge-base between main and the HEAD (ie. the point at which the branch for the PR was created).
>
> The idea what --diff_from_common_commit was to the use that and instead of using github.event.pull_request.base.ref we could just pass to clang-format main and HEAD. But we have that value from github so it doesn't really help us. Also, for some reason I didn't dig into new files were being ignored.

Sounds reasonable, thanks for elaborating! Let's just drop the flag as proposed then.

Regarding the LLVM change (comparing HEAD to HEAD~1) it seems to be purely a consideration of saving CI resources (at least that's what the commit message suggests), but for us the formatting workflow is fast enough.

PR #4796, opened 2025-08-23
ELFCodeLoader: Fixes BRK allocation with non-PT_DYN ELF files

Sonicadvance1:
When loading an ELF file (usually PT_EXEC), we would have BSS regions that ended up in the high pages of the address space. We would then attempt mapping BRK after whatever the highest address ending up being.

This was problematic because we used `MAP_FIXED` which means that the BRK region could cross in to 64-bit address space, overwrite VDSO, overwrite the stack, vsyscall, maybe a couple of other things.

Instead of letting that happen, reorder some of the logic so that in `LoadElfFile` will ensure the full `BRK_SIZE` is allocated (or return zero) and then we can ensure we're never overwriting other various memory regions.

This doesn't fix two fundamental issues that FEX has with brk:
* If BRK gets mapped below the ELF, that should mean it isn't mapped at all
  * Linux-isms, doubt this breaks new software
* We still map the full 8MB, which isn't correct
  * I'm working on fixing this issue, which is why I encountered this.

### DIFF
--- a/Source/Tools/FEXLoader/ELFCodeLoader.h
+++ b/Source/Tools/FEXLoader/ELFCodeLoader.h
@@ -48,15 +48,29 @@ class ELFCodeLoader final : public FEX::CodeLoader {
   uintptr_t BrkStart {};
   uintptr_t StackPointer {};
 
-  size_t CalculateTotalElfSize(const fextl::vector<Elf64_Phdr>& headers) {
-    auto first = std::find_if(headers.begin(), headers.end(), [](const Elf64_Phdr& Header) { return Header.p_type == PT_LOAD; });
-    auto last = std::find_if(headers.rbegin(), headers.rend(), [](const Elf64_Phdr& Header) { return Header.p_type == PT_LOAD; });
+  // This calculates the map size for ET_DYN type ELF files.
+  // Can't be used for ET_EXEC ELF files because they can have large virtual mapping holes.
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXLoader/ELFCodeLoader.h
```diff
@@ -48,15 +48,29 @@ class ELFCodeLoader final : public FEX::CodeLoader {
   uintptr_t BrkStart {};
   uintptr_t StackPointer {};
 
-  size_t CalculateTotalElfSize(const fextl::vector<Elf64_Phdr>& headers) {
-    auto first = std::find_if(headers.begin(), headers.end(), [](const Elf64_Phdr& Header) { return Header.p_type == PT_LOAD; });
-    auto last = std::find_if(headers.rbegin(), headers.rend(), [](const Elf64_Phdr& Header) { return Header.p_type == PT_LOAD; });
+  // This calculates the map size for ET_DYN type ELF files.
+  // Can't be used for ET_EXEC ELF files because they can have large virtual mapping holes.
+  size_t CalculateDYNELFSize(const fextl::vector<Elf64_Phdr>& headers) {
```
lioncash: Off a small look, it seems like this function can be either static or const

Sonicadvance1: Made static

lioncash (approved):

PR #4797, opened 2025-08-24
More BRK fixes

Sonicadvance1:
This fixes some minor edge cases in FEX's BRK emulation, with these changes in place, FEXInterpreter can self-host the TestHarnessRunner built for x86-64.

Specifically fixes a couple of things:
* Unmap the whole BRK space at startup
  * Needs to be initially reserved to ensure nothing gets mapped in the space, but then unmapped to ensure the VA is free
  * Technically gives ~8MB more VA space to 32-bit applications
  * Required to support the visibility rules that Linux imposes on brk
* Changes brk map granularity to a page
  * Again required for VA mapping visibility rules
* Fixes a bug where I forgot to munmap the brk region if it was set to 0 length
* Removes a workaround for old kernels that don't support `MAP_FIXED_NOREPLACE`


lioncash:
Needs a rebase

Sonicadvance1:
Rebased.

PR #4798, opened 2025-08-25
Initial python generated files

Sonicadvance1:
This is the initial pull request for switching over to constexpr x86 tables with runtime selection of 32-bit/64-bit (and technically f64/f80 x87).

This hits the easiest two tables to showcase how the generated tables will work while leaving the additional work to a follow-up PR.


Sonicadvance1:
Interesting observation
```
ryanh@ryanh-TR5000:/mnt/Work/Work/work/FEXNew/Build$ /mnt/Work/Projects/bloaty/Build/bloaty ../Build_Radxa/FEXPost -- ../Build_Radxa/FEXPre
    FILE SIZE        VM SIZE
 --------------  --------------
  +129%  +108Ki  +129%  +108Ki    .data.rel.ro
  +0.0%    +256  +0.0%    +256    .text
  +0.1%    +152  +0.1%    +144    .eh_frame
  +0.0%     +48  +0.0%     +48    .rela.dyn
   +40%     +16   +40%     +16    [LOAD #2 [R]]
  [ = ]       0  [DEL]     -24    [LOAD #5 [RW]]
  [DEL]     -24  [ = ]       0    [Unmapped]
  -9.0%    -104  -9.0%    -104    .got
  [ = ]       0 -60.6%    -344    .relro_padding
 -13.4%  -108Ki -13.4%  -108Ki    .data
  +0.0%    +344  -0.0%      -8    TOTAL
  ```
  
While not an immediate goal of this, the tables have effectively moved from rw `.data` section to ro `.data.rel.ro` section. Which means that these pages are going to get shared between multiple FEX processes and reduce memory usage slightly. So kind of cool.

**Comment by neobrain** [comment]
Could you elaborate on what the requirements for runtime selection of 32-bit/64-bit are, why the current tables can't meet them, and how this approach does?

What options besides scripted code generation have you considered and why didn't they work out?

**Comment by Sonicadvance1** [comment]
> Could you elaborate on what the requirements for runtime selection of 32-bit/64-bit are, why the current tables can't meet them, and how this approach does?

The easy question to answer there is that we need to have both 32-bit and 64-bit instruction descriptions in our decoder tables. We currently "install" architecture specific instructions in to the tables on initialization depending on the operating mode of the application we loaded. Since x86-64 can switch between the two, this behaviour doesn't work.

The current layout is "description tables" and "OpcodeDispatcher tables".
Description tables are augmented at runtime to have additional 32-bit or 64-bit descriptions overwriting its base description. Additionally the OpcodeDispatcher entries are then installed in to the Description tables, with bitness being a selection criteria for which ones get installed there as well.

The new path has the python generate the Description Tables and OpcodeDispatch tables directly in to one constexpr table, which ends up in RO pages.
Example of the generated table:
```cpp
constexpr std::array<X86InstInfo, 2048> H0F38TableOps = {{
  {"PSHUFB", TYPE_INST, GenFlagsSameSize(SIZE_64BIT)  | FLAGS_MODRM | FLAGS_XMM_FLAGS | FLAGS_SF_MMX, 0, { .OpDispatch = &IR::OpDispatchBuilder::PSHUFBOp } },
  <...>
  ```

The forthcoming changes are for generating an "Indirect" path which will support runtime selection of bitness in the table by having one step of indirection.

Example of generated table with indirect step.
```cpp
constexpr std::array<const X86InstInfo[2], 1> Test_Arch = {{
  {
    {"32Bit_Inst", TYPE_INST, FLAGS_NONE, 0, { .OpDispatch = &IR::OpDispatchBuilder::POPOp} },
    {"64Bit_Inst", TYPE_INST, FLAGS_NONE, 0, { .OpDispatch = &IR::OpDispatchBuilder::PUSHOp} },
  }
}};

constexpr std::array<X86InstInfo, 2> Test = {{
  // Indirect operation
  {"", TYPE_ARCH_DISPATCHER, FLAGS_NONE, 0, { .Indirect = Test_Arch[0] } },
  // Direct instruction encoding
  {"CMPXCHG8B/16B", TYPE_INST, FLAGS_MODRM | FLAGS_SF_MOD_DST | FLAGS_SF_MOD_MEM_ONLY, 0, { .OpDispatch = &IR::OpDispatchBuilder::CMPXCHGPairOp }},
}};
```

How this works is that when the FrontEnd sees type `TYPE_ARCH_DISPATCHER` it will fetch the indirect X86InstInfo based on current bitness, which is in an interleaved array and continue parsing as usual.

> 
> What options besides scripted code generation have you considered and why didn't they work out?

I have tried hand-writing these multiple times over the past few months and it always became unwieldy due to limitations of working around C++'s requirements of constexpr.

Sonicadvance1:
A bit more with the initial implementation hammering through.

Indirect table:
```cpp
enum Arch_LUT_SecondInstGroupOps {
  ENTRY_264 = 0,
  ENTRY_265 = 1,
  ENTRY_266 = 2,
  ENTRY_267 = 3,
};

constexpr std::array<X86InstInfo[2], 4> Arch_LUT_Table_SecondInstGroupOps = {{
  {
    {"", TYPE_INVALID, FLAGS_NONE, 0, { .OpDispatch = nullptr } },
    {"RDFSBASE", TYPE_INST, FLAGS_MODRM | FLAGS_SF_MOD_DST | FLAGS_SF_MOD_REG_ONLY, 0, { .OpDispatch = &IR::OpDispatchBuilder::Bind<&IR::OpDispatchBuilder::ReadSegmentReg, IR::OpDispatchBuilder::Segment::FS> } },
  },
  {
    {"", TYPE_INVALID, FLAGS_NONE, 0, { .OpDispatch = nullptr } },
    {"RDGSBASE", TYPE_INST, FLAGS_MODRM | FLAGS_SF_MOD_DST | FLAGS_SF_MOD_REG_ONLY, 0, { .OpDispatch = &IR::OpDispatchBuilder::Bind<&IR::OpDispatchBuilder::ReadSegmentReg, IR::OpDispatchBuilder::Segment::GS> } },
  },
  {
    {"", TYPE_INVALID, FLAGS_NONE, 0, { .OpDispatch = nullptr } },
    {"WRFSBASE", TYPE_INST, GenFlagsDstSize(SIZE_64BIT) | FLAGS_MODRM | FLAGS_SF_MOD_REG_ONLY, 0, { .OpDispatch = &IR::OpDispatchBuilder::Bind<&IR::OpDispatchBuilder::WriteSegmentReg, IR::OpDispatchBuilder::Segment::FS> } },
  },
  {
    {"", TYPE_INVALID, FLAGS_NONE, 0, { .OpDispatch = nullptr } },
    {"WRGSBASE", TYPE_INST, GenFlagsDstSize(SIZE_64BIT) | FLAGS_MODRM | FLAGS_SF_MOD_REG_ONLY, 0, { .OpDispatch = &IR::OpDispatchBuilder::Bind<&IR::OpDispatchBuilder::WriteSegmentReg, IR::OpDispatchBuilder::Segment::GS> } },
  },
}};
```

Default table:
```cpp
constexpr std::array<X86InstInfo, 512> SecondInstGroupOps = {{
  {"SLDT", TYPE_INVALID, FLAGS_MODRM | FLAGS_SF_MOD_DST, 0, { .OpDispatch = nullptr } }, // 0
  {"STR", TYPE_INVALID, FLAGS_MODRM | FLAGS_SF_MOD_DST, 0, { .OpDispatch = nullptr } }, // 1
  <...>
  { nullptr, TYPE_ARCH_DISPATCHER, FLAGS_NONE, 0, { .Indirect = Arch_LUT_Table_SecondInstGroupOps[ENTRY_264] } }, // 264
  { nullptr, TYPE_ARCH_DISPATCHER, FLAGS_NONE, 0, { .Indirect = Arch_LUT_Table_SecondInstGroupOps[ENTRY_265] } }, // 265
  { nullptr, TYPE_ARCH_DISPATCHER, FLAGS_NONE, 0, { .Indirect = Arch_LUT_Table_SecondInstGroupOps[ENTRY_266] } }, // 266
  { nullptr, TYPE_ARCH_DISPATCHER, FLAGS_NONE, 0, { .Indirect = Arch_LUT_Table_SecondInstGroupOps[ENTRY_267] } }, // 267
  <...>
  ```
  
The indirect table can't be generated by a constexpr lambda and then consumed by another constexpr lambda generating the full table. This becomes a bear with anything beyond a handful of instruction handlers.

The frontend deocder even though it isn't interesting.
```cpp
  if (Info->Type == FEXCore::X86Tables::TYPE_ARCH_DISPATCHER) [[unlikely]] {
    return NormalOp(&Info->OpcodeDispatcher.Indirect[BlockInfo.Is64BitMode ? 1 : 0], Op);
  }
  ```

neobrain:
Thanks for the writeup! I'm only half following, perhaps since `ARCH_DISPATCHER` and `Indirect` aren't actually wired up to anything yet in the change itself, so it's tricky to see how the tables you're quoting above fit together (and why the same couldn't be done with `constexpr`).

> The indirect table can't be generated by a constexpr lambda and then consumed by another constexpr lambda generating the full table

What's the issue here?

> A bit more with the initial implementation hammering through.

Perhaps a smaller, isolated test case would be better suited for review than "hammering through" things? Piecing together the relevant transformations here from 2,500 changed lines of code is rather involved.

Sonicadvance1:
> Thanks for the writeup! I'm only half following, perhaps since `ARCH_DISPATCHER` and `Indirect` aren't actually wired up to anything yet in the change itself, so it's tricky to see how the tables you're quoting above fit together (and why the same couldn't be done with `constexpr`).
> 
> > The indirect table can't be generated by a constexpr lambda and then consumed by another constexpr lambda generating the full table
> 
> What's the issue here?

You can't reference a constexpr value that is generated through a consteval lambda from another consteval lambda creating a constexpr value.

> 
> > A bit more with the initial implementation hammering through.
> 
> Perhaps a smaller, isolated test case would be better suited for review than "hammering through" things? Piecing together the relevant transformations here from 2,500 changed lines of code is rather involved.

That's why I'm doing single tables at a time. Each table is the isolated test case. hammer hammer yo.

neobrain:
> You can't reference a constexpr value that is generated through a consteval lambda from another consteval lambda creating a constexpr value.

Do you have a code example? Seems like this would very obviously work since constexpr data (either global vars or parameters) can always be used by lambdas without breaking constexpr-ness, so I must be missing something between all the constevals.

Sonicadvance1:
Implementation moved to #4805

PR #4804, opened 2025-08-25
Github: Force FEX portable config

Sonicadvance1:
Noticed some runners had generated a .fex-emu/Config.json, run the tests in portable mode to ensure they don't pick up a spurious config option.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -13,6 +13,7 @@ env:
   BUILD_TYPE: Release
   CC: clang
   CXX: clang++
+  FEX_PORTABLE: 1
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/ccpp.yml
```diff
@@ -136,6 +137,8 @@ jobs:
     - name: FEXLinuxTests
       working-directory: ${{runner.workspace}}/build
       shell: bash
+      env:
+        FEX_PORTABLE: 0
```
neobrain, line 142: Probably merits a comment similar to the one in the PR description

Sonicadvance1: Added a comment.

PR #4805, opened 2025-08-26
FEXCore: Move most instruction tables to be constexpr

Sonicadvance1:
No generator script as requested.

Only AVX table remaining.


Sonicadvance1:
Implementation note from bloaty:
```bash
ryanh@ryanh-TR5000:/mnt/Work/Work/work/FEXNew/Build$ /mnt/Work/Projects/bloaty/Build/bloaty ../Build_Radxa/FEXPost -- ../Build_Radxa/FEXPre
    FILE SIZE        VM SIZE
 --------------  --------------
  +320%  +499Ki  +320%  +499Ki    .data.rel.ro
   +36%  +130Ki   +36%  +130Ki    .rela.dyn
  +0.0%     +84  +0.0%     +88    .eh_frame
  [NEW]     +48  [NEW]     +48    [LOAD #2 [R]]
  [NEW]     +20  [ = ]       0    [Unmapped]
  +0.7%      +4  +0.7%      +4    .gnu.version_r
  -0.0%      -4  -0.0%      -4    .gcc_except_table
  -0.0%      -4  -0.0%      -4    .gnu.hash
  -0.1%      -4  -0.1%      -4    .gnu.version
  -0.0%      -8  -0.0%      -8    .rodata
  [ = ]       0 -40.0%     -16    [LOAD #5 [RW]]
  -0.1%     -24  -0.1%     -24    .dynsym
  -0.1%     -28  -0.1%     -28    .eh_frame_hdr
  -0.1%     -68  -0.1%     -68    .dynstr
 -20.0%    -232 -20.0%    -232    .got
  [ = ]       0 -78.4% -1.39Ki    .relro_padding
  -0.1% -4.03Ki  -0.1% -4.03Ki    .text
 -50.9%  -409Ki -50.9%  -409Ki    .data
  +4.2%  +215Ki  +2.3%  +214Ki    TOTAL
  ```
Because an x87 table is now duplicated, the executable is slightly larger on the total, but 409K moved from data over to read-only which is pretty good. Not a goal of the PR, just interesting as a datapoint still.

File: FEXCore/Source/Interface/Core/Frontend.cpp [hunk 1]
```diff
@@ -315,6 +323,12 @@ void Decoder::DecodeModRM_64(X86Tables::DecodedOperand* Operand, X86Tables::ModR
 }
 
 bool Decoder::NormalOp(const FEXCore::X86Tables::X86InstInfo* Info, uint16_t Op, DecodedHeader Options) {
+  if (Info->Type == FEXCore::X86Tables::TYPE_ARCH_DISPATCHER) [[unlikely]] {
+    // Dispatcher Op.
+    // TODO: Move this in to `NormalOpHeader`, Dispatch tables have a bug currently where some subtables don't inherit flags correctly.
```
neobrain, line 328: What is the bug? Does it need to be fixed before merging? If not, how can it be reproduced?

Comments mentioning a quirk should have a clear reference or explanation that defines what quirk needs to be addressed before the comment can be removed.

Sonicadvance1: Removing the code and running our ASM tests is enough to catch the issue. It doesn't need to be fixed first, mostly because it's going to be involved walking through all of our tables and ensuring flags are inherited correctly.

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -231,6 +231,7 @@ enum InstType {
   TYPE_X87 = TYPE_INST,
   TYPE_INVALID,
   TYPE_COPY_OTHER,
+  TYPE_ARCH_DISPATCHER,
```
neobrain, line 236: Should probably have a comment explaining dispatcher behavior (indirection to another table given by OpDispatchPtrWrapper::Indirect) and intended use case (32-/64-bit switching).

neobrain (approved):
Thanks for looking into this approach again, glad you found a way to make it work!

FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
Sonicadvance1: Added a comment.

FEXCore/Source/Interface/Core/Frontend.cpp
Sonicadvance1: Extended the comment to explain running the asm tests is enough to reproduce the issue.

PR #4806, opened 2025-08-27
X86Tables: Convert AVX tables to constexpr

Sonicadvance1:
One set of tables for 128-bit and one set of tables for 256-bit.
This one took a bit longer since I needed to convert a few handlers over
to `Bind`. With this all of our x86 tables are constexpr so they end up
in RO mapped memory which is great.


lioncash (approved):
Noice

PR #4807, opened 2025-08-27
Remove visibility on functions that is no longer necessary

Sonicadvance1:
These used to be required because we leaked IR semantics across the API boundary. Since we no longer do, we can remove the visibility from these.

Other bits also can use hidden visibility.

### DIFF
--- a/Source/Common/CMakeLists.txt
+++ b/Source/Common/CMakeLists.txt
@@ -22,3 +22,9 @@ endif()
 add_library(${NAME} STATIC ${SRCS})
 target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json FEXHeaderUtils range-v3::range-v3)
 target_include_directories(${NAME} PRIVATE ${CMAKE_BINARY_DIR}/generated)
+
+set_target_properties(${NAME} PROPERTIES
+  C_VISIBILITY_PRESET hidden
+  CXX_VISIBILITY_PRESET hidden
+  VISIBILITY_INLINES_HIDDEN TRUE
+)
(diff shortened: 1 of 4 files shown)


Source/Common/CMakeLists.txt
```diff
@@ -22,3 +22,7 @@ endif()
 add_library(${NAME} STATIC ${SRCS})
 target_link_libraries(${NAME} FEXCore_Base cpp-optparse tiny-json FEXHeaderUtils range-v3::range-v3)
 target_include_directories(${NAME} PRIVATE ${CMAKE_BINARY_DIR}/generated)
+
+set_target_properties(${NAME} PROPERTIES C_VISIBILITY_PRESET hidden)
```
lioncash: Should these be coalesced like:

```cmake
set_target_properties(${NAME} PROPERTIES
  C_VISIBILITY_PRESET hidden
  CXX_VISIBILITY_PRESET hidden
  VISIBILITY_INLINES_HIDDEN TRUE
)
```

or something of the sort, so any properties added in the future just requires adding only the key and value in a convenient way

Sonicadvance1: Switched, forgot they could be combined like that.

lioncash (approved):

neobrain:
Did you verify library forwarding is still functional after this change? It relies on symbol visibility for some things, though in theory it should already be set explicitly for those.

Sonicadvance1:
> Did you verify library forwarding is still functional after this change? It relies on symbol visibility for some things, though in theory it should already be set explicitly for those.

Indeed I did, it was still working fine.

PR #4808, opened 2025-08-29
Dispatcher: Centralize SignalDelegator config creation

lioncash:
Noticed that we had a few things open-coded related to the Dispatcher and SignalDelegator that could be handled a little nicer. Since all of the information for the delegator config comes from the Dispatcher directly, we can have it return a config instance directly upon request, which tidies up InitCore() a little with regards to handling dispatcher/signal specifics.

A knock-on effect of all of this is that the members of the Dispatcher can all be non-public in some form (which also revealed an unused member), which makes reasoning about a few things a tad bit nicer, since everything is local to the class in some form now.


Sonicadvance1 (approved):
Nice nice.

PR #4810, opened 2025-08-29
FileManagement: Minor interface cleanup

lioncash:
We just had a few member functions that could be marked as const here, since they don't modify internal state.


Sonicadvance1 (approved):
Neat

PR #4813, opened 2025-09-02
unittests: Adds missing movsxd edge case

Sonicadvance1:
nasm doesn't let you encode all of these but we had already supported them in FEX.

It's a bit of a weird edge case behaviour, since it's supposed to be sign extended a 32-bit value in to a 64-bit register but with prefixes (or lack of) it can be a move with zero-extend or a 16-bit insert.

Make sure we're actually testing these edge cases.


Sonicadvance1:
gvisor test failed purely from #4812

PR #4817, opened 2025-09-02
CodeCache: Introduce new interfaces

neobrain:
Mainly renames and rerranges existing interfaces into a more coherent structure, whereas previously functions related to code caching were scattered across different modules. Furthermore, function signatures are adapted for better type-safety and for purposes of self-documentation. The term "AOTIR" is purged from the code base in favor of "code cache".

Functional changes:
* `SyscallHandler::LookupExecutableFileSection` (previously LookupAOTIRCacheEntry) will return "no match" for virtual memory ranges without an associated Resource.

Quick reference of renamed functions:
* `ContextImpl::FetchAOTIRCacheEntry` => `CodeCache::LoadData`
* `ContextImpl::FinalizeAOTIRCache` => `CodeCache::SaveData`
* `AOTIR::AOTIRCacheEntry` => `CodeCache::ExecutableFileInfo`
* `AOTIR::AOTIRCacheEntryLookupResult` => `CodeCache::ExecutableFileSectionInfo`
* `AOTIRCaptureCache::LoadAOTIRCacheEntry` => `ExecutableFileSectionInfo` (constructor)
* `AOTIRCaptureCache::UnloadAOTIRCacheEntry` => `~ExecutableFileSectionInfo` (destructor)
* `SyscallHandler::LookupExecutableFileSection` => `SyscallHandler::LookupAOTIRCacheEntry`


FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: MIT
```
neobrain, line 1: Review hint: This file was moved from AOTIR.cpp. I isolated the move in git history, but virtually every line in this file changed so GitHub doesn't show it as a move.

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
```diff
@@ -354,6 +354,10 @@ inline fextl::string GetFilename(const fextl::string& Path) {
   return PathToString(std::filesystem::path(Path).filename());
 }
 
+inline fextl::string GetFilename(std::string_view Path) {
+  return PathToString(std::filesystem::path(Path).filename());
```
neobrain: @bylaws Could you briefly weigh in on this: Would it be safe to return a sub-`string_view` of the input path here (for consistency with Linux), or are there weird cases for which `filename()` returns a filename that's not a substring of the input on Windows?

```suggestion
  auto Filename = PathToString(std::filesystem::path(Path).filename());
  return Path.substr(Path.size() - Materialized.size());
```

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -803,13 +803,12 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
...
-  Thread->CPUBackend->ClearRelocations();
-
-  if (IRCaptureCache.PostCompileCode(Thread, CompiledCode.BlockBegin, GuestRIP, StartAddr, Length, DebugData.get())) {
-    // Early exit
-    return (uintptr_t)CodePtr;
+  if (!CodeCache.IsGeneratingCache) {
+    Thread->CPUBackend->ClearRelocations();
   }
 
+  CodeCache.PostCompileCode(*Thread, CompiledCode.BlockBegin, GuestRIP, *DebugData.get());
```
bylaws: Might we want to pull this inline? the gdb and debugdata stuff isn't really logically tied to the code cache

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -0,0 +1,61 @@
...
+      }
+
+      if (CTX.Config.GDBSymbols()) {
+        GDBJITRegister(MappedSection->FileInfo, MappedSection->FileStartVA, GuestRIP, (uintptr_t)CodePtr, DebugData);
+      }
+    }
+  }
+}
+
+fextl::string CodeCache::ComputeCodeMapId(std::string_view Filename) {
```
bylaws: I feel this belongs to the frontend, at most we could have the frontend provide some opaque config blob or similar

FEXCore/include/FEXCore/Core/CodeCache.h
```diff
@@ -0,0 +1,67 @@
...
+namespace Core {
+  struct InternalThreadState;
+} // namespace Core
+
+namespace HLE {
+  struct SourcecodeMap;
+} // namespace HLE
+
+// Generic information associated with an executable file.
+// Lifetime managed by AbstractCodeCache
```
bylaws: This doesn't seem to hold true?

FEXCore/Source/Interface/Core/JIT/Arm64Relocations.cpp
```diff
@@ -81,54 +81,60 @@ void Arm64JITCore::InsertGuestRIPMove(ARMEmitter::Register Reg, uint64_t Constan
...
 
-bool Arm64JITCore::ApplyRelocations(uint64_t GuestEntry, uint64_t CodeEntry, uint64_t CursorEntry, size_t NumRelocations,
-                                    const char* EntryRelocations) {
-  size_t DataIndex {};
-  for (size_t j = 0; j < NumRelocations; ++j) {
-    const FEXCore::CPU::Relocation* Reloc = reinterpret_cast<const FEXCore::CPU::Relocation*>(&EntryRelocations[DataIndex]);
-    LOGMAN_THROW_A_FMT((DataIndex % alignof(Relocation)) == 0, "Alignment of relocation wasn't adhered to");
-
-    switch (Reloc->Header.Type) {
+bool Arm64JITCore::ApplyRelocations(uint64_t GuestEntry, std::span<std::byte> Code, std::span<const FEXCore::CPU::Relocation> Relocations) {
```
bylaws, line 84: I suppose a future PR will pull this out of JITCore and switch it to using it's own emitter?

bylaws (changes requested):

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: For context, this API will change to be independent from any runtime configuration. It will only generate a globally unique identifer for a binary, likely using the ELF debug id (and similar for PE if anything like that exists) with fallback using the hashed filename + headers. Any JIT configuration would factor into a separate "code cache id", so that we can support a one-to-many relationship between each code map and code caches generated from it. (The *code cache id* will likely be frontend-specific indeed.)

Over the last few days I realized we do want to compute the code map id even in WoA deployments, since there should be a viable path for code caching on general Linux devices that use WoA. For that we'll want support for global cache directories (instead of game-local ones), so files with the same filename need to be distinguishable.

As a nice bonus this would also enable code maps to be reused when switching a particular game configuration from WoA to Linux-FEX (or vice-versa), instead of having to gather a completely separate code map from scratch.

FEXCore/Source/Interface/Core/Core.cpp
neobrain: Yeah, makes sense.

FEXCore/Source/Interface/Core/JIT/Arm64Relocations.cpp
neobrain: Indeed! This will move to the CodeCache implementation since that's the only place where it's used.

FEXCore/include/FEXCore/Core/CodeCache.h
neobrain: Correct, this is an artifact of a previous iteration of this PR.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -0,0 +1,61 @@
...
+  auto base_filename = FHU::Filesystem::GetFilename(Filename);
+  if (!base_filename.empty()) {
+    auto filename_hash = XXH3_64bits(Filename.data(), Filename.size());
+
+    return fextl::fmt::format("{}-{:016x}-{}{}{}", base_filename, filename_hash,
```
neobrain: Accidentally slipped in a change here:
```suggestion
    return fextl::fmt::format("{}-{}-{}{}{}", base_filename, filename_hash,
```

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Yeah, I think that even in such cases binary parsing should be kept to the frontend. Codemaps on WoA are done with PIDs and generated caches already have a unique name using headers roughly as you described, just in a bundle rather than individual libraries

neobrain: If this is done consistently with how Linux would do it, why does the code map id need to be computed in the frontend? Seems like consistency is much easier to achieve (and guarantee going forward) by sharing the code.

bylaws: The codemap ID on windows is generated by reading headers from the mapped PE, using windows specific structs. 
https://github.com/bylaws/FEX/blob/arm64eccc/Source/Windows/Common/ImageTracker.h#L71

Not really sure what part of this would belong in fexcore, and since the frontend handles loading the cache etc it doesn't seem to me like it is a concept that needs to be handled there. Perhaps Source/Common if there are things that can be shafed

neobrain: > The codemap ID on windows is generated by reading headers from the mapped PE

That's happening on Linux as well for PE files, so we can use the same PE parsing + code map computation code in both deployments. (Doesn't make sense to use something else in WoA when we have our own code anyway)

bylaws: My thought is that with it being like 4 lines on windows as is there's no point trying to use our own fex side parsing in this case. It's simple enough that sharing code probably will increase complexity. Note that caching on windows is for an entire process, not per library, the game exe codemap will contain entries for all libraries loaded by the game etc so I don't think it would necessarily be portable with Linux.

bylaws: (this is done for simplicity, and the fact that libraries are basically never shared on windows and makes per exe configs easy, I don't think the complexity of doing anything else makes sense)

bylaws: I think that perhaps a Source/Common function to generate a unique pe id given x header fields would be best at least. Basically GetUniqueId as linked but moved

neobrain: > My thought is that with it being like 4 lines on windows as is there's no point trying to use our own fex side parsing in this case

PE parsing is pretty trivial even when done manually. Using consistent implementations ensures testing covers both deployments and it ensures behavior stays consistent (instead of slowly drifting apart by accident in later refactors).

We *really* shouldn't have the code caching implementations for Linux and WoA drift apart more than absolutely necessary. This seems like something that can be trivially shared for little to no extra work. Yes, WoA will be slightly more involved, but only in so far that it needs to refer to shared FEX code instead of using native Windows structs.

> game exe codemap will contain entries for all libraries loaded by the game etc so I don't think it would necessarily be portable with Linux

That's what's happening on Linux as well. Code maps aren't split into individual libraries until FEXServer processes them, so code maps generated under WoA could (in principle) trivially be reused by a later Linux run.

> Note that caching on windows is for an entire process, not per library,

That's why the *code cache id* (not the code map id) would be frontend-specific.

<hr>

FWIW it's worth repeating CodeCache.cpp is currently just a rename of AOTIR.cpp with new function names, so perhaps this discussion is a bit abstract without a final implementation.

bylaws: Agreed in this being a bit abstract, I think what I have pushed for WoA is very close to how I'd want things to work in general and that sharing codemaps probably isn't too useful. For what is literally a 3 line function though think it makes sense to drop this for now and then add back in an appropriate location when there is a more concrete implementation of both to review

FEXHeaderUtils/FEXHeaderUtils/Filesystem.h
neobrain: I'm going to assume this is fine and will adopt this approach.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Removed the code for now as it's unused, but this conversation will come again once code map writing lands. I don't see a convincing argument to diverge here on the WoA side, but I could be swayed to move code map id generation to the frontend if I saw a working prototype that produces FEX-Linux-compatible code maps (and vice-versa) without user intervention.

Sonicadvance1:
Just pending @bylaws requested change sign-off now.

PR #4821, opened 2025-09-02
Frontend: Fix tailcall handling when mono hacks are enabled

bylaws:
ValidateCode didn't work prior to this on arm64ec, I chose to rewrite it and clean things up instead of just fixing the incorrect register usage (w2 vs TMP2)

### DIFF
--- a/FEXCore/Source/Interface/IR/IRDumper.cpp
+++ b/FEXCore/Source/Interface/IR/IRDumper.cpp
@@ -274,6 +274,10 @@ static void PrintArg(fextl::stringstream* out, const IRListView*, FEXCore::IR::B
   }
 }
 
+static void PrintArg(fextl::stringstream* out, const IRListView*, const std::array<uint8_t, 0x10>& Arg) {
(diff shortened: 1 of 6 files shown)


FEXCore/Source/Interface/IR/IRDumper.cpp
```diff
@@ -274,6 +274,13 @@ static void PrintArg(fextl::stringstream* out, const IRListView*, FEXCore::IR::B
   }
 }
 
+static void PrintArg(fextl::stringstream* out, const IRListView*, const std::array<uint8_t, 0x10>& Arg) {
+  for (auto byte : Arg) {
+    *out << std::hex << std::setfill('0') << std::setw(2) << (unsigned int)byte;
+  }
+  *out << std::dec;
```
neobrain:
```suggestion
  *out << fextl::fmt::format("{:02x}", fmt::join(Arg, ""));
```

Sonicadvance1 (approved):

PR #4822, opened 2025-09-02
Dispatcher: Move the EC entrypoint check prior to the L1 lookup

bylaws:
Profiling shows this is by far the hottest path, as every DX call will hit this (for arm64ec functions in a vtable). L2 hits are much rarer and L1 almost never hits the dispatcher since that is looked up inline in most cases.


bylaws:
Dropped the L1 lookup entirely for a marginal geekbench improvement of .5%

Sonicadvance1 (approved):

PR #4825, opened 2025-09-02
Frontend: Force acq/rel semantics for known Unity ringbuffer offsets

bylaws:
Unity games crash with TSO disabled due to the SPSC GfxDevice ThreadedStreamBuffer read/write pointer updates missing acq/rel semantics. Rather than attempting to use heuristics to match cases like this, which could be overzealous and hit more accesses than necessary, just target the problem directly as this is consist across 32/64 bit Unity versions for at least the past 10 years. Gate this behind the existing Unity mono hacks to avoid false-positives in non-Unity games.

Draft - do we want this? should this be some sort of config for volatile displacements? this gives a 1.5-3x FPS improvement vs TSO enabled or any reasonable heuristic that could catch all of these without hardcoding the offsets.


bylaws:
Unmarked as draft since this has been flawless in my testing

Sonicadvance1 (approved):

Sonicadvance1:
Might as well as have fun with the hack toggle.

PR #4830, opened 2025-09-03
Core: Drop support for TSO auto migration

neobrain:
Originally added in https://github.com/FEX-Emu/FEX/pull/1752, TSO auto migration is a heuristic to disable TSO emulation until the application creates a second thread. This heuristic doesn't mesh well with code caching (see below), so I'm proposing we remove it entirely.

There are some other reasons why dropping support isn't too bad:
* The general impact of TSO emulation has been greatly reduced since 2022 (see e.g. https://github.com/FEX-Emu/FEX/pull/2760 and https://github.com/FEX-Emu/FEX/pull/3511)
* For specific applications, AppConfigs and TSO heuristics can be a good alternative (#4728, #4825)
* WoA deployments never used TSO auto migration and instead can use volatile metadata (#4378)
* Ever since #4479 was merged, we've been soft-leaking any CodeBuffer memory allocated before TSO auto migration

Specifically the interaction with #4479 also causes TSO auto migration to re-compile any code executed by the first application thread, which can have a minor loading time impact on the vast majority of applications that use multiple threads.

## Conflicts with code caching

The TSO heuristic carries design overhead when considering its interaction with code caching. The suggested resolution in the [TSO auto-migration PR](https://github.com/FEX-Emu/FEX/pull/1752) itself was to disable TSO automigration when code caching is enabled. Since there's no reason *not* to enable code caching unconditionally, this would have the same ultimate effect as removing the heuristic entirely.

Here are some specific problems that would need consideration if we keep the heuristic:

### Should FEX generate a non-TSO cache "just in case"?

Any benefit of TSO auto-migration would be defeated by loading a code cache compiled with TSO emulation enabled. So should FEX always generate two version of the same cache (one with TSO, one without)? It's dubious whether code caching brings any tangible benefit to the type of application that would benefit from TSO auto-migration anyway, so compiling two caches seems wasteful.

We could record whether an application uses multiple threads in a code map, but again this doesn't seem worth the increase in design complexity. Notably this also wouldn't work for script interpreters (e.g. python), where the effectiveness of TSO migration depends on the script being executed.

### Runtime interaction of TSO auto-migration with code cache loading

What should FEX do if a cache is present but TSO hasn't been auto-enabled yet? If we don't load the cache, we won't be able to retroactively do so after TSO auto-migration; if we do load the cache, the LookupCache clear triggered during auto-migration will throw it away.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -395,14 +395,6 @@
           "This is required to ensure a split-lock doesn't tear inside the process"
         ]
       },
-      "TSOAutoMigration": {
-        "Type": "bool",
-        "Default": "true",
-        "Desc": [
-          "Automatically enables TSO when shared memory is used.",
-          "Should work without issues in most cases."
-        ]
-      },
       "VolatileMetadata": {
         "Type": "bool",
         "Default": "true",
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -154,8 +154,6 @@ public:
     return CodeInvalidationMutex;
   }
 
-  void MarkMemoryShared(FEXCore::Core::InternalThreadState* Thread) override;
-
   void ConfigureAOTGen(FEXCore::Core::InternalThreadState* Thread, fextl::set<uint64_t>* ExternalBranches, uint64_t SectionMaxAddress) override;
 
   bool IsAddressInCodeBuffer(FEXCore::Core::InternalThreadState* Thread, uintptr_t Address) const override;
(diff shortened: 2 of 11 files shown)


neobrain:
Naturally this incurs a big InstCountCI change since InstCountCI is a single-threaded workload and hence effectively ran with TSO disabled. We could explicitly disable TSO to leave the reference data unchanged, but it seems reasonable to keep it enabled to have an environment closer to production. Opinions?

Sonicadvance1:
I'm going to do some testing on this. One of the big things this feature helped when it was added was pressure-vessel startup times, but it was on Orin which is bugged hardware so maybe it's not really an issue.

alyssarosenzweig:
Personally I'd default off TSO for instcountci, we already have _TSO.json files to test those paths and the extra nops make it harder to see what's going on, but I also don't get a say =)

unittests/InstructionCountCI/AFP/SVE256/Secondary_REP.json
```diff
@@ -31,12 +31,13 @@
       ]
     },
     "cvtsi2ss xmm0, qword [rax]": {
-      "ExpectedInstructionCount": 2,
+      "ExpectedInstructionCount": 3,
```
Sonicadvance1: Can you go through these InstCountCI files and add the DisabledHostFeatures option `TSO`.
We don't care about the TSO-ness of these instcountCI tests.

neobrain: >  We don't care about the TSO-ness of **these** instcountCI tests.

That's which tests specifically? If you mean all of them that don't currently enable TSO, is there any difference to disabling TSO altogether by default as I suggested above?

Sonicadvance1: All the instcountCI tests that don't explicitly enable it, yes.

neobrain:
Disabled TSO in instcountci by default, so there is no longer any diff in the codegen jsons.

PR #4835, opened 2025-09-04
Fix quiet and signalling nan propagation

pmatos:
Fix for __builtin_issignaling() test of SPEC2017 classify test.


pmatos:
> This seems relatively costly, could we get a config option for the 64-bit side at least so we can toggle it when testing performance. Maybe a new strenum like HostFeatures but x87 flags?

Would you be happy with a StrictReducedPrecision that would enable this and a couple of other features I am working on together with ReducedPrecision?

Sonicadvance1:
> > This seems relatively costly, could we get a config option for the 64-bit side at least so we can toggle it when testing performance. Maybe a new strenum like HostFeatures but x87 flags?
> 
> Would you be happy with a StrictReducedPrecision that would enable this and a couple of other features I am working on together with ReducedPrecision?

Sure, that would work.

pmatos:
> > > This seems relatively costly, could we get a config option for the 64-bit side at least so we can toggle it when testing performance. Maybe a new strenum like HostFeatures but x87 flags?
> > 
> > 
> > Would you be happy with a StrictReducedPrecision that would enable this and a couple of other features I am working on together with ReducedPrecision?
> 
> Sure, that would work.

WDYT?

FEXCore/Source/Interface/Core/Interpreter/Fallbacks/F80Fallbacks.h
```diff
@@ -77,6 +78,11 @@ struct OpHandlers<IR::OP_F80CVTTO> {
   FEXCORE_PRESERVE_ALL_ATTR static VectorRegType handle8(uint16_t FCW, double src, FEXCore::Core::CpuStateFrame* Frame) {
     FEXCORE_PROFILE_INSTANT_INCREMENT(Frame->Thread, AccumulatedFloatFallbackCount, 1);
     ScopedSoftFloatState State {FCW, Frame};
+    FEX_CONFIG_OPT(ReducedPrecisionMode, X87REDUCEDPRECISION);
```
Sonicadvance1: Would be nice if these were looking up the config from `Frame->Thread->CTX->Config` or something. While these are relatively lightweight because of the optimization to no longer do string conversions for every look up, they're still a map search each time. Do a bit of pointer dereferencing instead.

pmatos: Sure - makes sense.

PR #4837, opened 2025-09-04
X87: Better cache intermediate results in the X87 slowpath

bylaws:


bylaws:
Think I can probably get some more gains here by catching AbridgedFTW

FEXCore/Source/Interface/IR/IR.json
```diff
@@ -505,6 +505,16 @@
           "!($BaseOffset >= offsetof(Core::CPUState, xmm.avx.data[0]) && $BaseOffset < offsetof(Core::CPUState, xmm.avx.data[16])) && \"Can't StoreContextIndexed to XMM\""
         ]
       },
+      "GPR = FormContextAddress OpSize:#Size, GPR:$Index, u32:$Stride": {
+        "Desc": ["Forms an address into the context structure indexed by SSA value",
+                 "Dest = Ctx + Index * Stride",
+                 "This allows backends to compute the address once and reuse it for multiple memory operations"
+                ],
```
lioncash, line 513: Perhaps this should also include that the stride needs to be a power of 2

bylaws:
> Think I can probably get some more gains here by catching AbridgedFTW

Did this, further couple hundred inst improvement

bylaws:
Also cached values, shaving another 3000 insts off instcountci

bylaws:
This I should be able to do something similar for valid tag management here. Can defer and maintain a bitmask that we store rotated at the end

bylaws:
I think this is everything :)

pmatos:
@bylaws Nice! At first sight it looks great. Let me run a few tests as well.

bylaws:
There is some bug with invalid flag handling that makes it different to what is done currently, hence draft rn

pmatos:
> There is some bug with invalid flag handling that makes it different to what is done currently, hence draft rn

Where did you see this? Do you have a test for it?

bylaws:
> > There is some bug with invalid flag handling that makes it different to what is done currently, hence draft rn
> 
> Where did you see this? Do you have a test for it?

Oh just instcountci, haven't investigated

bylaws:
nvm, the instcountci change is a fix to a previous bug (the popcount check was `if (popcount())` rather than `if (popcount() == 1)`

bylaws:
Benchmarking this, MGS V is perhaps >10% faster if i limit the clock on my t14s to get it under the 60fps cap.

bylaws:
Ontop of #4859 to ensure correctness here.

unittests/InstructionCountCI/x87_f64.json
```diff
@@ -1833,21 +1761,18 @@
...
     },
     "fld st7": {
-      "ExpectedInstructionCount": 15,
+      "ExpectedInstructionCount": 12,
       "Comment": [
         "0xd9 11b 0xc7 /0"
       ],
       "ExpectedArm64ASM": [
         "ldrb w20, [x28, #1019]",
-        "add w21, w20, #0x7 (7)",
```
Sonicadvance1, line 1842: Flagging this one because I need to take a closer look at it. Seems like it is now storing the resulting add rather than a sub but maybe that's expected.

bylaws: This is modulo 8 arithmetic, so adding 7 = sub 1

PR #4870, opened 2025-09-11
FEXBash: Remove ArgLoader usage

Sonicadvance1:
ArgLoader is going to get removed. So delete this menial usage of it.

### DIFF
--- a/Source/Tools/FEXBash/FEXBash.cpp
+++ b/Source/Tools/FEXBash/FEXBash.cpp
@@ -6,9 +6,7 @@ desc: Launches bash under FEX and passes arguments via -c to it
 $end_info$
 */
 
-#include "Common/ArgumentLoader.h"
-
-#include <FEXCore/Config/Config.h>
+#include <FEXCore/fextl/fmt.h>
 
 #include <filesystem>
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXBash/FEXBash.cpp
```diff
@@ -6,23 +6,22 @@ desc: Launches bash under FEX and passes arguments via -c to it
...
 #include <filesystem>
 #include <string>
 #include <unistd.h>
 #include <vector>
 
 int main(int argc, char** argv, char** const envp) {
-  FEX::ArgLoader::ArgLoader ArgsLoader(FEX::ArgLoader::ArgLoader::LoadType::WITHOUT_FEXLOADER_PARSER, argc, argv);
-  auto Args = ArgsLoader.Get();
+  // Skip argv[0].
+  const bool ArgCount = argc - 1;
```
lioncash: Should this be an int? (or a sanitized uint32_t given it's used with a size_t loop counter below)

Sonicadvance1: oop, good catch. Fixed.

PR #4873, opened 2025-09-11
FEXRootFSFetcher: Removes argloader usage

Sonicadvance1:
Another case that ArgLoader was misused.

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/Main.cpp
+++ b/Source/Tools/FEXRootFSFetcher/Main.cpp
@@ -1108,8 +1108,7 @@ bool ExtractEroFS(const fextl::string& Path, const fextl::string& RootFS, const 
 } // namespace UnSquash
 
 int main(int argc, char** argv, char** const envp) {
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXRootFSFetcher/Main.cpp
```diff
@@ -1108,8 +1108,7 @@ bool ExtractEroFS(const fextl::string& Path, const fextl::string& RootFS, const
 } // namespace UnSquash
 
 int main(int argc, char** argv, char** const envp) {
-  auto ArgsLoader = fextl::make_unique<FEX::ArgLoader::ArgLoader>(FEX::ArgLoader::ArgLoader::LoadType::WITHOUT_FEXLOADER_PARSER, argc, argv);
```
lioncash, line 1111: Should the argloader include be dropped, or are there still dependencies that are going to be stripped out in the final removal?

Sonicadvance1: Header still needs to around for destructor definition for now. Can get removed once `FEX::Config::LoadConfig` completely drops its ArgLoader argument.

PR #4875, opened 2025-09-11
FEXInterpreter: Remove FEXLoader

Sonicadvance1:
Doesn't /quite/ remove the ArgumentLoader because it is intertwined with
LinuxEmulation in an annoying way that will take another step to remove.


Sonicadvance1:
Rebased.

PR #4878, opened 2025-09-11
unittests/ASM: Implement modrm OOB tests

### LINKED ISSUE #4769: Create ModRM OOB unittests
As seen in #4768, we had some instructions doing OOB loads on some instructions.

We should generate some unittests that test all modrm supporting instructions to ensure they don't read or write past the expected size of the instruction.

Should test:
- Doesn't read past the end in to unmapped page
- Doesn't write past the end in to an unmapped page
- Doesn't read /before/ in to an unmapped page (some instructions are tricky here)
- Doesn't write /before/ in to an unmapped page

Shouldn't be too hard, there's only a thousand or so mnemonics that support modrm, will just be tedious. Some nasm macros will help significantly here.

Really want to make sure that we squash all of these since it has happened multiple times over the years.

Sonicadvance1:
We've hit this multiple times that certain instructions either read or write too much data. Implement some basic tests that hit EVERY instruction that understands modrm to see if it does OOB.

Simple helpers, just hammer it out and if it reaches the end then it is a success.

Tables:

- [X] Primary
- [X] PrimaryGroup
- [X] Secondary
- [X] Secondary_REP
- [X] Secondary_REPNE
- [X] Secondary_OpSize
- [X] SecondaryModRM
- [X] SecondaryGroup
- [X] DDD
- [X] H0F38
- [X] H0F3A
- [X] X87
- [X] VEX
- [X] VEX Group

Fixes #4769

### DIFF
--- a/unittests/ASM/Disabled_Tests_Simulator
+++ b/unittests/ASM/Disabled_Tests_Simulator
@@ -107,3 +107,6 @@ Test_TwoByte/0F_31.asm
 
 # Simulator doesn't support executing a syscall
 Test_Secondary/09_F3_07.asm
+
+# Vixl sim at 256-bit vector width access too much memory with some AVX instructions
+Test_modrm_oob/VEX.asm


Sonicadvance1:
Found a real bug here. `movbe` with 16-bit memory destination is turning in to a 64-bit loadstore which is wrong.

Sonicadvance1:
Blam. All the instruction tables hammered out and even fixed the one bug with movbe.

lioncash (approved):

PR #4883, opened 2025-09-12
FEX: Rebrand FEXInterpreter as FEX

Sonicadvance1:
Still creates a copy of FEXInterpreter from FEX for downstream projects
to have some time to get off the old name. Creating a symlink is kind of
a pain in cmake so just doing an install copy is easy.


Source/Common/FEXServerClient.cpp
```diff
@@ -121,7 +121,7 @@ fextl::string GetServerMountFolder() {
     // This is due to pressure-vesssel being a chroot environment.
     // It by default maps the host-filesystem to `/run/host/` so we need to redirect.
     // After pressure-vessel is fully set up it will set the `FEX_ROOTFS` environment variable,
-    // which the FEXInterpreter will pick up on.
+    // which the FEX will pick up on.
```
neobrain: Using an article here is awkward when just referring to "FEX".
```suggestion
    // which FEX will pick up.
```

Source/Tools/FEXInterpreter/CMakeLists.txt
```diff
@@ -5,50 +5,56 @@ if (ENABLE_VIXL_SIMULATOR)
...
   RUNTIME
     DESTINATION bin
     COMPONENT Runtime
 )
 
+# Create a copy of FEX with legacy names until phased out.
+install(PROGRAMS ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/FEX
+  RENAME FEXInterpreter
+  DESTINATION bin
+  COMPONENT Runtime)
```
neobrain: Perhaps use a `LegacyRuntime` component for this? It'll still be installed with `ninja install` then, but everyone who already migrated can easily opt out of the redundant binary by installing just the "Runtime" component.

neobrain (approved):
:rocket: :rocket: :rocket: :rocket:

PR #4886, opened 2025-09-13
FEXLinuxTests: Adds a small sleep for consistency

Sonicadvance1:
Sleep a second to make this test more consistent. Sometimes the filesystem times are slightly different than wallclock times.


Sonicadvance1:
Added a comment.

PR #4889, opened 2025-09-15
SignalDelegator: Correctly save and restore x87 state. 

Sonicadvance1:
We need to read the current x87 TOP location and rotate the values when
saving and restoring the context state on Linux. Wasn't visible in the
ASM tests since we read the FEX's CPU state directly there.


lioncash (approved):

pmatos:
Ufff - nice catch. lgtm

bylaws:
You might want to extend this to fix xsave itself too in a future pr

PR #4891, opened 2025-09-16
FEX: Fixes SIGILL reporting

Sonicadvance1:
Two bug fixes here.
- We weren't reporting a correct ErrorRegister. Needs to contain
  PF_INSTR at least.
- We weren't setting siginfo->si_addr in the case of `FaultToTopAndGeneratedException`

Fixes Mafia 2 (Classic)


unittests/FEXLinuxTests/tests/signal/sigill_flags.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <sys/mman.h>
+#include <stdint.h>
+#include <cstdlib>
+#include <optional>
+
+__attribute__((naked, nocf_check)) static void SafeRet() {
+  __asm volatile(R"(
+  ret;
+  )");
+}
```
neobrain, line 14: What is this needed for?

unittests/FEXLinuxTests/tests/signal/sigill_flags.cpp
```diff
@@ -0,0 +1,80 @@
...
+    void* ptr = mmap(nullptr, 4096, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    REQUIRE(ptr != MAP_FAILED);
+    memset(ptr, 0xc3, 4096);
+    REQUIRE(mprotect(ptr, 4096, Prot.Prot) == 0);
+
+    using func_type = void (*)();
+    auto func = reinterpret_cast<func_type>(ptr);
+
+    data.reset();
+    func();
```
neobrain: Probably merits some explanations about what is going on here. I've looked at it for a few minutes and I can only guess.

Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator/GuestFramesManagement.cpp
```diff
@@ -721,17 +727,17 @@ uint64_t SignalDelegator::SetupRTFrame_ia32(FEXCore::Core::InternalThreadState*
   guest_uctx->uc.uc_stack.ss_size = GuestStack->ss_size;
 
   // Setup siginfo
+  // These three elements are in every siginfo
```
neobrain, line 745: Not sure what **three** elements this was originally referring to, but due to the move this needs to be updated.

EDIT: On second thought, was the comment previously misplaced, and it's now moved to clarify that it covers `si_signo`, `si_errno`, and `si_code`? In that case the new placement is fine of course. Removing the empty newline after `si_errno` and inserting one before the `Layout` declaration would help make this more apparent, as currently the visual grouping suggests a different grouping.

unittests/FEXLinuxTests/tests/signal/sigill_flags.cpp
Sonicadvance1: Added a comment in the signal handler that it is required for safe long-jump out through signal handler.

unittests/FEXLinuxTests/tests/signal/sigill_flags.cpp
Sonicadvance1: Added a couple comments about it being a  JIT space with only `ret` inside but with NOEXEC permissions.

Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator/GuestFramesManagement.cpp
Sonicadvance1: Yes, it's those three si_* variables. Shifted the whitespace around as suggested.

neobrain (approved):

PR #4892, opened 2025-09-18
FEXCore/JIT: Supports restarting JIT in case of encoding failure 

### LINKED ISSUE #4486: METAL GEAR RISING: REVENGEANCE Crash with multiblock enabled and x87 reduced precision disabled.
This game is usually capable of running, but when using full precision x87 and multiblock it seems like it crashes once reaching the title screen. It gets through its intro images.

| Multiblock | Reduced Precision On |  Reduced Precision Off |
| - | - | - |
| On | ✅Works | ❌Crashes |
| Off | ✅Works | ✅Works |

https://store.steampowered.com/app/235460/METAL_GEAR_RISING_REVENGEANCE

Couldn't bisect it far enough back since Multiblock was fairly sketchy a few months ago.

Sonicadvance1:
ARM64 branches have fairly small relative distances they can encode.
These can be +-1MB, or even +-32KB. The largest relative branch is
+-128MB, which we already set as an upper limit of our block JIT cache
size.

We have for a long time just compiled these without checking with the
expectation that things just happen to work. We didn't hit the asserts
so it was relatively low priority. Apparently now with Steam and a
MaxInst limit of 5000, we are now hitting an assert where we are
encoding too large of a range.

Implement support for long jumping from anywhere in the JIT for when a
long jump tries to be encoded and fails, allowing us to restart the JIT
at any moment. This is implemented as a long jump when this singular
feature could have gotten away with some sort of invasive check and
early exit path for two reasons. For one, that would be even more
invasive, effectively doing try-catch logic manually. And two, the next
step is supporting JIT buffer overflow for when our block size heuristic
fails.

This next step will mandate longjump on SIGSEGV (with cooperative
interaction with the frontend) from effectively /anywhere/ in the JIT.
One of the design goals of the CodeEmitter is that every code emission
function doesn't do a size remaining check to allow the compiler to do
some very effective optimization of emitting code blocks to memory (and
it works!).

But we lose the ability to sanely size check. When writing the emitter I
knew we were going to need to write this cooperative guard page handler,
and we're finally at a point where it needs to be done. This will be in
the next PR although.

Fixes #4486

### DIFF
--- a/CodeEmitter/CodeEmitter/BranchOps.inl
+++ b/CodeEmitter/CodeEmitter/BranchOps.inl
@@ -25,11 +25,11 @@ public:
     if (Imm >= -1048576 && Imm <= 1048575 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0101'010 << 25;
       Branch_Conditional(Op, 0, 0, Cond, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
   [[nodiscard]] BranchEncodeSucceeded b(ARMEmitter::Condition Cond, ForwardLabel* Label) {
     AddLocationToLabel(Label, ForwardLabel::Reference {.Location = GetCursorAddress<uint8_t*>(), .Type = ForwardLabel::InstType::BC});
@@ -37,7 +37,7 @@ public:
     Branch_Conditional(Op, 0, 0, Cond, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded b(ARMEmitter::Condition Cond, BiDirectionalLabel* Label) {
@@ -58,11 +58,11 @@ public:
     if (Imm >= -1048576 && Imm <= 1048575 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0101'010 << 25;
       Branch_Conditional(Op, 0, 1, Cond, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
 
   [[nodiscard]] BranchEncodeSucceeded bc(ARMEmitter::Condition Cond, ForwardLabel* Label) {
@@ -71,7 +71,7 @@ public:
     Branch_Conditional(Op, 0, 1, Cond, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded bc(ARMEmitter::Condition Cond, BiDirectionalLabel* Label) {
@@ -119,11 +119,11 @@ public:
     if (Imm >= -134217728 && Imm <= 134217724 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0001'01 << 26;
       UnconditionalBranch(Op, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
   [[nodiscard]] BranchEncodeSucceeded b(ForwardLabel* Label) {
     AddLocationToLabel(Label, ForwardLabel::Reference {.Location = GetCursorAddress<uint8_t*>(), .Type = ForwardLabel::InstType::B});
@@ -132,7 +132,7 @@ public:
     UnconditionalBranch(Op, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded b(BiDirectionalLabel* Label) {
@@ -155,11 +155,11 @@ public:
       constexpr uint32_t Op = 0b1001'01 << 26;
       UnconditionalBranch(Op, Imm >> 2);
 
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
   [[nodiscard]] BranchEncodeSucceeded bl(ForwardLabel* Label) {
     AddLocationToLabel(Label, ForwardLabel::Reference {.Location = GetCursorAddress<uint8_t*>(), .Type = ForwardLabel::InstType::B});
@@ -168,7 +168,7 @@ public:
     UnconditionalBranch(Op, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded bl(BiDirectionalLabel* Label) {
@@ -192,11 +192,11 @@ public:
     if (Imm >= -1048576 && Imm <= 1048575 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0011'0100 << 24;
       CompareAndBranch(Op, s, rt, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
 
   [[nodiscard]] BranchEncodeSucceeded cbz(ARMEmitter::Size s, ARMEmitter::Register rt, ForwardLabel* Label) {
@@ -207,7 +207,7 @@ public:
     CompareAndBranch(Op, s, rt, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded cbz(ARMEmitter::Size s, ARMEmitter::Register rt, BiDirectionalLabel* Label) {
@@ -230,11 +230,11 @@ public:
     if (Imm >= -1048576 && Imm <= 1048575 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0011'0101 << 24;
       CompareAndBranch(Op, s, rt, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
 
   [[nodiscard]] BranchEncodeSucceeded cbnz(ARMEmitter::Size s, ARMEmitter::Register rt, ForwardLabel* Label) {
@@ -245,7 +245,7 @@ public:
     CompareAndBranch(Op, s, rt, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded cbnz(ARMEmitter::Size s, ARMEmitter::Register rt, BiDirectionalLabel* Label) {
@@ -268,11 +268,11 @@ public:
     if (Imm >= -32768 && Imm <= 32764 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0011'0110 << 24;
       TestAndBranch(Op, rt, Bit, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
 
   [[nodiscard]] BranchEncodeSucceeded tbz(ARMEmitter::Register rt, uint32_t Bit, ForwardLabel* Label) {
@@ -283,7 +283,7 @@ public:
     TestAndBranch(Op, rt, Bit, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded tbz(ARMEmitter::Register rt, uint32_t Bit, BiDirectionalLabel* Label) {
@@ -306,11 +306,11 @@ public:
     if (Imm >= -32768 && Imm <= 32764 && ((Imm & 0b11) == 0)) [[likely]] {
       constexpr uint32_t Op = 0b0011'0111 << 24;
       TestAndBranch(Op, rt, Bit, Imm >> 2);
-      return true;
+      return BranchEncodeSucceeded::Success;
     }
 
     // Can't encode.
-    return false;
+    return BranchEncodeSucceeded::Failure;
   }
 
   [[nodiscard]] BranchEncodeSucceeded tbnz(ARMEmitter::Register rt, uint32_t Bit, ForwardLabel* Label) {
@@ -320,7 +320,7 @@ public:
     TestAndBranch(Op, rt, Bit, 0);
 
     // Forward label doesn't know if it can encode until Bind.
-    return true;
+    return BranchEncodeSucceeded::Success;
   }
 
   [[nodiscard]] BranchEncodeSucceeded tbnz(ARMEmitter::Register rt, uint32_t Bit, BiDirectionalLabel* Label) {
--- a/CodeEmitter/CodeEmitter/Emitter.h
+++ b/CodeEmitter/CodeEmitter/Emitter.h
@@ -586,14 +586,13 @@ concept IsXOrWRegister = std::is_same_v<T, XRegister> || std::is_same_v<T, WRegi
 template<typename T>
 concept IsQOrDRegister = std::is_same_v<T, QRegister> || std::is_same_v<T, DRegister>;
 
-struct BranchEncodeSucceeded final {
-  bool _________;
+template<typename T>
+concept IsLabel = std::is_same_v<T, ARMEmitter::ForwardLabel> || std::is_same_v<T, ARMEmitter::BackwardLabel> ||
+                  std::is_same_v<T, ARMEmitter::BiDirectionalLabel> || std::is_same_v<T, ARMEmitter::ForwardLabel::Reference>;
 
-  explicit operator bool() const {
-    return _________;
-  }
-  BranchEncodeSucceeded(bool a)
-    : _________ {a} {}
+enum class BranchEncodeSucceeded {
+  Success,
+  Failure,
 };
 
 // Whether or not a given set of vector registers are sequential
--- a/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
@@ -93,14 +93,12 @@ void Dispatcher::EmitDispatcher() {
 
   FillStaticRegs();
   ldr(RipReg, STATE_PTR(CpuStateFrame, State.rip));
-  if (!cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep)) [[unlikely]] {
-    LOGMAN_MSG_A_FMT("Dispatcher programming mistake");
-  }
+  (void)cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep);
 
   ARMEmitter::BiDirectionalLabel LoopTop {};
 
 #ifdef _M_ARM_64EC
-  b(&LoopTop);
+  (void)b(&LoopTop);
 
   AbsoluteLoopTopAddressEnterECFillSRA = GetCursorAddress<uint64_t>();
   ldr(STATE, EC_ENTRY_CPUAREA_REG, CPU_AREA_EMULATOR_DATA_OFFSET);
@@ -108,10 +106,10 @@ void Dispatcher::EmitDispatcher() {
 
   ldr(RipReg, STATE_PTR(CpuStateFrame, State.rip));
   // Force a single instruction block if ENTRY_FILL_SRA_SINGLE_INST_REG is nonzero entering the JIT, used for inline SMC handling.
-  cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep);
+  (void)cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep);
 
   // Enter JIT
-  b(&LoopTop);
+  (void)b(&LoopTop);
 
   AbsoluteLoopTopAddressEnterEC = GetCursorAddress<uint64_t>();
   // Load ThreadState and write the target PC there
@@ -132,7 +130,7 @@ void Dispatcher::EmitDispatcher() {
   ldp<ARMEmitter::IndexType::OFFSET>(TMP1, TMP2, REG_CALLRET_SP);
   // EC_CALL_CHECKER_PC_REG is REG_PF which isn't touched by any of the above
   sub(ARMEmitter::Size::i64Bit, TMP1, EC_CALL_CHECKER_PC_REG, TMP1);
-  cbnz(ARMEmitter::Size::i64Bit, TMP1, &LoopTop);
+  (void)cbnz(ARMEmitter::Size::i64Bit, TMP1, &LoopTop);
 
   // If the entry at the TOS is for the target address, pop it and return to the JIT code
   add(ARMEmitter::Size::i64Bit, REG_CALLRET_SP, REG_CALLRET_SP, 0x10);
@@ -144,9 +142,7 @@ void Dispatcher::EmitDispatcher() {
   // We want to ensure that we are 16 byte aligned at the top of this loop
   Align16B();
 
-  if (!Bind(&LoopTop)) [[unlikely]] {
-    LOGMAN_MSG_A_FMT("Dispatcher programming mistake");
-  }
+  (void)Bind(&LoopTop);
   AbsoluteLoopTopAddress = GetCursorAddress<uint64_t>();
 
   // Load in our RIP
@@ -173,20 +169,16 @@ void Dispatcher::EmitDispatcher() {
   ldr(TMP2, STATE_PTR(CpuStateFrame, Pointers.Common.ExitFunctionEC));
   br(TMP2);
 
-  if (!Bind(&l_NotECCode)) [[unlikely]] {
-    LOGMAN_MSG_A_FMT("Dispatcher programming mistake");
-  }
+  (void)Bind(&l_NotECCode);
 #endif
 
   ldrb(TMP1, STATE_PTR(CpuStateFrame, State.flags[X86State::RFLAG_TF_RAW_LOC]));
-  if (!cbnz(ARMEmitter::Size::i32Bit, TMP1, &CompileSingleStep)) [[unlikely]] {
-    LOGMAN_MSG_A_FMT("Dispatcher programming mistake");
-  }
+  (void)cbnz(ARMEmitter::Size::i32Bit, TMP1, &CompileSingleStep);
 
   ARMEmitter::ForwardLabel NoBlock;
 
   if (DisableL2Cache()) {
-    b(&NoBlock);
+    (void)b(&NoBlock);
   } else {
     // This is the block cache lookup routine
     // It matches what is going on it LookupCache.h::FindBlock
@@ -209,7 +201,7 @@ void Dispatcher::EmitDispatcher() {
       ldr(TMP1, TMP1, TMP2, ARMEmitter::ExtendedType::LSL_64, 3);
 
       // If page pointer is zero then we have no block
-      cbz(ARMEmitter::Size::i64Bit, TMP1, &NoBlock);
+      (void)cbz(ARMEmitter::Size::i64Bit, TMP1, &NoBlock);
 
       // Steal the page offset
       and_(ARMEmitter::Size::i64Bit, TMP2, TMP4, 0x0FFF);
@@ -224,10 +216,10 @@ void Dispatcher::EmitDispatcher() {
 
       // If the guest address doesn't match, Compile the block.
       sub(TMP2, TMP2, RipReg);
-      cbnz(ARMEmitter::Size::i64Bit, TMP2, &NoBlock);
+      (void)cbnz(ARMEmitter::Size::i64Bit, TMP2, &NoBlock);
 
       // Check the host address to see if it matches, else compile the block.
-      cbz(ARMEmitter::Size::i64Bit, TMP4, &NoBlock);
+      (void)cbz(ARMEmitter::Size::i64Bit, TMP4, &NoBlock);
 
       // If we've made it here then we have a real compiled block
       {
@@ -319,9 +311,7 @@ void Dispatcher::EmitDispatcher() {
 
   // Need to create the block
   {
-    if (!Bind(&NoBlock)) [[unlikely]] {
-      LOGMAN_MSG_A_FMT("Dispatcher programming mistake");
-    }
+    (void)Bind(&NoBlock);
 
     EmitSignalGuardedRegion([&]() {
       SpillStaticRegs(TMP1);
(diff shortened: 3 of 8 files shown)


Sonicadvance1:
Managed to find a bug in my original Long jump implementation. Turns out I was accidentally overwriting d15 with the stack pointer. Didn't seem to matter in the super limited use case in the Linux Syscall handler, but when I moved it to be used by the JIT it showed up as a problem.

Didn't appear as a problem in this PR, but did in the WIP code I have for buffer overrun detection.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -569,14 +588,23 @@ void Dispatcher::EmitDispatcher() {
     }
   }
 
-  Bind(&l_CTX);
+  if (!Bind(&l_CTX)) [[unlikely]] {
```
neobrain: Have you measured a performance impact for the `likely`/`unlikely` annotations added in this PR? If not,  I'd prefer not using them since they make simple expressions like this one overly noisy.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -826,16 +824,25 @@ void Arm64JITCore::EmitEntryPoint(ARMEmitter::BackwardLabel& HeaderLabel, bool C
 CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size, bool SingleInst, const FEXCore::IR::IRListView* IR,
                                                    FEXCore::Core::DebugData* DebugData, bool CheckTF) {
   FEXCORE_PROFILE_SCOPED("Arm64::CompileCode");
+  this->Entry = Entry;
+  this->DebugData = DebugData;
+  this->IR = IR;
+  RestartControl.NeedsLongJumps = false;
+
+  switch (FEXCore::LongJump::SetJump(RestartControl.RestartJump)) {
+  case FEXCore::ToUnderlying(RestartOptions::RestartOptionControl::Incoming):
```
neobrain: Instead of casting all possible enum values to their underlying type, why not `static_cast` the integer returned by SetJump to `RestartOptionControl`?

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -61,6 +62,15 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
   const bool HostSupportsRPRES {};
   const bool HostSupportsAFP {};
 
+  struct RestartOptions {
+    FEXCore::LongJump::JumpBuf RestartJump;
+    bool NeedsLongJumps {};
```
neobrain: Can we use `std::optional<JumpBuf>` here instead of manually managing validity?

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -315,12 +325,201 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+  requires (std::is_same_v<T, ARMEmitter::ForwardLabel> || std::is_same_v<T, ARMEmitter::BackwardLabel> ||
+            std::is_same_v<T, ARMEmitter::BiDirectionalLabel> || std::is_same_v<T, ARMEmitter::ForwardLabel::Reference>)
+  void b_OrRestart(T* Label) {
+    if (b(Label)) [[likely]] {
+      return;
+    }
+
+    // We can support this but currently unnecessary.
+    LOGMAN_MSG_A_FMT("Tried to branch larger than 128MB away!");
+    // Restart(RestartOptions::RestartOptionControl::NeedsLongJumps);
```
neobrain: Why is this here?

FEXCore/include/FEXCore/Utils/LongJump.h
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <FEXCore/Utils/CompilerDefs.h>
+
+#include <cstdint>
+
+// This is a custom long jump implementation that avoids the glibc implementation.
+// This is required behaviour because glibc's fortification checks don't understand stack pivots.
+// FEX requires a stack pivot to work through a long jump, so these two features are at odds with each other.
```
neobrain: This needs rewording (or moving), since "this is required behavior" only applies to the previous use site. It doesn't make any sense for a generic longjmp utility.

FEXCore/include/FEXCore/Utils/LongJump.h
```diff
@@ -0,0 +1,38 @@
...
+#else
+struct JumpBuf {
+  // Registers to preserve
+  // RBX, RSP, RBP, R12, R13, R14, R15,
+  // <return address>
+  uint64_t Registers[8];
+};
+#endif
+
+[[nodiscard]] FEX_DEFAULT_VISIBILITY uint64_t SetJump(JumpBuf& Buffer);
```
neobrain, line 35: Now that this is a public header, this should document its return value.

FEXCore/include/FEXCore/Utils/LongJump.h
```diff
@@ -0,0 +1,38 @@
...
+[[noreturn]] FEX_DEFAULT_VISIBILITY void LongJump(JumpBuf& Buffer, uint64_t Value);
```
neobrain, line 36: Now that this is a public header, this should document its `Value` parameter.

CodeEmitter/CodeEmitter/BranchOps.inl
```diff
@@ -20,23 +20,31 @@ public:
     constexpr uint32_t Op = 0b0101'010 << 25;
     Branch_Conditional(Op, 0, 0, Cond, Imm);
   }
-  void b(ARMEmitter::Condition Cond, const BackwardLabel* Label) {
+  [[nodiscard]] bool b(ARMEmitter::Condition Cond, const BackwardLabel* Label) {
```
neobrain: Could you change the unspecific `bool` returned by these functions to a self-documenting type like `struct BranchEncodingError { bool ___; }` (with an `explicit operator bool()`)?

At the moment, it's unclear why the functions that return this boolean need to do so, and conversely why other functions don't. It's also unclear which call sites need to check the boolean and which ones are fine to ignore it. Additionally, like other unnamed boolean values, it's prone to checking for the wrong boolean value. Using a named type largely addresses these problems. I think something like `BranchEncodeError` is clearest, but if there's too many conditions you'd need to flip around ,`BranchEncodeSucceeded` is probably also fine.

As a nice bonus we could add `Ignore()` and `AssertSuccess` functions to this new type instead of doing (void) casts and manual assertions, but I'll leave it at your discretion whether such functions would be useful.

FEXCore/Source/Interface/Core/JIT/JITClass.h
Sonicadvance1: That's not what is being checked here. This is a boolean for determining if the JIT code needs long jumps. `JumpBuf` is not optional.

FEXCore/Source/Interface/Core/JIT/JITClass.h
Sonicadvance1: Because the comment above, `// We can support this but currently unnecessary.`
If we deem it necessary to support >128MB then we can trvially add it with taht.

FEXCore/include/FEXCore/Utils/LongJump.h
Sonicadvance1: `// This return value behaves exactly like glibc setjump/longjump`?

CodeEmitter/CodeEmitter/BranchOps.inl
Sonicadvance1: Did it. I don't understand it, but it's there.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
Sonicadvance1: castered.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: It's already overly noisy with the checks. Noise on Noise it doesn't matter, no one is looking at this file every day anyway.

neobrain: The if-conditions are required to make this new feature work, whereas the likeliness-annotations don't have any proven benefit as far as I can see. Removing them shouldn't take long since the expressions are can easily be searched-and-replaced anyway.

CodeEmitter/CodeEmitter/Emitter.h
```diff
@@ -586,6 +586,16 @@ concept IsXOrWRegister = std::is_same_v<T, XRegister> || std::is_same_v<T, WRegi
 template<typename T>
 concept IsQOrDRegister = std::is_same_v<T, QRegister> || std::is_same_v<T, DRegister>;
 
+struct BranchEncodeSucceeded final {
+  bool _________;
```
neobrain: Come on, you can do better than that :)

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -61,6 +62,15 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
 
+  struct RestartOptions {
+    FEXCore::LongJump::JumpBuf RestartJump;
+    bool NeedsLongJumps {};
+    enum class RestartOptionControl : uint64_t {
+      Incoming = 0,
+      NeedsLongJumps = 1,
+    };
+  };
+  RestartOptions RestartControl {};
```
neobrain, line 80: This seems like a good place to document the restart mechanism and why it's needed.

FEXCore/Source/Interface/Core/JIT/JITClass.h
neobrain: If long jumps aren't needed, what's the alternative?

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -315,12 +325,201 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+  template<typename T>
+  requires (std::is_same_v<T, ARMEmitter::ForwardLabel> || std::is_same_v<T, ARMEmitter::BackwardLabel> ||
+            std::is_same_v<T, ARMEmitter::BiDirectionalLabel> || std::is_same_v<T, ARMEmitter::ForwardLabel::Reference>)
+  void b_OrRestart(T* Label) {
+    if (b(Label)) [[likely]] {
+      return;
+    }
+
+    // We can support this but currently unnecessary.
+    LOGMAN_MSG_A_FMT("Tried to branch larger than 128MB away!");
```
neobrain: `ERROR_AND_DIE_FMT` seems like a better fit here: The current code will break on execution regardless without a clear indication and in a way that's probably going be inconsistent to reproduce otherwise. A loud error can be reported by users and will immediately alert us about what needs to be fixed.

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -315,12 +325,201 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+    if (RestartControl.NeedsLongJumps) [[unlikely]] {
+      // This should have been caught before this point.
+      LOGMAN_MSG_A_FMT("Oops. Unhandled long bind.");
+      return;
+    }
+
+    Restart(RestartOptions::RestartOptionControl::NeedsLongJumps);
+  }
+
+  [[noreturn]] void Restart(RestartOptions::RestartOptionControl Control) {
```
neobrain: I don't see this function ever being called with anything other than `RestartOptionControl::NeedsLongJumps`. Any reason to have this parameter?

(If it is indeed removed, the function should be renamed to something like `RestartToEnableXYZ`)

FEXCore/include/FEXCore/Utils/LongJump.h
```diff
@@ -0,0 +1,36 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <FEXCore/Utils/CompilerDefs.h>
+
+#include <cstdint>
+
+// It's longjump without glibc fortification checks.
```
neobrain: Is skipping glibc fortification checks relevant for the JIT? Can't we use normal glibc setjmp/longjmp there?

FEXCore/include/FEXCore/Utils/LongJump.h
neobrain: You're right, it's already mentioned at the top of the file that these are setjmp/longjmp alternatives, so additional documentation here is not needed.

FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -826,16 +824,25 @@ void Arm64JITCore::EmitEntryPoint(ARMEmitter::BackwardLabel& HeaderLabel, bool C
...
+  this->DebugData = DebugData;
+  this->IR = IR;
+  RestartControl.NeedsLongJumps = false;
+
+  switch (static_cast<RestartOptions::RestartOptionControl>(FEXCore::LongJump::SetJump(RestartControl.RestartJump))) {
+  case RestartOptions::RestartOptionControl::Incoming:
+    // Nothing
+    break;
+  case RestartOptions::RestartOptionControl::NeedsLongJumps: RestartControl.NeedsLongJumps = true; break;
+  default: LOGMAN_MSG_A_FMT("Unhandled Arm64 restart condition!");
```
neobrain: This should be an `ERROR_AND_DIE_FMT`.

FEXCore/Source/Interface/Core/JIT/JITClass.h
neobrain: Overloading the term "LongJump" here to refer to far branches in our JIT is really confusing. Can you find a non-overloaded name and ideally add a comment to clarify what the purpose of the variable is?

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -61,6 +62,15 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
   const bool HostSupportsRPRES {};
   const bool HostSupportsAFP {};
 
+  struct RestartOptions {
+    FEXCore::LongJump::JumpBuf RestartJump;
+    bool NeedsLongJumps {};
+    enum class RestartOptionControl : uint64_t {
+      Incoming = 0,
+      NeedsLongJumps = 1,
```
neobrain: Probably better to rename this to `EnableLongJumps` (or `Enable<XYZ>` after renaming the `NeedsLongJumps` variable).

When seeing `RestartOptionControl::NeedsLongJumps`, I wouldn't have thought of reading it as `Restart *because* long jumps are needed`. It seems clearer to have it read as `Restart to enable long jumps` with the new name.

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -61,6 +62,15 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
   const bool HostSupportsRPRES {};
   const bool HostSupportsAFP {};
 
+  struct RestartOptions {
+    FEXCore::LongJump::JumpBuf RestartJump;
+    bool NeedsLongJumps {};
+    enum class RestartOptionControl : uint64_t {
```
neobrain: I think we're referring to this as `RestartOptions::RestartOptionControl` everywhere, so might as well drop the prefix:
```suggestion
    enum class Control : uint64_t {
```

FEXCore/Source/Interface/Core/JIT/JITClass.h
neobrain: Should this variable really be part of RestartOptions? IIUC it's just a general behavior toggle of the JIT, so perhaps it's better placed in Arm64JitCode itself? I found it a bit surprising to read `RestartControl.NeedsLongJumps` for code that wasn't itself related to the JIT restarts at all.

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -315,12 +325,201 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+  requires (std::is_same_v<T, ARMEmitter::ForwardLabel> || std::is_same_v<T, ARMEmitter::BackwardLabel> ||
+            std::is_same_v<T, ARMEmitter::BiDirectionalLabel> || std::is_same_v<T, ARMEmitter::ForwardLabel::Reference>)
+  void b_OrRestart(ARMEmitter::Condition Cond, T* Label) {
+    if (RestartControl.NeedsLongJumps) [[unlikely]] {
+      ARMEmitter::ForwardLabel Skip {};
+      // Invert check and skip.
+      (void)b(InvertCondition(Cond), &Skip);
+      if (!b(Label)) [[unlikely]] {
+        LOGMAN_MSG_A_FMT("Tried to branch larger than 128MB away!");
+      }
```
neobrain: Do I understand correctly that this is done here since unconditionally branches can do farther jumps than conditional ones?

Might be worth changing these comments to something like "Wrap a manual Cond check around an unconditional branch; this can encode larger offsets" then. The inverted Cond check itself seems self-explanatory to me.

CodeEmitter/CodeEmitter/BranchOps.inl
neobrain: This is somewhat better, but implicitly converting `bool->BranchEncodeSucceeded` loses the main benefit. I had one of these in mind instead:
  * EITHER an explicit constructor (invoked like `BranchEncodeSucceeded { true }`)
  * OR symbolic constants: `static constexpr BranchEncodeSucceeded Success { true }, Fail { false }`
  * OR member functions `bool Failed()`/`bool Succeeded()` while making the actual boolean member private.

> I don't understand it

An alternative you might find more intuitive is to use an `enum class BranchEncodeSucceeded { Success, Failure }` and to define an `bool operator !(BranchEncodeSucceeded Result) { Result != BranchEncodeSucceeded.Success; }`. This works since most result checks are for failure; for the few isolated success checks you'd have to use a double-negation like `!!b(Cond, Label)`.

Personally I find a `struct` with `Failed/Succeeded` member functions cleanest (since we do have checks for both failure and success, so it's good to be explicit), but the core goal here is to reduce boolean blindness, which either of these will do.

FEXCore/Source/Interface/Core/JIT/JITClass.h
neobrain: This also allows us to uncomment the line right below, so it at least gets syntax-checked to ensure it doesn't go out of sync with any future refactors.

FEXCore/Source/Interface/Core/JIT/JITClass.h
Sonicadvance1: Short jumps.

FEXCore/include/FEXCore/Utils/LongJump.h
Sonicadvance1: It's just stating a difference from and industrial standard library implementation. nothing more. Just because it is in FEXCore utils doesn't mean that all requirements for feature implementation must be solely dedicated to only what ends up in our JIT. The Frontend needs to to not hit fortification, the JIT core does not. I could delete the comment if you want.

neobrain: Sorry, I think I didn't word the intent clearly. The comment is fine. What I'm suggesting is that the new JIT restart code can use `std::setjmp`/`std::longjmp` (along with its fortification safety nets), whereas we use `FEXCore::SetJump/LongJump` only for code that has the exceptional requirement of omitting fortification checks (i.e. the Frontend).

Sonicadvance1: Nah, Gonna use the same long jump implementation for both.

neobrain: Why? Fortification checks are useful and help quickly catch bugs during development. I'd prefer to stick to one implementation too but in this case they're not feature-equivalent.

bylaws: We'll need a duplicate implementation for EC if that's done, just having something common seems reasonable to me

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -169,11 +167,11 @@ void Dispatcher::EmitDispatcher() {
   ldr(TMP2, STATE_PTR(CpuStateFrame, Pointers.Common.ExitFunctionEC));
   br(TMP2);
 
-  Bind(&l_NotECCode);
+  (void)!Bind(&l_NotECCode);
```
bylaws: Rather than ignoring the result for all of these, might it be better to keep the exception and have variants that return a result where needed? I can see arguments for either way but curious on thoughts

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -315,12 +329,197 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+      return;
+    }
+
+    // We can support this but currently unnecessary.
+    ERROR_AND_DIE_FMT("Tried to branch larger than 128MB away!");
+    FEXCore::LongJump::LongJump(RestartControl.RestartJump, FEXCore::ToUnderlying(RestartOptions::Control::EnableFarARM64Jumps));
+  }
+
+  template<typename T>
+  requires (std::is_same_v<T, ARMEmitter::ForwardLabel> || std::is_same_v<T, ARMEmitter::BackwardLabel> ||
```
bylaws: Perhaps just define a label concept in the emitter? avoids repeating this 10x

FEXCore/include/FEXCore/Utils/LongJump.h
neobrain: What's the constraint regarding EC?

bylaws: EC/WOW64 ship no libc, so we would need to implement ourselves

neobrain: Could you explain again what the difference is to the code map writing PR, where I thought we couldn't use certain primitives but only on specific threads?

I wish we had it written down somewhere what the exact constraints are, since it seems to be somewhat subtle.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
Sonicadvance1: No idea. I had checks, I removed check because it was stated to be too noisy. Collaborate and pick an evil.

neobrain: Which comment suggested removing the checks themselves? The only comment I made on this file was about removing unnecessary likeliness annotations. Was there another comment chain elsewhere that prompted this?

Sonicadvance1: It was stated as noisy, so I removed the entire check. It wasn't getting checked before, might as well as not get checked still.

FEXCore/include/FEXCore/Utils/LongJump.h
neobrain: @bylaws Any clarification on this?

bylaws: Both we have to use our own libc, but on external threads we cant call the initialisation funcs for our own libc or something windows apus most functions rely on

Source/Tools/LinuxEmulation/LinuxSyscalls/Utils/Threads.cpp
```diff
@@ -423,7 +287,7 @@ namespace PThreads {
     void* UserArg;
     void* Stack {};
 
-    LongJump::JumpBuf* _exit_resolver {};
+    FEXCore::LongJump::JumpBuf* _exit_resolver {};
```
neobrain, line 293: The usecase-specific documentation should be moved here or to some similar place:

```suggestion
    // Use FEXCore's LongJump to avoid fortification checks.
    // This avoids a false positive since glibc does not understand stack pivots.
    FEXCore::LongJump::JumpBuf* _exit_resolver {};
```

FEXCore/include/FEXCore/Utils/LongJump.h
neobrain: The overly playful tone here isn't suitable for technical documentation. Also longjump is spelled incorrectly.

```suggestion
// Reimplementation of longjmp without glibc fortification checks.
// This is useful to avoid false positives reported by glibc.
```

neobrain: Alright, I'll make a note to write this down in the documentation sometime.

For this PR, if "implement it ourselves" just means adding setjmp/longjmp wrappers around FEXCore reimplementations to WoA, that would be preferable. We don't need to unconditionally disabling fortification for other deployments then, and it ensures LongJump continues to have a clearly defined purpose within FEXCore.

Sonicadvance1:
Rebased. Commentato'd.

FEXCore/unittests/Emitter/ALU_Tests.cpp
```diff
@@ -9,167 +9,167 @@ using namespace ARMEmitter;
...
+    (void)Bind(&Label);
     dc32(0);
 
-    LongAddressGen(Reg::r30, &Label);
+    (void)LongAddressGen(Reg::r30, &Label);
     CHECK(DisassembleEncoding(1) == 0x10fffffe);
   }
   {
-    // Will generate nop + adr.
+    // Will generate nop + (void)adr.
```
neobrain: ?

neobrain: Seems to happen a bunch of times throughout the file.

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -331,12 +345,177 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+      return;
+    }
+
+    // We can support this but currently unnecessary.
+    ERROR_AND_DIE_FMT("Long ADRP currently unsupported!");
+    FEXCore::LongJump::LongJump(RestartControl.RestartJump, FEXCore::ToUnderlying(RestartOptions::Control::EnableFarARM64Jumps));
+  }
+
+  template<ARMEmitter::IsLabel T>
+  void BindOrRestart(T* Label) {
```
neobrain, line 505: Can we just use this everywhere instead of the existing `Bind` in some places and BindOrRestart in others?

It's currently unclear why there's two variants of the same thing, and using a single one lets us avoid the return-discarding `(void)` all over the place. We can also use the simpler name `Bind` then.

Applies to the other instruction types as well, I think.

(Not sure I asked the same question before, but it caught my eye again now that the vast majority of calls discard the result)

FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -331,12 +345,177 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
...
+    if (Bind(Label)) {
+      return;
+    }
+
+    if (RequiresFarARM64Jumps) {
+      // This should have been caught before this point.
+      ERROR_AND_DIE_FMT("Oops. Unhandled long bind.");
```
neobrain:
```suggestion
      ERROR_AND_DIE_FMT("Unhandled long bind");
```

FEXCore/include/FEXCore/Utils/LongJump.h
neobrain: Upon reflection, implementing the libc APIs for our WoA CRT is premature since the PR just adds one SetJump and 10 LongJump calls. We can just move the LongJump calls back to a helper function and then handle the WoA/glibc differences with two lines of `#ifdef _WIN32`.

FEXCore/Source/Interface/Core/JIT/JITClass.h
Sonicadvance1: Bind is a CodeEmitter function. `BindOrRestart` can't be moved to CodeEmitter. Both need to live.

Sonicadvance1:
Rebased from merge conflicts due to #4945

bylaws:
j.d -o CMakeFiles/FEXCore_object.dir/Interface/Core/JIT/MiscOps.cpp.obj -c /home/me/Documents/proton-arm64ec/build/src-fex-arm64ec/FEXCore/Source/Interface/Core/JIT/MiscOps.cpp
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:101:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  101 |   b(&LoopTop);
      |   ^ ~~~~~~~~
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:109:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  109 |   cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep);
      |   ^~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:112:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  112 |   b(&LoopTop);
      |   ^ ~~~~~~~~
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:133:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  133 |   cbnz(ARMEmitter::Size::i64Bit, TMP1, &LoopTop);
      |   ^~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:163:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  163 |   tbz(TMP1, 0, &l_NotECCode);
../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/JIT/JIT.cpp:790:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result]
  790 |   cbz(ARMEmitter::Size::i32Bit, TMP2, &l_NoSuspend);

Introduces these warnings

Sonicadvance1:
> j.d -o CMakeFiles/FEXCore_object.dir/Interface/Core/JIT/MiscOps.cpp.obj -c /home/me/Documents/proton-arm64ec/build/src-fex-arm64ec/FEXCore/Source/Interface/Core/JIT/MiscOps.cpp ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:101:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 101 | b(&LoopTop); | ^ ~~~~~~~~ ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:109:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 109 | cbnz(ARMEmitter::Size::i32Bit, ENTRY_FILL_SRA_SINGLE_INST_REG, &CompileSingleStep); | ^~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:112:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 112 | b(&LoopTop); | ^ ~~~~~~~~ ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:133:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 133 | cbnz(ARMEmitter::Size::i64Bit, TMP1, &LoopTop); | ^~~~ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp:163:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 163 | tbz(TMP1, 0, &l_NotECCode); ../../../src-fex-arm64ec/FEXCore/Source/Interface/Core/JIT/JIT.cpp:790:3: warning: ignoring return value of function declared with 'nodiscard' attribute [-Wunused-result] 790 | cbz(ARMEmitter::Size::i32Bit, TMP2, &l_NoSuspend);
> 
> Introduces these warnings

Looks like I missed a couple from the last rebase.

Sonicadvance1:
Cleared the warnings.

neobrain:
Submitted #5024 as a followup to the review discussion.

I also looked into using std::longjmp but found it made some of the trickier bits of the restart mechanism even trickier to follow, so I left that part as is. We can revisit if we adopt more uses of longjmp across the code base.

Issue #4877, opened 2025-09-11
JIT CodeBuffer guard page support


Sonicadvance1:
Currently we use heuristics to determine if code is likely to fit in our JIT cache with excessively large bounds to try and ensure code will fit.
Current heuristic at time of this issue being created is one page plus 24-bytes per SSA node, which...usually works at least but we definitely have SSA nodes that are larger than six instructions.

Instead of relying solely on that heuristic (which we still want to reduce fault overheads), we support a guard page on the end of the JIT buffer that is can be hit when the heuristic fails and we can then longjump back to a safe location in the JIT. Pretty straightforward and we already have a longjump implementation that we can use.

Limitations:
- Can't have nested longjumps
  - Not worth supporting currently
- Can't do unwinding
  - Should be fine, the JIT itself doesn't really need it. 

Overheads:
- Each JIT instance calls the `SetJump` handler in to a TLS `JumpBuf` and sets a flag that it is currently in a long-jump region
  - Function is low cost, 14 instructions.
- Fault check is low cost on Linux, fairly high on Arm64ec
  - On Linux just check the flag and restore the context from the JumpBuf
  - On Arm64ec it needs to do some ridiculous triple fault and a couple hundred syscalls or something, basically same cost as handling SMC/SIGBUS today. Main reason to still require the heuristic.

PR #4893, opened 2025-09-19
FEX: Implements support for JIT CodeBuffer guard page restart 

Sonicadvance1:
When the JIT CodeBuffer overflows, we will now catch accesses to the
guard page and longjump while restarting the JIT with a larger buffer
request.

Fixes https://github.com/FEX-Emu/FEX/issues/4877

Depends on #4892 getting merged first.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -824,6 +824,7 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
   this->DebugData = DebugData;
   this->IR = IR;
   RequiresFarARM64Jumps = false;
+  SSANodeMultiplier = 24;
 
   // Prepare restart via long jump in case branch encoding fails.
   // This uses UncheckedLongJump since we don't implement std::longjmp in WoA setups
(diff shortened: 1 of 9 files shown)


Sonicadvance1:
Need to investigate this one. Switching over to a guard page might be having some...weird behaviour.

bylaws (approved):

Source/Tools/LinuxEmulation/ArchHelpers/MContext.h
```diff
@@ -123,6 +123,10 @@ static inline uint64_t GetPc(void* ucontext) {
   return GetMContext(ucontext)->pc;
 }
 
+static inline uint64_t* GetArmPc(void* ucontext) {
+  return reinterpret_cast<uint64_t*>(&GetMContext(ucontext)->pc);
```
bylaws, line 127: Perhaps make getpc etc return a reference, and drop this?

Sonicadvance1: If this was switched over to a reference then it would still need a reinterpret_cast anyway because mcontext type for member `pc` is `unsigned long` (I think it differs between types in different aarch64 libraries even). So it's less filthy to just cast to uint64_t*.

Sonicadvance1:
Found out what the problem was. Turns out that my `Split out SIGSEGV handler` was too ambitious and I broke deferred signal handling. Fixed that and now everything is groovy.

PR #4899, opened 2025-09-22
Removes anonymous lambda usage from 32-bit syscalls

Sonicadvance1:
These need to not be anonymous, so just remove the lambdas entirely.
This is the first step towards syscall dispatcher tables being constexpr and supporting both 32-bit and 64-bit tables being executable inside the same process.

NFC.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/EPoll.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/EPoll.cpp
@@ -29,80 +29,80 @@ struct CpuStateFrame;
 }
 
 namespace FEX::HLE::x32 {
+auto epoll_wait(FEXCore::Core::CpuStateFrame* Frame, int epfd, compat_ptr<FEX::HLE::x32::epoll_event32> events, int maxevents, int timeout)
+  -> uint64_t {
+  fextl::vector<struct epoll_event> Events(std::max(0, maxevents));
+  uint64_t Result = ::syscall(SYSCALL_DEF(epoll_pwait), epfd, Events.data(), maxevents, timeout, nullptr, 8);
+
+  if (Result != -1) {
+    FaultSafeUserMemAccess::VerifyIsWritable(events, sizeof(FEX::HLE::x32::epoll_event32) * Result);
+    for (size_t i = 0; i < Result; ++i) {
+      events[i] = Events[i];
+    }
+  }
+  SYSCALL_ERRNO();
+}
+
+auto epoll_ctl(FEXCore::Core::CpuStateFrame* Frame, int epfd, int op, int fd, compat_ptr<FEX::HLE::x32::epoll_event32> event) -> uint64_t {
+  struct epoll_event Event;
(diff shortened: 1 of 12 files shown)


neobrain:
>  This is the first step towards syscall dispatcher tables being constexpr and supporting both 32-bit and 64-bit tables being executable inside the same process.

How is this related? Lambda expressions decay to function pointers, so I would expect them to be equivalent to defining free functions in that sense.

Sonicadvance1:
> > This is the first step towards syscall dispatcher tables being constexpr and supporting both 32-bit and 64-bit tables being executable inside the same process.
> 
> How is this related? Lambda expressions decay to function pointers, so I would expect them to be equivalent to defining free functions in that sense.

The act of them being a lambda or function doesn't matter in this case. It's the act of them being anonymous being the issue since I need the symbols visible for the dispatch arrays.

neobrain:
> The act of them being a lambda or function doesn't matter in this case. It's the act of them being anonymous being the issue since I need the symbols visible for the dispatch arrays.

Ah, yeah that sounds reasonable.

neobrain:
How far off is a full PR that enables 32-bit/64-bit table switching? I think it would be easier judge the change as a whole than looking at preliminary refactors.

Sonicadvance1:
> How far off is a full PR that enables 32-bit/64-bit table switching? I think it would be easier judge the change as a whole than looking at preliminary refactors.

Main thing is pushing all these functions in to the C namespace, that way we can use C symbol non-mangling to take advantage of variadic function signatures to mangle them in to constexpr tables.


```cpp
  extern "C" {
  auto fex_exit (FEXCore::Core::CpuStateFrame* Frame, ...) -> uint64_t;
  }

  auto Syscalls_32 = []() {
    std::array<SyscallHandler::SyscallFunctionDefinition, 512> Array {{
    {}, // TODO: restart_syscall.
    {{.PtrVariadic = &fex_exit}, 2}
    // TODO: Remaining 510 syscalls.
    }};
    return Array;
  }();
  ```

Generate a table for 64-bit and 32-bit, and then runtime will select which jumptable to use depending on if the entrypoint is 32-bit or 64-bit. Saving 8KB of generated memory allocations per process, and removing an insignificantly small startup cost.

That's as far as I've gotten before review happened.

Alternative reality, just make it `const static` and `reinterpret_cast<void*>` the function pointers and check at the end that the compiler generated a constant array.

neobrain:
I think it's still unclear to me why all this needs to happen at compile-time instead of using an incrementally (runtime-)populated table during FEX startup. A free saving of 8KiB would be nice, but also perhaps not the best guiding metric to design code around.

neobrain (changes requested):

neobrain:
There's 3 of these changesets that have been sitting in the pull request list for half a year now (on top of the 4 `cta_atexit` PRs). Can we move forward here or could we close the PRs until we have more time to get back to it so that the PR list isn't quite so cluttered?

neobrain:
Closing to de-clutter the PR list, let's reopen if needed later.

PR #4905, opened 2025-09-25
x87: optimize fild/fist memcpy

pmatos:


pmatos:
Still unsure how far to take this since it would be possible to keep the integers as integers in RP for memcpy without every having float conversions. I wonder how much of a gain that would be - have some testing to do here, so keeping this as a draft for now.

bylaws:
Is there a specific program this is intending to help?

pmatos:
> Is there a specific program this is intending to help?

Not really - however when I first wrote the x87 stack optimization pass, I included this InterpretAs flag that I wanted to use for something like this but never did, so when you mentioned on discord something like this for multiblocks, I remembered I could try going back and implementing memcpy on integers. There are some interesting issues with the optimization that make it more complex to implement, which is why it's in draft mode.

bylaws:
right - i think its worth finding some conccrete cases if this helps before persuing perhaps?

PR #4906, opened 2025-09-25
CodeCache: Introduce code maps

neobrain:
## Overview

Code maps form the basis of code caching: They collect metadata about emulated binaries (executables/libraries in ELF/PE format) such as relative x86 block entry addresses and loaded shared libraries. If enabled, FEX emits code map files during regular execution so that a code cache (to be implemented in a future PR) can be generated from them.

A key property of code maps is that they are independent of the FEX configuration and version used to generate them (with the exception of multiblock, see below). This allows code caches to be regenerated on configuration changes or after updating FEX without having to re-run the game.

## Core vs Frontend

FEXCore defines the general code map file layout (`FEXCore::CodeMap`) and filesystem-agnostic serialization (`FEXCore::CodeMapWriter`). Additionally, it defines code map naming conventions, including a `CodeMapFileID` that uniquely identifies binaries.

The frontend (SyscallHandler) takes care of filesystem storage, file management, and future cache generation. This is important since FEXInterpreter (Linux) will support a fully automatic cache generation flow that's concurrently used by multiple processes, whereas we'll have a simpler setup on WoA at the expense of requiring manual cache generation (at least for now).

## FEXInterpreter (Linux) implementation

Code map files are created and managed by FEXServer. FEXInterpreter sends a request to ask for them. This makes it easier to handle multiple concurrent processes sharing code caches for the main executable and libraries.

Code maps are written to `$XDG_CACHE/fex-emu/codemaps/<basefilename>-<fileid>[-nomb].N.bin`, where `N` is an integer that counts up for each generated code map. For example, after running `ls` 5 times, the following files will have been created:
```
~/.cache/fex-emu/codemap/ls-d55ce1dfc01cddc5.0.bin
~/.cache/fex-emu/codemap/ls-d55ce1dfc01cddc5.1.bin
~/.cache/fex-emu/codemap/ls-d55ce1dfc01cddc5.2.bin
~/.cache/fex-emu/codemap/ls-d55ce1dfc01cddc5.3.bin
~/.cache/fex-emu/codemap/ls-d55ce1dfc01cddc5.4.bin
```

Once code cache generation is added, FEXServer will also merge any previously written code maps and split them into per-library ones.

## Multiblock considerations

By definition, non-multiblock (no-MB) runs of FEX will hit significantly more block entries than multiblock (MB) runs. There is no way to identify the "true" MB entries from a no-MB code map, so a no-MB code map would blow up an MB code cache by a factor of ~20.

As a consequence, no-MB runs will add a `-nomb` suffix to codemap names. This is the only instance of FEX configuration leaking into code maps.

## Identifying files via CodeMapFileIds

Over time, FEX will most likely execute binaries that share the same name but are different (such as different versions of the same libraries, or games that have both a 32-bit and a 64-bit build). Identifying code maps by binary filename is hence not good enough; instead, a unique uint64_t identifier called `CodeMapFileId` is introduced.

Currently, a CodeMapFileId is simply computed by taking an xxhash over the absolute filename. This will need to be changed to be device-independent in the future by factoring in other sources of information.

## TODO

* [x] Add FEX config option control code map dumping
* [x] Fix WoA builds
* [x] Fix fork

### DIFF
--- a/FEXCore/include/FEXCore/Core/CodeCache.h
+++ b/FEXCore/include/FEXCore/Core/CodeCache.h
@@ -117,6 +117,14 @@ public:
     KnownFileIds.clear();
   }
 
+  bool IsBackingFD(int FD) const {
+    if (FD == CodeMapFD) {
+      LogMan::Msg::DFmt("Hiding directory entry for code map FD");
+      return true;
+    }
+    return false;
+  }
+
   void AppendBlock(const FEXCore::ExecutableFileSectionInfo&, uint64_t Entry);
   void AppendLibraryLoad(const FEXCore::ExecutableFileInfo&);
   void AppendSetMainExecutable(const FEXCore::ExecutableFileInfo&);
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -335,6 +335,8 @@ int SyscallHandler::OpenCodeMapFile() {
     return -1;
   }
 
+  FM.SetProtectedCodeMapFD(CodeMapFD);
+
   // Ensure the file descriptor is closed on exec
   auto flags = fcntl(CodeMapFD, F_GETFD);
   fcntl(CodeMapFD, F_SETFD, flags | FD_CLOEXEC);
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.cpp
@@ -1289,18 +1289,42 @@ void FileManager::UpdatePID(uint32_t PID) {
   }
 }
 
-bool FileManager::IsRootFSFD(int dirfd, uint64_t inode) const {
+bool FileManager::IsProtectedFile(int ParentDirFD, uint64_t inode) const {
   // Check if we have to hide this entry
-  if (inode == RootFSFDInode || inode == ProcFDInode) {
+  const char* Match = nullptr;
+  if (inode == RootFSFDInode) {
+    Match = "RootFS";
+  } else if (inode == ProcFDInode) {
+    Match = "/proc";
+  } else if (inode == CodeMapInode) {
+    Match = "code map";
+  }
+  if (Match) {
     struct stat Buffer;
-    if (fstat(dirfd, &Buffer) >= 0) {
+    if (fstat(ParentDirFD, &Buffer) >= 0) {
       if (Buffer.st_dev == ProcFSDev) {
-        LogMan::Msg::DFmt("Hiding directory entry for RootFSFD");
+        LogMan::Msg::DFmt("Hiding directory entry for {} FD", Match);
         return true;
       }
     }
   }
   return false;
+}
+
+void FileManager::SetProtectedCodeMapFD(int FD) {
+  if (FD == -1) {
+    CodeMapInode = 0;
+    return;
+  }
+
+  auto FDPath = fextl::fmt::format("self/fd/{}", FD);
+  struct stat Buffer {};
+  auto Result = fstatat(ProcFD, FDPath.c_str(), &Buffer, AT_SYMLINK_NOFOLLOW);
+  if (Result >= 0) {
+    CodeMapInode = Buffer.st_ino;
+  } else {
+    CodeMapInode = 0;
+  }
 }
 
 } // namespace FEX::HLE
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/FileManagement.h
@@ -94,7 +94,10 @@ public:
   uint64_t Statfs(const char* path, void* buf);
 
   void UpdatePID(uint32_t PID);
-  bool IsRootFSFD(int dirfd, uint64_t inode) const;
+  // Helper to detect FEX-internal files from their inode and parent directory FD.
+  // This is useful to deal with Chromium/CEF, which closes any FDs reported in /proc/self/fd/.
+  bool IsProtectedFile(int ParentDirFD, uint64_t inode) const;
+  void SetProtectedCodeMapFD(int FD);
 
   fextl::string GetEmulatedPath(const char* pathname, bool FollowSymlink = false) const;
   fextl::string GetHostPath(fextl::string& Path, bool AliasedOnly) const;
@@ -188,6 +191,7 @@ private:
   int ProcFD {0};
   int64_t RootFSFDInode = 0;
   int64_t ProcFDInode = 0;
+  int64_t CodeMapInode = 0;
   dev_t ProcFSDev;
 };
 } // namespace FEX::HLE
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.cpp
@@ -122,7 +122,7 @@ uint64_t GetDentsEmulation(int fd, T* dirp, uint32_t count) {
 
       TmpOffset += Tmp->d_reclen;
 
-      if (FEX::HLE::_SyscallHandler->FM.IsRootFSFD(fd, Outgoing->d_ino)) {
+      if (FEX::HLE::_SyscallHandler->FM.IsProtectedFile(fd, Outgoing->d_ino)) {
         continue;
       }
 
@@ -885,6 +885,9 @@ void SyscallHandler::LockBeforeFork(FEXCore::Core::InternalThreadState* Thread) 
 
 void SyscallHandler::UnlockAfterFork(FEXCore::Core::InternalThreadState* LiveThread, bool Child) {
   if (Child) {
+    // Code maps are closed upon fork in the child
+    FM.SetProtectedCodeMapFD(-1);
+
     VMATracking.Mutex.StealAndDropActiveLocks();
   } else {
     VMATracking.Mutex.unlock();
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/FD.cpp
@@ -604,7 +604,7 @@ void RegisterFD(FEX::HLE::SyscallHandler* Handler) {
       for (size_t i = 0, num = 0; i < Result; ++num) {
         linux_dirent_64* Incoming = (linux_dirent_64*)(reinterpret_cast<uint64_t>(dirp) + i);
         Incoming->d_off = num;
-        if (FEX::HLE::_SyscallHandler->FM.IsRootFSFD(fd, Incoming->d_ino)) {
+        if (FEX::HLE::_SyscallHandler->FM.IsProtectedFile(fd, Incoming->d_ino)) {
           Result -= Incoming->d_reclen;
           memmove(Incoming, (linux_dirent_64*)(reinterpret_cast<uint64_t>(Incoming) + Incoming->d_reclen), Result - i);
           continue;
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x64/FD.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x64/FD.cpp
@@ -116,7 +116,7 @@ void RegisterFD(FEX::HLE::SyscallHandler* Handler) {
       // Check for and hide the RootFS FD
       for (size_t i = 0; i < Result;) {
         linux_dirent_64* Incoming = (linux_dirent_64*)(reinterpret_cast<uint64_t>(dirp) + i);
-        if (FEX::HLE::_SyscallHandler->FM.IsRootFSFD(fd, Incoming->d_ino)) {
+        if (FEX::HLE::_SyscallHandler->FM.IsProtectedFile(fd, Incoming->d_ino)) {
           Result -= Incoming->d_reclen;
           memmove(Incoming, (linux_dirent_64*)(reinterpret_cast<uint64_t>(Incoming) + Incoming->d_reclen), Result - i);
           continue;


FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -3,10 +3,135 @@
...
+
+  std::string_view base_filename = FHU::Filesystem::GetFilename(std::string_view {MainExecutable.Filename});
+  if (FileId != 0xffff'ffff'ffff'ffff) {
+    return fextl::fmt::format("{}-{:016x}{}", base_filename, MainExecutable.FileId, AddNombSuffix ? "-nomb" : "");
+  }
+
+  return "";
+}
+
+fextl::map<CodeMapFileId, CodeMap::ParsedContents> CodeMap::ParseCodeMap(std::ifstream& File) {
```
Sonicadvance1, line 28: This API usage will need to be changed before merging occurs. https://github.com/FEX-Emu/FEX/blob/main/docs/ProgrammingConcerns.md#stdfstream

bylaws: This should be fine - the API isn't used within fex itself, only tools (perhaps this is a reason to move to Source)

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -349,6 +355,7 @@ class SyscallHandler : public FEXCore::HLE::SyscallHandler, FEXCore::HLE::Source
 
   FEX::HLE::SignalDelegator* SignalDelegation;
   FEX::HLE::ThunkHandler* ThunkHandler;
+  std::unique_ptr<FEXCore::CodeMapWriter> CodeMapWriter;
```
neobrain:
```suggestion
  fextl::unique_ptr<FEXCore::CodeMapWriter> CodeMapWriter;
```

Source/Tools/LinuxEmulation/LinuxSyscalls/CodeMapWriter.h
```diff
@@ -0,0 +1,91 @@
...
+      Flush(GetBufferOffset());
+      close(*CodeMapFD);
+    }
+  }
+
+  // To be used in a child process after fork to enable destroying this
+  // code map writer without flushing any pending data to disk.
+  void InvalidateFDs() {
+    CodeMapFD.reset();
+  }
```
neobrain: In addition to this, some mechanism is needed to prevent guest-side closes of this FD. (Similar to how IsRootFSFD is used in x64/FD.cpp for example.)

Source/Tools/LinuxEmulation/LinuxSyscalls/CodeMapWriter.h
```diff
@@ -0,0 +1,91 @@
...
+      auto ProgramFD = open((RootFSPath() + ProgramName.value()->c_str()).c_str(), O_RDONLY);
+      if (ProgramFD == -1) {
+        ProgramFD = open(ProgramName.value()->c_str(), O_RDONLY);
+      }
+      if (ProgramFD == -1) {
+        CodeMapFD = -1;
+        return false;
+      }
+
+      CodeMapFD = FEXServerClient::RequestCodeMapFD(FEXServerClient::GetServerFD(), ProgramFD, Multiblock);
```
neobrain: The FD should be added to the tracked FD set.

Source/Tools/LinuxEmulation/LinuxSyscalls/CodeMapWriter.h
```diff
@@ -0,0 +1,91 @@
...
+      if (CodeMapFD == -1) {
+        return false;
+      }
+
+      // Ensure the file descriptor is closed in fork children
+      auto flags = fcntl(CodeMapFD.value(), F_GETFD);
+      fcntl(CodeMapFD.value(), F_SETFD, flags | FD_CLOEXEC);
+    }
+
+    return true;
```
neobrain:
```suggestion
    return CodeMapFD != -1;
```
This handles the (unlikely) scenario that we raced another thread for code map FD initialization *and* that initialization failed in the other thread.

Source/Tools/LinuxEmulation/LinuxSyscalls/CodeMapWriter.h
```diff
@@ -0,0 +1,91 @@
...
+    // TODO: Also disable Wine
+    if (Section.FileInfo.Filename.starts_with("/run/pressure-vessel")) {
+      return false;
+    }
+
+    std::unique_lock<std::mutex> FileOpenLock {FileOpenMutex, std::defer_lock};
+    if (!CodeMapFD) {
+      FileOpenLock.lock();
+      // Re-check CodeMapFD now to avoid race conditions
+    }
```
neobrain: This is a strange way to say
```suggestion
    if (CodeMapFD) {
        return true;
    }
    std::unique_lock<std::mutex> FileOpenLock {FileOpenMutex};
    // Re-check CodeMapFD now to avoid race conditions
```

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -170,6 +170,17 @@ void SyscallHandler::InvalidateGuestCodeRange(FEXCore::Core::InternalThreadState
   InvalidateCodeRangeIfNecessary(Thread, Start, Length);
 }
 
+void SyscallHandler::MarkGuestCodeRoot(FEXCore::Core::InternalThreadState& Thread, uint64_t Address) {
+  if (!CodeMapWriter) {
+    return;
+  }
+
+  auto Region = LookupExecutableFileSection(Thread, Address);
+  if (Region && CodeMapWriter->IsWriteEnabled(*Region)) {
```
neobrain: LookupExecutableFileSection needs to fail if `!EntryIt->second.Resource->MappedFile`, otherwise it will dereference a null pointer which then manifests in an observed crashes here.

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -369,6 +379,48 @@ void HandleSocketData(fasio::tcp_socket& Socket) {
       buffer += sizeof(FEXServerClient::FEXServerRequestPacket::Header);
       break;
     }
+
+    case FEXServerClient::PacketType::TYPE_QUERY_CODE_MAP:
+    case FEXServerClient::PacketType::TYPE_QUERY_CODE_MAP_NO_MULTIBLOCK: {
+      char Tmp[PATH_MAX];
+      int TmpLen = FEX::get_fdpath(inFD, Tmp);
+      assert(TmpLen != -1);
+      fmt::print("Requested code map FD for: {}\n", std::string_view(Tmp, TmpLen));
```
neobrain: This is unnecessarily spammy.

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -463,6 +463,7 @@ void RegisterThread(FEX::HLE::SyscallHandler* Handler) {
     ThreadObject->StatusCode = status;
 
     FEX::HLE::_SyscallHandler->UninstallTLSState(ThreadObject);
+    FEX::HLE::_SyscallHandler->FlushAndCloseCodeMap();
```
neobrain: `exit` only exits the current thread, so this needs to be moved to e.g. `ThreadManager::DestroyThread`, where we can check if this is the last remaining thread.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -3,10 +3,135 @@
...
+  AppendData(std::as_bytes(std::span {Data, TotalSize}));
+}
+
+void CodeMapWriter::AppendData(std::span<const std::byte> Data) {
+  std::shared_lock Lock {Mutex};
+  auto Offset = BufferOffset.fetch_add(Data.size_bytes());
+  if (Offset + Data.size_bytes() > Buffer.size()) {
+    Lock.unlock();
+    Flush(Offset);
+    AppendData(Data);
```
neobrain, line 193: There's a race condition here: If the buffer is almost exhausted (`BufferOffset=0xffc`) and two threads simultaneously append data, both will call Flush. Worse yet, one of the threads will flush with an offset larger than the buffer size (trying to cover data that hasn't even been written to the buffer!).

neobrain:
I've removed unused capabilities and revamped the interface accordingly.

In the interest of moving things forward, I made the call to move most of the code into Core and leave only file opening to syscalls. I tried the CodeMapWriter-in-frontend approach in my WIP branch, but there was way too much passing back and forth of object pointers (particularly for cache validation). That's not needed anymore because the implementation is now much more localized.

This may get in the way of background operation on WoA, which should be acceptable given the typical size of code maps (<1 MB for the full app). If needed, `io_uring` would fit the new model.

neobrain:
I also added an option to control code caching (EnableCodeCachingWIP), using a WIP suffix to be removed later so we can reliably flip on code caching by default eventually.

bylaws:
I think what was prior was somewhat cleaner for woa? Like literally just a frontend function and headers to mark the entry point is all we need. Shouldn't need to pass around any pointers given that.

I wouldn't mind if this approach better fitted Linux just ifdefing off CodemapWriter for win and including structure headers, unfortunate if that needs to be done rather than a generic API though. But using FDs etc on Windows for a hot path isn't very preferable

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -3,10 +3,201 @@
...
+    {
+      std::unique_lock ExclusiveLock {Mutex};
+      IsResponsibleForFlush = (Offset <= Buffer.size());
+      if (IsResponsibleForFlush) {
+        Flush(Offset, ExclusiveLock);
+      }
+    }
+    if (!IsResponsibleForFlush) {
+      // Wait for the buffer to be flushed on the responsible thread
+      while (BufferOffset > Buffer.size()) {
```
bylaws, line 189: cvar?

Source/Common/Config.cpp
```diff
@@ -602,6 +602,12 @@ fextl::string GetConfigDirectory(bool Global, const PortableInformation& Portabl
   return ConfigDir;
 }
 
+fextl::string GetCacheDirectory() {
```
Sonicadvance1, line 605: I'll change this to be configurable in a follow-up PR. (Already opened but I'll rebase once this is in.)

FEXCore/Source/Interface/Core/CodeCache.cpp
Sonicadvance1, line 189: Is `std::vector::size` an Atomic read? I don't think it is.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -3,10 +3,201 @@
...
+        std::this_thread::yield();
```
Sonicadvance1, line 190: I'll change this to a WFE spin post-merge.

FEXCore/Source/Interface/Core/CodeCache.cpp
Sonicadvance1: https://godbolt.org/z/sxWE4P8br Not atomic so this needs to be atomic. Either I can rewrite the fextl::vector implementation post-merge alongside the WFE spin on the size, or you can choose some other atomic size mechanism reporting yourself.

neobrain: Buffer has a fixed size so it doesn't need atomicity here. Flushing only resets BufferOffset to 0 without clearing the `fextl::vector`. Should I clarify the comment?

neobrain: This loop is rarely hit in practice since the function entry is protected by a `shared_lock` (and the conditions for multiple threads racing here are very specific), so ultimately I don't think there would be an significant difference, would there?

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Note that the loop is rarely hit in practice as mentioned above, but feel free.

FEXCore/include/FEXCore/Core/CodeCache.h
```diff
@@ -33,10 +42,118 @@ struct ExecutableFileSectionInfo {
   uintptr_t FileStartVA;
 };
 
+using CodeMapFileId = uint64_t;
+
+/**
+ * TODO. Captures information required for offline compilation during execution.
```
neobrain: Oops, should fix the leftover TODO.

FEXCore/Source/Interface/Core/CodeCache.cpp
Sonicadvance1: Ah, I misread the directionality here, and the buffer offset is already atomic.

neobrain:
> But using FDs etc on Windows for a hot path isn't very preferable

Without background threads, do Windows APIs offer performance benefits over buffered FD IO? Wouldn't Wine ultimately translate these APIs to FDs under the hood anyway?

WoA being slightly more laborious to implement isn't great, but I think sharing the logic around library loads etc is going to be worth it in the long run. (It's not the most complicated code, but still easy to slip in surprising differences between the two impls)

Source/Common/Config.cpp
neobrain: I can also cherry-pick your commit on the next push. Meant to do so before but then forgot about it.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: To elaborate, for races to happen here on a significant scale it would need sustained multi-threaded JIT-hammering. The JIT-hammering would need to be heavy enough that at least two threads finish compiling at exactly the same time so that they are both able to acquire the `std::shared_mutex` before one of them promotes to an exclusive lock. This is only a problem every 300 multiblocks, since that's how many entries fit into the write buffer.

Even if this happens, a thread spinlocking until 1 kB of data is written to disk isn't the end of the world. It would need to happen periodically.

Sonicadvance1: Once this is in then we can get a shm stat just like the cachelock cycles to see how frequently we're spinning, or how much the spinning is contributing to judder/stutter.

bylaws:
> > But using FDs etc on Windows for a hot path isn't very preferable
> 
> Without background threads, do Windows APIs offer performance benefits over buffered FD IO? Wouldn't Wine ultimately translate these APIs to FDs under the hood anyway?

FEX is the one doing the translation fwiw, in general it's just somewhat indirect/not explicit whereas most other things on hot paths quite easily resolve out to raw API calls, especially if we end up wanting to do any sharing between processes etc.
> 
> WoA being slightly more laborious to implement isn't great, but I think sharing the logic around library loads etc is going to be worth it in the long run. (It's not the most complicated code, but still easy to slip in surprising differences between the two impls)


In what way? Considering the code is already written to do it in the simpler/faster threaded way for Windows and works, I want to understand the full structure to see why that's not ideal.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: What bylaws said, though I kept it in FEXCore since moving it to a different module just so it doesn't get compiled into FEX felt a bit weird. No strong opinions on this though.

Sonicadvance1: We can keep it in FEXCore as long as we guarantee it never gets called from FEXInterpreter. Maybe at some point in the future some quarantined folder but that doesn't need to occur today.

Source/Common/Config.cpp
Sonicadvance1: Whichever you want to do. Commits are free after-all.

neobrain:
> In what way? Considering the code is already written to do it in the simpler/faster threaded way for Windows and works, I want to understand the full structure to see why that's not ideal.

Changing the code map logic is more reliable when I can rely on the logic being 1:1 the same. I can't do that if WoA duplicates the code in a similar but different way (due to adding threading and Windows-specific APIs). Some divergence is expected, but given the scope of code caching, it's better to share any code possible than to slightly diverge in details all over the place.

neobrain:
Fixed all relevant remaining TODOs. Notably the code map FD is hidden from Chromium/CEF's sight now to avoid it being closed (repurposing the helpers we have in place for the RootFS FD).

bylaws:
> > In what way? Considering the code is already written to do it in the simpler/faster threaded way for Windows and works, I want to understand the full structure to see why that's not ideal.
> 
> Changing the code map logic is more reliable when I can rely on the logic being 1:1 the same. I can't do that if WoA duplicates the code in a similar but different way (due to adding threading and Windows-specific APIs). Some divergence is expected, but given the scope of code caching, it's better to share any code possible than to slightly diverge in details all over the place.


Right but what are the concrete changes that need to be made from the code that already exists for EC/is presented here for linux? I can understand if there's going to be significant churn in this area but otherwise favour the more optimal solution, especially as windows will be arriving afterwards anyway and needing frontend library tracking anyway. and yeah point about extensibility if we need to do any sharing between processes or anything that does take longer still stands.  It also allows for writing out library loads immediately rather than deferring seems slightly cleaner for any sort of AOT etc using SEH md for arm64ec libs with a small amount of mixed-code that may not be immediately hit.

FEXCore/include/FEXCore/Core/CodeCache.h
```diff
@@ -33,10 +42,126 @@ struct ExecutableFileSectionInfo {
...
+
+private:
+  // Queues data into an internal ring buffer.
+  // Call Flush() to commit the data to disk.
+  void AppendData(std::span<const std::byte> Data);
+
+  // Commit given data range to disk
+  void Flush(size_t Offset, std::unique_lock<std::shared_mutex>&);
+
+  std::shared_mutex Mutex;
```
neobrain, line 143: @Sonicadvance1 Does it need a ForkableSharedMutex here? CompileBlock already takes a signal deferring lock, so I think it's fine without.

neobrain:
(I started writing a reply while you were still editing, going to just post what I have now)

> Right but what are the changes that need to be made? I can understand if there's going to be significant churn in this area but otherwise favour the more optimal solution, especially as windows will be arriving afterwards anyway.

We don't know what changes are needed yet but it's reasonable to assume there will be changes. I don't expect threaded code map writing to have a crucial impact on user experience, so I'm favoring maintainability over a theoretical performance improvement here. If we measure a bottleneck, we have the options of increasing the buffer size and of exploring `io_uring` (which would benefit both WoA and FEXInterpreter instead of just the former).

> and yeah point about extensibility if we need to do any sharing between processes or anything that does take longer still stands.

Not sure I understand this sentence or what it's referring to.

> It also allows for writing out library loads immediately rather than deferring seems slightly cleaner for any sort of AOT etc or handling arm64ec libs with a small amount of mixed-code

Library loads could be written eagerly on Linux too. But writing them lazily avoids creating no-op code maps that *only* contain library loads and no actual block offsets. (This is the type of subtle consideration I'd rather deal with in common code rather than having each frontend do on their own!)

bylaws:
> (I started writing a reply while you were still editing, going to just post what I have now)
> 
> > Right but what are the changes that need to be made? I can understand if there's going to be significant churn in this area but otherwise favour the more optimal solution, especially as windows will be arriving afterwards anyway.
> 
> We don't know what changes are needed yet but it's reasonable to assume there will be changes. I don't expect threaded code map writing to have a crucial impact on user experience, so I'm favoring maintainability over a theoretical performance improvement here. If we measure a bottleneck, we have the options of increasing the buffer size and of exploring `io_uring` (which would benefit both WoA and FEXInterpreter instead of just the former).

This is my point, if things work as is with code caching and any future changes are unknown and hypothetical I'm not sure it makes sense to rely on them existing, it's kinda impossible to discuss trade-offs in any meaningful way against non-existent changes. But doing I/O when compiling certainly could cause stutters and a sep thread avoids that for no cost on WoA, especially with I/O in general being more costly on wine. Also consider  say steam stub which I think may need a fair amount of integration with library tracking and is Windows specific in how to do it. Additionally io uring in wine isn't a thing and certainly couldn't be done in a generic way with FDs so we'd introduce a ton of platform specific code anyway. One could use overlapped io but this is super windows specific and would warrant most of the impl being done on the windows side anyway. We also want a helper thread with a workqueue anyway for future suspend opts on windows, so it's not like this reduces any complexity there.


> 
> > and yeah point about extensibility if we need to do any sharing between processes or anything that does take longer still stands.
> 
> Not sure I understand this sentence or what it's referring to.

If we wanted to share codemap files between multiple processes etc, as linux fexserver does but for woa. 

> 
> > It also allows for writing out library loads immediately rather than deferring seems slightly cleaner for any sort of AOT etc or handling arm64ec libs with a small amount of mixed-code
> 
> Library loads could be written eagerly on Linux too. But writing them lazily avoids creating no-op code maps that _only_ contain library loads and no actual block offsets. (This is the type of subtle consideration I'd rather deal with in common code rather than having each frontend do on their own!)

No-op codemaps would explicitly be useful for AOT and analysis, if it's a problem it's trivial to fix on both and also negligible to diverge on both.

bylaws:
Did just check, the threaded windows impl for this is like 100 lines, effectively nothing.

FEXCore/include/FEXCore/Core/CodeCache.h
Sonicadvance1: As long as an upper Forkable{Shared,}Mutex is guaranteed to be held when a fork occurs then it doesn't need to be one. But if there is any chance of it being held when a fork occurs then that'll be a deadlock. Signal deferring alone doesn't guarantee that.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -312,6 +313,36 @@ uint64_t SyscallHandler::GuestMremap(bool Is64Bit, FEXCore::Core::InternalThread
...
+
+  int CodeMapFD = FEXServerClient::RequestCodeMapFD(FEXServerClient::GetServerFD(), ProgramFD, Multiblock);
+  close(ProgramFD);
+  if (CodeMapFD == -1) {
+    return -1;
+  }
+
+  FM.SetProtectedCodeMapFD(CodeMapFD);
+
+  // Ensure the file descriptor is closed in fork children
```
Sonicadvance1: I've seen this comment a few times. `CLOEXEC` specifically automatically closes the FD on execve{at,}. It doesn't close on fork. In /most/ cases it's usually always fork+exec but doesn't need to be.

FEXCore/include/FEXCore/Core/CodeCache.h
Sonicadvance1: Looked at how your mutex interacts with the other mutexes and it looks like it might already be gated by `CodeInvalidationMutex`? Might have missed an edge case but I think it's safe for now.

Sonicadvance1 (approved):

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
neobrain: Ah, good catch, I didn't know that. Changed the logic in ResetAfterFork accordingly to close the FD in the child.

FEXCore/include/FEXCore/Core/CodeCache.h
neobrain: Great, should be all gated behind the CodeInvalidationMutex indeed. Thanks for checking!

neobrain:
> This is my point, if things work as is with code caching and any future changes are unknown and hypothetical I'm not sure it makes sense to rely on them existing, it's kinda impossible to discuss trade-offs in any meaningful way against non-existent changes. But doing I/O when compiling certainly could cause stutters and a sep thread avoids that for no cost on WoA, especially with I/O in general being more costly on wine.

We can split things up later if we find this to be a critical bottleneck. As mentioned, I don't think it's going to be an issue since total code map data is small, especially in subsequent runs. (And for the first run, IO stutters would typically coincide with points of heavy JIT activity.)

> Also consider say steam stub which I think may need a fair amount of integration with library tracking and is Windows specific in how to do it

This seems solvable with some additional abstraction work, which has the benefit of allowing non-WoA deployments to cache such Windows applications as well.

> We also want a helper thread with a workqueue anyway for future suspend opts on windows, so it's not like this reduces any complexity there.

Do you want to elaborate on this so I can factor it into the code caching ADRs?

---

All things considered I want to move forward with the current approach. Let's prefer centralizing the code so it's consistent and works everywhere before adding WoA-specific divergence. We need to start somewhere with code maps, and I consider this more sensible than diverging by default (because realistically we're not going to re-common-ize code once it's branched out).

If there's improvements to the common interfaces that make WoA life easier, that should definitely be considered though. It's been tricky for me since we have no reference for the constraints in that environment.

bylaws:
Perhaps I can benchmark both approaches once I get to EC and see what's faster? Might be a reasonable solution rather than discussing now

Sonicadvance1 (approved):

PR #4912, opened 2025-09-29
FEXCore: Implement support for AVX gathers with overflow 

Sonicadvance1:
Falls down the emulated path so we can zero extend the address
calculation. No way for SVE gathers to use 32-bit addressing.

Still doesn't enable AVX on 32-bit by default. I need more 32-bit unittests to prove that all overflow paths are working as intended.


FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -171,6 +171,11 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
     return ConvertSize(Op);
   }
 
+  [[nodiscard]]
+  ARMEmitter::Size ConvertSize(const IR::OpSize Size) {
```
lioncash, line 175: Seems like all of these ConvertSize members can be static, but I can just move those over in a follow-up

PR #4933, opened 2025-10-05
RedundantFlagCalculationElimination: Eliminate unnecessary vector copy

lioncash:
Just a small thing I noticed (alongside minor tidying changes)


Sonicadvance1 (approved):
Nice catch.

PR #4937, opened 2025-10-06
Allocator: Name FEX's VMA regions for allocation

Sonicadvance1:
Will allow external tools to track how much memory FEX allocates. Necessary since we can't use traditional memory usage tools to track FEX memory allocation independently of guest allocations. Plus most tools like heaptrack hook allocation symbols, which break under jemalloc.

Using this information I can see with Steam loaded with my library that FEX consumes ~825MB. Total process resident memory is 1204M, accounting for around 379MB being used by steam itself. This is /relatively/ close to my desktop running steam at around 261MB. There's a bit of variance due to what Steam chooses to do at startup.

This tracking will be the first step towards seeing where our memory usage is going.

### DIFF
--- a/FEXCore/Source/Utils/Allocator.cpp
+++ b/FEXCore/Source/Utils/Allocator.cpp
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 #include "Utils/Allocator/HostAllocator.h"
+#include "Utils/PrctlUtils.h"
 #include <FEXCore/Utils/Allocator.h>
 #include <FEXCore/Utils/CompilerDefs.h>
 #include <FEXCore/Utils/LogManager.h>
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Utils/Allocator.cpp
```diff
@@ -50,6 +52,18 @@ void* FEX_mmap(void* addr, size_t length, int prot, int flags, int fd, off_t off
     errno = -(uint64_t)Result;
     return (void*)-1;
   }
+
+  if (flags & MAP_ANONYMOUS) {
+#ifndef PR_SET_VMA
```
lioncash: Should we have a PrctlUtils.h or something, so we don't have to have definition checks like this in multiple files?

Sonicadvance1: Maybe? I could probably whip up a header for it quick.

Sonicadvance1: Added a little header.

lioncash (approved):

PR #4940, opened 2025-10-06
Unaligned ec

cjacek:
On ARM64EC, EC code is allowed to perform unaligned atomic memory operations, similar to how FEX allows them in JIT code. From experiments on Windows, it appears that the kernel emulates some of these operations, specifically those that don’t cross a 16-byte alignment boundary, but leaves the remaining cases to be handled by the JIT module. The issue can be reproduced on Windows, where it works with xtajit but not with FEX. (To some extent of “works”, since xtajit doesn’t appear to make any attempt to keep those operations atomic. This PR does better than that.)

In my case, I was investigating how to handle this in Wine. With this PR, FEX is able to emulate atomic operations correctly in all cases. To better match Windows behavior, we could consider emulating the “kernel” part of this logic using signals within Wine, though I’m not sure it’s worth the complexity, especially with @bylaws’s kernel patch is on the horizon and this PR as a fallback.


FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2016,6 +2016,29 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
       }
+    } else if (!IsJIT) {
+      if ((Instr & ArchHelpers::Arm64::LDAXR_MASK) == ArchHelpers::Arm64::LDAXR_INST) { // LDAXR*
+        if (ArchHelpers::Arm64::HandleAtomicLoad(Instr, GPRs, 0, &Thread->ExclusiveStore)) {
+          return 4;
+        }
+      } else if ((Instr & ArchHelpers::Arm64::STLXR_MASK) == ArchHelpers::Arm64::STLXR_INST) { // STLXR*
+        uint32_t StatusReg = Instr << 11 >> 27;
+        // // Emulate exclusive store by validating the address and value against the last unaligned LDAXR*.
+        if (GPRs[AddrReg] != Thread->ExclusiveStore.Addr || Size > Thread->ExclusiveStore.Size) {
```
Sonicadvance1, line 2026: I don't think `Size > Thread->ExclusiveStore.Size` is correct. According to ARM documentation, a mismatched size between load and store exclusive is predictable undefined. `B2.12.5 Load-Exclusive and Store-Exclusive instruction usage restrictions` for the full documentation there.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -2016,6 +2016,29 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
...
+          if (StatusReg != 31) {
+            GPRs[StatusReg] = 1;
+          }
+          return 4;
+        }
+        if (std::optional<uint64_t> Prev = DoCAS(Size, GPRs[DataReg], Thread->ExclusiveStore.Store, GPRs[AddrReg], StrictSplitLockMutex)) {
```
Sonicadvance1, line 2032: Is this actually correct or are we just assuming sane application behaviour here?
Since theoretically two fighting ll/sc pairs on different processors could be invalidating the exclusive monitor and still have this CAS succeed.

cjacek: It depends on sane application behavior to a limited extent. As you noted, false positives are possible. When they occur, we should still be fine as long as the computation between the load and store does not depend on another value modified by a different thread. The ARM user restrictions effectively mandate this assumption by advising against explicit memory accesses between an exclusive load and store. When the computation depends only on the current thread state and the value observed by the load, it produces exactly the same result with CAS as it would if the exclusive load occurred after other threads’ memory modifications.

For comparison, Microsoft’s implementation doesn’t seem to account for this at all. It produces rather inaccurate results, similar to my earlier version of this patch, which didn’t attempt to emulate exclusive stores and simply performed a load/store.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
cjacek: The spec says that the implementation may require matching sizes, not that it must. I tested on both M2 and Snapdragon X, and they’re fine with an exclusive load being larger than the store.

FEXCore/Source/Utils/ArchHelpers/Arm64.cpp
```diff
@@ -1976,9 +1975,10 @@ HandleUnalignedAccess(FEXCore::Core::InternalThreadState* Thread, UnalignedHandl
 
   auto CTX = static_cast<Context::ContextImpl*>(Thread->CTX);
   uint32_t* StrictSplitLockMutex {CTX->Config.StrictInProcessSplitLocks ? &CTX->StrictSplitLockMutex : nullptr};
+  bool IsJIT = CTX->IsAddressInCodeBuffer(Thread, ProgramCounter);
```
Sonicadvance1: Instead of moving this check to here, can we instead add the argument to `FEXCore::ArchHelpers::Arm64::HandleUnalignedAccess` so that the various frontends only check this once?

cjacek: I changed that in the new version, thanks. I propagated it all the way from `ResetToConsistentStateImpl` so that we don't duplicate the check in `Exception::HandleUnalignedAccess`.

PR #4945, opened 2025-10-06
FEXCore: Adds option to disable L2 cache lookups

Sonicadvance1:
This saves a whole bunch of memory. Cutting `Just Cause 2`'s title
screen from 1132MB anonymous FEX memory down to 438MB. 629MB in L2
alone.

L2 is primarily a means to reduce overhead in map queries, so it's all
about performance. But because it consumes a lot of people it's kind of
hard.

One idea is that the L2 lookups can be moved to shared data structures,
since we already pull the shared lock when doing an L2 lookup this is
already halfway there.

Side note, we're using unique locks even with read-only code paths
which we can't use the shared lock because this terrible recursive
mutex!

Instead of outright changing L2 behaviour and potentially wrecking
havoc, add a config option for now so testing can happen over time.

before:
```
Total FEX Anon memory resident: 1132 mB
    JIT resident:             60 mB
    OpDispatcher resident:    97 mB
    Frontend resident:        37 mB
    CPUBackend resident:      500 kB
    Lookup cache resident:    629 mB
    Lookup L1 cache resident: 108 mB
    ThreadStates resident:    436 kB
```

after:
```
Total FEX Anon memory resident: 438 mB
    JIT resident:             62 mB
    OpDispatcher resident:    56 mB
    Frontend resident:        22 mB
    CPUBackend resident:      496 kB
    Lookup cache resident:    0 (null)
    Lookup L1 cache resident: 109 mB
    ThreadStates resident:    436 kB
```

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -161,6 +161,14 @@
         "Desc": [
           "Allows the user to pass additional arguments to the application"
         ]
+      },
+      "DisableL2Cache": {
+        "Type": "bool",
+        "Default": "false",
+        "Desc": [
+          "Disables FEXCore's JIT L2 cache lookup. Saving memory.",
+          "Can potentially introduce more stutters."
+        ]
       }
     },
     "Debug": {
--- a/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
@@ -177,60 +177,64 @@ void Dispatcher::EmitDispatcher() {
   ldrb(TMP1, STATE_PTR(CpuStateFrame, State.flags[X86State::RFLAG_TF_RAW_LOC]));
   cbnz(ARMEmitter::Size::i32Bit, TMP1, &CompileSingleStep);
 
-  // This is the block cache lookup routine
-  // It matches what is going on it LookupCache.h::FindBlock
-  ldr(TMP1, STATE_PTR(CpuStateFrame, Pointers.Common.L2Pointer));
-
-  // Mask the address by the virtual address size so we can check for aliases
-  uint64_t VirtualMemorySize = CTX->Config.VirtualMemSize;
-  if (std::popcount(VirtualMemorySize) == 1) {
-    and_(ARMEmitter::Size::i64Bit, TMP4, RipReg.R(), VirtualMemorySize - 1);
-  } else {
-    LoadConstant(ARMEmitter::Size::i64Bit, TMP4, VirtualMemorySize);
-    and_(ARMEmitter::Size::i64Bit, TMP4, RipReg.R(), TMP4);
-  }
-
   ARMEmitter::ForwardLabel NoBlock;
 
-  {
-    // Offset the address and add to our page pointer
-    lsr(ARMEmitter::Size::i64Bit, TMP2, TMP4, 12);
+  if (DisableL2Cache()) {
+    b(&NoBlock);
+  } else {
+    // This is the block cache lookup routine
+    // It matches what is going on it LookupCache.h::FindBlock
+    ldr(TMP1, STATE_PTR(CpuStateFrame, Pointers.Common.L2Pointer));
 
-    // Load the pointer from the offset
-    ldr(TMP1, TMP1, TMP2, ARMEmitter::ExtendedType::LSL_64, 3);
+    // Mask the address by the virtual address size so we can check for aliases
+    uint64_t VirtualMemorySize = CTX->Config.VirtualMemSize;
+    if (std::popcount(VirtualMemorySize) == 1) {
+      and_(ARMEmitter::Size::i64Bit, TMP4, RipReg.R(), VirtualMemorySize - 1);
+    } else {
+      LoadConstant(ARMEmitter::Size::i64Bit, TMP4, VirtualMemorySize);
+      and_(ARMEmitter::Size::i64Bit, TMP4, RipReg.R(), TMP4);
+    }
 
-    // If page pointer is zero then we have no block
-    cbz(ARMEmitter::Size::i64Bit, TMP1, &NoBlock);
-
-    // Steal the page offset
-    and_(ARMEmitter::Size::i64Bit, TMP2, TMP4, 0x0FFF);
-
-    // Shift the offset by the size of the block cache entry
-    add(TMP1, TMP1, TMP2, ARMEmitter::ShiftType::LSL, (int)log2(sizeof(FEXCore::LookupCache::LookupCacheEntry)));
-
-    // The the full LookupCacheEntry with a single LDP.
-    // Check the guest address first to ensure it maps to the address we are currently at.
-    // This fixes aliasing problems
-    ldp<ARMEmitter::IndexType::OFFSET>(TMP4, TMP2, TMP1, 0);
-
-    // If the guest address doesn't match, Compile the block.
-    sub(TMP2, TMP2, RipReg);
-    cbnz(ARMEmitter::Size::i64Bit, TMP2, &NoBlock);
-
-    // Check the host address to see if it matches, else compile the block.
-    cbz(ARMEmitter::Size::i64Bit, TMP4, &NoBlock);
-
-    // If we've made it here then we have a real compiled block
     {
-      // update L1 cache
-      ldr(TMP1, STATE_PTR(CpuStateFrame, Pointers.Common.L1Pointer));
+      // Offset the address and add to our page pointer
+      lsr(ARMEmitter::Size::i64Bit, TMP2, TMP4, 12);
 
-      and_(ARMEmitter::Size::i64Bit, TMP2, RipReg.R(), LookupCache::L1_ENTRIES_MASK);
-      add(TMP1, TMP1, TMP2, ARMEmitter::ShiftType::LSL, 4);
-      stp<ARMEmitter::IndexType::OFFSET>(TMP4, RipReg, TMP1);
+      // Load the pointer from the offset
+      ldr(TMP1, TMP1, TMP2, ARMEmitter::ExtendedType::LSL_64, 3);
 
-      // Jump to the block
-      br(TMP4);
+      // If page pointer is zero then we have no block
+      cbz(ARMEmitter::Size::i64Bit, TMP1, &NoBlock);
+
+      // Steal the page offset
+      and_(ARMEmitter::Size::i64Bit, TMP2, TMP4, 0x0FFF);
+
+      // Shift the offset by the size of the block cache entry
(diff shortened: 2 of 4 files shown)


bylaws:
Might the inline L2 lookup in the dispatcher work want to be changed also? I'll try to have a play around with this

Sonicadvance1:
> Might the inline L2 lookup in the dispatcher work want to be changed also? I'll try to have a play around with this

Oop, you're right. Disabled it in the dispatcher as well. I guess I got a clear in my stats after a bunch of blocks were already linked.

Sonicadvance1:
> I think this is conceivably very heavy handed, and we might be able to get almost equivalent gains without such an option - i think that should be tested first.
> 
> testing with multiblock on, if you kept the dispatcher L2 lookup logic, and simply only ever populated L2 if we got a hit of a block from this call
> 
> https://github.com/FEX-Emu/FEX/blob/e0c17672f20ed55496cb06effcef43d09385956a/FEXCore/Source/Interface/Core/Core.cpp#L769
> 
> (and explicitly not
> https://github.com/FEX-Emu/FEX/blob/e0c17672f20ed55496cb06effcef43d09385956a/FEXCore/Source/Interface/Core/JIT/JIT.cpp#L543
> 
> ). I would be interested in results from that, l1 indirect misses should be quite rare. if they showed removing l2 still provided a significant improvement then agree it makes sense.

If I'm reading your request correctly. For the three instances of `FindBlock` in our source. You want only the one in `Core.cpp:CompileBlock` to populate L2 (aka read and write it).
While the one in `Core.cpp:CompileCode` and `JIT.cpp:ExitFunctionLink` to explicitly avoid touching L2?

Additionally, keeping the L2 lookup logic enabled in the Dispatcher. Thus allowing it to zero-page populate it on reads?

Bayonetta title screen:
Proposed changes:
```
Total FEX Anon memory resident: 867 MiB
    JIT resident:             79 MiB
    OpDispatcher resident:    48 MiB
    Frontend resident:        18 MiB
    CPUBackend resident:      628 KiB
    Lookup cache resident:    381 MiB
    Lookup L1 cache resident: 88 MiB
    ThreadStates resident:    484 KiB
    BlockLinks resident:      7 MiB
          Misc resident:      0 KiB
    JEMalloc resident:        0 KiB
```

This PR with the L2Cache disabled:
```
Total FEX Anon memory resident: 491 MiB
    JIT resident:             79 MiB
    OpDispatcher resident:    68 MiB
    Frontend resident:        21 MiB
    CPUBackend resident:      584 KiB
    Lookup cache resident:    0 KiB
    Lookup L1 cache resident: 87 MiB
    ThreadStates resident:    488 KiB
    BlockLinks resident:      7 MiB
          Misc resident:      0 KiB
    JEMalloc resident:        0 KiB
```

Main:
```
Total FEX Anon memory resident: 956 MiB
    JIT resident:             79 MiB
    OpDispatcher resident:    68 MiB
    Frontend resident:        18 MiB
    CPUBackend resident:      608 KiB
    Lookup cache resident:    452 MiB
    Lookup L1 cache resident: 88 MiB
    ThreadStates resident:    476 KiB
    BlockLinks resident:      7 MiB
          Misc resident:      38 MiB
    JEMalloc resident:        0 KiB
```

Maybe it lowers L2 cache residency by 16%?

Sonicadvance1:
```diff
diff --git a/FEXCore/Source/Interface/Core/Core.cpp b/FEXCore/Source/Interface/Core/Core.cpp
index a30e01a95..0c7ae6429 100644
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -724,7 +724,7 @@ ContextImpl::CompileCodeResult ContextImpl::CompileCode(FEXCore::Core::InternalT
   // but this would increase lock contention. Redundant frontend runs aren't
   // as expensive and are easily reverted.
   if (MaxInst != 1) {
-    if (auto Block = Thread->LookupCache->FindBlock(GuestRIP)) {
+    if (auto Block = Thread->LookupCache->FindBlock(GuestRIP, true)) {
       Thread->OpDispatcher->DelayedDisownBuffer();
       return {.CompiledCode = {.BlockBegin = reinterpret_cast<uint8_t*>(Block), .EntryPoints = {{GuestRIP, reinterpret_cast<uint8_t*>(Block)}}},
               .DebugData = nullptr,
@@ -765,7 +765,7 @@ uintptr_t ContextImpl::CompileBlock(FEXCore::Core::CpuStateFrame* Frame, uint64_
 
   // Is the code in the cache?
   // The backends only check L1 and L2, not L3
-  if (auto HostCode = Thread->LookupCache->FindBlock(GuestRIP)) {
+  if (auto HostCode = Thread->LookupCache->FindBlock(GuestRIP, false)) {
     return HostCode;
   }
 
diff --git a/FEXCore/Source/Interface/Core/JIT/JIT.cpp b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
index 5bf618c0c..75a9e4a2b 100644
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -540,7 +540,7 @@ uint64_t Arm64JITCore::ExitFunctionLink(FEXCore::Core::CpuStateFrame* Frame, FEX
       // Guard the LookupCache lock with the code invalidation mutex, to avoid issues with forking
       auto lk_inval =
         GuardSignalDeferringSection<std::shared_lock>(static_cast<Context::ContextImpl*>(Thread->CTX)->CodeInvalidationMutex, Thread);
-      HostCode = Thread->LookupCache->FindBlock(GuestRip);
+      HostCode = Thread->LookupCache->FindBlock(GuestRip, true);
     }
     if (!HostCode) {
       // Hold a reference to the code buffer, to avoid linking unmapped code if compilation triggers a recreation.
diff --git a/FEXCore/Source/Interface/Core/LookupCache.h b/FEXCore/Source/Interface/Core/LookupCache.h
index 4ea48ef40..f920e8020 100644
--- a/FEXCore/Source/Interface/Core/LookupCache.h
+++ b/FEXCore/Source/Interface/Core/LookupCache.h
@@ -126,7 +126,7 @@ public:
     Shared = &NewMap;
   }
 
-  uintptr_t FindBlock(uint64_t Address) {
+  uintptr_t FindBlock(uint64_t Address, bool SkipL2) {
     // Try L1, no lock needed
     auto& L1Entry = reinterpret_cast<LookupCacheEntry*>(L1Pointer)[Address & L1_ENTRIES_MASK];
     if (L1Entry.GuestCode == Address) {
@@ -136,7 +136,7 @@ public:
     // L2 and L3 need to be locked
     auto lk = Shared->AcquireLock();
 
-    if (!DisableL2Cache()) {
+    if (!DisableL2Cache() && SkipL2 != true) {
       // Try L2
       const auto PageIndex = (Address & (VirtualMemSize - 1)) >> 12;
       const auto PageOffset = Address & (0x0FFF);
@@ -160,7 +160,7 @@ public:
     // Try L3
     auto HostCode = Shared->FindBlock(Address, lk);
     if (HostCode) {
-      CacheBlockMapping(Address, HostCode.value());
+      CacheBlockMapping(Address, HostCode.value(), SkipL2);
       return HostCode.value();
     }
 
@@ -263,13 +263,13 @@ public:
   }
 
 private:
-  void CacheBlockMapping(uint64_t Address, uintptr_t HostCode) {
+  void CacheBlockMapping(uint64_t Address, uintptr_t HostCode, bool SkipL2) {
     // Do L1
     auto& L1Entry = reinterpret_cast<LookupCacheEntry*>(L1Pointer)[Address & L1_ENTRIES_MASK];
     L1Entry.GuestCode = Address;
     L1Entry.HostCode = HostCode;
 
-    if (!DisableL2Cache()) {
+    if (!DisableL2Cache() && SkipL2 != true) {
       // Do ful map
       auto FullAddress = Address;
       Address = Address & (VirtualMemSize - 1);
@@ -286,7 +286,7 @@ private:
         if (!NewPageBacking) {
           // Couldn't allocate, clear L2 and retry
           ClearL2Cache();
-          CacheBlockMapping(Address, HostCode);
+          CacheBlockMapping(Address, HostCode, SkipL2);
           return;
         }
         Pointers[Address] = NewPageBacking;
```

For the test.

bylaws:
Hmm. Interesting results, if that much space is truly saved then we're hitting l3 an awful lot more than I expected? Returns should almost always hit callret but I guess in SMC scenarios things could be bad? Will try to think through this. I really would rather avoid disabling this outright and taking that as a solution when there could exist some alternative, since games do hit L2 in hot paths (and benchmarks, iirc geekbench clang)

Sonicadvance1:
> Hmm. Interesting results, if that much space is truly saved then we're hitting l3 an awful lot more than I expected? Returns should almost always hit callret but I guess in SMC scenarios things could be bad? Will try to think through this. I really would rather avoid disabling this outright and taking that as a solution when there could exist some alternative, since games do hit L2 in hot paths (and benchmarks, iirc geekbench clang)

If we're hitting L3 enough that it is becoming a problem then I would expect to see it in perf-top consistently, but I haven't been able to see that today. Bigger issue seems to be multiple threads blocking each other in JIT time when loading spikes occur, but I don't know if that is actual jitting or just reads. Will need to add a new shm stat to track  contended fetch time versus code actually compiling. But that's unrelated to this PR anyway.

bylaws (approved):

Sonicadvance1:
> Hmm. Interesting results, if that much space is truly saved then we're hitting l3 an awful lot more than I expected? Returns should almost always hit callret but I guess in SMC scenarios things could be bad? Will try to think through this. I really would rather avoid disabling this outright and taking that as a solution when there could exist some alternative, since games do hit L2 in hot paths (and benchmarks, iirc geekbench clang)

Anything from thinking on this?

bylaws:
> > Hmm. Interesting results, if that much space is truly saved then we're hitting l3 an awful lot more than I expected? Returns should almost always hit callret but I guess in SMC scenarios things could be bad? Will try to think through this. I really would rather avoid disabling this outright and taking that as a solution when there could exist some alternative, since games do hit L2 in hot paths (and benchmarks, iirc geekbench clang)
> 
> Anything from thinking on this?

I realised spawning new threads will have to repopulate l1 from l2 each time which probably causes this. Considering that my thought is that tieing l2 usage to dynamic l1 usage might make sense? Like once l1 stops expanding enable l2 etc.

Sonicadvance1:
> > > Hmm. Interesting results, if that much space is truly saved then we're hitting l3 an awful lot more than I expected? Returns should almost always hit callret but I guess in SMC scenarios things could be bad? Will try to think through this. I really would rather avoid disabling this outright and taking that as a solution when there could exist some alternative, since games do hit L2 in hot paths (and benchmarks, iirc geekbench clang)
> > 
> > 
> > Anything from thinking on this?
> 
> I realised spawning new threads will have to repopulate l1 from l2 each time which probably causes this. Considering that my thought is that tieing l2 usage to dynamic l1 usage might make sense? Like once l1 stops expanding enable l2 etc.

I guess we can tinker with it more as we move on. It doesn't need to be perfect the first time, this lets us iterate if we want to change it.

PR #4951, opened 2025-10-08
LookupCache: Adds an option to dynamically scale L1 cache 

Sonicadvance1:
L1 cache residency can get quite large. Solution, start out small and
scale quickly on L1 cache misses but L2/L3 cache hits.

Some stats on L1 cache residency change:
- Teardown: 40MB -> 16MB (40%)
- Ender Lilies: 79MB -> 32MB (40.5%)
- Death Stranding: 186MB -> 93MB (50%)
- Steam: 75MB -> 7MB (9.3%)

The cost of this option is effectively free in our JIT. It changes a
single LDR to be a single LDP, which on Cortex CPUs cost the same. We do
this by moving the L1 pointer mask in to the CPUState object, making it
dynamic so it lives next to the L1 pointer. We then use that directly
rather than having the hardcoded value.

The lookup cache does a little bit of additional tracking and heuristics
to determine when the current L1 cache should increase or decrease in
size. From 128KB to 16MB per thread, allocating the full VA range as
previously.

Once the heuristic determines that L1 should be increased, it simply
changes the max and the L1 pointer size to compensate, the kernel will
fault in whichever pages are necessary.

Decreasing the size is a little bit more complex, as we want to madvise
the resulting L1 range to ensure we don't have that memory as resident
anymore. Same heuristic but going in the opposite direction otherwise.

Tends to be the case that L1 cache increases a bit on loading screens
then backs down once in-game.

These heuristic values are exposed for increasing and decreasing because
while I think I've picked reasonable values, we will likely need some
more fine tuning over time. Kind of expert user toggles at that point.

Full tracked stats from steam as an example of where we are:
```
Total (1000 millisecond sample period):
       JIT Time: 0.486630 ms/second (0.00 percent)
    Signal Time: 0.065880 ms/second (0.00 percent)
     SIGBUS Cnt: 38 (38.160780 per second)
        SMC Cnt: 0
  Softfloat Cnt: 0
FEX JIT Load: 0.004585 (cycles: 552510)
Total FEX Anon memory resident: 368 mB
    JIT resident:             95 mB
    OpDispatcher resident:    38 mB
    Frontend resident:        8 mB
    CPUBackend resident:      624 kB
    Lookup cache resident:    0 (null)
    Lookup L1 cache resident: 7 mB
    ThreadStates resident:    460 kB
    Unaccounted resident:     217 mB
```


bylaws:
It would be useful to get start-up time benchmarks with this option, running node.js with hyperfine is usually a decent proxy for this

Sonicadvance1:
> It would be useful to get start-up time benchmarks with this option, running node.js with hyperfine is usually a decent proxy for this

`node-v22.20.0-linux-x64` and `node-v24.10.0-linux-x64` doesn't seem to work today under FEX. It has some `DelayedTaskSche` (duler?) thread that just spins 100% of its CPU time on epoll_pwait, epoll_ctl.
So benchmarking startup time with that process doesn't seem viable.

Sonicadvance1:
Apparently `node-v20.19.5-linux-x64` works. Locked performance governor to CPU cores to make things more consistent.

No caches disabled:
```
ryanh@orion-o6:~/node-v20.19.5-linux-x64/bin$ FEX_DISABLEL2CACHE=0 FEX_DYNAMICL1CACHE=0 taskset 0xff1 hyperfine -i --warmup 10 -r 50 'FEX ./node a'
Benchmark 1: FEX ./node a
  Time (mean ± σ):     506.5 ms ±   6.6 ms    [User: 457.6 ms, System: 53.6 ms]
  Range (min … max):   496.3 ms … 532.9 ms    50 runs
```

Only L2 disabled:
```
ryanh@orion-o6:~/node-v20.19.5-linux-x64/bin$ FEX_DISABLEL2CACHE=1 FEX_DYNAMICL1CACHE=0 taskset 0xff1 hyperfine -i --warmup 10 -r 50 'FEX ./node a'
Benchmark 1: FEX ./node a
  Time (mean ± σ):     506.7 ms ±   7.6 ms    [User: 455.8 ms, System: 55.7 ms]
  Range (min … max):   494.7 ms … 533.1 ms    50 runs
```

Only L1 dynamic:
```
ryanh@orion-o6:~/node-v20.19.5-linux-x64/bin$ FEX_DISABLEL2CACHE=0 FEX_DYNAMICL1CACHE=1 taskset 0xff1 hyperfine -i --warmup 10 -r 50 'FEX ./node a'
Benchmark 1: FEX ./node a
  Time (mean ± σ):     495.1 ms ±   6.0 ms    [User: 452.2 ms, System: 47.5 ms]
  Range (min … max):   485.3 ms … 511.9 ms    50 runs
```

L1 Dynamic and L2 disabled:
```
ryanh@orion-o6:~/node-v20.19.5-linux-x64/bin$ FEX_DISABLEL2CACHE=1 FEX_DYNAMICL1CACHE=1 taskset 0xff1 hyperfine -i --warmup 10 -r 50 'FEX ./node a'
Benchmark 1: FEX ./node a
  Time (mean ± σ):     486.4 ms ±   6.0 ms    [User: 449.2 ms, System: 41.4 ms]
  Range (min … max):   476.3 ms … 514.5 ms    50 runs
```

Sonicadvance1:
Whew, luckily I rebenched this after the rebase. I messed up a line ordering and it increased Proton startup time from ~23 seconds to ~47 seconds. With that fixed, the difference in startup time is falling within noise again.

Sonicadvance1:
Rebased from merge conflict.

Sonicadvance1:
Rebased.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -131,46 +131,93 @@ class LookupCache {
...
+        if (AveragePerSecond >= DynamicL1CacheIncreaseCountHeuristic()) {
+          if (CurrentL1Entries < MAX_L1_ENTRIES) {
+            CurrentL1Entries <<= 1;
+            L1PointerMask = CurrentL1Entries - 1;
+
+            // Update the thread's L1 pointer mask to increase how much cache it uses.
+            // Since we're in C-code, this is safe to update here.
+            Thread->CurrentFrame->State.L1Mask = GetScaledL1PointerMask();
+          }
+        } else if (AveragePerSecond < DynamicL1CacheDecreaseCountHeuristic()) {
```
bylaws: Trying to understand the logic here, will this not result in a shrink if we have everything we need in L1 and never end up hitting L2/L3 aside from on rare new block compiles? It doesn't seem an ideal steady state to have. Dynamic growth I think makes sense to enable by default but shrinking perhaps wants more evaluation

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -131,46 +131,93 @@ class LookupCache {
...
       }
     }
 
-    // Try L3
-    auto HostCode = Shared->FindBlock(Address, lk);
-    if (HostCode) {
-      CacheBlockMapping(Address, HostCode.value(), lk);
-      return HostCode.value();
+    if (HostPtr && DynamicL1Cache()) {
+      // If host pointer was found in L2 or L3, then add it to the counter.
```
bylaws: Think this is better in a new function

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -131,46 +131,93 @@ class LookupCache {
...
+          }
+        } else if (AveragePerSecond < DynamicL1CacheDecreaseCountHeuristic()) {
+          if (CurrentL1Entries > MIN_L1_ENTRIES) {
+            CurrentL1Entries >>= 1;
+            L1PointerMask = CurrentL1Entries - 1;
+
+            // Madvise the entries that we are dropping. Gives the memory back to the OS.
+            LookupCacheEntry* FirstZeroL1Entry = &reinterpret_cast<LookupCacheEntry*>(L1Pointer)[CurrentL1Entries];
+            size_t ZeroMemorySize = (MAX_L1_ENTRIES - CurrentL1Entries) * sizeof(LookupCacheEntry);
+            FEXCore::Allocator::VirtualDontNeed(FirstZeroL1Entry, ZeroMemorySize, true);
```
bylaws: Might be better off to have recommit as false here

Sonicadvance1: Switched to false.

FEXCore/Source/Interface/Core/LookupCache.h
Sonicadvance1: Moved to its own function.

FEXCore/Source/Interface/Core/LookupCache.h
Sonicadvance1: Yes, which is why it's a heuristic. It'll likely wobble slightly for anything that is riding that edge.

bylaws: Right - thoughts on defaulting to 0 then? At least for games I think that makes sense.

Sonicadvance1: That hurts larger engine games like Death Stranding where most of its code can't fit in to L1 anyway, so it scales up quickly on loading, and then can never scale back down again. With the entire dynamic option being disabled by default I don't see a reason to partial disable a heuristic. That's the point of a heuristic, there's no perfect answer.

bylaws: Sure, while this is disabled by default I agree but if there are plans to enable it I don't think the memory tradeoff would be worth it for basically wiping half the cache every second

Sonicadvance1: It's only wiping half the cache /if/ it's riding that edge, which isn't the common case.
For example in Death Stranding, without L1 dynamic cache, we are getting a cache miss count of ~3800/second. WITH dynamic L1, that jumps to ~4200/second. With no discernible performance difference but still cutting its L1 cache residency in half. (This is all thread accumulated stats so each individually settles out).

If we default the heuristic to never reduce (DynamicL1CacheDecreaseCountHeuristic=0) then it still is around ~3800/s (because it grows until it falls within the noise of the grow metric of 250/s). While only saving 22.5% residency (186MB -> 144MB).

It'll always be a trade-off, we can fine tune the heuristic if we determine it is causing problems in something, that's why the knobs are exposed. Disable entirely to see if the problem goes away, disable decreasing to see if it goes away, etc etc.

Sonicadvance1: Alternative reading. If without dynamic L1 cache, we're spending ~5ms/second in the JIT doing lookups (Across 54 threads), and with dynamic L1 cache that increases to 8-12ms/second then that is...effectively imperceptible?
If we find a game where it does start becoming perceptible then this is a number that can be changed to be more or less aggressive.

bylaws (approved):

PR #4953, opened 2025-10-09
FEXCore: Remove InlineSyscall

Sonicadvance1:
Due to IR changes we can no longer do this, its use was fairly limited anyway.


Sonicadvance1:
This snowballed a bit harder than expected, but still NFC since it's just a bunch of dead code removed.

PR #4960, opened 2025-10-09
HostFeatures: Enable AVX for 32-bit by default

Sonicadvance1:
With the gather overflow fixes in place, I've been having this running for a while. Now that we just kicked out a release, enable AVX even on 32-bit.

We'll need to eventually create a list of games that explode with AVX enabled, but that same list would match what happens on real x86 hosts, so there can be some collaboration there.


Sonicadvance1:
Fixed the bug and audited the rest of the codebase from it. Turns out this was technically the only instruction where we were listening to VEX.W incorrectly on 32-bit. The frontend decoder has a quirk where it enables "widening" when VEX.W and 64bit occurs, which was used for the handful of AVX instructions that use GPRs for one of their arguments. I added a unittest that hits all the relevant instructions anyway, and fixed the scalar FMA that messed up the logic.

PR #4964, opened 2025-10-10
SyscallsVMATracking: Make list management internally linked

lioncash:
These don't require being bound to class state directly, and so the list management can be completely opaque to the outside (also means less rebuilding if these change)


neobrain:
If you keep making more changes to this code, could you batch your changes into a single PR? This part of the code heavily interferes with code caching rebases, and it'd be much easier to deal with the fallout in isolation rather than with multiple split PRs.

lioncash:
Whoops, will do in the future

neobrain:
No problem, thanks! The cleanups are good but yeah we aren't exactly pressed on landing them in such quick succession either :)

PR #4965, opened 2025-10-10
FEXCore: Remove the last recursive_mutex

Sonicadvance1:
Every time I see this recursive mutex I glare at it. Remove the last one so that we no longer need to deal with it.

The only reason why this recursive mutex still existed today was because it is fairly intertwined with the ContextImpl and tracing it all was a pain.

Peel back the layers and follow the idiom to have ContextImpl pull the write mutex when requiredand pass it through by reference to ensure it stays alive. This allows us to entirely give rid of the recursive nature of the mutex, which means that `FindBlock` can eventually be switched over to a read-lock to improve multiple threads reading the caches at the same time.

I didn't do that exercise since that can be followed up in a subsequent PR.

### DIFF
--- a/FEXCore/Source/Interface/Core/LookupCache.h
+++ b/FEXCore/Source/Interface/Core/LookupCache.h
@@ -9,23 +9,27 @@
 #include <FEXCore/fextl/memory_resource.h>
 
 #include <cstdint>
-#include <functional>
 #include <stddef.h>
 #include <utility>
 #include <mutex>
 
 namespace FEXCore {
 
-struct GuestToHostMap {
-  std::recursive_mutex WriteLock;
+struct LookupCacheWriteLockToken {
+private:
+  // Only constructible by GuestToHostMap
+  friend struct GuestToHostMap;
+  LookupCacheWriteLockToken(std::mutex& Mutex)
+    : Lock {Mutex} {}
+  std::lock_guard<std::mutex> Lock;
+};
 
-  struct LockToken {
-    std::lock_guard<std::recursive_mutex> Lock;
-  };
+struct GuestToHostMap {
+  std::mutex WriteLock;
 
   [[nodiscard]]
-  LockToken AcquireLock() {
-    return LockToken {std::lock_guard {WriteLock}};
+  LookupCacheWriteLockToken AcquireWriteLock() {
+    return LookupCacheWriteLockToken {WriteLock};
   }
 
   struct BlockLinkTag {
@@ -61,7 +65,7 @@ struct GuestToHostMap {
   GuestToHostMap();
 
   // Adds to Guest -> Host code mapping
-  void AddBlockMapping(uint64_t Address, void* HostCode, const LockToken&) {
+  void AddBlockMapping(uint64_t Address, void* HostCode, const LookupCacheWriteLockToken&) {
     // This may replace an existing mapping
     // NOTE: Generally no previous entry should exist, however there is one exception:
     //       If the backend updates the active thread's CodeBuffer, the new associated LookupCache
@@ -70,7 +74,7 @@ struct GuestToHostMap {
     BlockList[Address] = (uintptr_t)HostCode;
   }
 
-  std::optional<uintptr_t> FindBlock(uint64_t Address, const LockToken&) {
+  std::optional<uintptr_t> FindBlock(uint64_t Address, const LookupCacheWriteLockToken&) {
     auto HostCode = BlockList.find(Address);
     if (HostCode == BlockList.end()) {
       return std::nullopt;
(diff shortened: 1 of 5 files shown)


FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -9,22 +9,22 @@
...
 #include <mutex>
+#include <shared_mutex>
 
 namespace FEXCore {
 
-struct GuestToHostMap {
-  std::recursive_mutex WriteLock;
+struct LockToken {
+  std::lock_guard<std::shared_mutex> Lock;
+};
```
neobrain: Extending use of this type outside of GuestToHostMap requires stronger protection to avoid accidental misuse (such as by passing an entirely unrelated mutex to it!).

```suggestion
struct LockToken {
private:
  // Only constructible by GuestToHostMap
  friend struct GuestToHostMap;

  LockToken(std::shared_mutex& Mutex)
    : Lock { Mutex } {}

  std::lock_guard<std::shared_mutex> Lock;
};
```

neobrain: Also, this needs a more specific name now that it lives in the quasi-global FEXCore namespace.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -226,14 +224,14 @@ class LookupCache {
...
+                    const FEXCore::Context::BlockDelinkerFunc& delinker, const LockToken& lk) {
     Shared->AddBlockLink(GuestDestination, HostLink, delinker, lk);
   }
 
-  void ClearCache();
-  void ClearL2Cache();
-  void ClearThreadLocalCaches();
+  void ClearCache(const LockToken& lk);
+  void ClearL2Cache(const LockToken& lk);
+  void ClearThreadLocalCaches(const LockToken& lk);
```
neobrain: `lk` adds no information here:
```suggestion
  void ClearCache(const LockToken&);
  void ClearL2Cache(const LockToken&);
  void ClearThreadLocalCaches(const LockToken&);
```

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -9,22 +9,22 @@
...
-  std::recursive_mutex WriteLock;
+struct LockToken {
+  std::lock_guard<std::shared_mutex> Lock;
+};
 
-  struct LockToken {
-    std::lock_guard<std::recursive_mutex> Lock;
-  };
+struct GuestToHostMap {
+  std::shared_mutex WriteLock;
```
neobrain: This should be a regular `mutex`, shouldn't it? We never call `lock_shared` on it, nor should we.

Sonicadvance1: L2 cache lookup should be a shared_lock in the future. As stated in the commit message `which means that FindBlock can eventually be switched over to a read-lock to improve multiple threads reading the caches at the same time.`

FEXCore/Source/Interface/Core/LookupCache.h
Sonicadvance1: Switched it over to your new implementation. Pick a name.

FEXCore/Source/Interface/Core/LookupCache.h
neobrain: That sounds like a design decision to be made separately. It's unrelated to this refactor.

FEXCore/Source/Interface/Core/LookupCache.h
neobrain: You're the author of this change - could you try?

Sonicadvance1: You're the one asking for a name change when I'm okay with `LockToken`. The ball has been passed.

Sonicadvance1: With 24 hours passed, I have changed the object's name to something that allows zero ambiguity as to what it is doing.

neobrain: Thanks. Note that we have to enforce the main limitation here by convention: The type may *only* be used for  GuestToHostMap mutexes but not any other ones (otherwise the idiom won't properly ensure thread-safety). A name highlighting this convention helps enforce this contract.

MutexScopedGuardWrapper is a good name for a generic utility, but something like `LookupCacheWriteLockToken` would be more suitable here.

Sonicadvance1: Renamed. Thanks for the name.

neobrain (approved):

PR #4968, opened 2025-10-11
InstallFEX:  Add support for 25.10

### LINKED ISSUE #4967: Apple Silicon M1: 'Ubuntu 25.10' is not a supported distro
Hi, I downloaded Parallels, downloaded Ubuntu 25.10, and updated.
Then before downloading Wine or Steam, I pasted:
curl --silent https://raw.githubusercontent.com/FEX-Emu/FEX/main/Scripts/InstallFEX.py | python3
'ubuntu 25.10' is not a supported distro

Why isn't the emulator working? What am I doing wrong? Thanks

Image

Sonicadvance1:
The PPA was already updated for this last month.

Fixes #4967

### DIFF
--- a/Scripts/InstallFEX.py
+++ b/Scripts/InstallFEX.py
@@ -81,11 +81,7 @@ def IsSupportedDistro():
 
     # We only support Ubuntu
     if Distro[0] == "ubuntu":
-        # We only support what is available in ppa:fex-emu/fex
-        return Distro[1] == "22.04" or \
-            Distro[1] == "24.04" or \
-            Distro[1] == "24.10" or \
-            Distro[1] == "25.04"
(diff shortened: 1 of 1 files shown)


Scripts/InstallFEX.py
```diff
@@ -85,7 +85,8 @@ def IsSupportedDistro():
         return Distro[1] == "22.04" or \
```
lioncash: Perhaps collapse this into:

```py
return Distro[1] in {"22.04", "24.04", "24.10", "25.04", "25.10"}
```

so it's a little more concise (and only the version tag needs to be added in the future, even with line breaking)

Sonicadvance1: changed.

PR #4969, opened 2025-10-11
Switch over to rpmalloc instead of jemalloc. 

Sonicadvance1:
rpmalloc is currently very aggressively configured which causes
significant reductions in resident memory over jemalloc.

In Bayonetta's title screen it went from 963MB down to 834MB resident.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -17,9 +17,6 @@
   shallow = true
 	path = External/fex-gcc-target-tests-bins
 	url = https://github.com/FEX-Emu/fex-gcc-target-tests-bins.git
-[submodule "External/jemalloc"]
-	path = External/jemalloc
-	url = https://github.com/FEX-Emu/jemalloc.git
 [submodule "External/fmt"]
 	path = External/fmt
 	url = https://github.com/fmtlib/fmt.git


bylaws:
What is the difference in memory usage with the regular rpmalloc configuration? It seems conceivable to me there is a JIT time trade off for this that should be considered. I think game/node start up times would need to be evaluated, jemalloc has a lot more features that help alloc perf

Sonicadvance1:
> What is the difference in memory usage with the regular rpmalloc configuration? It seems conceivable to me there is a JIT time trade off for this that should be considered

I still need to do more testing to see where we should land the config for rpmalloc. I saw at least 30MB saved from jemalloc with default config. Waiting for the L1/L2 cache PRs to land before doing extensive testing.

neobrain:
`docs/jemalloc_usage.md` needs updating before we merge this, though I'd also be curious about more detailed analysis before switching such an elementary block based on a single metric.

Also we should move the glibc_jemalloc module over to rpmalloc as well, since having *two* custom allocators is rather overkill.

Sonicadvance1:
> Also we should move the glibc_jemalloc module over to rpmalloc as well, since having _two_ custom allocators is rather overkill.

That will need to be evaluated over time. We already have two custom allocators called jemalloc and jemalloc_glibc. That's nothing new.
For switching the glibc allocator to rpmalloc there will need more work outside of FEX.
- rpmalloc needs to grow namespace support.
  - Unlike jemalloc, it doesn't support this so you can't have two instances in one executable
- rpmalloc will need to gain robustness support
  - Unlike jemalloc, it's not as robust against bad users. Last time I tried to switch glibc over to rpmalloc resulted in spurious crashes outside of FEX control. This is due to a lack of robustness that jemalloc supports.
- Test if their thread_local pools actually get correctly tracked on thread creation and teardown
  - This isn't guaranteed since their TLS heaps are supposed to be distinct without cross-thread interactions.

Sonicadvance1:
Some benchmark numbers for node startup time. Difference between jemalloc and rpmalloc falls within noise.
```
jemalloc:
  Time (mean ± σ):     508.4 ms ±   3.7 ms    [User: 458.7 ms, System: 54.5 ms]
  Range (min … max):   501.4 ms … 516.8 ms    50 runs

rpmalloc:
  Time (mean ± σ):     508.1 ms ±   6.3 ms    [User: 453.1 ms, System: 58.2 ms]
  Range (min … max):   499.1 ms … 543.4 ms    50 runs
```

I've also thrown Necrodancer at it, which absolutely hammers SMC if you cause the player to rub its face against a wall. JIT percantage time for a single thread in both cases hovered around 70%. Perf top was inconclusive with both jemalloc and rpmalloc routines taking ~1-2% CPU time.

I'm going to continue doing testing on this, but I don't think the aggressive config is detrimental to our workload due to how we've setup when we allocate memory.

Sonicadvance1:
If others want to dogfood this, that would be good, I'll be poking at this more over the weekend to try any bad cases.
This is a fairly ridiculous residency change when I was testing. 113MB -> 34MB (or 208MB -> 20MB in another test). With dynamic L1 and Disabled L2 then our memory usage reduction this month will be enormous.

Also one step away from relying on a dead project which is nice.

bylaws:
If something large and multi threaded like GoW or rachet and clank has similar startup time with this, seems good for me. (You would want to profile EC separately, will try to find time for this but quite busy)

Sonicadvance1:
> If something large and multi threaded like GoW or rachet and clank has similar startup time with this, seems good for me. (You would want to profile EC separately, will try to find time for this but quite busy)

Tested a GoW, Ratchet launch times between their launcher and to the title screen and PV launch times.
Image_2025-10-19_00-39-00

Percentage difference basically falls within noise.

Bigger problem is that rpmalloc does use a thread local variable, so when `get_thread_heap` is called, it causes a nullptr exception since it's just accessing random TLS index. Which I only realized after I tried testing out arm64ec.
Is there a way to get thread_local working or do I need to manually consume emulator TEB slots for wow64 and arm64ec? I think we discussed this before and I'll need to manually do this?

bylaws:
You can use flsalloc or tlsalloc (dynamic tls) windows apis, jemalloc does this rn

Sonicadvance1:
> You can use flsalloc or tlsalloc (dynamic tls) windows apis, jemalloc does this rn

I see. I'll look in to implementing that in to rpmalloc tomorrow.

Sonicadvance1:
Implemented the dynamic tls option in rpmalloc, was easier than expected. I did try to implement the FLS path that rpmalloc has but it resulted in weird hangs. Oh well, something to look in to later.

In any case, benching arm64ec GoW and Ratchet had the results fall within one standard deviation from each other, so no change. Didn't check resident memory since the savings should be roughly equivalent to the Linux side.

Source/Windows/ARM64EC/Module.cpp
```diff
@@ -56,6 +56,7 @@ desc: Implements the ARM64EC BT module API using FEXCore
 #include <winternl.h>
 #include <winnt.h>
 #include <wine/debug.h>
+#include <rpmalloc/rpmalloc.h>
```
bylaws: Move all this to (init,dinit)CRTThread, will also implicitly fix some bugs this would introduce with terminating different threads

bylaws: (or ideally use a tls callback but this is fine)

FEXCore/Source/Utils/AllocatorHooks.cpp
```diff
@@ -1,66 +1,206 @@
...
+static void* FEX_rp_mmap(size_t size, size_t alignment, size_t* offset, size_t* mapped_size) {
+#define pointer_offset(ptr, ofs) (void*)((char*)(ptr) + (ptrdiff_t)(ofs))
+  // If the alignment is less than the operating page size then alignment is guaranteed. Just remove it.
+  if (alignment < global_config.page_size) {
+    alignment = 0;
+  }
+
+  size_t map_size = AlignUp(size + alignment, global_config.page_size);
+#ifdef _WIN32
+  auto ptr = VirtualAlloc(0, map_size, MEM_RESERVE | MEM_COMMIT, PAGE_READWRITE);
```
bylaws: Do we want to actually commit here? The explicit commit func suggests not but unsure of rpmalloc internals

bylaws: Nevermind, seems rpmalloc bundles its own implementation of all these. and windows never calls `InitializeAllocator`. I think we should just drop the windows path for these and stop compiling for it, there's no need to use any of this file

bylaws (changes requested):

Source/Windows/ARM64EC/Module.cpp
Sonicadvance1: Moved over

FEXCore/Source/Utils/AllocatorHooks.cpp
Sonicadvance1: Wrapped in win32 checks to make sure it doesn't attempt to get used.

Source/Windows/WOW64/Module.cpp
```diff
@@ -586,7 +588,11 @@ void BTCpuProcessInit() {
   }
 }
 
-void BTCpuProcessTerm(HANDLE Handle, BOOL After, ULONG Status) {}
+void BTCpuProcessTerm(HANDLE Handle, BOOL After, ULONG Status) {
+  if (After) {
```
bylaws: Remove this for both, there's no point finalising if it's the current process and if it's a different process (see the Handle arg) this will cause things to explode. This is mostly meant for flushing files etc

Source/Windows/WOW64/Module.cpp
```diff
@@ -506,6 +507,7 @@ class WowSyscallHandler : public FEXCore::HLE::SyscallHandler, public FEXCore::A
 };
 
 void BTCpuProcessInit() {
+  rpmalloc_initialize(nullptr);
```
bylaws: Think this makes more sense in InitCRTProcess

Source/Windows/WOW64/Module.cpp
Sonicadvance1: Removed.

Source/Windows/WOW64/Module.cpp
Sonicadvance1: Moved over.

Sonicadvance1:
With the changes on the win32 side, I double checked and made sure wow64/arm64ec still worked.

PR #4971, opened 2025-10-12
unittests: Fixes mixture of code and data in the same page

Sonicadvance1:
Test behaviour themselves not changed at all, just data moved or aligned.

For tests that aren't explicitly testing out SMC behaviour, we were accidentally relying on some aggressive SMC tracking by mixing data and code in the same page. To fix this just align the test's data to the next page boundary which means FEX's SMC tracking won't get triggered since it is no longer living in the same page.

This has been a thorn for a while, so just get rid of it. We obviously still have ASM tests that still exist that /do/ rely on SMC, and those are still expected to work.


bylaws:
I would revert https://github.com/FEX-Emu/FEX/commit/2556acb82d08108d437e1e7d12cf1a5ddd75dc4d if this is ensured

Sonicadvance1:
> I would revert [2556acb](https://github.com/FEX-Emu/FEX/commit/2556acb82d08108d437e1e7d12cf1a5ddd75dc4d) if this is ensured

Good shout, reverted the commit and it also managed to find a few tests that I missed.

Sonicadvance1:
Rebased.

PR #4978, opened 2025-10-15
unittests/ThunkLibs: Fix build with LLVM 21

dramforever:
I'm not familiar with this codebase and just sought to fix unit test build errors we ran into in Nixpkgs. I hope this makes sense. Please advise.

### DIFF
--- a/unittests/ThunkLibs/common.h
+++ b/unittests/ThunkLibs/common.h
@@ -18,9 +18,15 @@ class TestDiagnosticConsumer : public clang::TextDiagnosticPrinter {
   std::optional<std::string> first_error;
 
 public:
+#if LLVM_VERSION_MAJOR >= 21
+  TestDiagnosticConsumer(bool silent_, clang::DiagnosticOptions& diag_opts)
+    : clang::TextDiagnosticPrinter(llvm::errs(), diag_opts)
+    , silent(silent_) {}
+#else
   TestDiagnosticConsumer(bool silent_)
     : clang::TextDiagnosticPrinter(llvm::errs(), new clang::DiagnosticOptions)
     , silent(silent_) {}
+#endif
 
   void HandleDiagnostic(clang::DiagnosticsEngine::Level level, const clang::Diagnostic& diag) override {
     if (level >= clang::DiagnosticsEngine::Error && !first_error) {
@@ -118,7 +124,13 @@ struct fex_gen_config;
 
   auto invocation = clang::tooling::ToolInvocation(args, &action, files.get(), std::make_shared<clang::PCHContainerOperations>());
 
+#if LLVM_VERSION_MAJOR >= 21
+  clang::DiagnosticOptions diag_opts;
+  TestDiagnosticConsumer consumer(silent, diag_opts);
+#else
   TestDiagnosticConsumer consumer(silent);
+#endif
+
   invocation.setDiagnosticConsumer(&consumer);
 
   // Process the actual ToolAction.
--- a/unittests/ThunkLibs/generator.cpp
+++ b/unittests/ThunkLibs/generator.cpp
@@ -188,13 +188,21 @@ SourceWithAST::SourceWithAST(std::string_view input, bool silent_compile)
 
     bool runInvocation(std::shared_ptr<clang::CompilerInvocation> invocation, clang::FileManager* files,
                        std::shared_ptr<clang::PCHContainerOperations> pch, clang::DiagnosticConsumer* diag_consumer) override {
-#if LLVM_VERSION_MAJOR >= 20
+#if LLVM_VERSION_MAJOR >= 21
+      auto diagnostics =
+        clang::CompilerInstance::createDiagnostics(files->getVirtualFileSystem(), invocation->getDiagnosticOpts(), diag_consumer, false);
+#elif LLVM_VERSION_MAJOR == 20
       auto diagnostics =
         clang::CompilerInstance::createDiagnostics(files->getVirtualFileSystem(), &invocation->getDiagnosticOpts(), diag_consumer, false);
 #else
       auto diagnostics = clang::CompilerInstance::createDiagnostics(&invocation->getDiagnosticOpts(), diag_consumer, false);
 #endif
+
+#if LLVM_VERSION_MAJOR >= 21
+      ast = clang::ASTUnit::LoadFromCompilerInvocation(invocation, std::move(pch), nullptr, std::move(diagnostics), files);
+#else
       ast = clang::ASTUnit::LoadFromCompilerInvocation(invocation, std::move(pch), std::move(diagnostics), files);
+#endif
       return (ast != nullptr);
     }
   } tool_action {ast};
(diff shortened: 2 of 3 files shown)


unittests/ThunkLibs/generator.cpp
```diff
@@ -188,13 +188,21 @@ SourceWithAST::SourceWithAST(std::string_view input, bool silent_compile)
...
+        clang::CompilerInstance::createDiagnostics(files->getVirtualFileSystem(), invocation->getDiagnosticOpts(), diag_consumer, false);
+#elif LLVM_VERSION_MAJOR == 20
       auto diagnostics =
         clang::CompilerInstance::createDiagnostics(files->getVirtualFileSystem(), &invocation->getDiagnosticOpts(), diag_consumer, false);
 #else
       auto diagnostics = clang::CompilerInstance::createDiagnostics(&invocation->getDiagnosticOpts(), diag_consumer, false);
 #endif
+
+#if LLVM_VERSION_MAJOR >= 21
+      ast = clang::ASTUnit::LoadFromCompilerInvocation(invocation, std::move(pch), nullptr, std::move(diagnostics), files);
```
dramforever, line 202: I'm not sure about this. Here's what I've found:

`nullptr` here should make sense, following this example https://github.com/llvm/llvm-project/blob/llvmorg-21.1.1/clang/lib/Tooling/Tooling.cpp#L651-L657

AFAICT it's just a lifetime hack to make sure the diagnostics options stay alive. If it's kept alive externally `nullptr` should be fine.

unittests/ThunkLibs/common.h
```diff
@@ -19,8 +19,14 @@ class TestDiagnosticConsumer : public clang::TextDiagnosticPrinter {
 
 public:
   TestDiagnosticConsumer(bool silent_)
-    : clang::TextDiagnosticPrinter(llvm::errs(), new clang::DiagnosticOptions)
-    , silent(silent_) {}
+    :
+#if LLVM_VERSION_MAJOR >= 21
+    clang::TextDiagnosticPrinter(llvm::errs(), *(new clang::DiagnosticOptions))
```
neobrain: I must've misunderstood the ownership semantics when writing this code; storing the DiagnosticOptions as a `unique_ptr` member so it doesn't leak after each test would be preferable.

dramforever: Made a drive-by fix

unittests/ThunkLibs/common.h
```diff
@@ -16,11 +17,19 @@ class TestDiagnosticConsumer : public clang::TextDiagnosticPrinter {
   bool silent;
 
   std::optional<std::string> first_error;
+  std::unique_ptr<clang::DiagnosticOptions> diag_opts;
 
 public:
   TestDiagnosticConsumer(bool silent_)
-    : clang::TextDiagnosticPrinter(llvm::errs(), new clang::DiagnosticOptions)
-    , silent(silent_) {}
+    : diag_opts(std::make_unique<clang::DiagnosticOptions>())
```
neobrain: Sorry, this is exactly what I had in mind but I'm now realizing this won't work: C++ always initializes base classes before members of the inheriting class, so (despite appearances) `diag_opts` is still a null pointer when you're dereferencing it in the `clang::TextDiagnosticPrinter` lines.

This helper class is only used once, so let's just manage the lifetime externally and pass it in from the caller:
```cpp
  TestDiagnosticConsumer(bool silent_, clang::DiagnosticOptions& diag_opts)
    : clang::TextDiagnosticPrinter(llvm::errs(), (&)diag_opts) {
```

```cpp
  auto diag_opts = std::make_unique<clang::DiagnosticOptions>();
  TestDiagnosticConsumer consumer(silent, *diag_opts);
```

dramforever: Then we don't need `std::unique_ptr`, right? Just need to have `clang::DiagnosticOptions diag_opts;` as a local variable

neobrain: Ah yes, good catch.

dramforever: Fixed the drive-by fix...

dramforever:
oh no... that still doesn't work

neobrain:
> oh no... that still doesn't work

Is it just the TextDiagnosticPrinter thing that's somehow crashing currently? Reverting back to the old behavior of leaking the allocation would be fine then. It seemed like an easy thing to fix, but maybe clang is throwing around more dangling references behind the scenes than anticipated.

dramforever:
It seems that LLVM 20 uses `IntrusiveRefCntPtr` here, so the `new` was actually correct. And then it's gone in LLVM 21 for https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f

dramforever:
https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f#diff-e44e848f7f9fc985845d5df39c392d545a4d30775bb86eaa5e746ebb7b980140 file `clang-tools-extra/clang-change-namespace/tool/ClangChangeNamespace.cpp` would be an example of how to migrate, i think

neobrain:
> It seems that LLVM 20 uses `IntrusiveRefCntPtr` here, so the `new` was actually correct. And then it's gone in LLVM 21 for [llvm/llvm-project@13e1a2c](https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f)

> https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f#diff-e44e848f7f9fc985845d5df39c392d545a4d30775bb86eaa5e746ebb7b980140 file clang-tools-extra/clang-change-namespace/tool/ClangChangeNamespace.cpp would be an example of how to migrate, i think

Thanks for the investigation. Sounds like this should work then?
* unconditionally manage an `IntrusiveRefCntPtr` at call-site and pass it by-value to TestDiagnosticConsumer
* LLVM 20: Pass it by-value to clang::TextDiagnosticPrinter (bumping the refcount)
* LLVM >=21: Dereference it and pass it by reference to clang::TextDiagnosticPrinter (leaving the refcount unchanged)

dramforever:
> unconditionally manage an IntrusiveRefCntPtr at call-site and pass it by-value to TestDiagnosticConsumer

So this won't work because https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f made `DiagnosticOptions` no longer have a CRTP mixin `public llvm::ThreadSafeRefCountedBase<DiagnosticOptions>`, so we can't use `IntrusiveRefCntPtr<DiagnosticOptions>` in LLVM 21

I decided to just not fix the LLVM <= 20 code because it was not broken. `clang::TextDiagnosticPrinter` takes the pointeer and converts it into a `IntrusiveRefCntPtr` and handles the lifetime itself, like a `std::shared_ptr`, which is correct and doesn't actually leak it. So, "if it ain't broke, don't fix it".

As of the LLVM >= 21 code I've written it in the style found in https://github.com/llvm/llvm-project/commit/13e1a2cb2246dc5e9a4afcdacabed4d43154ec3f where I pass a reference in externally, which should work given that it is how LLVM does it

neobrain:
Oof, fair enough. Looks like this passes CI now, so that'll do. Have you been able to run the tests on your end with LLVM 21 as well?

dramforever:
Yeah I added this PR as a patch file to 2510 in Nixpkgs and it fixes the unit tests build, see: https://github.com/dramforever/nixpkgs/compare/NixOS:master...fex-fix-llvm-21

(I'm hoping to combine that into https://github.com/NixOS/nixpkgs/pull/450537)

neobrain (approved):
Great. Thanks for the help!

PR #4980, opened 2025-10-15
LinuxSyscalls: Fix base address inference for ELF binaries

neobrain:
## Overview

Code caching (and code map generation) critically relies on FEX being able to compute the offset of any given code block from the library's base in **virtual address space**. `LookupExecutableFileSection` has always used **file offsets** instead, presumably based on the incorrect assumption that this would yield the same result.

In fact for most libraries, the file offset of a file mapping won't always match the address offset in virtual memory from the base file mapping. `mmap()` parameters don't contain the information needed to resolve this reliably. Instead, fixing the problem requires FEX to be aware of the executable format (PE/ELF) so that it can infer the required information from section headers.

One further complication arises because ELF files might be loaded multiple times in the same process. This breaks a previous assumption that there is one unambiguous base address for each mapped file. This is now resolved by creating separate MappedResources for each re-mapping of an ELF file header section, so that FEX always knows how to look up the correct library base address for any given file mapping.

## Implementation

The base address inference has been tricky to get right in all cases. Separating the core logic to a new `InferMappingBaseAddress` function allows unit-tests to be written to ensure any future changes don't regress behavior.

Locally, I also have code that makes this logic compatible with PE files. This is subject for a future PR.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -423,11 +423,17 @@ std::optional<SyscallHandler::LateApplyExtendedVolatileMetadata> SyscallHandler:
         Resource->MappedFile = fextl::make_unique<FEXCore::ExecutableFileInfo>();
         Resource->MappedFile->Filename = fextl::string(Tmp, PathLength);
 
-        // Read ELF headers if applicable
-        Resource->ProgramHeaders = ReadELFHeaders(fd, std::span {reinterpret_cast<std::byte*>(addr), length});
-        // If this assumption is broken, we can't reliably cluster subsequent mappings by their base address
-        LOGMAN_THROW_A_FMT(Resource->ProgramHeaders.empty() || offset == 0, "Expected file offset 0 for the first mapping of an ELF "
-                                                                            "file");
+        // Read ELF headers if applicable.
+        // For performance, skip ELF checks if we're not mapping the file header
+        bool CheckForElfFile = (offset == 0);
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+        CheckForElfFile = true;
+#endif
+        if (CheckForElfFile) {
+          Resource->ProgramHeaders = ReadELFHeaders(fd, std::span {reinterpret_cast<std::byte*>(addr), length});
+          LOGMAN_THROW_A_FMT(Resource->ProgramHeaders.empty() || offset == 0, "Expected file offset 0 for the first mapping of an ELF "
+                                                                              "file");
+        }
       } else if (ResourceIt->second.ProgramHeaders.empty()) {
         // Not an ELF file, so we don't need to distinguish between different base addresses
       } else {


Sonicadvance1:
> the file offset of a file mapping won't always match the address offset in virtual memory from the base file mapping. mmap() parameters don't contain the information needed to resolve this reliably.

What do you mean by this? If `!MAP_ANONYMOUS` then fd + mmap offset argument will reliably give the offset from the base of the file mapped.

> One further complication arises because ELF files might be loaded multiple times in the same process. This breaks a previous assumption that there is one unambiguous base address for each mapped file.

I'm genuinely surprised that this causes complications, if we're tracking files by inode then even if a file is mapped to multiple addresses then they should all be mapped to the same `MappedResource`, with mirrors and everything.
So saying I map the same executable FD in two base addresses, say 0xA'0000 and 0xB'0000, and I execute both regions, I would assume the cache would get filled for both but going to the same backing resource. If relocations differ then obviously copying in to the active codebuffer would cause it to differ.

neobrain:
Just for clarity, note the PR description tries to distinguish between "library base address" and "file mapping base address". The former describes the virtual address that the ELF header is mapped to (file offset = 0), whereas the latter is the base address given to any mmap call (with file offset possibly non-zero). (And then there's the term "base file mapping", which is the file mapping that is mapped to the library base address)

> What do you mean by this? If !MAP_ANONYMOUS then fd + mmap offset argument will reliably give the offset from the base of the file mapped.

It's a different address space. Relative **file offsets** aren't the same as relative **virtual address offsets**, but FEXOfflineCompiler ultimately needs virtual address offsets (relative to the base file mapping).

I was initially (>6 months ago) hoping relative file offsets could be used but never could get it working. It would fall apart in practice due to ambiguities coming up either when writing code maps or when FEXOfflineCompiler would have to process codemap data. Bugs arising from such ambiguities would be laborious to debug. Sadly I didn't manage to collect a list of examples that can illustrate this (and LinuxSyscalls doesn't lend itself to making unit tests), so this would need to be revisited if we need specific rationale to move forward (doing so would likely take at least a few weeks until an alternative is implemented and proven as robust as what I have today, if possible at all).

To illustrate just one type of the problems that came up: FEXOfflineCompiler needs to unambiguously map any recorded offsets back to the virtual address. This is impossible to decide generally (since a single chunk of file data may be mapped by multiple file sections). Perhaps in practice it could be disambiguated purely on the FEXOfflineCompiler side by factoring in the section access flags there, under the assumption that executable code is only ever mapped once as executable (though this is *also* something I tried before moving to virtual address offsets).

> if we're tracking files by inode then even if a file is mapped to multiple addresses then they should all be mapped to the same MappedResource, with mirrors and everything

That *is* what's happening today. The problem is that FEX can't reliably figure out where the `file offset = 0` mapping is if we do this, because there might be more than one.

> So saying I map the same executable FD in two base addresses, say 0xA'0000 and 0xB'0000, and I execute both regions, I would assume the cache would get filled for both but going to the same backing resource

Is the scenario you're outlining that two different sections of the same file are mapped without anything special happening or that the same file offset is mapped twice at 0xa0000 and 0xb0000? The former is not the problem described by this PR, and for the latter the problem is that caches aren't managed per section but per-file. Per-section caches are an idea I considered at some point, and I would have to revisit why I went for file-based caching instead.

unittests/APITests/FileMappingBaseAddress.cpp
```diff
@@ -0,0 +1,83 @@
...
+  fextl::vector<Elf64_Phdr> Headers = {
+    {.p_type = PT_LOAD, .p_flags = PF_R | PF_X, .p_offset = 0x0000, .p_vaddr = 0x0000, .p_paddr = 0x0000, .p_filesz = 0x00040d, .p_memsz = 0x00040d},
+    {.p_type = PT_LOAD, .p_flags = PF_R, .p_offset = 0x1000, .p_vaddr = 0x1000, .p_paddr = 0x1000, .p_filesz = 0x00036c, .p_memsz = 0x00036c},
+    {.p_type = PT_LOAD, .p_flags = PF_W, .p_offset = 0x1dc8, .p_vaddr = 0x2dc8, .p_paddr = 0x2dc8, .p_filesz = 0x000238, .p_memsz = 0x000240},
+  };
+
+  fextl::vector<Mapping> Mappings = {
+    {.Addr = BaseAddr + 0x1000, .Size = 0x1000, .FileOffset = 0x1000, .Flags = PF_R},
+    {.Addr = BaseAddr + 0x2000, .Size = 0x1000, .FileOffset = 0x1000, .Flags = PF_W},
+    // TODO: How does this work?
```
neobrain: Should be removed before merge

unittests/APITests/FileMappingBaseAddress.cpp
```diff
@@ -0,0 +1,83 @@
...
+  uint64_t BaseAddr = 0x123400000;
+
+  fextl::vector<Elf64_Phdr> Headers = {
+    {.p_type = PT_LOAD, .p_offset = 0x00000, .p_vaddr = 0x00000, .p_paddr = 0x00000, .p_filesz = 0x7bdd5, .p_memsz = 0x7bdd5},
+    {.p_type = PT_LOAD, .p_offset = 0x7c000, .p_vaddr = 0x7c000, .p_paddr = 0x7c000, .p_filesz = 0x6f3a8, .p_memsz = 0x6f3a8},
+    {.p_type = PT_LOAD, .p_offset = 0xebbd0, .p_vaddr = 0xecbd0, .p_paddr = 0xecbd0, .p_filesz = 0x434, .p_memsz = 0x440},
+  };
+
+  fextl::vector<Mapping> Mappings = {
+    // {.Addr = BaseAddr, .Size = 0x7c000, .FileOffset = 0x00000},
```
neobrain: Should be removed before merge

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -383,19 +399,50 @@ std::optional<SyscallHandler::LateApplyExtendedVolatileMetadata> SyscallHandler:
...
+    if (ResourceIt == ResourceEnd || MappedELFHeaderAgain) {
+      // Create a new MappedResource for previously unseen file and for re-mappings of an ELF header
+      ResourceIt = VMATracking.InsertMappedResource(mrid, {nullptr, nullptr, 0});
+      ResourceIt->second.Iterator = ResourceIt;
+      Inserted = true;
+    }
+    Resource = &ResourceIt->second;
 
+    // Only handle FDs that are backed by regular files that are executable
+    if (PathLength != -1 && S_ISREG(buf.st_mode) && (buf.st_mode & S_IXUSR)) {
```
Sonicadvance1, line 420: Should this entire section of code only be enabled if caching is enabled? Right now it seems like it does a bunch of work that is unncessary when caching isn't going to be used.

neobrain: I realized ReadELFHeaders was previously called for the first mapping of *any* file, even when it was mapped from a non-zero offset. This is useful for the extra-assertion, but could indeed add unneeded overhead. I've changed the code to only try ReadELFHeaders when mapping from offset 0.

The remaining work should be negligible:
* For non-ELF files…
  * … on first map if offset == 0, ReadELFHeaders will return immediately after reading a few bytes from the already-mapped file header (no extra syscalls)
  * … on first map if offset != 0, no extra work is done
  * … on subequent maps, no work is added since `ProgramHeaders.empty() == true`
* For ELF files…
  * … on first map, the full ReadELFHeaders logic will run but the emulated application is most likely about to read the file anyway (using slow, emulated x86 logic)
  * … on subsequent maps, we run `InferMappingBaseAddress` almost always exactly once, so it's just a linear scan over the saved ELF headers

We could further constrain the logic to only run when cache-related settings are enabled as suggested, but I think it's better to keep the logic uniform given the light overhead. Conditional enablement would change how MappedResources are grouped, which then has ripple effects on the VMATracking logic.

unittests/APITests/FileMappingBaseAddress.cpp
neobrain: (For reference, this was just the bss section; not relevant for testing)

neobrain:
Should be good to go now!

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
Sonicadvance1: This decision needs to be re-evaluated. It's breaking thunks with Silksong's Linux build.
```
D 2644474|2644474 7425.508 LoadLib: libGL -> /usr/lib/aarch64-linux-gnu/fex-emu/HostThunks/libGL-host.so
D 2644474|2644474 7425.511 Loaded 6209 syms
D 2644474|2644474 7425.511 Thunks: Adding host trampoline for guest function 0x7ffff55f2800 via unpacker 0x7ffff55fd130
D 2644474|2644474 7425.511 Thunks: Adding host trampoline for guest function 0x7ffff5109a50 via unpacker 0x7ffff55fd150
D 2644474|2644474 7425.511 Thunks: Adding host trampoline for guest function 0x7ffff51046c0 via unpacker 0x7ffff55fd170
D 2644474|2644474 7425.511 Thunks: Adding host trampoline for guest function 0x7ffff50ed360 via unpacker 0x7ffff55fd190
A 2644474|2644474 7426.343 SyscallsSMCTracking.cpp:450, TrackMmap: ERROR: Could not find base for file mapping at 0x7ffff282a000 (offset 0x1a000)
```

neobrain: This was a minor issue already fixed in my WIP branch, see https://github.com/neobrain/FEX/commit/e8a537b997ca3030bfeaffd81fb0fa8754550e91 . Will upstream a patch later today.

neobrain: BTW if the same error message comes up again elsewhere, the most useful information to have is the full list of mappings in the failing process (gdb `info proc mappings` or similar). From that we can infer what's going on and add a unit test accordingly.

PR #4981, opened 2025-10-16
Revert quiet/signaling nan detection

pmatos:
This work broke Bayonetta startup. I am working on finding a fix for it, but if in the meantime we want to have a cleaner main - lets land this revert. I hope to get a fix for it in the next couple of days.

If we don't land this before the fix. I will close this and open a new PR for the fix.


pmatos:
This is unrelated to the reduced precision code path - the issue is with full precision btw.

PR #4985, opened 2025-10-17
Support (downstream) kernel-side unaligned atomic handling (The rebase sequel)

Sonicadvance1:
Just support the downstream kernel patches for unaligned atomic handling.


Sonicadvance1:
Slamming this in.

PR #4987, opened 2025-10-17
FEXCore/Config: Support fetching config options as a single shot

Sonicadvance1:
There's quite a few locations where we fetch a config value once, but the current define always defines a variable so it can't be used inline in if-statements. This usually leaves us with a long-lived variable that exists far beyond what is necessary, or makes a single if-statement turn in to a two liner.

Define a new define that creates an anonymous variable that can be checked directly inside of if-statements or passed as arguments or whatever is necessary. It still does a map lookup for each variable fetch as usual, so regular caution still needs to be used.

eg:
`if (FEX_CONFIG_OPT_SINGLE(MULTIBLOCK)) {`
`LogMan::Msg::DFmt("Inst: {}", FEX_CONFIG_OPT_SINGLE(MAXINST));`

eg to not do: (Does a map lookup for each iteration)
```
for (size_t i = 0; i < FEX_CONFIG_OPT_SINGLE(MAXINST); ++i) {
```

Instead do something like this in this case?
```
for (size_t i = 0, max = FEX_CONFIG_OPT_SINGLE(MAXINST); i < max; ++i) {
```

### DIFF
--- a/FEXCore/include/FEXCore/Config/Config.h
+++ b/FEXCore/include/FEXCore/Config/Config.h
@@ -123,6 +123,12 @@ namespace DefaultValues {
     FEXCore::Config::DefaultValues::enum                                    \
   }
 
+// Get a configuration option as a single-shot.
+#define FEX_CONFIG_OPT_SINGLE(enum)                                    \
+  FEXCore::Config::Value<FEXCore::Config::DefaultValues::Type::enum> { \
+    FEXCore::Config::CONFIG_##enum,                                    \
+    FEXCore::Config::DefaultValues::enum                               \
+  }()
 #undef P
 } // namespace DefaultValues
 


FEXCore/include/FEXCore/Config/Config.h
```diff
@@ -123,6 +123,12 @@ namespace DefaultValues {
     FEXCore::Config::DefaultValues::enum                                    \
   }
 
+// Get a configuration option as a single-shot.
+#define FEX_CONFIG_OPT_SINGLE(enum)                                    \
```
neobrain, line 127: I'm not fan of adding a separate macro for such a small gain. It doesn't seem worth having to look up which macro is the right one each time a config check is needed (since `FEX_CONFIG_OPT` and `FEX_CONFIG_OPT_SINGLE` aren't exactly the most distinguishable names).

The way this macro is implemented also seems likely to produce bad error messages on syntax errors (which are easier to make with e.g. the if/for examples you gave), so it seems preferable to make people stick to the statement-syntax instead at the expense of an extra line of code.

neobrain: A vaguely thought-through alternative might be to get rid of the macros entirely. A templated type `ConfigFetcher<FEXCore::Config::CONFIG_MULTIBLOCK>` could be implemented that can either be used in variable declarations or in expressions. We'd need a way to map a `CONFIG_` value to its type (pretty easy) and we'd need to move the default value lookup to that type's destructor.

neobrain: Got this implemented as a prototype, will look into PRing it later this week.

neobrain: Submitted in #5050.

neobrain:
Superseded by the now-merged https://github.com/FEX-Emu/FEX/pull/5050 .

PR #4991, opened 2025-10-20
Removes some hardcoded 4096 constants

Sonicadvance1:
Use our defined variable instead.

### DIFF
--- a/Source/Tools/CommonTools/Linux/Utils/ELFContainer.cpp
+++ b/Source/Tools/CommonTools/Linux/Utils/ELFContainer.cpp
@@ -209,7 +209,7 @@ bool ELFContainer::LoadELF_32() {
   DynamicProgram = Header._32.e_type != ET_EXEC;
 
   // Default BRK size
-  BRKSize = 4096;
+  BRKSize = FEXCore::Utils::FEX_PAGE_SIZE;
 
   return true;
 }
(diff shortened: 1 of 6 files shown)


Source/Common/Config.cpp
```diff
@@ -91,7 +91,7 @@ static char* SaveLayerToJSON(char* JsonBuffer, const FEXCore::Config::Layer* Lay
 }
 
 void SaveLayerToJSON(const fextl::string& Filename, const FEXCore::Config::Layer* Layer, const fextl::unordered_map<fextl::string, bool>& HostLibs) {
-  char Buffer[4096];
+  char Buffer[FEXCore::Utils::FEX_PAGE_SIZE];
```
bylaws: I feel as if not using page size makes more sense for this specific case

Sonicadvance1: Sure. Removed.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -47,7 +48,7 @@ static bool LoadFileImpl(T& Data, const fextl::string& Filepath, size_t FixedSiz
   } else {
     // The file is either empty or its size is unknown (e.g. procfs data).
     // Try reading in chunks instead
-    constexpr size_t READ_SIZE = 4096;
+    constexpr size_t READ_SIZE = FEXCore::Utils::FEX_PAGE_SIZE;
```
neobrain: Another instance where using the PAGE_SIZE falsely suggests an intentional choice when this code does not relate to page sizes (the chunk size is arbitrary).

Sonicadvance1: Sure. Removed.

FEXCore/Source/Utils/FileLoading.cpp
```diff
@@ -1,5 +1,6 @@
 // SPDX-License-Identifier: MIT
 #include <FEXCore/Utils/FileLoading.h>
+#include <FEXCore/Utils/TypeDefines.h>
```
neobrain: This header isn't used anymore now

Sonicadvance1: Removed.

PR #4992, opened 2025-10-21
FEXCore: Remove Paranoid TSO mode.

Sonicadvance1:
This mode has been broken for a long time because it's mostly untested. Barriers, and backpatching while slow have proven that they work. Maintain the one TSO path, at least until all ARM hardware gains support for x86-TSO memory model mode.


bylaws:
I don't think this makes sense, with kernel side atomic handling ParanoidTSO is often faster than non paranoid tso. I understand removing it for the emulation case however. But I would not want to remove outright because I use it often to test with kernel handling

Sonicadvance1:
> I don't think this makes sense, with kernel side atomic handling ParanoidTSO is often faster than non paranoid tso. I understand removing it for the emulation case however. But I would not want to remove outright because I use it often to test with kernel handling

Do you mean the FEX side code is faster with paranoid TSO enabled? Because it's roughly equivalent code on the FEX side, just some NOPs added and forcing vector loadstores in to GPRs. I don't see how the FEX side code could cause anything other than noise level performance difference.

Or do you mean that handling the unaligned atomic in the kernel without backpatching is faster?
Because we could still retain an option that when kernel unaligned atomic handling is enabled, to disable backpatching or not.

bylaws:
I mean without backpatching yeah

Sonicadvance1:
> I mean without backpatching yeah

Then when the kernel unaligned atomic handling is merged, it should just gain an option for controlling kernel-side backpatching, optionally controlling if NOPs are generated in the JIT or not.

bylaws:
We should be able to remove some atomic handling cases from our emulation too now

Sonicadvance1:
> We should be able to remove some atomic handling cases from our emulation too now

Indeed, `HandleAtomicVectorStore` was removed. The other handlers are still used by various bits.

PR #4995, opened 2025-10-24
FEXConfig: add an option to disable L2 cache on CPU tab

hrw:
The DisableL2Cache option is CPU related so show it in proper place.

### DIFF
--- a/Source/Tools/FEXConfig/main.qml
+++ b/Source/Tools/FEXConfig/main.qml
@@ -719,6 +719,11 @@ ApplicationWindow {
                 text: qsTr("Disable JIT optimization passes")
                 config: "O0"
             }
+
+            ConfigCheckBox {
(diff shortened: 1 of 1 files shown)


neobrain:
Honestly this seems preliminary, so it's a -1 from me.

There's already a checkbox for this setting in the Advanced tab, and this it's too niche to have a checkbox on the "main" level (where options should be at least somewhat comprehensible for technical users; a tweak to a deep inner implementation detail is the opposite!).

I'm open to revisiting this though once we've decided whether to enable the setting by default or not; if there are prominent use cases that require toggling the setting on a regular base, this would be a good reason to add a second check box.

neobrain:
Closing for now, let's revisit it at a later point.

Thanks for the contribution regardless!

hrw:
It is perfectly fine for me anyway ;D

PR #4993, opened 2025-10-22
FEXCore: Adds some more per-thread stats.

Sonicadvance1:
- Cache miss counts
  - Useful for determining if L2 cache or dynamic cache could help
- Cache read/write lock contention times
  - Useful to see if threads are blocking each other on contention
  - Read lock is the case where a read-lock is beneficial, even if we
    currently use a write lock.
- JIT count
  - Useful to see if any new JIT blocks are generating

eg:
Middle of compiling
```
       JIT Time: 277.647490 ms/second (2.31 percent)
    Signal Time: 0.139980 ms/second (0.00 percent)
     SIGBUS Cnt: 294 (295.820556 per second)
        SMC Cnt: 0
  Softfloat Cnt: 27160
  CacheMiss Cnt: 21633 (21766.959497 per second)
    $RDLck Time: 0.934010 ms/second (0.01 percent)
    $WRLck Time: 2.319050 ms/second (0.02 percent)
        JIT Cnt: 5697 (5732.277920 percent)
```

Idle:
```
       JIT Time: 1.396660 ms/second (0.01 percent)
    Signal Time: 0.000000 ms/second (0.00 percent)
     SIGBUS Cnt: 0 (0.000000 per second)
        SMC Cnt: 0
  Softfloat Cnt: 170648
  CacheMiss Cnt: 34 (34.171477 per second)
    $RDLck Time: 0.004350 ms/second (0.00 percent)
    $WRLck Time: 0.002880 ms/second (0.00 percent)
        JIT Cnt: 8 (8.040348 percent)
```

Necrodancer bad case:
```
Total (1000 millisecond sample period):
       JIT Time: 672.021240 ms/second (5.60 percent)
    Signal Time: 1.884800 ms/second (0.02 percent)
     SIGBUS Cnt: 4 (4.005191 per second)
        SMC Cnt: 3
  Softfloat Cnt: 0
  CacheMiss Cnt: 15219 (15238.751873 per second)
    $RDLck Time: 0.983350 ms/second (0.01 percent)
    $WRLck Time: 0.680150 ms/second (0.01 percent)
        JIT Cnt: 5334 (5340.922695 percent)
```

PR #4998, opened 2025-10-26
LookupCache: Convert mutex to new WritePriorityMutex 

Sonicadvance1:
Reduces stutters, lowers lock-contention overhead.

### DIFF
--- a/FEXCore/Source/Interface/Core/LookupCache.h
+++ b/FEXCore/Source/Interface/Core/LookupCache.h
@@ -3,6 +3,8 @@
 #include "Interface/Context/Context.h"
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/Utils/SHMStats.h>
+#include "Utils/WritePriorityMutex.h"
+
 #include <FEXCore/fextl/map.h>
 #include <FEXCore/fextl/memory_resource.h>
 #include <FEXCore/fextl/robin_map.h>
@@ -15,22 +17,35 @@
 #include <mutex>
 
 namespace FEXCore {
-
 struct LookupCacheWriteLockToken {
 private:
   // Only constructible by GuestToHostMap
   friend struct GuestToHostMap;
-  LookupCacheWriteLockToken(std::mutex& Mutex)
+  LookupCacheWriteLockToken(FEXCore::Utils::WritePriorityMutex::Mutex& Mutex)
     : Lock {Mutex} {}
-  std::lock_guard<std::mutex> Lock;
+  std::lock_guard<FEXCore::Utils::WritePriorityMutex::Mutex> Lock;
+};
+
+struct LookupCacheReadLockToken {
+private:
+  // Only constructible by GuestToHostMap
+  friend struct GuestToHostMap;
+  LookupCacheReadLockToken(FEXCore::Utils::WritePriorityMutex::Mutex& Mutex)
+    : Lock {Mutex} {}
+  std::shared_lock<FEXCore::Utils::WritePriorityMutex::Mutex> Lock;
 };
 
 struct GuestToHostMap {
-  std::mutex WriteLock;
+  FEXCore::Utils::WritePriorityMutex::Mutex Lock {};
 
   [[nodiscard]]
   LookupCacheWriteLockToken AcquireWriteLock() {
-    return LookupCacheWriteLockToken {WriteLock};
+    return LookupCacheWriteLockToken {Lock};
+  }
+
+  [[nodiscard]]
+  LookupCacheReadLockToken AcquireReadLock() {
+    return LookupCacheReadLockToken {Lock};
   }
 
   struct BlockLinkTag {
@@ -75,7 +90,7 @@ struct GuestToHostMap {
     BlockList[Address] = (uintptr_t)HostCode;
   }
 
-  std::optional<uintptr_t> FindBlock(uint64_t Address, const LookupCacheWriteLockToken&) {
+  std::optional<uintptr_t> FindBlock(uint64_t Address, const LookupCacheReadLockToken&) {
     auto HostCode = BlockList.find(Address);
     if (HostCode == BlockList.end()) {
       return std::nullopt;
@@ -144,7 +159,7 @@ public:
     {
       std::optional<FEXCore::SHMStats::AccumulationBlock<uint64_t>> LockTime(
         Thread->ThreadStats ? &Thread->ThreadStats->AccumulatedCacheReadLockTime : nullptr);
-      auto lk = Shared->AcquireWriteLock();
+      auto lk = Shared->AcquireReadLock();
       LockTime.reset();
 
       if (!DisableL2Cache()) {
@@ -302,7 +317,7 @@ public:
   }
 
   void ClearCache(const LookupCacheWriteLockToken&);
-  void ClearL2Cache(const LookupCacheWriteLockToken&);
+  void ClearL2Cache(const LookupCacheReadLockToken&);
   void ClearThreadLocalCaches(const LookupCacheWriteLockToken&);
 
   uintptr_t GetL1Pointer() const {
@@ -330,7 +345,7 @@ public:
   }
 
 private:
-  void CacheBlockMapping(uint64_t Address, uintptr_t HostCode, const LookupCacheWriteLockToken& lk) {
+  void CacheBlockMapping(uint64_t Address, uintptr_t HostCode, const LookupCacheReadLockToken& lk) {
     // Do L1
     auto& L1Entry = reinterpret_cast<LookupCacheEntry*>(L1Pointer)[Address & L1PointerMask];
     L1Entry.GuestCode = Address;
--- a/FEXCore/Source/Interface/Core/LookupCache.cpp
+++ b/FEXCore/Source/Interface/Core/LookupCache.cpp
@@ -76,7 +76,7 @@ LookupCache::~LookupCache() {
   // These will get freed when their memory allocators are deallocated.
 }
 
-void LookupCache::ClearL2Cache(const FEXCore::LookupCacheWriteLockToken& lk) {
+void LookupCache::ClearL2Cache(const FEXCore::LookupCacheReadLockToken& lk) {
   // Clear out the page memory
   // PagePointer and PageMemory are sequential with each other. Clear both at once.
   FEXCore::Allocator::VirtualDontNeed(reinterpret_cast<void*>(PagePointer),


Sonicadvance1:
Rebased since #4993 was merged.

Sonicadvance1:
> I'm curious, how does this stack up to shared_mutex in measured perf for the GuestToHostMap lock?

My code is faster in the read-only uncontested case, I didn't even attempt in the unique contested case because it'll be a nightmare without priority.

bylaws:
> > I'm curious, how does this stack up to shared_mutex in measured perf for the GuestToHostMap lock?
> 
> My code is faster in the read-only uncontested case, I didn't even attempt in the unique contested case because it'll be a nightmare without priority.

Right but like how much faster I suppose :)

Sonicadvance1:
> > > I'm curious, how does this stack up to shared_mutex in measured perf for the GuestToHostMap lock?
> > 
> > 
> > My code is faster in the read-only uncontested case, I didn't even attempt in the unique contested case because it'll be a nightmare without priority.
> 
> Right but like how much faster I suppose :)

In a fairly idle read-only scene, I got 9.5% to 21% faster than pthreads. Which is okay. Bigger win is that our power-consumption and latency will likely be lower due to the use of WFE and a /very/ small back-off.

bylaws:
ooh, what game was this?

Sonicadvance1:
> ooh, what game was this?

This was Death Stranding, it wasn't an FPS improvement but the amount of time reduced in lock-contention.

FEXCore/Source/Utils/SpinWaitLock.h
```diff
@@ -184,6 +186,26 @@ template bool Wait<uint16_t>(uint16_t*, uint16_t, const std::chrono::nanoseconds
 template bool Wait<uint32_t>(uint32_t*, uint32_t, const std::chrono::nanoseconds&);
 template bool Wait<uint64_t>(uint64_t*, uint64_t, const std::chrono::nanoseconds&);
 
+template<typename T>
+static inline T OneShotWFEBitComparison(T* Futex, T Mask, T Comp) {
+  std::atomic<T>* AtomicFutex = reinterpret_cast<std::atomic<T>*>(Futex);
```
bylaws: atomic_ref perhaps?

Sonicadvance1: Oop, you right. This commit was based on code before everything switched to atomic_ref. Fixed.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -15,24 +17,48 @@
 #include <mutex>
 
 namespace FEXCore {
-
 struct LookupCacheWriteLockToken {
+  // Move-only type
+  LookupCacheWriteLockToken(const LookupCacheWriteLockToken&) = delete;
+  LookupCacheWriteLockToken& operator=(const LookupCacheWriteLockToken&) = delete;
+  LookupCacheWriteLockToken(LookupCacheWriteLockToken&& rhs) = delete;
+  LookupCacheWriteLockToken& operator=(LookupCacheWriteLockToken&&) = delete;
```
neobrain: It's not actually a move-only type since you made it non-moveable. `lock_guard` itself already disables all these constructors though, so the added constructors are unnecessary:

```suggestion
```

If you wanted to make the type moveable, replacing `lock_guard` with `unique_lock` (while still removing all manual constructor declarations) would achieve that as well.

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -15,24 +17,48 @@
...
+  std::lock_guard<FEXCore::Utils::WritePriorityMutex::Mutex> Lock;
+};
+
+struct LookupCacheReadLockToken {
+  // Move-only type
+  LookupCacheReadLockToken(const FEXCore::Utils::WritePriorityMutex::Mutex&) = delete;
+  LookupCacheReadLockToken& operator=(const FEXCore::Utils::WritePriorityMutex::Mutex&) = delete;
+  LookupCacheReadLockToken(FEXCore::Utils::WritePriorityMutex::Mutex&& rhs) = delete;
+  LookupCacheReadLockToken& operator=(FEXCore::Utils::WritePriorityMutex::Mutex&&) = delete;
+
```
neobrain: Same as above, though this time it's actually moveable:

```suggestion
  // Move-only type
```

FEXCore/Source/Interface/Core/LookupCache.h
```diff
@@ -15,24 +17,48 @@
...
-  LookupCacheWriteLockToken(std::mutex& Mutex)
+  LookupCacheReadLockToken(FEXCore::Utils::WritePriorityMutex::Mutex& Mutex)
     : Lock {Mutex} {}
-  std::lock_guard<std::mutex> Lock;
+  std::shared_lock<FEXCore::Utils::WritePriorityMutex::Mutex> Lock;
 };
 
 struct GuestToHostMap {
-  std::mutex WriteLock;
+  FEXCore::Utils::WritePriorityMutex::Mutex WriteLock {};
```
neobrain: Calling it WriteLock when it's actually used to guard reads too is misleading:
```suggestion
  FEXCore::Utils::WritePriorityMutex::Mutex Lock {};
```

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+#include <FEXCore/Utils/LogManager.h>
+#endif
+
+#include "Utils/SpinWaitLock.h"
+
+namespace FEXCore::Utils::WritePriorityMutex {
+
+// A mutex that optimizes around the idea that read-locks should be insignificantly low-cost.
+// While a write-lock while rare, should be given priority because its work needs to get in and out to reduce contention on shared locks.
```
neobrain: > // While a write-lock while rare

Looks like a typo. The whole sentence has a bit wonky grammar though, could you double-check this?

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+// Goals:
+//   - One or two 32-bit atomic CAS in the read-contended or uncontended read-lock case, respectively.
```
neobrain: What's a "read-lock case"? Sounds like trying to read data protected by a locked mutex, but that seems to be a wrong interpretation since you qualified it as "uncontended".

Overall the entire comment block seems to suffer from sometimes using the term "write-lock"/"read-lock" to refer to the mutex itself while using it to describe contention scenarios other times without giving clear context.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+//   - A 32-bit atomic CAS in the uncontended write-lock case.
+//   - An WFE backoff in both contended read/write cases, assuming write-lock will unlock quickly.
+//   - A kernel futex fallback (with WAIT_BITSET!) to allow write-write handoff in the write-contended case.
+//     - Write-write handoff avoids write-stampede
```
neobrain: What's write-stampede? Doesn't seem like a standard term from a quick web-search.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+    while (true) {
+      bool Sleep = false;
+
+      do {
+        if ((Expected & WRITE_OWNED_BIT) == 0 && (Expected & READ_OWNER_COUNT_MASK) == 0) {
+          // If not write-owned, and no read-owners, try to acquire.
+
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+          LOGMAN_THROW_A_FMT((Expected & WRITE_WAITER_COUNT_MASK) != 0, "Underflow in write-waiters!");
+#endif
```
neobrain: `LOGMAN_THROW_A_FMT` trivially optimizes to a no-op in non-assertion builds:
```suggestion
          LOGMAN_THROW_A_FMT((Expected & WRITE_WAITER_COUNT_MASK) != 0, "Underflow in write-waiters!");
```

Same below.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+  Mutex() = default;
+
+  // Move-only type
+  Mutex(const Mutex&) = delete;
+  Mutex& operator=(const Mutex&) = delete;
+  Mutex(Mutex&& rhs) = delete;
+  Mutex& operator=(Mutex&&) = delete;
+
+  void lock() {
+    // Try an uncontended lock first.
```
neobrain: Non-blocking is the better term here. "uncontended" is a property of the underlying mutex (referring to the number of threads that simultaneously try to lock it), which isn't what you wanted to describe.

```suggestion
    // Try a non-blocking lock first
```

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+  void FutexWakeReaders() {
+    // Wake all readers.
+    ::syscall(SYS_futex, &Futex, FUTEX_PRIVATE_FLAG | FUTEX_WAKE_BITSET, INT_MAX, nullptr, nullptr, FUTEX_BITSET_WAIT_READERS);
+  }
+
+  // Reuse the SpinWaitLock WFE implementations for read/write lock acquiring with WFE.
+  // Can't reuse the spin-lock directly as some bit-representations are different.
+  // WFE-write-lock is less likely to occur the more read-lock threads are participating. Can still occur so good to try.
+  // WFE-read-lock is actually quite likely to succeed.
+  bool Attempt_WFE_WriteLock() {
```
neobrain, line 264: Return value isn't documented.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+
+  // Only spin on WFE for 0.01ms (10k ns).
+  constexpr static uint64_t CYCLECOUNT_DIVISOR = 1'000'000'000ULL / 10'000U;
+
+  // Layout:
+  //    Bits[31]: Write-lock bit.
+  // Bits[30:16]: Write-waiter count.
+  //    Bits[15]: Read-waiter bit.
+  //  Bits[14:0]: Read-owner count.
+  uint32_t Futex {};
```
neobrain, line 350: Have you considered using bitfields here?
```suggestion
  struct {
    uint32_t WriteLockBit : 1;
    uint32_t WriteWaiterCount : 15;
    uint32_t ReadWaiterBit : 1;
    uint32_t ReadOwnerCount : 15
  } Futex {};
```
That'll clean up the bitmasking logic and allows for dropping the comments and some of the magic constants above.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+    if ((Desired & WRITE_WAITER_COUNT_MASK) && (Desired & READ_OWNER_COUNT_MASK) == 0) {
+      FutexWakeWriter();
+    }
+  }
+
+  bool try_lock() {
+    auto AtomicFutex = std::atomic_ref<uint32_t>(Futex);
+
+    // Expected is zero
+    uint32_t Expected {};
```
neobrain: No need to state the obvious if you just use `= 0`:
```suggestion
    uint32_t Expected = 0;
```

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+#include "Utils/SpinWaitLock.h"
+
+namespace FEXCore::Utils::WritePriorityMutex {
+
+// A mutex that optimizes around the idea that read-locks should be insignificantly low-cost.
+// While a write-lock while rare, should be given priority because its work needs to get in and out to reduce contention on shared locks.
+// Goals:
+//   - One or two 32-bit atomic CAS in the read-contended or uncontended read-lock case, respectively.
+//   - A 32-bit atomic CAS in the uncontended write-lock case.
+//   - An WFE backoff in both contended read/write cases, assuming write-lock will unlock quickly.
```
neobrain:
```suggestion
//   - A WFE backoff in both contended read/write cases, assuming write-lock will unlock quickly.
```

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+    uint32_t Expected = AtomicFutex.load(std::memory_order_relaxed);
+    uint32_t Desired {};
+    do {
+      // Increment the number of write waiters.
+      Desired = Expected + WRITE_WAITER_INCREMENT;
+
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+      LOGMAN_THROW_A_FMT((Desired & WRITE_WAITER_COUNT_MASK) != 0, "Overflow in write-waiters!");
+#endif
+    } while (AtomicFutex.compare_exchange_strong(Expected, Desired, std::memory_order_acq_rel, std::memory_order_acquire) == false);
```
bylaws, line 67: fetch_add?

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+
+    // Thread added to waiter list.
+    Expected = Desired;
+
+    while (true) {
```
bylaws, line 76: I think you could make this cleaner as something like

while (true) {
   if (...)
{
if (...cas.. successful) return
} else {
Futexwait()
Expected = ..
}
}

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+    uint32_t Desired {};
+    do {
+#if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
+      LOGMAN_THROW_A_FMT((Expected & WRITE_OWNED_BIT) != WRITE_OWNED_BIT, "Trying to read-unlock something write-locked!");
+      LOGMAN_THROW_A_FMT((Expected & READ_OWNER_COUNT_MASK) != 0, "Trying to read-unlock something not read-locked!");
+#endif
+
+      // Decrement the shared counter.
+      Desired = Expected - READ_OWNER_INCREMENT;
+    } while (AtomicFutex.compare_exchange_strong(Expected, Desired, std::memory_order_acq_rel, std::memory_order_acquire) == false);
```
bylaws, line 194: fetch_sub?

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -0,0 +1,398 @@
...
+//   - A kernel futex fallback (with WAIT_BITSET!) to allow write-write handoff in the write-contended case.
+//     - Write-write handoff avoids write-stampede
+//     - Or write-read hand-off in the non-write-contended case. (Read Stampede)
+//
+// Non-Goals:
+//  - No FIFO/Fair scheduling.
+//  - No Priority-Inheritence.
+//  - No recursive mutex support.
+//  - No safety rails in release builds.
+#if !defined(_WIN32)
```
bylaws: Move the ifdef above the comment I think

FEXCore/Source/Interface/Core/LookupCache.h
Sonicadvance1: I just removed these, and the others, Added them while tinkering and it's unnecessary.

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: Definitely a typo, `While a write-lock is rare,` fixed.

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: This pattern works with my brain juices better so I don't really want to tinker with it. Doing some tricky stuff with having the CAS's returned data falling through to the return assert check and the follow up futex.

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: I thought about it, but then decided against it. I'm not going to go back and change it to use bitmasks now.

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: I don't do the writing good. It makes sense in my brain when I read it. It's describing scenarios and performance characteristics of the underlying mutex implementation.

neobrain: Maybe you could give this another look?

I would propose a better wording, but I don't understand what the text is trying to say.

FEXCore/Source/Utils/WritePriorityMutex.h
neobrain: Is there a reason not to do it? It's a straightforward change and given how much it would simplify the implementation I'm not sure why we wouldn't change it.

bylaws: fwiw I think I prefer as is done rn personally, especially with how we're using ops like fetch_add etc

neobrain: > fwiw I think I prefer as is done rn personally, especially with how we're using ops like fetch_add etc

Yeah, that's a good point. You'd have to turn the `fetch_add` into something silly like `Expected.raw = AtomicFutex.fetch_add(FutexData { .write_waiters = 1 }.raw)`. At that point it's not such a clear improvement overall after all.

Let's stick to the current implementation then.

PR #5004, opened 2025-10-28
Remove InterpretAsFloat from x87StackOptimizationPass

pmatos:
The InterpretAsFloat was never properly made use of. There's a couple of issues that are fixed more easily with this gone, so lets remove it.

If there's a specific optimization that requires this, we can bring it back at a later time. This should not have any effect on the current code generation.


bylaws:
I do want to go through this

PR #5006, opened 2025-10-28
Introduce two-pass code invalidation model

bylaws:
See individual commit messages for details, but speeds up invalidation by an order of magnitude.

### DIFF
--- a/Source/Windows/Common/InvalidationTracker.cpp
+++ b/Source/Windows/Common/InvalidationTracker.cpp
@@ -56,11 +56,7 @@ void InvalidationTracker::HandleMemoryProtectionNotification(uint64_t Address, u
 
   if (NeedsInvalidate) {
     // IntervalsLock cannot be held during invalidation
-    std::scoped_lock Lock(CTX.GetCodeInvalidationMutex());
-    FEXCore::Context::InvalidatedEntryAccumulator Accumulator;
-    for (auto Thread : Threads) {
-      CTX.InvalidateGuestCodeRange(Thread.second, Accumulator, AlignedBase, AlignedSize);
-    }
+    InvalidateIntervalInternal(AlignedBase, AlignedSize);
   }
 }
 
@@ -115,13 +111,8 @@ InvalidationTracker::InvalidateContainingSectionResult InvalidationTracker::Inva
          reinterpret_cast<uint64_t>(Info.AllocationBase) == SectionBase) {
     SectionSize += Info.RegionSize;
   }
-  {
-    std::scoped_lock Lock(CTX.GetCodeInvalidationMutex());
-    FEXCore::Context::InvalidatedEntryAccumulator Accumulator;
-    for (auto Thread : Threads) {
-      CTX.InvalidateGuestCodeRange(Thread.second, Accumulator, SectionBase, SectionSize);
-    }
-  }
+
+  InvalidateIntervalInternal(SectionBase, SectionSize);
 
   if (Free) {
     std::unique_lock Lock(IntervalsLock);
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Exciting! I'll have to take a look at this when I'm not stuck in the optimizations mines (probably early next week or the weekend). In that time it might be worth getting the WritePriorityMutex PR out of the way so these don't conflict with other.

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -873,50 +883,37 @@ uintptr_t ContextImpl::CompileSingleStep(FEXCore::Core::CpuStateFrame* Frame, ui
...
+    } else {
+      it = CodeBufferList.erase(it);
+    }
+  }
+}
 
-  auto lower = CodePages.lower_bound(Start >> 12);
-  auto upper = CodePages.upper_bound((Start + Length - 1) >> 12);
+void ContextImpl::InvalidateThreadCachedCodeRange(FEXCore::Core::InternalThreadState* Thread, uint64_t Start, uint64_t Length) {
+  LogMan::Throw::AFmt(CodeInvalidationMutex.try_lock() == false, "CodeInvalidationMutex needs to be unique_locked here");
```
neobrain: `LOGMAN_THROW_A_FMT` should be preferred since it includes the source location in asserts (and further it lazily evaluates any printed arguments).

(I can't actually think of any reason to use AFmt over the macro at all, perhaps we should remove it entirely.)

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -873,50 +883,37 @@ uintptr_t ContextImpl::CompileSingleStep(FEXCore::Core::CpuStateFrame* Frame, ui
...
+void ContextImpl::InvalidateCodeBuffersCodeRange(uint64_t Start, uint64_t Length) {
+  FEXCORE_PROFILE_SCOPED("InvalidateCodeBuffersCodeRange");
 
-  auto lk = Thread->LookupCache->AcquireWriteLock();
-  auto& CodePages = Thread->LookupCache->Shared->CodePages;
+  LogMan::Throw::AFmt(CodeInvalidationMutex.try_lock() == false, "CodeInvalidationMutex needs to be unique_locked here");
+  std::scoped_lock lk {CodeBufferListLock};
+  auto it = CodeBufferList.begin();
+  while (it != CodeBufferList.end()) {
+    if (auto Strong = it->lock(); Strong) {
```
neobrain: Fun fact, you don't actually need to repeat the variable name as a separate condition:
```suggestion
    if (auto Strong = it->lock()) {
```

neobrain (commented):
Just adding two drive-by comments. Nice speedup :)

Sonicadvance1:
Did testing on this, it consumes some additional memory but it's a necessary evil as it speeds up code invalidation dramatically. Just needs rebase plus the comments.

Sonicadvance1:
```
Before -> After -> unordered_set
Steam: 520MB -> 550MB (+5.7%) -> 497MB (-9.6% Lots of noise in this one because Steam is a very active process)
EnderLilies Linux: 535MB -> 585MB (+9.3%) -> 580MB (-0.8% Basically noise)
God Of War (2018): 476MB -> 589MB (+23.7%) -> 568MB (-3.5% Probably just pooling difference)
Death Stranding: 536MB -> 611MB (+13.9%) -> 595MB (-2.6% Probably just noise from pooling)
```
I think this is a reasonable compromise considering how much we can save by dynamic L1 and disabled L2. We could definitely make up some of the difference by making our pooling structures have the ability to drain out more aggressively.

bylaws:
How does that compare to unordered_set?

bylaws:
I checked against numbers you reported, looks to be roughly half. Seems good then

Sonicadvance1:
Bonus round added to the edited comment. These numbers aren't comparable to the previous numbers that I posted a few days ago.
Mostly just falls within noise on memory usage, but the added benefit of tsl's speed would outweight the small percentages here.

PR #5007, opened 2025-10-29
JIT: Restore behaviour of emitting interrupt checks at every block entry

bylaws:
This is needed to handle suspend in infinite loops that occur as a result of block-size constraints or indirect jumps. Fixes grow home.


bylaws:
> Oops, how did we manage to break that...?

I had thought back-edges was enough to catch all real-world cases I could find. Obviously not :P

PR #5008, opened 2025-10-29
FEXConfig: Fix crash caused by no longer recognized option

neobrain:
This option was removed in #4997 but the corresponding GUI code wasn't updated accordingly.


Sonicadvance1 (approved):
Neat.

neobrain:
Fast tracking this PR since the fix is trivial and FEXConfig can't be used at all without it.

PR #5009, opened 2025-10-30
f80 stack xchg optimization for fast path

pmatos:
We were calling the generic function, making the impl. itself simpler but doing more work than needed in this particular instance.


bylaws (changes requested):
See instcountci ci, this regresses it

pmatos:
> See instcountci ci, this regresses it

huh, thanks for pointing this out. weird. I was pretty sure it was improving it when I started splitting the patches.

pmatos:
> See instcountci ci, this regresses it

Yeah, my mistake sorting out the split. It's updated now.

PR #5011, opened 2025-10-30
Refactoring of storing code in x87 opt. stack pass

pmatos:
Enables memcpy optimization of 80bit floats on reduced precision.

Also uncovered a bug where if we had done 80bit memcpy
optimization, we wouldn't have properly stored the 80bits.
This was caught by the existing tests when we enabled the optimization.

Depends on #5004. (once it lands, I will rebase this one)


FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -290,23 +292,27 @@ class X87StackOptimization final : public Pass {
...
-      , InterpretAsFloat(Float) {}
+      , Source({Size, Source}) {}
     Ref StackDataNode {}; // Reference to the data in the Stack.
                           // This is the source data node in the stack format, possibly converted to 64/80 bits.
     struct StackMemberData final {
       OpSize Size;
       Ref Node;
     };
+
+    static const StackMemberInfo& Invalid() {
```
bylaws: What the benefit of this over ctor?

pmatos: The benefit of this is mostly that it avoid misuse. It's very easy to default to contract an empty StackMemberInfo, which before we removed InterpretAsFloat was initializing all floats as Integers. 

Now it acts as being self-documenting. You call `Invalid()` to create an invalid node rather than creating an empty stack member info.

bylaws: I would make this a constexpr static member variable Invalid then rather than function

pmatos: I don't think it can be constexpr, but I added it as a static member variable.

pmatos: @bylaws let me know if you have further comments on this.

pmatos:
Rebased it on current main to keep it up-to-date.

PR #5017, opened 2025-11-01
ArchHelpers: Adds ESR name helper

Sonicadvance1:
Just helps when an unhandled ESR occurs, it was always a case of needing to go in to the ARM ARM to decode it which was a bit of a pain. Add a textual representation of it.


Source/Tools/LinuxEmulation/ArchHelpers/MContext.h
```diff
@@ -203,9 +203,12 @@ constexpr static uint64_t ESR1_DataAbort_Level_EL2 = 0b01;
 constexpr static uint64_t ESR1_DataAbort_Level_EL1 = 0b10;
 constexpr static uint64_t ESR1_DataAbort_Level_EL0 = 0b11;
 
+const char* GetESRName(uint64_t ESR);
```
lioncash: Perhaps this should be a string_view, since this only ever returns valid strings

Sonicadvance1: Switched over to string_view.

PR #5020, opened 2025-11-03
FEXCore/Allocator: Fixes two bugs

Sonicadvance1:
Fixes a false OOM that could occur due to incorrect FreeSpace tracking, and fixes a nasty corruption that could occur when MAP_FIXED was too large for a VMA region.

Adds unittests for both.

### DIFF
--- a/FEXCore/unittests/APITests/Allocator.cpp
+++ b/FEXCore/unittests/APITests/Allocator.cpp
@@ -34,7 +34,7 @@ TEST_CASE("Allocator - Fixed replacement") {
     REQUIRE(Base == NewBase);
   }
 
-  Alloc::OSAllocator::ReleaseAllocatorWorkaround(Allocator);
+  Alloc::OSAllocator::ReleaseAllocatorWorkaround(std::move(Allocator));
 }
 
 TEST_CASE("Allocator - Non-Fit") {
@@ -62,5 +62,5 @@ TEST_CASE("Allocator - Non-Fit") {
   // - If it errors, then it doesn't fit in the VMA region.
   REQUIRE((NewBase == Base || HasSyscallError(NewBase)));
 
-  Alloc::OSAllocator::ReleaseAllocatorWorkaround(Allocator);
+  Alloc::OSAllocator::ReleaseAllocatorWorkaround(std::move(Allocator));
 }
--- a/FEXCore/Source/Utils/Allocator.cpp
+++ b/FEXCore/Source/Utils/Allocator.cpp
@@ -140,7 +140,7 @@ void ClearHooks() {
   FEXCore::Allocator::mmap = ::mmap;
   FEXCore::Allocator::munmap = ::munmap;
 
-  Alloc::OSAllocator::ReleaseAllocatorWorkaround(Alloc64);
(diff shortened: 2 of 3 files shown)


lioncash (approved):

FEXCore/unittests/APITests/Allocator.cpp
```diff
@@ -0,0 +1,68 @@
...
+  for (size_t i = 0; i < (RegionSize / 4096); ++i) {
+    auto NewBase = Allocator->Mmap(Base, 4096, PROT_NONE, MAP_FIXED | MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    REQUIRE(Base == NewBase);
+  }
+
+  // Leak the memory for now. Crashes if we try.
+  Allocator.release();
+}
+
+TEST_CASE("FlexBitSet - Non-Fit") {
```
neobrain: FlexBitSet?

Sonicadvance1: Oop, started off as me tracking them as FlexBitSet bugs, not allocator bugs. renamed.

FEXCore/unittests/APITests/Allocator.cpp
```diff
@@ -0,0 +1,68 @@
...
+  REQUIRE(!HasSyscallError(Base));
+
+  // Allocate perfectly overlapping pages. Allocate as many pages as the region.
+  // FEX had a bug where the allocator could run out of memory with MAP_FIXED.
+  for (size_t i = 0; i < (RegionSize / 4096); ++i) {
+    auto NewBase = Allocator->Mmap(Base, 4096, PROT_NONE, MAP_FIXED | MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+    REQUIRE(Base == NewBase);
+  }
+
+  // Leak the memory for now. Crashes if we try.
```
neobrain: What's this about? Is this a general issue with the allocator (if yes, why does it not cause issues elsewhere?) or something specific about the test?

Sonicadvance1: General issue with the allocator: https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Utils/Allocator.cpp#L143-L146

Sonicadvance1: I guess that comment also doesn't claim the issue of because the allocator object is intrusively allocated, it tries to double free itself.

neobrain: Perhaps we could move this pattern to a dedicated public interface, so we have a central place for documenting this quirk and so that we can easily find all instances of it? Seems like it's just these three times for now, but they're scattered across two files so it would help build a direct link between these `release` calls.

```
void ReleaseAllocatorWorkaround(fextl::unique_ptr<...> Allocator) {
  // explanatory comment goes here 
  Allocator.release();
}
```

Bit confused about the original comment though - tests aren't affected by lack of static initializers. Is there more than just that to it?

Sonicadvance1: > I guess that comment also doesn't claim the issue of because the allocator object is intrusively allocated, it tries to double free itself.

Moved it to a helper inline function.

PR #5025, opened 2025-11-05
Reimplementation of quiet and signalling nan propagation on x87

pmatos:
Reimplementation of quiet and signalling nan propagation on x87

Initially implemented in #4835,  later reverted in #4981.

Depends on #5011 .


bylaws (changes requested):
Perhaps I'm missing something but I would expect the strict option here to control the silencenan code emission in storestackmem, FromF64_PerserveNan is definitely cheap enough as to not require an option at least. Perhaps it would be nicer to make reduced precision and enum rather than adding a new option

bylaws:
Whats the instcountci blow up if this is enabled btw? I'm curious

pmatos:
> Perhaps I'm missing something but I would expect the strict option here to control the silencenan code emission in storestackmem, FromF64_PerserveNan is definitely cheap enough as to not require an option at least. Perhaps it would be nicer to make reduced precision and enum rather than adding a new option

I understand that from a performance perspective we could do the check in `FromF64_PerserveNan` but it does feel weird to me to do the check sometimes but not others. IMO, it makes more sense to have all the checks behind the same flag.

The downside of making reduced precision an enum is that this is not backwards compatible and will require everyone to update their config files. I think it was initially @Sonicadvance1 who suggested a stricter version of reduced precision in my initial pr. Wonder what he would think of making reduced precision and strenum. :)

pmatos:
> Whats the instcountci blow up if this is enabled btw? I'm curious

I will get these for you today.

bylaws:
I was researching if theres a better way to do this, and came up with below ASM:
```
   movi    v31.4s, #0x40 
    shl     v31.4s, v31.4s, #22
    fcmeq   v1.4s, v0.4s, v0.4s
    orr     v2.8b, v0.8b, v31.8b
    bif     v0.8b, v2.8b, v1.8b     
```
@Sonicadvance1 think we could just enable by default given this?  given in most cases it's effectively 3 insts, could get to two insts with sve (albeit same per mca since fcmuo has much lower throughput)

pmatos:
> I was researching if theres a better way to do this, and came up with below ASM:
> 
> ```
>    movi    v31.4s, #0x40 
>     shl     v31.4s, v31.4s, #22
>     fcmeq   v1.4s, v0.4s, v0.4s
>     orr     v2.8b, v0.8b, v31.8b
>     bif     v0.8b, v2.8b, v1.8b     
> ```
> 
> @Sonicadvance1 think we could just enable by default given this? given in most cases it's effectively 3 insts, could get to two insts with sve (albeit same per mca since fcmuo has much lower throughput)

Nice - I took a stab at trying that out.

pmatos:
@bylaws that was great. take a look. :)

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -326,6 +327,13 @@ class ContextImpl final : public FEXCore::Context::Context, public CPU::CodeBuff
     }
   }
 
+  void UpdateX87PrecisionConfig() {
```
Sonicadvance1: This function isn't correct. This'll tear the config across what `FEX_CONFIG_OPT` has already been loaded versus not. This should be moved to `FEXCore/Source/Interface/Config/Config.cpp` in `ReloadMetaLayer` towards the end. You can see how single stepping changes the MaxInst config and you could do something similar for this there.

pmatos: Sure. I will take a look. regarding @bylaws comment, should we then keep the strict mode even with the reduced performance impact?

Sonicadvance1: Could we get some perf numbers from the x87 heavy games? I can see Psychonauts multiblock things have doubled in size.

bylaws:
You'll want to pool the casted constant through RA, rather than materializing it every time before benchmarking

FEXCore/Source/Interface/Context/Context.h
pmatos: Are we talking about FPS numbers here or do you want something more specific?

bylaws:
Right now you emit the cast every time, you can pool the casted variant instead to avoid an extra mov. Maybe better to use a separate named function to avoid duplicating the magic constant :)

FEXCore/Source/Interface/Context/Context.h
Sonicadvance1: FPS numbers and/or microbench. I'm curious what the impact will be on like Bayonetta where it happens x87.

PR #5026, opened 2025-11-05
CMake: Don't use system libraries when cross-compiling.

Sbte:
This fixes compilation for me when building the Windows dlls with libfmt present on the system. I'm not sure if this is the right approach, but something similar was already done for xxhash.


neobrain:
The typical way to prevent host packages from being picked up would be to add
```cmake
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```
to your toolchain file. Does that work for your use case?

Sbte:
I'm using `FEX/Data/CMake/toolchain_mingw.cmake`, which already has this.

neobrain:
Could you post the full build log showing the build error you see with this?

Perhaps setting the other `CMAKE_FIND_*_MODE` options helps?
```cmake
set(CMAKE_FIND_ROOT_PATH_MODE_PROGRAM NEVER)
set(CMAKE_FIND_ROOT_PATH_MODE_LIBRARY ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_INCLUDE ONLY)
set(CMAKE_FIND_ROOT_PATH_MODE_PACKAGE ONLY)
```

Sbte:
Hmm, yeah, I tried that as well, but that didn't help. What did seem to help was setting `CMAKE_FIND_ROOT_PATH` to a non-existing path. If `CMAKE_FIND_ROOT_PATH` is empty, it doesn't work. I'll dig a bit more. It may be an issue with the latest CMake version, or maybe just with my system. I did also needed to update the xxhash submodule to be able to build with the latest CMake, so I imagine no one tried that yet. Feel free to close the PR.

neobrain:
Sounds good, if you find any tweaks we can easily add to the bundled toolchain files let us know!

PR #5033, opened 2025-11-09
CPUID: Fixes APICID for processor count calculation. 

Sonicadvance1:
Primary fix here is returning the current CPU index in function 01h.
Intel Quartus uses this alongside affinity setting to check if all cores
can be used for its calculation. Since we had hardcoded apicid 0 here,
it assumed to only have one core and never generated worker threads.

Additional fix for apicid size. This is the size of the bitmask required
for apic ids, we weren't calculating this correctly at all. This mask is
a "maximum" number of APICs that the CPU reserves in power of two.
Say the core supports 256 APICs, but the processor only supports 16, or
any other combination.

This CPU count issue with Quartus has been reported a couple of times over the years, but never had a good idea what to look for until today.

### DIFF
--- a/unittests/FEXLinuxTests/tests/cpu/cpu_count.cpp
+++ b/unittests/FEXLinuxTests/tests/cpu/cpu_count.cpp
@@ -1,0 +1,104 @@
+#include <catch2/catch_test_macros.hpp>
+
+#include <cpuid.h>
+#include <optional>
+#include <unistd.h>
+#include <thread>
+
+TEST_CASE("cpu count - libraries") {
+  const auto hw_concurrency = std::thread::hardware_concurrency();
+  CHECK(hw_concurrency == sysconf(_SC_NPROCESSORS_CONF));
+  CHECK(hw_concurrency == sysconf(_SC_NPROCESSORS_ONLN));
+}
+
+struct core_info {
+  uint32_t local_apicid;
+  uint32_t max_addressible_ids;
+  uint32_t cores;
+  uint32_t threads;
+  bool HTT;
+};
+
+struct cpuid_fn {
+  uint32_t eax, ebx, ecx, edx;
+};
+
+cpuid_fn get_cpuid(uint32_t func, uint32_t leaf = 0) {
+  cpuid_fn fn {};
+  __cpuid_count(func, leaf, fn.eax, fn.ebx, fn.ecx, fn.edx);
+  return fn;
+}
+
+std::optional<core_info> cpuid_calculate_core_info() {
+  core_info info {};
+
+  // Legacy path
+  const auto cpuid_fn_0 = get_cpuid(0);
+  if (cpuid_fn_0.eax < 1) {
+    return std::nullopt;
+  }
+
+  const auto cpuid_fn_1 = get_cpuid(1);
+
+  info.local_apicid = cpuid_fn_1.ebx >> 24;
+  info.HTT = (cpuid_fn_1.edx >> 28) & 1;
+
+  const auto cpuid_fn_8000_0000 = get_cpuid(0x8000'0000U);
+
+  if (cpuid_fn_8000_0000.eax < 0x8000'0008) {
+    return std::nullopt;
+  }
+
+  const auto cpuid_fn_8000_0008 = get_cpuid(0x8000'0008U);
+  const uint32_t apic_id_size = (cpuid_fn_8000_0008.ecx >> 12) & 0xF;
+
+  // E.5.2: MNLP (Maximum Number of Logical Processors)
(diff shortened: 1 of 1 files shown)


FEXCore/include/FEXCore/Utils/MathUtils.h
```diff
@@ -28,6 +28,17 @@ constexpr T ilog2(T Value) {
...
+template<typename T>
+requires (std::is_unsigned_v<T>)
+[[nodiscard]]
+constexpr T next_pow2(T Value) {
+  LOGMAN_THROW_A_FMT(Value != 0, "next_pow2 requires not-zero");
+  constexpr auto num_bits = sizeof(T) * 8;
+  const auto zeroes = std::countl_zero(Value - 1);
+  return 1ULL << (num_bits - zeroes);
+}
+
```
neobrain: C++20 has a function for this: `std::bit_ceil`

That also avoids ambiguity since `next_pow2(4)` sounds like it would return 8 even though it doesn't.

unittests/FEXLinuxTests/tests/cpu/cpu_count.cpp
```diff
@@ -0,0 +1,104 @@
+#include <catch2/catch_test_macros.hpp>
```
neobrain, line 1: Does this test any FEX functionality at all or are you just verifying assumptions made about Linux interfaces? Should have a comment if it's the latter, since it's not an uncommon use of unit tests.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -168,6 +171,7 @@ class Mutex final {
         Desired &= ~READ_WAITER_BIT;
       }
     } while (AtomicFutex.compare_exchange_strong(Expected, Desired, std::memory_order_acq_rel, std::memory_order_acquire) == false);
+    // LogMan::Msg::DFmt("WUL");
```
neobrain: ?

FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -509,7 +510,7 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_01h(uint32_t Leaf) const {
             (1 << 25) | // SSE
             (1 << 26) | // SSE2
             (0 << 27) | // Self Snoop
-            (1 << 28) | // Max APIC IDs reserved field is valid
+            (0 << 28) | // (HTT) Max APIC IDs reserved field is valid
```
neobrain, line 513: "HTT APIC" yields no useful results via websearch, so this should be changed to something more explicit, otherwise you're not adding information here.

FEXCore/Source/Utils/WritePriorityMutex.h
```diff
@@ -232,6 +236,12 @@ class Mutex final {
     return AtomicFutex.compare_exchange_strong(Expected, Desired, std::memory_order_acq_rel, std::memory_order_acquire);
   }
 
+  void StealAndDropActiveLocks() {
```
neobrain: I think you didn't mean to add this file to the patch?

FEXCore/Source/Interface/Core/CPUID.cpp
Sonicadvance1: Websearch results don't cache Intel and AMD system architecture manuals that explicitly point this out, just like the rest of the comments in this file. HTT, aka Hyper-Threading Technology, but the bit also shares the bit with the APIC IDs being valid (All CPUs with HTT need to have valid APIC IDs unless querying through other extended interfaces).
Image_2025-11-11_11-21-31

Image_2025-11-11_11-22-32

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: oops.

FEXCore/Source/Utils/WritePriorityMutex.h
Sonicadvance1: oops.

unittests/FEXLinuxTests/tests/cpu/cpu_count.cpp
Sonicadvance1: It's comparing OS returned values with cpuid returned results, which means it's testing FEX functionality.

FEXCore/include/FEXCore/Utils/MathUtils.h
Sonicadvance1: Removed.

lioncash (approved):

PR #5035, opened 2025-11-11
Use gradual memory growth

pmatos:
Use min instead of max, otherwise we are always using `MAX_STATS_SIZE`.


pmatos:
@Sonicadvance1 While trying to track something else I came across this calculation and although I don't understand all the code flow here, I think this should be `std::min`, otherwise we'll always be changing newsize to be max memory, right?

Sonicadvance1 (approved):
:surprised-pikachu-face:

PR #5038, opened 2025-11-11
unittests/FEXLinuxTests: Add the beginning of ELF mapping tests

Sonicadvance1:
To stress our code caching paths. Won't pass CI until #4980 is fixed.

### DIFF
--- a/unittests/FEXLinuxTests/tests/fd/elf_mapping_tests.cpp
+++ b/unittests/FEXLinuxTests/tests/fd/elf_mapping_tests.cpp
@@ -1,0 +1,69 @@
+// SPDX-License-Identifier: MIT
+#include <catch2/catch_test_macros.hpp>
+
+#include <elf.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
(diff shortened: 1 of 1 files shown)


neobrain:
The issue you mentioned in #4980 is unrelated to this unit test. I had already fixed it in a commit I referenced there before it was reported.

---

In general, I don't think synthetic unit tests are a good guide for design evolution. There's unsurprisingly a ton of memory mapping patterns that could hypothetically break code caching but won't practically matter if they don't occur in the wild. Conversely overoptimizing for such hypothetical scenarios is going to yield an implementation that performs poorly elsewhere.

We should instead observe what games do in the wild, quantify how often this happens in practice, and what impact it has on cache efficacy. We can then adapt the implementation and write unit tests around that as needed (by extending the tests [I already added](https://github.com/FEX-Emu/FEX/blob/main/unittests/APITests/FileMappingBaseAddress.cpp) modelling real-world use cases).

Sonicadvance1:
Okay, I'll keep this test around locally so I can extend it.

PR #5039, opened 2025-11-12
FEX: Moves FEX thunk callback function generation to the frontend 

Sonicadvance1:
Adds it to the VDSO handling, it's not necessarily a VDSO function but
it behaves as such as it is in every single process. This means we get
to reuse the mapped page for every process when thunks are built,
shaving a page out of 32-bit processes.

Also, fixes a bug in guest VDSO symbol loading where clang sticks all
symbols in to `.dynsym` where gcc sticks them in to `.symtab`. Search
both. This effectively meant the couple of guest VDSO symbols were
always failing to get found, causing us to allocate yet another page on
32-bit. So effectively three pages stolen.

This also means we can remove the Linux specific X86HelperGen stuff from
FEXCore, only passing a single "VDSO" function pointer to the backend
for the dispatcher. Once again moving the Linux stuff to the frontend is
good.

Fixes an assert about about untracked noexec code `NoExec
instruction in entry block: FFFFE000` whenever thunk callbacks were
used.


Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -608,136 +608,188 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
 }};
 
 void LoadGuestVDSOSymbols(bool Is64Bit, char* VDSOBase) {
-  // We need to load symbols we care about.
-  if (Is64Bit) {
-    // We don't care about any 64-bit symbols right now.
-    return;
-  }
+  auto LoadSymbols = [&]<bool Is64Bit>(char* VDSOBase) {
+    using ELFHeaderType = typename std::conditional<Is64Bit, Elf64_Ehdr, Elf32_Ehdr>::type;
```
lioncash:
```suggestion
    using ELFHeaderType = std::conditional_t<Is64Bit, Elf64_Ehdr, Elf32_Ehdr>;
```

etc

Sonicadvance1: Fixed.

lioncash (approved):

ThunkLibs/libVDSO/libVDSO_Guest.cpp
```diff
@@ -61,4 +61,12 @@ __attribute__((naked)) void __kernel_rt_sigreturn() {
...
 ssize_t __vdso_getrandom(void*, size_t, uint32_t, void*, size_t) __attribute__((alias("fexfn_pack_getrandom")));
 #endif
+
+__attribute__((naked)) void __fex_callback_ret() {
+  // CALLBACKRET FEX Instruction
+  asm volatile(R"(
+  .byte 0x0f, 0x3e;
+  )" ::
+                 : "memory");
+}
```
neobrain, line 71: Can't we just define this directly in Thunks.cpp since it's part of the runtime and is manually assembled already? We already manage a memory page for executable memory there (MakeHostTrampolineForGuestFunction) as well, so there's no need to add a new one.

I get that it seems convenient to put it in VDSO, but it really doesn't make sense to put an essential part of the library forwarding runtime in a random guest library.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -608,136 +608,188 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
 }};
 
 void LoadGuestVDSOSymbols(bool Is64Bit, char* VDSOBase) {
-  // We need to load symbols we care about.
-  if (Is64Bit) {
-    // We don't care about any 64-bit symbols right now.
-    return;
-  }
+  auto LoadSymbols = [&]<bool Is64Bit>(char* VDSOBase) {
```
neobrain: Why do we need 64-bit support here now? Is this purely because the new function is defined as part of VDSO or for other reason that's not apparent from the commit message?

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -608,136 +608,188 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
...
+          }
         }
       }
     }
+  };
+
+  if (Is64Bit) {
+    LoadSymbols.template operator()<true>(VDSOBase);
+  } else {
+    LoadSymbols.template operator()<false>(VDSOBase);
```
neobrain: If we need to keep 64-bit support: LoadGuestVDSOSymbols is only used internally, so we can just make that function be a template and invoke the proper variant at the call-site. That's better than wrapping its entire logic in a template lambda and then using .template operator()<true> syntax that few people even recognize.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -608,136 +608,188 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
...
+        uint64_t offset = SymTableHeader->sh_offset + i * SymTableHeader->sh_entsize;
+        auto Symbol = reinterpret_cast<const ELFSymbolType*>(VDSOBase + offset);
+        if (ELF32_ST_VISIBILITY(Symbol->st_other) != STV_HIDDEN && Symbol->st_value != 0) {
+          const char* Name = &StrTab[Symbol->st_name];
+          if (Name[0] != '\0') {
+            if (!Is64Bit && strcmp(Name, "__kernel_sigreturn") == 0) {
+              VDSOPointers.VDSO_kernel_sigreturn = VDSOBase + Symbol->st_value;
+            } else if (!Is64Bit && strcmp(Name, "__kernel_rt_sigreturn") == 0) {
+              VDSOPointers.VDSO_kernel_rt_sigreturn = VDSOBase + Symbol->st_value;
+            } else if (strcmp(Name, "__fex_callback_ret") == 0) {
```
neobrain: Can you move the change that takes care of `__fex_callback_ret` and related functionality to a dedicated patch? It's difficult to review how code has moved when everything is munged together.

EDIT: Actually not sure if that's what's munged together or something else. I haven't yet found what the symtab fix is about. Definitely would help to have this split up in logical commits either way.

ThunkLibs/libVDSO/libVDSO_Guest.cpp
Sonicadvance1: It's not random as /every/ application is required to have a VDSO. It's a hack when it /doesn't/ exist. VDSO is very explicitly the kernel support library, we're the kernel in the eyes of the x86 application. It's the right place to put it in to.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Peeled the symbol lookup out of this PR in to https://github.com/FEX-Emu/FEX/pull/5056

Once that is merged I can rebase this.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Because the thunk callback is in VDSO. Currently only 32-bit has two symbols that /can't/ be missing otherwise the world breaks, sigreturn and rt_sigreturn. thunk callret increases that to three symbols for 32-bit and one for 64-bit.

ThunkLibs/libVDSO/libVDSO_Guest.cpp
neobrain: > It's a hack when it /doesn't/ exist

That's not the argument I was making.

The fex-specific helper function isn't part of the kernel interface. VDSO is the last place people (including me) think of when looking for elements of the library forwarding runtime.

More pragmatically, Thunks.cpp seems like a cleaner place to place this function in overall.

FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
```diff
@@ -489,7 +488,7 @@ void Dispatcher::EmitDispatcher() {
 
     // Now push the callback return trampoline to the guest stack
     // Guest will be misaligned because calling a thunk won't correct the guest's stack once we call the callback from the host
-    LoadConstant(ARMEmitter::Size::i64Bit, ARMEmitter::Reg::r0, CTX->X86CodeGen.CallbackReturn);
+    LoadConstant(ARMEmitter::Size::i64Bit, ARMEmitter::Reg::r0, CTX->SignalDelegation->GetThunkCallbackRET());
```
neobrain, line 491: Will this require a relocation with code caching?

bylaws: The dispatcher presumably won't be cached

neobrain: Err, fair point.

ThunkLibs/libVDSO/libVDSO_Guest.cpp
Sonicadvance1: FEX /is/ the kernel interface, I am just merging the semantics between what it means to be a host VDSO versus what FEX provides which is where you're disliking the semantics blending.
If you want me to rename the entire world to FEXVDSO so that it can have additional FEX "VDSO"-style semantics in it, then I can rename it. It's a place that is guaranteed to exist, thunk callbacks need to be guaranteed to exist. It's the place to stick it.

Sonicadvance1:
Rebased now that #5056 is merged

ThunkLibs/libVDSO/libVDSO_Guest.cpp
neobrain: Not sure if I would like FEXVDSO better since my core concern is spreading FEX-specific runtime into surprising corners of the code base. But as mentioned in #5056 I'm not going to block this PR over it since 2 people have already nodded this off now.

Thanks for taking the time to engage in the argument!

PR #5044, opened 2025-11-12
Scripts: Clean-up 

antonkesy:
I had a glance over the scripts and noticed some unused imports and unaligned comments, and quickly fixed them. This should not break anything 👀 but who knows when using Python ...  
_I currently don't have access to my local test runner, so I hope the CI catches any mistakes_

A nice next step might be to add formatter/style guide for Python or something similar.
The existing code is already nearly identical to the output of the [black](https://github.com/psf/black) formatter.

I'll open an issue/PR to open a discussion/show a PoC in the next few days (if I'm not told otherwise).

### DIFF
--- a/Scripts/testharness_runner.py
+++ b/Scripts/testharness_runner.py
@@ -76,4 +76,4 @@ if (is_known_failure):
         sys.exit(1)
 else:
     # Just return the result code if we don't have this test as a known failure
-    sys.exit(ResultCode);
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Oh hey thanks for the PR, seems reasonable enough to me. I'm terrible at Python and bash so I trend towards leaving random garbage in the files.

PR #5045, opened 2025-11-13
JIT: Handle long ADR/ADRP

lioncash:
Wires up the long address handler into the ADR/ADRP restart handlers.

### DIFF
--- a/CodeEmitter/CodeEmitter/ALUOps.inl
+++ b/CodeEmitter/CodeEmitter/ALUOps.inl
@@ -100,16 +100,22 @@ public:
   }
 
   [[nodiscard]] BranchEncodeSucceeded LongAddressGen(ARMEmitter::Register rd, const BackwardLabel* Label) {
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -483,23 +483,31 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
 
   template<ARMEmitter::IsLabel T>
   void adr_OrRestart(ARMEmitter::Register rd, T* Label) {
+    if (RequiresFarARM64Jumps) {
+      if (LongAddressGen(rd, Label) == ARMEmitter::BranchEncodeSucceeded::Failure) {
```
Sonicadvance1, line 487: Oh heck, I actually just looked at the implementation of `LongAddressGen` in the CodeEmitter and it only uses adr or adrp+add.
We actually need a new label type that is guaranteed to emit 3-nops to be able to encode the 48-bit movz+movk+movk in the worst case. (Long Long address gen? It's a good advertising campaign reference).

PR #5047, opened 2025-11-13
Align code with clang-format

antonkesy:
Automatically done by running: `find . \( -path './External' -prune \) -o \( -iname '*.cc' -o -iname '*.cpp' -o -iname '*.hpp' -o -iname '*.h' -o -iname '*.c' \) -print | xargs clang-format --style=file -i`

```bash
> clang-format --version
19.1.1
```

### DIFF
--- a/ThunkLibs/Generator/gen.cpp
+++ b/ThunkLibs/Generator/gen.cpp
@@ -525,8 +525,8 @@ void GenerateThunkLibsAction::OnAnalysisComplete(clang::ASTContext& context) {
     for (auto& import : thunked_api) {
       const auto& function_name = import.function_name;
       const char* variadic_ellipsis = import.is_variadic ? ", ..." : "";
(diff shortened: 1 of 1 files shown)


unittests/FEXLinuxTests/tests/syscalls/syscalls_efault.cpp
antonkesy, line 1: I prefer the manual line breaks more. 

I think this could be archived using clang-format with:
```
AllowAllConstructorInitializersOnNextLine: false
AllowAllArgumentsOnNextLine: false
AllowAllParametersOfDeclarationOnNextLine: false
```

antonkesy: Okay, this issue was on my side, because I accidentally used clang-format-20 instead of 19 ...

neobrain:
Thanks! FWIW we also have a script to trigger reformatting for all files, see `Scripts/reformat.sh` (basically a thin wrapper around `git ls-files | clang-format-19`).

PR #5050, opened 2025-11-14
Config: Refactor value getter interface

neobrain:
This PR adds a template-based `Getter` interface that unifies various macros we used previously. The interface is extended by functions like `Get_DUMPIR`, which allow reading config values in arbitrary C++ expressions (as opposed to `FEX_CONFIG_OPT`, which can only be a standalone statement).

Thin wrappers around the new `Getter` interface are kept to avoid having to mass-migrate the entire code base. If desired, this could be done in a followup.

Previously:
```cpp
  FEX_CONFIG_OPT(IsMultiblock, MULTIBLOCK);
  if (IsMultiblock) {
    FEX_CONFIG_OPT(MaxInst, MAXINST);
    LogMan::Msg::DFmt("Inst: {}", MaxInst);
  }
```
Now:
```cpp
  if (FEXCore::Config::Get_MULTIBLOCK()) {
    LogMan::Msg::DFmt("Inst: {}", FEXCore::Config::Get_MAXINST()());
  }
```

This caught a bug in FEXInterpreter.cpp, where we didn't update the global `SilentLog` variable correctly due to a shadowing `Value` instance.

~~Also cleans up the StringArrayType specialization of `Value` by moving it to its own specialization.~~ Dropped c27ed5aa90f859ec0f4dbecdbd8dc80f16ef9afd because a clang 16 bug interferes with external member function implementations for template classes.

---

Alternative to #4987.

### DIFF
--- a/FEXCore/include/FEXCore/Config/Config.h
+++ b/FEXCore/include/FEXCore/Config/Config.h
@@ -103,11 +103,7 @@ static inline std::optional<fextl::string> EnumParser(const ArrayPairType& EnumP
   return fextl::fmt::format("{}", EnumMask);
 }
 
-namespace DefaultValues {
-  namespace Type {
-    using StringArrayType = fextl::list<fextl::string>;
-  } // namespace Type
-} // namespace DefaultValues
+using StringArrayType = fextl::list<fextl::string>;
 
 namespace detail {
   template<ConfigOption Option>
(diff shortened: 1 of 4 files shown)


FEXCore/include/FEXCore/Config/Config.h
```diff
@@ -293,6 +289,38 @@ class FEX_DEFAULT_VISIBILITY Value {
...
+    : Value<typename OptionInfo::Type> {Option, OptionInfo::Default()} {}
 };
+
+/**
+ * Helper for reading a config value with caching.
+ *
+ * Typically this is used to declare class members so that the value is read
+ * on construction of the parent.
+ */
+#define FEX_CONFIG_OPT(name, enum) FEXCore::Config::Getter<FEXCore::Config::ConfigOption::CONFIG_##enum> name;
```
lioncash: Is leaving the semicolon at the end intentional (just a subtle declarative change I noticed between the previous version of the define)

neobrain: Oh right, that shouldn't be there indeed. Thanks!

neobrain: In fact, I also added a default initializer `{}` to make the macro slightly harder to use incorrectly.

PR #5051, opened 2025-11-14
Config: Add interface to query cache directory

Sonicadvance1:
Also allow override of cache target directory, as it will be used.

### DIFF
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -603,6 +603,11 @@ fextl::string GetConfigDirectory(bool Global, const PortableInformation& Portabl
 }
 
 fextl::string GetCacheDirectory() {
+  const char* CacheOverride = getenv("FEX_APP_CACHE_LOCATION");
+  if (CacheOverride) {
+    return CacheOverride;
+  }
+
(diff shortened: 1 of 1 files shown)


Source/Common/Config.cpp
```diff
@@ -602,6 +602,17 @@ fextl::string GetConfigDirectory(bool Global, const PortableInformation& Portabl
...
 
+fextl::string GetCacheDirectory() {
+  const char* CacheOverride = getenv("FEX_APP_CACHE_LOCATION");
+  if (CacheOverride) {
+    return CacheOverride;
+  }
+
+  const char* HomeDir = GetHomeDirectory();
+  const char* CacheXDG = getenv("XDG_CACHE_HOME");
+  return (CacheXDG ? fextl::string {CacheXDG} : fextl::string {HomeDir} + "/.cache") + "/fex-emu/";
```
bylaws:
```suggestion
  return (CacheXDG ? fextl::string {CacheXDG} : (fextl::string {HomeDir} + "/.cache")) + "/fex-emu/";
```

Sonicadvance1:
Solved by #4906

PR #5052, opened 2025-11-14
FEXCore/Config: Adds a new `FileCodeCache` option

Sonicadvance1:
Not wired up, but before committing any live code, I want to make sure we wrap it all in this check.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -199,6 +199,13 @@
           "Can potentially introduce more stutters, more likely the higher the number.",
           "Don't have this number larger than the increase count!"
         ]
+      },
+      "FileCodeCache": {
+        "Type": "bool",
+        "Default": "false",
+        "Desc": [
+          "Control if file code caches are generated and used."
+        ]
       }
     },
     "Debug": {


bylaws:
Would this gate codemap output also? Of opinion it would be nice to have that as a different setting to using code caches

neobrain:
> Would this gate codemap output also? Of opinion it would be nice to have that as a different setting to using code caches

My plan was to have 3 toggles:
* Enable code cache subsystem (the only setting users should care about; if disabled, codemaps/caches are neither written nor used)
* If subsystem enabled:
  * Read-only code maps (for debugging issues in codemap -> cache generation)
  * Read-only code caches (for debugging issues of already generated caches)

Do we have user scenarios besides debugging where the latter two settings would be used, or scenarios where other settings would be useful?

Sonicadvance1:
> > Would this gate codemap output also? Of opinion it would be nice to have that as a different setting to using code caches
> 
> My plan was to have 3 toggles:
> 
> * Enable code cache subsystem (the only setting users should care about; if disabled, codemaps/caches are neither written nor used)
> * If subsystem enabled:
>   
>   * Read-only code maps (for debugging issues in codemap -> cache generation)
>   * Read-only code caches (for debugging issues of already generated caches)
> 
> Do we have user scenarios besides debugging where the latter two settings would be used, or scenarios where other settings would be useful?

Expectation is that currently we care about the all-or-nothing setting, although I could see the appeal of making an additional option for read-only.

Sonicadvance1:
Superseded by #4906 now

PR #5056, opened 2025-11-15
FEX/VDSO: Fixes symbol lookup

Sonicadvance1:
Fixes a bug in guest VDSO symbol loading where clang sticks all symbols in to .dynsym where gcc sticks them in to .symtab. Search both. This effectively meant the couple of guest VDSO symbols were always failing to get found, causing us to allocate yet another page on 32-bit. So effectively three pages stolen.

Peeled out of #5039

### DIFF
--- a/Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
+++ b/Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
@@ -607,7 +607,12 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
   },
 }};
 
-void LoadGuestVDSOSymbols(bool Is64Bit, char* VDSOBase) {
+template<bool Is64Bit>
+void LoadGuestVDSOSymbols(char* VDSOBase) {
+  using ELFHeaderType = std::conditional_t<Is64Bit, Elf64_Ehdr, Elf32_Ehdr>;
+  using ELFSHeaderType = std::conditional_t<Is64Bit, Elf64_Shdr, Elf32_Shdr>;
+  using ELFSymbolType = std::conditional_t<Is64Bit, Elf64_Sym, Elf32_Sym>;
+
   // We need to load symbols we care about.
   if (Is64Bit) {
     // We don't care about any 64-bit symbols right now.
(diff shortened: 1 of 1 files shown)


Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -607,52 +607,66 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
   },
 }};
 
-void LoadGuestVDSOSymbols(bool Is64Bit, char* VDSOBase) {
+template<bool Is64Bit>
```
neobrain, line 610: This isn't needed for the bugfix AFAICS, and it doesn't look like it's strongly interwoven with the core of the patch.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -607,52 +607,66 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
...
+  auto RawShdrs = reinterpret_cast<const ELFSHeaderType*>(VDSOBase + Header->e_shoff);
 
-  const Elf32_Shdr* StrHeader = &RawShdrs[Header->e_shstrndx];
+  const auto StrHeader = &RawShdrs[Header->e_shstrndx];
   const char* SHStrings = VDSOBase + StrHeader->sh_offset;
 
-  const Elf32_Shdr* SymTableHeader {};
-  const Elf32_Shdr* StringTableHeader {};
+  const ELFSHeaderType* SymTableHeader {};
+  const ELFSHeaderType* StringTableHeader {};
```
neobrain: These should be moved into the `sym_table` loop since their values are only valid during the respective loop iteration.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
```diff
@@ -607,52 +607,66 @@ static std::array<FEXCore::IR::ThunkDefinition, 7> VDSODefinitions = {{
...
-      StringTableHeader = &RawShdrs[SymTableHeader->sh_link];
-      break;
+  struct SymbolTypes {
+    const char* name;
+    int sh_type;
+  };
+
+  constexpr std::array<SymbolTypes, 2> symbol_table_names = {{{".dynsym", SHT_DYNSYM}, {".symtab", SHT_SYMTAB}}};
+
+  for (auto sym_table : symbol_table_names) {
```
neobrain, line 638: I would've gone for `std::pair` over a named struct and then using `auto [name, sh_type]` here. Can leave it as is though, just mentioning in case it's interesting.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Why? It doesn't matter and the other PR is going to re-add it anyway. It only serves as a pointless dance of removing and adding code in these PRs to waste time when it can just be added and be done with it.

neobrain: Up to you - if I understood correctly it's unnecessary with the counterproposal in the other PR, so not adding it here allows us to merge the patch without settling that discussion first.

neobrain: Democracy has spoken on the other PR. Let's only address the [nit](https://github.com/FEX-Emu/FEX/pull/5056#discussion_r2537257632) with the variable scope and get this merged so the fix is documented as its own commit.

Source/Tools/LinuxEmulation/VDSO_Emulation.cpp
Sonicadvance1: Moved.

PR #5058, opened 2025-11-16
Add SteamOS to supported Linux distributions list

ShaunRoselt:


Sonicadvance1:
Good commit, but we can't support something that doesn't yet exist. Let's leave that out until something changes.

PR #5059, opened 2025-11-16
Async: Move file descriptor checks from read_some() to read() 

Sonicadvance1:
This allows using read_some for incoming messages with an optional FD.
Doing so fits the purpose of read_some more closely, which is to read *any*
non-empty amount of data.

From the working branch.

### DIFF
--- a/Source/Common/Async.h
+++ b/Source/Common/Async.h
@@ -18,6 +18,7 @@
 #include <FEXCore/fextl/functional.h>
 #include <FEXCore/fextl/map.h>
 #include <FEXCore/fextl/vector.h>
+#include <FEXCore/Utils/LogManager.h>
 
 namespace fasio {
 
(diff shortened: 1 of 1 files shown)


neobrain:
I'm sure you had a lot of fun coming up with the branch name but FTR, I would prefer my patches to show up under a more reasonable branch name than warkwarkwarkwarkwarkwarkwarkwarkwarkwarkwarkwarkwarkwark, if nothing else because it's a pain to navigate git logs when each merge commit is unreadable.

neobrain:
Just realized this actually introduces an assertion failure to FEX because this patch relied on another (unfinished) patch that wasn't included...

Could you revert this for the time being? We should focus on integrating #4906 (which I'm working on), after that these patches are easy to integrate. Rushing these patches out just risks introducing regressions that can easily be avoided.

Sonicadvance1:
> Just realized this actually introduces an assertion failure to FEX because this patch relied on another (unfinished) patch that wasn't included...
> 
> Could you revert this for the time being? We should focus on integrating #4906 (which I'm working on), after that these patches are easy to integrate. Rushing these patches out just risks introducing regressions that can easily be avoided.

I'll fix the assertion failure.

PR #5060, opened 2025-11-16
FEXCore/Config: Expose GetConv members

Sonicadvance1:
From working branch commit 8244ca1666796267ce25741cdf1103eef4f7539d `Make cache generation aware of FEX configuration`


neobrain:
I would prefer these commits to stay in the context that justified adding them. On its own, it's a bit pointless to add.

Sonicadvance1:
I'm working through the minutiae.

PR #5065, opened 2025-11-18
FEXCore/CodeCache: Move spin-loop over to a WFE loop

Sonicadvance1:
Implements support for WFE spin-loops with a comparison predicate. Moving the original `Wait` over to using the `EQ` predicate. Adding an `LE` version specifically for the CodeCache spin-loop.

We can't use generic predicates here because of stack frames getting inbetween load exclusive and wfe which break it. I'm not anticipating many, if any, more predicate types to be added.

### DIFF
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -1,4 +1,6 @@
 // SPDX-License-Identifier: MIT
+#include "Utils/SpinWaitLock.h"
+
 #include <Interface/Context/Context.h>
 
 #include <FEXCore/HLE/SourcecodeResolver.h>
@@ -186,9 +188,7 @@ void CodeMapWriter::AppendData(std::span<const std::byte> Data) {
     }
     if (!IsResponsibleForFlush) {
       // Wait for the buffer to be flushed on the responsible thread
-      while (BufferOffset > Buffer.size()) {
-        std::this_thread::yield();
-      }
+      Utils::SpinWaitLock::WaitPred<std::less_equal<>, size_t>(reinterpret_cast<size_t*>(&BufferOffset), Buffer.size());
     }
     AppendData(Data);
     return;


Sonicadvance1:
Seriously mingw?!

Sonicadvance1:
Oh wow, okay actually. I can't use predicate functions here because it accesses memory and we can't have that. We'll need to use "comparison operations".

I was thinking the predicate would be fully inlined as a lambda and I was so terribly mistaken.

Sonicadvance1:
Okay, this was actually trickier than expected since we're dancing around assembly code. Basically required to do template instantiation with function declarations passed through.

Sonicadvance1:
Now that the codemaps PR is merged I have rebased it and it's a fairly small change overall, just a little tricky because of the frame stuff.

neobrain:
> Oh wow, okay actually. I can't use predicate functions here because it accesses memory and we can't have that. We'll need to use "comparison operations".

Yeah, that's a fun issue. We can restore some flexibility by moving the predicate back to the function parameters - but as a template parameter! The predicate must still be default-constructible now, but it's much cleaner and should produce the same assembly then. Seems to work fine in this toy example at least: https://godbolt.org/z/c7hT6fKqb

Sonicadvance1:
> > Oh wow, okay actually. I can't use predicate functions here because it accesses memory and we can't have that. We'll need to use "comparison operations".
> 
> Yeah, that's a fun issue. We can restore some flexibility by moving the predicate back to the function parameters - but as a template parameter! The predicate must still be default-constructible now, but it's much cleaner and should produce the same assembly then. Seems to work fine in this toy example at least: https://godbolt.org/z/c7hT6fKqb

You're right, that does work. Is does give the user the ability to shoot themselves in the foot if the predicate is complex. But that can be known that if someone decides to not use `std::equal_to` or one of the others in the functional helpers, that it needs to be sanity checked.

Moved the implementation over.

neobrain:
> Is does give the user the ability to shoot themselves in the foot if the predicate is complex

What's the problem with complex predicates?

Sonicadvance1:
> > Is does give the user the ability to shoot themselves in the foot if the predicate is complex
> 
> 
> 
> What's the problem with complex predicates?
> 
> 

If a complex predicate accesses memory at all then the exclusive monitor will clear before the WFE instruction executes.

neobrain:
> If a complex predicate accesses memory at all then the exclusive monitor will clear before the WFE instruction executes.

I see. Seems fine for now since we only use the predicate-function in this one place.

Issue #5069, opened 2025-11-19
Writes to argv[0] are not reflected in `/proc/self/cmdline`, breaking app name detection


wcampbell-nv:
**Motivation**

When launched, Wine rewrites argv to begin with the target .exe name. This change is normally reflected in `/proc/self/cmdline`. The NVIDIA driver relies on this renaming to enable some app-specific workarounds based on the value of `/proc/self/cmdline`'s first argument.

This behavior does not work when Wine is launched inside of FEX. `/proc/self/cmdline` keeps its initial value and the app specific workarounds fail to apply.

**Repro**

I've attached a simple repro script: [rename_process.c](https://github.com/user-attachments/files/23635610/rename_process.c)

```
➜ gcc rename_process.c
➜ ./a.out
/proc/self/cmdline: ./a.out
Rewrote argv[0] to RENAMED
/proc/self/cmdline: RENAMED
➜ x86_64-linux-gnu-gcc ./rename_process.c
➜ FEXInterpreter ./a.out
/proc/self/cmdline: ./a.out
Rewrote argv[0] to RENAMED
/proc/self/cmdline: ./a.out
```

**Root cause**

This is because FEX sets up a new stack for the emulated program, placing its argv data at a new memory location that the OS is not aware of.

wcampbell-nv:
Related, it would be nice if FEXInterpreter renamed the process name and cmdline to the target executable's, matching Wine's behavior. This would make name matching of Linux executables work the same whether or not they're emulated.

Sonicadvance1:
It's a known problem. I don't believe there is any way for us to rewrite the cmdline that other processes see which is why we capture opens of `/proc/self/cmdline`. Even binfmt_misc doesn't save us here as it gets injected regardless.

wcampbell-nv:
> we capture opens of `/proc/self/cmdline`

Interesting, this solution could work for our driver workarounds, which are in the same process. But it doesn't work in my example above where the emulated app modifies argv[0]. Looks like the cmdline_handler is outputting stashed string values rather than the argv array on the emulated app's stack.

>  I don't believe there is any way for us to rewrite the cmdline that other processes see

It can be written the same way Wine does it, by writing to argv[0]. Copying the argv memory from the emulated app's stack to the original OS provided argv would do this. The hard part is knowing when to repeat this copy, if the app modifies the data.

Amanieu:
[...]

PR #5070, opened 2025-11-22
Reflect application changes to argv[0] in /proc/self/cmdline

wcampbell-nv:
This is a simple fix for issue https://github.com/FEX-Emu/FEX/issues/5069, adjusting `/proc/self/cmdline` emulation for ELFCodeLoader to output the contents of the stack's argv data rather than the stashed argument strings. This means application writes to argv[0] will be visible to in-process `/proc/self/cmdline` reads. The change is still not visible to external processes.

I've tested that the rename_process.c script attached to the issue now works the same in and out of FEX:

```
➜ gcc rename_process.c
➜ ./a.out
/proc/self/cmdline: ./a.out
Rewrote argv[0] to RENAMED
/proc/self/cmdline: RENAMED

➜ x86_64-linux-gnu-gcc ./rename_process.c
➜ FEXInterpreter ./a.out
/proc/self/cmdline: ./a.out
Rewrote argv[0] to RENAMED
/proc/self/cmdline: RENAMED
```

### DIFF
--- a/Source/Tools/CommonTools/CodeLoader.h
+++ b/Source/Tools/CommonTools/CodeLoader.h
@@ -59,11 +59,9 @@ public:
    */
   virtual void WriteCmdlineFD(int32_t fd) const {
     const auto& Args = GetApplicationArguments();
-    const char NullChar {};
     // cmdline is an array of null terminated arguments
     for (const auto& Arg : Args) {
-      write(fd, Arg.c_str(), Arg.size());
-      write(fd, &NullChar, sizeof(uint8_t));
+      write(fd, Arg.c_str(), Arg.size() + 1); // Add 1 to include the null terminator
     }
   }
 
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -817,6 +817,11 @@ public:
   }
 
   void WriteCmdlineFD(int32_t fd) const override {
+    // '/proc/self/cmdline' typically maps to the stack's argv data.
+    // Applications can write to this to change the cmdline value,
+    // but in FEX they'll write to the emulated stack rather than the OS'.
+    //
+    // Expose the contents of the emulated stack's argument data.
     write(fd, reinterpret_cast<const void*>(StackPointer + ArgumentOffset), ArgumentBackingSize);
   }
 


Sonicadvance1:
Oh wow, I completely missed that would have been a possible work around. Good job!

wcampbell-nv:
I wouldn't have known this was possible without your comment :)

I see one test is failing, but it looks unrelated to my change. Could this be a test issue?

Sonicadvance1:
That's definitely a spooky spurious failure.

Source/Tools/CommonTools/CodeLoader.h
```diff
@@ -53,6 +54,19 @@ class CodeLoader {
...
+   * Writes out the arguments in the format of /proc/self/cmdline
+   */
+  virtual void WriteCmdlineFD(int32_t fd) const {
+    const auto& Args = GetApplicationArguments();
+    const char NullChar {};
+    // cmdline is an array of null terminated arguments
+    for (const auto& Arg : Args) {
+      write(fd, Arg.c_str(), Arg.size());
+      write(fd, &NullChar, sizeof(uint8_t));
+    }
```
neobrain, line 65: fextl::string is null-terminated:

```suggestion
    const auto& Args = GetApplicationArguments();
    // cmdline is an array of null terminated arguments
    for (const auto& Arg : Args) {
      write(fd, Arg.c_str(), Arg.size() + 1);
    }
```

Btw is this implementation used anywhere in practice or do we only ever call the `override` from ELFCodeLoader.h?

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -816,6 +816,10 @@ class ELFCodeLoader final : public FEX::CodeLoader {
     };
   }
 
+  void WriteCmdlineFD(int32_t fd) const override {
+    write(fd, reinterpret_cast<const void*>(StackPointer + ArgumentOffset), ArgumentBackingSize);
```
neobrain, line 825: This line probably merits an explanatory comment.

neobrain:
> The change is still not visible to external processes

Have you tried using the `prctl(PR_SET_MM, PR_SET_MM_MAP, ...)` approach that was suggested in the issue? Seems like that would allow us to drop the `/proc/self/cmdline` hook entirely even.

Source/Tools/CommonTools/CodeLoader.h
wcampbell-nv: There is another `HarnessCodeLoader` that could end up using this

wcampbell-nv:
I've updated the review to address @neobrain's feedback. In testing further I found that this cmdline interception isn't used for native thunk'd code, unfortunately it doesn't resolve the use case I had in mind. It only works within emulated x86 code. This is still an improvement.

I did experiment with `prctl(PR_SET_MM, PR_SET_MM_MAP, ...)`, I'll discuss that on https://github.com/FEX-Emu/FEX/issues/5069.

neobrain:
> I've updated the review to address @neobrain's feedback. In testing further I found that this cmdline interception isn't used for native thunk'd code, unfortunately it doesn't resolve the use case I had in mind. It only works within emulated x86 code. This is still an improvement.
> 
> I did experiment with prctl(PR_SET_MM, PR_SET_MM_MAP, ...), I'll discuss that on https://github.com/FEX-Emu/FEX/issues/5069.

Thanks for taking a look! One thing in particular I observed is that Plasma's task bar will show "FEX" as the primary title for many applications due to how Plasma's heuristics are chosen, so fixing external visibility might resolve that.

PR #5071, opened 2025-11-22
Github: Add a steamrt4 builder

Sonicadvance1:
This ensures we don't break downstream projects.

### DIFF
--- a/Source/Common/HostFeatures.cpp
+++ b/Source/Common/HostFeatures.cpp
@@ -625,6 +625,10 @@ FEXCore::HostFeatures FetchHostFeatures(FEX::CPUFeatures& Features, bool Support
   HostFeatures.SupportsSHA = true;
   HostFeatures.SupportsPMULL_128Bit = true;
   HostFeatures.SupportsAES256 = true;
+
+  // Simulator doesn't support these
+  HostFeatures.SupportsRPRES = false;
+  HostFeatures.SupportsAFP = false;
 #else
   // Check if we can support cacheline clears
   if (Features.GetDCZID().SupportsDCZVA()) {


Sonicadvance1:
This is now ready to go in. Since this is only a build-test the overhead per run is fairly low, but I'll be adding another runner to the build-farm to help mitigate some of the overhead anyway. I've been sitting on this Mac Mini M4 for quite a while and haven't done anything with it.

smcv:
This is Steam Runtime 4 (`steamrt4`), not "sniper 4". sniper is specifically the codename of Steam Runtime 3.

Sonicadvance1:
> This is Steam Runtime 4 (`steamrt4`), not "sniper 4". sniper is specifically the codename of Steam Runtime 3.

Thanks! Moved the name over.

Sonicadvance1:
Found a bug in the vixl simulator features that was only exposed on newer hardware. AFP and RPRES aren't supported by the simulator but the new hardware in CI supports AFP (rpres as well but not in the VM) which was causing failures.

.github/workflows/steamrt4.yml
```diff
@@ -0,0 +1,64 @@
...
+    - name : distrobox setup
+      run: |
+        distrobox create -Y -i registry.gitlab.steamos.cloud/steamrt/steamrt4/sdk/arm64:4.0.20251117.183306 steamrt4 || true
+        distrobox upgrade steamrt4
+        distrobox enter --name steamrt4 -- sudo apt-get install -y \
+          git cmake ninja-build ccache \
+          lld clang \
+          libclang-dev llvm-dev \
+          libstdc++-14-dev-i386-cross libgcc-14-dev-i386-cross \
+          libstdc++-14-dev-amd64-cross libgcc-14-dev-amd64-cross
```
neobrain, line 51: Can our runners not use the regular container workflow syntax? https://docs.github.com/en/actions/how-tos/write-workflows/choose-where-workflows-run/run-jobs-in-a-container

Seems wasteful to re-create the container on each run, whereas that way parts of it would be cached across runs.

Sonicadvance1: It's not actually recreated, it just ensures that if it doesn't exist, that it does get created.

neobrain (approved):

PR #5075, opened 2025-11-24
Config: Document the new `FEX_APP_CACHE_LOCATION` option

Sonicadvance1:
I forgot to document this in the man page.

### DIFF
--- a/FEXCore/Scripts/config_generator.py
+++ b/FEXCore/Scripts/config_generator.py
@@ -200,6 +200,15 @@ def print_man_environment_tail():
     ],
     "''", True)
 
+    print_man_env_option(
+    "APP_CACHE_LOCATION",
+    [
+    "Allows the user to override where FEX stores and loads cache files",
+    "By default FEX will look in $XDG_CACHE_HOME/fex-emu/ or $HOME/.cache/fex-emu/",
+    "This will override the full path, trailing forward-slash is expected to exist",
+    ],
+    "''", True)
+
 def print_man_header():
     header ='''.Dd {0}
 .Dt FEX


FEXCore/Scripts/config_generator.py
```diff
@@ -200,6 +200,15 @@ def print_man_environment_tail():
     ],
     "''", True)
 
+    print_man_env_option(
+    "APP_CACHE_LOCATION",
+    [
+    "Allows the user to override where FEX stores and loads cache files",
+    "By default FEX will look in $XDG_CACHE_HOME/fex-emu or $HOME/.cache/fex-emu/",
+    "This will override the full path",
```
neobrain: The path is expected to include a trailing directory separator, which should either be documented or GetCacheDirectory should append one automatically.

Sonicadvance1: Fixed the comment so both paths have the trailing forward-slash, and updated the comment to say it expects it.

FEXCore/Scripts/config_generator.py
```diff
@@ -200,6 +200,15 @@ def print_man_environment_tail():
     ],
     "''", True)
 
+    print_man_env_option(
+    "APP_CACHE_LOCATION",
+    [
+    "Allows the user to override where FEX stores and loads cache files",
+    "By default FEX will look in $XDG_CACHE_HOME/fex-emu/ or $HOME/.cache/fex-emu/",
+    "This will override the full path, trailing froward-slash is expected to exist",
```
neobrain:
```suggestion
    "This will override the full path, trailing forward-slash is expected to exist",
```

Sonicadvance1: Oop, fixed.

neobrain (approved):

PR #5077, opened 2025-11-24
Scripts/InstallFEX: Fixes two issues

Sonicadvance1:
- Fixes the issue of software-properties-common not being installed and trying to use it
- Fixes the issue of stderr from `apt` output messing with FEXRootFSFetcher

### DIFF
--- a/Scripts/InstallFEX.py
+++ b/Scripts/InstallFEX.py
@@ -202,8 +202,24 @@ def UpdatePPA():
 
     return DidUpdate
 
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
There's another issue that it would be good to skip the binfmt packages if the script detects that systemd-binfmt isn't running, but I don't quite know how to detect that right now. So punting until a different PR.

lioncash (approved):

PR #5091, opened 2025-12-01
JIT: Prepare FEX relocations for code caching

neobrain:
## Overview

When generating code caches, FEX-internal relocations are written in addition to the binary code itself. They are used to generate byte-by-byte consistent cache files across generations/machines, and conversely to restore ASLR-dependent guest base addresses at runtime. We have already been emitting relocations in many places, but this PR completes the infrastructure to make them ready for code caching.

Note: These FEX-internal relocations are unrelated to the relocations in ELF/PE files, though we may want to generate the former from the latter in the future.

## Changes
* Missing relocations are added
* A new relocation type is added to describe literals relative to the binary base address
* Relocation descriptors are now properly zero-initialized to ensure code cache consistency
* Relocation offsets are relative to the code buffer start instead of the block start (because code caching operates on full binaries rather than blocks)

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -1002,9 +1002,17 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
   // CodeSize not including the header or tail data.
   const uint64_t CodeOnlySize = GetCursorAddress<uint8_t*>() - CodeBegin;
 
-  // Add the JitCodeTail
+  // Add the JitCodeTail (written later)
   Align(alignof(JITCodeTail));
   const auto JITBlockTailLocation = GetCursorAddress<uint8_t*>();
+  CodeHeader->OffsetToBlockTail = JITBlockTailLocation - CodeData.BlockBegin;
+
+  JITCodeTail JITBlockTail {
+    .RIP = Entry,
+    .GuestSize = Size,
+    .SpinLockFutex = 0,
+    .SingleInst = SingleInst,
+  };
 
   // Entries that live after the JITCodeTail.
   // These entries correlate JIT code regions with guest RIP regions.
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Core/JIT/Arm64Relocations.cpp
```diff
@@ -37,45 +34,65 @@ void Arm64JITCore::InsertNamedThunkRelocation(ARMEmitter::Register Reg, const IR
   Relocations.emplace_back(MoveABI);
 }
 
-Arm64JITCore::NamedSymbolLiteralPair Arm64JITCore::InsertNamedSymbolLiteral(FEXCore::CPU::RelocNamedSymbolLiteral::NamedSymbol Op) {
+auto Arm64JITCore::InsertNamedSymbolLiteral(FEXCore::CPU::RelocNamedSymbolLiteral::NamedSymbol Op) -> NamedSymbolLiteralPair {
```
bylaws: out of interest, is this a codestyle change or is there a functional reason?

neobrain: No functional reason. Trailing return type allows dropping the Arm64JITCore prefix, but it's not too bad here so I reverted back to the style we use more commonly.

FEXCore/Source/Interface/Core/JIT/Arm64Relocations.cpp
```diff
@@ -40,42 +37,62 @@ void Arm64JITCore::InsertNamedThunkRelocation(ARMEmitter::Register Reg, const IR
...
   dc64(Lit.Lit);
   Relocations.emplace_back(Lit.MoveABI);
 }
 
+auto Arm64JITCore::InsertGuestRIPLiteral(uint64_t GuestRIP) -> NamedSymbolLiteralPair {
+  return {
+    .Lit = GuestRIP,
+    .MoveABI =
+      {
+        .GuestRIP = {.Header =
```
Sonicadvance1, line 79: Jeez, clang-format really makes a dog's breakfast of this declaration.

PR #5092, opened 2025-12-01
SyscallsSMCTracking: Workaround assert in ELF mapping

Sonicadvance1:
The ELF tracking thing has an expectation that only portions of ELF files that are described in the program headers will be mapped executable. This doesn't hold true as programs will remap random portions of ELF files as executable. In the case that this occurs, don't assert out and instead print a warning.

This was discovered as Node.js remaps a portion of itself executable that isn't described as such in the program headers. I also have a local unittest that exposes the same problem. I had discovered same problem in some other program with #5038.

Also fixes a bug where sometimes completely anonymously mapped executable sneak in and cause a crash, which is kind of silly.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -479,22 +479,30 @@ std::optional<SyscallHandler::LateApplyExtendedVolatileMetadata> SyscallHandler:
             (ProtMapping.Executable ? PF_X : 0) | (ProtMapping.Writable ? PF_W : 0) | (ProtMapping.Readable ? PF_R : 0));
           return std::ranges::find(ExpectedBases, Resource.FirstVMA->Base) != ExpectedBases.end();
         });
-        LOGMAN_THROW_A_FMT(ResourceIt != ResourceEnd, "ERROR: Could not find base for file mapping at {:#x} (offset {:#x})", addr, offset);
-        Resource = &ResourceIt->second;
+        if (ResourceIt == ResourceEnd) {
+          // This isn't necessarily a fatal exception. It just means the ELF section isn't a part of the ELF Program headers.
+          // Node.js hits this as it maps a section of itself that isn't a part of the program headers.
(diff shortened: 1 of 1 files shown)


neobrain:
I'll need to do some thinking on whether logging is good enough here. Will get back to it later this week.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -479,22 +479,30 @@ std::optional<SyscallHandler::LateApplyExtendedVolatileMetadata> SyscallHandler:
...
+          // Node.js hits this as it maps a section of itself that isn't a part of the program headers.
+          LogMan::Msg::IFmt("Warning: Could not find base for file mapping at {:#x} (offset {:#x}): {}", addr, offset,
+                            std::string_view(Tmp, PathLength));
+        } else {
+          Resource = &ResourceIt->second;
+        }
       }
 
-      const fextl::string Filename = FHU::Filesystem::GetFilename(Resource->MappedFile->Filename);
+      if (!Resource->MappedFile->Filename.empty()) {
```
neobrain: This is potentially dereferencing a nullptr now. You want to check for nullptr instead:
```suggestion
      if (Resource->MappedFile) {
```

There can never be a MappedFile with an empty filename (though even if there was, the code below would've handled it fine).

neobrain:
> Also fixes a bug where sometimes completely anonymously mapped executable sneak in and cause a crash, which is kind of silly.

All changed code is guarded by `if (!(flags & MAP_ANONYMOUS))`, so how would this affect such cases?

Sonicadvance1:
> > Also fixes a bug where sometimes completely anonymously mapped executable sneak in and cause a crash, which is kind of silly.
> 
> All changed code is guarded by `if (!(flags & MAP_ANONYMOUS))`, so how would this affect such cases?

Anonymous FD, not MAP_ANONYMOUS.

neobrain:
Ah right, that would explain it :)

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
Sonicadvance1: Switched over to this. Seems fine.

PR #5093, opened 2025-12-02
SteamRT4: Adds support for building the steam depot

Sonicadvance1:
Simple enough, adds the tooling necessary to work in conjunction with pressure-vessel. The tool is a fairly lightweight wrapper that just sets up some default configurations.

### DIFF
--- a/Source/Steam/CompatTool.cpp
+++ b/Source/Steam/CompatTool.cpp
@@ -83,6 +83,7 @@ fextl::string GenerateSteamAppConfig(const FEX::Config::PortableInformation& Por
     bool Multiblock = true;
     bool Thunks_GL = false;
     bool Thunks_Vulkan = false;
+    bool EnableLogging = false;
   };
   SteamOptions Options {};
 
@@ -95,6 +96,11 @@ fextl::string GenerateSteamAppConfig(const FEX::Config::PortableInformation& Por
   const auto steam_fex_multiblock = getenv("STEAM_FEX_MULTIBLOCK");
   if (steam_fex_multiblock) {
     Options.Multiblock = std::strtoull(steam_fex_multiblock, nullptr, 0) != 0;
+  }
+
+  const auto steam_fex_logging = getenv("STEAM_FEX_LOG");
+  if (steam_fex_logging) {
+    Options.EnableLogging = std::strtoull(steam_fex_logging, nullptr, 0) != 0;
   }
 
   // UI overrides.
@@ -123,6 +129,10 @@ fextl::string GenerateSteamAppConfig(const FEX::Config::PortableInformation& Por
     Dest = json_objOpen(Dest, "Config");
     Dest = json_str(Dest, "TSOEnabled", Options.TSO ? "1" : "0");
     Dest = json_str(Dest, "Multiblock", Options.Multiblock ? "1" : "0");
+    Dest = json_str(Dest, "SilentLog", Options.EnableLogging ? "0" : "1");
+    if (Options.EnableLogging) {
+      Dest = json_str(Dest, "OutputLog", "server");
+    }
     Dest = json_objClose(Dest);
   }
 
--- a/Source/Tools/FEXServer/Main.cpp
+++ b/Source/Tools/FEXServer/Main.cpp
@@ -206,7 +206,13 @@ int main(int argc, char** argv, char** const envp) {
     LogMan::Msg::DFmt("[FEXServer] Couldn't set subreaper.");
   }
 
-  if (Options.Foreground) {
+  bool EnableLoggingThread = Options.Foreground;
+#ifndef FEX_STEAM_SUPPORT
+  // If running with Steam support then always enable the logging thread.
+  EnableLoggingThread = true;
+#endif
+
+  if (EnableLoggingThread) {
     // Only start a log thread if we are in the foreground.
     // Prevents FEX from trying to log to nothing.
     Logger::StartLogThread();


bylaws:
We might want to commonise some config handling in the future so we can use on the proton side without duplicating, obviously lots of layers involved but will think through. Fine for now though

unittests/APITests/StringUtils.cpp
```diff
@@ -80,3 +80,19 @@ TEST_CASE("trim") {
   CHECK(Trim("\vFEXInterpreter\v") == "FEXInterpreter");
   CHECK(Trim(" FEXInterpreter ") == "FEXInterpreter");
 }
+
+TEST_CASE("InPlaceReplace") {
+  fextl::string Str = "";
+  CHECK(ReplaceAllInPlace(Str, "@", "#") == "");
+  // Replace with shorter.
+  Str = "@Test@";
+  CHECK(ReplaceAllInPlace(Str, "@Test@", "Yes") == "Yes");
```
neobrain: This will prevent the tested value from being printed on test failure. Using a wrapper lambda instead is better here and cleans the code up as well:
```cpp
auto ReplaceAll = [](fextl::string Str) { ReplaceAllInPlace(...); return Str; };
CHECK(ReplaceAll("@Test@", ...) == "Yes");
```

unittests/APITests/StringUtils.cpp
```diff
@@ -80,3 +80,19 @@ TEST_CASE("trim") {
   CHECK(Trim("\vFEXInterpreter\v") == "FEXInterpreter");
   CHECK(Trim(" FEXInterpreter ") == "FEXInterpreter");
 }
+
+TEST_CASE("InPlaceReplace") {
```
neobrain, line 84: Could you add a test for:
* replacing `@` with `@@` in the string `@@@`?
* replacing `@` with "" (empty string) in the string `@@@`?

CMakeLists.txt
```diff
@@ -33,6 +33,7 @@ option(ENABLE_FEXCORE_PROFILER "Enables use of the FEXCore timeline profiling ca
 set (FEXCORE_PROFILER_BACKEND "gpuvis" CACHE STRING "Set which backend to use for the FEXCore profiler (gpuvis, tracy)")
 option(ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT "Enables glibc memory allocation hooking with fault for CI testing")
 option(USE_PDB_DEBUGINFO "Builds debug info in PDB format" FALSE)
+option(BUILD_STEAM_SUPPORT "Builds support for steampipe depot creation" FALSE)
```
neobrain: We're changing behavior quite heavily beyond just adding one additional feature:

```suggestion
option(BUILD_STEAM_SUPPORT "Builds FEX for integration into Steam" FALSE)
```

Source/Steam/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+  PRIVATE
+  FEXCore Common CommonTools JemallocLibs)
+
+install(TARGETS fex-compat-tool
+  RUNTIME
+    DESTINATION /
+    COMPONENT Runtime
+)
+
+# Description json gets installed in to root of depot
```
neobrain:
```suggestion
# Description json gets installed into root of depot
```

Source/Steam/CompatTool.cpp
```diff
@@ -0,0 +1,165 @@
...
+
+  if (!File.IsValid()) {
+    return {};
+  }
+
+  File.Write(Buffer, strlen(Buffer));
+  return ConfigLocation;
+}
+
+int main(int argc, const char** argv, char** const envp) {
```
neobrain:
```suggestion
int main(int argc, const char** argv) {
```

Source/Steam/emulator.json
```diff
@@ -0,0 +1,12 @@
+{
+  "emulator_v0": {
+    "argv": "./usr/bin/FEX",
+    "environment": { "FEX_PORTABLE": "1" },
+    "container_argv": "./usr/bin/FEX",
```
smcv, line 5: I thought you'd come to the conclusion that we need to use a wrapper executable for the `container_argv` and not just the `main_argv`, to avoid FEX trying to write temporary files into its depot directory, which is mounted read-only inside the container? Or did you avoid that some other way?

Source/Steam/emulator.json
```diff
@@ -0,0 +1,12 @@
+{
```
smcv, line 1: A trick I've used in some other projects to work around JSON's lack of comments is to write the "source code" for a file like this in a format like YAML or TOML (which *can* have comments and is generally more human-authorable), and then run something similar to https://gitlab.steamos.cloud/steamrt/steamrt/-/blob/steamrt/scout/abi/yaml-to-json.py?ref_type=heads at build-time to convert it to JSON for consumption by lower-level components. It might be worth considering that here.

Source/Steam/CompatTool.cpp
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  fextl::string Data;
+  if (!FEXCore::FileLoading::LoadFile(Data, ConfigTemplatePath)) {
+    return {};
+  }
+
+  const char* RuntimeDir = getenv("XDG_RUNTIME_DIR");
+  if (!RuntimeDir) {
+    RuntimeDir = "/run/user/1000";
```
smcv: 1000 is a uid, not a magic number, so `geteuid()` (same as shell `id -u`) would be more appropriate. But in environments where `$XDG_RUNTIME_DIR` isn't set, `/run/user/$(id -u)` most likely isn't going to exist either, so it doesn't make a great fallback: attempts to write to it will just fail.

GLib falls back to `$XDG_CACHE_HOME` (or `~/.cache`) if `$XDG_RUNTIME_DIR` isn't set, and that seems like a reasonable "least-bad" approach to this.

Source/Steam/CompatTool.cpp
```diff
@@ -0,0 +1,165 @@
...
+    setenv("FEX_APP_CONFIG", AppConfigPath.c_str(), true);
+  }
+
+  const auto FEXInterpreterPath = PortableInfo.InterpreterPath + "usr/bin/FEX";
+
+  // Due to no arguments for this application, just replace argv[0] and execve again.
+  argv[0] = FEXInterpreterPath.c_str();
+  execv(FEXInterpreterPath.c_str(), const_cast<char* const*>(argv));
+
+  return -errno;
```
smcv: If you can't find/exec the executable I think it would be better to spam to stderr (`perror (argv[0])` or similar), otherwise there will be no trace in the Steam logs unless the user happens to have requested pressure-vessel debug output (in which case it will say something like `exit status 2`, but that's a lot less clear than `/path/to/FEX: No such file or directory`).

I know FEX tries not to write to stderr in general, because it's a general-purpose interpreter - but this compatibility tool wrapper can know that it's being invoked by pressure-vessel, or more generally by Steam, and in that situation it's OK to assume that stderr has been pointed somewhere useful (in practice a log file in `~/.steam/steam/logs/` or a pipe to `srt-logger`).

I'd also recommend returning an `env(1)`-style exit status: internal error → 125, `ENOENT` → 127, any other `execve()` failure → 126. pressure-vessel uses that convention, to try to preserve exit statuses from the invoked program as much as possible.

Source/Steam/emulator.json
```diff
@@ -0,0 +1,12 @@
...
+  "emulator_v0": {
+    "argv": "./usr/bin/FEX",
+    "environment": { "FEX_PORTABLE": "1" },
+    "container_argv": "./usr/bin/FEX",
+    "container_environment": { "FEX_ROOTFS": "" },
+    "main_argv": "./FEXCompatTool",
```
smcv, line 7: It's `FEXCompatTool` here, but `fex-compat-tool` in the build system when you compile it. Don't the two names need to be the same?

Source/Steam/emulator.json
Sonicadvance1: I've avoided this edge case by reading the steam environment variables directly inside of FEX's config system.

Source/Steam/emulator.json
Sonicadvance1: Oops.

Source/Steam/CompatTool.cpp
Sonicadvance1: Switched to ENOENT and added a perror.

Source/Steam/emulator.json
Sonicadvance1: Potentially worth adding in the future but not super necessary initially.

Source/Steam/CompatTool.cpp
```diff
@@ -0,0 +1,181 @@
...
+  const auto FEXInterpreterPath = PortableInfo.InterpreterPath + "usr/bin/FEX";
+
+  // Due to no arguments for this application, just replace argv[0] and execve again.
+  argv[0] = FEXInterpreterPath.c_str();
+  execv(FEXInterpreterPath.c_str(), const_cast<char* const*>(argv));
+
+  perror(argv[0]);
+
+  // ENOENT
+  return 127;
```
smcv: Ah, what I meant was something like this (untested!):


```suggestion
  execv(FEXInterpreterPath.c_str(), const_cast<char* const*>(argv));
  int saved_errno = errno;

  perror(argv[0]);

  if (saved_errno == ENOENT)
    return 127;

  return 126;
```

Sonicadvance1: Good shout, did the equivalent behaviour.

Sonicadvance1:
Added support for the new server_argv ready message. So this is good to go.

neobrain (approved):

Source/Steam/ServerManager.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+  // stdin will be a pipe, so wait until that FD is closed.
+  while (true) {
+    pollfd p {
+      .fd = STDIN_FILENO,
+      .events = POLLRDHUP,
+      .revents = 0,
+    };
+
+    int events = poll(&p, 1, -1);
```
smcv, line 83: Please could you also poll for `POLLIN`, and if it's seen, `read()` from the socket? Like my mockup server used in the PV unit tests, https://gitlab.steamos.cloud/steamrt/steam-runtime-tools/-/blob/wip/task870/tests/mock-emulator-server.c, does?

At the moment, if pressure-vessel for whatever reason writes more than `PIPE_BUF` bytes to the write end of the pipe you've received as stdin, we'll run out of pipe buffer space, and either block indefinitely or get an error. Right now we don't write anything (so the pipe will just close with 0 bytes available), but in future we could, and it would be a shame if we had to invent a capability-negotiation mechanism in order to make this work.

(My mockup doesn't actually use the Linux-specific `POLLRDHUP`, it just relies on `read()` returning 0 bytes if and only if we reached EOF.)

Source/Steam/ServerManager.cpp
```diff
@@ -0,0 +1,96 @@
...
+    if (events == -1 && errno == EINTR) {
+      continue;
+    }
+
+    if (events > 0 && (p.revents & (POLLRDHUP | POLLERR | POLLHUP | POLLNVAL))) {
+      // Error or pressure-vessel hung-up.
```
smcv, line 89: If you want to handle errors as being equivalent to EOF, you probably want `events < 0` as well.

Source/Steam/ServerManager.cpp
```diff
@@ -0,0 +1,96 @@
...
+  // Reload the meta layer
+  FEXCore::Config::ReloadMetaLayer();
+
+  auto pipes = get_pipe();
+
+  // Set the write side to close on exec.
+  fcntl(pipes.write_pipe, F_SETFD, FD_CLOEXEC);
+
+  // Give the read end of the pipe to FEXServer.
+  auto ServerFD = FEXServerClient::StartServer(PortableInfo.InterpreterPath, pipes.read_pipe);
```
smcv, line 57: What does the server do with the read end of the pipe? Does it, similarly, wait until EOF is reached and then exit?

If yes, then instead of implementing your own pipe, you could probably just `dup()` the server manager's original `STDIN_FILENO`, pass the duplicate down to the server, and close both stdin and the duplicate in the server manager.

smcv: Oh, but then you wouldn't know when the server manager can exit, so you'd probably have to `execve()` the server (replacing yourself with the server) rather than having run-and-wait logic. That would save us a process, but it's a rather different execution model, so I can understand if you'd rather not.

PR #5095, opened 2025-12-02
JIT: Add support for serializing relocations

neobrain:
... and move the relocation code to CodeCache.cpp, since that's the only user of ApplyCodeRelocations.

Serialized data must be independent of any data that is random at runtime, such as FEX-internal symbols or guest-addresses. This PR takes care of the former by allowing ApplyCodeRelocations to be used in reverse (zero-ing out FEX-internal symbols). Further, code is added to subtract the base guest address from any GuestRIP-relative constants when querying generated relocations for serialization.

This is the last piece needed before we can start writing code caches.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -88,14 +88,18 @@ public:
    * When loading a code cache, they are used to move cached code to the
    * dynamically chosen base address of the guest binary.
    *
+   * Conversely, relocations are applied in reverse when writing code caches
+   * to ensure consistency across generation runs.
+   *
    * Note that FEX relocations are unrelated to ELF/PE relocations.
    *
    * @param GuestDelta Guest address offset to apply to RIP-relative data
+   * @param ForStorage True for serializing data (producing deterministic output); false for de-serializing it (resolving dynamic symbols)
    *
    * @return Returns true on success
    */
   [[nodiscard]]
-  bool ApplyCodeRelocations(uint64_t GuestDelta, std::span<std::byte> Code, std::span<const CPU::Relocation> Relocations);
+  bool ApplyCodeRelocations(uint64_t GuestDelta, std::span<std::byte> Code, std::span<const CPU::Relocation> Relocations, bool ForStorage);
 };
 
 class ContextImpl final : public FEXCore::Context::Context, public CPU::CodeBufferManager {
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -591,10 +589,7 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
 
   fextl::vector<FEXCore::CPU::Relocation> Relocations;
 
-  ///< Relocation code loading
-  bool ApplyRelocations(uint64_t GuestEntry, std::span<std::byte> Code, std::span<const FEXCore::CPU::Relocation>);
-
-  fextl::vector<FEXCore::CPU::Relocation> TakeRelocations() override;
+  fextl::vector<FEXCore::CPU::Relocation> TakeRelocations(uint64_t GuestBaseAddress) override;
```
bylaws, line 599: nit: Maybe add a comment that they will be rebased relative to GuestBaseAddress?

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -80,6 +80,26 @@ class CodeCache : public AbstractCodeCache {
...
+   * When loading a code cache, they are used to move cached code to the
+   * dynamically chosen base address of the guest binary.
+   *
+   * Conversely, relocations are applied in reverse when writing code caches
+   * to ensure consistency across generation runs.
+   *
+   * Note that FEX relocations are unrelated to ELF/PE relocations.
+   *
+   * @param GuestDelta Guest address offset to apply to RIP-relative data
+   * @param ForStorage True for serializing data; false for de-serializing it
```
bylaws: maybe invert this and call it something like ResolveSymbols?

neobrain: I like the idea, but both settings trigger an active action (one writes 0, the other writes the new symbol address), so `ResolveSymbols` specifically wouldn't be a suitable name since it suggests that `false` would be a no-op.

bylaws: ah hmm, if can't think of a nicer name, detailing what it actually does in the comment would be fine too i feel

FEXCore/Source/Interface/Core/JIT/JITClass.h
neobrain: This is the only interface through which Relocations can leave the JIT module, so this would be an implementation detail. I'll add a comment though to explain the GuestBaseAddress parameter, since it's not exactly obvious where it should come from (and that the assumption is that all Relocations are generated from a single binary).

FEXCore/Source/Interface/Context/Context.h
neobrain: How about this?

`   * @param ForStorage True for serializing data (producing deterministic output); false for de-serializing it (resolving dynamic symbols)`

Sonicadvance1 (approved):

Issue #4792, opened 2025-08-21
Ubuntu 22.04 maintenance burdens


neobrain:
Tracking maintenance burdens from outdated packages in Ubuntu 22.04 here:
* Latest compiler version is clang-15:
  * `ranges::split` is broken (fixed in clang-16)
* FEXConfig must be built against Qt 5 since Qt 6 is not properly integrated

This list will be extended as we encounter new issues.

PR #5096, opened 2025-12-03
CodeCache: Implement offline compiler for cache generation

neobrain:
## Overview

Code caches should be compiled out-of-process to ensure a consistent environment and to minimize interference with emulation. For this purpose, this PR adds a new `FEXOfflineCompiler` executable that generates a cache from block offsets recorded in a user-provided code map.

To test out cache generation, enable the `EnableCodeCachingWIP` setting and follow these steps:
1. Run some application (e.g. SuperMeatBoy) and close it
2. Open a shell and verify that FEX created a file like `~/.cache/fex-emu/codemap/SuperMeatBoy-038d4ceff5e2271a.0.bin`
3. Run `FEXOfflineCompiler generate ~/.cache/fex-emu/codemap/SuperMeatBoy-038d4ceff5e2271a.0.bin --fileid 038d4ceff5e2271a`

The last command will also print a list of referenced libraries, which can be used to subsequently generate caches for these libraries by changing the `--fileid` argument:
> Parsed 709 codemap entries for ROOTFS/usr/lib/x86_64-linux-gnu/libc.so.6 (3d01bba8e94845de)

In a future change, this cache generation process will be automated and orchestrated by FEXServer.

### DIFF
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -2,20 +2,20 @@
 #include "Utils/SpinWaitLock.h"
 
 #include <Interface/Context/Context.h>
+#include <Interface/Core/ArchHelpers/Arm64Emitter.h>
 #include <Interface/Core/JIT/Relocations.h>
+#include <Interface/Core/LookupCache.h>
 
+#include <FEXCore/Core/Thunks.h>
 #include <FEXCore/HLE/SourcecodeResolver.h>
 
 #include <FEXHeaderUtils/Filesystem.h>
 
-#include <fstream>
-#include <thread>
+#include <git_version.h>
 
 #include <xxhash.h>
 
-#include <Interface/Core/ArchHelpers/Arm64Emitter.h>
-
-#include <FEXCore/Core/Thunks.h>
+#include <fstream>
 
 namespace FEXCore {
 
@@ -226,12 +226,106 @@ uint64_t CodeCache::ComputeCodeMapId(std::string_view Filename, int FD) {
   return XXH3_64bits(Filename.data(), Filename.size());
 }
 
+struct CodeCacheHeader {
+  char Magic[4] = {'F', 'X', 'C', 'C'};
+  uint32_t FormatVersion = 1;
+  char FEXVersion[8] = {};
+  uint32_t NumBlocks;
+  uint32_t NumCodePages;
+  uint32_t CodeBufferSize;
+  uint32_t NumRelocations;
+  uint64_t SerializedBaseAddress;
+  // TODO: Consider including information from LookupCache.BlockLinks
+};
+
 void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCacheFile, const ExecutableFileSectionInfo& GuestRIPLookup) {
   // TODO
 }
 
+template<typename T>
+static constexpr auto IsOrderedContainer(const T&) -> std::false_type;
+template<typename... T>
+static constexpr auto IsOrderedContainer(const std::map<T...>&) -> std::true_type;
+template<typename... T>
+static constexpr auto IsOrderedContainer(const std::set<T...>&) -> std::true_type;
+
 bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const ExecutableFileSectionInfo& SourceBinary, uint64_t SerializedBaseAddress) {
-  // TODO
+  auto CodeBuffer = CTX.GetLatest();
+  auto& LookupCache = *Thread.LookupCache->Shared;
+
+  auto Relocations = Thread.CPUBackend->TakeRelocations(SourceBinary.FileStartVA);
+
+  // Write file header
+  CodeCacheHeader header;
+  memcpy(&header.FEXVersion[0], GIT_SHORT_HASH, strlen(GIT_SHORT_HASH));
+  header.NumBlocks = LookupCache.BlockList.size();
+  header.NumCodePages = LookupCache.CodePages.size();
+  header.CodeBufferSize = CTX.LatestOffset;
+  header.NumRelocations = Relocations.size();
+  header.SerializedBaseAddress = SerializedBaseAddress;
+  ::write(fd, &header, sizeof(header));
+
+  // Dump guest<->host block mappings
+  {
+    // Cache contents must be deterministic, so copy the unordered block list and then sort by key
+    static_assert(!decltype(IsOrderedContainer(LookupCache.BlockList))::value, "Already deterministic; drop temporary container");
+    fextl::vector<std::pair<uint64_t, const GuestToHostMap::BlockEntry*>> BlockList;
+    BlockList.reserve(LookupCache.BlockList.size());
+    for (auto& [Guest, BlockEntry] : LookupCache.BlockList) {
+      static_assert(sizeof(Guest) == 8, "Breaking change in code cache data layout");
+      BlockList.emplace_back(Guest, &BlockEntry);
+    }
+    std::ranges::sort(BlockList);
+
+    for (auto [Guest, Host] : BlockList) {
+      static_assert(sizeof(Host->HostCode) == 8, "Breaking change in code cache data layout");
+      static_assert(sizeof(Host->CodePages[0]) == 8, "Breaking change in code cache data layout");
+
+      Guest -= SourceBinary.FileStartVA;
+      ::write(fd, &Guest, sizeof(Guest));
+      uint64_t HostCode = Host->HostCode - reinterpret_cast<uintptr_t>(CodeBuffer->Ptr);
+      ::write(fd, &HostCode, sizeof(HostCode));
+      uint64_t NumCodePages = Host->CodePages.size();
+      ::write(fd, &NumCodePages, sizeof(NumCodePages));
+      LOGMAN_THROW_A_FMT(std::ranges::is_sorted(Host->CodePages), "Code pages aren't sorted");
+      for (auto CodePage : Host->CodePages) {
+        CodePage -= SourceBinary.FileStartVA;
+        ::write(fd, &CodePage, sizeof(CodePage));
+      }
+    }
+  }
+
+  // Dump relocations
+  static_assert(sizeof(Relocations[0]) == 48, "Breaking change in code cache data layout");
+  ::write(fd, Relocations.data(), Relocations.size() * sizeof(Relocations[0]));
+
+  // Pad to next page in file so that the CodeBuffer can be mmap'ed into process on load
+  char Zero[64] {};
+  auto Off = lseek(fd, 0, SEEK_CUR);
+  while (Off != AlignUp(Off, Utils::FEX_PAGE_SIZE)) {
+    auto BytesToWrite = std::min(AlignUp(Off, Utils::FEX_PAGE_SIZE) - Off, sizeof(Zero));
+    ::write(fd, Zero, BytesToWrite);
+    Off += BytesToWrite;
+  }
+
+  // Dump the host code (relocated for position-independent serialization)
+  std::vector CodeBufferData(reinterpret_cast<std::byte*>(CodeBuffer->Ptr), reinterpret_cast<std::byte*>(CodeBuffer->Ptr) + CTX.LatestOffset);
+  if (!ApplyCodeRelocations(SerializedBaseAddress, CodeBufferData, Relocations, true)) {
+    LOGMAN_THROW_A_FMT(false, "Failed to apply code relocations");
+    return false;
+  }
+  ::write(fd, CodeBufferData.data(), CodeBufferData.size());
+
+  // Dump code pages
+  static_assert(decltype(IsOrderedContainer(LookupCache.CodePages))::value, "Non-deterministic data source");
+  for (auto& [Page, Entrypoints] : LookupCache.CodePages) {
+    static_assert(sizeof(Page) == 8, "Breaking change in code cache data layout");
+    ::write(fd, &Page, sizeof(Page));
+    uint64_t NumEntrypoints = Entrypoints.size();
+    ::write(fd, &NumEntrypoints, sizeof(NumEntrypoints));
+    ::write(fd, Entrypoints.data(), Entrypoints.size() * sizeof(Entrypoints[0]));
+  }
+
   return true;
 }
 
--- a/Source/Tools/FEXOfflineCompiler/Main.cpp
+++ b/Source/Tools/FEXOfflineCompiler/Main.cpp
@@ -1,0 +1,280 @@
+// SPDX-License-Identifier: MIT
+#include "../FEXInterpreter/ELFCodeLoader.h"
+#include <DummyHandlers.h>
+#include <PortabilityInfo.h>
+#include <Thunks.h>
+
+#include <FEXCore/Core/CodeCache.h>
+#include <FEXCore/Core/Context.h>
+#include <FEXCore/Core/HostFeatures.h>
+
+#include <Common/ArgumentLoader.h>
+#include <Common/Config.h>
+#include <Common/FEXServerClient.h>
+#include <Common/HostFeatures.h>
+
+#include <OptionParser.h>
+
+#include <fmt/printf.h>
+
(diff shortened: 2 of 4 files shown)


neobrain:
clang-15 doesn't support aggregate initialization with `std::piecewise_construct`, so here's another issue for #4792.

Source/Tools/FEXOfflineCompiler/Main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  }
+
+  auto CTX = FEXCore::Context::Context::CreateNewContext(HostFeatures);
+
+  auto SignalDelegation = std::make_unique<DummySignalDelegator>();
+
+  auto SyscallOSABI = Is64Bit ? FEXCore::HLE::SyscallOSABI::OS_LINUX64 : FEXCore::HLE::SyscallOSABI::OS_LINUX32;
+  auto SyscallHandler = std::make_unique<AOTSyscallHandler>(SyscallOSABI);
+
+  // TODO: How to handle VDSO with code caching? Does this need purpose-specific FEX relocations?
```
neobrain: @Sonicadvance1 Do we need to set up VDSO things for offline compilation? Doesn't seem to be needed, but wanted to double-check.

Sonicadvance1: Should be fine to avoid offline compilation of VDSO, we're not expecting those to grow to thousands of symbols, and most of them immediately thunk out anyway.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  struct Header {
+    char Magic[4] = {'F', 'X', 'C', 'C'};
+    uint32_t FormatVersion = 1;
+    char FEXVersion[8] = {};
+    uint32_t NumBlocks;
+    uint32_t NumCodePages;
+    uint32_t CodeBufferSize;
+    uint32_t NumRelocations;
+    uint64_t SerializedBaseAddress;
+    // TODO: Consider including information from LookupCache.BlockLinks
```
bylaws: I would expect that to be empty unless the code itself has been ran, or is the idea here to prelink?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
 bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const ExecutableFileSectionInfo& SourceBinary, uint64_t SerializedBaseAddress) {
-  // TODO
+  auto CodeBuffer = CTX.GetLatest();
+  auto& LookupCache = *Thread.LookupCache->Shared;
+
+  auto Relocations = Thread.CPUBackend->TakeRelocations(SourceBinary.FileStartVA);
+
+  // Write file header
+  struct Header {
+    char Magic[4] = {'F', 'X', 'C', 'C'};
```
bylaws: Maybe make this a constexpr member outside, same with version/the whole struct could go in a header. Perhaps have a ctor to populate it rather than doing in-line also

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  header.NumBlocks = LookupCache.BlockList.size();
+  header.NumCodePages = LookupCache.CodePages.size();
+  header.CodeBufferSize = CTX.LatestOffset;
+  header.NumRelocations = Relocations.size();
+  header.SerializedBaseAddress = SerializedBaseAddress;
+  ::write(fd, &header, sizeof(header));
+
+  // Dump guest<->host block mappings
+  {
+    fextl::vector<decltype(LookupCache.BlockList)::value_type> BlockList {LookupCache.BlockList.begin(), LookupCache.BlockList.end()};
```
bylaws: Unused?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+      }
+    }
+  }
+
+  // Dump relocations
+  ::write(fd, Relocations.data(), Relocations.size() * sizeof(Relocations[0]));
+
+  // Pad to next page in file so that the CodeBuffer can be mmap'ed into process on load
+  // TODO: lseek not currently available with our custom CRT
+#ifndef _WIN32
```
bylaws: You can just add this within the PR (https://github.com/bylaws/FEX/blame/arm64eccc/Source/Windows/Common/CRT/IO.cpp#L203) we already have _lseek so just adding alias.  Also note this code doesn't even need to work with the FEX CRT, just needs to link successfully. The windows offline compiler can depend on the standard CRT.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  {
+    fextl::vector<decltype(LookupCache.BlockList)::value_type> BlockList {LookupCache.BlockList.begin(), LookupCache.BlockList.end()};
+    for (auto [Guest, Host] : LookupCache.BlockList) {
+      Guest -= SourceBinary.FileStartVA;
+      ::write(fd, &Guest, sizeof(Guest));
+      Host.HostCode -= reinterpret_cast<uintptr_t>(CodeBuffer->Ptr);
+      ::write(fd, &Host.HostCode, sizeof(Host.HostCode));
+      uint64_t NumCodePages = Host.CodePages.size();
+      ::write(fd, &NumCodePages, sizeof(NumCodePages));
+      LOGMAN_THROW_A_FMT(std::ranges::is_sorted(Host.CodePages), "Code pages aren't sorted");
```
bylaws: This doesn't need to be sorted

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  header.NumCodePages = LookupCache.CodePages.size();
+  header.CodeBufferSize = CTX.LatestOffset;
+  header.NumRelocations = Relocations.size();
+  header.SerializedBaseAddress = SerializedBaseAddress;
+  ::write(fd, &header, sizeof(header));
+
+  // Dump guest<->host block mappings
+  {
+    fextl::vector<decltype(LookupCache.BlockList)::value_type> BlockList {LookupCache.BlockList.begin(), LookupCache.BlockList.end()};
+    for (auto [Guest, Host] : LookupCache.BlockList) {
```
bylaws: I was thinking it would be reasonable to add static asserts on the sizes of BlockList members/everything we write directly? Feel that would be helpful for parsing manually etc and avoid accidental mistakes. Same for e.g. size of a relocation structure, and comments pointing to code cache at the side of the struct definition.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -225,7 +231,80 @@ void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  auto Off = lseek(fd, 0, SEEK_CUR);
+  while (Off != AlignUp(Off, Utils::FEX_PAGE_SIZE)) {
+    auto BytesToWrite = std::min(AlignUp(Off, Utils::FEX_PAGE_SIZE) - Off, sizeof(Zero));
+    ::write(fd, Zero, BytesToWrite);
+    Off += BytesToWrite;
+  }
+#endif
+
+  // Dump the host code (relocated for position-independent serialization)
+  std::vector<std::byte> CodeBufferData((std::byte*)CodeBuffer->Ptr, (std::byte*)CodeBuffer->Ptr + CTX.LatestOffset);
```
bylaws: reinterpret_cast

Source/Tools/FEXOfflineCompiler/Main.cpp
```diff
@@ -0,0 +1,284 @@
...
+    }
+    return Ret;
+  }
+
+  uint64_t GuestMunmap(FEXCore::Core::InternalThreadState*, void* addr, uint64_t length) override {
+    return munmap(addr, length);
+  }
+};
+
+class DummySignalDelegator final : public FEXCore::SignalDelegator {};
```
bylaws: FEX::DummyHandlers::DummySignalDelegator

Source/Tools/FEXOfflineCompiler/Main.cpp
```diff
@@ -0,0 +1,284 @@
...
+  optparse::OptionParser Parser {};
+  Parser.add_option("--outdir").set_default(FEX::Config::GetCacheDirectory() + "cache").help("Output directory for generated cache files");
+  Parser.add_option("--fileid").help("Select binary to generate cache for");
+
+  optparse::Values Options = Parser.parse_args(argc, argv);
+  if (Parser.args().size() != 1) {
+    Parser.print_usage();
+    return 1;
+  }
+  const auto CodeMapPath = (std::string)Parser.args()[0];
```
bylaws: Did not know casts like this were even possible, maybe better to just not use auto and do std::string CodeMapPath(...)

Source/Tools/FEXOfflineCompiler/Main.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+  fextl::string OutDir(Options.get("outdir"));
+  if (!OutDir.ends_with('/')) {
+    OutDir.push_back('/');
+  }
+  std::filesystem::create_directories(OutDir);
+
+  const auto PortableInfo = FEX::ReadPortabilityInformation();
+  char* envp[] = {nullptr};
+  FEX::Config::LoadConfig("", envp, PortableInfo);
```
bylaws, line 253: It's worth making this an option or picking it up from codemaps etc, or if not at least a comment/TODO. Per-game configs should still work with caching ideally. (I realise FEXServer makes this awkward as you need to serialize env vars in a way WoA doesn't need as the compiler always runs under the game context)

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Yes, this would only be needed if we decide to prelink (across pages?) in the future. As-is, the block link list is always empty on cache generation.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: It does to produce deterministic code caches. All other contents are implicitly sorted by virtue of using ordered containers (or should be, anyway).

bylaws: Ah, a comment is probably warranted then to indicate determinism is desired here

Source/Tools/FEXOfflineCompiler/Main.cpp
neobrain: The config will be passed from FEXServer over shared memory in the future, so this code will change a fair bit. Seems easy to adapt for game configs either way, so I think it's fine without a TODO?

bylaws: Oh will FEXServer be invoking OfflineCompiler rather than calling it as a library?

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: ... in fact, we shouldn't iterate over LookupCache.BlockList but rather over its sorted contents, since that is *not* an ordered container.

Source/Tools/FEXOfflineCompiler/Main.cpp
neobrain: Yes, it always worked like this. We have no robustness guarantees in the JIT, so a crash during cache generation would take down FEXServer without helper processes.

(Also it's much easier to ensure a consistent compile environment if we start fresh every time instead of trying to clean up after previous runs)

bylaws: Ah neat, yeah agreed

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: It's a leftover of previous code that sorted the contents of LookupCache.BlockList, which I accidentally removed later :grimacing:

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: That sounds like it would express the same with more lines of code that is also less localized. I agree with moving the struct out-of-function though since it will be used by LoadData as well.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Not sure what type of `static_assert` would really be useful here, since this code itself is already agnostic to individual field sizes and asserting hardcoded sizes elsewhere seems somewhat arbitrary.

Perhaps a comment in LookupCache.h's BlockList declaration with a general warning about changes to data layout would help?

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: My feeling is that copying in the fex version etc makes sense in a ctor more than here, it logically goes together with the magic etc.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: I was meaning field sizes, both to make it easy to understand layout without bouncing back and forth, and to fail on any changes. The code may be agnostic but file format itself isn't imo

bylaws: just like static_assert(sizeof(Host.HostCode) == 4) etc.

neobrain: Found a reasonably clean way to add these while also adding checks for ordering of containers. Does this match what you had in mind?

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Moved the struct out of function, but leaving the use of direct initializers for now.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Yeah that's really nice :)

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -220,12 +226,106 @@ uint64_t CodeCache::ComputeCodeMapId(std::string_view Filename, int FD) {
...
+  uint32_t NumRelocations;
+  uint64_t SerializedBaseAddress;
+  // TODO: Consider including information from LookupCache.BlockLinks
+};
+
 void CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCacheFile, const ExecutableFileSectionInfo& GuestRIPLookup) {
   // TODO
 }
 
+template<typename T>
```
bylaws, line 245: Totally not necessary and fine as is, but came up with:
```
template<typename T>
concept OrderedContainer = requires {
    typename T::key_compare;
};
```

if you wanted to be fancy

Sonicadvance1 (approved):

Sonicadvance1:
Fancy can come in as a follow-up PR if desired or not. Let's go~!

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: That's a cute trick way to do it :)

I tried using `requires` statements inline, but they didn't produce usable error messages. I think it only works in a template context like yours, so I'll play around with it.

PR #5097, opened 2025-12-04
Remap /proc/pid/cmdline with PR_SET_MM_MAP

wcampbell-nv:
This change remaps the process' argument data to point to the argument array FEX creates on the emulated stack. This means `/proc/pid/cmdline` will output the same command line as if the program had been launched natively. The FEX exe is stripped from the cmdline and application writes to argv[0] are now (correctly) reflected in the cmdline. See issue https://github.com/FEX-Emu/FEX/issues/5069 for more context.

Credit to @Amanieu for the idea. The implementation is based on `man proc_pid_stat` and the kernel source code for `prctl_set_mm_map`.

**Testing**
I've confirmed that emulated code's reads of `/proc/self/cmdline` reflect application writes to argv[0], via the test program in https://github.com/FEX-Emu/FEX/issues/5069.

I've confirmed that thunk'd aarch64 code's reads of `/proc/self/cmdline` now reflect FEX's rewritten argument data, via logging in a Vulkan layer.

I've confirmed that other process see the remapped argument data, via `echo /proc/pid/cmdline`. The new command line data also shows up in htop, no more sifting through a million rows of 'FEX'!

### DIFF
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -17,6 +17,7 @@
 #include <FEXCore/Core/X86Enums.h>
 #include <FEXCore/Utils/LogManager.h>
 #include <FEXCore/Utils/TypeDefines.h>
+#include <FEXCore/Utils/FileLoading.h>
 #include <FEXCore/fextl/list.h>
 #include <FEXCore/fextl/string.h>
 #include <FEXCore/fextl/vector.h>
@@ -711,21 +712,14 @@ public:
 
   // Get the current memory map from /proc/self/stat
   static bool GetCurrentMap(struct prctl_mm_map& map) {
-    int fd = open("/proc/self/stat", O_RDONLY | O_CLOEXEC);
-    if (fd < 0) {
-      return false;
-    }
 
     // /proc/self/stat has 52 fields of at most 20 digits each (UINT64_MAX).
     // 52*20 = 1040, so 2048 is a conservative upper bound
     char stat_buffer[2048];
-    ssize_t bytes_read = read(fd, stat_buffer, sizeof(stat_buffer) - 1);
-    close(fd);
+    ssize_t bytes_read = FEXCore::FileLoading::LoadFileToBuffer("/proc/self/stat", stat_buffer);
 
-    if (bytes_read <= 0) {
-      return false;
-    }
-    stat_buffer[bytes_read] = '\0'; // So we don't read past to garbage data
+    // Ensure we don't read past the end into garbage data
+    stat_buffer[std::clamp(bytes_read, 0L, static_cast<ssize_t>(sizeof(stat_buffer)) - 1)] = '\0';
 
     // See man proc_pid_stat
     int items_read = sscanf(stat_buffer,
@@ -766,10 +760,10 @@ public:
 
       int r = prctl(PR_SET_MM, PR_SET_MM_MAP, &map, sizeof(map), 0L);
       if (r != 0) {
-        LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data. prctl failed: result {}, errno {}", r, errno);
+        LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data (prctl failed: result {}, errno {})", r, errno);
       }
     } else {
-      LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data. GetCurrentMap failed. ");
+      LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data (GetCurrentMap failed)");
     }
   }
 


wcampbell-nv:
Thanks, I didn't know of that restriction. I've swapped to read from a FD, the tests are passing now.

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -707,6 +709,70 @@ class ELFCodeLoader final : public FEX::CodeLoader {
...
+                            "%*d %*d %*d %*d %*d "      // 16 to 20
+                            "%*d %*u %*u %*d %*u "      // 21 to 25
+                            "%llu %llu %llu %*u %*u "   // 26 to 30
+                            "%*u %*u %*u %*u %*u "      // 31 to 35
+                            "%*u %*u %*d %*d %*u "      // 36 to 40
+                            "%*u %*u %*u %*d %llu "     // 40 to 45
+                            "%llu %llu %llu %llu %llu " // 46 to 50
+                            "%llu",                     // 51
+                            &map.start_code, &map.end_code, &map.start_stack, &map.start_data, &map.end_data, &map.start_brk,
+                            &map.arg_start, &map.arg_end, &map.env_start, &map.env_end);
```
Amanieu, line 738: You shouldn't need to read from /proc/self/stat, you already have all of the data available from the ELF loader and in the stack setup.

wcampbell-nv: We know these addresses on the emulated stack we create, but do we know the OS's original addresses? This change only remaps the argument data. I don't know the implications of remapping the rest of them.

Amanieu: For everything except `brk` and `start_brk`, this only affects data exposed in `/proc/self/{stat,auxv,cmdline,environ}`. You have to leave `exe_file = -1` since changing that field actually requires elevated permissions.

The brk-related stuff is tricky. You can set it to the application's brk (which allows it to use the native brk syscall), but then you need to ensure that FEX itself never uses brk ever again, including via libc (which is used by thunked libraries). It's probably not worth it.

Sonicadvance1: We actually hard disable brk early in the init process by allocating the next page inside of it and consuming all bytes until that page.

Amanieu: That doesn't stop glibc from actually trying to use the brk syscall though, and it will get very confused if the brk suddenly shifts to a completely different address.

Sonicadvance1: True, might get weird even when we control all allocations.

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -707,6 +709,70 @@ class ELFCodeLoader final : public FEX::CodeLoader {
...
+    if (fd < 0) {
+      return false;
+    }
+
+    // /proc/self/stat has 52 fields of at most 20 digits each (UINT64_MAX).
+    // 52*20 = 1040, so 2048 is a conservative upper bound
+    char stat_buffer[2048];
+    ssize_t bytes_read = read(fd, stat_buffer, sizeof(stat_buffer) - 1);
+    close(fd);
+
```
neobrain: We have a helper to load a whole file into a buffer:
```suggestion
    // /proc/self/stat has 52 fields of at most 20 digits each (UINT64_MAX).
    // 52*20 = 1040, so 2048 is a conservative upper bound
    char stat_buffer[2048];
    auto bytes_read = LoadFileToBuffer("/proc/self/stat", stat_buffer);
```

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -707,6 +709,70 @@ class ELFCodeLoader final : public FEX::CodeLoader {
...
+    if (GetCurrentMap(map)) {
+      map.arg_start = NewArgStart;
+      map.arg_end = NewArgStart + ArgSize;
+
+      int r = prctl(PR_SET_MM, PR_SET_MM_MAP, &map, sizeof(map), 0L);
+      if (r != 0) {
+        LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data. prctl failed: result {}, errno {}", r, errno);
+      }
+    } else {
+      LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data. GetCurrentMap failed. ");
```
neobrain: Nit:
```suggestion
      LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data (GetCurrentMap failed)");
```

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -707,6 +709,70 @@ class ELFCodeLoader final : public FEX::CodeLoader {
...
+  // Point the OS to our new stack's argument data
+  void RemapArgumentData(uintptr_t NewArgStart, uint64_t ArgSize) {
+    struct prctl_mm_map map {};
+    if (GetCurrentMap(map)) {
+      map.arg_start = NewArgStart;
+      map.arg_end = NewArgStart + ArgSize;
+
+      int r = prctl(PR_SET_MM, PR_SET_MM_MAP, &map, sizeof(map), 0L);
+      if (r != 0) {
+        LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data. prctl failed: result {}, errno {}", r, errno);
```
neobrain:
```suggestion
        LogMan::Msg::EFmt("Failed to remap /proc/pid/cmdline data (prctl failed: result {}, errno {})", r, errno);
```

Source/Tools/FEXInterpreter/ELFCodeLoader.h
neobrain: (If you want to keep supporting the exceptional cases, we can replace the `<=` check with `stat_buffer[std::clamp(bytes_read, 0, sizeof(stat_buffer) - 1] = 0;`)

Source/Tools/FEXInterpreter/ELFCodeLoader.h
wcampbell-nv: So I gather we'll want to continue querying `start_brk` from `/proc/self/stat`. I prefer leaving the rest the mapping alone to lower the surface area of this change. We can update the rest of the mappings in a follow up change if desired.

PR #5099, opened 2025-12-04
Tools/pidof: Fixes FEXpidof after #5097

Sonicadvance1:
FEX argument will no longer exist in `/proc/<pid>/cmdline` after this PR. Ensure this tool still works.

After #5097 is merged, one can use regular `pidof` for Linux applications, but this is still useful as it finds Wine applications running with FEX as well.

### DIFF
--- a/Source/Tools/pidof/pidof.cpp
+++ b/Source/Tools/pidof/pidof.cpp
@@ -328,6 +328,12 @@ int main(int argc, char** argv) {
       continue;
     }
 
+    // If matching all "FEX" instances then add arm64ec/wow64 FEX to the matched list.
+    if (ProgramArg != -1 && Config::AllFEX) {
+      MatchedPIDs.emplace(pid.pid);
+      continue;
+    }
+
     ProgramPair Arg = FindEmulatedWineArgument(ProgramArg, Args, IsWine);
     bool Matched = false;
     for (const auto& CompareProgram : Config::Programs) {


Sonicadvance1:
Added support for `FEXpidof FEX` special case so that `kill -9 $(FEXPidof FEX)` still gives us equivalent behaviour as `kill -9 $(pidof FEX)`.

wcampbell-nv:
> but this is still useful as it finds Wine applications running with FEX as well.

I'm not sure this is necessary anymore, Wine similarly re-writes its argv[0] data so that `/proc/pid/cmdline` begins with the target .exe.

Sonicadvance1:
> > but this is still useful as it finds Wine applications running with FEX as well.
> 
> I'm not sure this is necessary anymore, Wine similarly re-writes its argv[0] data so that `/proc/pid/cmdline` begins with the target .exe.

hm. I was going to say that it still did a bit more than pidof, but actually because we rename cmdline away, we actually can't reliably detect wine anymore. It does still provide use for `FEXpidof FEX` but it'll miss Wine things.

```bash
ryanh@orion-o6:/mnt/Work/Work/work/FEXNew/Build_Radxa$ FEXpidof SonicMania.exe
ryanh@orion-o6:/mnt/Work/Work/work/FEXNew/Build_Radxa$ pidof SonicMania.exe
12373
```

I don't think there would be a good way to actually resolve this now.

Sonicadvance1:
I'll need to rethink how to determine if a process is FEX or not since it's tricky at this point.

wcampbell-nv:
`/proc/pid/exe` should always point to the original exe, maybe you can use that. `readlink /proc/pid/exe` outputs the original path.

Sonicadvance1:
Mostly just need to ensure it works in the three main modalities of FEX which will take a bit of time.
- FEX - Running a linux app
- FEX - Running x86 Wine and extracting the .exe out of it
- Arm64ec/wow64 - Running inside of arm64 Wine and extracting the .exe file out of it.

Sonicadvance1:
There we go. Fixed now and it sees FEX in all the various iterations. `FEXpidof FEX` will now see all FEX instances, including FEX+Linux, FEX+Wine, arm64ec/wow64+FEX. So `kill -9 $(FEXpidof FEX)` is an omega kill command that hits everything.

Sonicadvance1:
Oh, slightly incorrect. FEX+Wine can't find the executable name because it can't determine that the process is wine anymore but eh. Can't really solve that.
It still determines that it is a FEX process so `FEXpidof FEX` at least captures it.

PR #5105, opened 2025-12-06
Syscalls: Fix DEBUG_STRACE printing

gio3k:
Replaces the % strings with {} in syscall TraceFormatString / Def.StraceFmt

Looks something like this on main:
```
D write(%ld, %ld, %ld) = %ld
D write(%ld, %ld, %ld) = %ld
D write(%ld, %ld, %ld) = %ld
```

Looks like this now:
```
D write(9, 140737488343440, 128) = 128
D write(9, 140737488343440, 34) = 34
D write(9, 94078319113395, 1) = 1
D openat(4294967196, 94078319267264, 524288, 0) = 10
D read(10, 140737488340936, 832) = 832
D fstat(10, 7fffffffc660) = 0
```

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
@@ -394,19 +394,19 @@ struct ArgToFmtString;
   };
 
 // Base types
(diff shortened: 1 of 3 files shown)


Sonicadvance1 (approved):
You can definitely tell how long it has been since I've used this.

PR #5110, opened 2025-12-10
ARM64Emitter: Force NOP padding to be enabled

neobrain:
When loading code caches, these constants get patched up for the new guest address. The new value may be larger than the original, so the padding bytes ensure the maximum of 16 bytes of encoding space is always available.

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: MIT
 #pragma once
 
+#include <FEXCore/Config/Config.h>
+
 #ifdef VIXL_DISASSEMBLER
 #include <aarch64/disasm-aarch64.h>
-#include <FEXCore/Config/Config.h>
 #include <FEXCore/fextl/memory.h>
 #include <FEXCore/fextl/vector.h>
 #endif
@@ -272,6 +273,8 @@ protected:
 
   FEX_CONFIG_OPT(Disassemble, DISASSEMBLE);
 #endif
+
+  FEX_CONFIG_OPT(EnableCodeCaching, ENABLECODECACHINGWIP);
 };
 
 } // namespace FEXCore::CPU
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -418,6 +418,11 @@ FEXCore::X86State::X86Reg Arm64Emitter::GetX86RegRelationToARMReg(ARMEmitter::Re
 }
 
 void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, uint64_t Constant, bool NOPPad) {
+  if (EnableCodeCaching) {
+    // Force NOP padding to ensure relocated constants always have enough encoding space available
+    NOPPad = true;
+  }
+
   bool Is64Bit = s == ARMEmitter::Size::i64Bit;
   int Segments = Is64Bit ? 4 : 2;
 


FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
```diff
@@ -108,7 +108,7 @@ class Arm64Emitter : public ARMEmitter::Emitter {
 public:
   Arm64Emitter(FEXCore::Context::ContextImpl* ctx, void* EmissionPtr = nullptr, size_t size = 0);
 
-  void LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, uint64_t Constant, bool NOPPad = false);
+  void LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, uint64_t Constant, bool NOPPad = true);
```
Sonicadvance1: plez to only force NOP padding when code caching is enabled.

Sonicadvance1:
```diff
diff --git a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
index ae6c39d6c..33995edf3 100644
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.cpp
@@ -418,6 +418,9 @@ FEXCore::X86State::X86Reg Arm64Emitter::GetX86RegRelationToARMReg(ARMEmitter::Re
 }

 void Arm64Emitter::LoadConstant(ARMEmitter::Size s, ARMEmitter::Register Reg, uint64_t Constant, bool NOPPad) {
+  if (EnableCodeCachingWIP()) {
+    NOPPad = true;
+  }
   bool Is64Bit = s == ARMEmitter::Size::i64Bit;
   int Segments = Is64Bit ? 4 : 2;

diff --git a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
index 626ed15b6..5b3e92c80 100644
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
@@ -272,6 +272,7 @@ protected:

   FEX_CONFIG_OPT(Disassemble, DISASSEMBLE);
 #endif
+  FEX_CONFIG_OPT(EnableCodeCachingWIP, ENABLECODECACHINGWIP);
 };

 } // namespace FEXCore::CPU
```

bylaws:
I'm not sure this behavior is desirable for actually using the jit at runtime right? Only for offline compilation, so wouldn't really want to check the code caching option here,

bylaws:
Also looks like a lot of the inserted nops in instcountci don't correspond to address constants, those should definitely have pad switched to false also. Only cases where an instruction defined constant could actually be placed should be padded

Sonicadvance1:
> Also looks like a lot of the inserted nops in instcountci don't correspond to address constants, those should definitely have pad switched to false also

I plan on auditing all of our `LoadConstant` usage after this is in so that only things that actually need to be relocated pad out when code caching is enabled. So it being excessive currently is a non-issue, just needs to not force pad unless its actually been enabled currently. NOP isn't free.

bylaws:
> > Also looks like a lot of the inserted nops in instcountci don't correspond to address constants, those should definitely have pad switched to false also
> 
> I plan on auditing all of our `LoadConstant` usage after this is in so that only things that actually need to be relocated pad out when code caching is enabled. So it being excessive currently is a non-issue, just needs to not force pad unless its actually been enabled currently. NOP isn't free.

Hmm when you get to that would be good to see the instcountci diff with it to confirm

neobrain:
I'm guessing you're referring to this comment?

> I'm not sure this behavior is desirable for actually using the jit at runtime right? Only for offline compilation, so wouldn't really want to check the code caching option here,

Not entirely sure what a good way to wire up the interfaces to handle this is, since we *do* want the same behavior at runtime when cache validation is enabled. Since it won't impact operation while `EnableCodeCachingWIP` is still off by default, maybe we can defer this until code validation landed and Ryan's LoadConstant audit is complete?

bylaws:
> I'm guessing you're referring to this comment?
> 
> > I'm not sure this behavior is desirable for actually using the jit at runtime right? Only for offline compilation, so wouldn't really want to check the code caching option here,
> 
> Not entirely sure what a good way to wire up the interfaces to handle this is, since we _do_ want the same behavior at runtime when cache validation is enabled. Since it won't impact operation while `EnableCodeCachingWIP` is still off by default, maybe we can defer this until code validation landed and Ryan's LoadConstant audit is complete?

I was thinking IsGeneratingCache could just be checked - would it be a problem to set that when validating

neobrain:
> I was thinking IsGeneratingCache could just be checked - would it be a problem to set that when validating

CodeCache logically sits on top of Arm64JitCore, so it should pass state down into the JIT instead of the JIT peeking up the stack. But further the validation run doesn't even set `CodeCache.IsGeneratingCache` (it would be rather weird for a reference compile run to set it when it neither uses a code cache nor generates one!)

For what it's worth, I'll submit the cache validation PR  later today / early tomorrow. Would still be interesting to see where exactly we end up needing NOP-padded LoadConstant calls, but I'm guessing it won't change how we generally need to pass around state.

PR #5111, opened 2025-12-10
CodeCache: Implement cache loading

neobrain:
Building on the cache generation feature added in #5096, this PR allows FEX to load already generated caches.

Some constraints apply that will be resolved in later PRs:
* Caches must be generated manually
* Caches for the main executable and for `ld.so` aren't yet loaded (only for its libraries)
* Caches aren't aware of FEX configuration yet

To test this feature, enable the `EnableCodeCachingWIP` setting and follow these steps:
1. Run some application (e.g. SuperMeatBoy) and close it
2. Open a shell and verify that FEX created a file like ~/.cache/fex-emu/codemap/SuperMeatBoy-038d4ceff5e2271a.0.bin
3. Run `FEXOfflineCompiler generate ~/.cache/fex-emu/codemap/SuperMeatBoy-038d4ceff5e2271a.0.bin --fileid <some library file id>`
4. Run the application again, observe FEXServer logs. If all worked well, you should see a message like `Cache load:  1584 blocks; base=0x7fffea5d8000; off=  0x19000-0x0057000; 12a95018966b8c93 libGLX_mesa.so.0.0.0`

Testing requires #5110.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
@@ -217,6 +217,32 @@ static fextl::vector<Elf64_Phdr> ReadELFHeaders(int FD, std::span<std::byte> Hea
   return std::move(Parser.phdrs);
 }
 
+static void LoadCodeCache(FEXCore::Core::InternalThreadState& Thread, FEXCore::ExecutableFileSectionInfo& Section, uint64_t CodeCacheConfigId) {
+  auto CacheFilename = fextl::fmt::format("{}cache/{}-{:016x}", FEX::Config::GetCacheDirectory(),
+                                          FEXCore::CodeMap::GetBaseFilename(Section.FileInfo, false), CodeCacheConfigId);
+  int CacheFD = open(CacheFilename.c_str(), O_RDONLY);
+  if (CacheFD == -1) {
+    LogMan::Msg::IFmt("Cache file does not exist: {}", CacheFilename);
+    return;
+  }
+
+  struct stat buf;
+  if (fstat(CacheFD, &buf) != 0) {
+    LogMan::Msg::EFmt("Invalid cache file: {}", CacheFilename);
+    close(CacheFD);
+    return;
+  }
+
+  auto CacheFileSize = buf.st_size;
+  auto MappedCache = (std::byte*)FEXCore::Allocator::mmap(nullptr, CacheFileSize, PROT_READ, MAP_PRIVATE, CacheFD, 0);
+  LOGMAN_THROW_A_FMT(MappedCache, "Failed to map code cache into memory");
+  if (!Thread.CTX->GetCodeCache().LoadData(Thread, MappedCache, Section)) {
+    // TODO: Delete this cache file
+  }
+  FEXCore::Allocator::munmap(MappedCache, CacheFileSize);
+  close(CacheFD);
+}
+
 void* SyscallHandler::GuestMmap(bool Is64Bit, FEXCore::Core::InternalThreadState* Thread, void* addr, size_t length, int prot, int flags,
                                 int fd, off_t offset) {
   LOGMAN_THROW_A_FMT(Is64Bit || (length >> 32) == 0, "values must fit to 32 bits");
@@ -258,22 +284,7 @@ void* SyscallHandler::GuestMmap(bool Is64Bit, FEXCore::Core::InternalThreadState
   }
 
   if (EnableCodeCaching && CachedSection) {
-    auto CacheFilename = fextl::fmt::format("{}cache/{}-{:016x}", FEX::Config::GetCacheDirectory(),
-                                            FEXCore::CodeMap::GetBaseFilename(CachedSection->FileInfo, false), CodeCacheConfigId);
-    int FD = open(CacheFilename.c_str(), O_RDONLY);
-    struct stat buf;
-    if (FD != -1 && (fstat(FD, &buf) == 0)) {
-      auto CacheFileSize = buf.st_size;
-      auto MappedCache = (std::byte*)FEXCore::Allocator::mmap(nullptr, CacheFileSize, PROT_READ, MAP_PRIVATE, FD, 0);
-      LOGMAN_THROW_A_FMT(MappedCache, "Failed to map code cache into memory");
-      if (!CTX->GetCodeCache().LoadData(*Thread, MappedCache, *CachedSection)) {
-        // TODO: Delete this cache file
-      }
-      FEXCore::Allocator::munmap(MappedCache, CacheFileSize);
-    }
-    if (FD != -1) {
-      close(FD);
-    }
+    LoadCodeCache(*Thread, *CachedSection, CodeCacheConfigId);
   }
 
   return reinterpret_cast<void*>(Result);


Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -250,6 +257,24 @@ void* SyscallHandler::GuestMmap(bool Is64Bit, FEXCore::Core::InternalThreadState
...
 
+  if (EnableCodeCaching && CachedSection) {
+    auto CacheFilename = fextl::fmt::format("{}cache/{}-{:016x}", FEX::Config::GetCacheDirectory(),
+                                            FEXCore::CodeMap::GetBaseFilename(CachedSection->FileInfo, false), CodeCacheConfigId);
+    int FD = open(CacheFilename.c_str(), O_RDONLY);
+    struct stat buf;
+    if (FD != -1 && (fstat(FD, &buf) == 0)) {
+      auto CacheFileSize = buf.st_size;
+      auto MappedCache = (std::byte*)FEXCore::Allocator::mmap(nullptr, CacheFileSize, PROT_READ, MAP_PRIVATE, FD, 0);
+      close(FD);
```
Sonicadvance1: FD leak here if open succeeds but fstat does not.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -250,6 +257,24 @@ void* SyscallHandler::GuestMmap(bool Is64Bit, FEXCore::Core::InternalThreadState
...
   }
 
+  if (EnableCodeCaching && CachedSection) {
+    auto CacheFilename = fextl::fmt::format("{}cache/{}-{:016x}", FEX::Config::GetCacheDirectory(),
+                                            FEXCore::CodeMap::GetBaseFilename(CachedSection->FileInfo, false), CodeCacheConfigId);
+    int FD = open(CacheFilename.c_str(), O_RDONLY);
+    struct stat buf;
+    if (FD != -1 && (fstat(FD, &buf) == 0)) {
+      auto CacheFileSize = buf.st_size;
+      auto MappedCache = (std::byte*)FEXCore::Allocator::mmap(nullptr, CacheFileSize, PROT_READ, MAP_PRIVATE, FD, 0);
```
Sonicadvance1: Should at least throw an LOGMAN assert here if the allocation fails.

Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
neobrain: Had fixed this in a couple of places for later patches but missed this one apparently, thanks!

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+  if (header.NumBlocks == 0) {
+    // Valid caches are never empty
+    LogMan::Msg::IFmt("Code cache empty, aborting");
+    return false;
+  }
+
+  char ExpectedVersion[8] = GIT_SHORT_HASH;
+  std::ranges::fill(std::ranges::find(ExpectedVersion, 0), std::end(ExpectedVersion), 0);
+  if (!std::ranges::equal(header.FEXVersion, ExpectedVersion)) {
+    LogMan::Msg::IFmt("Cache generated from old FEX version {}, current is {}; skipping\n", fmt::join(header.FEXVersion, ""),
```
bylaws: Do we want to verify format version etc before this/performing any other operations?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+    auto MaxBound = CTX.SyscallHandler->LookupExecutableFileSection(Thread, max_val->first + BinarySection.FileStartVA);
+    if (&MinBound->FileInfo != &BinarySection.FileInfo || &MaxBound->FileInfo != &BinarySection.FileInfo) {
+      ERROR_AND_DIE_FMT("Cached blocks offsets {:#x}-{:#x} out of bounds for guest library {} ({:016x} @ {:#x}) while trying to load "
+                        "section {:#x}-{:#x}!",
+                        min_val->first, max_val->first, BinarySection.FileInfo.Filename, BinarySection.FileInfo.FileId,
+                        BinarySection.FileStartVA, BinarySection.BeginVA, BinarySection.EndVA);
+    }
+
+    // Constrain BlockList to the given ExecutableFileSectionInfo
+    namespace ranges = std::ranges;
```
bylaws: may as well keep this at the top and use throughout

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+    LOGMAN_THROW_A_FMT(ranges::is_sorted(BlockList, [](auto& a, auto& b) { return a.first < b.first; }), "Expected sorted block list");
+    auto begin = ranges::lower_bound(BlockList, BinarySection.BeginVA - BinarySection.FileStartVA, std::less {}, &BlockListEntry::first);
+    auto end =
+      ranges::upper_bound(begin, BlockList.end(), BinarySection.EndVA - BinarySection.FileStartVA - 1, std::less {}, &BlockListEntry::first);
+    BlockList.erase(end, BlockList.end());
```
bylaws, line 397: I was kinda wondering what the point of this pruning is, if we change to reject all caches that cover a greater range than the mapped executable above it seems a simpler way to achieve a similar result. Is this partial mapping some common case on linux somehow?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+  // Read relocations
+  fextl::vector<FEXCore::CPU::Relocation> Relocations(header.NumRelocations, FEXCore::CPU::Relocation::Default());
+  ::memcpy(Relocations.data(), MappedCacheFile, Relocations.size() * sizeof(Relocations[0]));
+  MappedCacheFile += Relocations.size() * sizeof(Relocations[0]);
+
+  // Pad to next page in file, which contains CodeBuffer data
+  MappedCacheFile = reinterpret_cast<std::byte*>(AlignUp(reinterpret_cast<uintptr_t>(MappedCacheFile), Utils::FEX_PAGE_SIZE));
+
+  // Prepare CodeBuffer: Page aligned and big enough to hold all cached data
+  auto Lock = std::unique_lock {CTX.CodeBufferWriteMutex};
```
bylaws, line 415: We need to support a null thread here as discussed.  This should grab a codebuffer from just context and create new one if needed, any threads will switch to it when they compile code etc

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+    Thread.LookupCache->ChangeGuestToHostMapping(*Prev, *CTX.GetLatest()->LookupCache, lk);
+  }
+
+  auto CodeBuffer = CTX.GetLatest();
+  LOGMAN_THROW_A_FMT(header.CodeBufferSize <= CodeBuffer->Size, "CodeBuffer too small to load code cache");
+  LOGMAN_THROW_A_FMT(reinterpret_cast<uintptr_t>(CodeBuffer->Ptr) % 0x1000 == 0, "Expected CodeBuffer base to be page-aligned");
+  const auto Delta = AlignUp(CTX.LatestOffset, 0x1000) - CTX.LatestOffset;
+  CTX.LatestOffset += Delta;
+
+  while (CTX.LatestOffset + header.CodeBufferSize > CodeBuffer->Size - Utils::FEX_PAGE_SIZE) {
```
bylaws, line 428: This is a temporary hack until we support mmapping additional codebuffers right?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+  MappedCacheFile += header.CodeBufferSize;
+  CTX.LatestOffset += header.CodeBufferSize;
+
+  // Apply FEX relocations
+  auto Ret = ApplyCodeRelocations(BinarySection.FileStartVA, CodeBufferRange, Relocations, false);
+  LOGMAN_THROW_A_FMT(Ret == true, "Failed to apply code cache relocations");
+
+  // Register blocks to LookupCache
+  {
+    auto& LookupCache = *CodeBuffer->LookupCache;
```
bylaws, line 446: Need to acq lock here?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+    for (auto& [Guest, Host] : BlockList) {
+      LookupCache.BlockList[Guest + BinarySection.FileStartVA].HostCode = Host.HostCode + reinterpret_cast<uintptr_t>(CodeBufferRange.data());
+
+      for (auto& CodePage : Host.CodePages) {
+        CodePage += BinarySection.FileStartVA;
+      }
+      LookupCache.BlockList[Guest + BinarySection.FileStartVA].CodePages = std::move(Host.CodePages);
```
bylaws: AddBlockMapping? might want slight api addition but better than open-coding

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +323,156 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+    memcpy(&NumEntrypoints, MappedCacheFile, sizeof(NumEntrypoints));
+    MappedCacheFile += sizeof(NumEntrypoints);
+
+    Entrypoints.resize(NumEntrypoints);
+    memcpy(Entrypoints.data(), MappedCacheFile, NumEntrypoints * sizeof(Entrypoints[0]));
+    MappedCacheFile += NumEntrypoints * sizeof(Entrypoints[0]);
+
+    EntrypointsSet.clear();
+    EntrypointsSet.insert(Entrypoints.begin(), Entrypoints.end());
+    if (Thread.LookupCache->AddBlockExecutableRange(&Thread, EntrypointsSet, CodePage, FEXCore::Utils::FEX_PAGE_SIZE)) {
```
bylaws: Can switch this to use iterators to avoid the intermediary set

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Resizing the code buffer to accommodate the memory requirements isn't a hack, but indeed this won't be needed with persistent cache mappings.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: The CodeBufferWriteLock should take care of it, shouldn't it?

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: What's the suggestion here?

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: I'm meaning in terms of the way it's done, like a minimum size parameter to getlatestcodebuffer would be a lot cleaner if this is gonna stay around

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: I think I'd prefer deferring this since type-erasing over iterators is annoying to implement, and I'm not sure it makes a significant dent in performance given we're in IO code.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Ah yeah nw, I mentally indented the lock block

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: To use LookupCache.AddBlockMapping()/extend it work here instead

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: It's already at the top - what other operations are there?

The log line should stay above since it provides context for this message.

neobrain: Oh, regarding format version - there isn't any check on this currently since it's implied by the FEX version check. The format version can only change if FEX itself changed, so the field is only really useful for any potential external tooling.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Would have to re-check where I encountered this, but it's easy enough to support either way.

Not sure what the logic behind rejecting all caches that cover a greater range would be. Either it's a valid cache that we should just extract all relevant data from, or we've somehow entered an invalid state that shouldn't just be skipped over.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: >  This should grab a codebuffer from just context and create new one if needed, any threads will switch to it when they compile code etc

Note that much of the CodeBuffer stuff here (along with the Thread use) goes away with persistently mapped caches, since we'll only be touching the LookupCache then. WoA can just ifdef this out for the time being.

For the CheckCodeBufferUpdate call specifically: Was not having a thread a condition that only happens during startup? In that case no code buffer update would be needed either way, so the update can be guarded behind an if check.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Right - NumBlocks is read before checking this, I think it makes sense to be defensive and put hash check before that. Or at least a magic check

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: I was thinking more on the lines that if we're already in an invalid state, there's no harm being cautious and giving up. At least if it's code that should never really be exercised outside of such invalid cases, if this is common due to some reason then nevermind.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Sure, maybe just move the vector decl out of the loop then and fine

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: I think we might be talking about two slightly different things:
1. The range given to LoadData is smaller than the binary size (but covers the entire code cache contents)
2. The code cache is larger than the range given to LoadData (e.g. when there's multiple executable sections)

My previous comment targeted point 2, but it sounds like you're wondering about point 1. Such partial loads are actually the default case outside of WoA, since FEX hooks into individual mmap calls rather than some magic `LoadBinary` function.

(Point 1 doesn't justify having this code, it's really about point 2.)

bylaws: Right, makes sense, ty

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Yeah, that would avoid needing to use a pesky loop instead of a plain conditional, but not sure it's bad enough to warrant changing unrelated interfaces since it's only temporary anyway.

(CodeBufferManager::GetLatest in particular can't do this since it's a base class of Context, not Context itself, so it doesn't have access to all the state used here)

bylaws: Yeah exactly:)

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: It's already outside :)

Will move the `fextl::set` back in though, since I don't think that container can benefit from reusing allocations.

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: Yep

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Ah yes. Will adopt this order: Magic > Version > NumBlocks. Thanks!

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Ah yes, should apply, cheers.

bylaws:
Was thinking this through - do we want to differentiate a dirty FEX git tree when loading caches?

FEXCore/Source/Interface/Core/CodeCache.cpp
bylaws: > I think I'd prefer deferring this since type-erasing over iterators is annoying to implement, and I'm not sure it makes a significant dent in performance given we're in IO code.

I was looking at this and a little confused, can one literally not just template both funcs?

bylaws: Also - this should use CodeBuffer LookupCache->AddBlockExecutableRange rather than going through thread

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +325,162 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
```
bylaws: Either needs to be CodePage << 12 or the on-disk format needs to shift (I'm a fan of latter)

bylaws (changes requested):

neobrain:
> Was thinking this through - do we want to differentiate a dirty FEX git tree when loading caches?

I can see the appeal, though personally it would be inconvenient to have caches cleared every time I touch any file. It also seems more reasonable to expect people making changes to codegen to be aware of caches, since we can't differentiate between "dirty" and "dirtier" trees anyway (so any potential issue would just be delayed, not avoided).

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: > I was looking at this and a little confused, can one literally not just template both funcs?
> Also - this should use CodeBuffer LookupCache->AddBlockExecutableRange rather than going through thread

Putting these two together it's indeed that simple :)

Would've preferred a non-template solution, but since it's just needed for one of them it's probably fine. The function signature has grown rather incomprehensible these past months so want to not make it any worse by adding `template` everywhere.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Done

neobrain:
@bylaws I think all comments are covered now, was there anything else to add?

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -329,6 +325,161 @@ bool CodeCache::SaveData(Core::InternalThreadState& Thread, int fd, const Execut
...
+      for (auto& CodePage : Host.CodePages) {
+        CodePage += BinarySection.FileStartVA;
+      }
+      auto HostCode = reinterpret_cast<void*>(Host.HostCode + reinterpret_cast<uintptr_t>(CodeBufferRange.data()));
+      LookupCache.AddBlockMapping(Guest + BinarySection.FileStartVA, std::move(Host.CodePages), HostCode, WriteLock);
+    }
+
+    // Register loaded code ranges
+    fextl::vector<uint64_t> Entrypoints;
+    fextl::set<uint64_t> EntrypointsSet;
```
bylaws: Dead var

bylaws (approved):
Just this and looks all good

neobrain:
Great, thanks for the review :)

PR #5119, opened 2025-12-15
BranchOps: Use RIP relocs for direct branch targets

bylaws:


bylaws:
> Just double-checking, did the `LoadConstant(ARMEmitter::Size::i64Bit, TMP1, NewRIP);` below also need a relocation or is that one of the things we take care of post-compile in JIT.cpp?

oh that it does.

Sonicadvance1 (approved):

PR #5120, opened 2025-12-15
Add Regex Engine to FEX for config option loading

badumbatish:
Hi there, i'm opening the PR to add a small regex engine to the Fex codebase. The motivation is to help dynamic modify some config option based on the name of the config file. Please see the following screenshot for more context.
Screenshot 2025-12-15 at 7 25 58 AM.

The regex engine is an NFA based one, implemented from the dragon book 2nd ed.

Implementation is in `FEXCore/Source/Utils/Regex.cpp`.
Test case is `unittests/APITests/Regex.cpp`.

I took the liberty to refactor and reuse a bit of the code in parsing the json. The traversing of a json list of pair of strings is now a helper function called ListApplier that accepts a `json_t`.

The default behavior is that the config loader goes through each regex and see if it matches the config name, if it does, it applies the config option specified and then stops checking anymore regex (this can change based on the need and reviews of the PR).

I'm not sure where should i add the test case for the new config loader. I tested locally via `/home/ubuntu/.fex-emu/Config.json` and it works, would love some suggestions
```
{
  "Config":
    {"RootFS":"Ubuntu_22_04"},
  "ThunksDB":{},
  "RegexConfig" : {
    ".*" : {
      "RootFS":"Ubuntu_24_04",
    }
  }
}
```


bylaws:
I'm unsure on a full regex engine, I think a very naive glob impl might be enough. albeit this is very neat

FEXCore/include/FEXCore/Utils/Regex.h
```diff
@@ -0,0 +1,75 @@
+#pragma once
```
Sonicadvance1: I have a concern about including this regex_engine that has no license. When it says "Inspiration taken from ..." and it's pretty much the code from https://github.com/0xsh4dy/regex_engine .
Attribution doesn't really matter when no license exists.

Sonicadvance1 (changes requested):
Other than the concern about the licensing of the regex engine used, I'm fine with this type of change.

FEXCore/include/FEXCore/Utils/Regex.h
badumbatish: i've contacted the repo author and he's added an MIT license to it, what other changes should i add to the PR for this to go through?

Sonicadvance1: Make sure to add the SPDX identifier to both Regex.h and Regex.cpp
```cpp
/*
 * Copyright (c) 2025 Rakshit Awasthi
 * SPDX-License-Identifier: MIT
 */
```

In both the files should cover it.

FEXCore/include/FEXCore/Utils/Regex.h
```diff
@@ -0,0 +1,79 @@
...
+  NFA parseConcatenation();
+
+  // INFO: "a*", ".*"
+  NFA parseStarPlusHuhhhh();
+
+  // INFO: "(abc)" or a
+  NFA parseAtom();
+
+public:
+  static inline fextl::string Alphabet = "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ,./<>?;':\"[]\\{}|1234567890!@#$%^&*()-=_+";
```
neobrain: Does this break detection for patterns that contain non-ASCII characters?

FEXCore/Source/Utils/Regex.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+  if (Pos >= Pattern.size()) {
+    return NFA::createForEpsilon();
+  }
+  char curChar = Pattern[Pos++];
+
+  if (Escaped) {
+    Escaped = false;
+    if (AcceptableEscapable.find(curChar) == fextl::string::npos) {
+      fprintf(stderr, "Expected an acceptable escapable character which "
```
neobrain: Sounds like this should use one of our assertion macros, probably either `LOGMAN_THROW_*` or `ERROR_AND_DIE_FMT`. Since this is user-controlled input, there needs to be a message that indicates what the source of the problem is (to a sufficiently well-educated user at least).

FEXCore/include/FEXCore/Utils/Regex.h
badumbatish: i suspect so this is a naive alphabet I added just enough for test cases to pass. Are we expecting any non ascii in the json file name?

FEXCore/Source/Utils/Regex.cpp
badumbatish: yep will do

badumbatish: hmm I removed this feature from the regex engine so there isn't a need for `LOGMAN_THROW_` so it seems we won't be needing it here

badumbatish:
I've gone ahead and removed some features like unions, escapables, question, plus. Please let me know if this is sufficient for the PR

FEXCore/Source/Utils/Regex.cpp
```diff
@@ -0,0 +1,190 @@
...
+fextl::set<State *> NFA::move(const fextl::set<State *> &states, const char c) {
+  fextl::set<State *> result;
+  for (auto *state : states) {
+    const decltype(state->transitions) &transitionMap = state->transitions;
+    if (auto itr = transitionMap.find(c); itr != transitionMap.end()) {
+      for (auto *transition : itr->second) {
+        result.insert(transition);
+      }
+    }
+  }
```
neobrain: Given that we only need simple wildcard support (`*` without a dot), can't we just add any character that's not `*` here without keeping track of an alphabet or a `transitions` map?

I think that way we can drop the `transitions` vector and the limited alphabet entirely. I'm not fully grokking the NFA stuff, but intuitively it still seems overly complex for a simple wildcard matcher.

FEXCore/include/FEXCore/Utils/Regex.h
neobrain: Application authors choose the filenames so it could be anything, though not sure we'd realistically use wildcard matching for any such executable names.

Can we just treat any character that's not `*` as part of the "alphabet"? (see also my other comment). That would elegantly avoid this surprising restriction.

FEXCore/Source/Utils/Regex.cpp
badumbatish: hmm now that im thinking about it, let's drop the NFA and the transitions vectors if all we're doing is simple wildcard, i'll rewrite the regex engine and test the new engine on the existing test case

unittests/APITests/Regex.cpp
```diff
@@ -0,0 +1,33 @@
+#include "FEXCore/fextl/string.h"
```
neobrain: Could you add these?
* Empty wildcard at the beginning/end:
```cpp
  CHECK(Regex::matches("a*", "a"));
  CHECK(Regex::matches("*a", "a"));
  CHECK(Regex::matches("*a*", "a"));
```
* wildcard in the middle:
```cpp
  CHECK(Regex::matches("test*pattern", "test__pattern"));
```
* multiple wildcards:
```cpp
  CHECK(Regex::matches("test*pattern*more", "test__pattern__more"));
  CHECK(Regex::matches("test**pattern", "test_pattern"));
```
* Also might be worth adding a test for something like `Regex::matches("a*", "a*")`. Not exactly relevant to the use case, but should be covered by the helper implementation.

unittests/APITests/Regex.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+TEST_CASE("Singular regex") {
+  CHECK(Regex::matches("a", "a") == true);
+  CHECK(Regex::matches("a*", "aaaaaaa") == true);
+}
+
+TEST_CASE("Concat regex") {
+  CHECK(Regex::matches("aaa", "aaa") == true);
+  CHECK(Regex::matches("ab", "ab") == true);
+  CHECK(Regex::matches("a", "ab") == false);
```
neobrain: This is easier to parse since it doesn't require jumping to the end of the line to see what you're testing for:
```suggestion
  CHECK(Regex::matches("ab", "ab"));
  CHECK(!Regex::matches("a", "ab"));
```

unittests/APITests/CMakeLists.txt
```diff
@@ -6,7 +6,9 @@ set(TESTS
   FileMappingBaseAddress
   Filesystem
   InterruptableConditionVariable
-  StringUtils)
+  StringUtils
+  Regex
```
neobrain: It's no longer a Regex engine, so rename to something like WildcardMatch(er)?

Source/Common/Config.cpp
```diff
@@ -43,22 +43,59 @@ namespace JSON {
...
-      if (!ConfigString) {
-        LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ConfigName);
+      if (!RegexNamedList) {
+        LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, RegexName);
         return;
       }
 
-      Func(ConfigName, ConfigString);
+      // Matches the first and then get out
+      // Needs PR review on this
```
neobrain: What do you need feedback on?

Source/Common/Config.cpp
```diff
@@ -43,22 +43,59 @@ namespace JSON {
...
+          return;
+        }
+        Func(ConfigName, ConfigString);
+      }
+
+    };
+
+    ListApplier(ConfigList);
+
+    const json_t* RegexList = json_getProperty(json, "RegexConfig");
```
neobrain: Can we rename this property to `AppOverrides`? It can be used without wildcards after all.

Source/Common/Config.cpp
```diff
@@ -43,22 +43,59 @@ namespace JSON {
...
+    if (!RegexList) {
+      // This is a non-error if the configuration file exists but no RegexConfigList section
+      return;
+    }
+
+    using FEXCore::Utils::Regex;
```
neobrain: This is only used once IIUC, so probably better to just fully qualify the name below.
```suggestion
```

Source/Common/Config.cpp
```diff
@@ -43,22 +43,59 @@ namespace JSON {
       return;
     }
 
-    for (const json_t* ConfigItem = json_getChild(ConfigList); ConfigItem != nullptr; ConfigItem = json_getSibling(ConfigItem)) {
-      const char* ConfigName = json_getName(ConfigItem);
-      const char* ConfigString = json_getValue(ConfigItem);
+    auto ListApplier = [&Config, &Func](const json_t* jsonList) {
+      for (const json_t* ConfigItem = json_getChild(jsonList); ConfigItem != nullptr; ConfigItem = json_getSibling(ConfigItem)) {
```
neobrain: Inline lambdas are hard to make sense of as a reader since they require a lot of jumping around to get a basic grasp of what a function does.

In this case, one way to improve the structure is to first check the `RegexList` and collect all matching `json_t`s into an `fextl::vector`. Then, we can turn the existing loop into something like this:

```suggestion
    for (auto ConfigBlock : ConfigBlocks) {
      for (const json_t* ConfigItem = json_getChild(ConfigBlock); ConfigItem != nullptr; ConfigItem = json_getSibling(ConfigItem)) {
```

FEXCore/include/FEXCore/Utils/Regex.h
```diff
@@ -0,0 +1,11 @@
...
+
+#pragma once
+#include <string_view>
+
+namespace FEXCore::Utils {
+class Regex {
+public:
+  static bool matches(std::string_view pattern, std::string_view text);
+};
+} // namespace FEXCore::Utils
```
neobrain: No need to wrap this in a class:

```suggestion
namespace FEXCore::Utils::Wildcard {
bool Matches(std::string_view pattern, std::string_view text);
}
```
(Also note `Matches` rather than `matches`, which is how existing functions are named)

neobrain:
Oh, you'll also want to run clang-format-19 on any newly added code. (I think most of us have our IDEs set up to auto-format on save.)

badumbatish:
> Oh, you'll also want to run clang-format-19 on any newly added code. (I think most of us have our IDEs set up to auto-format on save.)

gotacha i'll install it

badumbatish:
Apologies for the delay i was on vacation. I've addressed the comments

Source/Common/Config.cpp
badumbatish: I mainly would want to know if the current approach of applying the config is good or not

neobrain:
> Apologies for the delay i was on vacation. I've addressed the comments

No worries, though you arrived just in time for *my* vacation and sadly no one else was around to take over the review.

Not exactly sure what's going on with the CI build failures - could you rebase the PR onto latest main to see if it still happens then?

Sonicadvance1:
The CI build failures come from pulling in headers from a FEXCore library but the test isn't linking to that library.

neobrain:
Development continued in https://github.com/FEX-Emu/FEX/pull/5388.

PR #5123, opened 2025-12-15
Guest relocation support

bylaws:
Implements guest relocations by decoding relocated immediates as block entrypoint relative offsets and then handling such cases with _EntrypointOffset in OpcodeDispatcher. This currently supports all cases observed in my tests on 32-bit/64-bit Windows libraries.

CC: @neobrain

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -655,7 +655,8 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
               LogMan::Msg::EFmt("Invalid or Unknown instruction: {} 0x{:x}", TableInfo->Name ?: "UND", Block.Entry - GuestRIP);
             }
 
-            if (Block.BlockStatus == Frontend::Decoder::DecodedBlockStatus::INVALID_INST) {
+            if (Block.BlockStatus == Frontend::Decoder::DecodedBlockStatus::INVALID_INST ||
+                Block.BlockStatus == Frontend::Decoder::DecodedBlockStatus::BAD_RELOCATION) {
               Thread->OpDispatcher->InvalidOp(DecodedInfo);
             } else {
               Thread->OpDispatcher->NoExecOp(DecodedInfo);
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -162,7 +162,8 @@ std::pair<uint64_t, bool> Decoder::ReadData(uint8_t Size) {
       } else if (It->second == GuestRelocationType::Rel64 && Size == 8) {
         return {static_cast<int64_t>(Res) - static_cast<int64_t>(EntryPoint), true};
       } else {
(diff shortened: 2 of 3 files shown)


FEXCore/Source/Interface/Core/X86Tables/X86Tables.h
```diff
@@ -159,16 +176,13 @@ struct DecodedOperand {
     } GPR;
 
     struct {
-      int32_t Displacement;
+      int64_t Displacement;
```
Sonicadvance1, line 179: Does it make sense for any of these to be `int64_t`? Nothing in x86 can encode a 64-bit RIP literal. Any time `DecodedOperand` becomes larger, this causes our decoding size to grow exponentially since /every/ argument eats the cost in the decoded x86 instruction.

bylaws: I was thinking that for int32_t to work one has to rely on 32 bit wrap around but if you had a 64 bit module using 32 bit relocs that just confined itself to the lower 4gb of memory via flags then wraparound wouldn't be applied in such cases and int32_t would be unable to represent the full rang

Sonicadvance1: Sure, you would need to sign extend the incoming `int32_t` but that's what `DecodeAddress` and the {V,}SIB handling stuff already does. We aren't ever decoding an x86 instruction with 64-bit RIP literals in it, nor are we calculating the address and then sticking it back in to `DecodedOperand`.

Once the address is decoding in the OpcodeDispatcher, it either gets handled directly like VSIB, or stuffed in to the `AddressMode` struct.

bylaws: I was meaning there's physically not enough range, if the offset were > 2gb backwards

Sonicadvance1: There's also not physically enough range in the instruction encoding itself? So if a guest relocation was trying to relocate farther than 2GB away, it couldn't ever fit to begin with. It would need to do a `mov <reg>, <64-bit constant>; add <addr_reg>, <reg>;`

bylaws: The instruction encoding can cover any arbitrary 32 bit constant, they are absolute not relative. The full range for relocation can only be achieved with a 32 bit constant if we assume wraparound when adding to entrypoint offset, but if a 32 bit relocation were used on x64 and binary set to load only in low memory then no wraparound is applied and the full range is not reachable

neobrain:
Looking really forward to playing with this!

Just a note though since I see you've been doing this a lot recently, please let's stick to not throwing a bunch of random characters into branch names. It makes navigating git logs a real pain when half of it is just unintelligible garbage.

bylaws:
I hadn't realized branch names were observable once merged into main... Yeah apologies then lmao

bylaws:
Extended with fixes for deferring invalid relocation handling

neobrain:
@Sonicadvance1 Would've preferred some effort to coordinate merging this since I had already said I was going to take a look. I had reported on Discord that as-is this breaks reliable code caching on Linux, didn't think it was necessary to repeat here.

PR #5127, opened 2025-12-16
CodeCache: Implement automatic cache generation

neobrain:
## Overview

While #5111 added support for loading code caches generated by FEXOfflineCompiler (#5096), this process still needed to be performance manually by the user. This PR turns FEXServer into a cache orchestrator to automate this process:

* The `~/.cache/fex-emu/codemap` directory is automatically monitored for new code maps; any incoming data is grouped by binary and then aggregated into a single code map (for each binary) ready for processing
* When new code map data is available, FEXOfflineCompiler is invoked to produce the cache for the corresponding binary

This process is triggered once every FEX application startup using the interface added in #5113.

### Code map types

Since code maps can be in one of many states now, it's good to establish common terminology:

* `ready`: A single code map per executable/library, consisting of ordered data aggregated from one or multiple code maps dumped by FEX
* `new`: A raw code map dump produced by FEX with unsorted data, which can be in either of two states:
  * `pending`: Code map is still opened by FEX and hence not ready for aggregation
  * `incoming`: FEX is done writing this code map so it can used by FEXServer

Functionally, `ready` code maps are placed in `~/.cache/fex-emu/codemap/ready`, whereas `new` ones are placed in `~/.cache/fex-emu/codemap/new`. `pending` code maps are `flock()`ed by FEX, whereas `incoming` ones are not.

### Filename management

Filenames are chosen carefully to prevent various subtle issues:

* Code caches are first written to a file ending in `.new` and then renamed to their proper filename to avoid tearing issues (i.e. FEX opening a stale cache that is being overwritten by FEXOfflineCompiler).

* Code maps are suffixed with a counter so that multiple FEX processes can dump them in parallel. Using a counter instead of a randomized ID allows FEXServer to quickly enumerate all incoming code maps without walking the entire codemap directory.

### Exclusive file locks

FEXServer shouldn't attempt to read pending code maps, since FEX may append new data at them at any time (and conversely FEXServer shouldn't delete a pending code map underneath a FEX process that's still writing to it). To avoid this, `flock()`-based file locking is used. FEXServer will then process code maps only if their owning FEX process has already exited (or crashed).

There is one important race condition that could occur when FEX just created the code map but hasn't acquired the exclusive lock over it, yet. To avoid FEXServer stealing ownership in this short period, FEXServer will avoid processing empty code maps. This implies that FEX may never exit with an empty code map (which is a design constraint that was followed in previous PRs).

An alternative could be to mark pending code maps as part of a per-client data storage in FEXServer. This idea didn't occur to me until I implemented a per-client data storage for other purposes, so I haven't investigated it any further as the `flock()`-based approach is already working.

### Cache regeneration triggers

Some typical scenarios that trigger cache generations are:
* No cache exists yet
* No cache exists yet... for the current FEX configuration
* A cache exists, but new code map data has become available
* A cache exists and we didn't find any new code maps, but it's older than the corresponding code map (which can happen when flipping between two FEX configurations)

Proper support for FEX configuration handling will be implemented later, but for now simple checks based on file existence and timestamps are used to reliably trigger regeneration in all other scenarios.

We will also want to avoid cache generation when there already is a large cache and only a few new blocks have become available, which will require further work.

### DIFF
--- a/Source/Tools/FEXServer/ProcessPipe.cpp
+++ b/Source/Tools/FEXServer/ProcessPipe.cpp
@@ -385,7 +385,7 @@ static void WriteNewCodeMap(const FEXCore::ExecutableFileInfo& File, const std::
   }
 
   for (auto& Block : Blocks) {
-    OutputCodeMap.AppendBlock(FEXCore::ExecutableFileSectionInfo {const_cast<FEXCore::ExecutableFileInfo&>(File), 0}, Block);
+    OutputCodeMap.AppendBlock(FEXCore::ExecutableFileSectionInfo {File, 0}, Block);
   }
 }
 
--- a/FEXCore/Source/Interface/Core/Frontend.h
+++ b/FEXCore/Source/Interface/Core/Frontend.h
@@ -145,7 +145,7 @@ private:
   fextl::set<uint64_t> VisitedBlocks;
   fextl::set<uint64_t>* ExternalBranches {nullptr};
 
-  fextl::unordered_map<uint32_t, GuestRelocationType>* Relocations {nullptr};
+  const fextl::unordered_map<uint32_t, GuestRelocationType>* Relocations {nullptr};
 
   // ModRM rm decoding
   using DecodeModRMPtr = void (FEXCore::Frontend::Decoder::*)(X86Tables::DecodedOperand* Operand, X86Tables::ModRMDecoded ModRM);
--- a/FEXCore/include/FEXCore/Core/CodeCache.h
+++ b/FEXCore/include/FEXCore/Core/CodeCache.h
@@ -39,7 +39,9 @@ struct ExecutableFileInfo {
   ExecutableFileInfo() = default;
 #endif
 
-  fextl::unique_ptr<HLE::SourcecodeMap> SourcecodeMap;
+  // This legacy field must be assignable through const-references
+  mutable fextl::unique_ptr<HLE::SourcecodeMap> SourcecodeMap;
+
   uint64_t FileId = 0;
   fextl::string Filename;
   fextl::unordered_map<uint32_t, GuestRelocationType> Relocations;
@@ -47,7 +49,7 @@ struct ExecutableFileInfo {
 
 // Information associated with a specific section of an executable file
 struct ExecutableFileSectionInfo {
-  ExecutableFileInfo& FileInfo;
+  const ExecutableFileInfo& FileInfo;
 
   // Start address that the file is mapped to.
   // NOTE: Since executable files may be mapped multiple times, this can depend on the queried section.


Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -275,6 +291,187 @@ void SendFDSuccessPacket(fasio::tcp_socket& Socket, int FD) {
...
+        OutputCodeMap.AppendLibraryLoad(Dependency);
+      }
+    }
+  } else {
+    // List only the library itself
+    OutputCodeMap.AppendLibraryLoad(File);
+  }
+
+  for (auto& Block : Blocks) {
+    OutputCodeMap.AppendBlock(FEXCore::ExecutableFileSectionInfo {const_cast<FEXCore::ExecutableFileInfo&>(File), 0}, Block);
```
bylaws: Hmm, can the ExecutableFileInfo reference be const in the section info struct?

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -275,6 +291,187 @@ void SendFDSuccessPacket(fasio::tcp_socket& Socket, int FD) {
...
+  // Accumulate information from new code maps
+  auto IncomingCodeMap = ImportPendingCodeMaps(MainFileId, HasMultiblock);
+  for (auto& [File, _] : IncomingCodeMap) {
+    Result.emplace(std::piecewise_construct, std::forward_as_tuple(nullptr, File.FileId, File.Filename),
+                   std::forward_as_tuple(NeedsCacheRefresh::No));
+  }
+
+  // For each referenced library, add referenced offsets to that library's reference code map
+  for (auto& [File, Blocks] : IncomingCodeMap) {
+    const auto BinaryName = (std::string)FEXCore::CodeMap::GetBaseFilename(File, !HasMultiblock);
```
bylaws:
```suggestion
    const auto BinaryName = std::string {FEXCore::CodeMap::GetBaseFilename(File, !HasMultiblock)};
```

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -275,6 +291,187 @@ void SendFDSuccessPacket(fasio::tcp_socket& Socket, int FD) {
...
+    // List only the library itself
+    OutputCodeMap.AppendLibraryLoad(File);
+  }
+
+  for (auto& Block : Blocks) {
+    OutputCodeMap.AppendBlock(FEXCore::ExecutableFileSectionInfo {const_cast<FEXCore::ExecutableFileInfo&>(File), 0}, Block);
+  }
+}
+
+enum class NeedsCacheRefresh {
```
bylaws, line 392: I feel like struct with a bool is better than this, unless this will be extended in the future to not just be a boolean

neobrain: I would've preferred this too, but the results would need to be stored in an `std::vector` then, which makes lookup into the container rather awkward (and makes the order of following operations less predictable, which isn't ideal when debugging).

An alternative would be `std::set`, but that requires a bunch of boilerplate code for the struct (`std::hash` etc) and it doesn't allow editing keys after insertion (so it makes flipping the boolean awkward).

Let me know if you see a better way though.

Source/Tools/FEXServer/ProcessPipe.cpp
neobrain: The SourcecodeMap assignment in Core.cpp prevents this currently.

bylaws: Mutable?

neobrain: Not ideal since there is no logical const-ness here, but sounds like the lesser evil than having `const_cast` elsewhere. Good idea!

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -275,6 +291,187 @@ void SendFDSuccessPacket(fasio::tcp_socket& Socket, int FD) {
...
+    OutputCodeMap.AppendSetMainExecutable(File);
+
+    for (auto& [Dependency, _] : Dependencies) {
+      if (Dependency.FileId != File.FileId) {
+        OutputCodeMap.AppendLibraryLoad(Dependency);
+      }
+    }
+  } else {
+    // List only the library itself
+    OutputCodeMap.AppendLibraryLoad(File);
```
bylaws, line 384: I'm surprised this isn't needed in the other case too, doesn't the codemap still need to contain library load info for the main executable

bylaws: Oh I suppose it's implicit given there are blocks?
EDIT: nvm, yeah I think i'm missing something or this is broken

neobrain: You're right, I assumed the SetMainExecutable block would implicitly register the binary filename too, but it doesn't (it only adds the entry but leaves the Filename unset).

Luckily this only simplifies the logic, since we can just drop the `if (Dependency.FileId != File.FileId)` check then.

Source/Tools/FEXServer/ProcessPipe.cpp
```diff
@@ -275,6 +291,187 @@ void SendFDSuccessPacket(fasio::tcp_socket& Socket, int FD) {
...
+// Discovers any pending code maps, parses their contents into a runtime data structure, and deletes them
+static std::map<FEXCore::ExecutableFileInfo, fextl::set<uintptr_t>>
+ImportPendingCodeMaps(const FEXCore::ExecutableFileInfo& MainFileId, bool HasMultiblock) {
+  // Detect code maps by checking file name suffixes by counting up an index.
+  // Code maps that are ready for reading must be non-empty and flock(FLOCK_EX) must succeed:
+  // - If empty, we tried generating the cache before the client could even lock it
+  // - If exclusively lockable, we know the client either closed or crashed
+  std::vector<std::string> CodeMaps;
+  for (int Index = 0; true; ++Index) {
+    auto CodeMap = fmt::format("{}/{}.{}.bin", CodeMapDirectory, FEXCore::CodeMap::GetBaseFilename(MainFileId, !HasMultiblock), Index);
```
neobrain: Oops:
```suggestion
    auto CodeMap = fmt::format("{}/new/{}.{}.bin", CodeMapDirectory, FEXCore::CodeMap::GetBaseFilename(MainFileId, !HasMultiblock), Index);
```

neobrain:
> An API in Config.{h,cpp} that returns the new/ready codemap dir paths would be nice, saves constructing manually everywhere

Do you have a specific suggestion? CodeMapDirectory is somewhat expensive to compute, so it would need to be given as a parameter. Further, all current users of that function need to append a filename to the directory, so you end up with something more verbose since you only save the `{}/new/` part to begin with.

Perhaps this would work? `CodeMap::GetPathForIncomingData(std::string_view CodeMapDirectory, const ExecutableFileInfo& Info, int Index, bool AddNombSuffix)`, which would construct a full file path (though presumably `Index` doesn't work on WoA if using timestamps instead).

Source/Tools/FEXServer/ProcessPipe.cpp
neobrain: If we add more data to ExecutableFileInfo (like your relocation PR does), we'd need to make these new members mutable too, so I'll stick with the `const_cast` for now after all.

bylaws: > If we add more data to ExecutableFileInfo (like your relocation PR does), we'd need to make these new members mutable too, so I'll stick with the `const_cast` for now after all.

I don't think so, relocations isn't assigned to by core and the frontend should have direct access to the underlying struct elsewhere, unless this is a Linux quirk

neobrain: Oh, yes you're right. I've been coding too much Rust recently and forgot non-const references can still exist elsewhere... :D

bylaws:
> > An API in Config.{h,cpp} that returns the new/ready codemap dir paths would be nice, saves constructing manually everywhere
> 
> Do you have a specific suggestion? CodeMapDirectory is somewhat expensive to compute, so it would need to be given as a parameter. Further, all current users of that function need to append a filename to the directory, so you end up with something more verbose since you only save the `{}/new/` part to begin with.

Perhaps even just a constant string vars ( "codemap/new/" "codemaps/ready") somewhere? Then frontends/offline compiler etc can just pull those in.
> 
> Perhaps this would work? `CodeMap::GetPathForIncomingData(std::string_view CodeMapDirectory, const ExecutableFileInfo& Info, int Index, bool AddNombSuffix)`, which would construct a full file path (though presumably `Index` doesn't work on WoA if using timestamps instead).

I'm unsure on this, think it's fine to leave this kind of thing to frontends.

neobrain:
> Perhaps even just a constant string vars ( "codemap/new/" "codemaps/ready") somewhere? Then frontends/offline compiler etc can just pull those in.

I don't think exposing these two specific things as constants would help overall readability since there's no good name for the concept "two components of a path", and if we can't common-ize a full API to choose codemap paths then I don't think sharing this specific part makes any sense.

For ProcessPipe.cpp, I think I'll drop `CodeMapDirectory` and instead replace it with `NewCodeMapDirectory` and `ReadyCodeMapDirectory`, so there'll be exactly one mention of each path only. Perhaps that's an option for WoA too?

bylaws:
> > Perhaps even just a constant string vars ( "codemap/new/" "codemaps/ready") somewhere? Then frontends/offline compiler etc can just pull those in.
> 
> I don't think exposing these two specific things as constants would help overall readability since there's no good name for the concept "two components of a path", and if we can't common-ize a full API to choose codemap paths then I don't think sharing this specific part makes any sense.
> 
> For ProcessPipe.cpp, I think I'll drop `CodeMapDirectory` and instead replace it with `NewCodeMapDirectory` and `ReadyCodeMapDirectory`, so there'll be exactly one mention of each path only. Perhaps that's an option for WoA too?

Sure, I was more thinking in terms of making it easy to change paths in future

neobrain:
> Sure, I was more thinking in terms of making it easy to change paths in future

Yeah, I think the main thing to ensure here would be to make all instances of the `codemap/new` and `codemap/ready` be greppable as such (instead of the `{}/{}` dance ProcessPipe is doing currently) and ideally only spelling them out in one place (per frontend) to begin with.

For sharing and maintaining code caching code, I'm less concerned about symbolic constants like these and more about general flow of logic that's harder to match up between subtrees.

Source/Tools/FEXServer/ProcessPipe.cpp
neobrain: #5123 added code that breaks exactly what I described above, so I will be moving back to `const_cast` to unblock this PR.

bylaws: > #5123 added code that breaks exactly what I described above, so I will be moving back to `const_cast` to unblock this PR.

How so?

neobrain: Apologies, apparently I was glancing over the build error a bit too quickly. It needed *more* const, not less.

Presumably making the `Relocations` pointer const is not a problem, right?

bylaws: Correct

PR #5129, opened 2025-12-16
Windows: Introduce ImageTracker for tracking per-loaded-image data 

bylaws:
and use it to commonise volatile metadata code (adding WOW64 image
voltmd support in the process). This will be extended to track mapped
images for code caching.

### DIFF
--- a/Source/Windows/WOW64/Module.cpp
+++ b/Source/Windows/WOW64/Module.cpp
@@ -33,6 +33,7 @@ $end_info$
 #include "Common/Config.h"
 #include "Common/Exception.h"
 #include "Common/TSOHandlerConfig.h"
+#include "Common/ImageTracker.h"
 #include "Common/InvalidationTracker.h"
 #include "Common/OvercommitTracker.h"
 #include "Common/CPUFeatures.h"
@@ -41,7 +42,6 @@ $end_info$
 #include "Common/CRT/CRT.h"
 #include "Common/PortabilityInfo.h"
 #include "Common/Handle.h"
-#include "Common/VolatileMetadata.h"
 #include "DummyHandlers.h"
 #include "BTInterface.h"
 #include "Windows/Common/SHMStats.h"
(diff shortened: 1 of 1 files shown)


Source/Windows/ARM64EC/Module.cpp
```diff
@@ -135,6 +134,7 @@ fextl::unique_ptr<FEX::Windows::StatAlloc> StatAllocHandler;
 std::optional<FEX::Windows::InvalidationTracker> InvalidationTracker;
 std::optional<FEX::Windows::CPUFeatures> CPUFeatures;
 std::optional<FEX::Windows::OvercommitTracker> OvercommitTracker;
+std::optional<FEX::Windows::ImageTracker> ImageTracker;
```
neobrain, line 137: It's a bit misleading to use `std::optional` for things that aren't actually optional. Can we use `unique_ptr` instead? (Probably applies to the other members as well, but it's the first time I'm noticing this.)

bylaws: My feeling is for a static variable used at most once the inline storage of optional is a little nicer, and it avoids a level indirection. Considering this pattern is already used throughout

Source/Windows/Common/ImageTracker.h
```diff
@@ -0,0 +1,49 @@
...
+  void HandleImageUnmap(uint64_t Address, uint64_t Size);
+
+private:
+  FEXCore::Context::Context& CTX;
+
+  FEX_CONFIG_OPT(ExtendedVolatileMetadataConfig, EXTENDEDVOLATILEMETADATA);
+  fextl::unordered_map<fextl::string, FEX::VolatileMetadata::ExtendedVolatileMetadata> ExtendedMetaData;
+};
+
+} // namespace FEX::Windows
```
neobrain: Missing newline at EOF. (I'm surprised this doesn't get flagged by clang-format?)

Source/Windows/ARM64EC/Module.cpp
neobrain: "used throughout" isn't exactly meaningful when you're talking about 3 other uses, all of which are right above and which follow equivalent uses of `fextl::unique_ptr` :)

There's no reason to worry about levels of indirection or storage class for an object as rarely accessed as an ImageTracker. Meanwhile, when reading the code it's typical to wonder whether an object can be empty, and `std::optional` heavily suggests it can. `unique_ptr` suggests this to a much lesser degree.

Concerns of readability should be prioritized over theoretical performance effects here.

bylaws: I mean it's used on WOW64/ARM64EC both, above is unique ptr only due to FEX API limitations. How about `using Singleton = std::optional;` or DelayedInit or similar for the same effect here (also note that the null case does matter, it's actively checked for in callbacks for InvalidationTracker on ARM64EC and when handling access violations at which point the objects may not be init so thinking about it optional makes more sense anyway?) Overcommit tracker this is also the case

neobrain: Does std::optional have a more tangible benefit than "I like it better" for ImageTracker? I think I brought up valid points against it, so codifying its use behind an abstraction wouldn't really help.

If some of these objects can indeed be empty then it's an okay use (if still somewhat foreign for dynamic, non-POD objects), but I don't see any emptiness checks in this PR (and it's initialized unconditionally + never reset).

bylaws: It is technically better/cleaner, and your readability concerns/points against it can be solved with such a using line. I don't really see a point in adopting something worse

PR #5130, opened 2025-12-16
CodeCache: Implement runtime cache validation

neobrain:
## Overview

This PR adds a validation pass for loaded code caches: All cached code is recompiled on-the-fly using the normal JIT code paths to ensure the offline-compiled cache perfectly matches runtime behavior. This has proven tremendously useful for debugging while bringing up the initial code caching implementation and when debugging regressions afterwards, so we'll want to keep maintaining this feature going forward.

The new validation pass is enabled using a new `EnableCodeCacheValidation` setting that will be off default since this extra validation is so expensive. Upon detecting any differences, a fatal error is raised (and an assertion message with diagnostic information is sent to the logs).

Here is some example output printed on failure. The mismatching host code address is automatically mapped back to the guest-side block address base, allowing IR to be printed along the way:
```
I 34745|34745 935.827 Cache load:   229 blocks; base=0x7ffff7ec2000; off=   0x1000-0x002c000; dbd3d54d31294247 ROOTFS/usr/lib/x86_64-linux-gnu/ld-linux-x86-64.so.2
E 34745|34745 935.883 recorded rip 0: 0x7ffff7ec3000 (offset 0x1000)
E 34745|34745 935.883 recorded rip 1: 0x7ffff7ec3000 (offset 0x1000)
E 34745|34745 935.883 IR:
        (%0) IRHeader %2, #0x7ffff7ec3000, #1, #2
        (%2) CodeBlock %3, %4
                (%3 i0) BeginBlock %2
                (%7 i64) InlineConstant #0x0
                %8(R16) i64 = SubWithFlags R7, %7
                (%9 i0) InvalidateFlags #0x10
                (%10 i64) InlineEntrypointOffset #0x7
                (%11 i64) StorePF R16
                (%12 i64) ExitFunction %10, None, %Invalid, %Invalid
                (%4 i0) EndBlock %2

A 34745|34745 935.883 Cache validation failed at offset 0x4: e0ffff10 <-> 00ffff10 (at 0xaaab6c6cc000 <-> 0xaaadfeb32000, guest block 0x1000)
```

## Implementation

All code is compiled in a separate ContextImpl instance. This ensures normal operation is not affected, and it allows constructing a fake Thread object (which is needed for WoA, where a valid Thread object may not be available otherwise).

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -70,13 +70,30 @@ public:
   ~CodeCache();
 
   ContextImpl& CTX;
+  fextl::unique_ptr<ContextImpl> ValidationCTX;
+  fextl::unique_ptr<Core::InternalThreadState> ValidationThread;
+  FEXCore::Core::CPUState::gdt_segment ValidationGDT[32] {};
   bool IsGeneratingCache = false;
 
   FEX_CONFIG_OPT(EnableCodeCaching, ENABLECODECACHINGWIP);
+  FEX_CONFIG_OPT(EnableCodeCacheValidation, ENABLECODECACHEVALIDATION);
 
   uint64_t ComputeCodeMapId(std::string_view Filename, int FD) override;
   bool SaveData(Core::InternalThreadState&, int TargetFD, const ExecutableFileSectionInfo&, uint64_t SerializedBaseAddress) override;
   bool LoadData(Core::InternalThreadState&, std::byte* MappedCacheFile, const ExecutableFileSectionInfo&) override;
+
+  /**
+   * Performs expensive extra validation on the loaded code cache data.
+   *
+   * This kicks off an in-process recompile of all cached blocks and compares
+   * them with the cached data. Differences will be reported as fatal errors,
+   * which can uncover bugs like for example:
+   * - mismatches of the JIT configuration used during cache generation
+   * - hidden position dependencies due to missing FEX relocations
+   * - incorrect instruction padding
+   */
+  void Validate(const ExecutableFileSectionInfo&, fextl::set<uint64_t> GuestBlocks, const fextl::set<uint64_t>& HostBlocks,
+                std::span<std::byte> CachedCode);
 
   void InitiateCacheGeneration() override {
     IsGeneratingCache = true;
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -3,8 +3,12 @@
 
 #include <Interface/Context/Context.h>
 #include <Interface/Core/ArchHelpers/Arm64Emitter.h>
+#include <Interface/Core/Dispatcher/Dispatcher.h>
+#include <Interface/Core/JIT/DebugData.h>
 #include <Interface/Core/JIT/Relocations.h>
 #include <Interface/Core/LookupCache.h>
+#include <Interface/Core/OpcodeDispatcher.h>
+#include <Interface/IR/PassManager.h>
 
 #include <FEXCore/Core/Thunks.h>
 #include <FEXCore/HLE/SourcecodeResolver.h>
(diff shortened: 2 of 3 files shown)


FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,126 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+        auto tail = reinterpret_cast<CPU::CPUBackend::JITCodeTail*>(reinterpret_cast<uintptr_t>(header) + header->OffsetToBlockTail);
+        (i == 0 ? GuestBlockAddr : GuestBlockAddrRef) = tail->RIP - Section.FileStartVA;
+        LogMan::Msg::EFmt("recorded rip {}: {:#x} (offset {:#x})", i, tail->RIP, tail->RIP - Section.FileStartVA);
+
+        if (i == 1 && tail->RIP >= Section.BeginVA && tail->RIP < Section.EndVA) {
+          auto [IRView, TotalInstructions, TotalInstructionsLength, StartAddr, Length, _] =
+            ValidationCTX->GenerateIR(ValidationThread.get(), tail->RIP, false, FEXCore::Config::Get_MAXINST());
+          fextl::stringstream ss;
+          FEXCore::IR::Dump(&ss, &*IRView);
+          LogMan::Msg::EFmt("IR ({}):\n{}", ValidationCTX->Config.Is64BitMode(), ss.str());
```
neobrain: Should print something more explicit than `IR (true)` here.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,126 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+  }));
+  (void)ApplyCodeRelocations(Section.FileStartVA, CodeBufferRangeRef, NewRelocations, false);
+
+  if (ValidationCTX->LatestOffset <= CodeBufferRangeRef.size()) {
+    // Reference compilation produced fewer bytes than our cache, so validation is going to fail.
+    // Make sure we don't output any garbage bytes though.
+    CodeBufferRangeRef = CodeBufferRangeRef.subspan(0, ValidationCTX->LatestOffset);
+  }
+
+  CodeBufferRangeRef.data()[4] = std::byte {0};
```
neobrain: Remove debugging code

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,126 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+        LogMan::Msg::EFmt("recorded rip {}: {:#x} (offset {:#x})", i, tail->RIP, tail->RIP - Section.FileStartVA);
+
+        if (i == 1 && tail->RIP >= Section.BeginVA && tail->RIP < Section.EndVA) {
+          auto [IRView, TotalInstructions, TotalInstructionsLength, StartAddr, Length, _] =
+            ValidationCTX->GenerateIR(ValidationThread.get(), tail->RIP, false, FEXCore::Config::Get_MAXINST());
+          fextl::stringstream ss;
+          FEXCore::IR::Dump(&ss, &*IRView);
+          LogMan::Msg::EFmt("IR ({}):\n{}", ValidationCTX->Config.Is64BitMode(), ss.str());
+        } else {
+          LogMan::Msg::EFmt("Can't dump IR for out-of-range RIP {:#x}", tail->RIP);
```
neobrain: This is accidentally printed when i==0 currently.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,126 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+    // Align down to instruction size
+    auto Idx = AlignDown(std::distance(CodeBufferRangeRef.begin(), Mismatch), 4);
+
+    auto BlockIt = std::prev(HostBlocks.lower_bound(*HostBlocks.begin() + Idx + 1));
+    std::optional<uint64_t> GuestBlockAddr;
+    std::optional<uint64_t> GuestBlockAddrRef;
+    if (BlockIt != HostBlocks.end()) {
+      for (int i : {0, 1}) {
+        std::span Buffer = (i == 0 ? CachedCode : CodeBufferRangeRef);
+        auto header = reinterpret_cast<CPU::CPUBackend::JITCodeHeader*>(&Buffer[*BlockIt - *HostBlocks.begin()]);
```
neobrain: The data at this offset used to always be a JITCodeHeader, however since the addition of multiblocks with multiple entrypoints, that's no longer the case. Instead, we have to collect the `BlockBegin` members returned from CompileCode above to find the true multiblock starts. Alternatively, decode the `adr` instruction at `*BlockIt - *HostBlocks.begin() + 4` to get the offset to the JITCodeHeader.

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -70,14 +70,31 @@ class CodeCache : public AbstractCodeCache {
   ~CodeCache();
 
   ContextImpl& CTX;
+  fextl::unique_ptr<ContextImpl> ValidationCTX;
```
bylaws, line 73: I think this deserves to be wrapped in a ValidationContext struct or similar

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,126 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+    ValidationCTX->SetSyscallHandler(CTX.SyscallHandler);
+    ValidationCTX->SetThunkHandler(CTX.ThunkHandler);
+    if (!ValidationCTX->InitCore()) {
+      ERROR_AND_DIE_FMT("Failed to create cache load validation context");
+    }
+
+    ValidationThread.reset(ValidationCTX->CreateThread(0, 0, nullptr));
+
+    auto Frame = ValidationThread->CurrentFrame;
+    Frame->State.segment_arrays[FEXCore::Core::CPUState::SEGMENT_ARRAY_INDEX_GDT] = &gdt[0];
```
bylaws: I think gdt can be a local variable here right?

neobrain: FEX copies that pointer around internally, so the underlying storage must remain valid across calls to this function (since we don't recreate the context for each validation).

FEXCore/Source/Interface/Context/Context.h
neobrain: Does it? It's only 3 data members, and CodeCache itself only has two more. And arguably, within `CodeCache::Validate` it's easier to distinguish `ValidationCTX` from `CTX` than `ValidationState.CTX` would be.

Is there a specific concern this seeks to resolve? Perhaps moving initialization to a more isolated `PrepareValidation` helper? (I thought using a constructor might be neat, but such a helper would achieve the same cleanup with less boilerplate)

bylaws: This was mostly because of gdt existing as a member, and if that were renamed to ValidationGdt my thought was at that point just a nested struct defined internally would achieve the exact same. But I'm impartial if you'd prefer just that rename

neobrain: Fair point, renamed to ValidationGDT.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: I went for decoding `adr` here.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -476,9 +480,129 @@ bool CodeCache::LoadData(Core::InternalThreadState& Thread, std::byte* MappedCac
...
+
+    auto BlockIt = std::prev(HostBlocks.lower_bound(*HostBlocks.begin() + Idx + 1));
+    std::optional<uint64_t> GuestBlockAddr;
+    std::optional<uint64_t> GuestBlockAddrRef;
+    if (BlockIt != HostBlocks.end()) {
+      for (int i : {0, 1}) {
+        std::span Buffer = (i == 0 ? CachedCode : CodeBufferRangeRef);
+
+        // Second instruction is always a constant load for relative offset to the (multi)block start
+        int32_t addr = ((*(uint32_t*)&Buffer[*BlockIt - *HostBlocks.begin() + 4]) & 0x3ff'ffe0) << 11;
```
bylaws: reinterpret_cast and tbh I would use BlockStart for getting header since adr very much might not stay around but I suppose it's fine

neobrain: We don't store BlockBegin in the code cache, so there's no other way to reconstruct that information for one of the two iterations. We could optimistically assume that the two BlockBegin sets match, but having fewer such assumptions makes the output more reliable.

bylaws: Ah sure, yeah that can be adopted later then upon any such changes

PR #5134, opened 2025-12-22
Windows: Implement _[w]sopen file APIs

bylaws:


Source/Windows/include/winternl.h
```diff
@@ -10,6 +10,18 @@
...
+#define SH_DENYRW 0x10 /* Deny read/write */
+#define SH_DENYWR 0x20 /* Deny write */
+#define SH_DENYRD 0x30 /* Deny read */
+#define SH_DENYNO 0x40 /* Deny nothing */
+
+#define _SH_COMPAT SH_COMPAT
+#define _SH_DENYRW SH_DENYRW
+#define _SH_DENYWR SH_DENYWR
+#define _SH_DENYRD SH_DENYRD
+#define _SH_DENYNO SH_DENYNO
```
neobrain, line 23: What are these about?

bylaws: These are straight from wine headers. I think ms began without the underscore prefix but then added after to avoid falling afoul of posix or similar?

Issue #5125, opened 2025-12-15
Code cache subtask tracking


neobrain:
Tracking some specific subtasks required as we progress towards fully functional code caching.

* #5116
* #5117
* #5123 
* Persistently mmap caches
* Delete caches generated from old FEX versions
* Enable NOP padding only for cache generation+validation runs (see #5110)
* Debounce cache regeneration when the number of new blocks is low
* Ensure we never ever produce zero-size code maps or move to a non `flock()`-based method of guarding code map access (see #5127)
* Machine-agnostic CodeMapIds
* #5141
* CodeBuffer resizes should preserve data
* Drop CodeBuffer size limit (at least when generating caches and validating them)
* Multi-threaded cache generation (#5203)
* On-disk compression
* Cache header only allocates 8 bytes for version information
* Moving RootFS folder invalidates code maps
* `FEX_MAXINST` poisons code maps (similar to `FEX_MULTIBLOCK`)
* Consider full guest executable hashing for consistency checking of file IDs
* Consider de-virtualizing `CodeCache::ComputeCodeMapId`
* https://github.com/FEX-Emu/FEX/issues/5379

Sonicadvance1:
- `ThreadRemoveCodeEntry` IR op needs to be code caching aware. Would break the FullSMC path on codecaching otherwise.

PR #5140, opened 2025-12-22
First round of `LoadConstant` auditing

Sonicadvance1:
Audits the direct uses of `LoadConstant` for padding requirements.
Two things:
- `ThreadRemoveCodeEntry` uses the block entry but relocations aren't wired up to it. This will break under codecaching
  - Tracked in #5125
- `Constant` IR op needs to be audited through the OpcodeDispatcher, with padding data passed through from the dispatcher.

This should reduce a significant amount of padding from happening in the JIT with code caching enabled, even with this first pass. I'll be walking through the OpcodeDispatcher for their use of padding required constants and working on passing that information through to the backend.

### DIFF
--- a/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
+++ b/FEXCore/Source/Interface/Core/ArchHelpers/Arm64Emitter.h
@@ -117,7 +117,7 @@ public:
     // Choose to pad or not depending on if code-caching is enabled.
     AUTOPAD,
   };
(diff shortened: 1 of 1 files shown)


bylaws:
I feel that NOPAD should really be a default arg here, the cases where padding is required are purely relocations which are special anyway and one would be aware if padding were necessary when writing such code. It adds a lot of noise otherwise

Sonicadvance1:
> I feel that NOPAD should really be a default arg here, the cases where padding is required are purely relocations which are special anyway and one would be aware if padding were necessary when writing such code. It adds a lot of noise otherwise

I'll do that as a post-merge change. I want to ensure that if code-caching needs to get bisected through this that it doesn't break.

PR #5148, opened 2025-12-23
Interpreter: Moves around the thread and ELF initialization code

Sonicadvance1:
This allows the context and parent thread objects to be created earlier,
allowing the VDSO and ELFCodeLoader mapping functions to have a thread
object for tracking memory mappings through the regular guest routines.

This means that we no longer need to do any form of deferred handling
for code caching as all the state is ready early in the initialization
process.

A little bit of care needed to be taken to ensure we still close the
ELFCodeLoader's FDs later and that VDSO unmapping happens before tearing
down the parent thread, but overall this is mostly just passing the
InternalThreadState object around as normal.

I couldn't find any functional regression from this change alongside
code caching, but it would be good for @neobrain to double check this.

### DIFF
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -409,6 +409,12 @@ public:
       return false;
     }
 
+    if (Thread) {
+      // Update the thread persona.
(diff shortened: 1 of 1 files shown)


lioncash (approved):
LGTM as far as I can tell

PR #5153, opened 2025-12-24
WritePriorityMutex: Add some more documentation

Sonicadvance1:
Just my brain spinning as I try and determine what is causing some hanging. Seems to be WINE specific so might not even be in FEX code.

Good to have some more documentation so when I read this again I don't need to make some more logic deductions.


bylaws:
needs updating

Sonicadvance1:
Documentation updated.

bylaws (approved):

PR #5155, opened 2025-12-24
FEXConfig: use one qrc

crueter:
Rather than manually creating a Qt 5 and Qt 6 QRC, instead we create a
single qrc that references a flattened directory. Then, we copy each of
its dependents (main, dialogs, icon) into the binary directory as-is,
resulting in just one qrc rather than having to maintain one for each.


crueter:
So Qt 5 doesn't support QML modules from CMake. I guess qmldir it is 😵‍💫

crueter:
Note for self: Qt 5 supports only old var style https://doc.qt.io/archives/qt-5.15/qtcore-cmake-qt5-add-resources.html

No idea what to do at this point. The evil part of me says "drop Qt 5" but the 3 Ubuntu 20.04 users will impale me with pitchforks. The evil but slightly less pitchfork-risky solution is to just write the file manually. The slightly less evil idea is to do configure_file or something

Sonicadvance1:
> Note for self: Qt 5 supports only old var style https://doc.qt.io/archives/qt-5.15/qtcore-cmake-qt5-add-resources.html
> 
> No idea what to do at this point. The evil part of me says "drop Qt 5" but the 3 Ubuntu 20.04 users will impale me with pitchforks. The evil but slightly less pitchfork-risky solution is to just write the file manually. The slightly less evil idea is to do configure_file or something

Well, we only support back to Ubuntu 22.04, but it also doesn't have QT6 in the base packages (It's in backports I think?) which means the PPA builder can't rely on it.

crueter:
Actually the package search indicates 6.2 is just in Universe https://packages.ubuntu.com/search?suite=jammy&arch=any&searchon=names&keywords=qt6-declarative

crueter:
Okay so at this point there are two options:
1. Drop Qt 5 support since 22.04 has Qt 6 in base, and just use qt_add_qml_module
2. Use a single qrc, but flatten things in the CMAKE_CURRENT_BINARY_DIR

Final decision is not up to me, but option 1 should probably be done in a more formal manner where the deprecation/dropping is warned to users and developers + via the wiki. So my vote is for option 2 even though it somewhat pains me.

crueter:
>I forgot to git add

neobrain:
> Drop Qt 5 support since 22.04 has Qt 6 in base, and just use qt_add_qml_module

Last I checked this was not a properly integrated Qt6 configuration, lacking proper theming etc.

> Use a single qrc, but flatten things in the CMAKE_CURRENT_BINARY_DIR

As much as I'm personally annoyed by the current solution, does this have a meaningful benefit over leaving things as they are? After all there's always a low but non-zero risk these refactors lead to regressions, which we need to be particularly wary of when talking about Ubuntu 22.04 that personally none of us are regularly using at this point.

IMO the way forward here is to wait for the next LTS release and then drop the Qt 5 paths. Even if we continue to support 22.04, this should limit the impact to users who'll be fine with degraded Qt integration anyway.

crueter:
> As much as I'm personally annoyed by the current solution, does this have a meaningful benefit over leaving things as they are?

One thing I want to look at soon is to split the 800-ish line main.qml into several separate files. This makes it simpler to make individual changes to one aspect (e.g. one singular tab). This solution specifically makes it easier to *add* new files, because rather than having to remember to edit both qml5 and qml6 you instead just edit one. Though it does have the caveat of being moderately confusing to deal with imports... but let's be real here, the current solution really isn't any less confusing.



> Last I checked this was not a properly integrated Qt6 configuration, lacking proper theming etc.

I remember this only being the case for Qt Widgets. IIRC it had to do with qgtk3 just not activating because GNOME sucks, but I may be able to spin up a VM later today and check it out to verify.

> IMO the way forward here is to wait for the next LTS release and then drop the Qt 5 paths.

Assuming that 22.04 does have bad Qt integration that's probably better. To be fair I don't think anyone besides me has any desire to add new files to the QML...

neobrain:
> One thing I want to look at soon is to split the 800-ish line main.qml into several separate files. This makes it simpler to make individual changes to one aspect (e.g. one singular tab). This solution specifically makes it easier to add new files, because rather than having to remember to edit both qml5 and qml6 you instead just edit one. Though it does have the caveat of being moderately confusing to deal with imports... but let's be real here, the current solution really isn't any less confusing.

That refactor could be useful, but (besides such a change) we aren't exactly adding files to FEXConfig on a daily basis. As long as it's still as simple as adding the new files to qml5 and qml6 individually, I'll take the battle-tested existing code over any changes here.

> I remember this only being the case for Qt Widgets. IIRC it had to do with qgtk3 just not activating because GNOME sucks, but I may be able to spin up a VM later today and check it out to verify.

FTR I specifically verified this on 22.04 with Plasma, where it didn't even pick up a dark theme despite the desktop being configured as such.

crueter:
Ubuntu 22.04's Qt 6 is so fried it doesn't even launch.

```
QQmlApplicationEngine failed to load component
qrc:/main.qml:101:5: Type MessageDialog unavailable
qrc:/dialogs/MessageDialog.qml:4:1: FromQt.MessageDialog is not a type
Segmentation fault (core dumped)
```

I will close this for now. If I ever decide to make any changes to FEXConfig (probably random UI design stuff) I will reopen this and redo it; or when Ubuntu 22.04 support is dropped, go back to using `qt_add_qml_module`.

PR #5156, opened 2025-12-24
[cmake] propagate `-ISource` to all Tools

crueter:
Rather than individually adding `${CMAKE_SOURCE_DIR}/Source` as
an include directory to each target, just use `include_directories` once
in the Tools directory and each subsequent target will have this
propagated down.

Also removed a seemingly unnecessary `-I` in LinuxEmulation--maybe
needed? But I can't test compilation right now as I don't have an ARM
development environment on hand for the next day or two.

### DIFF
--- a/Source/Tools/CodeSizeValidation/CMakeLists.txt
+++ b/Source/Tools/CodeSizeValidation/CMakeLists.txt
@@ -4,7 +4,6 @@ set (SRCS Main.cpp)
 add_executable(CodeSizeValidation ${SRCS})
 target_include_directories(CodeSizeValidation
   PRIVATE
-    ${CMAKE_CURRENT_SOURCE_DIR}/Source/
     ${CMAKE_BINARY_DIR}/generated
 )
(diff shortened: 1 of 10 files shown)


Sonicadvance1:
oop, merge conflict.

Sonicadvance1:
You'll want to rebase rather than merge to get rid of extraneous merge commits. We don't do squash merging here.

Sonicadvance1:
Also interestingly, you deleted the added `include_directories(..)` in the tool's cmake and things are still compiling. Maybe has been unnecessary entirely for a while? :thinking:

crueter:
> Also interestingly, you deleted the added `include_directories(..)` in the tool's cmake and things are still compiling. Maybe has been unnecessary entirely for a while? 🤔

That was accidental, but yeah it looks like some include directories are getting propagated via PUBLIC/INTERFACE targets. If it works it works :P

p.s. working on another large PR, will fix history once I'm at a good stopping point

crueter:
okay, should be good now

PR #5157, opened 2025-12-24
[cmake] use `MINGW` builtin rather than custom detection

crueter:
CMake has had the `MINGW` builtin to describe MinGW targets since at
least version 3.2, so it can safely be used. This variable is also set
for the MSYS2 environments, so CLANGARM64 also correctly sets `MINGW`.

Note that this depends on https://github.com/FEX-Emu/jemalloc/pull/11.


Sonicadvance1:
> Note that this depends on https://github.com/FEX-Emu/jemalloc/pull/11.

Merged this.

crueter:
updated the submodule

Sonicadvance1 (approved):

PR #5159, opened 2025-12-24
WIP: [cmake] allow fetching deps with CPM instead of submodules

crueter:
Extremely WIP, thunks and tests are untested since I don't have an aarch64 development environment on me at the time. Will document this later.

Main TODOs:
- [ ] copy other modules over to CPM
- [ ] thunks wtf
- [ ] test
- [ ] robin map????????

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -35,7 +35,7 @@
 [submodule "External/robin-map"]
 	shallow = true
 	path = External/robin-map
-	url = https://github.com/FEX-Emu/robin-map.git
+	url = https://github.com/crueter/robin-map.git
 [submodule "External/Vulkan-Headers"]
 	shallow = true
 	path = External/Vulkan-Headers
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ out/
 .vs/
 *.pyc
 .cache
+.qtcreator
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -34,6 +34,7 @@ set (FEXCORE_PROFILER_BACKEND "gpuvis" CACHE STRING "Set which backend to use fo
 option(ENABLE_GLIBC_ALLOCATOR_HOOK_FAULT "Enables glibc memory allocation hooking with fault for CI testing")
 option(USE_PDB_DEBUGINFO "Builds debug info in PDB format" FALSE)
 option(BUILD_STEAM_SUPPORT "Builds FEX for integration into Steam" FALSE)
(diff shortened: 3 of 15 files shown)


Sonicadvance1:
I'm not sure how I feel about this one yet.

crueter:
I'll document more thoroughly when I actually finish it, but the gist is
- submodules are annoying (I have already been bitten by forgetting to `git submodule update` multiple times), whereas with this each commit sha/tag is its own cache entry in `.cache/cpm/<packagename>`. This also has the benefit of every build being 100% reproducible
- this uses source archives instead of full Git trees, which is net faster as it doesn't need to include the `.git` directory for each

.gitmodules
```diff
@@ -35,7 +35,7 @@
 [submodule "External/robin-map"]
 	shallow = true
 	path = External/robin-map
-	url = https://github.com/FEX-Emu/robin-map.git
+	url = https://github.com/crueter/robin-map.git
```
lioncash, line 38: tbh, it'd be nice if we could track the main [repo](https://github.com/Tessil/robin-map) for robin-map, especially since it seems that we're 2 minor versions behind now (i.e. 1.2.1, but 1.4.1 is out, which does seem to have some performance improvements, though small, and also unbreaks systems using CMake 4.0+)

Sonicadvance1: We have one patch on top of upstream so we can't track upstream. https://github.com/FEX-Emu/robin-map/commit/d5683d9f1891e5b04e3e3b2192b5349dc8d814ea

lioncash: oh, bleh. I guess in that case it would be good to reapply on top of the latest release then (if possible), so we don't have an eventual build break

crueter: To be honest I wish we could just use something like unordered_dense instead

In CPM land, applying patches is as simple as creating a patch and applying it with the `PATCHES` option. That's another benefit of using it over submodules.

crueter:
This needs a lot more time to bake, so I'm closing it for now and keeping it as a future reference. Self-notes:

- A separate PR should be made for preferring system dependencies if available. This makes the CPM porting work easier and reduces this PR's size--this depends on #5162 to avoid the robin-map patching thing.
- MERGE CONFLICTS
- Will probably want each CPM dependency to have an option like `<name>_DIR` where users can prefetch the dependency and use it in the same way you can submodules (see https://gpo.zugaina.org/app-emulation/FEX)
- Patches... ideally, none are used, but you can't win it all I guess. Could go with the same CPMUtil route of `.patch/<packagename>/0000-patchname.patch` perhaps.

PR #5162, opened 2025-12-25
FEXCore: use ankerl::unordered_dense over tsl

crueter:
ankerl::unordered_dense is faster on average and has less memory usage
than both std::unordered_map and tsl::robin_map. It's also seemingly
better maintained, and shouldn't need an extra patch on top of upstream--this has the added benefit of allowing users to use the system unordered_dense package, which I'm sure the Gentoo packagers will appreciate. Will make future externals handling refactors easier as well, wink wink

Obviously, this will need a lot of testing.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -32,10 +32,6 @@
 [submodule "External/Catch2"]
 	path = External/Catch2
 	url = https://github.com/catchorg/Catch2.git
-[submodule "External/robin-map"]
-	shallow = true
-	path = External/robin-map
-	url = https://github.com/FEX-Emu/robin-map.git
 [submodule "External/Vulkan-Headers"]
 	shallow = true
 	path = External/Vulkan-Headers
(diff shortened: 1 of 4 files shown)


Sonicadvance1:
In order to limit the potential fallout from this change. I would prefer if only tsl gets replaced with ankerl and leave the std implementation alone.

crueter:
done. Interestingly `unordered_multi{map,set}` don't seem to be used. Not sure what those are for, but I'll leave them be for now

Sonicadvance1:
> done. Interestingly `unordered_multi{map,set}` don't seem to be used. Not sure what those are for, but I'll leave them be for now

The fextl stuff is grown organically as we hit std containers that need wrapping with allocators. So it was likely used before and removed, or I overreached while implementing. Doesn't really matter, it'll sit around and be used if someone needs it.

PR #5163, opened 2025-12-25
CI: fix pr-code-format

crueter:
Previously, changes to externals were also tracked, resulting in
absolutely hilarious abominations like https://github.com/FEX-Emu/FEX/actions/runs/20498228017/job/58901276463?pr=5162

So rather than dealing with that weird action, just fetch the main
branch and run a `diff --name-only` with the PR's merge base with main.
This is how I've done several dozen diff-based scripts (license headers,
clang-format, etc) and it works perfectly, so let's just use this.
Should run quicker too but don't put any money on it.

I also went ahead and removed a bunch of unnecessary quotations cuz
wynaut.

### DIFF
--- a/.github/workflows/pr-code-format.yml
+++ b/.github/workflows/pr-code-format.yml
@@ -1,7 +1,7 @@
-# Inspired by LLVM's pr-code-format.yml at 
+# Inspired by LLVM's pr-code-format.yml at
 # https://github.com/llvm/llvm-project/blob/main/.github/workflows/pr-code-format.yml
 
-name: "Check code formatting"
+name: Check code formatting
 on:
   pull_request:
     branches:
@@ -13,7 +13,7 @@ jobs:
     if: github.repository == 'FEX-Emu/FEX'
 
     steps:
-      - name: Fetch FEX sources
+      - name: Checkout
         uses: actions/checkout@v4
         with:
           ref: ${{ github.event.pull_request.head.sha }}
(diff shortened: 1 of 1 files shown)


crueter:
did some final cleanups, should be good now

neobrain:
@pmatos Could you take a look at this?

pmatos:
> @pmatos Could you take a look at this?

Yep. Give me a couple of hours.

.github/workflows/pr-code-format.yml
```diff
@@ -27,43 +27,37 @@ jobs:
...
-
-      - name: "Listed files"
-        env:
-          CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          echo "Formatting files:"
-          echo "$CHANGED_FILES"
+          BASE=$(git merge-base main HEAD)
+          FILES=$(git diff --name-only "$BASE")
+          echo "CHANGED_FILES=$FILES" >> $GITHUB_ENV
```
pmatos, line 33: Thanks for the PR and sorry for the delay. I am bit concerned about this wrt newlines, spaces in filenames and how files are split. 

1. Would it be better to replace newlines by commas since that's what we have been using from `step-security/changed-files`? 
2. Need to confirm that GITHUB_ENV is happy with spaces in paths.

Otherwise I like dropping the dependency on another action.

pmatos: We definitely expect files to be split by commas in `code-format-helper.py`. We should probably do something like:
```
FILES=$(git diff --name-only "$BASE" | tr '\n' ',' | sed 's/,$//')
```

crueter: Fixed, for example changing vk headers commit causes:

`.github/workflows/pr-code-format.yml,External/Vulkan-Headers`

pmatos:
@crueter looks good. this is hard to test locally. Lets hope it holds up in practice. Thanks.

pmatos:
@crueter you might have to re-request a review so I can formally approve it. Weird github flow.

crueter:
augahahhhhaghha

PR #5164, opened 2025-12-28
ImageTracker: Track loaded PE images for LookupExecutableFileSection

bylaws:


Source/Windows/Common/ImageTracker.cpp
```diff
@@ -75,13 +118,26 @@ void ImageTracker::HandleImageMap(std::string_view Path, uint64_t Address, bool
 
   if (!VolatileInstructions.empty() || !VolatileValidRanges.Empty()) {
     LogMan::Msg::DFmt("Loaded volatile metadata for {:X}: {} entries", Address, VolatileInstructions.size());
-    std::scoped_lock Lock(CTX.GetCodeInvalidationMutex());
```
Sonicadvance1, line 78: Isn't this a problem? EVMD is dumped in to a shared data structure so it isn't thread safe to modify in this way.

bylaws: The lock scope is just increased to cover whole function

Sonicadvance1: Oop, somehow missed that.

Sonicadvance1 (approved):

neobrain:
What's the rationale behind `CodeCache: Use defaulted dtor for ExecutableFileInfo`? This adds a header dependency and shouldn't be needed.

bylaws:
I couldn't compile without it, when using it in some containers in headers. Cannot remember specifics

PR #5166, opened 2025-12-28
[cmake] refactor: compiler and architecture handling

crueter:
- Do compiler/architecture checks EARLY, don't waste time doing random
  configuration stuff if the user can't even compile in the first place
- MSVC is unsupported, I assume? So add a check to disallow. There's
  literally no MSVC or MSC_VER checks anywhere, so...
- Rather than using the MSVC architecture definitions, use our own
  `ARCHITECTURE_arm64` et al. Hijacking existing "standard" definitions
  is a very bad idea. Also makes it more readable in CMake
- Change the x86 host check to `x86|amd64`. Some systems still refer to
  themselves as x86 despite being 64-bit for... reasons, and I saw one a
  very long time ago that referred to it as amd64. This should
  basically never come up, nor is it really relevant given that FEX is
  for arm64... but it kinda annoyed me so whatever.

TODOs:
- Should we check `CMAKE_SIZEOF_VOID_P (equal) 64`? I don't think anyone
  is even trying to compile this thing on armv7 or older, but might as
  well? maybe?
- What's the status of *BSD, Solaris, macOS? Technically macOS does
  support Wine, not sure about the others.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,7 +70,8 @@ else ()
 endif()
 
 ## Architecture Handling ##
-if (CMAKE_SYSTEM_PROCESSOR MATCHES "x86|amd64")
+string(TOLOWER ${CMAKE_SYSTEM_PROCESSOR} processor)
+if (processor MATCHES "x86|amd64")
   option(ENABLE_X86_HOST_DEBUG "Enables compiling on x86_64 host" FALSE)
   if (NOT ENABLE_X86_HOST_DEBUG)
     message(FATAL_ERROR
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
> Rather than using the MSVC architecture definitions, use our own
ARCHITECTURE_arm64 et al. Hijacking existing "standard" definitions
is a very bad idea. Also makes it more readable in CMake

Today I learned that MSVC uses these for architecture definitions.

crueter:
May also be worth it to *explicitly* disallow all unsupported arches: all 32-bits, ppc, riscv(?), mips (RIP), loongarch, s390, etc. The proper way to do this would be a dedicated module to detect the architecture using all known symbols from GCC and Clang. There is actually a permissively licensed module that does this (shoutout merryhime) https://git.eden-emu.dev/eden-emu/dynarmic/src/branch/master/CMakeModules/DetectArchitecture.cmake

Adding new architectures isn't too difficult with that script too :) I think that should be in a separate PR, I have a notorious habit of making giant PRs and I want to avoid that here.

bylaws:
I lean towards using the standard __aarch64__/__x86_64__ defines, however arm64ec complicates this by not defining __aarch64__ and only __arm64ec__ so every arm check would need to check both that and aarch64.

Thoughts?

crueter:
Personally, I prefer `ARCHITECTURE_*` defines because they are easy to remember as opposed to the rather unorthodox `__*__` arch defines. As for arm64ec, it looks like every ifdef for that is standalone; I don't see any `arm64 || arm64ec` checks anywhere?

If there *are* commonalities what I would do is have both arm64 and arm64ec define a variable/define, e.g. `ARCHITECTURE_arm` alongside their individual variables. But I'm not really knowledgeable enough on the differences b/t/ arm64 and arm64ec to make the call re: duplication

neobrain:
Just for reference:

> What's the status of *BSD, Solaris, macOS? Technically macOS does support Wine, not sure about the others.

Out of these, *BSD is the only useful target for FEX, but none of us ever tested that platform.

PR #5168, opened 2025-12-28
[cmake] reduce usage of trivial variables

crueter:
Trivial variables like SRCS, NAME, etc. actually do more harm than good.
They *will* make your IDE mad, and are also less readable. Remember:
verbosity is not a bad thing! Usually

Also: did a few tiny cleanups that I missed from my `endpara` PR.

### DIFF
--- a/External/SoftFloat-3e/CMakeLists.txt
+++ b/External/SoftFloat-3e/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-set(SRCS
+add_library(softfloat_3e STATIC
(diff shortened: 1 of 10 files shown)


neobrain:
This is much appreciated, thanks!

Just a small note that applies in general: Our scripts for generating release notes require commits to be formatted consistently (`Subsystem: Change description`), so it'd be nice if you could adapt this for any pending/upcoming PRs. Thanks again!

crueter:
Noted

PR #5169, opened 2025-12-28
[cmake] better option descriptions + more consistent language

crueter:
- "Enables"/"Uses" -> Enable/Use
- Be a bit more descriptive with... the descriptions


Sonicadvance1 (approved):
Sure.

PR #5170, opened 2025-12-28
[cmake] more parenthesis cleanups, linker gc module, more same-line stuff

crueter:
I may have gotten carried away.

- I missed some stuff for end parenthesis because I accidentally
  searched within project files instead of the entire directory (so some
  thunk/test/windows stuff was missed), cleaned those up.
- `INTERFACE`, `PUBLIC`, `PRIVATE`, `RUNTIME`, `LIBRARY` should be on
  the same line as the target name. (I should really invest in making a
  style guide...)
- Some short statements were unnecessarily split across multiple
  lines--cleaned those up
- Made a common `LinkerGC` module that applies gc-sections etc. to a
  target in Release mode
- Usually for functions you want to have something on the first line,
  e.g. `FILES`/`DIRECTORY` for install, or the target/a positional
  argument, etc etc. Not always though, notably for some custom_command
  calls

TODO:
- What's with the `list(APPEND LIBS...)` stuff? It's used really
  inconsistently, sometimes not at all, sometimes it looks like there're
  duplicates? A more thorough cleanup is in order there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -203,7 +203,7 @@ if (HAS_CLANG_PRESERVE_ALL)
   else()
     message(STATUS "Has clang::preserve_all")
   endif()
-endif ()
+endif()
 
 if (ARCHITECTURE_arm64 AND HAS_CLANG_PRESERVE_ALL)
   add_definitions("-DFEX_PRESERVE_ALL_ATTR=__attribute__((preserve_all))" "-DFEX_HAS_PRESERVE_ALL_ATTR=1")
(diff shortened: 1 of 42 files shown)


Sonicadvance1:
Merge conflict now, needs a rebase.

Sonicadvance1:
Another conflict.

Sonicadvance1 (approved):

neobrain:
Nice cleanups!

>    What's with the list(APPEND LIBS...) stuff? It's used really
>    inconsistently, sometimes not at all, sometimes it looks like there're
>    duplicates? A more thorough cleanup is in order there.

I'm in support of dropping this pattern entirely. They make the build logic unnecessarily non-local and there's no compelling reason not to pass the target names to `target_link_libraries` directly.

PR #5171, opened 2025-12-28
Thunks/Vulkan: Update for v1.4.337 

Sonicadvance1:
Fixes DXVK/vkd3d-proton on newer mesa.

### DIFF
--- a/ThunkLibs/libvulkan/libvulkan_interface.cpp
+++ b/ThunkLibs/libvulkan/libvulkan_interface.cpp
@@ -56,6 +56,8 @@ template<>
 struct fex_gen_type<VkPhysicalDevice_T> : fexgen::opaque_type {};
 template<>
 struct fex_gen_type<VkQueue_T> : fexgen::opaque_type {};
+template<>
+struct fex_gen_type<VkExternalComputeQueueNV_T> : fexgen::opaque_type {};
 
 // Mark union types with compatible layout as such
 // TODO: These may still have different alignment requirements!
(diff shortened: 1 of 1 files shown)


lioncash (approved):

ThunkLibs/libvulkan/libvulkan_interface.cpp
```diff
@@ -2769,18 +2789,25 @@ template<>
...
-struct fex_gen_config<vkGetCudaModuleCacheNV> {};
-template<>
-struct fex_gen_config<vkCreateCudaFunctionNV> {};
-template<>
-struct fex_gen_config<vkDestroyCudaModuleNV> {};
-template<>
-struct fex_gen_config<vkDestroyCudaFunctionNV> {};
-template<>
-struct fex_gen_config<vkCmdCudaLaunchKernelNV> {};
+// These CUDA functions conflict with another extension.
```
neobrain, line 2792: What's this about? Merely having these function pointers shouldn't cause any issues?

Sonicadvance1: It does. Uncommenting them shows massive compile errors.

neobrain: Seems like the Vulkan headers changed how beta extensions are reported (?). DefinitionsExtract.py doesn't report those functions anymore, so you were trying to compile against definitions that no longer exist.

PR #5172, opened 2025-12-28
[cmake] Use a Find module for xxhash

crueter:
Port of #5159

Find modules are preferred for pkgconfig/otherwise non-CMake libraries.
Let's use that here for xxHash.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -303,6 +303,9 @@ set(CMAKE_LINKER_FLAGS_RELWITHDEBINFO "${CMAKE_LINKER_FLAGS_RELWITHDEBINFO} -fno
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fomit-frame-pointer")
 set(CMAKE_LINKER_FLAGS_RELEASE "${CMAKE_LINKER_FLAGS_RELEASE} -fomit-frame-pointer")
 
+## Modules ##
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):

CMakeLists.txt
```diff
@@ -303,6 +303,9 @@ set(CMAKE_LINKER_FLAGS_RELWITHDEBINFO "${CMAKE_LINKER_FLAGS_RELWITHDEBINFO} -fno
 set(CMAKE_CXX_FLAGS_RELEASE "${CMAKE_CXX_FLAGS_RELEASE} -fomit-frame-pointer")
 set(CMAKE_LINKER_FLAGS_RELEASE "${CMAKE_LINKER_FLAGS_RELEASE} -fomit-frame-pointer")
 
+## Modules ##
+list(APPEND CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/CMakeModules/)
```
neobrain, line 307: Could you move this folder to Data/CMake/Modules? Trying to avoid too many randomly placed top-level directories, and it's more sensible to keep this closer to our other CMake utilities.

crueter: A top-level CMakeModules directory is usually preferred for Find modules and toolchains, but I can move this and LinkerGC in a PR if needed

neobrain: Yeah, that convention probably makes sense for simpler projects, but our top-level got way too cluttered previously (to the point of hindering discoverability) with all the other things we juggle around. So a follow-up would be much appreciated!

PR #5179, opened 2025-12-29
CMake: Fix mingw if host has libxxhash-dev installed

Sonicadvance1:


Sonicadvance1:
Pushing this through to fix CI.

PR #5187, opened 2026-01-02
Accurate stack mapping, fixes #5149

### LINKED ISSUE #5149: FEX doesn't handle executable stack flags correctly.
Kernel documentation:
```
   *                 CPU: | lacks NX*  | has NX, ia32     | has NX, x86_64 |
   * ELF:                 |            |                  |                |
   * ---------------------|------------|------------------|----------------|
   * missing PT_GNU_STACK | exec-all   | exec-all         | exec-none      |
   * PT_GNU_STACK == RWX  | exec-stack | exec-stack       | exec-stack     |
   * PT_GNU_STACK == RW   | exec-none  | exec-none        | exec-none      |
```

FEX implementation:
```
   *                 CPU: | has NX, ia32     | has NX, x86_64 |
   * ELF:                 |                  |                |
   * ---------------------|------------------|----------------|
   * missing PT_GNU_STACK | exec-none        | exec-none      |
   * PT_GNU_STACK == RWX  | exec-all         | exec-all       |
   * PT_GNU_STACK == RW   | exec-none        | exec-none      |
```

Effectively if PT_GNU_STACK exists with RWX then we should only enable executable stack.
If PT_GNU_STACK /doesn't/ exist then we should enable exec-all on only x86, but leave it alone on x86-64.

Needs some unittests to ensure correct behaviours for the matrix of six behaviours (We don't emulate a system without NX these days).

Low prio as this isn't affecting anything we know of.

There's also the additional quirk that if an executable is mapped from a noexec filesystem that mappings should inherit that. Not sure if there is a clean way to detect that.... Might only matter with execve, not mmap?

svc64:
Also added unit tests for different `PT_GNU_STACK` permissions.

This is my first PR so hopefully code conventions and such are fine.

### DIFF
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -411,11 +411,7 @@ public:
 
     // Set the process personality here
     // Also, what about ADDR_LIMIT_3GB & co ?
-    uint32_t HostKernel = FEX::HLE::SyscallHandler::CalculateHostKernelVersion();
-    // In kernel 5.8+, executable stack doesn't mean that all memory is executable
-    if (-1 == personality(
-                PER_LINUX |
-                ((ExecutableStack && HostKernel < FEX::HLE::SyscallHandler::KernelVersion(5, 8, 0)) || ExecuteAll ? READ_IMPLIES_EXEC : 0))) {
+    if (-1 == personality(PER_LINUX | (ExecuteAll ? READ_IMPLIES_EXEC : 0))) {
       LogMan::Msg::EFmt("Setting personality failed");
       return false;
     }
--- a/unittests/FEXLinuxTests/tests/smc/smc-exec-stack.cpp
+++ b/unittests/FEXLinuxTests/tests/smc/smc-exec-stack.cpp
@@ -2,7 +2,6 @@
 #include <signal.h>
 #include <ucontext.h>
 #include <sys/mman.h>
-#include <sys/utsname.h>
 
 bool got_signal = false;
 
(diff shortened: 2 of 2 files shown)


Sonicadvance1:
Theoretically now that #5191 is merged, once you rebase this it should start passing CI.

svc64:
Anyways CI is passing now... I didn't test behavior on kernels older than 5.8, I don't have such machine.

Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -400,11 +403,19 @@ class ELFCodeLoader final : public FEX::CodeLoader {
...
     // Set the process personality here
-    // This needs some more investigation
-    // READ_IMPLIES_EXEC might be default for 32-bit elfs
     // Also, what about ADDR_LIMIT_3GB & co ?
-    if (-1 == personality(PER_LINUX | (ExecutableStack ? READ_IMPLIES_EXEC : 0))) {
+    uint32_t HostKernel = FEX::HLE::SyscallHandler::CalculateHostKernelVersion();
+    // In kernel 5.8+, executable stack doesn't mean that all memory is executable
+    if (-1 == personality(
+                PER_LINUX |
+                ((ExecutableStack && HostKernel < FEX::HLE::SyscallHandler::KernelVersion(5, 8, 0)) || ExecuteAll ? READ_IMPLIES_EXEC : 0))) {
```
Sonicadvance1: This check is unnecessary since our host-side kernel minspec is 5.15.

svc64: Makes sense, fixed

Sonicadvance1:
Thanks for fixing this weird edge case behaviour!

PR #5190, opened 2026-01-02
Windows: Invalidate code in freed memory after the free syscall

bylaws:
Windows will populate Size/Address with those of the underlying free region. This is important for cross-process free operations, where the the 'before' callback is still called after the memory is freed so querying the size of the section (which is not freed) fails.


WINDROID-EMU:
How can I compile a version of this to use in a project called micewine? It uses a modified version of Wine to work better on Android, and this project uses Box64, and I'm seeing that FEX performs better.

Sonicadvance1:
> How can I compile a version of this to use in a project called micewine? It uses a modified version of Wine to work better on Android, and this project uses Box64, and I'm seeing that FEX performs better.

https://wiki.fex-emu.com/index.php/Development:ARM64EC

PR #5197, opened 2026-01-05
Windows: Improve handling of RWX memory

bylaws:
- Fix a theoretical race
- Support issuing NtReadFile with RWX memory as a target safely


bylaws:
Needed to remove some abstraction due to infinitely blocking reads in some cases

lioncash (approved):

PR #5200, opened 2026-01-05
CMake: Move CMakeModules to Data/CMake

crueter:
See
https://github.com/FEX-Emu/FEX/pull/5172#pullrequestreview-3626560058.


neobrain:
Thanks for the follow-up!

PR #5204, opened 2026-01-07
FEXCore: Fixes circular dependency with thunk callback

Sonicadvance1:
Fixes crash in thunks that use callbacks, introduced in #5148.

The dispatcher would call the syscallhandler to get the VDSO thunk callback. But due to reordering initialization, the VDSO thunk would have not been loaded at that point. This would cause thunks that use callbacks to crash with a nullptr exception.

Instead, defer the thunk callback pointer loading until the thread starts executing, and load the pointer in to our thread state's pointer struct instead.

Didn't get caught in my initial test sweep since I didn't run a Wine game with thunks.


lioncash:
Seems like instcountci needs to be updated

lioncash:
Oh, nvm, sent the comment right as the PR was updated

PR #5205, opened 2026-01-07
FEXCore: Cleanup pointers structure 

Sonicadvance1:
There's no longer a distinction between AArch64 and x86 and everything
effectively falls under "Common" now. This means flattening the entire
structure just cleans it up.

NFC. (Although instcountCI will update because of a couple pointer
offsets changing)


lioncash (approved):
Nice!

PR #5208, opened 2026-01-08
Thunks/Vulkan: Fix beta extension symbol visibility

Sonicadvance1:


Sonicadvance1:
Pretty sure the problem here is that `VK_NV_cuda_kernel_launch` used to /not/ be a beta extension, but then NVIDIA moved it over to beta. Then instead promoting `VK_NVX_binary_import` out of beta, but still retaining the `NV_NVX` prefix instead of `VK_NV`.

These two extensions are nightmare fuel.

Sonicadvance1:
Pretty sure my original plan of not supporting any beta extensions still holds here, but this is a special weird one-off backtrack.

neobrain:
Considering you were fine with disabling these extensions without explicit note before, why don't we just leave all beta extensions (including this one) disabled?

As-is, this PR would complicate updating the interface files since it would make DefinitionsExtract spew out Vulkan functions we don't want, so I'm tentatively -1 on this unless there's a compelling reason to make an exception here.

PR #5210, opened 2026-01-09
LibraryForwarding: Redo Vulkan definitions update

neobrain:
https://github.com/FEX-Emu/FEX/commit/eb95a959bc8170686f509b9fb8d524bcf59a686a (https://github.com/FEX-Emu/FEX/pull/5171) did not keep the interface file in sync with the output of DefinitionExtract.py. Some definitions were reordered in the Vulkan headers, and diverging from that order makes future Vulkan header updates harder to diff and integrate.

There also is no "conflict" about CUDA extensions; instead, the previous update didn't reflect that the new headers fully hide beta extensions. (See also #5208)

Also refreshed the definitions used for 32-bit applications.


neobrain:
CC @lioncash : To note for future reviews of Vulkan header updates: DefinitionExtract output must be reproducible, and breaking it makes future updates needlessly involved.

Sonicadvance1 (approved):
I'll give this a test today.

neobrain:
Thanks for testing!

FYI git-diff supports highlighting moved lines with a separate color, using which it becomes apparent that the first two patches here are NFC.

PR #5211, opened 2026-01-09
Linux: Moves DRM LRU FD cache to be per-thread

Sonicadvance1:
This was a nasty race condition where each thread could be accessing the DRM cache at any given moment. Move it over to a per thread object that is only allocated once it gets used.

Also removes an `atexit` registration that contributes to crashing on exit.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
@@ -9,53 +9,18 @@ struct CpuStateFrame;
 }
 
 namespace FEX::HLE::x32 {
-class LRUCacheFDCache {
+class DRMLRUCacheFDCache final {
 public:
   using HandlerType = uint32_t (*)(FEXCore::Core::CpuStateFrame* Frame, int fd, uint32_t cmd, uint32_t args);
-  virtual ~LRUCacheFDCache() = default;
-  void SetFDHandler(uint32_t FD, HandlerType Handler) {
-    FDToHandler[FD] = Handler;
-  }
+  DRMLRUCacheFDCache();
+  void SetFDHandler(uint32_t FD, HandlerType Handler);
+  void DuplicateFD(int fd, int NewFD);
+  HandlerType FindHandler(int32_t FD);
+  uint32_t AddAndRunMapHandler(FEXCore::Core::CpuStateFrame* Frame, int fd, uint32_t cmd, uint32_t args);
 
-  void DuplicateFD(int fd, int NewFD) {
-    auto it = FDToHandler.find(fd);
-    if (it != FDToHandler.end()) {
-      FDToHandler[NewFD] = it->second;
-    }
-  }
-
-  HandlerType FindHandler(int32_t FD) {
-    HandlerType Handler {};
-    for (size_t i = 0; i < LRUSize; ++i) {
-      auto& it = LRUCache[i];
-      if (it.FD == FD) {
-        if (i == 0) {
-          // If we are the first in the queue then just return it
-          return it.Handler;
-        }
-        Handler = it.Handler;
-        break;
-      }
-    }
-
-    if (Handler) {
-      AddToFront(FD, Handler);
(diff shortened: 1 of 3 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
```diff
@@ -7,6 +7,16 @@ struct CpuStateFrame;
 }
 
 namespace FEX::HLE::x32 {
+class LRUCacheFDCache {
```
neobrain: Could you add a docstring to explain what this interface does and why it's called `*Cache*Cache`?

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
```diff
@@ -7,6 +7,16 @@ struct CpuStateFrame;
...
+public:
+  using HandlerType = uint32_t (*)(FEXCore::Core::CpuStateFrame* Frame, int fd, uint32_t cmd, uint32_t args);
+  virtual ~LRUCacheFDCache() = default;
```
neobrain: This doesn't actually need to be virtual, does it? There's only one implementation, and it only adds LRUSize/LRUCache/FDToHandler here, so we can just move it over.

Sonicadvance1: Moved as much to the header as possible without leaking DRM implementation details.

neobrain: This seems to have missed the intent of my comment. Instead of having two classes (one virtual, one "impl"), we can just have one. The function implementations themselves should stay in the cpp file (as opposed to their declarations).

Maybe I'm missing something that prevents all virtual from being dropped here, if so would appreciate taking a moment to explain.

Sonicadvance1: Indeed, there are some private DRM details in the `AddandRunHandler` that can't be exposed in the header.

neobrain: In the *implementation* of AddandRunHandler? So just declare the function in the header and implement it in the cpp file then?

Again I'm not suggesting moving any of the actual implementations into the header.

Sonicadvance1: Sure. Moved.

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
```diff
@@ -1,12 +1,73 @@
...
 namespace FEXCore::Core {
 struct CpuStateFrame;
 }
 
 namespace FEX::HLE::x32 {
+class DRMLRUCacheFDCache final {
+public:
+  using HandlerType = uint32_t (*)(FEXCore::Core::CpuStateFrame* Frame, int fd, uint32_t cmd, uint32_t args);
+  DRMLRUCacheFDCache();
+  virtual ~DRMLRUCacheFDCache() = default;
```
neobrain: Not needed anymore:
```suggestion
```

Source/Tools/LinuxEmulation/LinuxSyscalls/x32/IoctlEmulation.h
```diff
@@ -1,12 +1,73 @@
...
+  void SetFDHandler(uint32_t FD, HandlerType Handler) {
+    FDToHandler[FD] = Handler;
+  }
```
neobrain: I didn't propose moving the implementation to the header. Function implementations should stay in the cpp file:
```suggestion
  void SetFDHandler(uint32_t FD, HandlerType Handler);
```

(same for the other functions)

neobrain (commented):
Getting there. To reiterate, implementation details should stay in the cpp file.

PR #5213, opened 2026-01-10
FEXCore/Allocator: Remove unique_ptr behaviour from Alloc64

Sonicadvance1:
Even though we leak `Alloc64` on shutdown, that isn't good enough to avoid the `atexit` handler deallocating memory. So we need to use a raw pointer and leak it.

Removes an `atexit` registration that contributes to crashing on exit.

### DIFF
--- a/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
+++ b/FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
@@ -572,32 +572,23 @@ OSAllocator_64Bit::~OSAllocator_64Bit() {
   }
 }
 
-fextl::unique_ptr<Alloc::HostAllocator> Create64BitAllocator() {
-  return fextl::make_unique<OSAllocator_64Bit>();
+Alloc::HostAllocator* Create64BitAllocator() {
+  const auto size = sizeof(OSAllocator_64Bit);
+  const auto MinPage = FEXCore::AlignUp(size, FEXCore::Utils::FEX_PAGE_SIZE);
+
+  auto ptr = ::mmap(nullptr, MinPage, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  if (ptr == MAP_FAILED) {
+    ERROR_AND_DIE_FMT("Couldn't allocate memory region");
+  }
+
+  FEXCore::Allocator::VirtualName("FEXMem_Misc", reinterpret_cast<void*>(ptr), MinPage);
+
+  return ::new (ptr) OSAllocator_64Bit();
 }
-
-template<class T>
-struct alloc_delete : public std::default_delete<T> {
-  void operator()(T* ptr) const {
-    if (ptr) {
-      const auto size = sizeof(T);
-      const auto MinPage = FEXCore::AlignUp(size, FEXCore::Utils::FEX_PAGE_SIZE);
-
-      std::destroy_at(ptr);
-      ::munmap(ptr, MinPage);
-    }
-  }
-
-  template<typename U>
-  requires (std::is_base_of_v<U, T>)
-  operator fextl::default_delete<U>() {
-    return fextl::default_delete<U>();
-  }
-};
 
 template<class T, class... Args>
 requires (!std::is_array_v<T>)
-fextl::unique_ptr<T> make_alloc_unique(FEXCore::Allocator::MemoryRegion& Base, Args&&... args) {
+T* make_alloc(FEXCore::Allocator::MemoryRegion& Base, Args&&... args) {
   const auto size = sizeof(T);
   const auto MinPage = FEXCore::AlignUp(size, FEXCore::Utils::FEX_PAGE_SIZE);
   if (Base.Size < size || MinPage != FEXCore::Utils::FEX_PAGE_SIZE) {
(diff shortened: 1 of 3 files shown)


neobrain:
> Even though we leak Alloc64 on shutdown, that isn't good enough to avoid the atexit handler deallocating memory.

What exactly is happening there? Is this something specific about `unique_ptr`? Or does glibc somehow iterate over all pending allocations in atexit (why would it need to though?) ?

Sonicadvance1:
> > Even though we leak Alloc64 on shutdown, that isn't good enough to avoid the atexit handler deallocating memory.
> 
> What exactly is happening there? Is this something specific about `unique_ptr`? Or does glibc somehow iterate over all pending allocations in atexit (why would it need to though?) ?

Any static lifetime constructor member that allocates memory registers an atexit handler in order to destruct memory.
Static lifetime constructors will allocate memory before FEX has installed its hooks, once a host-side `atexit` handler is called, the handlers will try to deallocate memory and explode because now an incorrect allocator is trying to deallocate that memory.

We must remove all of FEX's global static constructors that allocate memory to mitigate the problem.

neobrain:
> Any static lifetime constructor member that allocates memory registers an atexit handler in order to destruct memory.

So global variables that have destructors are fine in principle, but if the constructor allocates memory it's not? But where would `std::unique_ptr`'s default constructor allocate memory? And why would its destructor try to de-allocate when we `release()` the associated pointer already?

Sonicadvance1:
> > Any static lifetime constructor member that allocates memory registers an atexit handler in order to destruct memory.
> 
> So global variables that have destructors are fine in principle, but if the constructor allocates memory it's not? But where would `std::unique_ptr`'s default constructor allocate memory? And why would its destructor try to de-allocate when we `release()` the associated pointer already?

In this PR's example:
`fextl::unique_ptr<Alloc::HostAllocator> Alloc64 {};`

This doesn't allocate memory, but it does register an `atexit` handler. This exit handler then deallocates memory that had been allocated from `SetupHooks`. This has the same failure mode why we have the workaround for with `Alloc::OSAllocator::ReleaseAllocatorWorkaround(std::move(Alloc64));`

In something like #5214,
```
static std::array<fextl::string, Paths::PATH_LAST> Paths;
static fextl::map<FEXCore::Config::LayerType, fextl::unique_ptr<FEXCore::Config::Layer>> ConfigLayers;
```

These also both register `atexit` handlers to free their contained data. Which will contain data from multiple VMA allocators. Once the exit handlers are called then things crash.

I've already removed all the static objects that up-front allocate over the last few months because it was causing spurious glibc fault CI failures depending on allocation order. But to ensure we don't have shutdown failure we need to ensure we don't have any atexit handlers that deallocate container memory. Once all of these are removed I'll be adding a CI validation step to ensure nothing registers an `atexit` handler.

neobrain:
Yes, I already understand the general problem you're trying to avoid, but it's not sufficiently clear to me why the `.release` call doesn't prevent it.

> In this PR's example:
> fextl::unique_ptr<Alloc::HostAllocator> Alloc64 {};
>
> This doesn't allocate memory, but it does register an atexit handler. This exit handler then deallocates memory that had been allocated from SetupHooks.

But we also dissociated the memory allocated in SetupHooks from the `Alloc64` object when calling ReleaseAllocatorWorkaround. By the time the exit handler runs, `Alloc64` is just a nullptr. Why would it try to deallocate memory?

Sonicadvance1:
> Yes, I already understand the general problem you're trying to avoid, but it's not sufficiently clear to me why the `.release` call doesn't prevent it.
> 
> > In this PR's example:
> > fextl::unique_ptrAlloc::HostAllocator Alloc64 {};
> > This doesn't allocate memory, but it does register an atexit handler. This exit handler then deallocates memory that had been allocated from SetupHooks.
> 
> But we also dissociated the memory allocated in SetupHooks from the `Alloc64` object when calling ReleaseAllocatorWorkaround. By the time the exit handler runs, `Alloc64` is just a nullptr. Why would it try to deallocate memory?

There's where the misunderstanding is happening.
If an application calls `exit` then the `atexit` handlers are called at the point of calling `exit` and our `release` function calls never occur! Because we never reach the end of main! atexit handlers work at the point of exit and also `main` returning!

neobrain:
Ah, I see.

Maybe it makes sense to add some LeakyPtr abstraction to ensure these things continue to be discoverable instead of just using raw pointers for them, what do you think? We can proceed with the raw pointers for now, and I can add a PR for the abstraction during some idle time later then.

Sonicadvance1:
> Ah, I see.
> 
> Maybe it makes sense to add some LeakyPtr abstraction to ensure these things continue to be discoverable instead of just using raw pointers for them, what do you think? We can proceed with the raw pointers for now, and I can add a PR for the abstraction during some idle time later then.

Sure. later PRs sound great.

FEXCore/Source/Utils/Allocator/64BitAllocator.cpp
```diff
@@ -572,32 +572,23 @@ OSAllocator_64Bit::~OSAllocator_64Bit() {
...
 
-  template<typename U>
-  requires (std::is_base_of_v<U, T>)
-  operator fextl::default_delete<U>() {
-    return fextl::default_delete<U>();
-  }
-};
+  FEXCore::Allocator::VirtualName("FEXMem_Misc", reinterpret_cast<void*>(ptr), MinPage);
+
+  return ::new (ptr) OSAllocator_64Bit();
```
neobrain, line 586: Why allocate a full page here instead of sticking closer to the original code?


```suggestion
  auto Storage = FEXCore::Allocator::aligned_alloc(alignof(OSAllocator_64Bit), sizeof(OSAllocator_64Bit));
  return ::new (Storage) OSAllocator_64Bit;
```

Sonicadvance1: It's reusing some of the code from below in `make_alloc_unique/make_alloc`. Which we would want this to be consistent between the two so that if/when we do deallocate at the end of main we know which allocator either one came from.

neobrain: I've been staring at this for a while but I don't get it - given that we just leak the pointer anyway, what does it matter what happens during deallocation? At least I'm not seeing any path that even tries to clean up the pointer now.

Sonicadvance1: > I've been staring at this for a while but I don't get it - given that we just leak the pointer anyway, what does it matter what happens during deallocation? At least I'm not seeing any path that even tries to clean up the pointer now.

```
 89 void ClearHooks() {
 90   SetJemallocMmapHook(::mmap);
 91   SetJemallocMunmapHook(::munmap);
 92   FEXCore::Allocator::mmap = ::mmap;
 93   FEXCore::Allocator::munmap = ::munmap;
 94
 95   Alloc::OSAllocator::ReleaseAllocatorWorkaround(std::move(Alloc64));
 96 }
 ```
 Attempts to clean it up but we need to leak it currently. One step at a time.

neobrain: If you want to take one step at a time, could you stick to the current allocation behavior and address only the issue that you've described?

As-is this looks like going half a step further, and it's unclear what direction it's taking.

neobrain (changes requested):

neobrain:

PR #5214, opened 2026-01-10
FEXCore/Config: Late initialize two objects

Sonicadvance1:
These need to be late initialized because they allocate memory and register an `atexit` handler to deallocate.
Leave the deallocation for our Shutdown handler.

Removes an `atexit` registration that contributes to crashing on exit.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.cpp
+++ b/FEXCore/Source/Interface/Config/Config.cpp
@@ -49,22 +49,24 @@ enum Paths {
   PATH_CONFIG_TELEMETRY_FOLDER,
   PATH_LAST,
 };
-static std::array<fextl::string, Paths::PATH_LAST> Paths;
+using PathsType = std::array<fextl::string, Paths::PATH_LAST>;
+static PathsType* Paths {};
+alignas(alignof(PathsType)) static char PathsPlacement[sizeof(PathsType)];
 
 void SetDataDirectory(const std::string_view Path, bool Global) {
-  Paths[PATH_DATA_DIR_LOCAL + Global] = Path;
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -234,19 +238,21 @@ void MetaLayer::MergeConfigMap(const LayerOptions& Options) {
 }
 
 void Initialize() {
+  Paths = new (PathsPlacement) PathsType {};
```
neobrain: Why does it need placement new here instead of plain `new`?

neobrain, line 241: Why does it need placement new here instead of plain new?

Sonicadvance1: plain new causes glibc allocation to occur, and this also gives a nicety of static allocation still.

neobrain: So we can just take FEXCore::Allocator or whatever it is that we normally use instead of `new`, right?

> a nicety of static allocation still

Does this translate to any practical benefit? I'd rather we keep the code as complex as needed only if not.

Sonicadvance1: Yes. Static allocation duration so we aren't mixing allocators.

neobrain: How would we be mixing allocators if we just used `FEXCore::Allocator` instead?

neobrain (changes requested):

neobrain:

PR #5215, opened 2026-01-10
FEXInterpreter: Removes three static objects

Sonicadvance1:
These three objects are easy enough to move out of the static namespace. They were causing atexit registrations for deallocating memory.

Removes three `atexit` registrations that contributes to crashing on exit.

### DIFF
--- a/Source/Tools/FEXInterpreter/FEXInterpreter.cpp
+++ b/Source/Tools/FEXInterpreter/FEXInterpreter.cpp
@@ -142,40 +142,44 @@ void Init() {
 } // namespace FEX::Logging
 
 namespace FEX::Allocator {
-static fextl::unique_ptr<FEX::HLE::MemAllocator> Allocator;
-static fextl::vector<FEXCore::Allocator::MemoryRegion> Base48Bit;
-static fextl::vector<FEXCore::Allocator::MemoryRegion> Low4GB;
 
-void Init(bool Is64Bit) {
+fextl::vector<FEXCore::Allocator::MemoryRegion> InitMemoryRegions(bool Is64Bit) {
   if (Is64Bit) {
     // Destroy the 48th bit if it exists
-    Base48Bit = FEXCore::Allocator::Setup48BitAllocatorIfExists();
-  } else {
-    // Reserve [0x1_0000_0000, 0x2_0000_0000).
-    // Safety net if 32-bit address calculation overflows in to 64-bit range.
-    constexpr uint64_t First64BitAddr = 0x1'0000'0000ULL;
-    Low4GB = FEXCore::Allocator::StealMemoryRegion(First64BitAddr, First64BitAddr + First64BitAddr);
-
-    // Setup our userspace allocator
-    FEXCore::Allocator::SetupHooks();
-    Allocator = FEX::HLE::CreatePassthroughAllocator();
-
-    // Now that the upper 32-bit address space is blocked for future allocations,
-    // exhaust all of jemalloc's remaining internal allocations that it reserved before.
-    // TODO: It's unclear how reliably this exhausts those reserves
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXInterpreter/FEXInterpreter.cpp
```diff
@@ -142,23 +142,26 @@ void Init() {
...
-
-void Init(bool Is64Bit) {
+struct AllocatorData {
+  fextl::unique_ptr<FEX::HLE::MemAllocator> Allocator;
+  fextl::vector<FEXCore::Allocator::MemoryRegion> Base48Bit;
+  fextl::vector<FEXCore::Allocator::MemoryRegion> Low4GB;
+};
+
+AllocatorData Init(bool Is64Bit) {
+  AllocatorData Data {};
```
neobrain: A struct is an unsuited means of grouping these objects since they have distinct lifetimes. At least `Allocator` should remain a separate object, and not sure if having `struct AllocatorData { vector Base48Bit, Low4GB; }` would be useful at that point.

Can we instead split off the memory region logic to a separate `ClaimReservedMemoryRegions()` function? This can just return a single `vector<MemoryRegion>` since Base48Bit and Low4GB are never used simultaneously anyway. That way, we only need to manage two objects in main then, which are independently destroyed.

Source/Tools/FEXInterpreter/FEXInterpreter.cpp
```diff
@@ -170,12 +173,13 @@ void Init(bool Is64Bit) {
     } while (reinterpret_cast<uintptr_t>(data) >> 32 != 0);
     free(data);
   }
+  return Data;
 }
 
-void Shutdown() {
+void Shutdown(AllocatorData& Data) {
```
neobrain: To force use of std::move at call-site:
```suggestion
void Shutdown(std::vector&& Data) {
```

This better highlights the lifetime end of this object.

Source/Tools/FEXInterpreter/FEXInterpreter.cpp
Sonicadvance1: Sure.

Source/Tools/FEXInterpreter/FEXInterpreter.cpp
```diff
@@ -142,40 +142,44 @@ void Init() {
...
+  do {
+    data = malloc(0x1);
+  } while (reinterpret_cast<uintptr_t>(data) >> 32 != 0);
+  free(data);
+
+  return Allocator;
 }
 
-void Shutdown() {
+void Shutdown(fextl::vector<FEXCore::Allocator::MemoryRegion>& MemoryRegions) {
```
neobrain: As outlined in the previously:
```suggestion
void Shutdown(fextl::vector<FEXCore::Allocator::MemoryRegion>&& MemoryRegions) {
```

Sonicadvance1:
```
/mnt/Work/Work/work/FEXNew/Source/Tools/FEXInterpreter/FEXInterpreter.cpp:626:28: error: rvalue reference to type 'vector<...>' cannot bind to lvalue of type 'vector<...>'
  626 |   FEX::Allocator::Shutdown(MemoryRegions);
      |                            ^~~~~~~~~~~~~
/mnt/Work/Work/work/FEXNew/Source/Tools/FEXInterpreter/FEXInterpreter.cpp:180:65: note: passing argument to parameter 'MemoryRegions' here
  180 | void Shutdown(fextl::vector<FEXCore::Allocator::MemoryRegion>&& MemoryRegions) {
      |                                                                 ^
```
      
???

Sonicadvance1: I guess std::move

neobrain (approved):

PR #5216, opened 2026-01-10
CodeCache: Remove an assumption about git hash length

Sonicadvance1:
Our short hash is typically 7-characters in length, but due to hash conflicts it can be more than that.

Remove a single assumption that the hash size is 8-bytes.

Gets the case working that the hash length is 8-characters, but doesn't attempt going further than that. As this fixes the compiler error I had due to a hash conflict.

### DIFF
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -351,8 +351,8 @@ bool CodeCache::LoadData(Core::InternalThreadState* Thread, std::byte* MappedCac
     return false;
   }
 
-  char ExpectedVersion[8] = GIT_SHORT_HASH;
-  ranges::fill(ranges::find(ExpectedVersion, 0), std::end(ExpectedVersion), 0);
+  constexpr std::string_view ExpectedVersion = GIT_SHORT_HASH;
+  static_assert(ExpectedVersion.size() <= sizeof(header.FEXVersion));
   if (!ranges::equal(header.FEXVersion, ExpectedVersion)) {
     LogMan::Msg::IFmt("Cache generated from old FEX version {}, current is {}; skipping", fmt::join(header.FEXVersion, ""),
                       fmt::join(ExpectedVersion, ""));


FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -351,8 +351,8 @@ bool CodeCache::LoadData(Core::InternalThreadState* Thread, std::byte* MappedCac
     return false;
   }
 
-  char ExpectedVersion[8] = GIT_SHORT_HASH;
-  ranges::fill(ranges::find(ExpectedVersion, 0), std::end(ExpectedVersion), 0);
+  constexpr std::string_view ExpectedVersion = GIT_SHORT_HASH;
+  static_assert(ExpectedVersion.size() <= sizeof(header.FEXVersion));
   if (!ranges::equal(header.FEXVersion, ExpectedVersion)) {
```
neobrain, line 356: This will fail now if `GIT_SHORT_HASH` is 7 characters or fewer, since FEXVersion is always 8 bytes.

Sonicadvance1: I think you read the comparison backwards. It's ensuring the GIT_SHORT_HASH is always smaller than the 8-byte `FEXVersion`.

```
generated/git_version.h
4:#define GIT_SHORT_HASH "1"
```
by overriding hash works.

Sonicadvance1: Meanwhile:
```
generated/git_version.h
4:#define GIT_SHORT_HASH "123456789"
```
If I override to longer then I get the static_assert.

neobrain: The comment is on the existing `ranges.equal` call, not the new `static_assert`. In your first example `"1"`, ExpectedVersion will be a 1-element container whereas `ExpectedVersion` has 8 entries, so `ranges::equal` will fail at runtime.

Sonicadvance1:
I don't know enough about C++ ranges to solve this without making the code ugly with C. So I'll just leave this alone.

PR #5217, opened 2026-01-10
FEX: Move SBRK handling to the frontend

Sonicadvance1:
This is fundamentally a frontend only problem, and also Linux only. Moves it to the frontend where it belongs.

There's likely more things in Allocator.cpp that can be moved to the frontend but this is the first thing.

NFC

### DIFF
--- a/FEXCore/Source/Utils/Allocator.cpp
+++ b/FEXCore/Source/Utils/Allocator.cpp
@@ -72,54 +72,6 @@ int FEX_munmap(void* addr, size_t length) {
   return Result;
 }
 
-// This function disables glibc's ability to allocate memory through the `sbrk` interface.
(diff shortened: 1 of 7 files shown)


neobrain:
We already have 5 files called `*Allocator.cpp` without clear distinction, plus headers. Can we rename this one to something like SBRKAllocations.cpp so that people don't need to check 6 files to find what they're looking for?

Sonicadvance1:
> We already have 5 files called `*Allocator.cpp` without clear distinction, plus headers. Can we rename this one to something like SBRKAllocations.cpp so that people don't need to check 6 files to find what they're looking for?

Moved.

neobrain (approved):

Issue #5206, opened 2026-01-08
User Mode Linux stuck before actually launched


fish4terrisa-MSDSM:
# Problem
I'm testing the new introduced seccomp support in user mode linux on FEX-emu.
I know that currently FEX-emu's seccomp is really poor at this time, and the program wont due to ptrace is not supported and seccomp emulation is heavily unfinished.
However, when working with previous versions of user mode linux, it can output the help message just fine, and will fail at the step of checking if ptrace can change syscall number(as expected this failes)
But with 6.18.2 version of user mode linux, it cannot lanuch at all. The program is stuck with no iutput and no help message is printed.
I made sure that the libs used on host is the same in FEX's rootfs.
This is the strace log:

[vmlinux.log](https://github.com/user-attachments/files/24499722/vmlinux.log)

It seems that it's looping so I interrupted after seen four same output.

[vmlinux.txt](https://github.com/user-attachments/files/24499794/vmlinux.txt)

This is the binary I used(It's a executable), tho if you dont trust it you can just compile it yourself. Mine is from the package `linux-usermode` in AUR, compiled on my laptop(with newest libs)

# What I expected
At least being able to output help message. Since nothing is checked or ran when printing help message, it should be expected to work even under FEX, and is working in previous versions of user mode linux.
Tested in `box64` and `qemu-user`, both at least succeeded in printing the help message, which indicate that this is a problem in `FEX` side.

fish4terrisa-MSDSM:
Btw it'll be kinda interesting if the seccomp emulation can be fully finished which allows the users to run user mode linux in FEX.

fish4terrisa-MSDSM:
This is a static build and it shares the same problem.

[kernel64.txt](https://github.com/user-attachments/files/24500006/kernel64.txt)

This is the config of the kernel:

[config.txt](https://github.com/user-attachments/files/24500046/config.txt)

The source code is official linux 6.18.2 source tarball.
No patches are applied to the source code.

Sonicadvance1:
With the latest PRs merged, it at least gets as far as my desktop
```bash
$ FEX_NEEDSSECCOMP=1 ./Bin/FEX /mnt/Work/Projects/linux seccomp=on
Core dump limits :
        soft - NONE
        hard - NONE
Checking environment variables for a tempdir...none found
Checking if /dev/shm is on tmpfs...OK
Checking PROT_EXEC mmap in /dev/shm...OK
Checking that seccomp filters can be installed...OK
[...]

PR #5218, opened 2026-01-10
Linux: Fixes personality inheritence

Sonicadvance1:
UML disables ASLR using a personality. FEX in the ELFCodeLoader was accidentally erasing personalities. This caused UML to do a bootloop trying to repeat disable ASLR.

It now gets ASLR disabled, but fails shortly afterwards.

For #5206

### DIFF
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -411,7 +411,9 @@ public:
 
     // Set the process personality here
     // Also, what about ADDR_LIMIT_3GB & co ?
-    if (-1 == personality(PER_LINUX | (ExecuteAll ? READ_IMPLIES_EXEC : 0))) {
+    uint32_t Personality = personality(~0ULL);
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -411,15 +411,21 @@ class ELFCodeLoader final : public FEX::CodeLoader {
 
     // Set the process personality here
     // Also, what about ADDR_LIMIT_3GB & co ?
-    if (-1 == personality(PER_LINUX | (ExecuteAll ? READ_IMPLIES_EXEC : 0))) {
+    uint32_t Personality = personality(~0ULL);
+    if (ExecuteAll) {
+      Personality |= READ_IMPLIES_EXEC;
+    } else {
+      Personality &= ~READ_IMPLIES_EXEC;
```
svc64: Why explicitly remove READ_IMPLIES_EXEC?

Sonicadvance1: You're right, nothing implying it should get removed. Would break inheritance over execve.

PR #5220, opened 2026-01-11
CI: Update tunables for artifact builds

Sonicadvance1:
These are what we want them to be.


Sonicadvance1:
I think /technically/ the arm64ec/wow64 builds ignore the flags, but good to set it explicitly anyway.

PR #5226, opened 2026-01-13
JIT: Ensure code cache consistency by zero-initializing padding bytes

neobrain:
Code caches are required to produce consistent results across runs (given unchanged input code maps), so this padding data must be initialized explicitly.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/JIT.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/JIT.cpp
@@ -1083,12 +1083,11 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
 
       // NOTE: 16-byte alignment of the new cursor offset must be preserved for block linking records
       SetBuffer(CurrentCodeBuffer->Ptr, CurrentCodeBuffer->AllocatedSize);
-      SetCursorOffset(AlignUp(CodeBuffers.LatestOffset, 16));
+      SetCursorOffset(CodeBuffers.LatestOffset);
+      Align16B();
       if ((GetCursorOffset() + TempSize) > CurrentCodeBuffer->UsableSize()) {
         CTX->ClearCodeCache(ThreadState);
       }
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/Core/JIT/JIT.cpp
```diff
@@ -1003,12 +1003,12 @@ CPUBackend::CompiledCode Arm64JITCore::CompileCode(uint64_t Entry, uint64_t Size
   const auto JITBlockTailLocation = GetCursorAddress<uint8_t*>();
   CodeHeader->OffsetToBlockTail = JITBlockTailLocation - CodeData.BlockBegin;
 
-  JITCodeTail JITBlockTail {
```
bylaws, line 1006: Wait huh should this not zero init everything?

neobrain: Only members, not padding bytes :/

bylaws: Ahh, thinking about it I wonder if including padding explicitly in this struct kinda makes sense? Especially if this is ending up in generated code, I find myself reading it manually with gdb on the occasion and explicit padding helps there and we already do for ThreadState. 

(Idm the change as is, but this occurred to me, up to you)

Sonicadvance1: I also would like us to describe the padding members for serialized structs. We already have a `_Pad` member inside of it, just add 4 to the array.

Sonicadvance1: This change chunk actually looks like it is entirely unnecessary, we already fully pad the JITCodeTail to be 40-bytes.

neobrain: Correct, I think some other FEX changes rendered this part of the patch obsolete after I originally wrote the code.

PR #5228, opened 2026-01-13
Linux: Updates minimum altstack requirements

Sonicadvance1:
Turns out the Linux kernel's definition of `MINSIGSTKSZ` and glibc's definition of `MINSIGSTKSZ` don't match.

The linux kernel has a massive comment about it in `arch/x86/kernel/signal.c`.

Also adds a unittest for it.

For #5206


lioncash:
Said [comment](https://github.com/torvalds/linux/blob/master/arch/x86/kernel/signal.c#L399) if anyone is interested.

neobrain:
If anyone is interested *and* from the future, here is a permalink to the comment: https://github.com/torvalds/linux/blob/24d479d26b25bce5faea3ddd9fa8f3a6c3129ea7/arch/x86/kernel/signal.c#L399

PR #5231, opened 2026-01-14
FEX: Pass VA/PA to CPUID and have cpuinfo populate from cpuid

Sonicadvance1:
Previously we were hardcoding 48-bit, which is reasonable. Query the VA space from the frontend and pass it back to CPUID instead. Then populate cpuinfo information from CPUID.


bylaws:
I'm not sure I think diverging from standard x86 by default is ideal here.

Sonicadvance1:
> I'm not sure I think diverging from standard x86 by default is ideal here.

You right.

PR #5233, opened 2026-01-14
CMake: Use a Find module for Zydis/Zycore

crueter:
Much like the prior xxhash PR. This one also has the advantage of making
it completely trivial to handle distros that ONLY ship zy{dis,core}-config.cmake, NOT the pkgconfig files (namely Gentoo and Arch Linux, I'm sure others do too)


pmatos:
This was what I originally implemented but according to @neobrain this is not a good idea as some distros don't ship cmake files.

crueter:
> This was what I originally implemented but according to @neobrain this is not a good idea as some distros don't ship cmake files.

That's why I have the fallback:

```cpp
find_package(Zycore QUIET CONFIG)

if (Zycore_CONSIDERED_CONFIGS)
    find_package_handle_standard_args(Zycore CONFIG_MODE)
else()
```

Basically if it finds a `zycore-config.cmake`, it will set `Zycore_CONSIDERED_CONFIGS` to `ON` and then `find_package_handle_standard_args` will pass that to the caller. If it doesn't find a `zycore-config.cmake` it sets `Zycore_CONSIDERED_CONFIGS` to `OFF` and goes through the standard pkgconfig procedure.

I tested on a distro without CMake files for zy* and it worked fine, though of course more tests on Fedora and the like would be nice.

neobrain:
> distros that also install zy{dis,core}-config.cmake alongside the PkgConfig files

Are there distributions that *don't* ship the pkgconfig files but only cmake ones? Not sure there's much upside in supporting the latter if we need the path for pkgconfig either way.

Would prefer to keep the configuration matrix small unless there's a compelling argument for supporting the cmake files here.

crueter:
> > distros that also install zy{dis,core}-config.cmake alongside the PkgConfig files
> 
> Are there distributions that _don't_ ship the pkgconfig files but only cmake ones? Not sure there's much upside in supporting the latter if we need the path for pkgconfig either way.
> 
> Would prefer to keep the configuration matrix small unless there's a compelling argument for supporting the cmake files here.

I actually slightly misspoke before. Behold, Gentoo:

```
$ equery f zycore-c | grep pkgconfig
$ equery f zydis | grep pkgconfig 
```

It looks like the pkgconfig files are ONLY installed if your distribution uses the Meson build, not CMake. <https://github.com/search?q=repo%3Azyantific%2Fzycore-c%20pkgconfig&type=code>

Arch Linux as well: <https://archlinux.org/packages/extra/x86_64/zydis/files/>

CMakeLists.txt
```diff
@@ -345,17 +345,11 @@ if (BUILD_TESTING OR ENABLE_VIXL_DISASSEMBLER OR ENABLE_VIXL_SIMULATOR)
...
 
-find_package(PkgConfig REQUIRED)
-
 if (ENABLE_ZYDIS)
-  pkg_search_module(Zydis QUIET IMPORTED_TARGET zydis)
-  pkg_search_module(Zycore QUIET IMPORTED_TARGET zycore)
-  if (TARGET PkgConfig::Zydis AND TARGET PkgConfig::Zycore AND NOT CMAKE_CROSSCOMPILING)
-    add_library(Zydis::Zydis ALIAS PkgConfig::Zydis)
-    add_library(Zycore::Zycore ALIAS PkgConfig::Zycore)
-    message(STATUS "Using system Zydis")
```
neobrain, line 356: Could you restore these logs?

crueter:
```cmake
cmake_minimum_required(VERSION 3.22)

project(test)

find_package(Zycore REQUIRED)
message(STATUS "ZYCORE FOUND: ${Zycore_FOUND} lowercase: ${zycore_FOUND}")
if (NOT TARGET Zycore::Zycore)
    message(FATAL_ERROR "Zycore::Zycore doesn't exist")
else()
    message(STATUS "Zycore found")
endif()

find_package(Zydis REQUIRED)
message(STATUS "ZYDIS FOUND: ${Zydis_FOUND} lowercase: ${zydis_FOUND}")
if (NOT TARGET Zydis::Zydis)
    message(FATAL_ERROR "Zydis::Zydis doesn't exist")
else()
    message(STATUS "Zydis found")
endif()
```

This works as expected:

```
-- ZYCORE FOUND: 1 lowercase: TRUE
-- Zycore found
-- ZYDIS FOUND: 1 lowercase: TRUE
-- Zydis found
```

However, transplanting those same calls into FEX doesn't:

```
CMake Error at CMakeLists.txt:348 (find_package):
  Found package configuration file:

    /usr/lib64/cmake/zycore/zycore-config.cmake

  but it set Zycore_FOUND to FALSE so package "Zycore" is considered to be
  NOT FOUND.
```

Weird case-sensitivity shenanigans I guess. :/

crueter:
PS. do not merge until tests on pkgconfig-only distros like Fedora are conducted. Will get a Fedora VM set up as soon as I stop being lazy and put in my new SSD but couldn't hurt to get some double checks

crueter:
Tested and working on:

- Ubuntu 22.04 (CMake)
- Arch Linux (CMake)
- Gentoo Linux (CMake)
- Fedora (pkgconfig) - with a small change

Truthfully I literally have no idea what could possibly be going on with these case sensitivity shenanigans, and after debugging for several hours I've got nothing. So I'm just going to leave it like this and pray the Gentoo packagers don't eat me.

CMakeLists.txt
```diff
@@ -345,20 +345,32 @@ if (BUILD_TESTING OR ENABLE_VIXL_DISASSEMBLER OR ENABLE_VIXL_SIMULATOR)
...
-  else()
-    message(STATUS "Using bundled Zydis")
+  find_package(zycore MODULE)
+
+  # For some grand reason, the Zycore find_dependency is case-sensitive here...
+  # (usually it isn't). So to work around this we have to do this awful hack and
+  # "transfer" the case
+  if(zycore_FOUND AND NOT Zycore_FOUND)
```
neobrain: I don't see anything in the CMake documentation suggesting `Zycore_FOUND` would ever be set (or checked in subsequent `find_package` calls). Did you try with a clean build directory?

neobrain:
> Truthfully I literally have no idea what could possibly be going on with these case sensitivity shenanigans, and after debugging for several hours I've got nothing. So I'm just going to leave it like this and pray the Gentoo packagers don't eat me.

If we can't get the CMake stuff working reliable without resorting to silly hacks like this, I think we're better off sticking to using pkgconfig files when available and falling back to the bundled module if not. There's a trade-off to be considered how much developer time is worth spending to support even the last niche of the configuration matrix out there, particularly when it's about a dependency only used by developers.

CMakeLists.txt
crueter: Zydis' `find_package` looks for `find_dependency(Zycore)`, which ONLY succeeds if `Zycore_FOUND` is set, but Zycore's config only sets `zycore_FOUND`. Normally it sets the capitalized version first but not in this instance for God knows what reason.

neobrain: > but Zycore's config only sets zycore_FOUND

Doesn't `find_package(SOMETHING)` itself set `SOMETHING_FOUND` irrespectively of the contents of the config file? If not, sounds like an upstream bug that should be fixed there, not worked around in FEX.

Upstream uses `find_package(Zycore)` in its main CMakeLists.txt, so clearly that's what they intend to work. We should stick to that recommendation.

crueter:
I'll try a few more things and drop the CMake support if absolutely nothing else works.

CMakeLists.txt
crueter: I don't think it's an upstream bug of any sort. As I posted earlier it works totally fine normally no matter the case, but no matter what I couldn't get it to work properly in FEX because it wasn't propagating those variable cases.

Could try a few more things

crueter:
Awesome, looks like all of my troubles may have actually just been caused by a stale build directory...

crueter:
Horrific hack was caused by my skill issue and has been removed. CMake and PkgConfig are both working as expected, also added a minimum version check since 4.0 is required (and Ubuntu 22.04 doesn't have 4.0).

CMakeLists.txt
```diff
@@ -345,20 +345,23 @@ if (BUILD_TESTING OR ENABLE_VIXL_DISASSEMBLER OR ENABLE_VIXL_SIMULATOR)
...
+
     set(ZYDIS_BUILD_TOOLS OFF CACHE BOOL "" FORCE)
     set(ZYDIS_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+
+    message(STATUS "Using bundled Zydis")
     add_subdirectory(External/zydis/)
+  else()
+    message(STATUS "Using system Zydis")
   endif()
 endif()
```
neobrain: Splitting up the configuration matrix to support scenarios like "system zycore but bundled zydis" seems silly. We're never going to realistically test this configuration (nor would I expect upstream to test strange version couplings) and it's just inviting future niche problems. Let's either use both bundled or both from the system:

```suggestion
  if (TARGET Zydis::Zydis AND TARGET Zycore::Zycore)
    message(STATUS "Using system Zydis")
  else()
    message(STATUS "Using bundled Zydis")
    set(ZYDIS_BUILD_TOOLS OFF CACHE BOOL "" FORCE)
    set(ZYDIS_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
    add_subdirectory(External/zydis/)
  endif()
endif()
```

PR #5235, opened 2026-01-14
Github: Enforce expected amount of cxa_atexit registrations

Sonicadvance1:
On top of the previous PRs.

There is only one registration that I've not been able to track down why
it happens, but it is something to do with std::ostream. This just
ensures in CI that we don't ever register anymore atexit handlers so we
don't regress on crashing on `exit`.


.github/workflows/ccpp.yml
```diff
@@ -233,6 +233,11 @@ jobs:
       working-directory: ${{runner.workspace}}/build
       run: mv ${{runner.workspace}}/build/Testing/Temporary/LastTest.log ${{runner.workspace}}/build/Testing/Temporary/LastTest_StructVerifier.log || true
 
+    - name: Ensure expected atexit registrations
+      working-directory: ${{runner.workspace}}/build
+      shell: bash
+      run: $GITHUB_WORKSPACE/Scripts/CountInstancesOfAtExit.py ${{runner.workspace}}/build/Bin/FEX "__cxa_atexit@plt>$" 1 ; echo $?
```
neobrain: Why not just define the commands inline here? Moving things to an extra script just makes this way more complicated to trace through.

```suggestion
run: |
  COUNT = `llvm-objdump -D thefile | grep thepattern | wc -l`
  if [ $COUNT -eq 123 ]
    echo "Expected 123 atexit handlers, found $COUNT"
    exit 1
  fi 
```

Sonicadvance1: Because I don't know bash scripting. So sure, I can switch it over.

Sonicadvance1:
Looks like the bash implementation doesn't work. I don't know bash well enough to know what's wrong. :shrug:

.github/workflows/ccpp.yml
```diff
@@ -233,6 +233,16 @@ jobs:
...
 
+    - name: Ensure expected atexit registrations
+      working-directory: ${{runner.workspace}}/build
+      shell: bash
+      run: |
+        COUNT = `llvm-objdump -D ${{runner.workspace}}/build/Bin/FEX | grep "__cxa_atexit@plt>$" | wc -l`
+        if [ $COUNT -eq 1 ]
+          echo "Expected 1 atexit handlers, found $COUNT"
+          exit 1
+        fi
```
neobrain:
```suggestion
        COUNT=`llvm-objdump -D ${{runner.workspace}}/build/Bin/FEX | grep '__cxa_atexit@plt>\\$' | wc -l`
        if [ $COUNT -ne 1 ]
          echo "Expected 1 atexit handlers, found $COUNT"
          exit 1
        fi
```
Something to this effect should be good. Otherwise you might need to open a terminal and run it once to find the last typo.

.github/workflows/ccpp.yml
```diff
@@ -233,6 +233,17 @@ jobs:
       working-directory: ${{runner.workspace}}/build
       run: mv ${{runner.workspace}}/build/Testing/Temporary/LastTest.log ${{runner.workspace}}/build/Testing/Temporary/LastTest_StructVerifier.log || true
 
+    - name: Ensure expected atexit registrations
+      working-directory: ${{runner.workspace}}/build
```
crueter:
```suggestion
COUNT=$(llvm-objdump -D build/Bin/FEX | grep '__cxa_atexit@plt>\\$' | wc -l)
[ "$COUNT" -eq 1 ] || { echo "Expected 1 atexit handlers, found $COUNT"; exit 1; }
```

Slightly more concise. Also ensure to remove the redundant working-directory and shell directives (cc. #5243 and #5245)

neobrain (changes requested):
(pending on the other PRs that are stalled; marking as "requesting changes" to help distinguish this from active PRs in our crowded PR list)

neobrain:

PR #5236, opened 2026-01-16
Switch over to rpmalloc instead of jemalloc

Sonicadvance1:
Just a rebase of the earlier PR (#4969). Haven't investigated why Asahi devices hate this.

---

rpmalloc is currently very aggressively configured which causes significant reductions in resident memory over jemalloc.

In Bayonetta's title screen it went from 963MB down to 834MB resident.

### DIFF
--- a/docs/allocator_usage.md
+++ b/docs/allocator_usage.md
@@ -1,9 +1,9 @@
-# JEMalloc usage
-FEX-Emu uses two different jemalloc heap allocators at once, each for different purposes:
-- jemalloc: The primary heap allocator (to keep FEX's internal allocations out of the 32-bit address space used by guest applications)
+# Dual allocator usage
+FEX-Emu uses two different heap allocators at once, each for different purposes:
+- rpmalloc: The primary heap allocator (to keep FEX's internal allocations out of the 32-bit address space used by guest applications)
 - jemalloc_glibc: The second heap allocator (to add allocation introspection features used by thunks)
 
-## jemalloc - primary heap allocator
+## rpmalloc - primary heap allocator
 This allocator overrides `mmap` and `munmap` by forwarding them to FEXCore's internal VMA region allocator.
 
 All of FEXCore's `fextl::` namespaced objects allocate memory with this method.


Sonicadvance1:
Looks like the asahi installer completely barfs on my partition table. Tried installing over a dead Fedora 40 install that no longer had the ability to update, and now it doesn't know how to install correctly. So doing some Asahi testing is going to be difficult.
`Bootability failed Error Domain=BYErrorDomain Code=112 "Preserved restore bundle in preboot is missing" UserInfo={BYErrorHint=VolumeInvalid, NSLocalizedDescription=Preserved restore bundle in preboot is missing}`
Screenshot 2026-01-16 at 13 04 48

Sonicadvance1:
Got Asahi installed. Tested Steam, Linux-native games through SLR, and Proton games through PV/SLR.
Couldn't reproduce @neobrain's previous problem, so it seems like it is working fine.

Now that I have an asahi install, if the problem crops up I should be able to reproduce it.

CMakeLists.txt
```diff
@@ -309,12 +309,12 @@ elseif (NOT MINGW)
...
-if (ENABLE_JEMALLOC)
-  # The jemalloc subproject that all FEXCore fextl objects allocate through.
-  add_subdirectory(External/jemalloc/)
-elseif (NOT MINGW)
-  message(STATUS
-    " jemalloc disabled!\n"
+if (ENABLE_FEX_ALLOCATOR)
+  # The rpmalloc subproject that all FEXCore fextl objects allocate through.
+  add_subdirectory(External/rpmalloc/)
+elseif (NOT MINGW_BUILD)
```
crueter: Rebase fail?

Sonicadvance1: Oops, fixed.

neobrain:
Thanks for updating the docs!

Sadly this change still consistently makes Steam fail to launch here (showing the typical "steamwebhelper is not responding" dialog), whereas both latest main and the base of this branch work fine.

muvm was restarted between tries (since it won't pick up FEX updates otherwise). Tried with a fresh build folder as well (`CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=OFF -DUSE_LINKER=lld -DENABLE_LTO=False -DENABLE_ASSERTIONS=False -G Ninja ..`). I also removed any traces of the GuestThunks/HostThunks folder just to make sure there was no stale VDSO version acting up or similar.

Sonicadvance1:
> Thanks for updating the docs!
> 
> 
> 
> Sadly this change still consistently makes Steam fail to launch here (showing the typical "steamwebhelper is not responding" dialog), whereas both latest main and the base of this branch work fine.
> 
> 
> 
> muvm was restarted between tries (since it won't pick up FEX updates otherwise). Tried with a fresh build folder as well (`CC=clang CXX=clang++ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTING=OFF -DUSE_LINKER=lld -DENABLE_LTO=False -DENABLE_ASSERTIONS=False -G Ninja ..`). I also removed any traces of the GuestThunks/HostThunks folder just to make sure there was no stale VDSO version acting up or similar.
> 
> 

Weird, seems to be working fine here. image

neobrain:
Presumably you're running your own RootFS on Fedora 42?

I already updated to F43 and I'm using the Fedora RootFS, so maybe that would help explain the difference.

Sonicadvance1:
> Presumably you're running your own RootFS on Fedora 42?
> 
> 
> 
> I already updated to F43 and I'm using the Fedora RootFS, so maybe that would help explain the difference.

Testing was done on Fedora 43 with FEX's Fedora 43 rootfs image.

Sonicadvance1:
Actually, double checked. I was on Fedora 42 but with FEX's Fedora 43 rootfs.

Sonicadvance1:
Updated to Fedora 43 and Steam is still working here with the FEX official rootfs.
But really according to the previous message you had in the revert PR, I don't think guest rootfs really mattered once PV was in the mix and couldn't find the guest application's interpreter.

Sonicadvance1:
Oh hey, I switched over to Fedora's rootfs and reproduced the problem. Finally progress.

Sonicadvance1:
Reproduced the error on my non-asahi platform as well by binding a rootfs to `/run/fex-emu/rootfs`, same rootfs configured to a path in my $HOME directory is fine. So that's weird.

Sonicadvance1:
Found the bug. It's unrelated to rpmalloc but was exposed /because/ of rpmalloc.

Sonicadvance1:
This PR is now cherry-picking the commit from #5288 and is now working.

neobrain (approved):
Confirmed working on my Asahi install as well now, cheers!

Sonicadvance1:
....Apparently "fixes" also closes PRs, derpy github.

PR #5241, opened 2026-01-17
CI: Merge InstCountCI Diff steps into one

crueter:
Uses `git --no-pager diff --exit-code HEAD` instead of the two-step check used before. Also shows the diff if it exists so you can see changes at a quick glance.


Sonicadvance1:
Merge conflict here now.

crueter:
Figgsed

PR #5245, opened 2026-01-19
CI: Refactor C/C++ tests to use a composite action

crueter:
Basically reduces all that boilerplate of run test target... move log...
blah blah down to a single composite action that handles it on its own.
Mostly just makes maintenance a little easier as adding tests is
trivial. Will move to others later.

Currently, this relies on GitHub's utterly horrific environment
inheritance garbage that is both undocumented and nonsensical, but it
should work (TM) pending some further tests and 40 attempts at getting
it right (sorry in advance to your runners Ryan).

Some todos:
- should each one run always, or should tests stop whenever a previous
  one fails?
- ENVIRONMENT INHERITANCE
- Port this to other actions

### DIFF
--- a/.github/workflows/test/action.yml
+++ b/.github/workflows/test/action.yml
@@ -32,7 +32,7 @@ runs:
 
     - name: Move results
       if: ${{ always() }}
-      run: mv "$LAST_LOG".log "$LAST_LOG"_GCC32.log || true
+      run: mv "$LAST_LOG".log "$LAST_LOG"_"${{ inputs.log-name }}".log || true
       shell: bash
       env:
         LAST_LOG: build/Testing/Temporary/LastTest


.github/workflows/test/action.yml
```diff
@@ -0,0 +1,39 @@
...
+    - name: Run tests
+      run: cmake --build build --config ${{ env.BUILD_TYPE }} --target ${{ inputs.target }}
+      shell: bash
+      env:
+        FEX_PORTABLE: ${{ inputs.portable }}
+        DISPLAY: ${{ inputs.display }}
+
+    - name: Move results
+      if: ${{ always() }}
+      run: mv "$LAST_LOG".log "$LAST_LOG"_GCC32.log || true
```
Sonicadvance1: Always moved to GCC32?

crueter: good catch!

neobrain:
> crueter:runner-workspace-aaaaaaaaaa

Branch names show up in git logs, so please stick to constructive branch names. "Funny" names become obnoxious when trying to look up something in commit history. (This is one of the milder cases, just mentioning it for awareness)

Besides that, I welcome the notion of fixing all the duplication in our workflows!

crueter:
I've become spoiled by squashing. :P

Issue #5242, opened 2026-01-19
CI cleanups


crueter:
- [ ] defaults.run.shell - I'm opening a PR to set this to bash, but is this even necessary? None of the actions use Windows, and the GHA runner for Linux *should* default to bash
- [ ] runner.workspace redundancy - you generally don't need to specify `${{ runner.workspace }}` everywhere
- [ ] `working-directory`: you can just do `cmake --build build` for instance
- [ ] variables - long file names and such can just be extracted to variables
- [ ] update deps - some of the action deps are outdated. note that this depends on having nodejs 24 for some.
- [ ] placeholder for more stuff

crueter:
Pretty much done with these, anything hereafter will just come if I or others get a hankering for cleanup.

PR #5246, opened 2026-01-19
CI: Composite testing action and general clean ups

crueter:
This is a general cleanup PR for all GHA workflows that reduces overall duplication and redundancy. Covers most of #5242 sans the actions upgrades.

- `working-directory` is kind of evil and usually you prefer to do `cmake --build build` in the root of the workspace, this lets you address EVERYTHING as relative paths
- `${{ runner.workspace }}` is (usually) unnecessary thanks to relative paths
- wine_dll builds and ALL tests have been moved to composite actions that do everything they did with less code.
- optimized some arch-specific steps to use POSIX `case` on `matrix.arch[1]` rather than explicit steps for each one.
- Enforced slightly better scheme for step names
- General formatting
- Lots of other tiny things I can't remember

Overall much less LoC for the same functionality, easier to add new testsets and such later.

### DIFF
--- a/.github/workflows/wine_dll_artifacts.yml
+++ b/.github/workflows/wine_dll_artifacts.yml
@@ -4,10 +4,6 @@ on:
   push:
     branches:
       - main
-  # TODO(CRUETER): REMOVE
-  pull_request:
-    branches:
-      - main
 
 env:
   BUILD_TYPE: Release


crueter:
TODO: evaluate the necessity of the `always()` guards.

crueter:
Be sure to review carefully, some of the changes were generated via a hastily written shell script and may not be 100% accurate. Also cross-verify some of the flags for wine_dll_artifacts and make sure I didn't make any silly typos when copy-pasting.

Will clean up history once reviewed

neobrain:
> Be sure to review carefully, some of the changes were generated via a hastily written shell script and may not be 100% accurate

Sounds like a carefully written shell script would've been the more time-saving method here overall. Could you do a careful self-review first before asking us to double-check things that may contain sloppy errors? Thanks.

Also not sure the changes like `run: cmake -E make_directory ${{runner.workspace}}/build` => `run: cmake -E make_directory build` are strictly necessary for this PR. If they're not, please factor them out into separate commits (`git add/restore -u -p`) to minimize review surface.

Still appreciate being able to throw out 100 lines of workflow code, neat.

crueter:
> Also not sure the changes like `run: cmake -E make_directory ${{runner.workspace}}/build` => `run: cmake -E make_directory build` are strictly necessary for this PR.

I removed those entirely and replaced them with just `cmake -S . -B build` which saves ~4 lines per workflow. Do you mean just the runner.workspace part?

also should I do a separate PR for that or just a commit?

crueter:
Did another few run-throughs of everything and all of the behavior matches what it was before. Script-generated parts all match previous naming and test schemes. Running workflows locally (at least the ones that I can) generates identical results to main.

.github/workflows/hostrunner.yml
```diff
@@ -36,46 +36,38 @@ jobs:
...
+        target: asm_tests
+        log-name: ASM
 
     - name: Truncate test results
       if: ${{ always() }}
-      working-directory: ${{runner.workspace}}/build
       # Cap out the log files at 20M in case something crash spins and dumps fault text
       # ASM tests get quite close to 10MB
-      run: truncate --size="<20M" ${{runner.workspace}}/build/Testing/Temporary/LastTest_*.log || true
+      run: truncate --size="<20M" build/Testing/Temporary/LastTest_*.log || true
```
neobrain, line 70: Is this a relevant change or just a bonus cleanup? If it's the latter, could you move it to a separate patch (and squash the rest so it's reviewable individually)?

There seem to a lot of these throughout the PR diff.

crueter: It's a bonus cleanup, so I'll move it when I have the time

crueter:
Will be split into other PRs

PR #5249, opened 2026-01-20
CI: Remove redundant runner.workspace and working-directory directives

crueter:
Generally speaking, `${{ runner.workspace }}`/`${{ github.workspace }}`/`$GITHUB_WORKSPACE` are unnecessary, and can actually cause problems when mixing and matching. Artifact uploads still use this though since they seemingly refuse to work with relative paths (?)

`working-directory` can also cause problems and can generally be replaced trivially with `cmake -S . -B build` or the like. Some of the test moves may benefit from `working-directory: ${{ runner.workspace }}/build/Testing/Temporary`, but that will be solved in a separate PR.

Also, try to avoid explicit env directives unless it's cleaner to include them. In cases like `DESTDIR` it's better to put it in-line

Separated from #5246 for reviewer's ease.

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -36,177 +36,146 @@ jobs:
         echo "FEX_ROOTFS=$HOME/Rootfs/" >> $GITHUB_ENV
 
     - name: Update RootFS cache
-      run: $GITHUB_WORKSPACE/Scripts/CI_FetchRootFS.py
+      run: python3 Scripts/CI_FetchRootFS.py
 
-    - name : submodule checkout
+    - name: Checkout Submodules
       # Need to update submodules
       run: |
         git submodule sync --recursive
         git submodule update --init --depth 1
 
     - name: Clean Build Environment
-      run: rm -Rf ${{runner.workspace}}/build
+      run: rm -Rf build
 
     - name: Create Build Environment
-      run: cmake -E make_directory ${{runner.workspace}}/build
+      run: cmake -E make_directory build
 
     - name: Configure CMake
-      working-directory: ${{runner.workspace}}/build
       run: |
-        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G Ninja -DENABLE_LTO=False -DENABLE_ASSERTIONS=True \
+        cmake -S . -B build -DCMAKE_BUILD_TYPE=$BUILD_TYPE -G Ninja -DENABLE_LTO=False -DENABLE_ASSERTIONS=True \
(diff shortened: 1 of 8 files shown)


.github/workflows/ccpp.yml
```diff
@@ -36,177 +36,150 @@ jobs:
...
+      run: truncate --size="<20M" build/Testing/Temporary/LastTest_*.log || true
 
     - name: Remove old SHM regions
       if: ${{ always() }}
-      working-directory: ${{runner.workspace}}/build
-      run: cmake --build . --config $BUILD_TYPE --target remove_old_shm_regions
+      run: |
+        cmake --build build --config $BUILD_TYPE --target remove_old_shm_regions
+
+        # TEST: PLEASE Remove
```
Sonicadvance1: For plez removing?

.github/workflows/wine_dll_artifacts.yml
```diff
@@ -22,75 +22,59 @@ jobs:
...
-    - name: Configure CMake wow64
-      shell: bash
-      working-directory: ${{runner.workspace}}/build_wow64
+    - name: Configure CMake (wow64)
       run: |
-        cmake $GITHUB_WORKSPACE -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_TOOLCHAIN_FILE=$GITHUB_WORKSPACE/Data/CMake/toolchain_mingw.cmake \
+        cmake -S . -B build_wow64 -DCMAKE_BUILD_TYPE=$BUILD_TYPE -DCMAKE_TOOLCHAIN_FILE=Data/CMake/toolchain_mingw.cmake \
           -DMINGW_TRIPLE=aarch64-w64-mingw32 -DCMAKE_INSTALL_LIBDIR=/usr/lib/wine/aarch64-windows -G Ninja \
-          -DENABLE_LTO=False -DENABLE_ASSERTIONS=False -DENABLE_JEMALLOC_GLIBC_ALLOC=False -DCMAKE_INSTALL_PREFIX=/usr \
+          -DENABLE_LTO=False -DENABLE_ASSERTIONS=False -DENABLE_JEMALLOC_GLIBC_ALLOC=False \
```
Sonicadvance1, line 57: Does removing the install prefix break things?

crueter: It's duplicated immediately thereafter:
```
          -DBUILD_TESTING=False -DCMAKE_INSTALL_PREFIX=/usr -DTUNE_ARCH=generic -DTUNE_CPU=none
```

Sonicadvance1: So it is.

Sonicadvance1 (approved):

PR #5256, opened 2026-01-21
WIP: CMake: Use CPM over Git submodules

crueter:
Incomplete re-do of #5159. Still need to do drm headers and tests. I'm opening this now to provide a place for general discussion of this change, since opinions are (rightfully) mixed thus far.

Using CPM instead of submodules has a whole host of pros and cons, but a general TL;DR is:
- Submodules need to be explicitly updated by the user with `git submodule update --recursive` every time they are changed. Here with CPM, each time the version of a submodule is changed, it's stored in a unique cache directory which doesn't conflict. Note that I have been bit by this multiple times.
- Users don't need to do a recursive clone. These take a VERY long time on slow internet (for example, it took me about 5 minutes for a recursive clone on my crappy home internet). With CPM, smaller source archives are fetched at configure time, and these generally take much less time (on my home internet it took about 1m15s for clone + configure).
- Externals are fetched *as-needed*. Rather than users needing to manually specify what submodules they need or just fetch all of them, CPM will only fetch those that are needed. This means that if you have most of the dependencies installed, you don't have to worry about having redundant submodules checked out.
- CPM does find_package stuff for us. So rather than having to do `find_package`, then check if the package was found, blah blah... CPM does that all for us!
- When making debugging changes in submodules, it's very easy to accidentally add the "dirty" modifier to the subproject's commit status. This pollutes the tree and can be somewhat of a pain to deal with.
  * On that same topic, reverting changes in submodules requires going into the directory and stashing/resetting. With CPM you can just delete the cache (`rm -rf .cache/cpm/<package>`) and it'll be good as new.
- Updating versions is easier with CPM because you can just change the tag or commit sha of a version in the CMake file directly. It's also trivial to enforce minimum version requirements.

Now, CPM is not a silver bullet, nor is it compromise-free:
- Submodules are slightly easier to use for network-sandboxed packagers i.e. Gentoo. I added `<PKG_NAME>_CUSTOM_DIR` to account for this, but it *is* extra work.
- The maintenance burden is slightly higher due to the "black box" nature of CPM and the wrapper. This is where docs come in! Which I will add

On a personal note: I have used [my own personal CPM wrapper](https://git.crueter.xyz/CMake/CPMUtil) for several projects, and received overwhelmingly positive feedback on its simplicity compared to vcpkg and submodules. Personally I have absolutely loved how easy it is to use and how flexible it is.


crueter:
```
/home/swurl/code/FEX/unittests/FEXLinuxTests/tests/.cache/cpm/catch2/v3.11.0/fuzzing/fuzz_textflow.cpp:39:45: error: unknown type name 'uint8_t'
   39 | extern "C" int LLVMFuzzerTestOneInput(const uint8_t *Data, size_t Size) {
      |                                             ^
```

bylaws:
Need to think about this but I think submodules will likely still be desired for proton, fetching build deps at build time there isn't ideal due to how things are structured and it copying source to a temporary directory

Sonicadvance1:
launchpad /probably/ doesn't care at least. No idea what package maintainers would think about this.
I personally don't know how I feel about this, I'm used to the submodule workflow.

crueter:
I added `<PKG_NAME>_CUSTOM_DIR` as an option such that packagers can pre-fetch the repository archives and use them from there. Worst case scenario with this PR, it's entirely possible to add back submodules for specific externals if needed.

crueter:
Might be better to keep Catch2 as a submodule for now since some of the subprojects use it, and it's not as easy to implant it as vk headers

bylaws:
Discussed with proton devs. On the proton side we would need to use custom dirs if this is implemented, but that would presumably lead to versioning annoyances for updates if FEX requires a newer dep. Am moderately against

crueter:
I could go all-in and port over the JSON parsing and selective parts of CPMUtil's tooling, such that you can fetch all needed deps with one script e.g. `Scripts/CPMFetch.sh <packages...>`. Could make it easier for the Proton peeps.

bylaws:
> I could go all-in and port over the JSON parsing and selective parts of CPMUtil's tooling, such that you can fetch all needed deps with one script e.g. `Scripts/CPMFetch.sh <packages...>`. Could make it easier for the Proton peeps.

I think non git external dependences as a whole are just not desirable.

crueter:
Why's that? It's not really any different than using `git submodule init`

bylaws:
'I would prefer not to deal with this. Build-time downloads have bit us many many times and caused a lot of random downtimes. We have a few dependencies using this mechanism that we are actively working to drop. Magic caches are terrible in the long run both locally for development (e.g. checking older version of the repo on a flaky internet) and for CI (you have to cache more than just the git repos in some form). If that happens I think we would end up having to use `<PKG_NAME>_CUSTOM_DIR`s and have those included as submodules ourselves. Git clones are slow, but git updates are usually faster and more predictable.'

crueter:
My experience was much the same on projects that *didn't* properly implement it but the total opposite when I took the time and actively worked on designing a system specifically to fix those exact problems. I get their concerns but I have very specifically designed around those potential flaws.

I still think this is absolutely a net positive given the universally positive feedback I have received from people who have interacted with this paradigm. Regardless, I suppose I will leave this here and potentially revisit it if I can work something else out, or discuss with the peanut gallery thereof.

crueter:
Idea:

- default behavior is find_package(whatever) and it will use that if available
- if not available, will check if associated submodule contains anything, and use it if so
- if THAT's not available, then fetch with CPM

(should) make everyone happy maybe?

neobrain:
> Idea:
> 
>  * default behavior is find_package(whatever) and it will use that if available
>  * if not available, will check if associated submodule contains anything, and use it if so
>  * if THAT's not available, then fetch with CPM
> 
> (should) make everyone happy maybe?

The general vibe from the comments above is that CPM doesn't help us solve any problems we (and our users) typically encounter. Also worth noting that many of the proposed benefits don't apply if we have to keep maintaining the existing dependency resolution methods anyway.

Closing as not planned. For this to be reconsidered, a strong value-add specific to FEX would need to be brought up along with a convincing analysis of potential migration risks.

PR #5261, opened 2026-01-27
CI: Move noisy build targets generation to a separate step

crueter:
As requested by neobrain. Makes log parsing easier because you can see
the "noisy" and usually unchanging targets in one step
(which usually succeed, I hope), and the more error-prone "the rest of FEX"
build in separate steps. Makes parsing FEX build errors easier

### DIFF
--- a/.github/workflows/ccpp.yml
+++ b/.github/workflows/ccpp.yml
@@ -55,6 +55,11 @@ jobs:
           -DENABLE_X86_HOST_DEBUG=True -DBUILD_FEX_LINUX_TESTS=True -DBUILD_THUNKS=True \
           -DCMAKE_INSTALL_PREFIX="$PWD"/build/install
 
+    # These steps make a lot of noise but rarely fail.
+    # Put them in a separate step to make normal build logs easier to parse
+    - name: Noisy Build Targets
+      run: cmake --build build --target asm_files 32bit_asm_files JemallocLibs Catch2 vixl cephes_128bit
+
     - name: Build
       run: cmake --build build
 
--- a/.github/workflows/glibc_fault.yml
+++ b/.github/workflows/glibc_fault.yml
@@ -63,6 +63,11 @@ jobs:
           -DENABLE_GLIBC_ALLOCATOR_HOOK_FAULT=True -DENABLE_JEMALLOC_GLIBC_ALLOC=False \
           -DCMAKE_INSTALL_PREFIX="$PWD"/build/install
 
+    # These steps make a lot of noise but rarely fail.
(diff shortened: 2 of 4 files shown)


Sonicadvance1:
We have more tests than just asm tests, so the step name is a bit weird.

crueter:
Oop, good catch, will change that when I finish this other PR

.github/workflows/ccpp.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           -DENABLE_X86_HOST_DEBUG=True -DBUILD_FEX_LINUX_TESTS=True -DBUILD_THUNKS=True \
           -DCMAKE_INSTALL_PREFIX="$PWD"/build/install
 
+    - name: Generate ASM Tests
+      run: cmake --build build --target asm_files 32bit_asm_files
```
neobrain: Do we maybe also want to add the FEX_jmalloc/FEX_jmalloc_glibc/Catch2/vixl/softfloat_3e/cephes_128bit targets here? Could rename the task to "Noisy build targets" or similar then.

.github/workflows/ccpp.yml
```diff
@@ -55,6 +55,9 @@ jobs:
           -DENABLE_X86_HOST_DEBUG=True -DBUILD_FEX_LINUX_TESTS=True -DBUILD_THUNKS=True \
           -DCMAKE_INSTALL_PREFIX="$PWD"/build/install
 
+    - name: Generate ASM Tests
```
neobrain: Can we add a comment here to explain why these are done in a separate build step?

.github/workflows/ccpp.yml
crueter: Could, yes. Would be a lot of targets in one line though, maybe another composite action time?

neobrain: Not sure, it's still one line and not like the list changes every other week.

FWIW I realized FEX_jemalloc+its glibc variant are grouped into the `JemallocLibs` target at least, and cephes pulls in `softfloat_3e`.

So the real list is more like `JemallocLibs Catch2 vixl cephes_128bit`, which isn't too bad.

crueter: Yeah that's more manageable, can be done

.github/workflows/ccpp.yml
crueter: yes

neobrain:
Ah, jemalloc_glibc still gets built since it's only an INTERFACE target of JemallocLibs. Anyway, this is fine - it's still a massive improvement and I'm fine scrolling over the few remaining lines. Thanks!

Issue #5234, opened 2026-01-14
syscall execveat behavior doesnt match x86_64 native machine when executing a ELF file loaded in memfd


fish4terrisa-MSDSM:
For some reason, the syscall `__NR_execveat`'s behavior doesnt match a x86_64 native machine when it's executing a x86_64 ELF loaded in memfd.
# Reproduce
POC:
```
#define _GNU_SOURCE
#define __NR_execveat 322
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <strings.h>
#include <unistd.h>
#include <fcntl.h>
#include <sys/mman.h>
#include <sys/stat.h>



int main(void)
{

 int fd = open("busybox", O_RDONLY); // Put a static amd64 busybox binary in the same dir as target
 if (fd == -1) return 1;

 struct stat st;
 int rv = fstat(fd, &st);

 if (rv == -1) return 2;

 void *buffer = malloc(st.st_size);
 if (buffer == NULL) return 3;

 ssize_t bytesRead = read(fd, buffer, st.st_size);
 if (bytesRead == -1) return 4;
 //
 // Create an memfd file in memory
 //
 int af = memfd_create("anonfile", MFD_CLOEXEC | MFD_ALLOW_SEALING);
 if (af == -1) return 5;
 char* argv[] = { "busybox", NULL };

 //
 // Write the ELF file into new anonymous file and execute it
 //

 ssize_t bytesWritten = write(af, buffer, st.st_size);
 if (bytesWritten == -1) return 6;

 syscall(__NR_execveat, af, (unsigned long)"", (unsigned long)argv, NULL, AT_EMPTY_PATH);
 close(fd);

}
```
compile:
```
gcc memexec.c -o memexec
./memexec # Make sure there is a static x86_64 build busybox put in the same folder
```
# Result
On a x86_64 native machine, you'll see busybox's help message(depend on the distro you are using, it might complain about missing applet if your busybox is cut. Either case you'll know busybox executed successfully.
In a FEX environment when running the same `memexec` binary with FEX, however, it'll have some problem.
The `busybox` executeable wont be lanuched, with `strace` it shown that it was complaining about `-ENOENT`. An interesting fact is that when I bind mount a static arm64 build of busybox to that x86_64 busybox's place, it'll get lanuched successfully. 
So it seems that in this case when using `execveat` executing an elf file in memfd, it can only lanuch arm64 binary for some reason, which breaks some x86_64 programs(like UML)
[...]

PR #5265, opened 2026-01-27
Linux: Ensure seccomp inheritence occurs unconditionally

Sonicadvance1:
Seccomp once enabled is a virus that passes onward to all children processes. Ensure we enable the option if we are inheriting a filter.

For #5234


Source/Tools/FEXInterpreter/FEXInterpreter.cpp
```diff
@@ -400,6 +400,11 @@ int main(int argc, char** argv, char** const envp) {
   FEXCore::Config::Set(FEXCore::Config::CONFIG_DISABLE_VIXL_INDIRECT_RUNTIME_CALLS, "0");
 #endif
 
+  if (FEXSeccompFD) {
+    // seccomp inheritence happens unconditionally.
```
neobrain:
```suggestion
    // seccomp inheritance happens unconditionally
```

PR #5268, opened 2026-01-27
CI: Enforce SPDX headers on newly changed files

crueter:
Just the identifier, year can be easily retrofitted by changing one
variable. This script will receive updates in the future, so can be
implanted when those occur.

Not sure if retroactively applying headers is a good idea, but that
would be a separate PR.

### DIFF
--- a/.github/workflows/license-header.yml
+++ b/.github/workflows/license-header.yml
@@ -1,0 +1,29 @@
+# SPDX-License-Identifier: MIT
+
+name: Check license headers
+on:
+  pull_request:
+    branches:
+      - main
+
+jobs:
+  license_header:
+    runs-on: [self-hosted, X64]
+    if: github.repository == 'FEX-Emu/FEX'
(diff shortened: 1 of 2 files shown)


crueter:
CC. this requires `license_header` to be changed to be a required check

crueter:
Unfortunately no, REUSE has a command-line interface but it just doesn't work right. I can rework the existing code into something a bit more compact by removing the add/update code.

crueter:
OK, consolidated it a bit

neobrain:
I'm still not sure this solves a problem we had. It's not a ton of code, but the cost of missing an SPDX header now and then is equally marginal. Maybe others want to weigh in on this.

Sonicadvance1:
I'm just going to say no to this. It's CI for the sake of CI rather than actually solving a problem. Good shout that some projects probably need this but I don't think that's us.

PR #5272, opened 2026-01-27
FEXCore: Fix initial PF flag value

pmatos:
Initialize pf_raw to 1 instead of 0 so that the reconstructed Parity
Flag matches x86 reset state (PF=0).

### DIFF
--- a/unittests/ASM/FEX_bugs/InitialPFFlag.asm
+++ b/unittests/ASM/FEX_bugs/InitialPFFlag.asm
@@ -1,0 +1,18 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x202"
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Might want to double check that the hostrunner is actually clearing flags to the "default" state. On test entry it is probably whatever.

pmatos:
> Might want to double check that the hostrunner is actually clearing flags to the "default" state. On test entry it is probably whatever.

Interesting - the test passes on my intel machine. I wonder what's going on in the host runner.

pmatos:
Seems to be related to the initialized state in the Hostrunner, I changed it to match what I get on my intel machine. 

@Sonicadvance1 wdyt?

PR #5274, opened 2026-01-28
FEXInterpreter: Override glibc set program invocation name

Sonicadvance1:
Mesa uses this to determine what the executable name is in a process for application profiles. Without thunking, the guest glibc sets this correctly, with thunking mesa would pick up `FEX` instead of the app.

With this fixed, it fixes Dead Island rendering when thunks are enabled. Plenty of other games in Mesa's application profiles that it would fix as well.

### DIFF
--- a/Source/Tools/FEXInterpreter/ELFCodeLoader.h
+++ b/Source/Tools/FEXInterpreter/ELFCodeLoader.h
@@ -886,7 +886,8 @@ public:
 
     // Setup ExecFN aux
     AuxExecFN->val = StackPointer + ExecFNLocation;
-    strncpy(reinterpret_cast<char*>(AuxExecFN->val), ApplicationArgs[0].c_str(), ApplicationArgs[0].size() + 1);
+    const auto InvocationName = reinterpret_cast<char*>(AuxExecFN->val);
+    strncpy(InvocationName, ApplicationArgs[0].c_str(), ApplicationArgs[0].size() + 1);
 
     // Stack setup
     // [0, 8):   Argument Count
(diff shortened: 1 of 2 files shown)


Source/Tools/FEXInterpreter/ELFCodeLoader.h
```diff
@@ -911,6 +912,21 @@ class ELFCodeLoader final : public FEX::CodeLoader {
     }
 
     RemapArgumentData(StackPointer + ArgumentOffset, ArgumentBackingSize);
+#if defined(HAS_PROGRAM_INVOCATION_NAME) && HAS_PROGRAM_INVOCATION_NAME
+    // Set the glibc invocation names to the process name.
+    // Mesa uses this to determine application profiles.
+    // Necessary when thunking is enabled otherwise mesa would only see FEX.
+
+    std::string_view INV = std::string_view(InvocationName, ApplicationArgs[0].size());
```
neobrain, line 920: How is this different from argv[0]? Seems strange to introduce an external feature check to the build system for just this one use.

Sonicadvance1: argv[0] is whatever is passed in to execveat, it could be potato, it could be null.

This is a set of global variables that glibc exposes as a data symbol for the executable path.

Would be a bit difficult for a library like libGL to find an argv[0] in memory anyway.

neobrain: Ah, I misread this. So host libGL will see the symbol as provided by host glibc, hence we need to override it, makes sense!

PR #5286, opened 2026-02-05
Fix Buffer Overflow in LoadFileImpl

### LINKED ISSUE #5285: Buffer Overflow in LoadFileImpl
The function `LoadFileImpl` in `FEXCore/Source/Utils/FileLoading.cpp` line 42 always calls `pread` with an offset of `0` instead of using `CurrentOffset`:

```cpp
while (CurrentOffset != FileSize && (Read = pread(FD, &Data.at(CurrentOffset), FileSize, CurrentOffset)) > 0) {
    CurrentOffset += Read;
}
```

If the `pread` function returns fewer bytes than requested (see read(2)) the while loop will get executed again. Since the offset is always `0` data is again read from the beginning of the file.
Furthermore, since the destination buffer is indexed via `CurrentOffset` a (heap) buffer overflow occurs on subsequent `pread` calls.

**Steps to reproduce**:
On Linux v6.7.0 calling `LoadFileImpl` on a large file (>= 2.1G) triggers the buffer overflow.

**Fix**:
Replace the offset `0` with `CurrentOffset` in `LoadFileImpl`.
Also update the `FileSize` argument to be the number of remaining bytes which still need to be read.

simon902:
Fix #5285

### DIFF
--- a/FEXCore/Source/Utils/FileLoading.cpp
+++ b/FEXCore/Source/Utils/FileLoading.cpp
@@ -39,8 +39,10 @@ static bool LoadFileImpl(T& Data, const fextl::string& Filepath, size_t FixedSiz
   if (FileSize) {
     // File size is known upfront
     Data.resize(FileSize);
-    while (CurrentOffset != FileSize && (Read = pread(FD, &Data.at(CurrentOffset), FileSize, 0)) > 0) {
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Whoa! Spicy! What a miss.

PR #5288, opened 2026-02-09
Config: Fixes bad vector to string cast

Sonicadvance1:
Turns out casting a vector of data to a string is a bad idea, who knew?! This bug was exposed by the rpmalloc PR because the `/run/host/container-manager` file isn't null terminated. So the cast fextl::string through std::vector::data had the /potential/ to not be null terminated.

This was just a bug hiding in the code, convert it over to use fextl::string directly and avoid the entire dance.

Fixes #5236

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.cpp
+++ b/FEXCore/Source/Interface/Config/Config.cpp
@@ -307,12 +307,10 @@ constexpr char ContainerManager[] = "/run/host/container-manager";
 fextl::string FindContainer() {
   // We only support pressure-vessel at the moment
   if (FHU::Filesystem::Exists(ContainerManager)) {
-    fextl::vector<char> Manager {};
+    fextl::string Manager {};
     if (FEXCore::FileLoading::LoadFile(Manager, ContainerManager)) {
       // Trim the whitespace, may contain a newline
-      fextl::string ManagerStr = Manager.data();
-      ManagerStr = FEXCore::StringUtils::Trim(ManagerStr);
-      return ManagerStr;
+      return FEXCore::StringUtils::Trim(Manager);
     }
   }
   return {};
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
This might also explain why some other people managed to randomly get logs about a bad ELF. Since this has been lurking in our source for a significant amount of time, and would change behaviour depending on random allocation behaviours.

FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -307,12 +307,11 @@ constexpr char ContainerManager[] = "/run/host/container-manager";
...
   if (FHU::Filesystem::Exists(ContainerManager)) {
-    fextl::vector<char> Manager {};
+    fextl::string Manager {};
     if (FEXCore::FileLoading::LoadFile(Manager, ContainerManager)) {
       // Trim the whitespace, may contain a newline
-      fextl::string ManagerStr = Manager.data();
-      ManagerStr = FEXCore::StringUtils::Trim(ManagerStr);
-      return ManagerStr;
+      Manager = FEXCore::StringUtils::Trim(Manager);
+      return Manager;
```
neobrain:
```suggestion
      return FEXCore::StringUtils::Trim(Manager);
```

FEXCore/Source/Interface/Config/Config.cpp
```diff
@@ -321,12 +320,11 @@ fextl::string FindContainer() {
...
   if (FHU::Filesystem::Exists(ContainerManager)) {
-    fextl::vector<char> Manager {};
+    fextl::string Manager {};
     if (FEXCore::FileLoading::LoadFile(Manager, ContainerManager)) {
       // Trim the whitespace, may contain a newline
-      fextl::string ManagerStr = Manager.data();
-      ManagerStr = FEXCore::StringUtils::Trim(ManagerStr);
-      if (strncmp(ManagerStr.data(), "pressure-vessel", Manager.size()) == 0) {
+      Manager = FEXCore::StringUtils::Trim(Manager);
+      if (strncmp(Manager.data(), "pressure-vessel", Manager.size()) == 0) {
```
neobrain:
```suggestion
      if (FEXCore::StringUtils::Trim(Manager) == "pressure-vessel") {
```

PR #5291, opened 2026-02-10
CI: Clean up syntax assertions and remove redundant log-name parameter

crueter:
Now logs are uploaded directly as the name of the test  instead of a
separate name.


Sonicadvance1:
This breaks artifacts uploads.
`
No files were found with the provided path: /home/buildbot/actions-runner/_work/FEX/FEX/build/Testing/Temporary/LastTest_*.log. No artifacts will be uploaded.
`

crueter:
yeah, forgot to change that path. Will do in a min

PR #5293, opened 2026-02-10
CMake: Avoid global include_directories in favor of propagation

crueter:
Uses ALIAS targets to propagate include directories to targets rather
than polluting the preprocessor with an extra include directory it
usually doesn't need.


crueter:
Depends on https://github.com/FEX-Emu/vixl/pull/8

PR #5300, opened 2026-02-11
Inline softfloat i16/i32/f32/f64 to extF80 conversions in dispatcher

pmatos:
This was mainly targetted to MirrorsEdge but hopefully will help other workloads that hit these softfloat conversions in a similar way.

### DIFF
--- a/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/Dispatcher/Dispatcher.cpp
@@ -26,7 +26,6 @@
 
 #include <array>
 #include <bit>
-#include <csignal>
 #include <cstring>
 
 namespace FEXCore::CPU {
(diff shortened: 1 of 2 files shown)


pmatos:
@Sonicadvance1 it would be great if you get the change to try this on MirrorsEdge. My experiments with MirrorsEdge saw Softfloat count going from 5M/s in the start of the game to about 6k/s.

Sonicadvance1:
```
Before:
Test, Total Cycles, Total Runs, Cycles Average, Internal Loops, Average cycles per internal, Average NS per internal, per/second
FLD32, 2137113050, 50000, 42742.3, 1000, 42.7423 (42.7423ns), 23396.048234
FLD64, 2092929490, 50000, 41858.6, 1000, 41.8586 (41.8586ns), 23889.959140

After:
Test, Total Cycles, Total Runs, Cycles Average, Internal Loops, Average cycles per internal, Average NS per internal, per/second
FLD32, 147746090, 50000, 2954.92, 1000, 2.95492 (2.95492ns), 338418.431242
FLD64, 148256340, 50000, 2965.13, 1000, 2.96513 (2.96513ns), 337253.705305
```
Holy hecking quick. The variability in Mirror's Edge is making it hard to bench, but maybe it went from 3FPS to up to 12 with wow64?
But it seems to break Proton Experimental running the game? Starts throwing a GLXBadFBConfig.

I'm adding a toggle per implementation to try and see which one breaks it.

Sonicadvance1:
Oop, derp in my game testing, it works fine.

Before: `Softfloat Cnt: 9,225,639 `
After: `Softfloat Cnt: 44,312 `

It still gets down to single digit FPS, but the average is higher. So instead of 3FPS, it's more like 6-9FPS on Proton-Exp.

Sonicadvance1:
Without PhsyX:
Before: `Softfloat Cnt: 7,415,463 `
After: `Softfloat Cnt: 48,010`

And the FPS is significantly improved, hitting 60 more consistently.

Sonicadvance1 (approved):
I'm all for a 14x performance improvement!

PR #5301, opened 2026-02-12
CMake: Print FEX version strings found

Sonicadvance1:
Useful debugging

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -540,6 +540,8 @@ else()
   set(GIT_SHORT_HASH "${OVERRIDE_HASH}")
 endif()
 
+message(STATUS "FEX version strings: git describe: '${GIT_DESCRIBE_STRING}'. git short hash: '${GIT_SHORT_HASH}'")
+
 if (ENABLE_IWYU)
   find_program(IWYU_EXE "iwyu")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -540,6 +540,8 @@ else()
   set(GIT_SHORT_HASH "${OVERRIDE_HASH}")
 endif()
 
+message(STATUS "FEX version strings: git describe: '${GIT_DESCRIBE_STRING}'. git short hash: '${GIT_SHORT_HASH}'")
```
crueter, line 543: FYI, in CMake land, long status messages like this are somewhat frowned upon. It's probably fine for Git describe but it may be preferable to change to this:

```cmake
message(STATUS "FEX Version: ${GIT_DESCRIBE_STRING}")
message(STATUS "FEX Commit: ${GIT_SHORT_HASH}")
```

Sonicadvance1: Sounds like something that would be a stylistic thing. Sure some people hate going beyond 72 columns because we're all programming on classic terminals, but they can learn to live in a widescreen world.

crueter: I consider myself very stylish

PR #5313, opened 2026-02-16
Revert "CodeCache: Use defaulted dtor for ExecutableFileInfo"

neobrain:
Couldn't reproduce WoA build failures reported in #5164; current `main` builds fine without this, so remove the false header dependency.


bylaws:
It might have been the offline compiler that broke rather than current code? Cannot quite remember. Manually overriding the destructor meant I couldn't do something I wanted to and this was the simplest way to fix.

Happy to merge but will need to be reverted again whenever that lands

PR #5315, opened 2026-02-19
Fix InstallFEX.py not running FEXRootFSFetcher correctly

### LINKED ISSUE #5310: FEXRootFSFetcher fails to execute at the end of our install script
```
Packages installed
RootFS not found. Running FEXRootFSFetcher to get rootfs
RootFS not found. Do you want to try and download one?
Response {y,yes,1} or {n,no,0}
Unknown response. Assuming no
FEX is now installed. Trying basic program run
Invalid or Unsupported elf file.
This is likely due to a misconfigured x86-64 RootFS
Current RootFS path set to ''
RootFS path doesn't exist. This is required on AArch64 hosts
Use FEXRootFSFetcher to download a RootFS
FEX failed to run. Not continuing
```

A bit weird. Tested on an Oracle Free Ampere Altra VPS.

carsongoodwin32:
Fixes #5310 

On a fresh install of ARM64 Ubuntu 25.10, running InstallFEX.py with the provided curl command produces:

```
RootFS not found. Do you want to try and download one?
Response {y,yes,1} or {n,no,0}
Unknown response. Assuming no
```

Which results in no RootFS being installed and the user having to either manually run FEXRootFSFetcher or forgo the provided command in the repository description, download InstallFEX.py, and run it directly.

This fixes the issue by explicitly opening /dev/tty and passing it in to stdin for the subprocess call. It also has an exception if that fails in a similar way to how the KeyboardError is implemented. 

I also changed subprocess.call() to subprocess.run() since that seems to be the recommended way of doing subprocesses post Python 3.5 (and changed all of the CmdResult checks to work with the output of subprocess.run() since it no longer returns an integer, but stores a return code inside of the returned object). I hope this is okay. 

Tested and working on ARM64 Ubuntu 25.10 running under UTM on macOS.

### DIFF
--- a/Scripts/InstallFEX.py
+++ b/Scripts/InstallFEX.py
@@ -161,8 +161,8 @@ def InstallPPA():
 
     DidInstall = False
     try:
-        CmdResult = subprocess.call(["sudo", "add-apt-repository", "-y", "ppa:fex-emu/fex"])
-        DidInstall = CmdResult == 0
+        CmdResult = subprocess.run(["sudo", "add-apt-repository", "-y", "ppa:fex-emu/fex"])
+        DidInstall = CmdResult.returncode == 0
     except KeyboardInterrupt:
         DidInstall = False
         pass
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Nice! I have confirmed locally that this also fixes the issue!

Python 3.5 as a minimum requirement is fine since the oldest Ubuntu version we support ships python 3.10.

PR #5317, opened 2026-02-20
JIT: Support avoiding spilling CPU flags

Sonicadvance1:
When thunks are jumping out, games are jumping /entirely/ out of their controlled code, which means we don't need to save and restore NZCV,PF,AF.

Some CPUs don't fully rename direct accesses to this register which adds up during thunking. FPCR is also in the same situation where it'll force pipeline flushes and isn't renamed away, but we can't really avoid that.

Improves performance at least in Detroit: Become human where the game spends ~48% CPU time inside of the thunk trampoline for `vkUpdateDescriptorSets`.

I plan on a follow-up PR where I converge all these options into a struct argument instead, but that's a follow-up since I don't want to burn a bunch of time right now.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
@@ -322,7 +322,7 @@ DEF_OP(Thunk) {
   // X0: CTX
   // X1: Args (from guest stack)
 
-  SpillStaticRegs(TMP1); // spill to ctx before ra64 spill
+  SpillStaticRegs(TMP1, true, ~0U, ~0U, false); // spill to ctx before ra64 spill
 
   PushDynamicRegs(TMP1);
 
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Interface/Core/JIT/BranchOps.cpp
```diff
@@ -322,7 +322,7 @@ DEF_OP(Thunk) {
   // X0: CTX
   // X1: Args (from guest stack)
 
-  SpillStaticRegs(TMP1); // spill to ctx before ra64 spill
+  SpillStaticRegs(TMP1, true, ~0U, ~0U, false); // spill to ctx before ra64 spill
```
lioncash, line 325: Should we perhaps convert the parameter list into an extendable struct that uses designated initializers to make call sites more self explanatory; not sure how much of a pain that'd be

Sonicadvance1: Indeed, that's what I meant by the commit message.

> I plan on a follow-up PR where I converge all these options into a struct argument instead, but that's a follow-up since I don't want to burn a bunch of time right now.

lioncash: Oh, turns out I'm illiterate this Thursday (oops)

lioncash (approved):

PR #5323, opened 2026-02-23
json_ir_generator: Generate docstring for IR op emitters

neobrain:
This allows IDEs to show the description texts when hovering over IR emitter functions in cpp files.


Sonicadvance1 (approved):
Sure.

PR #3584, opened 2024-04-16
Library Forwarding: Support libGL/libvulkan without forwarding libX11

neobrain:
~~*Marked as draft because this will need **a lot** of testing*~~

## Overview

Currently when forwarding libGL/libvulkan to host libraries, FEX generally requires libX11 calls to be forwarded to the aarch64 host as well to do anything interesting. This brings a number of problems, since the libX11 API uses complex data structures that are difficult to adapt for an aarch64 host without breaking the x86 client application. Most importantly, forwarding libX11 from a 32-bit application to the 64-bit host in the current form is impossible since the layout of these shared data structures will be inconsistent.

This PR removes FEX's current libX11 forwarding support and instead makes the host-side libGL/libvulkan interoperable with an X11 connection managed by the guest. This idea was initially brought up in #1735, but the implementation was reliant on adding tremendous amounts of boilerplate.

## Implementation

Interoperability is enabled by having our host-side libGL/libvulkan wrappers open a shadow connection to the X server (either via `xcb_connect` or `XOpenDisplay`). Since most libX11 objects are global identifiers (`XID`s) that are valid in any context, the host-side drivers can talk over this shadow connection as if it were the same as the guest-side connection.

This means we can pass XIDs across architecture boundaries without modification. Only the connection types (`xcb_connection_t` or `_XDisplay`) must be exchanged for the shadow connection (and vice-versa) when crossing architectures.

Furthermore, `XVisualInfo` is an exceptional libX11 type since it's neither a type alias for `XID` nor a connection. These objects can't directly be shared between guest and host due to the embedded `Visual*` object. However, we can use `XGetVisualInfo` to look up a valid `XVisualInfo` whenever we cross architectures.

Finally, some libX11 functions used by libGL allocate data internally and require the user to de-allocate it using XFree. Since XFree is a macro that expands to `free`, data allocated by the host-side GL driver (on the host heap) would be de-allocated as if it were on the guest heap. To handle this properly, FEX wraps data-allocating libGL functions and relocates their returned data to the guest stack.

## Impact

This PR has the potential to fix any game that could previously not use libGL/libvulkan forwarding due to issues in libX11.

This PR also has the potential to break any games that use the glX/WSI APIs in an unexpected way.

## Testing

I have thrown this PR at all games I can get running in my Linux VM with virgl / llvmpipe and could not find any regressions. Unless otherwise noted, these all work:
* 7 Billion Humans ✅
[...]

PR #5325, opened 2026-02-23
Remove unused X11 libraries

ChanthMiao:
Just remove some unused code that droped by #3584 .


neobrain:
Thanks!

Note that "thunks" is legacy terminology that should be avoided in favor of library forwarding or wrapping; the legacy term is too overloaded, and many of our function wrappers don't actually do any thunking.

PR #5326, opened 2026-02-23
Add vcmp**_full tests

OFFTKP:
Tests all 32 predicates with different values and NaNs

PR #5327, opened 2026-02-23
FEXCore: Fixes VEX float compare operations

Sonicadvance1:
The AMD documentation about this instruction is very vague and misleading in multiple ways. While the Intel documentation is much cleaner and explains how we need to implement these.

8 of these "new" operations are just inverted signaling versions of the original 8 SSE versions.
The remaining 16 new operations fill gaps in the original x86 version of the instructions, exposing the 5 bit truth tables directly, which is why we also have a "true" and "false" version as well.

Both scalar and vector wide.
Fixes #5326

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
@@ -562,31 +562,90 @@ template void OpDispatchBuilder::AVXInsertScalarRound<OpSize::i64Bit>(OpcodeArgs
 
 Ref OpDispatchBuilder::InsertScalarFCMPOpImpl(OpSize Size, IR::OpSize OpDstSize, IR::OpSize ElementSize, Ref Src1, Ref Src2,
                                               uint8_t CompType, bool ZeroUpperBits) {
-  switch (CompType & 7) {
-  case 0x0: // EQ
-    return _VFCMPScalarInsert(Size, ElementSize, Src1, Src2, FloatCompareOp::EQ, ZeroUpperBits);
-  case 0x1: // LT, GT(Swapped operand)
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -2425,25 +2484,61 @@ void OpDispatchBuilder::MOVBetweenGPR_FPR(OpcodeArgs, VectorOpType VectorType) {
...
+  case VectorCompareType::LT_OS: // GT(Swapped operand)
+  case VectorCompareType::LT_OQ: return _VFCMPLT(Size, ElementSize, Src1, Src2);
+  case VectorCompareType::LE_OS: // GE(Swapped operand)
+  case VectorCompareType::LE_OQ: return _VFCMPLE(Size, ElementSize, Src1, Src2);
+  case VectorCompareType::UNORD_Q:
+  case VectorCompareType::UNORD_S: return _VFCMPUNO(Size, ElementSize, Src1, Src2);
+  case VectorCompareType::NEQ_UQ:
+  case VectorCompareType::NEQ_US: return _VFCMPNEQ(Size, ElementSize, Src1, Src2);
+  case VectorCompareType::NLT_US: // NGT(Swapped operand)
+  case VectorCompareType::NLT_UQ: Result = _VFCMPLT(Size, ElementSize, Src1, Src2); return _VNot(Size, ElementSize, Result);
```
lioncash: Should this and NGE be braced so the result var can be local? Would make it a little more consistent with NGT_US below

Sonicadvance1: Oop yep, fixed.

Issue #5296, opened 2026-02-10
Implement remaining x87 alias encodings


Sonicadvance1:
#5289 implemented the alias encodings for FCOM and FCOMP, but according to wikipedia there are more.
We should implement FSTP, FXCHG, FCOMP (0xDE 0xD0+i), and also `FSTPNCE`?

- FSTPNCE is a weird one and will need a bit more investigation. 

As long as they are tested in CI on an x86 host then we can be reasonably sure they're real.

See this wikipedia page for more information. https://en.wikipedia.org/wiki/List_of_x86_instructions#cite_note-x87_alias-346

PR #5332, opened 2026-02-26
X87Tables: Add handling for FCOMP DE D0 aliases

lioncash:
Handles the single remaining alias for FCOMP.

Partially addresses #5296


Sonicadvance1:
Nice nice

PR #5334, opened 2026-02-26
X87Tables: Handle aliases for FSTP

lioncash:
Crosses off the straightforward aliases for #5296. Only FSTPNCE left now

### DIFF
--- a/unittests/ASM/X87/DF_D0.asm
+++ b/unittests/ASM/X87/DF_D0.asm
@@ -1,0 +1,35 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "MM6": ["0x8000000000000000", "0x4001"],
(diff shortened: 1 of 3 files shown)


unittests/ASM/X87/DF_D0.asm
```diff
@@ -0,0 +1,35 @@
...
+fld tword [rdx + 8 * 0]
+
+db 0xdf, 0xd1
+
+lea rdx, [rel data3]
+fld tword [rdx + 8 * 0]
+
+hlt
+
+align 8
```
Sonicadvance1: This should be `align 4096` so we don't test smc shenanigans. Same with the other test.

lioncash: Fixed

Sonicadvance1 (approved):

PR #5337, opened 2026-02-27
External/vixl: Use the right commit hash

crueter:
Totally forgot to update this when I opened that PR. oops!

FYI I would recommend merge-committing if I PR to other repos since it
preserves my GPG signature


Sonicadvance1 (approved):
Oops, don't think we setup the merge rules on those other repos.

PR #5338, opened 2026-02-27
Fix mingw builds with LLVM 22.

cjacek:


Sonicadvance1:
Oh, so that was what was causing the cassert thing that I was looking at yesterday. Good news is that this will fix building, bad news is that at runtime this will immediately explode since `GetActiveProcessorCount` is used for `hardware_concurrency` which we call immediately on startup.

I'll finish up the `GetActiveProcessorCount` implementation that I was working on so we can get this fixed.

PR #5339, opened 2026-02-28
unittests: add x87 FISTP m64int rounding tests

gigachadmma69:
## Summary

- Adds `Rounding_Bankers_64bit.asm`: Tests banker's rounding (round-to-nearest-even) at exact midpoints using `FISTP m64int`. Verifies the tie-breaking rule where exact `.5` values round to the nearest **even** integer (0.5→0, 1.5→2, 2.5→2, 3.5→4, -1.5→-2, -2.5→-2).

- Adds `Rounding_FISTP_64bit.asm`: Tests all 4 x87 rounding modes (nearest, down, up, toward-zero) with `FISTP m64int` for both positive and negative 1.5. The existing `Rounding.asm` only covers 16-bit and 32-bit `FIST`/`FISTP` destinations; this extends coverage to the 64-bit integer case.

## Motivation

The 64-bit integer store path (`FISTP m64int`) has different encoding and codegen from the 16/32-bit paths but had no dedicated rounding mode tests. These tests help catch rounding regressions, particularly banker's rounding edge cases (related: #779).

## Test plan

- Both test files follow the existing FEX unittest format with `%ifdef CONFIG` / `RegData` headers
- `Rounding_FISTP_64bit.asm` uses the `x87cw.mac` include for precision/rounding mode macros
- Expected values verified against x86 hardware behavior

### DIFF
--- a/unittests/ASM/X87/Rounding_Bankers_64bit.asm
+++ b/unittests/ASM/X87/Rounding_Bankers_64bit.asm
@@ -1,0 +1,65 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x0000000000000000",
+    "RBX": "0x0000000000000002",
+    "RCX": "0x0000000000000002",
+    "RDX": "0x0000000000000004",
+    "RSI": "0xFFFFFFFFFFFFFFFE",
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Is this an AI submission? We already test rounding.

PR #5345, opened 2026-03-05
FEXCore: Initial support for large page size code buffers

Sonicadvance1:
Some CPUs are affected by not being able to keep their pipelines fed when running at a 4kb page size. To help alleviate this, we can allocate large page size supporting buffers.

This can happen either directly with mmap, or with madvise and THP. Huge pages aren't always supported through mmap, so have a fallback for madvise so it at least tries to operate when possible.

mmap path is also disabled right now as FEX's VMA allocator doesn't support huge pages yet, so leaning heavily on the madvise implementation right now.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUBackend.cpp
+++ b/FEXCore/Source/Interface/Core/CPUBackend.cpp
@@ -349,15 +349,20 @@ namespace CPU {
     return *Buffer.LookupCache;
   }
 
-  CodeBuffer::CodeBuffer(size_t Size)
-    : AllocatedSize(Size) {
-    Ptr = static_cast<uint8_t*>(FEXCore::Allocator::VirtualAlloc(Size, true));
-    LOGMAN_THROW_A_FMT(!!Ptr, "Couldn't allocate code buffer");
+  CodeBuffer::CodeBuffer(size_t Size, uint32_t PageSize)
+    : AllocatedSize(Size)
+    , PageSize {PageSize} {
+    Ptr = static_cast<uint8_t*>(FEXCore::Allocator::VirtualAllocWithLargePageSupport(Size, PageSize, true));
+    LOGMAN_THROW_A_FMT(Ptr != MAP_FAILED, "Couldn't allocate code buffer");
+
+    if (PageSize == FEXCore::Utils::FEX_PAGE_SIZE) {
+      // Try and use THP when set to minimum page size.
+      FEXCore::Allocator::VirtualTHP(Ptr, Size);
+    }
 
     // Protect the last page of the allocated buffer to trigger SIGSEGV on write access
-    uintptr_t LastPageAddr = AlignDown(reinterpret_cast<uintptr_t>(Ptr) + Size - 1, FEXCore::Utils::FEX_PAGE_SIZE);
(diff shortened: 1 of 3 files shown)


FEXCore/Source/Interface/Core/CPUBackend.cpp
```diff
@@ -370,6 +375,65 @@ namespace CPU {
     FEXCore::Allocator::VirtualFree(Ptr, AllocatedSize);
   }
 
+  CodeBufferManager::CodeBufferManager() {
+#ifndef _WIN32
+    // Check for supported large-page sizes
+    constexpr std::array<uint32_t, 5> testing_sizes = {
```
lioncash: This only has 4 explicit entries, is the 5th zeroed entry intentional?

FEXCore/Source/Interface/Core/CPUBackend.cpp
```diff
@@ -370,6 +375,65 @@ namespace CPU {
...
+#endif
+  }
+
+  uint32_t CodeBufferManager::FindLargestPageSizeForAllocation(size_t AllocationSize) {
+    // TODO: Enable once FEX's allocator support huge-pages.
+    // 4kb page size always supported.
+    return FEXCore::Utils::FEX_PAGE_SIZE;
+
+    // Scan from largest page size to smallest.
+    uint32_t TestingLargePages = SupportedLargePages;
```
lioncash, line 406: perhaps this block of code downwards should be `#if 0`-ed out in the meantime, just so there's no analysis warns of dead code

FEXCore/Source/Interface/Core/CPUBackend.cpp
Sonicadvance1: Oops. I was poking at 1GB page sizes and forgot to reduce it.

FEXCore/Source/Interface/Core/CPUBackend.cpp
neobrain: Dead code should either not be added in the first place or it should be unconditionally compiled so it doesn't break in later changes. The `#if 0` tries to preempt a theoretical warning while making it much more likely this will fail to compile once re-enabled.

FEXCore/Source/Interface/Core/CPUBackend.cpp
```diff
@@ -396,7 +459,9 @@ namespace CPU {
     }
 #endif
 
-    auto Buffer = fextl::make_shared<CodeBuffer>(Size);
+    const auto page_size = FindLargestPageSizeForAllocation(Size);
```
neobrain, line 462: Wrong variable casing - sticks out a lot in this file.

FEXCore/Source/Interface/Core/CPUBackend.h
```diff
@@ -44,10 +44,12 @@ namespace CPU {
   struct CodeBuffer {
     uint8_t* Ptr;
     size_t AllocatedSize; // including guard page; see UsableSize()
+    const uint32_t PageSize;
```
neobrain, line 47: Doesn't seem like we use this state anywhere where it's not already available e.g. as a function parameter.

Sonicadvance1:
I have learned that explicit huge TLB is worthless and we should instead rely solely on THP. So I'm just going to open another PR that does a madvise instead.

PR #5347, opened 2026-03-06
CPUBackend: Enable Transparent Huge Pages on JIT buffers

Sonicadvance1:
If/When this works, the amount of iTLB misses drop dramatically, which reduces L2 TLB pressure, which just improves performance for our CPU cores that have itty-bitty L1 iTLB entry counts.

We can't use `mmap(MAP_HUGETLB)` directly for terrible reasons, so we are required to lean on madvise instead.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUBackend.cpp
+++ b/FEXCore/Source/Interface/Core/CPUBackend.cpp
@@ -363,6 +363,9 @@ namespace CPU {
 
     FEXCore::Allocator::VirtualName("FEXMemJIT", reinterpret_cast<void*>(Ptr), Size);
 
+    // Huge-pages reduce the amount of iTLB misses dramatically when it works.
+    FEXCore::Allocator::VirtualTHP(reinterpret_cast<void*>(Ptr), Size);
+
     LookupCache = fextl::make_unique<GuestToHostMap>();
   }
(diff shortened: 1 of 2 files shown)


lioncash:
lgtm, but should the 'terrible reason' be documented, or is it just something silly

Sonicadvance1:
> lgtm, but should the 'terrible reason' be documented, or is it just something silly

Turns out it's fundamental to how "Huge Pages" work in Linux (but not Transparent Huge Pages! The distinction matters). Which I didn't understand was a thing.
By default, Linux doesn't enable huge pages. You must have a service or something control the `nr_hugepages` files to set a number of huge pages that are allowed. Once that is set, then the kernel reserves that set number of pages in physical memory, consuming the RAM and never returning it.

So even if you have zero applications using Huge Pages, then RAM is gone and won't be returned. Reserve 500? Okay, 500 * <huge page size> of memory is now missing. This is an unacceptable limitation for Huge Pages for our use cases. So the recommendation is to just use THP as it doesn't suffer from this reservation problem, nor does it suffer from failing to allocate due to fragmentation (it'll just fail silently to use THP if PA is fragmented).

lioncash:
Ah, ok. gotcha

PR #5355, opened 2026-03-10
Handle zero length in ChangeProtectionFlags

wsxarcher:
Add a no-op for zero length in ChangeProtectionFlags.

This fixes AMD Vivado 2025.2 which tries to mprotect with 0 as size and merge strategies fails:

(Vivado needs `libncurses5`/`ncurses-compat-libs` in the RootFS and `_JAVA_OPTIONS="-Xint"` in env)

```
Unexpected ChangeProtectionFlags Merge strategy! [0x400000, 0x401000) Versus [0x0, 0x0)
```


Sonicadvance1 (approved):
hah. What a silly little thing. Thanks for the change!

PR #5356, opened 2026-03-10
MemoryOps: Drop MOPS handling into place for MemSet/MemCpy

lioncash:
Drops memset handling in first (and also makes some very old tests actually... test, instead of setting rcx to 0 and then doing nothing)


FEXCore/Source/Interface/Core/JIT/MemoryOps.cpp
```diff
@@ -1943,12 +1959,54 @@ DEF_OP(MemSet) {
...
+          // back like: (address - count) + 1.
+          //
+          // This lets us offset the address such that we can treat a backwards
+          // memset as if it were a forwards one.
+          if (IsBackwards) {
+            sub(TMP2, TMP2, TMP1);
+            add(ARMEmitter::Size::i64Bit, TMP2, TMP2, 1);
+          }
+
+          setp(TMP2, TMP1, Value.X());
```
Sonicadvance1, line 1995: setp touches nzcv. Are we making sure that nzcv isn't corrupted through this instruction?
Might be worth having a small unit test.

lioncash: Good catch, updated it to save NZCV

PR #5357, opened 2026-03-10
Config: Enable Dynamic L1 and Disabled L2 caches by default

Sonicadvance1:
Dramatically reduces memory consumption of FEX's per-thread lookup
structures. Primarily because L2 cache entirely goes away which can end
up reaching hundreds of megabytes or over a gigabyte of memory in some
cases, but also because L1 cache dynamically scales based on load.

Useful for conserving memory on systems 16GB of RAM or less and
are UMA, like Asahi users inside of muvm.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -195,7 +195,7 @@
       },
       "DisableL2Cache": {
         "Type": "bool",
-        "Default": "false",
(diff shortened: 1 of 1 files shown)


neobrain:
Any reason to single out Steam here instead of enabling the settings by default for other deployments too?

Sonicadvance1:
After discussions, we have decided to just flip FEX's defaults entirely.

PR #5362, opened 2026-03-12
FEX: Disable THP on key allocations that consume memory

Sonicadvance1:
Disables THP on some key locations that are fairly sparse
- rpmalloc
  - This is the big one as this allocates some heavy sparse buffers.
- CallRet stacks
  - These get in the hundreds of megabytes, while not being sparse they trend towards only using a handful of pages and ballooning to 2MB per thread is quite heavy.
- Lookup cache
  - L1 specifically gets hit here which adds a decent chunk of overhead due to sparsity.

Win32 for all of these also aren't handled, but that will need to be a followup.

### DIFF
--- a/FEXCore/Source/Utils/AllocatorHooks.cpp
+++ b/FEXCore/Source/Utils/AllocatorHooks.cpp
@@ -109,6 +109,9 @@ static void* FEX_rp_mmap(size_t size, size_t alignment, size_t* offset, size_t* 
 #define PR_SET_VMA_ANON_NAME 0
 #endif
     prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, ptr, map_size, global_config.page_name);
+
+    // Disable HUGEPAGE on allocation from rpmalloc.
+    madvise(ptr, map_size, MADV_NOHUGEPAGE);
   }
 
   if (ptr == nullptr) {
--- a/FEXCore/include/FEXCore/Utils/AllocatorHooks.h
+++ b/FEXCore/include/FEXCore/Utils/AllocatorHooks.h
@@ -27,6 +27,11 @@ enum class ProtectOptions : uint32_t {
 };
 FEX_DEF_NUM_OPS(ProtectOptions)
 
+enum class THPControl {
+  Enable,
+  Disable,
+};
+
 #ifdef _WIN32
 inline void* VirtualAlloc(void* Base, size_t Size, bool Execute = false, bool Commit = true) {
   // Allocate top-down to avoid polluting the lower VA space, as even on 64-bit some programs (i.e. LuaJIT) require allocations below 4GB.
@@ -83,7 +88,7 @@ inline bool VirtualProtect(void* Ptr, size_t Size, ProtectOptions options) {
 }
 
 inline void VirtualName(const char*, void*, size_t) {}
-inline void VirtualTHP(void* Ptr, size_t Size) {}
+inline void VirtualTHPControl(void* Ptr, size_t Size, THPControl Control) {}
 
 #else
 using MMAP_Hook = void* (*)(void*, size_t, int, int, int, off_t);
@@ -124,8 +129,8 @@ inline bool VirtualProtect(void* Ptr, size_t Size, ProtectOptions options) {
   return ::mprotect(Ptr, Size, prot) == 0;
 }
 
-inline void VirtualTHP(void* Ptr, size_t Size) {
-  ::madvise(Ptr, Size, MADV_HUGEPAGE);
+inline void VirtualTHPControl(void* Ptr, size_t Size, THPControl Control) {
+  ::madvise(Ptr, Size, Control == THPControl::Enable ? MADV_HUGEPAGE : MADV_NOHUGEPAGE);
 }
 
 #endif
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
@@ -189,6 +189,9 @@ FEX::HLE::ThreadStateObject* ThreadManager::CreateThread(uint64_t InitialRIP, ui
 
   FEXCore::Allocator::VirtualName("FEXMem_CallRetStacks", reinterpret_cast<void*>(AllocBase), CALLRET_STACK_ALLOC_SIZE);
 
+  // Disable HUGEPAGE on callret stacks.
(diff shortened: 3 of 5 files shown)


FEXCore/include/FEXCore/Utils/AllocatorHooks.h
```diff
@@ -83,7 +83,7 @@ inline bool VirtualProtect(void* Ptr, size_t Size, ProtectOptions options) {
 }
 
 inline void VirtualName(const char*, void*, size_t) {}
-inline void VirtualTHP(void* Ptr, size_t Size) {}
+inline void VirtualTHP(void* Ptr, size_t Size, bool Disable = false) {}
```
neobrain: Could you add a dedicated interface for disabling THP instead? Reading "disable THP" is much easier than requiring people to jump to the definition to understand what the `true` parameter does.

neobrain: Having a separate interface also allows for documenting *when* THP should be explicitly disabled. IIUC the logic here is "any time we mmap relatively small amounts of memory [particularly if done multiple times]"?

Sonicadvance1: Changed the argument over to an enum class so it is explicit.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -189,6 +189,10 @@ FEX::HLE::ThreadStateObject* ThreadManager::CreateThread(uint64_t InitialRIP, ui
 
   FEXCore::Allocator::VirtualName("FEXMem_CallRetStacks", reinterpret_cast<void*>(AllocBase), CALLRET_STACK_ALLOC_SIZE);
 
+  // Disable HUGETLB on callret stacks.
+  FEXCore::Allocator::VirtualTHPControl(reinterpret_cast<void*>(AllocBase), CALLRET_STACK_ALLOC_SIZE,
+                                        FEXCore::Allocator::THPControl::RequestNoHugeTLB);
```
neobrain: Was there a reason you're opting for a separate enum over a new function here The "No" barely stands out in this overly excessive amount of noise, which defeats the intent of making the functional difference clear. (I'd perhaps be less picky about this if the flag didn't entirely reverse what the function does.)

A separate function would avoid that problem since you don't need to mention all the namespaces twice.

---

Perhaps a better alternative: Move back to the previous boolean-based interface but use a function name that makes its semantics clear, like SetXYZEnabled.

Sonicadvance1: For the follow-up PR for win32, the THP control needs frontend functions to replace the implementation so less function pointers being passed around means my brain spins less.

If you want me to rename the function and just use a bool again, that's also fine. Named Enums feel nicer and semantically matches `MADV_HUGEPAGE` and `MADV_NOHUGEPAGE` semantics. Especially when we have so many PR discussions around naked values as arguments to functions.

neobrain: Fair - if we want to stick with the enum, can we rename the values to simply "Enable"/"Disable"? That's much shorter and doesn't bury the distinguishing "No" in the middle, which I think makes it as good as renaming the function to "VirtualSetTHPEnabled" then (probably better even for the reasons you mentioned).

Side note, not exactly sure what the TLB suffix intended to communicate before: From a quick glance, THP (`MADV_HUGEPAGE` in madvise) and TLB (`MAP_HUGETLB` in mmap) seem to be two related but entirely distinct categories?

Sonicadvance1: > Side note, not exactly sure what the TLB suffix intended to communicate before: From a quick glance, THP (MADV_HUGEPAGE in madvise) and TLB (MAP_HUGETLB in mmap) seem to be two related but entirely distinct categories?

They're the same thing, the enums are just named differently.

neobrain: Erm, they each have a dedicated chapter in the kernel docs, so I don't think we should throw the terms around interchangeably:
https://www.kernel.org/doc/html/latest/admin-guide/mm/transhuge.html
https://www.kernel.org/doc/html/latest/admin-guide/mm/hugetlbpage.html

You've probably presumably looked longer at this than me but we should stick to one of the terms if we don't care strongly about the difference (and if we do care about the difference, we should choose deliberately)

Sonicadvance1: Changed the enum to enable disable.

FEXCore/Source/Utils/AllocatorHooks.cpp
```diff
@@ -109,6 +109,9 @@ static void* FEX_rp_mmap(size_t size, size_t alignment, size_t* offset, size_t*
 #define PR_SET_VMA_ANON_NAME 0
 #endif
     prctl(PR_SET_VMA, PR_SET_VMA_ANON_NAME, ptr, map_size, global_config.page_name);
+
+    // Disable HUGETLB on allocation from rpmalloc.
+    madvise(ptr, map_size, MADV_NOHUGEPAGE);
```
neobrain: Referring to HugeTLB when this is actually THP.

Source/Tools/LinuxEmulation/LinuxSyscalls/ThreadManager.cpp
```diff
@@ -189,6 +189,9 @@ FEX::HLE::ThreadStateObject* ThreadManager::CreateThread(uint64_t InitialRIP, ui
 
   FEXCore::Allocator::VirtualName("FEXMem_CallRetStacks", reinterpret_cast<void*>(AllocBase), CALLRET_STACK_ALLOC_SIZE);
 
+  // Disable HUGETLB on callret stacks.
+  FEXCore::Allocator::VirtualTHPControl(reinterpret_cast<void*>(AllocBase), CALLRET_STACK_ALLOC_SIZE, FEXCore::Allocator::THPControl::Disable);
```
neobrain: Another instance of referring to HugeTLB when this is actually THP.

Sonicadvance1: Fixed.

FEXCore/Source/Utils/AllocatorHooks.cpp
Sonicadvance1: Fixed.

PR #5364, opened 2026-03-12
Win32: Enable support for virtual naming and THP control 

Sonicadvance1:
Allows WTF to work (mostly) with Wine by letting us VirtualName things,
and also allows madvise control of THP, which significantly cuts back
memory usage.

This works around the problem of Wine not giving us control of this by
using raw syscalls when wine is detected.

### DIFF
--- a/Source/Windows/Common/Allocator.cpp
+++ b/Source/Windows/Common/Allocator.cpp
@@ -1,0 +1,243 @@
+// SPDX-License-Identifier: MIT
+#include <FEXCore/Utils/AllocatorHooks.h>
+#include <FEXCore/Utils/TypeDefines.h>
+
+#include <array>
+#include <chrono>
+#include <libloaderapi.h>
+#include <sysinfoapi.h>
+#include <synchapi.h>
+#include <windef.h>
+#include <winternl.h>
+#include <winnt.h>
+#include <wine/debug.h>
+
+namespace FEX::Windows::Allocator {
+#define PR_SET_VMA 0x53564d41
+#define PR_SET_VMA_ANON_NAME 0
+
+#define MADV_HUGEPAGE 14
+#define MADV_NOHUGEPAGE 15
+
+namespace Trampoline {
+  struct madvise_data {
+    const void* addr;
+    size_t size;
+    int advise;
+  };
+
+  struct prctl_data {
+    int op;
+    uint64_t attr;
+    const void* addr;
+    size_t size;
+    const char* name;
+    uint64_t ret;
+  };
+
+  __attribute__((naked)) uint64_t wine_prctl(prctl_data* d) {
+    asm volatile(
+      R"(
+  .globl wine_prctl_begin
+  wine_prctl_begin:
+    mov x19, x0;
+    mov x8, 167; // prctl
+    ldr x0, [x19]; // op
+    ldp x1, x2, [x19, %[attr_offset]]; // {attr, addr}
+    ldp x3, x4, [x19, %[size_offset]]; // {size, name}
+    svc #0;
+    str x0, [x19, %[ret_offset]];
+    // Tell wine it was all groovy.
+    mov x0, 0;
+    ret;
+
+  .globl wine_prctl_end
+  wine_prctl_end:
+  )"
+      :
+      : [attr_offset] "i"(offsetof(prctl_data, attr)), [size_offset] "i"(offsetof(prctl_data, size)), [ret_offset] "i"(offsetof(prctl_data, ret))
+      : "memory");
+  };
+
+  __attribute__((naked)) uint64_t wine_madvise(madvise_data* d) {
+    asm volatile(R"(
+  .globl wine_madvise_begin
+  wine_madvise_begin:
+    mov x8, 233; // madvise
+    ldr x2, [x0, %[advise_offset]]; // advise
+    ldp x0, x1, [x0]; // {addr, size}
+    svc #0;
+    // Tell wine it was all groovy.
+    mov x0, 0;
+    ret;
+
+  .globl wine_madvise_end
+  wine_madvise_end:
+  )" ::[advise_offset] "i"(offsetof(madvise_data, advise))
+                 : "memory");
+  }
+
+  extern "C" uint64_t wine_madvise_begin;
+  extern "C" uint64_t wine_madvise_end;
+
+  void* const wine_madvise_begin_loc = &wine_madvise_begin;
+  void* const wine_madvise_end_loc = &wine_madvise_end;
+
+  extern "C" uint64_t wine_prctl_begin;
+  extern "C" uint64_t wine_prctl_end;
+
+  void* const wine_prctl_begin_loc = &wine_prctl_begin;
+  void* const wine_prctl_end_loc = &wine_prctl_end;
+
+  extern NTSTATUS(WINAPI* __wine_unix_call_dispatcher)(uint64_t, unsigned int, void*);
+  decltype(__wine_unix_call_dispatcher) WineUnixCall;
+
+  enum unix_function_indexes {
+    INDEX_PRCTL = 0,
+    INDEX_MADVISE = 1,
+    INDEX_MAX,
+  };
+  static std::array<void*, INDEX_MAX> unix_functions {};
+
+  static uint64_t wine_prctl_trampoline(int op, uint64_t attr, const void* addr, size_t size, const char* name) {
+    prctl_data d {
+      .op = op,
+      .attr = attr,
+      .addr = addr,
+      .size = size,
+      .name = name,
+    };
+    WineUnixCall(reinterpret_cast<uint64_t>(unix_functions.data()), INDEX_PRCTL, &d);
+    return d.ret;
+  }
+
+  static uint64_t wine_madvise_trampoline(const void* addr, size_t size, int advice) {
+    madvise_data d {
+      .addr = addr,
+      .size = size,
+      .advise = advice,
+    };
+    WineUnixCall(reinterpret_cast<uint64_t>(unix_functions.data()), INDEX_MADVISE, &d);
+    return 0;
+  }
+
+  void VirtualName(const char* Name, const void* Ptr, size_t Size) {
+    static bool Supports {true};
+    if (Supports) {
+      auto Result = wine_prctl_trampoline(PR_SET_VMA, PR_SET_VMA_ANON_NAME, Ptr, Size, Name);
+      if (Result != 0) {
+        // Disable any additional attempts.
+        Supports = false;
+      }
+    }
+  }
+
+  void VirtualTHPControl(const void* Ptr, size_t Size, FEXCore::Allocator::THPControl Control) {
+    wine_madvise_trampoline(Ptr, Size, Control == FEXCore::Allocator::THPControl::Disable ? MADV_NOHUGEPAGE : MADV_HUGEPAGE);
+  }
+} // namespace Trampoline
+
+// This code path will eventually crash once Wine and the kernel implements `userspace syscall dispatch`.
+// FEX will need to switch over to using WINE's unixlib syscall approach then.
+// See `SetupHooks` for why unixlib doesn't work today.
+namespace Illegal {
+  __attribute__((naked)) uint64_t prctl(int op, uint64_t attr, const void* addr, size_t size, const char* Name) {
+    asm volatile(R"(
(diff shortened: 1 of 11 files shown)


Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,235 @@
...
+  const auto Sym = GetProcAddress(ntdll, "__wine_unix_call_dispatcher");
+
+  if (!Sym) {
+    return;
+  }
+
+  FEXCore::Allocator::HookPtrs Ptrs {};
+
+  // Wine wants us to use trampolines through unixlib.
+  // This doesn't seem to work for some reason so skip it for now.
```
neobrain: Noticing how much code is attached to this, can we either fix this before merging or drop the dead code from the PR? Not sure what value it adds since there's no external references or similar for future orientation.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+  }
+
+  void VirtualTHPControl(void* Ptr, size_t Size, FEXCore::Allocator::THPControl Control) {
+    wine_madvise_trampoline(Ptr, Size, Control == FEXCore::Allocator::THPControl::Disable ? MADV_NOHUGEPAGE : MADV_HUGEPAGE);
+  }
+} // namespace Trampoline
+
+namespace Illegal {
```
neobrain, line 143: Needs an explanation what's "illegal" about it

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,235 @@
...
+      }
+    }
+  }
+
+  void VirtualTHPControl(void* Ptr, size_t Size, FEXCore::Allocator::THPControl Control) {
+    madvise(Ptr, Size, Control == FEXCore::Allocator::THPControl::Disable ? MADV_NOHUGEPAGE : MADV_HUGEPAGE);
+  }
+} // namespace Illegal
+
+// NTSTATUS __wine_unix_call_dispatcher( unixlib_handle_t, unsigned int, void * );
```
neobrain: This *looks* like a stale comment, but the intent seems to be to provide context. Maybe lead with a sentence? Hard to say what you're going for.

Also consider moving the comment into the function, since you're not documenting SetupHooks but an implementation detail of it.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,235 @@
...
+void SetupHooks(HMODULE ntdll) {
+  const auto Sym = GetProcAddress(ntdll, "__wine_unix_call_dispatcher");
+
+  if (!Sym) {
+    return;
+  }
+
+  FEXCore::Allocator::HookPtrs Ptrs {};
+
+  // Wine wants us to use trampolines through unixlib.
```
neobrain: In what sense does Wine "want" us to do things? Can you provide an external reference for context?

Source/Windows/Common/CallRetStack.h
```diff
@@ -25,6 +25,12 @@ void InitializeThread(FEXCore::Core::InternalThreadState* Thread) {
   const void* CallRetStackAlloc = ::VirtualAlloc(
     nullptr, FEXCore::Core::InternalThreadState::CALLRET_STACK_SIZE + 2 * FEXCore::Utils::FEX_PAGE_SIZE, MEM_RESERVE, PAGE_NOACCESS);
 
+  FEXCore::Allocator::VirtualName("FEXMem_CallRetStacks", const_cast<void*>(CallRetStackAlloc),
+                                  FEXCore::Core::InternalThreadState::CALLRET_STACK_SIZE + 2 * FEXCore::Utils::FEX_PAGE_SIZE);
+  FEXCore::Allocator::VirtualTHPControl(const_cast<void*>(CallRetStackAlloc),
```
neobrain: Why the `const_cast` instead of just changing CallRetStackAlloc's type?

Alternatively, maybe worth considering if VirtualTHPControl and VirtualName should take `const void*` pointers instead.

Sonicadvance1: Changed the function signatures.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Updated the comment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Moved the comment next to where we are getting the symbol pointer. As the code uses the symbol if it exists. The function definition is opaque otherwise.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Added a comment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Updated the comment to give more context as to why both code paths exist.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,244 @@
...
+  // - uint32_t is just an index in to that
+  // - void* is the user provided pointer, gets loaded in to x0 in for the unix_function called.
+  // - Return value - SUCCESS or other error.
+  //
+  // If this symbol doesn't exist, then we aren't running under WINE.
+  const auto Sym = GetProcAddress(ntdll, "__wine_unix_call_dispatcher");
+
+  if (!Sym) {
+    return;
+  }
```
neobrain: This is needlessly detached from the code that actually uses `Sym`. Should be moved into the `constexpr (false)` block below.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+  FEXCore::Allocator::HookPtrs Ptrs {};
+
+  // Wine has a strong desire to force us to use unixlib for calling helper routines.
+  // This is required on their end to use `userspace syscall dispatch` to capture rogue applications doing raw Windows syscalls.
```
neobrain: Thanks for adding the extra information, though I'm not sure why you replaced "wants" with "strong desire to force us" in this particular sentence. It's even more vague than the initial wording.

```suggestion
  // Wine requires us to use unixlib for calling helper routines that <WHY>.
  // It needs this for `userspace syscall dispatch` to capture rogue applications doing raw Windows syscalls.
```

Adding a placeholder there since it's still unclear why VirtualTHPControl needs this indirection whereas the vast majority of "helper routines" that we use don't need it. Is the key distinguishing factor the direct use of Linux system calls?

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,244 @@
...
+  // If FEX doesn't use unixlib, then when WINE and a kernel implements this, then our "illegal" path will start crashing.
+  //
+  // The problem is that attempting to use `unixlib` gets captured by FEX's `Call Checker` so can't be used until a workaround is
+  // implemented. Keep this code around since we will need to switch to it eventually.
```
neobrain: Do I understand correctly that Wine *currently* does not require this, but a future version potentially will?

Sonicadvance1: A future version definitely will, it'll happen at some point regardless of our desires. We just have a bit of time before the Arm64 kernel supports the feature.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Moved.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Changed the words.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  void VirtualTHPControl(const void* Ptr, size_t Size, FEXCore::Allocator::THPControl Control) {
+    wine_madvise_trampoline(Ptr, Size, Control == FEXCore::Allocator::THPControl::Disable ? MADV_NOHUGEPAGE : MADV_HUGEPAGE);
+  }
+} // namespace Trampoline
+
+// This code path will eventually crash once Wine and the kernel implements `userspace syscall dispatch`.
+// At which point FEX needs to switch over to using WINE's unixlib syscall approach.
```
neobrain: This is not a sentence.
```suggestion
// FEX will need to switch over to using WINE's unixlib syscall approach then.
```

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+  // If this symbol doesn't exist, then we aren't running under WINE.
+  const auto Sym = GetProcAddress(ntdll, "__wine_unix_call_dispatcher");
+
+  if (!Sym) {
+    return;
+  }
+
+  FEXCore::Allocator::HookPtrs Ptrs {};
+
+  // Wine requires us to use unixlib for calling helper routines that use Linux syscalls.
```
neobrain:
```suggestion
  // Wine will soon require us to use unixlib for calling helper routines that use Linux syscalls.
```

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+  // It needs this for `userspace syscall dispatch` to capture rogue applications doing raw Windows syscalls.
+  // If FEX doesn't use unixlib, then when WINE and a kernel implements this, then our "illegal" path will start crashing.
+  //
+  // The problem is that attempting to use `unixlib` gets captured by FEX's `Call Checker` so can't be used until a workaround is
+  // implemented. Keep this code around since we will need to switch to it eventually.
```
neobrain:
```suggestion
  // This code currently conflicts with FEX's `Call Checker` since the latter captures uses of `unixlib`.
  // We hence have to stick to the `Illegal::` functions until a workaround is implemented.
```

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+  }
+} // namespace Illegal
+
+void SetupHooks(HMODULE ntdll) {
+  // If this symbol doesn't exist, then we aren't running under WINE.
+  const auto Sym = GetProcAddress(ntdll, "__wine_unix_call_dispatcher");
+
+  if (!Sym) {
+    return;
+  }
```
neobrain, line 184: Seems like only the comment got moved, not the actual code. This can all go into the `constexpr (false)` branch.

Source/Windows/Common/Allocator.cpp
```diff
@@ -0,0 +1,243 @@
...
+    }
+  }
+
+  void VirtualTHPControl(const void* Ptr, size_t Size, FEXCore::Allocator::THPControl Control) {
+    madvise(Ptr, Size, Control == FEXCore::Allocator::THPControl::Disable ? MADV_NOHUGEPAGE : MADV_HUGEPAGE);
+  }
+} // namespace Illegal
+
+void SetupHooks(HMODULE ntdll) {
+  // If this symbol doesn't exist, then we aren't running under WINE.
```
neobrain, line 179: Do we have supported use cases that run this code outside of Wine?

Source/Windows/Common/Allocator.cpp
neobrain: Thanks, adding suggestions to make this clearer from the comment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: This can not go all in to the constexpr branch, as we need to disable both code paths if wine isn't detected. Like for example, we are executing inside of a Windows OS environment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Yes. Windows.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Replaced with your comment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Replaced with your comment.

Source/Windows/Common/Allocator.cpp
Sonicadvance1: Replaced with your comment.

PR #5365, opened 2026-03-13
Config: Fix issue with config overrides

Sonicadvance1:
Accidentally was checking for Config override in the combination of portable config and `FEX_APP_CONFIG_LOCATION`.

Fixes an early crash in PV.


Sonicadvance1:
Yes, I know the config code here is in a bit of a hot mess. I'm going to work on resolving that soon. Just...don't want to keep everyone broken while it happens.

PR #5368, opened 2026-03-13
CodeCache: Support ELF relocations

neobrain:
## Overview

Today, FEX can already cache ELF/PE binaries with relocations. However, since caches are loaded at `mmap` time, any loaded cache data will immediately be invalidated when ld/Wine start applying relocations to the data. In practice, this currently renders code caching ineffective for most 32-bit applications.

To fix this shortcoming, this PR builds on the basic infrastructure laid out in https://github.com/FEX-Emu/FEX/pull/5123 and adds major elements required for ELF relocation support: New ELF parsing code, a mechanism to delay cache loads, and a way to declare uncacheable blocks. 

FEXOfflineCompiler is furthermore changed to apply ELF relocations before cache generation, which ensures that x86 code seen at compile time matches the code at runtime.GuestRelocationType::Skip

Together, this enables effective code caching in Portal 1 (32-bit Linux-native).

### Delayed cache loading

When `ld` loads ELF files, code sections that would normally be read-only will be `mmap`ped as *writable* so that `ld` can apply relocations. Once done, the code section is `mprotect`ed as read-only. This behavior is triggered by the DT_TEXTREL/DF_TEXTREL entries in the ELF headers.

Normally FEX loads code caches at `mmap` time. But now if it finds those TEXTREL entries, the code cache load will be delayed until ld's mprotect call. If multiple code sections exist, their cache load will be triggered at once since per-section tracking would carry a big complexity cost while not making any difference in practice.

This heuristic also matches what Wine does when loading PE files.

### Uncacheable blocks

ELF has more relocation types than PE, and FEX can't handle some of them yet. These gaps may be closed in the future if it's deemed worth the effort, but for now the simplest thing is to just omit affected blocks from caching.

A new relocation type (`GuestRelocationType::Skip`) is introduced to signal this. When Core encounters this, it will stop compilation and discard any previously compiled block data. I took the first approach that worked here, so I'd welcome input for better ways to implement it.

### `robin_map` bug

UPDATE: Dropped this patch ca4776ddbcdbb02ef11ea99da553d763bfaece42 for now since I couldn't reproduce the problem in the latest iteration of the PR.

It seems that the large number of relocations used by Portal's `engine.so` triggers a bug in `robin_map`: I observed that `Relocations.contains(X)` would return `false` even though X was encountered when manually iterating through the map. Of course, this bug quickly manifests in runtime crashes due to wrong code generation.

This is highly concerning and should be looked into more closely since we use `robin_map` in many other places. For now, I just moved all relocation data to use `fextl::unordered_map` instead to unblock code caching work.

### DIFF
--- a/Source/Tools/CommonTools/Linux/Utils/ELFParser.h
+++ b/Source/Tools/CommonTools/Linux/Utils/ELFParser.h
@@ -461,6 +461,12 @@ private:
   static std::optional<FEXCore::GuestRelocationType> ClassifyRelocation32(uint32_t Type) {
     if (Type == R_386_RELATIVE || Type == R_386_32) {
       return FEXCore::GuestRelocationType::Rel32;
+    } else if (Type == R_386_PC32) {
+      // Currently not handled
+      return FEXCore::GuestRelocationType::Skip;
+    } else if (Type == R_386_TLS_TPOFF) {
+      // Currently not handled
+      return FEXCore::GuestRelocationType::Skip;
     }
     return std::nullopt;
   }
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -127,6 +127,7 @@ public:
 
   void ExecuteThread(FEXCore::Core::InternalThreadState* Thread) override;
 
+  bool CheckIfBlockIsCacheable(FEXCore::Core::InternalThreadState&, uint64_t GuestRIP, uint64_t MaxInst) override;
   void CompileRIP(FEXCore::Core::InternalThreadState* Thread, uint64_t GuestRIP) override;
   void CompileRIPCount(FEXCore::Core::InternalThreadState* Thread, uint64_t GuestRIP, uint64_t MaxInst) override;
 
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -502,6 +502,10 @@ static void IRDumper(FEXCore::Core::InternalThreadState* Thread, IR::IREmitter* 
   fextl::fmt::print(FD, "IR-ShouldDump-{} 0x{:x}:\n{}\n@@@@@\n", NewIR.PostRA() ? "post" : "pre", GuestRIP, out.str());
 };
 
+bool ContextImpl::CheckIfBlockIsCacheable(FEXCore::Core::InternalThreadState& Thread, uint64_t GuestRIP, uint64_t MaxInst) {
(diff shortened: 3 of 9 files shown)


Sonicadvance1:
> robin_map bug
> 
> It seems that the large number of relocations used by Portal's engine.so triggers a bug in robin_map: I observed that Relocations.contains(X) would return false even though X was encountered when manually iterating through the map. Of course, this bug quickly manifests in runtime crashes due to wrong code generation.
> 
> This is highly concerning and should be looked into more closely since we use robin_map in many other places. For now, I just moved all relocation data to use fextl::unordered_map instead to unblock code caching work.

Was this before or after we switched the robin_map implementation from `tsl::robin_map` to `ankerl::unordered_dense::map`? Would be good to know which one to throw blaim on to.

neobrain:
> Was this before or after we switched the robin_map implementation from tsl::robin_map to ankerl::unordered_dense::map? Would be good to know which one to throw blaim on to.

Oh right, I had forgotten that we switched the underlying implementation. I found the issue on 27 February, so that's after the switch.

Source/Tools/CommonTools/Linux/Utils/ELFParser.h
```diff
@@ -236,6 +240,181 @@ struct ELFParser {
...
+      return false;
+    }
+
+    auto phdr_it = std::ranges::find_if(phdrs, [](auto& phdr) { return phdr.p_type == PT_DYNAMIC; });
+    if (phdr_it == phdrs.end()) {
+      return false;
+    }
+    auto& phdr = *phdr_it;
+
+    if (type == ::ELFLoader::ELFContainer::TYPE_X86_32) {
```
Sonicadvance1, line 260: This can be deduplicated with a cute templated lambda or something that takes a typename for `{Elf32_Dyn,Elf64_Dyn}` to remove code duplication.

neobrain:
Dropping the robin_map change (ca4776ddbcdbb02ef11ea99da553d763bfaece42) for now since I can't reproduce the original problem anymore. Now that we know the problem, it's easy to keep an eye on the situation and dig out the patch again if needed.

Also for reference, here are some pending issues in Portal 1 that are unrelated to the PR and which require follow-up fixes:
* First launch will sometimes fail code cache validation in libLLVM (needs debugging)
* Starting a new game will fail code cache validation in steamclient.so (fix found, needs cleanup)

Source/Tools/CommonTools/Linux/Utils/ELFParser.h
neobrain: Done :+1:

PR #5371, opened 2026-03-15
Syscalls: Fixes crash in ELF parsing code

Sonicadvance1:
When an application maps a file as PROT_NONE, we can't check if it is an ELF. Was causing a crash in `Cisco Packet Tracer`.


Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -504,6 +504,11 @@ SyscallHandler::TrackMmap(FEXCore::Core::InternalThreadState* Thread, uint64_t a
 #if defined(ASSERTIONS_ENABLED) && ASSERTIONS_ENABLED
         CheckForElfFile = true;
 #endif
+        if (!(prot & PROT_READ)) {
+          // If this isn't mapped readable then it can't be checked.
```
neobrain, line 508: This is the wrong place to put this - if a MappedFile is created, it must be filled with valid ELF header data, which is now skipped entirely. The MappedFile is now put in a weird half-initialized state that the rest of the code is not written in mind for.

Could you move the check up into the `if (Inserted)` check instead so that MappedFile is always in a valid state?

PR #5372, opened 2026-03-16
Tests/ASM: Add FISTTP unit tests for 16, 32, 64-bit as well as negatives

### LINKED ISSUE #779: Write x87 tests for FIST*T*P, Rounding Control, Precision Control

CxnYusuf:
FISTTP was not covered by previous Unit-Tests. Added 6 Tests covering 16, 32 and 64-bit integer store variants as well as their negative counterparts to verify truncation towards 0.

Closes #779

### DIFF
--- a/unittests/ASM/X87/FISTTP_16bit.asm
+++ b/unittests/ASM/X87/FISTTP_16bit.asm
@@ -1,0 +1,29 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0",
+    "RBX": "1"
+  }
+}
+%endif
+
+; Test FISTTP with 16-bit integer store
+; FISTTP always truncates toward zero, ignoring rounding control word
+
+finit
+fld qword [rel .value]
+
+; Convert to int16 using truncation - 1.9 should become 1, not 2
+fisttp word [rel .result]
+
+fstsw ax
+and rax, 1
+
+; Load result to verify truncation worked
+movzx rbx, word [rel .result]
+
+hlt
+
+align 4096
+.value: dq 1.9
+.result: dw 0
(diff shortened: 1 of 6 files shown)


Sonicadvance1:
Is this an AI based submission?

CxnYusuf:
> Is this an AI based submission?

no sorry? i am very much a beginner and tried to follow the existing FIST Unit tests as a guideline, can you tell me what's wrong with it?

Sonicadvance1:
We just had an AI generated PR that targeted this same issue, so the two pull requests within close temporal proximity of each other and an account with no history raises red flags.

If AI wasn't used then it would be fine.

CxnYusuf:
> We just had an AI generated PR that targeted this same issue, so the two pull requests within close temporal proximity of each other and an account with no history raises red flags.
> 
> If AI wasn't used then it would be fine.

oh okay gotcha, that makes sense. i just looked for "good first issue" and unit tests seemed like they would be easier to get into. If you'd like to verify, im also on discord! sorry if this is a bother, but is everything else with the tests fine? I really do want to learn

unittests/ASM/X87/FISTTP_16bit.asm
```diff
@@ -0,0 +1,29 @@
...
+; Test FISTTP with 16-bit integer store
+; FISTTP always truncates toward zero, ignoring rounding control word
+
+finit
+fld qword [rel .value]
+
+; Convert to int16 using truncation - 1.9 should become 1, not 2
+fisttp word [rel .result]
+
+fstsw ax
```
Sonicadvance1, line 19: In each of these unittests you're storing the status word and then masking bit 0. I am /presuming/ this is for testing the result of C1 to ensure the value was rounded in the correct direction.

The problem is that C1 in the status word is at bit-offset 9, so it should be getting masked by 0x100 in all of these tests.

Sonicadvance1: Oh no, nevermind. Since you're copying from the tests this is actually checking for IE.

Sonicadvance1:
Seems reasonable enough to me for unit tests. Thanks for the tests!

Sonicadvance1 (approved):

PR #5383, opened 2026-03-19
FEXGetConfig: Test for showing fault granularity

Sonicadvance1:
Useful for seeing if behaviour has changed. Useful with the
`--tso-emulation-info` option to show hardware behaviour

Example of what it looks like.
Screenshot 2026-03-18 at 20 03 11

For ARMv8.0-a behaviour:
Screenshot 2026-03-18 at 20 10 14

### DIFF
--- a/Source/Tools/FEXGetConfig/Main.cpp
+++ b/Source/Tools/FEXGetConfig/Main.cpp
@@ -14,6 +14,8 @@
 #include <filesystem>
 #include <string>
 #include <sys/prctl.h>
+#include <sys/signal.h>
+#include <ucontext.h>
 
 namespace {
 struct TSOEmulationFacts {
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
I'll clean this up tomorrow, just wanted to slap it in a PR quick.

Sonicadvance1:
There we go, cleaned up and added 128-bit as well, which could become useful in the future.

Source/Tools/FEXGetConfig/Main.cpp
```diff
@@ -94,6 +96,104 @@ TSOEmulationFacts GetTSOEmulationFacts() {
...
+  act.sa_sigaction = HandleSIGBUS;
+  act.sa_flags = SA_SIGINFO;
+  sigaction(SIGBUS, &act, &act);
+  void* ptr = mmap(nullptr, 4096, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+
+  auto test_fault = [](bool* FaultOffsets, auto AccessFunction, auto AccessArray) {
+    FaultArray = FaultOffsets;
+    for (size_t i = 0; i < 64; ++i) {
+      auto ptr_offset = (decltype(AccessArray))((uintptr_t)AccessArray + i);
+      AccessFunction(ptr_offset, 1);
```
neobrain: The data type is already fixed by the AccessFunction - why not pass in `AccessArray` as an `std::byte*` so we don't need to do the ptr_offset dance?

Source/Tools/FEXGetConfig/Main.cpp
```diff
@@ -94,6 +96,104 @@ TSOEmulationFacts GetTSOEmulationFacts() {
...
+    }
+  };
+
+  auto print_granule = [](const char* size, bool* FaultArray) {
+    std::string output {};
+    for (size_t i = 0; i < 64; ++i) {
+      if (i && (i % 16 == 0) && ((i + 1) != 64)) {
```
neobrain: `(i % 16 == 0)` and `((i + 1) != 64)` can't be true at the same time.

```suggestion
      if (i && (i % 16 == 0)) {
```

Source/Tools/FEXGetConfig/Main.cpp
```diff
@@ -94,6 +96,104 @@ TSOEmulationFacts GetTSOEmulationFacts() {
...
+        output += "'";
```
neobrain: Why not just space?

Source/Tools/FEXGetConfig/Main.cpp
Sonicadvance1: Sure.

Source/Tools/FEXGetConfig/Main.cpp
Sonicadvance1: Sure.

Source/Tools/FEXGetConfig/Main.cpp
Sonicadvance1: Fixed.

Source/Tools/FEXGetConfig/Main.cpp
neobrain: The weirdly elaborate cast is still there though

Sonicadvance1: Oop. Missed out on a bit of my weird.

PR #5384, opened 2026-03-19
Cmake: Default to release builds with a message

Sonicadvance1:
People keep forgetting to set this and have a worse experience. Default to a Release build, which ensures optimizations are enabled and assertions are disabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -172,6 +172,13 @@ set(TUNE_ARCH "generic" CACHE STRING "Override the Arch the build is tuned for")
 set(OVERRIDE_VERSION "detect" CACHE STRING "Override the FEX version")
 set(OVERRIDE_HASH "detect" CACHE STRING "Override the FEX git hash")
 
+get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
+if (NOT IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE Release
+        CACHE STRING "Choose the type of build." FORCE)
+    message(STATUS "No build type set, defaulting to a Release build")
+endif()
+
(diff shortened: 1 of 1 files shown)


lioncash (approved):

crueter (changes requested):
Generators that support multiconfig (Ninja included) may actually have issues with this. I would recommend doing something akin to:

```cmake
get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
if (NOT IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)
    set(CMAKE_BUILD_TYPE Release
        CACHE STRING "Choose the type of build." FORCE)
    message(STATUS "No build type set, defaulting to a Release build")
endif()
```

Sonicadvance1:
> Generators that support multiconfig (Ninja included) may actually have issues with this. I would recommend doing something akin to:
> 
> ```cmake
> get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
> if (NOT IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)
>     set(CMAKE_BUILD_TYPE Release
>         CACHE STRING "Choose the type of build." FORCE)
>     message(STATUS "No build type set, defaulting to a Release build")
> endif()
> ```



> Generators that support multiconfig (Ninja included) may actually have issues with this. I would recommend doing something akin to:
> 
> ```cmake
> get_property(IS_MULTI_CONFIG GLOBAL PROPERTY GENERATOR_IS_MULTI_CONFIG)
> if (NOT IS_MULTI_CONFIG AND NOT CMAKE_BUILD_TYPE)
>     set(CMAKE_BUILD_TYPE Release
>         CACHE STRING "Choose the type of build." FORCE)
>     message(STATUS "No build type set, defaulting to a Release build")
> endif()
> ```

Sure. Copy-pasted.

PR #5386, opened 2026-03-20
Add Regex Engine to FEX for config option loading

Sonicadvance1:
Rebased version of #5120.

### DIFF
--- a/FEXCore/Source/Utils/WildcardMatcher.cpp
+++ b/FEXCore/Source/Utils/WildcardMatcher.cpp
@@ -1,13 +1,12 @@
 // SPDX-License-Identifier: MIT
 
 #include <FEXCore/Utils/WildcardMatcher.h>
-#include "FEXCore/fextl/map.h"
+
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
Well then, this is trickier than expected. I'll need to pick this up tomorrow.

Sonicadvance1:
Okay, the issue was that it wasn't linking FEXCore_Base directly. Which all the other tools do. Bit anti-climatic.

neobrain:
Why do we have this and #5388, seemingly the same change?

Sonicadvance1:
> Why do we have this and #5388, seemingly the same change?

If you look at the commits, this is just a rebase and code compilation fix. The other PR is on top of this and changes the behaviour.

neobrain:
I did see the difference, but there's no apparent reason for it to be a third PR rather than a personal branch (or a push to the original PR) since it's ultimately the same change proposal regardless.

PR #5388, opened 2026-03-20
Config: Finish wiring up Regex app overrides 

Sonicadvance1:
This wasn't quite wired up exactly how we wanted it. It was previously
matching against the opaque file config handle, which can be anything.

Instead compare it to the appname that now gets passed over to it for
matching.

This allows us to do the following:
```
{
    "Config": {
        "ProfileStats": "1",
        "X87ReducedPrecision": "1",
        "TSOEnabled": "1",
        "VectorTSOEnabled": "0",
        "MemcpySetTSOEnabled": "0",
        "HalfBarrierTSOEnabled":"1",
        "MaxInst": "500",
        "Multiblock": "1"
    },
    "AppOverrides" : {
        "setup*" : {
            "Comment": [
                "292030 - The Witcher 3: Wild Hunt"
            ],
            "X87ReducedPrecision": "0"
        }
     }
}
```

Based on #5386 which needs to get merged first.

### DIFF
--- a/FEXCore/Source/Utils/WildcardMatcher.cpp
+++ b/FEXCore/Source/Utils/WildcardMatcher.cpp
@@ -3,35 +3,21 @@
 #include <FEXCore/Utils/WildcardMatcher.h>
 
 namespace FEXCore::Utils::Wildcard {
-class WildcardMatcher {
-private:
-  std::string_view pattern;
-  std::string_view text;
-
-public:
-  WildcardMatcher(std::string_view pattern, std::string_view text)
-    : pattern(pattern)
-    , text(text) {}
-
-  bool matchHelper(size_t p_idx, size_t t_idx);
-};
-
-bool WildcardMatcher::matchHelper(size_t p_idx, size_t t_idx) {
+static bool matchHelper(std::string_view pattern, std::string_view text, size_t p_idx, size_t t_idx) {
   if (p_idx == pattern.size()) {
     // Pattern exhausted
     return (t_idx == text.size());
   } else if (pattern[p_idx] == '*') {
     // Wildcard: Try matching zero characters, or one or more characters
-    return matchHelper(p_idx + 1, t_idx) || (t_idx < text.size() && matchHelper(p_idx, t_idx + 1));
+    return matchHelper(pattern, text, p_idx + 1, t_idx) || (t_idx < text.size() && matchHelper(pattern, text, p_idx, t_idx + 1));
   } else {
     // Match normally
-    return (t_idx < text.size() && pattern[p_idx] == text[t_idx] && matchHelper(p_idx + 1, t_idx + 1));
+    return (t_idx < text.size() && pattern[p_idx] == text[t_idx] && matchHelper(pattern, text, p_idx + 1, t_idx + 1));
   }
 }
 
 bool Matches(std::string_view pattern, std::string_view text) {
-  WildcardMatcher matcher(pattern, text);
-  return matcher.matchHelper(0, 0);
+  return matchHelper(pattern, text, 0, 0);
 }
 
 } // namespace FEXCore::Utils::Wildcard
--- a/Source/Common/Config.cpp
+++ b/Source/Common/Config.cpp
@@ -29,7 +29,8 @@
 
 namespace FEX::Config {
 namespace JSON {
-  static void LoadJSonConfig(const fextl::string& Config, std::function<void(const char* Name, const char* ConfigString)> Func) {
+  static void LoadJSonConfig(const fextl::string& Config, std::optional<fextl::string> AppName,
+                             std::function<void(const char* Name, const char* ConfigString)> Func) {
     fextl::vector<char> Data;
     if (!FEXCore::FileLoading::LoadFile(Data, Config)) {
       return;
@@ -52,26 +53,17 @@ namespace JSON {
     fextl::vector<const json_t*> ConfigBlocks;
     ConfigBlocks.push_back(ConfigList);
 
-    const json_t* OverrideList = json_getProperty(json, "AppOverrides");
-    if (OverrideList) {
-      for (const json_t* Item = json_getChild(OverrideList); Item != nullptr; Item = json_getSibling(Item)) {
-        const char* ItemName = json_getName(Item);
-        const json_t* OverrideNamedList = json_getProperty(OverrideList, ItemName);
+    if (AppName) {
+      const json_t* OverrideList = json_getProperty(json, "AppOverrides");
+      if (OverrideList) {
+        for (const json_t* Item = json_getChild(OverrideList); Item != nullptr; Item = json_getSibling(Item)) {
+          const char* AppPattern = json_getName(Item);
 
-        if (!ItemName) {
-          LogMan::Msg::EFmt("JSON file '{}': Couldn't get config name for an item", Config);
-          break;
-        }
-
-        if (!OverrideNamedList) {
-          LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ItemName);
-          break;
-        }
-
-        // Find the first match, then break
-        if (FEXCore::Utils::Wildcard::Matches(ItemName, Config)) {
-          ConfigBlocks.push_back(OverrideNamedList);
-          break;
+          // Find the first match, then break
+          if (FEXCore::Utils::Wildcard::Matches(AppPattern, *AppName)) {
+            ConfigBlocks.push_back(Item);
+            break;
+          }
         }
       }
     }
@@ -80,11 +72,6 @@ namespace JSON {
       for (const json_t* ConfigItem = json_getChild(ConfigBlock); ConfigItem != nullptr; ConfigItem = json_getSibling(ConfigItem)) {
         const char* ConfigName = json_getName(ConfigItem);
         const char* ConfigString = json_getValue(ConfigItem);
-
-        if (!ConfigName) {
-          LogMan::Msg::EFmt("JSON file '{}': Couldn't get config name for an item", Config);
-          return;
-        }
 
         if (!ConfigString) {
           LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ConfigName);
@@ -196,22 +183,24 @@ protected:
 
 class MainLoader final : public OptionMapper {
 public:
-  explicit MainLoader(FEXCore::Config::LayerType Type);
-  explicit MainLoader(fextl::string ConfigFile);
+  explicit MainLoader(FEXCore::Config::LayerType Type, std::optional<fextl::string> AppName = std::nullopt);
+  explicit MainLoader(fextl::string ConfigFile, std::optional<fextl::string> AppName = std::nullopt);
   explicit MainLoader(FEXCore::Config::LayerType Type, std::string_view ConfigFile);
 
   void Load() override;
 
 private:
+  std::optional<fextl::string> AppName;
   fextl::string Config;
 };
 
 class AppLoader final : public OptionMapper {
 public:
-  explicit AppLoader(const fextl::string& Filename, FEXCore::Config::LayerType Type);
+  explicit AppLoader(const fextl::string& AppName, FEXCore::Config::LayerType Type);
   void Load();
 
 private:
+  const fextl::string AppName;
   fextl::string Config;
 };
 
@@ -250,12 +239,14 @@ void OptionMapper::MapNameToOption(const char* ConfigName, const char* ConfigStr
 #include <FEXCore/Config/ConfigOptions.inl>
 }
 
-MainLoader::MainLoader(FEXCore::Config::LayerType Type)
+MainLoader::MainLoader(FEXCore::Config::LayerType Type, std::optional<fextl::string> AppName)
   : OptionMapper(Type)
+  , AppName {AppName}
   , Config {FEXCore::Config::GetConfigFileLocation(Type == FEXCore::Config::LayerType::LAYER_GLOBAL_MAIN)} {}
 
-MainLoader::MainLoader(fextl::string ConfigFile)
+MainLoader::MainLoader(fextl::string ConfigFile, std::optional<fextl::string> AppName)
   : OptionMapper(FEXCore::Config::LayerType::LAYER_MAIN)
+  , AppName {AppName}
   , Config {std::move(ConfigFile)} {}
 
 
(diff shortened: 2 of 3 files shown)


Source/Common/Config.cpp
```diff
@@ -48,21 +50,52 @@ namespace JSON {
...
+          }
+
+          if (!OverrideNamedList) {
+            LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ItemName);
+            break;
+          }
+
+          // Matches the first and then get out
+          if (FEXCore::Utils::Wildcard::Matches(ItemName, *AppName)) {
+            // Safe to assume its just pairs of strings at this point?
```
neobrain: What's this?

Source/Common/Config.cpp
```diff
@@ -48,21 +50,52 @@ namespace JSON {
...
+            LogMan::Msg::EFmt("JSON file '{}': Couldn't get config name for an item", Config);
+            break;
+          }
+
+          if (!OverrideNamedList) {
+            LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ItemName);
+            break;
+          }
+
+          // Matches the first and then get out
```
neobrain: Inconsistent grammar; better:

```suggestion
          // Find the first match, then break
```

Source/Common/Config.cpp
```diff
@@ -28,7 +29,8 @@
 
 namespace FEX::Config {
 namespace JSON {
-  static void LoadJSonConfig(const fextl::string& Config, std::function<void(const char* Name, const char* ConfigSring)> Func) {
+  static void LoadJSonConfig(const fextl::string& Config, std::optional<fextl::string> AppName,
+                             std::function<void(const char* Name, const char* ConfigSring)> Func) {
```
neobrain: ConfigSring is a typo

EDIT: Just realized this was already present before

Source/Common/Config.cpp
```diff
@@ -11,7 +12,7 @@
 #include <FEXCore/Utils/FileLoading.h>
 #include <FEXHeaderUtils/Filesystem.h>
 #include <FEXHeaderUtils/SymlinkChecks.h>
-
+#include <FEXCore/Utils/WildcardMatcher.h>
```
neobrain: Hmm didn't clang-format used to reorder headers? Anyway:


```suggestion

#include <FEXCore/Config/Config.h>
#include <FEXCore/fextl/fmt.h>
#include "FEXCore/fextl/vector.h"
@@ -11,7 +12,7 @@
#include <FEXCore/Utils/FileLoading.h>
#include <FEXCore/Utils/WildcardMatcher.h>
#include <FEXHeaderUtils/Filesystem.h>
#include <FEXHeaderUtils/SymlinkChecks.h>
```

unittests/APITests/WildcardMatcher.cpp
```diff
@@ -0,0 +1,49 @@
...
+  CHECK(Matches("*", "setup.json"));
+  CHECK(Matches("test*pattern", "test__pattern"));
+  CHECK(!Matches("setup.*", "setupjson"));
+  CHECK(Matches("setup*", "setup.json"));
+  CHECK(Matches("setup*", "setup/setup.json"));
+  CHECK(Matches("*setup*", "setup/setup.json"));
+}
+
+
+// Tests potential usage inside fex itself
```
neobrain: Comment doesn't really say anything:
```suggestion
```

Source/Common/Config.cpp
```diff
@@ -320,11 +358,11 @@ fextl::unique_ptr<FEXCore::Config::Layer> CreateGlobalMainLayer() {
...
 }
 
-fextl::unique_ptr<FEXCore::Config::Layer> CreateMainLayer(const fextl::string* File) {
+fextl::unique_ptr<FEXCore::Config::Layer> CreateMainLayer(const fextl::string* File, const std::optional<fextl::string> AppName) {
   if (File) {
-    return fextl::make_unique<MainLoader>(*File);
+    return fextl::make_unique<MainLoader>(*File, AppName);
   } else {
-    return fextl::make_unique<MainLoader>(FEXCore::Config::LayerType::LAYER_MAIN);
+    return fextl::make_unique<MainLoader>(FEXCore::Config::LayerType::LAYER_MAIN, AppName);
```
neobrain: Parameter should be non-const (it's passed by value anyway) and moved into callees:
```suggestion
fextl::unique_ptr<FEXCore::Config::Layer> CreateMainLayer(const fextl::string* File, std::optional<fextl::string> AppName) {
if (File) {
    return fextl::make_unique<MainLoader>(*File, std::move(AppName));
} else {
    return fextl::make_unique<MainLoader>(FEXCore::Config::LayerType::LAYER_MAIN, std::move(AppName));
```

FEXCore/Source/Utils/WildcardMatcher.cpp
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+
+#include <FEXCore/Utils/WildcardMatcher.h>
+#include "FEXCore/fextl/map.h"
+namespace FEXCore::Utils::Wildcard {
+class WildcardMatcher {
+private:
+  std::string_view pattern;
+  std::string_view text;
+  fextl::map<std::pair<size_t, size_t>, bool> cache;
```
neobrain: Do we anticipate use cases where this will provide a speedup? The `*setup*` case we care about right now doesn't benefit from it, but it adds code bloat and is much more expensive than scanning linear memory in the general case.

(I briefly played around with this and couldn't actually find a pattern + string combination that would hit the cache)

FEXCore/Source/Utils/WildcardMatcher.cpp
```diff
@@ -0,0 +1,52 @@
...
+  else if (pattern[p_idx] == '*') {
+    // Try matching zero characters, or one or more characters
+    result = matchHelper(p_idx + 1, t_idx) || (t_idx < text.size() && matchHelper(p_idx, t_idx + 1));
+  }
+  // Match normally
+  else {
+    result = (t_idx < text.size() && pattern[p_idx] == text[t_idx] && matchHelper(p_idx + 1, t_idx + 1));
+  }
+  cache[key] = result;
+  return result;
```
neobrain: Removing the cache (see above), this simplifies down to:
```suggestion
  if (p_idx == pattern.size()) {
    // Pattern exhausted
    return (t_idx == text.size());
  } else if (pattern[p_idx] == '*') {
    // Wildcard: Try matching zero characters, or one or more characters
    return matchHelper(p_idx + 1, t_idx) || (t_idx < text.size() && matchHelper(p_idx, t_idx + 1));
  } else {
    // Match normally
    return (t_idx < text.size() && pattern[p_idx] == text[t_idx] && matchHelper(p_idx + 1, t_idx + 1));
  }
```

neobrain: (This also renders the WildcardMatcher class obsolete: matchHelper can just be a standalone function that takes the pattern and text as parameters)

Source/Common/Config.cpp
```diff
@@ -48,21 +50,52 @@ namespace JSON {
...
-        return;
+    fextl::vector<const json_t*> ConfigBlocks;
+    ConfigBlocks.push_back(ConfigList);
+
+    if (AppName) {
+      const json_t* OverrideList = json_getProperty(json, "AppOverrides");
+      if (OverrideList) {
+        for (const json_t* Item = json_getChild(OverrideList); Item != nullptr; Item = json_getSibling(Item)) {
+          const char* ItemName = json_getName(Item);
+          const json_t* OverrideNamedList = json_getProperty(OverrideList, ItemName);
```
neobrain: This is just recomputing `Item`, isn't it? Probably a leftover from a previous iteration of the PR.

Source/Common/Config.cpp
Sonicadvance1: I dunno, it was in the original PR which was already reviewed.

neobrain: Yes, I already flagged these comments in the original PR: https://github.com/FEX-Emu/FEX/pull/5120#discussion_r2675425349

Sonicadvance1: Removed the comment.

Source/Common/Config.cpp
Sonicadvance1: Replaced.

Source/Common/Config.cpp
Sonicadvance1: Fixed anyway, whatever.

Source/Common/Config.cpp
Sonicadvance1: Reordered headers.

unittests/APITests/WildcardMatcher.cpp
Sonicadvance1: Removed

FEXCore/Source/Utils/WildcardMatcher.cpp
Sonicadvance1: I dunno.

FEXCore/Source/Utils/WildcardMatcher.cpp
Sonicadvance1: Replaced their implementation with yours.

Source/Common/Config.cpp
Sonicadvance1: Removed const and added movers.

Source/Common/Config.cpp
Sonicadvance1: I don't know what you mean. ItemName and OverrideNamedList are both objects that are changing each iteration of the loop as it walks the AppOverrides.

FEXCore/Source/Utils/WildcardMatcher.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+public:
+  WildcardMatcher(std::string_view pattern, std::string_view text)
+    : pattern(pattern)
+    , text(text) {}
+
+  bool matchHelper(size_t p_idx, size_t t_idx);
+};
+
+bool WildcardMatcher::matchHelper(size_t p_idx, size_t t_idx) {
```
neobrain: The class is unnecessary now as pointed out in my previous comment:


```suggestion
bool WildcardMatcher::matchHelper(std::string_view pattern, std::string_view text, size_t p_idx, size_t t_idx) {
```

Source/Common/Config.cpp
neobrain: Maybe I'm misunderstanding. Looking at your example:
```json
    "AppOverrides" : {
        "setup*" : {
            "Comment": [
                "292030 - The Witcher 3: Wild Hunt"
            ],
            "X87ReducedPrecision": "0"
        }
     }
```

The way I'm breaking this down is:
* `OverrideList` refers to the `AppOverrides` node
* There's one loop iteration, in which...
  * `Item` is the `setup*` node
  * `json_getName(Item)` is `"setup*"`
  * `json_getProperty(OverrideList, "setup*");` is the `setup*` node again

So you've just done a roundtrip that recomputed the value you already had in `Item`, and you're performing redundant validity checks after.

neobrain:
> Based on https://github.com/FEX-Emu/FEX/pull/121 which needs to get merged first.

I'm guessing this is done now? Not sure what PR this was intended to reference since 121 probably wasn't it.

Sonicadvance1:
> > Based on https://github.com/FEX-Emu/FEX/pull/121 which needs to get merged first.
> 
> 
> 
> I'm guessing this is done now? Not sure what PR this was intended to reference since 121 probably wasn't it.

Oop, meant to be the PR number.

neobrain:
> Oop, meant to be the PR number.

Well, yes, but which one? Can you edit the description please?

Source/Common/Config.cpp
Sonicadvance1: Yes, you're misunderstanding.
- `OverrideList` refers to `AppOverride` node indeed.
- `Item = json_getChild(OverrideList)` Gets the `setup*` node as expected yes.
- `const char* ItemName = json_getName(Item);` is getting the string name of the `setup*` node, which is what is going to be used in the regex comparison
- `OverrideNamedList = json_getProperty(OverrideList, ItemName);` Gets the node contents, aka the overlaid config

eg:
```json
  "AppOverrides" : {
      "setup*" : {
          "Comment": [
              "292030 - The Witcher 3: Wild Hunt"
          ],
          "X87ReducedPrecision": "0"
      },
      "Wow.exe" : {
          "NeedsSeccomp": "1"
      },
      "pants" : {
          "X87ReducedPrecision": "1"
      }
   }
```

This will have three loop iterations.
* Item -> ItemName -> OverrideNamedList
** `setup*` NODE -> `setup*` STRING -> json properties
** `Wow.exe` NODE -> `Wow.exe` STRING -> json properties
** `pants` NODE -> `pants` STRING -> json properties.

FEXCore/Source/Utils/WildcardMatcher.cpp
Sonicadvance1: Removed the class. I'm down for some more recursive stack usage on initialization.

Sonicadvance1:
> Getting there. Could you also squash the commits by author? There's no need for this change set to be broken into 13 patches, 2 should suffice.

Squashed.

Source/Common/Config.cpp
Sonicadvance1: Oop nope. I'm wrong as usual. Removed.

FEXCore/Source/Utils/WildcardMatcher.cpp
neobrain: The code could rather easily be refactored to be non-recursive, but I'm happy enough that we got this file down from 280 lines of code to 23 :)

FEXCore/Source/Utils/WildcardMatcher.cpp
```diff
@@ -0,0 +1,23 @@
+// SPDX-License-Identifier: MIT
+
+#include <FEXCore/Utils/WildcardMatcher.h>
+
+namespace FEXCore::Utils::Wildcard {
+bool matchHelper(std::string_view pattern, std::string_view text, size_t p_idx, size_t t_idx) {
```
neobrain: Should be static

Source/Common/Config.cpp
```diff
@@ -48,21 +50,45 @@ namespace JSON {
...
+    if (AppName) {
+      const json_t* OverrideList = json_getProperty(json, "AppOverrides");
+      if (OverrideList) {
+        for (const json_t* Item = json_getChild(OverrideList); Item != nullptr; Item = json_getSibling(Item)) {
+          const char* ItemName = json_getName(Item);
+
+          if (!ItemName) {
+            LogMan::Msg::EFmt("JSON file '{}': Couldn't get config name for an item", Config);
+            break;
+          }
```
neobrain: I don't think this can ever happen, can it?

Source/Common/Config.cpp
```diff
@@ -48,21 +50,45 @@ namespace JSON {
...
-        LogMan::Msg::EFmt("JSON file '{}': Couldn't get config name for an item", Config);
-        return;
+    fextl::vector<const json_t*> ConfigBlocks;
+    ConfigBlocks.push_back(ConfigList);
+
+    if (AppName) {
+      const json_t* OverrideList = json_getProperty(json, "AppOverrides");
+      if (OverrideList) {
+        for (const json_t* Item = json_getChild(OverrideList); Item != nullptr; Item = json_getSibling(Item)) {
+          const char* ItemName = json_getName(Item);
```
neobrain:
```suggestion
          const char* AppPattern = json_getName(Item);
```

Source/Common/Config.cpp
```diff
@@ -48,21 +50,45 @@ namespace JSON {
...
-        LogMan::Msg::EFmt("JSON file '{}': Couldn't get value for config item '{}'", Config, ConfigName);
-        return;
-      }
+    for (auto ConfigBlock : ConfigBlocks) {
+      for (const json_t* ConfigItem = json_getChild(ConfigBlock); ConfigItem != nullptr; ConfigItem = json_getSibling(ConfigItem)) {
+        const char* ConfigName = json_getName(ConfigItem);
+        const char* ConfigString = json_getValue(ConfigItem);
 
-      Func(ConfigName, ConfigString);
+        if (!ConfigName) {
```
neobrain: Another instance where nullptr can't happen. Not sure about the getValue check though, presumably `{}` would return nullptr upon casting to `char*`, so that one is fine to stay.

Source/Common/Config.cpp
Sonicadvance1: Supposedly it can according to the json parser source, but I don't know what the edge case is. I've never seen it deleted.

Source/Common/Config.cpp
Sonicadvance1: The getValue can return nullptr in the case of it being an json object or array. So misconfigured stuff.

Sonicadvance1:
> Last round, good to merge afterwards. (Though note we're back to 6 commits now...)

That's okay, commits are free.

PR #5389, opened 2026-03-20
SMCTracking: Remove relocation log

Sonicadvance1:
Holy jeez does this thing spam.


Source/Tools/LinuxEmulation/LinuxSyscalls/SyscallsSMCTracking.cpp
```diff
@@ -223,7 +223,7 @@ static ReadELFHeadersResult ReadELFHeaders(int FD, std::span<std::byte> HeaderDa
   Parser.ReadElf(dup(FD));
 
   auto Relocations = Parser.PopulateRelocations();
-  if (!Relocations.empty()) {
+  if (EnableCodeCaching && !Relocations.empty()) {
```
neobrain: The log message was mainly for debugging - we can remove it entirely. If we find the information useful it should be re-added at the call site instead.

Sonicadvance1: Sure. Deleted.

PR #5395, opened 2026-03-25
CPUID: Add basic stub handling for AVX10 info

lioncash:
Just gets the feature bit handling stuff in place for various facilities, so it can be easily expanded in the future.

### DIFF
--- a/FEXCore/Source/Interface/Core/CPUID.cpp
+++ b/FEXCore/Source/Interface/Core/CPUID.cpp
@@ -756,6 +756,95 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_07h(uint32_t Leaf) const {
               (0 << 29) |                  // Arch capabilities - Speculative side channel mitigations
               (0 << 30) |                  // Arch capabilities - MSR module specific
               (0 << 31);                   // SSBD - Speculative Store Bypass Disable
+  } else if (Leaf == 1) {
+    Res.eax = (0U << 0) |  // SHA512
+              (0U << 1) |  // SM3
+              (0U << 2) |  // SM4
+              (0U << 3) |  // RAO_INT
+              (0U << 4) |  // AVX_VNNI
+              (0U << 5) |  // AVX512_BF16
+              (0U << 6) |  // LASS (Linear Address Space Separation)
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -834,6 +923,25 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_1Ah(uint32_t Leaf) const {
   return Res;
 }
 
+FEXCore::CPUID::FunctionResults CPUIDEmu::Function_24h(uint32_t Leaf) const {
```
Sonicadvance1, line 926: If we add this cpuid function range, we could want to bump function 0h to return 24h in eax no?

lioncash: ah, good catch

FEXCore/Source/Interface/Core/CPUID.cpp
```diff
@@ -834,6 +923,25 @@ FEXCore::CPUID::FunctionResults CPUIDEmu::Function_1Ah(uint32_t Leaf) const {
...
+  FEXCore::CPUID::FunctionResults Res {};
+
+  if (Leaf == 0) {
+    // EAX indicates the maximum number of subleaves.
+    Res.eax = 0;
+
+    // Bits 19-31 reserved
+    Res.ebx = (1U << 0) |     // (bits 0-7 specify the vector ISA version)
+              (0b111U << 16); // Defined as always 0b111
```
Sonicadvance1: Actually, I have a concern that even though this is officially documented as "reserved" and read as 0b111. It's original intent was to document the supported vector widths.
I feel like we should be safe and return full reserved zero values until AVX10 is supported (Or we have hardware that tells us otherwise). Probably also then knocking back 0h return size to the previous value since we're returning reserved values anyway.

Of course keeping that documentation with a note saying what it should return once supported? Sorry for yanking you around asking for these changes, I'm also reading up what the expected CPUID values are in the meantime.

lioncash: Updated

PR #5406, opened 2026-03-31
FEXRootFSFetcher: Improve hashing performance

Sonicadvance1:
Don't use pread, instead map the file and madvise larger blocks. This removes copying overhead as its just mapping file pages in instead. Also splits the implementation of file reading from hashing to make tinkering less involved, as if I want more performance out of this (say due to live hashing) then it's easier to tinker.

Improves hashing performance from ~2.2GB/s to ~3.6GB/s on my system, which is CPU bounded by xxhash and ping-ponging the kernel here.

### DIFF
--- a/Source/Tools/FEXRootFSFetcher/XXFileHash.cpp
+++ b/Source/Tools/FEXRootFSFetcher/XXFileHash.cpp
@@ -5,12 +5,74 @@
 #include <fcntl.h>
 #include <fmt/format.h>
 #include <unistd.h>
-#include <vector>
 #include <xxhash.h>
+#include <functional>
 
 namespace XXFileHash {
-// 32MB blocks
-constexpr static size_t BLOCK_SIZE = 32 * 1024 * 1024;
+class Reader {
+public:
+  Reader(int fd, size_t Size)
+    : fd {fd}
+    , Size {Size} {}
+
+  virtual ~Reader() = default;
+
+  bool Initialized() const {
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXRootFSFetcher/XXFileHash.cpp
```diff
@@ -5,12 +5,74 @@
...
+
+    while (ReadPtr < ReadEndPtr) {
+      ReadSize = std::min<size_t>(READ_BLOCK_SIZE, ReadEndPtr - ReadPtr);
+
+      if (!cb(ReadPtr, ReadSize)) {
+        return false;
+      }
+
+      // Only allow a single block read to be resident.
+      ::madvise(reinterpret_cast<void*>(ReadPtr), ReadSize, MADV_DONTNEED);
```
ShadowCurse, line 61: Is there a reason to call `madvise` on each just read chunk? Since you just read the file, it's content will stay in the page cache and kernel can evict it whenever it needs to anyway.

Sonicadvance1: For large files, the memory still becomes resident. I want to ensure we don't spin residency to upwards of 8GB, so it seems like a reasonable trade-off.

ShadowCurse: ok got it

PR #5418, opened 2026-04-14
ArchHelpers: Allow atomic memory operations in non-JIT handler

Sonicadvance1:
`Detroit: Become Human` decided to use unaligned CriticalSections. So this workarounds that.


cjacek (approved):
LGTM

lioncash (approved):

PR #5422, opened 2026-04-16
FEXCore: Add support for developer single stepping, read/write watching.

Sonicadvance1:
This is a fairly heavy handed approach to debugging but is super handy when dissecting a problem.  Basically toggle the constexpr `BLOCK_DEBUGGING` option, and you get watchpoints and single-stepping support. The watchpoints trigger /before/ the read/write because of how the implementation works.


Sonicadvance1:
This has been sitting here for a month, which is far longer than our policy of allowing uncontended PRs to merge in a week.
I'll be merging this week if this doesn't change.

lioncash (approved):

PR #5423, opened 2026-04-16
WOW64: Support disabling DEP

bylaws:
Required for older 32-bit games that assumes the execute bit is implicit from read. Needs a corresponding wine patch.


Source/Windows/Common/InvalidationTracker.cpp
```diff
@@ -60,6 +84,48 @@ void InvalidationTracker::HandleMemoryProtectionNotification(uint64_t Address, u
   }
 }
 
+void InvalidationTracker::HandleProcessExecuteFlagsChange(ULONG Flags) {
```
Sonicadvance1, line 87: Don't we need to invalidate code caches on DEP state change? I know of at least one game that execs read-only code, pings on the SIGILL(SIGSEGV?) and disables DEP and continues executing.

Also do we actually need to track DEP regions, or just consider all read-only regions as read-execute without tracking the regions like on the Linux side? Since this is a process-wide toggle right?

bylaws: Eww I see, I wonder if we shouldn't cache the code for non executable blocks? That would solve the problem in a way I think is probably cleaner? 

> Also do we actually need to track DEP regions, or just consider all read-only regions as read-execute without tracking the regions like on the Linux side? Since this is a process-wide toggle right?

We need to track the regions to be able to return the right extents when querying executable ranges, and if we merged them into XRegions without the separate DEP variant it woild be slightly more awkward to restore afterwards and we need to track separately for RW regions anyway to be able to restore the right permissions so seems cleanest to me.

Sonicadvance1: I guess if we never cache NX blocks that would be good. Just need to double check that multiblock doesn't bring them in as branch targets and whatever. Until we hit a game that constantly generates NX blocks and overflows our cache constantly.

bylaws: Right - will check this Monday

bylaws:
needs some testing but I think this should work

bylaws:
probably worth merging/testing soon

Sonicadvance1:
I'll take a look at the PR this week.

Sonicadvance1 (approved):

WINDROID-EMU:
Could you give me a hand with the game Call of Duty Black Ops 2? To this day, nobody has managed to get this game to run, not even on the Xbox One X or the Xbox 64. It's a huge challenge.

PR #5424, opened 2026-04-17
Windows: Introduce a FEX unixlib and use for all linux syscalls

bylaws:
This avoids the ever increasing need to expose obscure or fex-specific linux features in wine patches and instead allows FEX to run its own linux code nicely - this could pave way for faster sigbus handling or smc  (with far too much effort to the degree i think it's still a bad idea over hashing) 

This is somewhat of a breaking change in that wine fex users will also need to install the unixlibs for full functionality, but seems far better in long-run.

Needs testing with shm stats and madvise naming to confirm it still works though before merging.

Mostly authored by claude, which is much better now than jan I must say

### DIFF
--- a/Source/Windows/Common/Allocator.cpp
+++ b/Source/Windows/Common/Allocator.cpp
@@ -2,242 +2,51 @@
 #include <FEXCore/Utils/AllocatorHooks.h>
 #include <FEXCore/Utils/TypeDefines.h>
 
-#include <array>
-#include <chrono>
-#include <libloaderapi.h>
 #include <sysinfoapi.h>
-#include <synchapi.h>
 #include <windef.h>
 #include <winternl.h>
-#include <winnt.h>
-#include <wine/debug.h>
+#include <wine/unixlib.h>
 
-namespace FEX::Windows::Allocator {
-#define PR_SET_VMA 0x53564d41
-#define PR_SET_VMA_ANON_NAME 0
+#include "Unixlib/FEXUnixlib.h"
 
 #define MADV_HUGEPAGE 14
 #define MADV_NOHUGEPAGE 15
 
-namespace Trampoline {
-  struct madvise_data {
-    const void* addr;
-    size_t size;
-    int advise;
-  };
+namespace FEX::Windows::Allocator {
 
-  struct prctl_data {
-    int op;
-    uint64_t attr;
-    const void* addr;
-    size_t size;
-    const char* name;
-    uint64_t ret;
-  };
-
-  __attribute__((naked)) uint64_t wine_prctl(prctl_data* d) {
-    asm volatile(
-      R"(
-  .globl wine_prctl_begin
-  wine_prctl_begin:
-    mov x19, x0;
-    mov x8, 167; // prctl
-    ldr x0, [x19]; // op
-    ldp x1, x2, [x19, %[attr_offset]]; // {attr, addr}
-    ldp x3, x4, [x19, %[size_offset]]; // {size, name}
-    svc #0;
-    str x0, [x19, %[ret_offset]];
-    // Tell wine it was all groovy.
-    mov x0, 0;
-    ret;
-
-  .globl wine_prctl_end
-  wine_prctl_end:
-  )"
-      :
-      : [attr_offset] "i"(offsetof(prctl_data, attr)), [size_offset] "i"(offsetof(prctl_data, size)), [ret_offset] "i"(offsetof(prctl_data, ret))
-      : "memory");
-  };
-
-  __attribute__((naked)) uint64_t wine_madvise(madvise_data* d) {
-    asm volatile(R"(
-  .globl wine_madvise_begin
-  wine_madvise_begin:
-    mov x8, 233; // madvise
-    ldr x2, [x0, %[advise_offset]]; // advise
-    ldp x0, x1, [x0]; // {addr, size}
-    svc #0;
-    // Tell wine it was all groovy.
-    mov x0, 0;
-    ret;
-
-  .globl wine_madvise_end
-  wine_madvise_end:
-  )" ::[advise_offset] "i"(offsetof(madvise_data, advise))
-                 : "memory");
-  }
-
-  extern "C" uint64_t wine_madvise_begin;
-  extern "C" uint64_t wine_madvise_end;
-
-  void* const wine_madvise_begin_loc = &wine_madvise_begin;
-  void* const wine_madvise_end_loc = &wine_madvise_end;
-
-  extern "C" uint64_t wine_prctl_begin;
-  extern "C" uint64_t wine_prctl_end;
-
-  void* const wine_prctl_begin_loc = &wine_prctl_begin;
-  void* const wine_prctl_end_loc = &wine_prctl_end;
-
-  extern NTSTATUS(WINAPI* __wine_unix_call_dispatcher)(uint64_t, unsigned int, void*);
-  decltype(__wine_unix_call_dispatcher) WineUnixCall;
-
-  enum unix_function_indexes {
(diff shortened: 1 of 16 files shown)


Source/Windows/Unixlib/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
+cmake_minimum_required(VERSION 3.14)
+project(FEXUnixlib CXX)
+
+include(GNUInstallDirs)
+
+add_library(wow64fex_unixlib SHARED FEXUnixlib.cpp)
+set_target_properties(wow64fex_unixlib PROPERTIES OUTPUT_NAME "wow64fex" PREFIX "lib" SUFFIX ".so")
```
crueter, line 7:
```suggestion
set_target_properties(wow64fex_unixlib PROPERTIES OUTPUT_NAME "wow64fex")
```

These are already set by default on *NIX

Source/Windows/Unixlib/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+target_link_libraries(wow64fex_unixlib PRIVATE rt)
+install(TARGETS wow64fex_unixlib LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT Runtime)
+
+add_library(arm64ecfex_unixlib SHARED FEXUnixlib.cpp)
+set_target_properties(arm64ecfex_unixlib PROPERTIES OUTPUT_NAME "arm64ecfex" PREFIX "lib" SUFFIX ".so")
```
crueter, line 12:
```suggestion
set_target_properties(arm64ecfex_unixlib PROPERTIES OUTPUT_NAME "arm64ecfex")
```

Source/Windows/Unixlib/CMakeLists.txt
```diff
@@ -0,0 +1,14 @@
...
+target_link_libraries(arm64ecfex_unixlib PRIVATE rt)
+install(TARGETS arm64ecfex_unixlib LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT Runtime)
```
crueter, line 14: Actually, I think it would be better to name the output based on the current architecture, and use a singular target. For example:

```cmake
if (ARCHITECTURE_arm64ec)
    set(libname arm64ecfex)
elseif (ARCHITECTURE_arm64)
    set(libname wow64fex)
endif()

add_library(unixlib SHARED FEXUnixlib.cpp)
set_target_properties(unixlib PROPERTIES OUTPUT_NAME "${libname}")
target_link_libraries(unixlib PRIVATE rt)
install(TARGETS unixlib LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR} COMPONENT Runtime)
```

crueter (changes requested):

Source/Windows/Unixlib/CMakeLists.txt
bylaws: I don't think this quite works, this is a separate project we want to ompile once, we don't invoke CMake twice per arch here

crueter: https://github.com/FEX-Emu/FEX/blob/main/.github/workflows/wine_dll_artifacts.yml#L37-L45

`wine_build` is run twice, once for arm64ec and once for wow64. I realized I'm wrong in that arm64/arm64ec isn't ever set in this specific step (since it's built in Linux land, not MinGW)--however as it stands two identical copies of each .so are being built. So there are two solutions:

1. Add an option to the unixlib for arm64ec/wow64
2. Keep the CMake as it is, but move the Unixlib build step outside of wine_build and into wine_dll_artifacts

Frankly I'm not entirely certain why wow64/arm64ec need different shared libraries for this, since they are both built identically to one another--is it possible to merge them somehow?

bylaws: >wine_build is run twice, once for arm64ec and once for wow64.

Ah this is an artefact of how the CI was setup. It's only needed once, will see if I can fix that. There's no way to unify these on the wine side we always need separate file copies as such I thought maybe it could be useful to just build the file twice in case it happens to be useful somehow, the alternative is just to cp/a CMake target that results in such

crueter: In that case, yeah you would either copy the library, *or* create an object library that links into two separate shared libraries. But since they're literally identical, it's probably best to just `cp` them, either in CMake or in the workflow

bylaws: Did using OBJECTS? think this seems cleanest

bylaws:
updated

Source/Windows/Unixlib/CMakeLists.txt
crueter: 'Twould work. LGTM

Source/Windows/Common/Allocator.cpp
```diff
@@ -2,242 +2,51 @@
...
 
-  if (!Sym) {
+void SetupHooks(bool IsWine) {
+  if (!IsWine) {
     return;
   }
 
-  FEXCore::Allocator::HookPtrs Ptrs {};
-
-  // Wine will soon require us to use unixlib for calling helper routines that use Linux syscalls.
```
neobrain, line 188: Could you double-check that the context/background provided in the original comments isn't just dropped (to the extent that it still applies)? The new comments are very sparse, and it seems easy enough to avoid repeating the lengthy conversation we had to produce the original comment.

bylaws: I don't think any of it is really meaningful to have as a code comment tbh, the comment exists to justify a codepath that no longer exists

bylaws:
updated

Sonicadvance1:
Since this is primarily authored by AI tools, I'm going to need to spend extra effort reviewing this PR, which is why I've been putting it off.

Source/Windows/Common/Unixlib.cpp
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+#ifdef ARCHITECTURE_arm64ec
+  DispatcherDirect = *DispatcherPtr;
+  Dispatcher = TrampolineCall;
+#else
+  Dispatcher = *DispatcherPtr;
+#endif
+
+  if (NtQueryVirtualMemory(NtCurrentProcess(), &__ImageBase, MemoryWineUnixFuncs, &UnixlibHandle, sizeof(UnixlibHandle), nullptr)) {
```
cjacek, line 39: Current Wine has `MemoryWineLoadUnixLibByName`, which allows passing unixlib name. With that, it should be possible to share unixlib among wow64 and arm64ec builds. We may backport that to Proton if needed.

bylaws: Oh nice, that's much cleaner will switch :)

Source/Windows/Common/Allocator.cpp
neobrain: // Wine will soon require us to use unixlib for calling helper routines that use Linux syscalls.
  // It needs this for `userspace syscall dispatch` to capture rogue applications doing raw Windows syscalls.

That's the least bit that's still relevant context. Probably makes sense to reword it with the new implementation, but without this context the code begs the same questions we went through when adding this code initially (see discussion over there, trying to avoid repeating it here since adapting existing comments doesn't take that much time).

Source/Windows/Common/Unixlib.cpp
Sonicadvance1: Was this switched?

Sonicadvance1:
This is now implemented through a series of non-LLM related PRs, and I even understand it now.

PR #5419, opened 2026-04-15
JIT-inline F64SCALE and F64F2XM1 for reduced precision x87 path

pmatos:
Inlines reduced precision fscale and f2xm1, similarly to what we did with trig functions.

PR #5343, opened 2026-03-03
JIT-inline F64SIN, F64COS, and F64TAN for reduced precision x87 path

pmatos:

PR #5425, opened 2026-04-17
JIT-inline F64ATAN and F64FYL2X for reduced precision x87 path

pmatos:
In the same direction of #5419 and #5343.
Still trying to stick as close as possible to the suggestions in https://github.com/ARM-software/optimized-routines/
I am not an expert in how some of these algorithms work but they pass the tests and in-game testing, in addition to being faster than what we had before.


Sonicadvance1:
I'll run this through my usual suite of testing tomorrow to confirm nothing breaks.

Sonicadvance1 (approved):
Weee

PR #5427, opened 2026-04-19
FEXCore: Detect and document Snapdragon X2 Elite new alignment behaviour

Sonicadvance1:
Specifically Oryon-3 has added support for RMW instructions to
be unaligned within a 64B granule as long as they don't cross the
granule. The weird thing is that unaligned atomics for LRCPC  that
are still within the 64B granule aren't affected, so it doesn't improve
performance in the "regular" load/store case, only the `LOCK`
instruction cases.

There's no way to detect this other than detecting the CPU.

### DIFF
--- a/FEXCore/include/FEXCore/Core/HostFeatures.h
+++ b/FEXCore/include/FEXCore/Core/HostFeatures.h
@@ -42,6 +42,7 @@ struct HostFeatures {
   bool Supports3DNow {};
   bool SupportsSSE4a {};
   bool SupportsMOPS {};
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
Okay, actually we don't need to detect this because we don't need to change anything on the JIT side. It'll just occur and our SIGBUS handler will just get used less frequently as cross-16B granule RMW operations will not fault.

Atomic loads and stores will still fault crossing a 16B granule, which will backpatch to the regular barrier+non-atomic, which is still a huge performance bottleneck.

PR #5428, opened 2026-04-20
Steam/CompatTool: Support `STEAM_COMPAT_GRAPHICS_PROVIDER` for rootfs path

Sonicadvance1:
If we have been provided a graphics provider path through an environment variable, then use that path directly rather than searching.

### DIFF
--- a/Source/Steam/CompatTool.cpp
+++ b/Source/Steam/CompatTool.cpp
@@ -26,24 +26,25 @@ fextl::string GenerateSteamConfigTemplate(const FEX::Config::PortableInformation
     return {};
   }
 
-  // Try and find a mount point.
   fextl::string MountPoint {};
+
+  // If the graphics provider was provided through an environment variable, then use that.
+  // Otherwise, try to find a mount point.
+  const char* GraphicsProvider = getenv("STEAM_COMPAT_GRAPHICS_PROVIDER");
   const char* RuntimeDir = getenv("XDG_RUNTIME_DIR");
-  if (RuntimeDir) {
+  const char* CacheDir = getenv("XDG_CACHE_HOME");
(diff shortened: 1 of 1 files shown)


Source/Steam/CompatTool.cpp
```diff
@@ -26,22 +26,29 @@ fextl::string GenerateSteamConfigTemplate(const FEX::Config::PortableInformation
...
+  if (GraphicsProvider) {
+    // If the graphics provider was provided through an environment variable, then use that.
+    MountPoint = GraphicsProvider;
   } else {
-    const auto UserDirectory = fextl::fmt::format("/run/user/{}", geteuid());
-    if (FHU::Filesystem::Exists(UserDirectory)) {
-      MountPoint = fextl::fmt::format("{}/fexrootfs/", UserDirectory);
+    // Try and find a mount point.
+    const char* RuntimeDir = getenv("XDG_RUNTIME_DIR");
+    if (RuntimeDir) {
```
neobrain, line 38:
```suggestion
    // If the graphics provider was provided through an environment variable, then use that.
    // Otherwise, try to find a mount point.
  const char* GraphicsProvider = getenv("STEAM_COMPAT_GRAPHICS_PROVIDER");
  const char* RuntimeDir = getenv("XDG_RUNTIME_DIR");
  const auto UserDirectory = fextl::fmt::format("/run/user/{}", geteuid());
  if (GraphicsProvider) {
    MountPoint = GraphicsProvider;
  } else if (RuntimeDir) {
    ...
  } else if (FHU::Filesystem::Exists(UserDirectory)) {
    ...
  } else {
    ...
  }
```

Sonicadvance1: Sure.

PR #5438, opened 2026-04-27
FEXCore: Fix stack access for 0x66 PUSH/POP FS/GS

### LINKED ISSUE #5437: Incorrect stack adjustment for POP/PUSH FS/GS when prefixed with 0x66
The translation for the instructions `pop FS`, `pop GS`, `push FS` and `push GS` when preceded by the 0x66 operand-size override prefix is incorrect in **64** bit mode.

**Expected Behavior:**
The 0x66 prefix should force a 16-bit operation: The CPU should push/pop exactly 2 bytes of data to/from the stack and adjust RSP by exactly 2.

**Actual Behavior:**
The 0x66 prefix is **ignored**: FEX pushes/pops 8 bytes of data and adjusts RSP by 8.

simon902:
Previously, PUSH/POP FS/GS were incorrectly flagged, causing the 0x66 operand-size override prefix to be ignored and defaulting to an 8-byte stack operation.

This change updates the instructions to use `SIZE_64BITDEF` instead of `SIZE_64BIT` ensuring that the if statement in `Frontend.cpp:474` is taken and the `SIZE_16BIT` to be correctly applied:

```cpp
    } else if (HasNarrowingDisplacement &&
               (SrcSizeFlag == FEXCore::X86Tables::InstFlags::SIZE_DEF || SrcSizeFlag == FEXCore::X86Tables::InstFlags::SIZE_64BITDEF)) {
      // See table 1-2. Operand-Size Overrides for this decoding
      // If the default operating mode is 32bit and we have the operand size flag then the operating size drops to 16bit
      DecodeInst->Flags |= DecodeFlags::GenSizeSrcSize(DecodeFlags::SIZE_16BIT);
}
```

Fixes #5437

### DIFF
--- a/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
+++ b/FEXCore/Source/Interface/Core/X86Tables/SecondaryTables.cpp
@@ -31,19 +31,19 @@ constexpr std::array<X86InstInfo[2], ENTRY_MAX> Secondary_ArchSelect_LUT = {{
   },
   {
     {"PUSH FS", TYPE_INST, GenFlagsSrcSize(SIZE_16BIT) | FLAGS_DEBUG_MEM_ACCESS | FLAGS_NO_OVERLAY, 0, { .OpDispatch = &IR::OpDispatchBuilder::Bind<&IR::OpDispatchBuilder::PUSHSegmentOp, FEXCore::X86Tables::DecodeFlags::FLAG_FS_PREFIX> } },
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (commented):
That's a good catch! Thanks for the fix.
I opened #5439 which also adds 64-bit side unittests to ensure we don't break it.

lioncash:
Closing, since this was cherry-picked into #5439 (again, thanks for the fix!)

PR #5441, opened 2026-04-28
CodeCache: Implement lazy code loading

neobrain:
## Overview

Currently, FEX eagerly loads the entire contents of a code cache when the guest app loads the corresponding library. But typically, not all cache contents are needed in every single run (and especially not immediately after loading), so this not only increases memory use but also adds undesirable disk IO compared to an uncached run.

This PR implements lazy loading of cached code pages to alleviate these effects: Pages are now loaded from the cache file into memory on-demand as they're being executed (or otherwise accessed) by setting up a segfault handler on yet-untouched code pages. This new behavior is off by default and controlled by a new setting called `EnableLazyCodeCachingWIP`.

## Implementation details

Notably, we can't rely on the OS's standard mechanism for memory-mapped files here since FEX needs to apply its JIT relocations on loaded code sections.

### 3-step cache setup

Cache files are now loaded in 3 steps:
* LoadCache: Read cache file headers, perform basic validation (version checks, ...), 
* EnableLoadedSection: Sets up LookupCache entries to point into the protected cache file (and performs extended validation if enabled)
* FinalizeCodePages: Performs the actual lazy loading of code pages (called on first access by the segfault handler)

When lazy cache loading is *disabled*, the 3 steps are still used, but EnableLoadedSection will just call FinalizeCodePages directly with the entire memory range.

### Page-loading granularity

Lazy loading ("page finalization") will generally be as granular as individual pages *if possible*. There are two deviations from the rule of thumb:
* Page-crossing JIT relocations would be cut off, so we always pull in the adjacent page from disk in that case
* Similarly, page-crossing block linking records (40 bytes) trigger the same behavior, but I couldn't figure out why. I assumed they would just trigger a segmentation fault from within ExitFunctionLinker that could be handled, but I couldn't get that approach working in practice (maybe I missed some implicit requirement this code has?). For now, ensuring we never cut off block linking records isn't a big deal, so we can worry about this at a later point.

The logic for this is implemented in SelectCodeRangeToFinalize.

### Thread-safety

A new `CodeCachePatchingMutex` is added (though perhaps the existing CodeInvalidationMutex would be a better fit?) to guard page finalization. During finalization, pages are unprotected in three steps:
* `PROT_NONE`->`write-only`
* apply JIT relocations
* `write-only`->`RWX`

This avoids potential race conditions where other code might read constant pools while we're still in the process of applying relocations. Such code instead now triggers another segfault, which will wait for the pending page finalization to finish and then continue.

### Fork safety

Code page finalization requires two mutexes to be held (VMATracking and CodeCachePatchingMutex). A naive extension of Core::LockAfterFork will run into deadlocks, so a simple deadlock-avoidance algorithm is required where partial locks are reverted and locking is retried until success. I'm not sure if there's a better way without significantly refactoring the interfaces here.

## Future work

* The initial LoadCache call will now do some extra work and bookkeeping (like splitting FEX relocations into per-page groups) that can be avoided by changing the code cache format. This is largely straightforward and won't require interfaces changes.
* Tune page selection algorithm: Currently the smallest fragment possible is used, but we may want to extend e.g. to the full multiblock. However it's not obvious what the ideal choice here is, and and page selection must be moved to FEXOfflineCompiler for this to be feasible first.

## Miscellaneous notes

* `LookupCache::ClearCache` clears all references to cached code without unloading the cache file. Is there a good way to preserve those references on clears? (not critical for merging though)
* Are there potential race conditions where other threads might be writing to a region of cached code while it's being finalized? (mutexes should take care of it)
* Does page-crossing TSO backpatching face similar problems as page-crossing JIT relocations? (probably, but it's extremely rare in practice. Best fixed later, once page clustering is moved to FEXOfflineCompiler)

### DIFF
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -1,4 +1,9 @@
 // SPDX-License-Identifier: MIT
+#include "FEXCore/Utils/LogManager.h"
+#include "FEXCore/Utils/MathUtils.h"
+#include "FEXCore/Utils/TypeDefines.h"
+#include "FEXCore/fextl/memory.h"
+#include <FEXCore/Utils/Profiler.h>
 #include <FEXCore/Utils/SpinWaitLock.h>
 
 #include <Interface/Context/Context.h>
@@ -16,9 +21,13 @@
 
 #include <FEXHeaderUtils/Filesystem.h>
 
+#include <algorithm>
 #include <git_version.h>
 
+#include <span>
 #include <xxhash.h>
+
+#include <FEXCore/Utils/AllocatorHooks.h>
 
 #include <fstream>
 
@@ -31,6 +40,32 @@ ExecutableFileInfo::ExecutableFileInfo(fextl::unique_ptr<HLE::SourcecodeMap> Map
   , Filename(Filename) {}
 #endif
 ExecutableFileInfo::~ExecutableFileInfo() = default;
+
+MappedCodeCacheFile::~MappedCodeCacheFile() {
+  if (CacheManager) {
+    CacheManager->UnregisterMappedCodeBuffer(*this);
+  }
+}
+
+void AbstractCodeCache::RegisterMappedCodeBuffer(MappedCodeCacheFile& Code) {
+  MappedCodeBuffers.push_back(Code.CodeBuffer);
+  // Unregister on destruction of Code
+  Code.CacheManager = this;
+}
+
+void AbstractCodeCache::UnregisterMappedCodeBuffer(MappedCodeCacheFile& Code) {
+  std::erase_if(MappedCodeBuffers, [&](const auto& Elem) { return Elem.data() == Code.CodeBuffer.data(); });
+}
+
+bool AbstractCodeCache::IsAddressInMappedCodeBuffer(uintptr_t Address) const {
+  for (const auto& Range : MappedCodeBuffers) {
+    auto Start = reinterpret_cast<uintptr_t>(Range.data());
+    if (Address >= Start && Address < Start + Range.size_bytes()) {
+      return true;
+    }
+  }
+  return false;
+}
 
 fextl::string CodeMap::GetBaseFilename(const ExecutableFileInfo& MainExecutable, bool AddNombSuffix) {
   auto FileId = MainExecutable.FileId;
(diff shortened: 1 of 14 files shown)


bylaws:
ooh nice, do you have benchmarks on startup times of games with vs without this? would be cool to see

neobrain:
> ooh nice, do you have benchmarks on startup times of games with vs without this? would be cool to see

No fancy benchmarks besides sanity testing. The main goal is to get the interface updates out of the way, as well as reducing the risk of FEX discarding all caches just because we resized a CodeBuffer (which would actually add a penalty compared to uncached operation).

neobrain:
The previous approach still had a tiny window where read-accesses would return unfinalized code data, so I had to rework the approach quite a bit. I've added this as a separate commit for the sake of documentation.

With the new approach, code finalization happens in an isolated virtual memory range that isn't known externally. Only once fully complete, that memory range is atomically remapped onto the memory range reserved for execution. The logic itself is more complex now, but atomicity makes verifying correctness much easier.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -792,29 +817,67 @@ void CodeCache::FinalizeCodePages(MappedCodeCacheFile& Code, std::span<std::byte
...
 
-  ARMEmitter::Emitter::ClearICache(CodeRange.data(), CodeRange.size_bytes());
-  if (!Allocator::VirtualProtect(CodeRange.data(), CodeRange.size_bytes(),
-                                 Allocator::ProtectOptions::Read | Allocator::ProtectOptions::Write | Allocator::ProtectOptions::Exec)) {
-    ERROR_AND_DIE_FMT("VirtualProtect failed");
+  // Atomically make the finalized code data visible by remapping the staging
+  // buffer onto the requested CodeBuffer window. MREMAP_DONTUNMAP is used to
+  // leave the old VA range reserved so that we can cleanly deallocate it
+  // through Allocator.
+  void* RemapResult = ::mremap(Staging, Size, Size, MREMAP_FIXED | MREMAP_MAYMOVE | MREMAP_DONTUNMAP, CodeRange.data());
```
Sonicadvance1, line 863: This line of code actually spooked me out until I read the comment.

FEXCore/Source/Interface/Core/CodeCache.cpp
```diff
@@ -32,6 +41,38 @@ ExecutableFileInfo::ExecutableFileInfo(fextl::unique_ptr<HLE::SourcecodeMap> Map
 #endif
 ExecutableFileInfo::~ExecutableFileInfo() = default;
 
+MappedCodeCacheFile::~MappedCodeCacheFile() {
+  if (CacheManager) {
+    CacheManager->UnregisterMappedCodeBuffer(*this);
+  }
+
+#ifndef _WIN32
+  if (CodeBuffer.empty()) {
```
bylaws, line 50: This seems wrong?

Sonicadvance1:
This was delayed due to 2605 release, but I'll get back to reviewing this soon.

FEXCore/Source/Interface/Core/CodeCache.cpp
neobrain: Oops, must've slipped in when I added WoA build support. Should be `!CodeBuffer.empty()` of course (which should always be true anyway though).

Sonicadvance1 (approved):
Here we go.

PR #5444, opened 2026-04-29
VectorOps: Fix MAXPS/MAXPD NaN and signed-zero tie selection

bylaws:
fcmgt returns false on NaN, so the existing polarity in the non-SVE
fcmgt+bit sequences and in the SVE predicate-merge picked the wrong
source on NaN/tie. Swap the compare operands and flip bit<->bif / add
a predicate not to match x86 second source wins behaviour.


Sonicadvance1 (approved):
Solve all my problems Claude.

PR #5445, opened 2026-04-29
OpcodeDispatcher: Fix 8/16-bit RCR/RCL effective count

bylaws:
The effective count for 8-bit RCR/RCL is (CL & 0x1F) MOD 9 and for
16-bit is (CL & 0x1F) MOD 17. We were masking by 0x1F only, so CL=9
(r8) / CL=17 (r16) ran the cascade body instead of being the required
no-op, clobbering CF/OF. Fold the MOD into the count using a new
ArithRef::URem helper so constant-folded counts still fold at emit
time.


bylaws:
Seems this is some edge case Intel/AMD divergence, sadge

PR #5446, opened 2026-04-29
OpcodeDispatcher: Fix FIST 16-bit IE for denormal inputs 

bylaws:
The special-value check masked the biased exponent with 0x7fff and
then used TestNZ, so it triggered on Exp==0 (denormals) instead of
Exp==0x7fff (NaN/Inf). Replace the TestNZ with SubWithFlags against
0x7fff so EQ only fires for genuine NaN/Inf.


Sonicadvance1 (approved):
AI is paving the way to a productive future.

PR #5447, opened 2026-04-29
SoftFloat: Fix FSCALE(0, +Inf) to raise IE and return a quiet NaN

bylaws:
The lhs==0 short-circuit in X80SoftFloat::FSCALE returned lhs
unchanged without calling extF80_mul, so the 0*Inf invalid-operation
case wasn't fully handled. Detect +Inf rhs explicitly
in the zero-lhs path and raise the flag, returning QNaN.


Sonicadvance1 (approved):
LLMs are great at logical deductions.

PR #5451, opened 2026-04-30
arm64ec: Fixes crash in many games with SDL+Dualsense

Sonicadvance1:
We were pointing to an incorrect function pointer and exploding when a pending suspend doorbell had occured.


bylaws (approved):
Nice catch

PR #5452, opened 2026-05-01
Wine: Add support for enabling Mono heuristic earlier

Sonicadvance1:
The Mono TSO heuristic by default doesn't really do anything when TSO is
enabled. So add a detection during wow64/arm64ec initialization that
runs through a handful of checks to see if the Mono TSO heuristic should
be enabled.

This ensures that we don't need to do any code cache clearing or
anything partway through the game loading mono, while still allowing
overrides if it has explicitly been set.

Specifically:
- If `SteamAppId` isn't set then skip heuristic entirely.
- Treat `STEAM_FEX_TSOENABLED` as a tristate.
  - If it exists, skip the heuristic detection even if it is 0.
- Treat `STEAM_COMPAT_FEX_CONFIG` as a tristate.
  - Same as above

It explicitly checks with the current executing path with the common
path to the mono runtime. Might need to add more in the future.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -34,6 +34,7 @@ $end_info$
 #include "Common/Exception.h"
 #include "Common/ImageTracker.h"
 #include "Common/InvalidationTracker.h"
+#include "Common/MonoHeuristic.h"
 #include "Common/OvercommitTracker.h"
 #include "Common/TSOHandlerConfig.h"
 #include "Common/CPUFeatures.h"
@@ -583,6 +584,9 @@ NTSTATUS ProcessInit() {
   FEX::Config::LoadConfig(fextl::string {ExecutableName}, _environ, FEX::ReadPortabilityInformation());
   FEXCore::Config::ReloadMetaLayer();
   FEX::Windows::Logging::Init();
+
+  // Handle Mono TSO heuristic early.
+  FEX::Windows::HandleMonoTSOHeuristicConfig(FEX::Windows::BaseName(FEX::Windows::GetExecutableFilePathW()));
 
   FEXCore::Config::Set(FEXCore::Config::CONFIG_IS64BIT_MODE, "1");
 
--- a/Source/Windows/Common/CMakeLists.txt
+++ b/Source/Windows/Common/CMakeLists.txt
@@ -9,6 +9,7 @@ target_include_directories(CommonWindowsRuntime PRIVATE "${CMAKE_SOURCE_DIR}/Sou
 add_library(CommonWindows STATIC
   Allocator.cpp
   CPUFeatures.cpp
+  MonoHeuristic.cpp
   SHMStats.cpp
   InvalidationTracker.cpp
   ImageTracker.cpp
--- a/Source/Windows/Common/Module.h
+++ b/Source/Windows/Common/Module.h
@@ -22,6 +22,17 @@ inline fextl::string GetExecutableFilePath() {
   return Path;
 }
 
+inline std::wstring GetExecutableFilePathW() {
+  std::array<WCHAR, PATH_MAX> Buf;
+  UNICODE_STRING PathW {.Length = 0, .MaximumLength = Buf.size() * sizeof(WCHAR), .Buffer = Buf.data()};
+
+  if (LdrGetDllFullName(nullptr, &PathW)) {
+    return {};
+  }
+
+  return std::wstring(PathW.Buffer);
+}
+
 inline fextl::string GetSectionFilePath(uint64_t Address) {
   struct {
     MEMORY_SECTION_NAME Info;
@@ -43,4 +54,8 @@ inline fextl::string GetSectionFilePath(uint64_t Address) {
 inline std::string_view BaseName(std::string_view Path) {
   return Path.substr(Path.find_last_of('\\') + 1);
 }
+inline std::wstring_view BaseName(std::wstring_view Path) {
+  return Path.substr(Path.find_last_of(L'\\') + 1);
+}
+
 } // namespace FEX::Windows
--- a/Source/Windows/Common/MonoHeuristic.cpp
+++ b/Source/Windows/Common/MonoHeuristic.cpp
@@ -1,0 +1,100 @@
+#include <FEXCore/Config/Config.h>
+
+#include <cstddef>
+#include <ntstatus.h>
+#include <shlwapi.h>
+#include <windef.h>
+#include <winternl.h>
+
+namespace FEX::Windows {
+std::wstring GetCurrentPath() {
+  std::wstring PathStr(4096, L'\0');
+  DWORD Result {};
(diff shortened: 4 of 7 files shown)


bylaws:
I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct

bylaws:
There are several older versions of mono this would cause crashes on as the hack doesn't work there

Sonicadvance1:
> I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct

If it detects Mono in the suspected path, then disable TSO so the mono heuristic does what it needs to? Maybe I just need to add an additional UnityEngine.dll check on top of this to guarantee constraints harder.

> There are several older versions of mono this would cause crashes on as the hack doesn't work there

How many of those are shipping in Unity?

bylaws:
> > I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct
> 
> If it detects Mono in the suspected path, then disable TSO so the mono heuristic does what it needs to? Maybe I just need to add an additional UnityEngine.dll check on top of this to guarantee constraints harder.

This is also not sufficient, a fair proportion of games have that and a non supported mono version.

> 
> > There are several older versions of mono this would cause crashes on as the hack doesn't work there
> 
> How many of those are shipping in Unity?

Several, a lot of 32 bit games and some 64 bit older unity ones. Mini metro I know off top of my head. The only way to do this correctly afaik is to just invalidate all code for the unity player image and main image,  I don't believe that would really impact that much as mono is very lean to compile.

Sonicadvance1:
> > > I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct
> > 
> > 
> > If it detects Mono in the suspected path, then disable TSO so the mono heuristic does what it needs to? Maybe I just need to add an additional UnityEngine.dll check on top of this to guarantee constraints harder.
> 
> This is also not sufficient, a fair proportion of games have that and a non supported mono version.

So a fair proportion of games today would pick up the mono heuristic and still crash today without this PR is what you're saying?

> 
> > > There are several older versions of mono this would cause crashes on as the hack doesn't work there
> > 
> > 
> > How many of those are shipping in Unity?
> 
> Several, a lot of 32 bit games and some 64 bit older unity ones. Mini metro I know off top of my head. The only way to do this correctly afaik is to just invalidate all code, I don't believe that would really impact that much as mono is very lean to compile.

You're saying that the heuristic would already pick up Mini Metro as an incompatible version and crash with today's code anyway? So I just need to make sure to enable the additional mono FEXCore toggle that the Image tracker does right from the start instead? And then we tag Mini Metro as known not being able to use this?

bylaws:
> > > > I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct
> > > 
> > > 
> > > If it detects Mono in the suspected path, then disable TSO so the mono heuristic does what it needs to? Maybe I just need to add an additional UnityEngine.dll check on top of this to guarantee constraints harder.
> > 
> > 
> > This is also not sufficient, a fair proportion of games have that and a non supported mono version.
> 
> So a fair proportion of games today would pick up the mono heuristic and still crash today without this PR is what you're saying?

Not with TSO enabled, no.

> 
> > > > There are several older versions of mono this would cause crashes on as the hack doesn't work there
> > > 
> > > 
> > > How many of those are shipping in Unity?
> > 
> > 
> > Several, a lot of 32 bit games and some 64 bit older unity ones. Mini metro I know off top of my head. The only way to do this correctly afaik is to just invalidate all code, I don't believe that would really impact that much as mono is very lean to compile.
> 
> You're saying that the heuristic would already pick up Mini Metro as an incompatible version and crash with today's code anyway? So I just need to make sure to enable the additional mono FEXCore toggle that the Image tracker does right from the start instead? And then we tag Mini Metro as known not being able to use this?

If tso is enabled it would not crash, no. And this is not a mini metro specific crash, of the 4 games I checked when I developed this mini metro was 1/4. Manually invalidating code on backpatcher detection really should be fine and cover all cases. I would benchmark and test that before breaking a subset of unity games by default.

Sonicadvance1:
> > > > > I'm confused, this shouldn't work? What is this trying to achieve. It is unsafe to enable any heuristics till we actually hit the address of the backpatcher static detection is not correct
> > > > 
> > > > 
> > > > If it detects Mono in the suspected path, then disable TSO so the mono heuristic does what it needs to? Maybe I just need to add an additional UnityEngine.dll check on top of this to guarantee constraints harder.
> > > 
> > > 
> > > This is also not sufficient, a fair proportion of games have that and a non supported mono version.
> > 
> > 
> > So a fair proportion of games today would pick up the mono heuristic and still crash today without this PR is what you're saying?
> 
> Not with TSO enabled, no.

Sure, the idea is to be more aggressive about disabling TSO emulation with Unity since we have the heuristic.

> 
> > > > > There are several older versions of mono this would cause crashes on as the hack doesn't work there
> > > > 
> > > > 
> > > > How many of those are shipping in Unity?
> > > 
> > > 
> > > Several, a lot of 32 bit games and some 64 bit older unity ones. Mini metro I know off top of my head. The only way to do this correctly afaik is to just invalidate all code, I don't believe that would really impact that much as mono is very lean to compile.
> > 
> > 
> > You're saying that the heuristic would already pick up Mini Metro as an incompatible version and crash with today's code anyway? So I just need to make sure to enable the additional mono FEXCore toggle that the Image tracker does right from the start instead? And then we tag Mini Metro as known not being able to use this?
> 
> If tso is enabled it would not crash, no. And this is not a mini metro specific crash, of the 4 games I checked when I developed this mini metro was 1/4. Manually invalidating code on backpatcher detection really should be fine and cover all cases. I would benchmark and test that before breaking a subset of unity games by default.

bylaws:
Yes, I agree being more aggressive makes sense. If tso enablement was delayed till after the backpatcher was hit and only unityplayer.dll was invalidated (but the entirety of it) I think that would probably be the best solution.
Unityplayer.dll is where 90% of the non TSO perf delta originates, everything else is pretty much negligible iirc. You might even be able to just add it as a AddForceTSOInformation range and then keep the rest as TSO enabled?

Sonicadvance1:
Dropping since we have the turnip workaround now. Other drivers will still be in a bad spot obviously.

bylaws:
> Dropping since we have the turnip workaround now. Other drivers will still be in a bad spot obviously.

Don't most unity games use dxvk regardless? Should be fine all round

Sonicadvance1:
Not all games are DXVK.

bylaws:
> Not all games are DXVK.

Unity games this patch would detect are though?

PR #5458, opened 2026-05-03
make 64bit symbols visible to enhance portability (musl)

### LINKED ISSUE #5457: musl build fails: large-file types missing (incomplete type)
Building using musl fails with:

```
FEX/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Types.h:548:5: error: member access into incomplete type 'const struct statfs64'
  548 |     COPY(f_bsize);
      |     ^
```

peppergrayxyz:
building using clang/musl fails with:

```
FEX/Source/Tools/LinuxEmulation/LinuxSyscalls/x32/Types.h:548:5: error: member access into incomplete type 'const struct statfs64'
  548 |     COPY(f_bsize);
      |     ^
```

add `_LARGEFILE64_SOURCE` to define large-file feature macros

fix: #5457


CMakeLists.txt
```diff
@@ -194,6 +194,8 @@ if (ENABLE_GDB_SYMBOLS)
   add_compile_definitions(GDB_SYMBOLS_ENABLED=1)
 endif()
 
+add_compile_definitions(-D_LARGEFILE64_SOURCE)
```
Sonicadvance1: https://cmake.org/cmake/help/latest/command/add_compile_definitions.html
> Added in version 3.26: Any leading -D on an item will be removed.

Our minimum prescribed cmake version is 3.14.

Sonicadvance1 (approved):
Reasonable enough.

PR #5475, opened 2026-05-03
switch to portable move_if_noexcept wrapper

### LINKED ISSUE #5474: [libc++] error: std::function does not support implementing fextl::move_only_function
building using libc++ fails with:

```
FEX/FEXCore/include/FEXCore/fextl/functional.h:55:21: error: static assertion failed due to requirement 'noexcept(this->internal = std::move(wrapped_lambda))': This implementation of std::function does not support implementing fextl::move_only_function
   55 |       static_assert(noexcept(internal = std::move(wrapped_lambda)), "This implementation of std::function "
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
FEX/Source/Common/AsyncNet.h:253:7: note: in instantiation of function template specialization 'fextl::move_only_function<fasio::post_callback (fasio::error)>::move_only_function<(lambda at /home/pepper/dev/ghrepos/FEX/Source/Common/AsyncNet.h:253:7)>' requested here
  253 |       [ServerFD = FD, &Reactor = Reactor, OnAccept = std::move(OnAccept)](error ec) mutable {
      |       ^
1 error generated.
```

peppergrayxyz:
building using libc++ fails with:
```
FEX/FEXCore/include/FEXCore/fextl/functional.h:55:21: error: static assertion failed due to requirement 'noexcept(this->internal = std::move(wrapped_lambda))': This implementation of std::function does not support implementing fextl::move_only_function
   55 |       static_assert(noexcept(internal = std::move(wrapped_lambda)), "This implementation of std::function "
      |                     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
FEX/Source/Common/AsyncNet.h:253:7: note: in instantiation of function template specialization 'fextl::move_only_function<fasio::post_callback (fasio::error)>::move_only_function<(lambda at /home/pepper/dev/ghrepos/FEX/Source/Common/AsyncNet.h:253:7)>' requested here
  253 |       [ServerFD = FD, &Reactor = Reactor, OnAccept = std::move(OnAccept)](error ec) mutable {
      |       ^
1 error generated.
```

This works with libstdc++ because its std::function converting constructor and converting assignment are conditionally noexcept based on whether the target can be initialized without throwing.

libc++ does not expose the same guarantee on the converting assignment operator. In current libc++, the `std::function` overload is declared without `_NOEXCEPT`.

The fix uses a function pointer in the case that the guaranty is not provided directly from the standard library instead of working around `std::function` behavior.

For the straight forward cases the reference is stored in `internal` for the other case in `internal_invoker`. `internal_invoker` is initialized with a reference to a static function that calls `internal`. In case `internal` is used this ensures that it is invoked accordingly, in case `internal_invoker` is used, writing to it overrides the default behavior. The choice which one to invoke is therefore branchless.

fix: #5474

### DIFF
--- a/FEXCore/include/FEXCore/fextl/functional.h
+++ b/FEXCore/include/FEXCore/fextl/functional.h
@@ -32,30 +32,20 @@ public:
       internal = std::move(f);
     } else {
       // Other arguments require allocation, which is a problem since
-      // std::function doesn't allow allocator customization. Implementations
-      // are generally able to avoid allocation for lambdas with a single
-      // pointer capture however. We can exploit this special case by wrapping
-      // the actual argument in a lambda that points an external storage
-      // location.
+      // std::function doesn't allow allocator customization.
 
       static_assert(!std::is_pointer_v<F>, "Pointer types must manually be dereferenced");
 
-      // First, relocate argument to a location returned from FEX's allocators
+      // Relocate argument to a location returned from FEX's allocators
       using Fnoref = std::remove_reference_t<F>;
       storage = Alloc(alignof(Fnoref), sizeof(Fnoref));
-      auto moved_lambda = new (storage) Fnoref {std::move(f)};
+      new (storage) Fnoref {std::move(f)};
 
-      // Second, wrap the relocated argument in a single-capture lambda
-      auto wrapped_lambda = [moved_lambda](Args... args) {
+      // Create a wrapper and store a pointer to it
+      internal_invoker = +[](const move_only_function* self, Args... args) -> R {
+        auto* moved_lambda = reinterpret_cast<Fnoref*>(self->storage);
         return (*moved_lambda)(std::forward<Args>(args)...);
       };
-
-      // Third, assign the result to std::function, ensuring it's indeed
-      // allocation-free by checking for nothrow-constructibility
-      static_assert(noexcept(internal = std::move(wrapped_lambda)), "This implementation of std::function "
-                                                                    "does not support implementing "
-                                                                    "fextl::move_only_function");
-      internal = std::move(wrapped_lambda);
 
       // Finally, if a destructor must be called, generate a pointer to its destructor
       if constexpr (!std::is_trivially_destructible_v<Fnoref>) {
@@ -78,6 +68,7 @@ public:
       this->~move_only_function();
     }
     internal = std::exchange(other.internal, nullptr);
+    internal_invoker = std::exchange(other.internal_invoker, nullptr);
     internal_destructor = std::exchange(other.internal_destructor, nullptr);
     storage = std::exchange(other.storage, nullptr);
     return *this;
@@ -91,16 +82,21 @@ public:
   }
 
   R operator()(Args... args) const {
-    return internal(std::forward<Args>(args)...);
+    return internal_invoker(this, std::forward<Args>(args)...);
   }
 
   explicit operator bool() const noexcept {
-    return (bool)internal;
+    return (bool)internal || internal_invoker;
   }
 
 private:
+  static R InvokeInternal(const move_only_function* self, Args... args) {
+    return self->internal(std::forward<Args>(args)...);
+  }
+
   std::function<R(Args...)> internal;
   void (*internal_destructor)(move_only_function*) = nullptr;
+  R (*internal_invoker)(const move_only_function*, Args...) = InvokeInternal;
   void* storage = nullptr;
 };
 } // namespace fextl


FEXCore/include/FEXCore/fextl/functional.h
```diff
@@ -47,15 +47,16 @@ class move_only_function<R(Args...), Alloc, Dealloc> {
...
       auto wrapped_lambda = [moved_lambda](Args... args) {
-        return (*moved_lambda)(std::forward<Args>(args)...);
+        return std::invoke(*moved_lambda, std::forward<Args>(args)...);
       };
 
       // Third, assign the result to std::function, ensuring it's indeed
       // allocation-free by checking for nothrow-constructibility
-      static_assert(noexcept(internal = std::move(wrapped_lambda)), "This implementation of std::function "
+      using T = std::decay_t<decltype(wrapped_lambda)>;
+      static_assert(std::is_nothrow_move_constructible<T>::value,   "This implementation of std::function "
```
lioncash:
```suggestion
      static_assert(std::is_nothrow_move_constructible_v<T>,   "This implementation of std::function "
```

neobrain (changes requested):
> This does not prove, that assigning the wrapper to std::function is noexcept or allocation-free, but avoids relying on libstdc++'s non-portable noexcept annotation and accepts libc++ as long as the wrapper type itself has the required nothrow move behavior.

The current code *guarantees* that the use of `std::function` is non-allocating, which is a hard requirement for FEX. Failing to compile when this guarantee cannot be provided is a feature. We can add a separate implementation for libc++ if needed, but it must provide the guarantee some other way then.

peppergrayxyz:
@neobrain I will further dig into libc++ and mark this as draft for now!

peppergrayxyz:
I couldn't find a way to assign to libc++ `std::function` while getting the same guaranties like from libstdc++.

However in the third branch, there is an intervention and storage allocation, so the benefit of the first two branches of being sleek and fast doesn't really hold here. So I was wondering why workaround `std::function` at all. Since there is a wrapper already, we could use its function pointer instead.

I set up a repo, extracted `functional.h` into a standalone test app and implemented an invoke function as fallback. [move_only_function, v1](https://github.com/peppergrayxyz/move_only_function/blob/a802753504a4433014458a99df7d2c726b717274/src/main_new.cpp#L70-L96):
```c++
  static_assert(!std::is_pointer_v<F>, "Pointer types must manually be dereferenced");

  // First, relocate argument to a location returned from FEX's allocators
  using Fnoref = std::remove_reference_t<F>;
  storage = Alloc(alignof(Fnoref), sizeof(Fnoref));
  auto moved_lambda = new (storage) Fnoref {std::move(f)};

  // Second, wrap the relocated argument in a single-capture lambda
  auto wrapped_lambda = [moved_lambda](Args... args) {
    return (*moved_lambda)(std::forward<Args>(args)...);
  };

  if constexpr (noexcept(internal = std::move(wrapped_lambda))) {
    // Third, assign the result to std::function, ensuring it's indeed
    // allocation-free by checking for nothrow-constructibility
    internal = std::move(wrapped_lambda);
  } else {
    // If this implementation of std::function does not support
    // nothrow-constructibility use a function pointer 
    auto invoker = +[](const move_only_function* self, Args... args) -> R {
      auto* moved_lambda = reinterpret_cast<Fnoref*>(self->storage);
      return (*moved_lambda)(std::forward<Args>(args)...);
    };
    static_assert(noexcept(internal_invoker = invoker));
    internal_invoker = invoker;
```
The alternative path doesn't rely on `std::function` behavior and actually saves on cpu cycles, so it could be just used alone. [move_only_function, v2](https://github.com/peppergrayxyz/move_only_function/blob/e8d999bd93fe93a47ab0fb2f5fd9c83cb038824b/src/main_new.cpp#L63-L77):
```c++
  // Other arguments require allocation, which is a problem since
  // std::function doesn't allow allocator customization. 
  static_assert(!std::is_pointer_v<F>, "Pointer types must manually be dereferenced");

  // Relocate argument to a location returned from FEX's allocators
  using Fnoref = std::remove_reference_t<F>;
  storage = Alloc(alignof(Fnoref), sizeof(Fnoref));
  new (storage) Fnoref {std::move(f)};

  // Create a wrapper and store a pointer to it
  internal_invoker = +[](const move_only_function* self, Args... args) -> R {
    auto* moved_lambda = reinterpret_cast<Fnoref*>(self->storage);
    return (*moved_lambda)(std::forward<Args>(args)...);
  };
```

I updated the PR accordingly to reflect this approach!

Sonicadvance1:
> I couldn't find a way to assign to libc++ `std::function` while getting the same guaranties like from libstdc++.
> 
> However in the third branch, there is an intervention and storage allocation, so the benefit of the first two branches of being sleek and fast doesn't really hold here. So I was wondering why workaround `std::function` at all. Since there is a wrapper already, we could use its function pointer instead.
> 
> I set up a repo, extracted `functional.h` into a standalone test app and implemented an invoke function as fallback. [move_only_function, v1](https://github.com/peppergrayxyz/move_only_function/blob/a802753504a4433014458a99df7d2c726b717274/src/main_new.cpp#L70-L96):
> 
> ```c++
>   static_assert(!std::is_pointer_v<F>, "Pointer types must manually be dereferenced");
> 
>   // First, relocate argument to a location returned from FEX's allocators
>   using Fnoref = std::remove_reference_t<F>;
>   storage = Alloc(alignof(Fnoref), sizeof(Fnoref));
>   auto moved_lambda = new (storage) Fnoref {std::move(f)};
> 
>   // Second, wrap the relocated argument in a single-capture lambda
>   auto wrapped_lambda = [moved_lambda](Args... args) {
>     return (*moved_lambda)(std::forward<Args>(args)...);
>   };
> 
>   if constexpr (noexcept(internal = std::move(wrapped_lambda))) {
>     // Third, assign the result to std::function, ensuring it's indeed
>     // allocation-free by checking for nothrow-constructibility
>     internal = std::move(wrapped_lambda);
>   } else {
>     // If this implementation of std::function does not support
>     // nothrow-constructibility use a function pointer 
>     auto invoker = +[](const move_only_function* self, Args... args) -> R {
>       auto* moved_lambda = reinterpret_cast<Fnoref*>(self->storage);
>       return (*moved_lambda)(std::forward<Args>(args)...);
>     };
>     static_assert(noexcept(internal_invoker = invoker));
>     internal_invoker = invoker;
> ```
> 
> The alternative path doesn't rely on `std::function` behavior and actually saves on cpu cycles, so it could be just used alone. [move_only_function, v2](https://github.com/peppergrayxyz/move_only_function/blob/e8d999bd93fe93a47ab0fb2f5fd9c83cb038824b/src/main_new.cpp#L63-L77):
> 
> ```c++
>   // Other arguments require allocation, which is a problem since
>   // std::function doesn't allow allocator customization. 
>   static_assert(!std::is_pointer_v<F>, "Pointer types must manually be dereferenced");
> 
>   // Relocate argument to a location returned from FEX's allocators
>   using Fnoref = std::remove_reference_t<F>;
>   storage = Alloc(alignof(Fnoref), sizeof(Fnoref));
>   new (storage) Fnoref {std::move(f)};
> 
>   // Create a wrapper and store a pointer to it
>   internal_invoker = +[](const move_only_function* self, Args... args) -> R {
>     auto* moved_lambda = reinterpret_cast<Fnoref*>(self->storage);
>     return (*moved_lambda)(std::forward<Args>(args)...);
>   };
> ```
> 
> I updated the PR accordingly to reflect this approach!

As you can see by our CI, it's a requirement from the API that it is move-only and allocation-free. This is confirmed by libstdc++'s implementation having small-object optimization not allocating any memory.

If libstdc++ can do it, then libc++ should be able to do so as well.

peppergrayxyz:
I think this issue is above my pay grade :'(

Sonicadvance1:
Sounds like this isn't going anywhere right now.

Issue #3785, opened 2024-06-30
AVX128: vpshufhw can be improved


Sonicadvance1:
Similar to #3784, this is just using a simple vinselement loop instead of anything smart, can be improved dramatically.

PR #5478, opened 2026-05-03
Perf/pshufw shuffle search

Hinoaaaaaf212:


Hinoaaaaaf212:
**OpcodeDispatcher: Specialize MMX PSHUFW shuffles**
  
  `PShufWLane` was only special-casing the identity copy and the four single-element broadcasts; everything else fell
  through to a `TBL`.

  For MMX `pshufw` the shuffled half is the entire register, so 11 more immediates can be done in a single ARM64 NEON op
   (rotates via `EXT`, reversals via `REV32`/`REV64`, even/odd via `ZIP`/`UZP`/`TRN`). Same approach as #3871 did for
  `PSHUFD`.

  128-bit `PSHUFLW`/`PSHUFHW` still need to preserve the untouched half so they've been left.

  Fixes part of https://github.com/FEX-Emu/FEX/issues/3785

Sonicadvance1:
Thanks for the attempted LLM implementation, but actually went through and solved it in #5482 instead.

Hinoaaaaaf212:
Hi, this is my first time working on something like this so it might seem fairly basic, I didn't use an LLM to code but to explain for my comment, if you have the time to if you're willing to point out what I did wrong/do better I'd be really happy, thank you in advance! :)

PR #5462, opened 2026-05-03
determine sigset_t fieldname to enhance portability (musl)

peppergrayxyz:
musl build fails due to access to internal glibc member:

```
FEX/Source/Tools/LinuxEmulation/LinuxSyscalls/SignalDelegator.cpp:655:39: error: no member named '__val' in '__sigset_t'
  655 |       .SigMask = _context->uc_sigmask.__val[0],
      |                  ~~~~~~~~~~~~~~~~~~~~ ^
1 error generated.
```

add check to determine private glibc or musl member name or throw an error.

fixes: #5461

PR #5480, opened 2026-05-04
SignalDelegator: Simplify support for musl's sigset_t

neobrain:
No need to do build system trickery to handle both musl and glibc, just use structured bindings instead.

See #5462 for context.


neobrain:
@peppergrayxyz Could you double-check this works with musl? It doesn't look like it differs much from glibc internally, but it's better to verify since I don't have a musl build setup to test with.

peppergrayxyz:
@neobrain I checked it on Gentoo and Alpine and it compiles! 

I think it would be a good addition to add some testing, that glibc and musl set the bits the same way and as a precaution if this is to be extended to another c-lib at some point.

Sonicadvance1 (approved):
As long as the offsets are the same this doesn't matter between musl and glibc anyway. This is struct passed directly from the kernel to userspace on the stack, we're just reusing the libc struct definition because it's easy.

PR #5482, opened 2026-05-04
OpcodeDispatcher: Optimize MMX pshufw 

### LINKED ISSUE #3785: AVX128: vpshufhw can be improved

Sonicadvance1:
Found through writing a shuffle solver rather than an LLM.

Fixes https://github.com/FEX-Emu/FEX/issues/3785


lioncash (approved):
shuffling all the things

PR #5489, opened 2026-05-11
FEXBash: Drop implicit -c and add colored PS1

neobrain:
Previously, `FEXBash steam.sh -silent` didn't actually pass `-silent` to the script, which differs from the behavior of `bash steam.sh -silent`. The culprit is the implicit use of `-c`, which can't easily be opted out of. Instead, just drop this behavior and require `-c` to be passed manually if its semantics are desired (just like with regular bash).

Additionally, the `FEXBash` string in the PS1 command line now has colors matching the FEX logo for no particular reason other than being cute:
Screenshot_20260511_203629

### DIFF
--- a/Source/Tools/FEXBash/FEXBash.cpp
+++ b/Source/Tools/FEXBash/FEXBash.cpp
@@ -2,7 +2,7 @@
 /*
 $info$
 tags: Bin|FEXBash
-desc: Launches bash under FEX and passes arguments via -c to it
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
Luckily this won't break anything anymore. Its behaviour used to be load bearing but that isn't the case anymore.

Sonicadvance1 (approved):

Issue #5399, opened 2026-03-27
World of Warcraft fails to launch


chrisRidgers:
Platform: Asahi Linux
Version: FC34
Wine Version: wine-10.15 (system default version, exists inside muvm stack)
Proton Version: Proton Experimental

Battle.net runs via Steam/ Proton in Asahi's muvm/ FEX stack.  However, trying to launch World of Warcraft *nothing* happens.  *Nothing* being a process that runs but never get's as far as creating a window - doesn't seem to crash out though.  Through steams cli output I can see the process eventually get's shut down if you exit Battle.net and ultimately Steam.

Managed to get some more output from by running proton manually from muvm and FEXBash.  Looks like the process is just running the same thing repeatedly until it's killed.

Spoke to the Aasahi community and they suggested opening a ticket here.

So using the following command:

```
chris@snow:~$ PROTON_LOG=1 STEAM_COMPAT_DATA_PATH=~/.local/share/Steam/steamapps/compatdata/4263109722 STEAM_COMPAT_CLIENT_INSTALL_PATH=~/.local/share/Steam ~/.local/share/Steam/steamapps/common/Proton\ -\ Experimental/proton run ~/.local/share/Steam/steamapps/compatdata/4263109722/pfx/drive_c/Program\ Files\ \(x86\)/World\ of\ Warcraft/_retail_/Wow.exe 
```

Get's an output of this:

```
E 310|310 0.000 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
E 310|310 0.030 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
I 310|310 0.156 CLONE_CLEAR_SIGHAND passed to clone3. Returning EINVAL.
I 310|310 0.156 clone: Unsupported flags w/o CLONE_THREAD (Shared Resources), 4100
E 315|315 0.165 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
fsync: warning: a previous shm file /wine-5f71a9-fsync was not properly removed
fsync: up and running.
E 318|318 0.310 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
E 320|320 0.683 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
E 324|324 1.066 Failed to remap /proc/pid/cmdline data (prctl failed: result -1, errno 1)
[...]

PR #5492, opened 2026-05-13
Windows: Trace interrupt translation and prototype INT 0x29 fail-fast mapping

FrontMage:
## Summary

This adds a narrow interrupt-tracing path for the Windows bridge and prototypes a Windows-style fail-fast mapping for raw `INT 0x29`.

## What changed

- Add opt-in `FEX_TRACE_INTERRUPTS=1` logging in:
  - `OpcodeDispatcher::INTOp()`
  - Windows exception conversion
  - ARM64EC exception rethrow bridge
- Add opt-in `FEX_WIN_INT29_FAILFAST=1` handling for guest `INT 0x29`
- Map `INT 0x29` to `STATUS_STACK_BUFFER_OVERRUN (0xC0000409)` instead of the current `EXCEPTION_ILLEGAL_INSTRUCTION`
- Propagate one exception parameter from the guest register state as a provisional fast-fail code

## Why

Direct probe executables show:

- `INT3` currently reaches user mode as `EXCEPTION_BREAKPOINT` and is resumable
- `INT 0x2D` currently reaches the existing breakpoint-style special case
- baseline `INT 0x29` currently reaches user mode as `EXCEPTION_ILLEGAL_INSTRUCTION (0xC000001D)`
- with this patch enabled, `INT 0x29` becomes `0xC0000409` and carries one parameter

This makes the `INT 0x29` path closer to Windows fail-fast semantics and gives us a cleaner basis for debugging Windows loaders that use trap-driven control flow.

## Notes

- The fail-fast mapping is intentionally gated behind `FEX_WIN_INT29_FAILFAST=1`
- The current implementation uses `CX` as the provisional source for `ExceptionInformation[0]`
- This is intended as a debugging/validation step, not a final claim that all Windows `__fastfail` details are now correct

## Validation

Validated with minimal Windows x64 probes for:

- `INT3`
- `INT 0x2D`
- `INT 0x29`

Observed behavior:

- baseline `INT 0x29` -> `0xC000001D`
- patched `INT 0x29` -> `0xC0000409`
- patched path propagates one parameter (`0x1234` in the probe)

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -496,7 +496,7 @@ static void RethrowGuestException(const EXCEPTION_RECORD& Rec, ARM64_NT_CONTEXT&
   EFlags &= ~(1 << FEXCore::X86State::RFLAG_TF_RAW_LOC);
   CTX->SetFlagsFromCompactedEFLAGS(Thread, EFlags);
 
-  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X0);
+  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X8, Args->Context.X0);
   if (Args->Rec.ExceptionCode == EXCEPTION_SINGLE_STEP) {
     Args->Context.Cpsr &= ~(1 << 21); // PSTATE.SS
   } else if (Args->Rec.ExceptionCode == EXCEPTION_BREAKPOINT) {
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -496,7 +496,7 @@ static void RethrowGuestException(const EXCEPTION_RECORD& Rec, ARM64_NT_CONTEXT&
   EFlags &= ~(1 << FEXCore::X86State::RFLAG_TF_RAW_LOC);
   CTX->SetFlagsFromCompactedEFLAGS(Thread, EFlags);
 
-  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X8);
+  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X0);
   if (Args->Rec.ExceptionCode == EXCEPTION_SINGLE_STEP) {
     Args->Context.Cpsr &= ~(1 << 21); // PSTATE.SS
   } else if (Args->Rec.ExceptionCode == EXCEPTION_BREAKPOINT) {
--- a/Source/Windows/Common/Exception.h
+++ b/Source/Windows/Common/Exception.h
@@ -11,7 +11,7 @@
 namespace FEX::Windows {
 template<typename TReg>
 static inline EXCEPTION_RECORD
-HandleGuestException(FEXCore::Core::CpuStateFrame::SynchronousFaultDataStruct& Fault, const EXCEPTION_RECORD& Src, TReg& Rip, TReg Rax) {
+HandleGuestException(FEXCore::Core::CpuStateFrame::SynchronousFaultDataStruct& Fault, const EXCEPTION_RECORD& Src, TReg& Rip, TReg Cx) {
   EXCEPTION_RECORD Dst = Src;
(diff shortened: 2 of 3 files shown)


FrontMage:
This addresses #5399 .

bylaws:
I think I would rather see this implementing full handling of the fast fail stuff since it's not much more, and no need to gate this behind env bar

bylaws:
(wine tests should describe behavior requires for this in full)

Sonicadvance1:
There's a bunch of unrelated changes in this PR, some of which already exist in upstream.
Plus I don't think we care about tracing, instead just implementing the fast-fail handling?

FrontMage:
@Sonicadvance1 Sorry for the premature PR. I pushed the earlier branch too early while it still contained local probing/tracing scaffolding and a provisional env-gated implementation.

I have force-updated this branch to current `FEX-Emu:main` and rebuilt the PR as a single narrow commit: `Windows: Handle INT 0x29 as fast-fail`.

The reason this started as a traced prototype was that the captured guest logs showed raw `INT 0x29` being decoded as `CD 29`, converted through the generated `#GP` path, and reaching user mode as `EXCEPTION_ILLEGAL_INSTRUCTION (0xC000001D)` before this handling existed. With the prototype handling enabled, the same probe path converted it to `0xC0000409` with one exception parameter:

```text
[FEX-INT] decode rip=0x140001B0D op=0xCD imm=0x29 signal=11 trap=13 err=0x14A
[FEX-INT] convert.int29 mode=failfast rip=0x140001B0D code=0xC0000409 info0=0x1234
[FEX-INT] bridge.post ... exception=0xC0000409 addr=0x140001B0D params=1 info0=0x1234

**Comment by bylaws** [comment]
>The current implementation uses RAX as the provisional source for ExceptionInformation[0]
this is incorrect for the breakpoint path


see https://github.com/bylaws/wine/blob/arm64ec/dlls/ntdll/unix/signal_x86_64.c#L2144

**Comment by FrontMage** [comment]
> > The current implementation uses RAX as the provisional source for ExceptionInformation[0]
> > this is incorrect for the breakpoint path
> 
> see https://github.com/bylaws/wine/blob/arm64ec/dlls/ntdll/unix/signal_x86_64.c#L2144

In the recent commit, I've changed it to Cx.

File: Source/Windows/ARM64EC/Module.cpp [hunk 1]
```diff
@@ -496,7 +496,7 @@ static void RethrowGuestException(const EXCEPTION_RECORD& Rec, ARM64_NT_CONTEXT&
   EFlags &= ~(1 << FEXCore::X86State::RFLAG_TF_RAW_LOC);
   CTX->SetFlagsFromCompactedEFLAGS(Thread, EFlags);
 
-  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X8);
+  Args->Rec = FEX::Windows::HandleGuestException(Fault, Rec, Args->Context.Pc, Args->Context.X0);
```
Sonicadvance1: Why change which register is used? What was wrong with using rax like the 0x2d exception below before it was modified??

bylaws: It needs both as an argument

FrontMage: You're right, sorry about that. I focused on correcting INT 0x29 to use CX/ECX and missed that switching the shared path globally would regress the existing INT 0x2D handling.

I've pushed a follow-up fix so `HandleGuestException()` receives both RAX and CX now. INT 0x29 uses CX/ECX, while INT 0x2D keeps using RAX/EAX.

I validated this with probes where EAX != ECX:
- INT 0x2D: EAX=0x2D2D, ECX=0xCACA, delivered `0x80000003` with `info0=0x2D2D` and resumed.
- INT 0x29: EAX=0x1111, ECX=0x2222, delivered `0xC0000409` with `info0=0x2222`.

So the older INT 0x2D path should no longer be broken by the INT 0x29 fix.

Sonicadvance1: Okay, the new changes make more sense now.

FrontMage: Thx for the review! Fixed some formatter issues and switched it to the open state now.

Sonicadvance1 (approved):
Seems reasonable to me now.

PR #5500, opened 2026-05-17
CodeCache: Accept multiple threads in SaveData

bylaws:
Needs some consideration of if this is the cleanest way, but this is certainly the simplest.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -23,6 +23,13 @@
           "Enable the code caching subsystem"
         ]
       },
+      "EnableLazyCodeCachingWIP": {
+        "Type": "bool",
+        "Default": "false",
+        "Desc": [
+          "Enable lazy loading of chunks in code caches"
+        ]
+      },
       "EnableCodeCacheValidation": {
         "Type": "bool",
         "Default": "false",
--- a/FEXCore/Source/Interface/Context/Context.cpp
+++ b/FEXCore/Source/Interface/Context/Context.cpp
@@ -53,6 +53,6 @@ FEXCore::CPUID::FunctionResults FEXCore::Context::ContextImpl::RunCPUIDFunctionN
 }
 
 bool FEXCore::Context::ContextImpl::IsAddressInCodeBuffer(FEXCore::Core::InternalThreadState* Thread, uintptr_t Address) const {
-  return Thread->CPUBackend->IsAddressInCodeBuffer(Address);
+  return Thread->CPUBackend->IsAddressInCodeBuffer(Address) || CodeCache.IsAddressInMappedCodeBuffer(Address);
 }
 } // namespace FEXCore::Context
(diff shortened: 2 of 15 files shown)


Sonicadvance1:
Rebase necessary.

neobrain (changes requested):
Could you PR an actual implementation of multithreaded writing instead so that there's a meaningful base for discussion? The interface change to SaveData on its own isn't very useful.

Regarding mmap writing, I would prefer this to stay in a PR for now. What we have today is easier to iterate on, and I've got upcoming implementation changes in SaveData that would interfere unnecessarily. If serialization time is critical, simple buffered writing would work as a stop-gap.

bylaws:
>Regarding mmap writing, I would prefer this to stay in a PR for now. What we have today is easier to iterate on, and I've got upcoming implementation changes in SaveData that would interfere unnecessarily. If serialization time is critical, simple buffered writing would work as a stop-gap.

I think it should be fairly easy to iterate on, like I explicitly did it in a generic way that is trivial to extend without new address logic? I think unless the code quality is meaningfully bad it would be best to merge instead of asking to reinvent the wheel again, delaying it again just means more conflict and more effort for whoever ends up needing to clean it up. And like it makes things 10x faster or smth so may as well take the win.

Could you PR an actual implementation of multithreaded writing instead so that there's a meaningful base for discussion? The interface change to SaveData on its own isn't very useful.
This is dependent on all my existing PRs landing, I can do so but it would mean there's like 5 stacked PRs, if you feel that's best sure I can do

neobrain:
> I think it should be fairly easy to iterate on, like I explicitly did it in a generic way that is trivial to extend without new address logic? I think unless the code quality is meaningfully bad it would be best to merge instead of asking to reinvent the wheel again, delaying it again just means more conflict and more effort for whoever ends up needing to clean it up. And like it makes things 10x faster or smth so may as well take the win.

I'm only measuring a 2x improvement (both for small and large caches), which can equally be achieved with simple buffering that doesn't require us to awkwardly split IO logic or to do a double-pass. (My guess is your 10x measurement was made before I addressed the trivial low-hanging fruit in SaveData?)

> > Could you PR an actual implementation of multithreaded writing instead so that there's a meaningful base for discussion? The interface change to SaveData on its own isn't very useful.
> This is dependent on all my existing PRs landing, I can do so but it would mean there's like 5 stacked PRs, if you feel that's best sure I can do

How is it dependent? Surely this can already be illustrated Linux-side with everything being upstream there already?

bylaws:
How is it dependent? Surely this can already be illustrated Linux-side with everything being upstream there already?

I have not implemented this for the Linux side and do not have time to.  If you want me to pr windows side I can.

I'm only measuring a 2x improvement (both for small and large caches), which can equally be achieved with simple buffering that doesn't require us to awkwardly split IO logic or to do a double-pass. (My guess is your 10x measurement was made before I addressed the trivial low-hanging fruit in SaveData?)

I really don't think it's that awkward to do two passes, 2x improvement is pretty good and meaningful. Especially in the future mmap is much more flexible that just chained writes, and FEX is pretty well acquainted with mmap anyway :)

PR #5507, opened 2026-05-22
CodeCache: Resolve comment https://github.com/FEX-Emu/FEX/pull/5441#discussion_r3255384588

Sonicadvance1:


lioncash (approved):

neobrain:
Did you test if this actually works as intended? The lifetime management around this is rather tricky and I was going to verify before updating the PR.

Sonicadvance1:
> Did you test if this actually works as intended? The lifetime management around this is rather tricky and I was going to verify before updating the PR.

Regardless of span lifetime management, this would have been incorrect as it would then be dereferencing an empty span and passing it to munmap instead. It's non-fatal in a typical setup in both cases, better to at least be correct for the commented line. The lifetime management problems can be sorted after the fact.

neobrain:
> The lifetime management problems can be sorted after the fact.

Are you planning on taking over addressing this then? Seems bold to make that decision on behalf of the code's author without any coordination, especially if your assumption "non-fatal in typical setup" isn't backed by testing.

PR #5509, opened 2026-05-22
Windows: Repair missed executable range queries

dllu:
This is PR 2 out of 4 that allows Starcraft II to work on my DGX Spark, the others being:

* #5508
* #5510 
* #5511 

BEFORE:

image

AFTER:

image

## Problem

The Windows invalidation tracker can lose an executable interval while the
guest address is still committed executable according to `VirtualQuery`.
When the frontend queries that address during compilation, the stale interval
state makes FEX treat real executable guest memory as non-executable.

StarCraft II x64 triggers this during startup after its protected exception
sequence. Removing this change from the reduced ARM64 Proton fix stack stops
the Battle.net-owned x64 launch before a new graphics log is created and
returns StarCraft II startup error `2:1046976`.

## Reproduction

The game-level reproduction is:

```sh

  --exec="launch S2"
```

The lower-level pattern is a Windows executable-range notification sequence
that invalidates an interval while the memory state remains committed
executable, followed by compilation from that entrypoint.

## Fix

Keep the existing interval query fast path. On a miss, ask `VirtualQuery`
whether the reached guest address is still committed executable. If it is,
insert the aligned executable interval directly under the interval lock and
retry the query. The fallback avoids the normal notification path because the
decoder already holds the code invalidation lock while querying.

## Risk

The fallback trusts current Windows memory protection state after an interval
miss. That is narrower than broadly ignoring interval tracking, but it can
recreate a missed executable range that tracking meant to remove if the
protection query is stale or a caller relies on the miss for something other
than current protection state. The locking behavior is also worth review
because this runs from compilation.

### DIFF
--- a/Source/Windows/ARM64EC/Module.cpp
+++ b/Source/Windows/ARM64EC/Module.cpp
@@ -846,17 +846,21 @@ NTSTATUS NotifyMapViewOfSection(void* Unk1, void* Address, void* Unk2, SIZE_T Si
 }
 
 void NotifyUnmapViewOfSection(void* Address, BOOL After, NTSTATUS Status) {
+  static thread_local FEX::Windows::InvalidationTracker::InvalidateContainingSectionResult PendingUnmap {};
+
   if (!InvalidationTracker || !GetCPUArea().ThreadState()) {
     return;
   }
 
   if (!After) {
     ThreadCreationMutex.lock();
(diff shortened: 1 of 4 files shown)


bylaws:
Why exactly is the change in the permissions being missed in the tracking data structure, this isn't intended to be lazily populated and doing so might break things. Could you figure out what is causing the change to be missed?

bylaws:
this is wrong - but i get the vibe you're going for. If you wanna make a c program that reprooduces i can fix it in a cleaner way, but I can probably figure out from this issue.

PR #5511, opened 2026-05-22
ARM64EC: Repair copied Wine syscall thunks before decoding

dllu:
This is PR 4 out of 4 that allows Starcraft II to work on my DGX Spark, the other ones being:

* #5508
* #5509 
* #5510 

BEFORE:

image

AFTER:

image

## Problem

Wine's ARM64EC syscall exports can point at small ARM64 syscall wrappers. Some
x64 guests copy those wrapper bytes into private executable memory and later
reach the copy from x64 control flow. The ARM64EC image metadata no longer
describes the private copy, so FEX's x64 frontend attempts to decode ARM64
instructions as x64 guest code.

StarCraft II x64 performs this copy during protected startup. After the earlier
privileged-fault and executable-range prerequisites are fixed, this copied
wrapper boundary is the remaining decoder stop before the client renders.

## Reproduction

A focused reproducer can copy a Wine ARM64EC `Nt*` syscall export wrapper to a
private executable allocation and jump to the copy from x64 guest control flow.

The game-level reproduction is:

```sh

  --exec="launch S2"
```

The reduced integration stack reaches rendered StarCraft II x64 menu and replay
gameplay after this branch is present. Without copied-wrapper handling, the
startup path tries to decode copied ARM64 Wine thunk bytes as x64 code.

## Fix

Add a pre-compile guest-code preparation hook. The ARM64EC Windows handler uses
it only for committed private executable memory whose first six instructions
match Wine's ARM64 syscall-wrapper signature. It extracts the Wine syscall id,
rewrites the reached 24-byte private copy to Wine's equivalent x64 syscall
thunk, flushes the instruction cache, and invalidates the patched guest range.

## Risk

This is intentionally a narrow runtime code rewrite, but it still matches an
instruction signature in private executable memory and temporarily changes page
protection around the reached entrypoint. Review should focus on signature
specificity, cross-page protection span handling, cache invalidation, and
whether Wine can expose a less copy-sensitive thunk shape instead.

### DIFF
--- a/FEXCore/Source/Interface/Core/CodeCache.cpp
+++ b/FEXCore/Source/Interface/Core/CodeCache.cpp
@@ -47,7 +47,7 @@ MappedCodeCacheFile::~MappedCodeCacheFile() {
   }
 
 #ifndef _WIN32
-  if (CodeBuffer.empty()) {
+  if (!CodeBuffer.empty()) {
     FEXCore::Allocator::munmap(CodeBuffer.data(), CodeBuffer.size_bytes());
   }
 #endif


bylaws:
This isn't the right fix for the problem. The way to do this is if you look at apply_arn64x_relocatioms in wine, you can make NtReadFile detect if the source file is a dll and apply relocations into the target buffer.

PR #5524, opened 2026-05-22
split unit and integration tests

### LINKED ISSUE #5523: tests/fex_linux_tests_all relies on install step ("Failed to open lib")
When running `fex_linux_tests_all` the test fails at `thunk_testlib.32.jit.flt` with an error:

```
/run/host/home/pepper/dev/ghrepos/FEX/unittests/FEXLinuxTests/tests/thunks/thunk_testlib.cpp:44
...............................................................................

/run/host/home/pepper/dev/ghrepos/FEX/unittests/FEXLinuxTests/tests/thunks/thunk_testlib.cpp:44: FAILED:
due to unexpected exception with message:
  Failed to open lib
```

The search path is set to the install-dir, which does not exist at this point:

https://github.com/FEX-Emu/FEX/blob/268081e5d088c4a848adc68b8ca0c2486499f0eb/unittests/FEXLinuxTests/CMakeLists.txt#L91-L95

peppergrayxyz:
running `fex_linux_tests_all` tests againts the install path. Running
the test target fails therefore at `thunk_testlib.32.jit.flt` with a
"Failed to open lib" error, when the path does not exist (yet).

This commit splits `fex_linux_tests_all` into
- `fex_linux_portable_tests_all` -> test in `${CMAKE_BINARY_DIR}/install`
- `fex_linux_global_tests_all`   -> test in `${CMAKE_INSTALL_PREFIX}`

The portable test calls the install target with --prefix pointing to the
build dir. Then runs the test with `FEX_PORTABLE=1` from there.

The global test calls the target in the install dir and runs with
`FEX_PORTABLE=0`. Its tests are labeled as `integration`.

A configurtion for the default test target `ninja test` is provided,
that excludes all tests with the `integration` label and a `ninja
test_integration` test target to run them after install.

This enables to run the default test target without running into errors
from the integration tests, but run them separately after install:

```
cd build
CC=clang CXX=clang++ cmake -DBUILD_TESTING=True -DBUILD_THUNKS -G Ninja ..
ninja
ninja test
sudo ninja install
ninja test_integration
```

Fix: https://github.com/FEX-Emu/FEX/issues/5523


Sonicadvance1:
This isn't a correct fix. The thunklib test is expected to be able to dlopen `libfex_thunk_test.so` and it gets transparently replaced with `libfex_thunk_test-guest.so`, which is one of the points of the test!
This is why our CI does a `make install` step before running these!

Sonicadvance1:
https://github.com/FEX-Emu/FEX/blob/main/.github/workflows/ccpp.yml#L51-L56
build+install step here.

peppergrayxyz:
That explains a lot! I wasn't aware that this is not only a unit, but also an integration test.

I still think there is value in running `ninja test` and have all unit tests executed without some of them failing, because the install step has not been run. I updated the PR to run the unit and integration tests individually. This should work correctly now!

neobrain:
Thanks for looking into this, but I don't think "run tests locally without installation" is a workload we're interested in supporting. It just adds a bunch of build system complexity for a configuration we'll never use, and requiring installation isn't that outrageous.

PR #5535, opened 2026-05-28
FEXCore: handle LODS address-size override and invalid LOCK prefixes

FrontMage:
## Summary

This handles two x86 instruction edge cases that showed up while debugging the Windows loader path:

- `LODS` with an address-size override in 64-bit mode
- clearly invalid `LOCK` prefixes that should raise an invalid-op path instead of being reported as missing lock-handler coverage

This addresses #5399.

## What changed

- Allow `LODS` in 64-bit mode to honor `0x67` address-size override semantics.
- Use 32-bit `ESI` address calculation for overridden `LODS`, then zero-extend the `RSI` writeback.
- Keep 32-bit mode `LODS` address-size override unsupported, matching the existing string-op guard style.
- Detect invalid `LOCK` prefix forms before opcode dispatch for:
  - `LOCK MOV`
  - register-destination ALU encodings such as `LOCK ADD reg, [mem]`
- Add an ASM test for `0x67 lodsd` to verify the load address and `RSI` writeback behavior.

## Why

The loader path hit `0x67 lodsd` while running in 64-bit mode. x86-64 defines the address-size override for string operations, so this should use the 32-bit string address register rather than failing decode.

The same investigation also exposed invalid `LOCK` prefixes. Those should be treated as invalid guest instructions when the form is architecturally invalid, rather than falling through to the generic "Missing LOCK HANDLER" diagnostic.

## Notes

- The `LOCK` handling here is intentionally narrow. It does not try to classify every possible invalid `LOCK` prefix form.
- Valid locked memory forms continue to use the existing opcode-handler paths.
- The new `LOCK` pre-check only handles forms that are clearly invalid before dispatch.

## Validation

- Changed files are limited to:
  - `FEXCore/Source/Interface/Core/Core.cpp`
  - `FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp`
  - `unittests/ASM/Primary/Primary_AD_addrmod.asm`
- The new ASM test assembles under the FEX harness pattern with `BITS 64` prepended and a trailing `ret` appended.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -121,6 +121,54 @@ static GetFrameBlockInfoResult GetFrameBlockInfo(FEXCore::Core::CpuStateFrame* F
   return {InlineHeader, nullptr};
 }
 
+static bool IsInvalidLockPrefix(const X86Tables::DecodedInst* DecodedInfo, const X86Tables::X86InstInfo* TableInfo) {
+  if (!(DecodedInfo->Flags & X86Tables::DecodeFlags::FLAG_LOCK)) {
+    return false;
+  }
+
+  if (!TableInfo || !TableInfo->Name) {
+    return true;
+  }
+
+  const std::string_view Name {TableInfo->Name};
+  if (Name == "MOV") {
+    return true;
+  }
+
+  // These ALU encodings have a register destination and a memory/register source,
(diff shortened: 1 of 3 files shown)


Sonicadvance1:
I'll look in to reimplementing these as they aren't quite exactly how I want them to be laid out in the JIT.

Sonicadvance1:
Fixed by #5543 and #5542

PR #5536, opened 2026-05-28
LibraryForwarding: Implement support for CUDA

Sonicadvance1:
This is enough to get less complex cuda applications running, and is a good starting spot to slowly finish off the remaining implementation.

Some information:
- 429 functions in total
- 153 only compiled for 64-bit (35.6%)
- 7 functions disabled entirely (1.6%)

The main thing /not/ working with this initial implementation is .cu files compiled in to an ELF using the static cuda runtime. This is due to the `cuGetExportTable` function being stubbed out and the static cuda RT requires at least two interfaces from that function before it continues.

This function isn't publicly documented by NVIDIA but has been publicly reverse engineered to be fairly trivial. It's just a jump table with the first element being the size of the table in bytes.

That will be the next step of the implementation.


ThunkLibs/libcuda/fixup_defines.h
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <cstddef>
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef void (*CUhostFn)(void*);
```
neobrain, line 8: What's the purpose of this file?

ThunkLibs/libcuda/cuda_defines.h
```diff
@@ -0,0 +1,1017 @@
+// SPDX-License-Identifier: MIT
+// Generated file. Be careful if modifying
```
neobrain, line 2: How was it generated? Should have reproducible instructions for re-generating in the future.

ThunkLibs/libcuda/cuda_defines.h
```diff
@@ -0,0 +1,1017 @@
...
+#pragma once
```
neobrain, line 3: Why can't we use the proper cuda.h headers?

ThunkLibs/libcuda/libcuda_Guest.cpp
```diff
@@ -0,0 +1,84 @@
...
+});
+
+extern "C" {
+
+// This variable controls the behavior of cuGetProcAddress for functions we don't know the signature of:
+// - if false (default), we return a nullptr (since the application might have a fallback code path)
+// - if true, we return a stub function that fatally errors upon being called
+constexpr bool stub_unknown_functions = false;
+
+// Fatally erroring function with a thunk-like interface. This is used as a placeholder for unknown CUDA functions
```
neobrain, line 31: What's a "thunk-like" interface? Would prefer not unnecessarily overloading "thunk" even more than we already do - it's a meaningless term at this point.

ThunkLibs/libcuda/libcuda_Host.cpp
```diff
@@ -0,0 +1,76 @@
...
+  return ret;
+}
+
+CUresult FEXFN_IMPL(cuGetExportTable)(guest_layout<const void**> ppExportTable, const CUuuid* pExportTableId) {
+  // https://github.com/vosen/ZLUDA/blob/1b9ba2b2333746c5e2b05a2bf24fa6ec3828dcdf/zluda_dark_api/src/lib.rs#L197
+  //
+  // [0x6b, 0xd5, 0xfb, 0x6c, 0x5b, 0xf4, 0xe7, 0x4a, 0x89, 0x87, 0xd9, 0x39, 0x12, 0xfd, 0x9d, 0xf9]
+  // - cudart interface (0x68 bytes?)
+  // [0xa0, 0x94, 0x79, 0x8c, 0x2e, 0x74, 0x2e, 0x74, 0x93, 0xf2, 0x08, 0x00, 0x20, 0x0c, 0x0a, 0x66]
+  // - Tools runtime callback hooks (0x38 bytes?)
```
neobrain: Could you add some context to this? Right now that's a list of random numbers.

ThunkLibs/libcuda/libcuda_Host.cpp
```diff
@@ -0,0 +1,76 @@
...
+  std::string uuid {};
+  for (size_t i = 0; i < sizeof(CUuuid); ++i) {
+    const bool last = (i + 1) == sizeof(CUuuid);
+    char tmp[8];
+    auto size = snprintf(tmp, 8, "0x%02x%s", pExportTableId->_0[i], last ? "" : ", ");
+    uuid += std::string_view(tmp, size);
+  }
```
neobrain, line 75: We can use fmt in these modules, so this can simply be

```suggestion
auto uuid = fmt::format("{:#02x}", fmt::join(CUuuid, ", "));
```

neobrain (commented):
What type of applications are working with this initial implementation?

ThunkLibs/libcuda/libcuda_Host.cpp
Sonicadvance1: We can't use format for these. I had already tried that but CI cross-compilers are too old for it.

ThunkLibs/libcuda/libcuda_Guest.cpp
Sonicadvance1: This comment was copied wholesale from the Vulkan thunks. So however you want this comment, it should be updated there as well.

ThunkLibs/libcuda/cuda_defines.h
Sonicadvance1: We can't and don't want to guarantee cuda to be installed in CI for building cuda. It's not an expected dependency unlike our other thunks.

ThunkLibs/libcuda/fixup_defines.h
Sonicadvance1: Providing type declarations from cuda that can't be auto generated easily.

Sonicadvance1:
> What type of applications are working with this initial implementation?

Various sample programs that dlopen the cuda interface and query hardware details. Technically kernel running could work but I didn't have any on hand to attempt. I wasn't going to attempt any complex applications until I have the `cuGetExportTable` UUID handling implemented, since a lot of things link the static cuda runtime and mandate it.

ThunkLibs/libcuda/libcuda_Guest.cpp
neobrain: Oops noted, thanks! Fine to leave as-is for this PR then, I'll note it down for rewording later.

neobrain:
> Various sample programs that dlopen the cuda interface and query hardware details. Technically kernel running could work but I didn't have any on hand to attempt. I wasn't going to attempt any complex applications until I have the cuGetExportTable UUID handling implemented, since a lot of things link the static cuda runtime and mandate it.

Anything specific you recommend for testing? Not familiar with the CUDA ecosystem, so if there's any particular example repo and binaries within that are expected to work, that would be helpful for testing.

Sonicadvance1:
> > Various sample programs that dlopen the cuda interface and query hardware details. Technically kernel running could work but I didn't have any on hand to attempt. I wasn't going to attempt any complex applications until I have the cuGetExportTable UUID handling implemented, since a lot of things link the static cuda runtime and mandate it.
> 
> Anything specific you recommend for testing? Not familiar with the CUDA ecosystem, so if there's any particular example repo and binaries within that are expected to work, that would be helpful for testing.

I just looked up cuda samples.

https://github.com/NVIDIA/cuda-samples/tree/master/cpp These ones.

ThunkLibs/libcuda/libcuda_Host.cpp
Sonicadvance1: Added a comment.

ThunkLibs/libcuda/cuda_defines.h
Sonicadvance1: https://codeberg.org/Sonicadvance1/thunkgen_desc
```
time ./Bin/def_extract -d /mnt/Work/Work/FEX/ThunkLibs/libcuda/cuda_defines.h -i /mnt/Work/Work/FEX/ThunkLibs/libcuda/libcuda_interface.cpp -o test.json -j ../formats/cuda/cuda.json -- ../formats/cuda/cuda.h
```

ThunkLibs/libcuda/libcuda_Guest.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+struct override_entry {
+  std::string_view name;
+  void* ptr;
+};
+
+constexpr static std::array<override_entry, 2> proc_override = {
+  {{"cuGetProcAddress", (void*)cuGetProcAddress_v2}, {"cuGetProcAddress_v2", (void*)cuGetProcAddress_v2}}};
+
```
neobrain, line 61: This fails to compile on my system (clang 21); is this not covered by CI?

```
libcuda_Guest.cpp:59:48: error: constexpr variable 'proc_override' must be initialized by a constant expression
   59 | constexpr static std::array<override_entry, 2> proc_override = {
      |                                                ^               ~
   60 |   {{"cuGetProcAddress", (void*)cuGetProcAddress_v2}, {"cuGetProcAddress_v2", (void*)cuGetProcAddress_v2}}};
      |   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
libcuda_Guest.cpp:60:25: note: cast that performs the conversions of a reinterpret_cast is not allowed in a constant expression
   60 |   {{"cuGetProcAddress", (void*)cuGetProcAddress_v2}, {"cuGetProcAddress_v2", (void*)cuGetProcAddress_v2}}};
```

Sonicadvance1: This is covered by CI.

PR #5546, opened 2026-06-03
CodeCache: Support targeting WOW64/ARM64EC in FEXOfflineCompiler

neobrain:
Alternative to #5502 (largely reusing Billy's implementation).

This makes the existing FEXOfflineCompiler usable as a PE executable that will generate FEX code caches for WOW64/ARM64EC binaries. In combination with other upcoming patches, this has been end-to-end-tested and verified to produce functional caches that pass validation when loaded at runtime.

Using a single executable makes it easier to stick to consistent conventions across our target platforms, particularly with PE support for x86 Wine running under FEX eventually being added here too. We can still diverge where needed but will be more explicit about it.


Sonicadvance1:
Needs a rebase.

Sonicadvance1 (approved):

PR #5547, opened 2026-06-03
FEXOfflineCompiler: Add "process-all" verb

neobrain:
Adds a new command: `FEXOfflineCompiler process-all` (no arguments).

This operation will take care of any pending code cache operations:
* import new code maps from `CACHE_DIR/codemap/new` and process them to `CACHE_DIR/codemap/ready`
* generate caches for updated code maps with new blocks (detected via file timestamps)
* ensure caches already exist for all other code maps (and generate them if needed)

One notable detail is that libraries (DLL) are processed in the *context* of an executable (EXE), since the latter is what FEXOfflineCompiler will eventually determine the FEXCore configuration from. The proposed patch will already correctly handle if the same library is used by multiple executables (by compiling them only once if FEXCore config is identical, or compiling them twice if not).

In the future we can merge some of the code map processing logic back into FEXServer, but for now I'm keeping it separate until the dust settles. In principle it's very similar though, except that it operates on all code maps wheres FEXServer operates on an individual executable. 

Note that this PR implements cache updating logic *and* code map merging, whereas #5502 lacks the latter.


bylaws:
code map merging, whereas https://github.com/FEX-Emu/FEX/pull/5502 lacks the latter... thats because the compile frontend does it. which is a separate binary and unrelated pr

Sonicadvance1 (approved):

Issue #3799, opened 2024-07-01
SVE256: AVX implementation fails to retain upper 128-bits of results


Sonicadvance1:
We don't have any unit tests that aggressively mix SSE and AVX unit tests. In some (most?) cases the SVE256 bit implementation of our AVX emulation doesn't correctly retain the upper 128-bits of the register when an SSE operation occurs.

This can be seen in the following unit test:
```asm
%ifdef CONFIG
{
  "RegData": {
    "XMM0": ["0xfdecd28fab3fa4a5", "0x7d7ccd8836d09fc2", "0xccdbcfc31f3ff0f3", "0x108390defebac4be"],
    "XMM1": ["0xc4be43cc1ad6970b", "0x4559bd7eb46a1278", "0", "0"],
    "XMM2": ["0xc4be43cc1ad6970b", "0x4549bd7eb46a1278", "0xf793ef673dac6e4c", "0xbb7b5b3d85d34271"],
    "XMM3": ["0x000043cc1ad6970b", "0x4549bd7eb46a1278", "0xf793ef673dac6e4c", "0xbb7b5b3d85d34271"]
  }
}
%endif

vmovaps ymm0, [rel .data]
vmovaps ymm1, [rel .data + (1 * 32)]
vmovaps ymm2, [rel .data + (2 * 32)]
vmovaps ymm3, [rel .data + (3 * 32)]

addpd xmm0, xmm1
vaddpd xmm1, xmm2, xmm3

hlt

align 32
.data:
dq 0xfdecd28fab3fa4a5, 0x7d7ccd8836d09fc2, 0xccdbcfc31f3ff0f3, 0x108390defebac4be
dq 0x43cc1ad6970b4549, 0xbd7eb46a1278f793, 0xef673dac6e4cbb7b, 0x5b3d85d342718be9
dq 0xc4be43cc1ad6970b, 0x4549bd7eb46a1278, 0xf793ef673dac6e4c, 0xbb7b5b3d85d34271
dq 0x000043cc1ad6970b, 0x4549bd7eb46a1278, 0xf793ef673dac6e4c, 0xbb7b5b3d85d34271
```

This unit test works with ASIMD and SVE128, but fails with SVE256. Looking at an instcountci test shows why this occurs

First with ASIMD/SVE128:
```json
    "addpd xmm0, xmm1": {
      "ExpectedInstructionCount": 1,
      "Comment": [
        "Map 1 0b00 0x10 128-bit"
      ],
[...]

PR #5557, opened 2026-06-17
[SVE256] Handle SSE insertions for ROUNDPD/ROUNDPS

lioncash:
See #3799


unittests/ASM/Disabled_Tests_Simulator
```diff
@@ -85,6 +85,8 @@ Test_H0F3A/66_09.asm
 Test_H0F3A/66_0A.asm
 Test_H0F3A/66_0B.asm
 Test_OpSize/66_5B.asm
+Test_SSELanePreservation/roundpd.asm
```
Sonicadvance1: I assume these were tested and the only failure behaviour was the rounding rather than the truncation?

lioncash: This is actually a holdover from during bringup that I forgot to remove (whoops)

Sonicadvance1 (approved):

PR #5565, opened 2026-06-17
[SVE256] Handle SSE insertions for XOR special case

lioncash:
See #3799

Ensures that our special case maintains insertion behavior


Sonicadvance1 (approved):
Nice catch.

PR #5571, opened 2026-06-19
Fix incorrect RSP update for 16bit leave

simon902:
When executing the `leave` instruction with an `o16` prefix, the operand size is correctly reduced to 16 bits. However, the previous implementation erroneously used this 16-bit size to also update the `RSP`. This caused only the lower 16 bits of `RSP` to be updated.
However, instead the native stack address size (`GPRSize`) should be used for the update.


Sonicadvance1 (approved):
Looks reasonable enough to me. Thanks!

PR #5572, opened 2026-06-20
[SVE256] Handle SSE insertions for PCMPESTRM/PCMPISTRM ops

lioncash:
See #3799

### DIFF
--- a/unittests/ASM/SSELanePreservation/pcmpestrm.asm
+++ b/unittests/ASM/SSELanePreservation/pcmpestrm.asm
@@ -1,0 +1,37 @@
+%ifdef CONFIG
+{
+  "HostFeatures": ["AVX"],
+  "RegData": {
+    "XMM0": ["0x00000000000060a0", "0x0000000000000000", "0x0000000000000000", "0x0000000000000000"],
+    "XMM1": ["0x00000000000060a0", "0x0000000000000000", "0xFFFFFFFFFFFFFFFF", "0xEEEEEEEEEEEEEEEE"],
+    "XMM2": ["0x6463626144434241", "0x6C6B6A694C4B4A49", "0xFFFFFFFFFFFFFFFF", "0xEEEEEEEEEEEEEEEE"],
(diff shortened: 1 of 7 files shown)


unittests/ASM/SSELanePreservation/pcmpestrm.asm
```diff
@@ -0,0 +1,34 @@
...
+}
+%endif
+
+vmovaps ymm0, [rel .data]
+vmovaps ymm1, [rel .data + (1 * 32)]
+vmovaps ymm2, [rel .data]
+vmovaps ymm3, [rel .data + (1 * 32)]
+
+pcmpistrm xmm2, xmm3, 0b00000000
+vmovaps ymm1, ymm0
```
Sonicadvance1, line 22: Would you want to reload ymm0 after moving the resulting modified ymm0 in to ymm1, to ensure the following `vpcmpistrm` is using the same data as the first `pcmpistrm`? (Also this test is incorrectly using pcmpistrm when it is supposed to be testing for pcmpestrm per the test name)

Then the same for the `pcmpistrm` test.

Sonicadvance1 (commented):
Missing equivalent `pcmpestri` and `pcmpistri` tests?

unittests/ASM/SSELanePreservation/pcmpestrm.asm
lioncash: Not really, since xmm0 is only used as a destination and isn't read from. We just shunt the value from the first pcmpestrm into ymm1 afterwords, which is used to ensure the insertion occurred. The following vpcmpestrm ensures that the upper lanes then get overwritten in ymm0. Did fix the usage of pcmpistrm though

lioncash:
pcmpistri and pcmpestri don't modify any of the xmm registers, only ecx/rcx

unittests/ASM/SSELanePreservation/pcmpestrm.asm
Sonicadvance1: Sounds good

Sonicadvance1 (approved):

Sonicadvance1:
> pcmpistri and pcmpestri don't modify any of the xmm registers, only ecx/rcx

Right. Forgot about that.

PR #5577, opened 2026-06-21
Context: Add support for single-step RIP ranges

Sonicadvance1:
Useful when debugging a range.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -247,6 +247,7 @@ public:
   struct TrackingEmpty {
     // RIP stepping handling
     virtual void AddSingleStepTarget(uint64_t GuestRIP) {}
+    virtual void AddSingleStepTargetRange(uint64_t RIPBegin, uint64_t RipEnd) {}
     virtual void AllTargetSingleStep() {}
     virtual void RemoveSingleStepTarget(uint64_t GuestRIP) {}
     virtual bool IsSingleStepTarget(uint64_t GuestRIP) {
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Context/Context.h
```diff
@@ -302,6 +307,20 @@ class ContextImpl final : public FEXCore::Context::Context, public CPU::CodeBuff
     fextl::set<uint64_t> SingleStepTargets {};
     fextl::set<uint64_t> WatchWriteTargets {};
     fextl::set<uint64_t> WatchReadTargets {};
+    struct Range {
+      uint64_t Begin, End;
+    };
+    fextl::vector<Range> SingleStepRanges {};
+
+    bool IsInRange(uint64_t RIP) {
+      for (const auto& range : SingleStepRanges) {
```
lioncash: small nit, but could be collapsed into:

```cpp
return std::ranges::any_of(SingleStepRanges, [RIP](const auto& range) {
    return RIP >= range.Begin && RIP <= range.End;
});
```

and the function could also be made const

Sonicadvance1: Sure thing, switched over to that.

Sonicadvance1:
Apparently old libstdc++ can't do construct_at semantics fully.
```diff
-      SingleStepRanges.emplace_back(RIPBegin, RIPEnd);
+      SingleStepRanges.emplace_back(Range {RIPBegin, RIPEnd});
```
Minor change to work around that.

lioncash (commented):
One small nit, but looks good otherwise

FEXCore/Source/Interface/Context/Context.h
```diff
@@ -302,6 +307,14 @@ class ContextImpl final : public FEXCore::Context::Context, public CPU::CodeBuff
     fextl::set<uint64_t> SingleStepTargets {};
     fextl::set<uint64_t> WatchWriteTargets {};
     fextl::set<uint64_t> WatchReadTargets {};
+    struct Range {
+      uint64_t Begin, End;
+    };
+    fextl::vector<Range> SingleStepRanges {};
+
+    bool IsInRange(uint64_t RIP) {
```
lioncash:
```suggestion
    bool IsInRange(uint64_t RIP) const {
```

Sonicadvance1: nit resolved.

lioncash (approved):

PR #5584, opened 2026-06-22
[SVE256] More comprehensively test SSE insertions for scalar comparisons

lioncash:
See #3799

Drops in the facilities to ensure all of the available SSE scalar comparison paths are tested for proper insertion behavior.


Sonicadvance1 (approved):
Woo data

PR #5585, opened 2026-06-22
[SVE256] Remove heavy handed moves from scalar compares

lioncash:
(See #3799)

I had a feeling #5569 was a little overkill, but was just getting everything up to a functional baseline at the time. Now, with the tests added in #5584 to test all SSE paths, I was able to see which comparisons in particular were the ones that would have deviating behavior (NLT and NLE)

This lets us safely restore the behavior without the excessive moves on hardware that makes use of FEAT_AFP.


Sonicadvance1 (approved):
Oh fancy.

PR #5587, opened 2026-06-23
[SVE256] Add fast paths for trivial VSHUFPD flags (0b0000, and 0b1111)

lioncash:
Lets us at least flatten down two paths from 20 instructions to 1.


Sonicadvance1 (approved):
Nice.

PR #5590, opened 2026-06-23
AVX: Reduce inserts in VBLENDPS/VPBLENDD/VPBLENDW

lioncash:
Lets us reduce inserts by seeing which bits in the selector mask indicates a particular source is used more than the other one, and then just uses that as the base to be inserted into, cutting down on overall insertion overhead.

In some cases, this can be quite drastic, like with:

```asm
vpblendw ymm0, ymm1, ymm2, 0b00000001
```

being cut down from 98 instructions to 14.


Sonicadvance1 (approved):
Fancy Fancy.

PR #5592, opened 2026-06-23
JIT: Arm64: fix the loop in CacheLineClear/Clean

ShadowCurse:
These functions need to clean at least 64 bytes of cache since this is the default on x86_64, but previously they could clean less if DCacheLineSize was smaller than 64 bytes.


Sonicadvance1:
Turns out the original code is actually incorrect, so this change is even more incorrect.
The loops should still exist, just invert the check `64U / CTX->HostFeatures.DCacheLineSize`

ShadowCurse:
yeah you are right, my bad, fixed now

Sonicadvance1 (approved):
Alright cool, this is now a proper fix. Thanks!

PR #5594, opened 2026-06-24
AVX: Wire up lane helper for VPSHUFD/VPSHUFLW/VPSHUFHW

lioncash:
Lets the AVX implementation get all the optimizations that the SSE variant has, reducing the overhead a little. Even with the individual lane handling, this is still leagues better than all of the individual inserts that are pretty beefy with SVE.

For example:

```asm
vpshufd ymm0, ymm1, 0b00000011
```

drops from 50 instructions to 9.

Still some more work we can do to drop those down further, but this should be good while that's done.


Sonicadvance1 (approved):
Ah, I see.

PR #5596, opened 2026-06-24
AVX: Wire up lane helper for VPERMILPS imm variant

lioncash:
Makes for some more trivial savings. Will need handling for VPERMILPD added separately, since selector behavior is different.


Sonicadvance1 (approved):
nice nice

PR #5597, opened 2026-06-24
OpcodeDispatcher: Fix incorrect comment for BTOp. ZF must be preserved.

simon902:
Incorrect comment for BTOp, the ZF flag is not undefined but must be preserved.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -2427,9 +2427,10 @@ void OpDispatchBuilder::BTOp(OpcodeArgs, uint32_t SrcIndex, BTAction Action) {
     auto BitSelect = (Size == (LshrSize * 8)) ? Src : Src.And(Mask);
     auto LshrOpSize = IR::SizeToOpSize(LshrSize);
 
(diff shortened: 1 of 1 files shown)


Sonicadvance1:
By whose specification? According to AMD ZF is undefined, only CF is defined.
Screenshot 2026-06-24 at 08 02 33

Sonicadvance1:
Apparently Intel specification states that ZF must be preserved? Sounds like we need to have the comment state the difference between vendors.
Screenshot 2026-06-24 at 08 04 12

simon902:
Interesting. I didnt know there were such small details where AMD and Intel diverge.

Regarding changing the comment: In the same function, there is already another comment which also mentions that ZF must be preserved:

```
// OF/SF/AF/PF undefined. ZF must be preserved. We choose to preserve OF/SF
// too since we just use an rmif to insert into CF directly. We could
// optimize perhaps.
```

Sonicadvance1:
Could just put in the comment.
```cpp
// AMD: OF/SF/ZF/AF/PF undefined.
// Intel: OF/SF/AF/PF undefined. ZF must be preserved.
```

simon902:
I have updated both comments to reflect the vendor specific behaviour.

Sonicadvance1 (approved):
Looks reasonable enough to me. Thanks for the comment.

PR #5598, opened 2026-06-24
AVX: Wire up helper to VPERMILPD

lioncash:
We can just leverage the shuffle handler for this, since VPSHUFD essentially functions like VPERMILPD


Sonicadvance1 (approved):
Nice nice.

PR #5605, opened 2026-06-25
AVX: Skip identity insertions in VPERMQ

lioncash:
In the slower case, if our iteration index and the selector index match, then all that means is that we'd be inserting the same data that already exists at that location, so we can skip the insertion in that case.


Sonicadvance1 (approved):
Nice.

PR #5608, opened 2026-06-26
OpcodeDispatcher: Fix CRC32 operand size

simon902:
Update `OpDispatchBuilder::CRC32` to use `OpSizeFromSrc(Op)` for loading the source operand instead of the `GPRSize`.
Previosuly, using `GPRSize` for the source operand caused the assert `(Width + lsb) <= IR::OpSizeAsBits(Size)` in `_Bfe` when `crc32` was used with an 8-bit high register.


Sonicadvance1:
I don't think this is the proper way to resolve this, it's going to cause unnecessary zero extends and extractions.

Sonicadvance1:
Thanks for the unit test and bringing up the problem. It's now resolved.

PR #5610, opened 2026-06-26
FEXCore: Pass host type that changes codegen to FEXCore

Sonicadvance1:
Because these compile options change codegen, we need to make sure these are runtime selected rather than compile-time selected. Will reduce code-cache variance.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -4778,35 +4778,36 @@ void OpDispatchBuilder::INTOp(OpcodeArgs) {
   case 0xCD: { // INT imm8
     uint8_t Literal = Op->Src[0].Literal();
 
-#ifndef _WIN32
-    constexpr uint8_t SYSCALL_LITERAL = 0x80;
-    if (Literal == SYSCALL_LITERAL) {
-      if (Is64BitMode) [[unlikely]] {
-        LogMan::Msg::EFmt("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");
-        UnhandledOp(Op);
+    if (CTX->HostFeatures.HostType == FEXCore::HostFeatures::HostTypeEnum::Linux) {
+      constexpr uint8_t SYSCALL_LITERAL = 0x80;
+      if (Literal == SYSCALL_LITERAL) {
+        if (Is64BitMode) [[unlikely]] {
+          LogMan::Msg::EFmt("[Unsupported] Trying to execute 32-bit syscall from a 64-bit process.");
+          UnhandledOp(Op);
+          return;
+        }
+        // Syscall on linux
+        SyscallOp(Op, false);
         return;
       }
-      // Syscall on linux
-      SyscallOp(Op, false);
-      return;
-    }
-#else
-    constexpr uint8_t SYSCALL_LITERAL = 0x2E;
-    if (Literal == SYSCALL_LITERAL) {
-      // Can be used for both 64-bit and 32-bit syscalls on windows
-      SyscallOp(Op, false);
-      return;
-    }
-#endif
+    } else if (CTX->HostFeatures.HostType == FEXCore::HostFeatures::HostTypeEnum::Wow64 ||
+               CTX->HostFeatures.HostType == FEXCore::HostFeatures::HostTypeEnum::Arm64ec) {
+      constexpr uint8_t SYSCALL_LITERAL = 0x2E;
+      if (Literal == SYSCALL_LITERAL) {
+        // Can be used for both 64-bit and 32-bit syscalls on windows
+        SyscallOp(Op, false);
+        return;
+      }
 
-#ifdef ARCHITECTURE_arm64ec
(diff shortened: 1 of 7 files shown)


lioncash (approved):

neobrain:
Does this require corresponding changes in FEXOfflineCompiler on WoA and have you ensured code cache validation is still clean after this change?

Sonicadvance1:
It will need one yes. Nothing changed there, it'll continue generating broken code in this instance as expected.

neobrain:
> It will need one yes. Nothing changed there, it'll continue generating broken code in this instance as expected.

How has it been generating broken code and why are you bringing this up just now?

Anyway, are you going to submit a followup PR for it?

Sonicadvance1:
> > It will need one yes. Nothing changed there, it'll continue generating broken code in this instance as expected.
> 
> How has it been generating broken code and why are you bringing this up just now?
> 
> Anyway, are you going to submit a followup PR for it?

There were basically three compile time code paths
- If Linux: Generate 32-bit syscall
- If Win32: Generate generic syscall
- if arm64ec: Generate code for non-scaled cyclecounter

Depending on which OfflineCodeCompiler build flags and target platform, the tool would never be unable to generate the correct syscall code for all targets. Or the other way of thinking about it, the tool would generate code for a different platforms needs.

These are fairly uncommon instructions so it's likely that it wasn't encountered before. There needs to be a flag passed to either the OfflineCodeCompiler or a flag saved in to the code map to determine what the original target was. I assume codemap is going to be the primary path, considering on win32 we have no way to know if a process is wow64 or arm64ec after the fact.

neobrain:
> considering on win32 we have no way to know if a process is wow64 or arm64ec after the fact.

Am I missing something? A 32-bit binary is wow64 and a 64-bit one is arm64ec, isn't it? So the information is right there. Also note that trying to compile for ARM64EC with FEXOfflineCompiler32 isn't intended to happen, just that it isn't enforced (yet).

Either way this seems unrelated to the PR though. Why didn't you update FEXOfflineCompiler along the other changes? Relying on me to catch this stuff when it was submitted + merged over a weekend isn't exactly good team play.

Sonicadvance1:
> > considering on win32 we have no way to know if a process is wow64 or arm64ec after the fact.
> 
> Am I missing something? A 32-bit binary is wow64 and a 64-bit one is arm64ec, isn't it? So the information is right there. Also note that trying to compile for ARM64EC with FEXOfflineCompiler32 isn't intended to happen, just that it isn't enforced (yet).
> 
> Either way this seems unrelated to the PR though. Why didn't you update FEXOfflineCompiler along the other changes? Relying on me to catch this stuff when it was submitted + merged over a weekend isn't exactly good team play.

Glad to see that you concur then that the original code was broken and the new code is equally broken in that environment. The OfflineCompiler needs to be patched up to fetch host CPU features and set the host type the same as the original Linux/wow64/arm64ec code paths if it wants to produce identical output.

Sonicadvance1:
> > > considering on win32 we have no way to know if a process is wow64 or arm64ec after the fact.
> > 
> > 
> > Am I missing something? A 32-bit binary is wow64 and a 64-bit one is arm64ec, isn't it? So the information is right there. Also note that trying to compile for ARM64EC with FEXOfflineCompiler32 isn't intended to happen, just that it isn't enforced (yet).
> > Either way this seems unrelated to the PR though. Why didn't you update FEXOfflineCompiler along the other changes? Relying on me to catch this stuff when it was submitted + merged over a weekend isn't exactly good team play.
> 
> Glad to see that you concur then that the original code was broken and the new code is equally broken in that environment. The OfflineCompiler needs to be patched up to fetch host CPU features and set the host type the same as the original Linux/wow64/arm64ec code paths if it wants to produce identical output.

Looks like I might have misread this but I'll make it a bit more clear. **Before this change**, depending on build flags of FEXOfflineCompiler (Linux, wow64 Windows, arm64ec Windows), the syscall instruction was emitting different code incorrectly depending on case on the Windows side. Primarily boiling down to the behaviour of `int  0x81` changing depending on if the build flag contained **ARCHITECTURE_arm64ec** or not. Because **FEXOfflineCompiler.exe** compiles with either one depending on if the target is libwow64fex.dll or libarm64ecfex.dll.

How regular wow64/arm64.dll code emits:

| constant | wow64 | arm64ec |
| - | - | - |
| 0x81 | SIGSEGV | CNTVCT |

What FEXOfflineCompiler emits, depending on build flag:
| flag | wow64 | arm64ec |
| - | - | - |
| ARCHITECTURE_arm64ec | CNTVCT | CNTVCT |
| ARCHITECTURE_arm64 (wow) | SIGSEGV | SIGSEGV |

This is why the flag must be passed in from the frontend, because it was already broken. I think because the frontend isn't currently passing the operating mode to the backend, there is **now** just some additional breakage that **FEXOfflineCompiler.exe** is currently claiming it is HostType of Linux because it's using the wrong `FetchHostFeatures`.

The wrong HostFeatures just means that some features are also incorrectly enabled or disabled which changes codegen as well. But I didn't change that in this particular PR, as it should be changed independently.

neobrain:
For closure, the additional breakage that I pointed out was fixed in #5635.

PR #5611, opened 2026-06-26
Windows: Adds empty Linux side unix library

Sonicadvance1:
We are going to need a unix library. Going to take this one step at a time without AI/ML so I fully understand all the pieces of the puzzle, and to ensure we don't lose any functionality before we're ready.

This only ensures that we are building the Linux facing .so files for arm64ec and wow64, but they are empty today. Next PR will be initializing it on the PE side.

### DIFF
--- a/Source/Windows/UnixLib/FEXUnixLib.h
+++ b/Source/Windows/UnixLib/FEXUnixLib.h
@@ -1,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+
+enum class FEXUnixLibFunctions {
+  // Empty for now.
+  MAX,
+};
--- a/.github/workflows/wine_build/action.yml
+++ b/.github/workflows/wine_build/action.yml
@@ -33,3 +33,17 @@ runs:
     - name: Install
       shell: bash
       run: DESTDIR="$PWD"/install cmake --build build_${{ inputs.target }} -t install
+
+    - name: Configure UnixLib
+      shell: bash
+      run: |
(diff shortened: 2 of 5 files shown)


Source/Windows/UnixLib/FEXUnixLib.h
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+
+enum FEXUnixLibFunctions {
+  // Empty for now.
+  MAX,
```
lioncash, line 6: Perhaps this should be made an enum class, or just classified with a longer name like `UNIX_LIB_FUNC_MAX`, just so it isn't putting a general name like `MAX` in the global namespace.

lioncash (commented):
One minor nit, but LGTM overall

Source/Windows/UnixLib/FEXUnixLib.h
Sonicadvance1: Switched over to an enum class. Had to add in FEXCore/Util's ToUnderlying which will have some knock-on effects to the following PRs, but nothing terrible.

lioncash (approved):

neobrain:
> Going to take this one step at a time without AI/ML so I fully understand all the pieces of the puzzle

Why submit 4 tiny pull requests for a single logical change? If you need to take steps at a time, that's what git commits allow you to do, and needlessly splitting out context into multiple PRs hardly makes it any easier to review than LLM-generated code...

PR #5612, opened 2026-06-26
Windows: Load unixlib if possible 

Sonicadvance1:
Currently does nothing other than load it (as the library also doesn't
do anything yet). Ensured it was working by temporarily creating a test
entrypoint and doing `Call` on to it.

Next step after this is to reimplement some of the nasty hacks FEX is
doing inside the unixlib code itself.


Sonicadvance1:
Rebased.

lioncash (approved):

PR #5613, opened 2026-06-26
Windows/UnixLib: Adds support for Hardware TSO support

Sonicadvance1:
Including fallback to non-unixlib path because we need to support both.

Showcases how these are going to be implemented without throwing the
entire world at it right away. Next PR will be implementing the
remaining four necessary unixlib handlers that we will require:

- Kernel unaligned atomic control
- shm_stats thing
- madvise operation
- prctl vma naming


Sonicadvance1:
Rebased.

lioncash (approved):

PR #5617, opened 2026-06-28
OpcodeDispatcher: Fix missing zeroing for vcvtps2ph

simon902:
The vcvtps2ph instruction should always zero the upper 128 bits of the destination if it is a  **register**. See the Intel Manual:

```
VEX.128 version: The source operand is a XMM register. The destination operand is a XMM register or 64-bit memory location. If the destination operand is a register then the upper bits (MAXVL-1:64) of corresponding register are zeroed.

VEX.256 version: The source operand is a YMM register. The destination operand is a XMM register or 128-bit memory location. If the destination operand is a register, the upper bits (MAXVL-1:128) of the corresponding destination register are zeroed.
```

This PR adds a fix and an ASM unit test.
I also ran the `instcountci_update_tests` target which updated the `unittests/InstructionCountCI/AVX128/VEX_map3.json` file, however, there is also `unittests/InstructionCountCI/VEX_map3.json` which did not get updated by this. Is this legacy code or should it also be updated?

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/AVX_128.cpp
@@ -2293,9 +2293,8 @@ void OpDispatchBuilder::AVX128_VCVTPS2PH(OpcodeArgs) {
     _PopRoundingMode(OldFPCR);
   }
 
-  // We need to eliminate upper junk if we're storing into a register with
-  // a 256-bit source (VCVTPS2PH's destination for registers is an XMM).
-  if (Op->Src[0].IsGPR() && SrcSize == OpSize::i256Bit) {
+  // We need to zero the upper 128 bits if we're storing into a register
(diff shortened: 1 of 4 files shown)


lioncash:
> Is this legacy code or should it also be updated?

That is the 256-bit SVE instcountci file, which already handles the zeroing behavior properly. It doesn't need to be updated.

Sonicadvance1:
The test need to be added to `unittests/ASM/Disabled_Tests_Simulator` because Vixl simulator doesn't support instructions that touch the rounding mode.

Sonicadvance1:
For updating the instcount ci tests, it's easiest to just run `ninja && ninja instcountci_tests ; ninja instcountci_update_tests`

simon902:
Done. Added the test to `Disabled_Tests_Simulator` and updated InstcountCI.

Sonicadvance1 (approved):
Alright, looking good to me!

PR #5618, opened 2026-06-28
Core: Start a new block for the next op in full SMC check

wsxarcher:
Under FEX_SMCCHECKS=full, FEX splits each x86 instruction into its own block but forgot to clear its per-block vector-constant cache at those splits. So back-to-back pmovmskb instructions shared one constant across blocks, and the block-local register allocator then reused the same register for the constant and its operand, producing VAnd v1, v1. That made pmovmskb return garbage, breaking glibc strlen (when hwcaps x86-64-v3), which corrupted loader paths and made Vivado fail. The fix clears all creating a new block.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -650,6 +650,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
           Thread->OpDispatcher->SetTrueJumpTarget(InvalidateCodeCond, CodeWasChangedBlock);
 
           Thread->OpDispatcher->SetCurrentCodeBlock(CodeWasChangedBlock);
+          Thread->OpDispatcher->StartNewBlock();
           Thread->OpDispatcher->_ThreadRemoveCodeEntry();
           Thread->OpDispatcher->ExitFunction(Thread->OpDispatcher->_InlineEntrypointOffset(GPRSize, InstAddress - GuestRIP));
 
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -657,6 +657,7 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
           Thread->OpDispatcher->SetFalseJumpTarget(InvalidateCodeCond, NextOpBlock);
           Thread->OpDispatcher->SetCurrentCodeBlock(NextOpBlock);
+          Thread->OpDispatcher->ClearCachedNamedConstants();
```
Sonicadvance1: This isn't fully correct. After `SetCurrentCodeBlock` we should be calling `StartNewBlock` instead of `ClearCachedNamedConstants`.

wsxarcher: Confirmed it works, changing the PR

FEXCore/Source/Interface/Core/Core.cpp
Sonicadvance1, line 652: I think there needs to be another `Thread->OpDispatcher->StartNewBlock();` after this `SetCurrentCodeBlock` as well to be completely correct.

wsxarcher: Added and verified!

Sonicadvance1:
Make sure to squash these commits down, we don't do merge squashing.

Sonicadvance1 (approved):
Alright looks reasonable now. This is definitely a less tested code path. Thanks!

PR #5619, opened 2026-06-28
AVX: Handle trivial cases better for VDPPS

lioncash:
Just a few trivial cases where we can have much better output.

There's more we can do here, but this just gets the simple ones out of the way first.


Sonicadvance1 (approved):
`"ExpectedInstructionCount": 1,` lol

PR #5620, opened 2026-06-28
Fix UAF of PS1

wsxarcher:
PS1 std::string was getting deallocated before the usage as c_str in execve.
The bug was not really triggered as there was no allocation after that but the memory was in fact deallocated.


Sonicadvance1 (approved):
This one seems fine.

PR #5621, opened 2026-06-28
Do not exec FEX if it is a folder in FEXBash

wsxarcher:
FEXBash crashes when there is a FEX named folder (for example the repo checkout) in the cwd because it tries to execve it. Checks only regular files and not folders.

### DIFF
--- a/Source/Tools/FEXBash/FEXBash.cpp
+++ b/Source/Tools/FEXBash/FEXBash.cpp
@@ -53,14 +53,14 @@ int main(int argc, char** argv, char** const envp) {
 
   // Check if a local FEX to FEXBash exists
   // If it does then it takes priority over the installed one
-  if (!std::filesystem::exists(FEXPath)) {
+  if (!std::filesystem::is_regular_file(FEXPath)) {
     std::error_code ec;
     auto FEXBashPath = std::filesystem::read_symlink("/proc/self/exe", ec);
     if (!ec) {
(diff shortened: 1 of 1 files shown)


Source/Tools/FEXBash/FEXBash.cpp
```diff
@@ -53,14 +53,14 @@ int main(int argc, char** argv, char** const envp) {
 
   // Check if a local FEX to FEXBash exists
   // If it does then it takes priority over the installed one
-  if (!std::filesystem::exists(FEXPath)) {
+  if (!std::filesystem::is_regular_file(FEXPath)) {
```
Sonicadvance1, line 56: Do these throw an assert if the file doesn't exist? Might need the version that returns `system_error` so a throw doesn't occur?

wsxarcher: Yes it throws but also exists throwed so I kept the same

wsxarcher: Sorry i meant it throws only when there is a permission error like exists

wsxarcher: When the file does not exists it just return false

Sonicadvance1: Okay, good to know that we aren't introducing any new asserts

Sonicadvance1 (approved):
Seems fine to me.

PR #5622, opened 2026-06-28
AVX: Handle trivial UZP/ZIP operations in VPERMQ

lioncash:
Handles cases where a permutation can be simplified into a single zip/unzip operation.


Sonicadvance1 (approved):
Nice

PR #5623, opened 2026-06-28
Windows/UnixLib: Adds remaining helpers

Sonicadvance1:
Centralizes all the nasty behaviour that will end up breaking when WINE
eventually turns on userspace syscall dispatch. Pushes all of the logic
in to the UnixLib. Support both paths until everyone is migrated to
supporting the UnixLib, then we can delete the bit of code duplication
between the PE side and UnixLib side.

Helpful that everything that gets punched through the UnixLib is
optional, so worst case some optional bits can break for a while.


Sonicadvance1:
Rebased.

lioncash (approved):

neobrain:
@lioncash Could we stick to the policy of not merging PRs within less than 24 working day hours? It's fine for non-trivial changes, but a +389/-293 is far from that, especially if it's merged alongside 14 other pull requests over the same weekend.

PR #5624, opened 2026-06-28
AVX: Slightly trim codegen for VPSADW 256-bit case

lioncash:
We can massage this a little bit to be slightly better. At least gets rid of the heavyweight inserts.


Sonicadvance1 (approved):
wee

PR #5635, opened 2026-06-30
FEXOfflineCompiler: Fixes HostFeature detection under Win32

Sonicadvance1:
CPUFeature detection is marginally different between Linux and Windows. FOC was only using the Linux path which had two broken things happening to it.
- Feature detection was incorrect and enabling/disabling features differently from wow64/arm64ec .dll files
- HostType was being set as Linux even though it was generating code for WINE

Ensure that when built for Win32 that it uses the correct feature fetching.

One thing that is still incorrect is that 64-bit or 32-bit is determined at compile time on win32, whereas the Linux side parses an ELF and determines bitness at runtime. This doesn't fix that remaining problem there.


neobrain (approved):
Thanks!

> One thing that is still incorrect is that 64-bit or 32-bit is determined at compile time on win32, whereas the Linux side parses an ELF and determines bitness at runtime. This doesn't fix that remaining problem there.

Yes, I have had a patch locally that addresses this.

PR #5636, opened 2026-06-30
AVX: Remove unnecessary moves from VPSHUF{D, HW, LW}

lioncash:
These aren't necessary anymore, since these use operations that already zero extend.


Sonicadvance1 (approved):
Nice nice.

Sonicadvance1:
Well, maybe not so nice that CI fails. But in theory :smile:

lioncash:
There we go, just needed to handle the identity case

Sonicadvance1 (approved):
ezpz

PR #5637, opened 2026-06-30
Windows/UnixLib: Fix loading with new MemoryWineLoadUnixLibByName mechanism

mrpippy:
`NtQueryVirtualMemory(MemoryWineLoadUnixLibByName)` takes the name as a `UNICODE_STRING *`, not `char *`.

If desired, an implementation of `RTL_CONSTANT_STRING()` or similar would allow the `UNICODE_STRING` to be initialized at compile-time.


Sonicadvance1 (approved):
Looks like this probably works. Thanks!

PR #5638, opened 2026-06-30
IR: Add constant for swapping midsections of 256-bit vectors around 

lioncash:
Lets us trim some more things down until we can make use of the Q operations (ZIP1Q, etc) introduced in SVE2.1.

Inspect the second commit specifically to see the actual codegen changes, since the main view is more cluttered with the constant indices changing.


Sonicadvance1 (approved):
Nice

PR #5641, opened 2026-07-01
VectorOps: Reduce temporary usage in 64-bit AdvSIMD min max paths

lioncash:
We can reorganize these such that they only use one temporary in the worst case instead of two.


Sonicadvance1:
I guess this just isn't picked up in instcountci because there's no 64-bit min/max until AVX512?

lioncash:
Yeah. I just happened to notice the paths could have better output while I was looking at something else

Sonicadvance1 (approved):

PR #5648, opened 2026-07-04
[SVE256] Ensure SSE insertion behavior for AES/SHA/PCLMUL operations

lioncash:
Forgot that x86 can do this whole crypto algorithm thing. Drops in tests to make sure these behave as expected as well.

Another one to tag onto #3799


Sonicadvance1 (approved):
Wacky that SHA doesn't have an AVX variant, but I'm not complaining.

PR #5650, opened 2026-07-04
[SVE256] Handle 256-bit AES operations

lioncash:
Currently we split these into two 128-bit operations since VIXL doesn't have support for the unified SVE operations yet.

Now we fully support VAES on SVE256. Better to have it working slightly below average performance-wise than not at all.


Sonicadvance1 (approved):

Sonicadvance1:
Really looks like we need to implement a bunch of missing operations in vixl.

PR #5651, opened 2026-07-04
unittests: Add stress tests for VPERMIL{PD, PS}

lioncash:
While unlikely to be used in practice over other kind of shuffling and blending, these should also have stress tests to make sure they do the right thing.


Sonicadvance1:
Oh snap, found a bug in the avx128 implementation?

Sonicadvance1 (approved):

PR #5654, opened 2026-07-05
Allocator: fix the caching of host va size

ShadowCurse:
`HostVASize` global was never set and so each call to `DetermineVASize` would redo all the work. Set the global to fix this.
Additionally do some renaming to specify that the returned value is a number of bits and not the size of the VA.


Sonicadvance1 (approved):
Oops. Thanks for catching that. Luckily it only gets called a handful of times.

PR #5660, opened 2026-07-06
Wow64: Spin loop on atomic with WFE

Sonicadvance1:
Instead of burning roughly a million watts, put this spinloop on a WFE. This tends to occur on a crash during shutdown that isn't fully able to be avoided. The least we can do is not consume all the power in the world.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/SpinWaitLock.h
+++ b/FEXCore/include/FEXCore/Utils/SpinWaitLock.h
@@ -144,6 +144,21 @@ static inline void WaitPred(T* Futex, T ComparisonValue) {
   }
 }
 
+template<typename T, typename Pred>
+static inline void WaitBitMaskPred(T* Futex, T BitMask, T ComparisonValue, Pred Predicate) {
+  auto AtomicFutex = std::atomic_ref<T>(*Futex);
+  T Result = AtomicFutex.load();
+
+  while (!Predicate(Result & BitMask, ComparisonValue)) {
+    Result = LoadExclusive(Futex);
+    if (Predicate(Result & BitMask, ComparisonValue)) {
+      return;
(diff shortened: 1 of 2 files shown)


FEXCore/include/FEXCore/Utils/SpinWaitLock.h
```diff
@@ -144,6 +144,21 @@ static inline void WaitPred(T* Futex, T ComparisonValue) {
   }
 }
 
+template<typename Pred, typename T>
+static inline void WaitBitMaskPred(T* Futex, T BitMask, T ComparisonValue) {
```
lioncash: Rewriting the signature so the template args are reversed, and passing the predicate in like so:

```suggestion
template<typename T, typename Pred>
static inline void WaitBitMaskPred(T* Futex, T BitMask, T ComparisonValue, Pred Predicate) {
```

and then invoking it as:
```cpp
SpinWaitLock::WaitBitMaskPred(TLS.ControlWordAddress(), ControlBits::IN_JIT, 0U, std::equal_to<>());
```
and removing the braces from `Pred {}` in the implementation would make the interface a little more flexible (and also allow passing regular lambdas in too).

Sonicadvance1: Looks like it still correctly inlines in that case so I'm fine with this change. Although generic lambda support doesn't make much sense for these since the only reason why they aren't a big block of inline assembly is to be nicer toward readability. As soon as there is any memory access that isn't the exclusive monitor address this entire thing falls on its face.

Image_2026-07-07_16-09-37

ASM looks correct in this instance at least. Some redundant moves but that's expected with this arrangement.

lioncash (approved):

PR #5663, opened 2026-07-08
JIT/ALUOps: Fix operand overlapping bug for pdep

simon902:
The software implementation for `pdep` incorrectly compares the original mask operand to 0 instead of using the scratch register. When the destination and mask operands overlap (e.g., `pdep rbx, rax, rbx`), clearing the destination register first zeroes out the mask, causing an erroneous early exit.

This PR fixes the bug and adds a unit test.


Sonicadvance1:
Oh, what a fun mistake.

Sonicadvance1 (approved):

PR #5664, opened 2026-07-08
OpcodeDispatcher: Fix 32bit cmpxchg zero extension with eax as first operand

simon902:
This PR fixes a bug in `CMPXCHGOp` where the destination register was not properly zero extended when `eax` was used as the first operand.
Also removed an incorrect comment in the function.


Sonicadvance1 (approved):
So many edge cases with this instruction.

lioncash (approved):

PR #5665, opened 2026-07-08
[SVE256] Handle 256-bit blend operations much more efficiently

lioncash:
We can massage a given selector into a valid predicate register bitmask and then simply perform a merging move, which eliminates most busywork around optimizing 256-bit blends.

In the future, once we drop SVE2.1 support in, we can use PMOV to eliminate the load from memory and related constant management.


Sonicadvance1 (approved):
Fancy fancy

PR #5667, opened 2026-07-09
Fix cvttss2si when prefixed with 0x66

simon902:
When the instruction `cvttss2si` is prefixed with the operand-size override prefix (0x66), FEX computes the result only for the lower 16bits. However, `cvttss2si` does not support 16bit register operands.
Hence, for such cases fall back to 32bit operations.


Sonicadvance1:
Could probably make the frontend smarter to handle these, but eh. This is good enough.

Sonicadvance1 (approved):

PR #5669, opened 2026-07-09
Fix LZCNT tests reading out of bounds

OFFTKP:
Each of these tests have 15 data words but read 16. Due to a `ret` inserted by a `sed` in `ASM/CMakeLists.txt`, the byte after the data section is 0xc3. Unsure if this ret is necessary in any tests, but just added an extra word in every test that produces the same result in r15.


Sonicadvance1 (approved):
lol, oops. Thanks for discovering that.

PR #5674, opened 2026-07-09
Vector: Trim one instruction off insertq

lioncash:
We can fold a bitwise not and and pair into a bic


Sonicadvance1 (approved):
BIC, the saviour we don't deserve.

PR #5675, opened 2026-07-09
VectorOps: Join identical branches in VFMLS/VFNMLS

lioncash:
Same thing, just a little less redundant.


Sonicadvance1 (approved):
hah, how silly.

PR #5677, opened 2026-07-09
Windows/UnixLib: Fix enabling TSO through legacy Proton codepath

mrpippy:
I don't have a setup to test this, but I don't think it could be working: `sizeof(BOOL) == 4`, `sizeof(bool) == 1`


Sonicadvance1 (approved):
Reasonable enough to me. Looks like the legacy path would have returned `STATUS_INFO_LENGTH_MISMATCH` before this. Yolo.

PR #5680, opened 2026-07-10
IR: Remove unused VUShraI

lioncash:
Given that this is currently unused and that we don't have the signed equivalent implemented, we can just remove this for now.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/VectorOps.cpp
@@ -3201,48 +3201,6 @@ DEF_OP(VUShrI) {
   }
 }
 
-DEF_OP(VUShraI) {
-  const auto Op = IROp->C<IR::IROp_VUShraI>();
-  const auto OpSize = IROp->Size;
-
-  const auto BitShift = Op->BitShift;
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
Fun, was used back for sha stuff for implementing element rotates. https://github.com/FEX-Emu/FEX/pull/3271
But new version just hard requires SHA and is quite a bit more efficient. https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp#L21-L41

Would still be nice to have an element rotate by immediate ARM.

Sonicadvance1:
> Fun, was used back for sha stuff for implementing element rotates. #3271 But new version just hard requires SHA and is quite a bit more efficient. https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp#L21-L41
> 
> Would still be nice to have an element rotate by immediate ARM.

Oh hey, SVE2 and SHA3 extensions both introduce XAR which is XOR with rotate by immediate. SHA3 XAR only rotates 64-bit elements (we need 32-bit) while SVE2 XAR allows you to select element size! It's destructive and requires a zero register to be generated but that is a legitimate optimization to the SHA instruction if you want to implement it.

PR #5688, opened 2026-07-10
Thread: Amend pidfd_open check

lioncash:
Checks for success.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls/Thread.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls/Thread.cpp
@@ -158,8 +158,8 @@ FEX::HLE::ThreadStateObject* CreateNewThread(FEXCore::Context::Context* CTX, FEX
   if (flags & CLONE_PIDFD) {
     // Use pidfd_open to emulate this flag
     const int pidfd = ::syscall(SYSCALL_DEF(pidfd_open), Result, 0);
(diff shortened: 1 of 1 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls/Thread.cpp
```diff
@@ -158,7 +158,7 @@ FEX::HLE::ThreadStateObject* CreateNewThread(FEXCore::Context::Context* CTX, FEX
   if (flags & CLONE_PIDFD) {
     // Use pidfd_open to emulate this flag
     const int pidfd = ::syscall(SYSCALL_DEF(pidfd_open), Result, 0);
-    if (Result == ~0ULL) {
+    if (pidfd == -1) {
       LogMan::Msg::EFmt("Couldn't get pidfd of TID {}\n", Result);
```
Sonicadvance1: Should probably change this Result usage in the log to be errno, so it's a useful log.

Sonicadvance1: Maybe result and errno, so it shows which tid it failed on, and the errno :)

lioncash: Updated it

Sonicadvance1 (approved):

Sonicadvance1 (approved):

PR #5700, opened 2026-07-10
Common/Bitset: Minor API changes

lioncash:
A fix to clearing and tidying up. Much of the interface is unused, since the only usage of this is in the IR validation pass


Sonicadvance1 (approved):

Sonicadvance1:
The not clearing elements here was definitely a bit spicy.

PR #5701, opened 2026-07-10
Thread: Amend new thread handling in HandleNewClone()

lioncash:
Ensures that newly cloned threads get tracked properly.


Sonicadvance1:
I'm going to run this one through a small gauntlet of testing before committing, just because it's so spicy :sweat_drops:

Sonicadvance1 (approved):

PR #5703, opened 2026-07-10
x64/Signals: Amend set size in rt_sigtimedwait

lioncash:
We should be checking the size passed in, not the sizeof of it.


Sonicadvance1 (approved):
Oopsies.

Sonicadvance1 (approved):

PR #5713, opened 2026-07-11
x32/FD: Correct total word calculation in select() variants

lioncash:
Previously this would result in a larger amount of words specified than necessary.

e.g. Given nfds = 1:
With AlignUp(1, 32) / 4, we'd end up with supposedly eight words, when it should only be one word.

On the other extreme, given a full fd set of 1024 fds, then we'd end up with 256 words, when it should only be 32.


Sonicadvance1 (approved):
woops

PR #5719, opened 2026-07-12
64BitAllocator: Avoid overwriting Region[0] in Create64BitAllocatorWithRegions

lioncash:
Since this was a reference, this would end up overwriting Region[0] with whatever the smallest region was instead of just being a running pointer to what happened to be the current smallest region.

We can switch over to a pointer to avoid obliterating the first memory region.


Sonicadvance1 (approved):
Oopsies.

PR #5721, opened 2026-07-12
AllocatorHooks: Amend VirtualProtect for Windows

lioncash:
VirtualProtect returns non-zero on success, also the old protection flag parameter isn't allowed to be null.


Sonicadvance1 (approved):
Oh wacky, that would cause early `STATUS_ACCESS_VIOLATION` errors. That's dangerous.

PR #5722, opened 2026-07-12
ElfContainer: Amend entry iteration in GetDynamicLibs()

lioncash:
These were using i in the termination condition, which is for section headers, not entries.


Sonicadvance1 (approved):
And I oop.

PR #5724, opened 2026-07-12
WinAPI/IO: Fix handling of end of file offset in SetFilePointerEx

lioncash:
This just means the end of the file is being used as the base offset.

Also note that according to the documentation for SetFilePositionEx, that setting the position beyond the current file size is not considered an error as far as the API is concerned.


Sonicadvance1 (approved):
Nice that it behaves just like lseek then.

PR #5726, opened 2026-07-12
Utils/File: Handle dual read/write case

lioncash:
According to POSIX open docs, this is a completely separate flag that isn't a combination of O_RDONLY and O_WRONLY (TIL), so we need to handle this separately.

Makes the codepath behaviorally symmetric with the Windows one.


Sonicadvance1 (approved):
```cpp
#define O_RDONLY        00000000
#define O_WRONLY        00000001
#define O_RDWR          00000002
```
Huh, indeed.

PR #5727, opened 2026-07-12
x32/FD: Make use of SanitizeIOCount for vector construction in vmsplice

lioncash:
Makes this consistent with the other fd syscalls that make temporary buffers.


Sonicadvance1 (approved):
Might actually need to look in to these a little further. The Sanitize only ensures negatives of `int` works, where a bunch of these syscalls are `unsigned long` or `uint32_t`. Need to see what the kernel does.

But since this is also used just above in the verify, let's make it consistent first.

PR #5731, opened 2026-07-12
Crypto: Make use of XAR in SHA1NEXTE when available

### LINKED ISSUE #5730: Use SVE `XAR` when available for SHA
When SHA rotates by 30-bits, we have to do a silly duplication dance to get the rotate in to the correct element.
When SVE is available, we can use its XAR instruction plus a zero-vector register to save an instruction or two for hardware that supports it.

We can't use ASIMD XAR that FEAT_SHA3 added because it doesn't match the behaviour that we want.

lioncash:
Lets us shave an instruction off on hardware that supports XAR.

Closes #5730

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp
@@ -26,15 +26,23 @@ void OpDispatchBuilder::SHA1NEXTEOp(OpcodeArgs) {
   Ref Dest = LoadSourceFPR(Op, Op->Dest, Op->Flags);
   Ref Src = LoadSourceFPR(Op, Op->Src[0], Op->Flags);
 
-  // ARMv8 SHA1 extension provides a `SHA1H` instruction which does a fixed rotate by 30.
-  // This only operates on element 0 rather than element 3. We don't have the luxury of rewriting the x86 SHA algorithm to take advantage of this.
-  // Move the element to zero, rotate, and then move back (Using duplicates).
-  // Saves one instruction versus that path that doesn't support SHA extension.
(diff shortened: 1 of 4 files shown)


FEXCore/Source/Interface/Core/OpcodeDispatcher/Crypto.cpp
```diff
@@ -26,15 +26,23 @@ void OpDispatchBuilder::SHA1NEXTEOp(OpcodeArgs) {
...
-  // Saves one instruction versus that path that doesn't support SHA extension.
-  auto Duplicated = _VDupElement(OpSize::i128Bit, OpSize::i32Bit, Dest, 3);
-  auto Sha1HRotated = _VSha1H(Duplicated);
-  auto RotatedNode = _VDupElement(OpSize::i128Bit, OpSize::i32Bit, Sha1HRotated, 0);
-  auto Tmp = _VAdd(OpSize::i128Bit, OpSize::i32Bit, Src, RotatedNode);
-  auto Result = _VInsElement(OpSize::i128Bit, OpSize::i32Bit, 3, 3, Src, Tmp);
+  Ref Result {};
+  if (CTX->HostFeatures.SupportsSVE128) {
+    auto ZeroVec = LoadZeroVector(OpSize::i128Bit);
+    auto Tmp = _VInsElement(OpSize::i128Bit, OpSize::i32Bit, 3, 3, ZeroVec, Dest);
```
Sonicadvance1, line 32: If you `VExtr` here instead of VIns it should convert the `mov+ins` pair in to a single `ext`

Sonicadvance1: Oh wait, I'm a derp. I was thinking element 0 in to element 3. nevermind.

Sonicadvance1 (approved):

PR #5733, opened 2026-07-13
64BitAllocator: Avoid madvising more than necessary in InitializeVMARegionsUsed()

lioncash:
Because our underlying bitset type is uint64_t, then that means `Memory + ManagedSize` is more like: `Memory + (ManagedSize * 8)`, which is way larger of a base than we need.


Sonicadvance1:
I'll spend a bit of time tomorrow validating all the bits to see how wrong this was before. Been a while since I've looked at this code.

Sonicadvance1 (approved):
Yea okay, confirmed the logic. That was pretty bad.

PR #5737, opened 2026-07-13
VectorOps: Simplify SVE 256-bit VOrn with BSL2N

lioncash:
Lets us shave off an instruction and also avoid using a temporary register in some cases. We can also tweak our worst case that requires a predicate to eliminate the temporary as well.

We can also expand our cmpps cases, so that we can reflect the BSL2N usages in instcountci.


Sonicadvance1 (approved):
Oh fancy, wasn't aware of bsl{1,2}n.

PR #5741, opened 2026-07-13
Utils/File: Fix handle releasing

lioncash:
ShouldClose was never being set in the event we opened a regular file. The only time it was set (to false) is when it's used to encapsulate stderr and stdout.

So anything opened by a File instance was essentially held open.


Sonicadvance1 (approved):
Spicy.

PR #5742, opened 2026-07-13
x32/Semaphore: Fix storing of message type in msgrcv

lioncash:
This was previously storing into the local compat handler, not the actual managed message.


Sonicadvance1 (approved):
Whoops

PR #5743, opened 2026-07-13
StringUtils: Handle strings entirely composed of whitespace in trims

lioncash:
Previously this wouldn't handle fully whitespaced strings.


Sonicadvance1 (approved):
I love unittests.

PR #5744, opened 2026-07-13
AtomicOps: Avoid constrained unpredictable case in TelemetrySetValue()

lioncash:
STLXR cannot use the same register as both the status register and the value register, otherwise it's architecturally unpredictable behavior.

Only applies to hardware without FEAT_LSE, so this only meaningfully affects hardware using the v8.0 spec, since FEAT_LSE becomes mandatory in v8.1 and newer.


Sonicadvance1 (approved):
Oop

PR #5745, opened 2026-07-14
VectorOps: Simplify 256-bit VAddV

lioncash:
Didn't read the manual close enough on the first read when I dropped this in award.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/VectorOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/VectorOps.cpp
@@ -1199,20 +1199,16 @@ DEF_OP(VAddV) {
   const auto Vector = GetVReg(Op->Vector);
 
   if (HostSupportsSVE256 && Is256Bit) {
-    // SVE doesn't have an equivalent ADDV instruction, so we make do
-    // by performing two Adv. SIMD ADDV operations on the high and low
-    // 128-bit lanes and then sum them up.
-
-    const auto Mask = PRED_TMP_32B.Zeroing();
-    const auto CompactPred = ARMEmitter::PReg::p0;
-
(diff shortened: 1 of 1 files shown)


FEXCore/Source/Interface/Core/JIT/VectorOps.cpp
```diff
@@ -1199,20 +1199,8 @@ DEF_OP(VAddV) {
...
-    const auto CompactPred = ARMEmitter::PReg::p0;
-
-    // Select all our upper elements to run ADDV over them.
-    not_(CompactPred, Mask, PRED_TMP_16B);
-    compact(ARMEmitter::SubRegSize::i64Bit, VTMP1.Z(), CompactPred, Vector.Z());
-
-    addv(SubRegSize.Vector, VTMP2.Q(), Vector.Q());
-    addv(SubRegSize.Vector, VTMP1.Q(), VTMP1.Q());
-    add(SubRegSize.Vector, Dst.Q(), VTMP1.Q(), VTMP2.Q());
+    uaddv(SubRegSize.Vector, Dst.D(), Mask, Vector.Z());
```
Sonicadvance1: I don't think this exactly matches behaviour. Looks like this always zero extends smaller elements to be 64-bit, so overflow for something like 16-bit additions could return garbage in the upper 48-bits instead of truncating to a 16-bit scalar.

Which most operations immediately feed in to a VExtractToGPR  so it can't see that, but something like `PSADBWOpImpl` isn't so it would be visible there.

Might want a unittest for the few instructions that /can/ overflow, and maybe a bool in the IR operation to force an element size mask for operations that need it? ptrue should be able to generate the mask, which will add a `ptrue + mov` pair if masking is absolutely required?

Sonicadvance1 (approved):

PR #5750, opened 2026-07-15
VectorOps: Make use of unpredicated shifts

lioncash:
Same behavior, just without introducing a predicate register dependency. Largely just an artifact from initial bringup.


Sonicadvance1 (approved):
ooo, fancy.

PR #5752, opened 2026-07-15
FEXGetConfig: Add convenience option for dumping system/tso info

lioncash:
Lets you get a broad overview of TSO emulation, fault granularity, and the feature regs all at once instead of needing to type out every long command for them individually.

Now it's easier to be lazy and just pass "-e", or "--all-emu-info".


Sonicadvance1 (approved):

Sonicadvance1:
oop, merge conflict.

PR #5760, opened 2026-07-16
Arm64Emitter: Fix incorrect condition for constant NOP padding

neobrain:
This needs to be enabled when *generating* caches, not at runtime when we're loading them (unless we're compiling for validation).

Previous code would incorrectly disable NOP padding in FEXOfflineCompiler and instead enable it at runtime when it wasn't needed.


Sonicadvance1 (approved):
Needs a rebase.

Ari4ka (approved):

PR #5761, opened 2026-07-16
LinuxEmulation: Resolve missing prototype warnings

lioncash:
Ensures all functions are marked whether they're intended to be internally linked or not.

### DIFF
--- a/Source/Tools/LinuxEmulation/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
+++ b/Source/Tools/LinuxEmulation/LinuxSyscalls/EmulatedFiles/EmulatedFiles.cpp
@@ -67,7 +67,7 @@ static void SealTmpFD(int fd) {
   }
 }
 
-fextl::string GenerateCPUInfo(FEXCore::Context::Context* ctx, uint32_t CPUCores) {
+static fextl::string GenerateCPUInfo(FEXCore::Context::Context* ctx, uint32_t CPUCores) {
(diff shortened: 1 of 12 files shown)


Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
```diff
@@ -74,6 +74,39 @@ class SyscallHandler;
 class SignalDelegator;
 class ThunkHandler;
 
+namespace x32 {
+void RegisterEpoll(FEX::HLE::SyscallHandler* Handler);
```
Sonicadvance1: These end up being defined in their respective subdirectories already in Syscalls.cpp. Should those be removed and keep these ones?
Or not at these? The scoping on these are setup to only be on the x32/x64 respective sides before this change.

Sonicadvance1 (commented):
Code formatting says hello!

Source/Tools/LinuxEmulation/LinuxSyscalls/Syscalls.h
lioncash: Moved them over to their respective .h file to be more consistent. The reason for the warnings is that they weren't visible themselves to their implementations.

Sonicadvance1: I see, I shee.

Sonicadvance1 (approved):

PR #5762, opened 2026-07-17
FEXCore: Resolve missing prototype warnings

lioncash:
Makes sure we mark everything internally linked as necessary, or make declarations visible to their implementation.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.cpp
+++ b/FEXCore/Source/Interface/Config/Config.cpp
@@ -1,9 +1,10 @@
 // SPDX-License-Identifier: MIT
 #include "Common/StringConv.h"
-#include "FEXCore/Utils/EnumUtils.h"
+#include "Utils/Config.h"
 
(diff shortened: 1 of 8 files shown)


FEXCore/Source/Interface/GDBJIT/GDBJIT.cpp
```diff
@@ -35,7 +35,7 @@ struct jit_descriptor {
 constinit jit_descriptor __jit_debug_descriptor = {.version = 1};
 
 /* GDB puts a breakpoint in this function.  */
-void __attribute__((noinline)) __jit_debug_register_code() {
+static void __attribute__((noinline)) __jit_debug_register_code() {
```
Sonicadvance1: Technically this one isn't actually internal, the symbol needs to be publicly visible. It's just likely been broken ever since we enabled hidden visibility by default.

Sonicadvance1 (approved):

Ari4ka (approved):

PR #5764, opened 2026-07-17
LinuxAllocator: Reduce MemAllocator32Bit size by 16 bytes

lioncash:
These constants don't need to be member vars.


Sonicadvance1 (approved):
oopsies

PR #5780, opened 2026-07-22
Windows: Fixes around using Unicode functions

mrpippy:
Some small fixes based on what I think are best practices on Win32:

- Use a *W function instead of *A if it's easy to do so, since you'll avoid unnecessary string conversions (i.e. `GetModuleHandleW(L"ntdll.dll")` instead of `GetModuleHandleA("ntdll.dll")`
- Always use the *A/*W function directly, rather than relying on the `#define` which changes based on the definition of `UNICODE` (i.e. use `CreateFileA()` or `CreateFileW()`, not `CreateFile()`)
- Define `UNICODE` project-wide so it's harder to accidentally use the *A functions.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -195,6 +195,9 @@ if (ENABLE_GDB_SYMBOLS)
 endif()
 
 add_compile_definitions(_LARGEFILE64_SOURCE)
+if (WIN32)
+  add_compile_definitions(UNICODE _UNICODE)
+endif()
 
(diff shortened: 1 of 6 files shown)


Sonicadvance1:
Well, let's see if compilation explodes with that define. It's not like we implement every Win32 API inside of `Source/Windows/Common/WinAPI/`

Sonicadvance1 (approved):
Certainly nothing bad can come from this.

PR #5781, opened 2026-07-23
FEXCore: Isolate multiblock error state per block

FrontMage:
## What changed

- Reset `HadDispatchError` and `HadInvalidInst` for each decoded block in a multiblock compilation unit.
- Add an ASM regression where an unreachable block contains a valid but unsupported `mov fs, ax`, while a later independent block must still compile and return the expected result.

## Why

These flags describe whether the current decoded block needs an early exit, but they were scoped to the entire multiblock compilation unit. A dispatcher failure in one independently reachable block therefore leaked into later blocks and caused them to emit premature exits.

Keeping the state block-local lets the failing block terminate without truncating unrelated entry paths.

## Validation

ARM64 Ubuntu, Clang 21.1.8:

| Configuration | Before | After |
| --- | --- | --- |
| `jit_1` | Pass | Pass |
| `jit_500` | Pass | Pass |
| `jit_500_m` | Fail (`RAX=0x10`) | Pass (`RAX=0x20`) |

All 64-bit Multiblock ASM tests pass after the change: 6/6.

This is one generic frontend correctness issue encountered while investigating #5399; it is not intended to be a complete fix for that issue.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -560,10 +560,14 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
     }
 #endif
 
-    bool HadDispatchError {false};
-    bool HadInvalidInst {false};
     for (size_t j = 0; j < CodeBlocks.size(); ++j) {
       const auto& Block = CodeBlocks[j];
+
+      // Dispatch failures and invalid instructions terminate only the decoded
+      // block that contains them. Other block targets in the same multiblock
+      // compilation unit are independent entry paths.
+      bool HadDispatchError {false};
+      bool HadInvalidInst {false};
 
 #ifdef ZYDIS_DISASSEMBLER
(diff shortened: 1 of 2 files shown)


FrontMage:
This is one of the fixes required to fix the official World of Warcraft #5399

FEXCore/Source/Interface/Core/Core.cpp
```diff
@@ -560,11 +560,15 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
     }
 #endif
 
-    bool HadDispatchError {false};
-    bool HadInvalidInst {false};
     for (size_t j = 0; j < CodeBlocks.size(); ++j) {
       const auto& Block = CodeBlocks[j];
```
Sonicadvance1, line 565: This comment is incorrect. `HadDispatchError` throws away all compilation results on dispatch error, while `HadInvalidInst` ends the block early.

Sonicadvance1: Er no, missed the additional comparisons.

Sonicadvance1 (approved):
Drop the AI overviews next time.

PR #5782, opened 2026-07-23
IRValidation: Minor cleanups

lioncash:
Just a few trivial things I noticed while in the area.


Sonicadvance1 (approved):
Receiving validation.

PR #5783, opened 2026-07-23
FEXCore: Ignore inactive JIT guard ranges

FrontMage:
This is one of the fixes required for #5399.


Sonicadvance1 (approved):
lol, what a funny failure mode.

FrontMage:
If it weren't for WoW's Eidolon loader, I would never have found this failure; that thing is the ultimate edge case tester, might want to consider putting it in the unit test :)

PR #5784, opened 2026-07-24
Frontend: Prioritize instruction fetch faults

FrontMage:
This is one of the fixes required for #5399.

PR #5785, opened 2026-07-24
FEXCore: Report instruction fetch fault address

FrontMage:
This is one of the fixes required for #5399.

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -96,6 +96,9 @@ bool Decoder::CheckRangeExecutable(uint64_t Address, uint64_t Size) {
     ExecutableRangeWritable = RangeInfo.Writable;
 
     if (RangeInfo.Size == 0) {
+      if (NonExecutableAddress == 0) {
+        NonExecutableAddress = Address;
+      }
       return false;
     }
 
@@ -1086,12 +1089,14 @@ Decoder::DecodedBlockStatus Decoder::DecodeInstruction(uint64_t PC) {
   // Will be set if DecodeInstructionImpl tries to read non-executable memory
   HitNonExecutableRange = false;
   HitBadRelocation = false;
+  NonExecutableAddress = 0;
   auto ErrorDuringDecoding = DecodeInstructionImpl(PC);
 
   if (ErrorDuringDecoding != DecodedBlockStatus::SUCCESS || HitNonExecutableRange || HitBadRelocation) [[unlikely]] {
     // Put an invalid instruction in the stream so the core can raise SIGILL if hit
     // Error while decoding instruction. We don't know the table or instruction size
     DecodeInst->TableInfo = nullptr;
+    DecodeInst->FaultAddress = NonExecutableAddress;
     auto Result = ErrorDuringDecoding != DecodedBlockStatus::SUCCESS ? ErrorDuringDecoding :
                   DecodeInst->InstSize                               ? DecodedBlockStatus::PARTIAL_DECODE_INST :
                   HitNonExecutableRange                              ? DecodedBlockStatus::NOEXEC_INST :
--- a/FEXCore/Source/Interface/Core/Frontend.h
+++ b/FEXCore/Source/Interface/Core/Frontend.h
@@ -126,6 +126,7 @@ private:
   uint64_t ExecutableRangeEnd {};
   bool ExecutableRangeWritable {};
   bool HitNonExecutableRange {};
+  uint64_t NonExecutableAddress {};
   bool HitBadRelocation {};
 
   struct DecodeStream {
--- a/FEXCore/Source/Interface/Core/JIT/MiscOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/MiscOps.cpp
@@ -76,6 +76,9 @@ DEF_OP(Break) {
   LoadConstant(ARMEmitter::Size::i64Bit, TMP1, Constant);
   str(TMP1, STATE, offsetof(FEXCore::Core::CpuStateFrame, SynchronousFaultData));
 
+  LoadConstant(ARMEmitter::Size::i64Bit, TMP1, Op->Reason.FaultAddress);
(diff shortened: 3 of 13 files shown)


dragosol:
Real-world validation datapoint for this PR: **it is the fix that makes Overwatch 2 get past the Battle.net "Eidolon" loader.**

Setup: aarch64 Arch Linux guest under QEMU on Apple Silicon, Mesa venus -> virglrenderer -> MoltenVK -> Metal, Steam + Proton under FEX, FEX built from `main` (FEX-2608 + ~130 commits).

Without this PR, on `main`, the loader deterministically emits **39 `NoExec instruction in entry block`** errors across ~40 unique addresses inside `Overwatch.exe`'s PE image (0x140000000/0x142000000), then spins forever re-decoding the same RIP - the process never progresses. We bisected the onset (`git bisect --first-parent`, 968 merges, 9 build+test cycles) to the merge of PR #5024 "feature_jit_encoder_recovery" (`3ccdf6508`, 2025-11-05); its first-parent predecessor `fd33cf1ce` is clean.

Note that #5781, #5783 and #5784 alone (i.e. current `main`) are **not** sufficient - we verified the livelock still reproduces on `main` without #5785.

With #5785 applied on top of `main`:

- the loader completes deobfuscation,
- .NET/mscoree initialises,
- DXVK starts and creates a D3D11 device,
- the game reaches its login screen, authenticates with Battle.net, loads the main menu, and we have since run the in-game Practice Range for 22 minutes continuously at 1080p.

So from our side this PR is confirmed load-bearing for a second Blizzard title beyond WoW, and we would very much like to see it land. Happy to provide logs or re-test any revision.

(One unrelated bug we hit immediately afterwards, in case it helps anyone else chasing this class: the thunk layer's `CallHostFunction` passes the host callee address in `r11`, which the stack-protector prologue clobbers for >=7-argument functions. Filed separately.)


(Filed the thunk bug separately as #5878.)

PourOutYourPowerAndLove:
@dragosol Really this PR fixes Overwatch 2? I haven't tested SC2 and WoW, only HotS and to my understanding basically all the Blizzard games were broken at least until recently, unverified.  I might be able to test it myself, but only if it installs over the steam client because the battle.net woes right now under emulation seem to be coming back. AFAIK Eidolon has been the target of this work and if it works it brings WoW, Diablo series, and Overwatch a lot closer to running under FEX. SC2 and Heroes of the Storm afaik do not use eidolon and seems whatever issue is related to relocations is holding both of those games back from running

Also, you ran FEX under QEMU on macOS? I thought crossover provides FEX builds for ARM64EC wines at the very least, i'm not sure about being able to build a FEX binary to run x86 wine, but FEX on macOS is probably unsupported anyway.

dragosol:
Yeah, genuinely — but let me be precise about the setup, because I think I described it badly and it sounds more exotic than it is.

**FEX is not running on macOS.** It's a bog-standard aarch64 Arch Linux guest (Hyprland, Steam, Proton) running under QEMU with HVF acceleration on an M3 Max. Inside that guest everything is normal Linux FEX. macOS only provides the hypervisor and the GPU: guest Mesa **venus** → virtio-gpu → virglrenderer → MoltenVK → Metal. So no FEX-on-macOS weirdness, and no ARM64EC/CrossOver involvement — this is the plain "x86 Steam + x86 Proton entirely under FEX" config.

**On your Battle.net concern: not an issue here, and this is the useful bit for you.** It's the **Steam** version of OW2 (appid 2357570), so there's no Battle.net client involved at all. Steam launches `Overwatch.exe` directly and the account link happens in-game. So if the Battle.net app is what's blocking you on HotS, OW2-via-Steam sidesteps it entirely and should be a much easier test case.

What I actually see with #5785 on top of `main`:

- without it: 39 `NoExec instruction in entry block` across ~40 unique addresses in the PE image, then an endless re-decode spin at one RIP
- with it: loader clears, mscoree inits, DXVK creates a D3D11 device, login screen → Battle.net auth → main menu → Practice Range, 22 min continuous at 1080p

**Limitations, so nobody thinks this is "OW2 works, ship it":**

- I also needed a second fix that isn't in this PR — the thunk layer's `CallHostFunction` passes the host callee address in `r11`, which the stack-protector prologue clobbers for ≥7-arg functions. Without that, it dies in `vkGetQueryPoolResults` during D3D11 device creation. Filed as #5878. You'll likely hit it too.
- Graphics quality has to be **Low**. The game auto-detects Ultra (venus reports a 49 GB "Apple M3 Max" GPU 😄) and Ultra makes Metal command buffers long enough to trip Apple's GPU watchdog, killing the VkDevice. That's a MoltenVK/Metal ceiling, nothing to do with FEX.
- ~37–58 fps at 1080p, with real stutter. GPU sits at only ~40% util, so I'm likely CPU-bound in emulation.
- Colours are wrong (red/blue swapped, everyone looks like a Smurf) — that one's a bug in my own host-side virglrenderer, not FEX.
- Mouse aim doesn't work yet — my QEMU pointer device config, also not FEX.

So: FEX's side of this is the part that's actually working. Everything still broken is on my end of the stack.

**What I'm chasing overall:** a single machine that's simultaneously a good macOS box and a near-native Linux gaming VM — full control over every translation layer so I can actually fix the stuttering rather than accept it. That's why the VM instead of CrossOver/Whisky; I want the layers visible and patchable.

Happy to share exact build/config, or re-test any revision if that helps get this landed.

dragosol:
Also, in the interest of being upfront since the write-up probably reads a bit *too* polished: I've been driving this with Claude doing the grunt work — the bisects, the log spelunking, the disassembly reading — while I prompt and steer each step and review what comes back.

The findings themselves are from real runs on real hardware, not generated prose: the bisect was 9 actual build+test cycles, the `NoExec` counts and addresses come out of actual FEX logs, and the r11/canary thing in #5878 was caught with a live gdb catchpoint conditioned on `si_addr == __stack_chk_guard`. I can produce the raw logs, the disassembly, or a full config dump for any claim in either report if you want to check something specific.

Flagging it because I know AI-generated patches/reports have a bad reputation in this space and I'd rather you know the provenance up front than wonder. Any number in there you doubt, ask and I'll paste the source.

PR #5790, opened 2026-07-28
Fix building for Windows with libc++ 23

mstorsjo:
Add missing includes, add a couple of function stubs.


Sonicadvance1 (approved):
Seems reasonable enough to me. Thanks!

PR #5795, opened 2026-08-01
Fix XMM register bug in SpillSRA

Iaying6564:
in the SpillSRA function the XMM register memory layout is determined by checking the SupportsAVX flag alone. This assumes that SupportsAVX means the host has 256-bit registers, however SupportsAVX is a guest-facing flag that determines whether or not AVX is being emulated. While this does affect memory layout, it does not solely determine if the host is using the AVX (256-bit) memory layout. In order to fix this we add a `&& SupportsSVE256` after the SupportsAVX so that it uses the correct memory layout when the machine does not have SVE256

a test that repros this bug is also attached


Sonicadvance1:
Thanks for the PR. That was quite a bad bug indeed. I've merged it over at #5795 so you don't need to deal with the code formatting.
Thanks for the fix!

PR #5064, opened 2025-11-18
Fix building under clang-22

zeyi2:
Closes #5032

PR #5804, opened 2026-08-05
thunkgen: satisfy the clang 22 ComplierInstance VFS invariant.

sunshineinabox:
Fixes 

Clang 22's CompilerInstance expects its virtual file system to match the FileManager's VFS. Set it when building against LLVM 22+.

Follow up to [#5064](https://github.com/FEX-Emu/FEX/pull/5064).

Context: https://github.com/ROCKNIX/distribution/pull/3103


Sonicadvance1 (approved):
Huh. Seems fine to me. Thanks.

PR #5806, opened 2026-08-05
Fix precision loss for F64 to i32 conversion

simon902:
Fixes the precision issue the comment in the non-SVE fallback path in `Vector_F64ToI32` already described.

First convert f64 to i64 and then narrow the i64 to an i32.
This means nothing is rounded before the integer conversion, so sources needing more than f32's 24-bit mantissa survive intact.

Contrary to what the comment assumed, there shouldn't be a performance penalty: the new sequence is two vector instructions, exactly like the `fcvtn` + `fcvtzs` pair it replace (not the per-element scalar `fcvtzs` to GPR the comment had in mind).

The code uses `sqxtn` rather than `xtn` because `Vector_CVT_Float_To_Int32Impl` only fixes up
positive overflow, so the clamp to `INT32_MIN` is what gives negative out-of-range sources the `0x80000000` x86 expects (integer-indefinite value).

Also `RoundMode::TowardsZero` no longer needs its `frintz`, since `fcvtzs` already truncates.

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/ConversionOps.cpp
+++ b/FEXCore/Source/Interface/Core/JIT/ConversionOps.cpp
@@ -558,16 +558,18 @@ DEF_OP(Vector_F64ToI32) {
     }
   } else {
     ///< Round float to integral depending on rounding mode.
+    ///< skip TowardsZero as fcvtzs below already truncates toward zero on its own
+    auto CVTReg = Dst.Q();
     switch (Round) {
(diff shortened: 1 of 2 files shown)


Sonicadvance1 (approved):
Sounds good to me, and Animal Well's music is still working correctly.

I'm guessing the comment is just old before the caller handled the overflow case which is why this wasn't a potential path forward before.
Thanks!

PR #5809, opened 2026-08-06
FCOMI fails to clear OF, SF, AF

simon902:
For FCOMI/FCOMIP/FUCOMI/FUCOMIP: the Intel SDM states these instructions "set OF, SF, and AF to 0" regardless of the comparison result.
The AMD manual doesn't mention OF/SF/AF for these instructions at all.

Testing on real hardware (at least 3 AMD systems) confirms OF, SF, and AF are unconditionally cleared to 0 after these instructions, regardless of their prior value.

This PR makes it such that the affected instructions always clear OF, SF and AF to zero.


Sonicadvance1:
Looks like all three of the ARMv8.4 runners decided to off themselves simultaneously for some reason today. That'll stall out all PRs for the day until I can figure out what happened there.

simon902:
Hey @Sonicadvance1, do the runners work again? Otherwise we could retrigger the CI for this PR? Let me know if you need anything else.

Sonicadvance1 (approved):
Let's assume this is fine and just yolo.

PR #5812, opened 2026-08-07
POPF and PUSHF fail to preserve and clear specific flag bits

simon902:
## POPF

FEX executes in CPL=3 with IOCPL < CPL.
Per the Intel Manual entry for `POPF/POPFD/POPFQ` this then means that the flags should be updated the following way:

| Operand Size | CPL | IOPL | ID (21) | VIP (20) | VIF (19) | AC (18) | VM (17) | RF (16) | NT (14) | IOPL (13:12) | OF (11) | DF (10) | IF (9) | TF (8) | SF (7) | ZF (6) | AF (4) | PF (2) | CF (0) |
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
| 16 | 1-3 | < CPL | N | N | N | N | N | 0 | S | N | S | S | N | S | S | S | S | S | S |
| 32, 64 | 1-3 | < CPL | S | N | N | S | N | 0 | S | N | S | S | N | S | S | S | S | S | S |

| Key | Description |
|---|---|
| S | Updated from stack |
| SV | Updated from IF (bit 9) in FLAGS value on stack |
| N | No change in value |
| X | No EFLAGS update |
| 0 | Value is cleared |

I.e., `VIP, VIF, VM, IOPL, IF` should remain unchanged and `RF` cleared to zero.
For the case the instruction operates on 16bit also the `ID, AC` flags need to be preserved.

## PUSHF

Furthermore, also per the SDM, `VM, RF` need to be cleared to zero for `PUSHF`.

---

This PR, preserves and clears the aforementioned flags and also adds a unittest.

### DIFF
--- a/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
+++ b/FEXCore/Source/Interface/Core/OpcodeDispatcher.cpp
@@ -3625,7 +3625,10 @@ void OpDispatchBuilder::BSWAPOp(OpcodeArgs) {
 void OpDispatchBuilder::PUSHFOp(OpcodeArgs) {
   const auto Size = OpSizeFromSrc(Op);
 
-  Push(Size, GetPackedRFLAG());
+  // VM, RF cleared to 0
+  constexpr uint32_t FlagsMask = ~((1U << FEXCore::X86State::RFLAG_VM_LOC) | (1U << FEXCore::X86State::RFLAG_RF_LOC));
+
+  Push(Size, GetPackedRFLAG(FlagsMask));
 }
 
 void OpDispatchBuilder::POPFOp(OpcodeArgs) {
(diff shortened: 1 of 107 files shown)


Sonicadvance1:
Do any of the flags you're now allowing to save and restore actually need to have side-effects implemented?
Also looking suspiciously like an AI overview. Was this written with ML?

simon902:
1. What do you mean by `need to have side-effects implemented`?
2. Now as you mention it, it kind of looks like AI xD. But no, i did not use AI to write the text. I just copied the tables from the SDM to make it clear what i am referring to and added heading to distinguish between pop and push as they are essentially two different bug cases.

simon902:
Hey @Sonicadvance1, just pinging again since maybe it got lost. Let me know if I can provide more context to my previous message!

simon902:
Fixed the test to not fail on the native runner, reformatted code and rebased against main

PR #5815, opened 2026-08-07
FEXCore/Utils: Implements a new atomic segmented bitmap allocator 

Sonicadvance1:
This is tailored towards our needs for our JIT and eventually replacing
the linear allocator. Allowing us to reallocate memory for code blocks
that have been invalidated, letting us keep a single code buffer around
for longer and using less memory overall.

In particular, high-invalidation games that ship anti-tamper tend to
emit millions of ~128-byte blocks in just a handful of minutes which
causes our current linear allocator to consume gigabytes very quickly.
This will allow us to more aggressively reuse the allocation space and
reduce the memory load in those situations.

There's some additional resize tuning that needs some work whence it is
in situ which doesn't need to be done now.

### DIFF
--- a/FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
+++ b/FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
@@ -1,0 +1,474 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <FEXCore/Utils/TypeDefines.h>
+#include "Utils/atomic_bitset.h"
+
+#include <array>
+#include <optional>
+
+namespace FEXCore::Utils {
+struct bitmap_allocator_settings {
+  std::array<uint32_t, 3> bucket_sizes;
+  std::array<uint32_t, 3> bucket_allocation_percentages;
+};
+
+// Default allocation settings passed through template for easy tinkering.
+constexpr static bitmap_allocator_settings fex_default_allocator_settings = {
+  .bucket_sizes =
+    {
+      16,   // 16B - 1KB
+      512,  // 512B - 32KB
+      2048, // 2KB - 128KB
+    },
+  .bucket_allocation_percentages =
+    {
+      // 16B bucket special cased to allocate remaining space left over from other buckets.
+      100,
+      // 40% in the 512B bucket.
+      40,
+      // 10% in the 2KB bucket.
+      10,
+    },
+};
+
+/**
+ * A bitmap allocator that is segmented in to three partitioned buckets, with atomic memory allocation.
+ *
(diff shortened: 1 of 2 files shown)


FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
```diff
@@ -0,0 +1,477 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <FEXCore/Utils/TypeDefines.h>
+#include "Utils/atomic_bitset.h"
+
+#include <optional>
+
+namespace FEXCore::Utils {
+struct bitmap_allocator_settings {
+  uint32_t bucket_sizes[3] {};
```
lioncash: std::array, perhaps

FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
```diff
@@ -0,0 +1,477 @@
...
+  uint8_t* arena_base {};
+  size_t arena_base_size {};
+
+  // Bitset tracking.
+  uint8_t* bitset_base {};
+  size_t bitset_base_size {};
+
+  // Bucket sizes are important for how much can be allocated in a single 64-bit atomic operation.
+  // Ensure these two buffers are sorted by size.
+  constexpr static std::array<uint32_t, 3> bucket_granule_sizes = {{
```
lioncash: Coupled with the above, this seems like it can be a direct assignment

lioncash (commented):
Just two nits, but overall looks good

FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
Sonicadvance1: Ah, so it can be.

FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
Sonicadvance1: Changed over.

FEXCore/Source/Utils/atomic_segmented_bitmap_allocator.h
```diff
@@ -0,0 +1,473 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+#include <FEXCore/Utils/TypeDefines.h>
+#include "Utils/atomic_bitset.h"
+
+#include <optional>
+
+namespace FEXCore::Utils {
+struct bitmap_allocator_settings {
+  std::array<uint32_t, 3> bucket_sizes;
```
lioncash, line 11: `<array>` needs to be included

Sonicadvance1: oop, you right. Fixed.

lioncash (approved):

PR #5818, opened 2026-08-09
unittests/ASM: Ensure ffreep always has an operand

OFFTKP:
It seems that on nasm `ffreep` without an operand assembles to `dfc1` which is `ffreep st1`. 
Added an operand to all the ffreep instructions that didn't have one to avoid surprises, as I assume st0 was the intended operand.


Sonicadvance1 (approved):
Oh weird

PR #5819, opened 2026-08-10
Fix warnings in llvm-mingw builds

cjacek:


Sonicadvance1 (approved):
Sounds fine to me. Thanks!

PR #5820, opened 2026-08-11
Decoder: fix illegal LEA encoding

javelina-pkwy:
Fixes subtle decoder bug allowing execution of illegal register-to-register variants of LEA.

For example, before this fix these bytes:
`0x66, 0x8D, 0xC0 (ie. LEA ax, ax)`

Would be translated as 
```
uxth w20, w4
bfxil x4, x20, #0, #16
```

We now treat all of these encodings as illegal.


Sonicadvance1 (approved):
Seems benign enough. Thanks.

PR #5823, opened 2026-08-13
SharedCodeBufferManager: Allocate JIT space atomically.

Sonicadvance1:
This removes the fairly long lived lock that the buffer allocator held while doing significantly more work than intended while holding that lock.

As the first step towards moving over to the atomic bitmap allocator, change this to be atomic to closer match what the new allocator is doing. Since we are just doing linear allocations, this is an easy convert and should give a good stutter improvement.

(Sadly I didn't have any lock tracking on this lock to know how contended it is. So even though its an improvement, it would be hard to quantify anything.)

### DIFF
--- a/FEXCore/Source/Interface/Core/JIT/JITClass.h
+++ b/FEXCore/Source/Interface/Core/JIT/JITClass.h
@@ -624,6 +624,8 @@ private:
 
   void EmitEntryPoint(ARMEmitter::BackwardLabel& HeaderLabel, bool CheckTF);
 
+  [[nodiscard]] SharedCodeBufferManager::CodeBufferAllocation AllocateCodeBufferInSharedCache(size_t Size);
+
 #define DEF_OP(x) void Op_##x(IR::IROp_Header const* IROp, IR::Ref Node)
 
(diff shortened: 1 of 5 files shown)


FEXCore/Source/Interface/Core/JIT/JITClass.h
```diff
@@ -624,6 +624,8 @@ class Arm64JITCore final : public CPUBackend, public Arm64Emitter {
 
   void EmitEntryPoint(ARMEmitter::BackwardLabel& HeaderLabel, bool CheckTF);
 
+  SharedCodeBufferManager::CodeBufferAllocation AllocateCodeBufferInSharedCache(size_t Size);
```
lioncash: Since this one returns concrete allocation info, maybe this should be annotated with `[[nodiscard]]` so there's never a case where it's unintentionally unused.

Sonicadvance1: Glued on `[[nodiscard]]`

lioncash (approved):

PR #5825, opened 2026-08-13
FEXCore: prevent deadlock when branching to MAX_UINT64

javelina-pkwy:
Something I found when running my own tests is that deadlocks are possible when guest code branches to an invalid address and then has a signal handler registered for that fault which does not sigreturn.

What seems to be happening is that when we dispatch to an invalid address (ie. 0xFFFFFFFF) CompileBlock() acquires the CodeInvalidationMutex, but when we attempt to fetch the bytes at the invalid address, we take a fault. Since the guest has a signal handler registered for that fault, we go there first. In the test program below, we longjump out of the signal handler, and then map+unmap a section, which requires FEX to acquire the CodeInvalidationMutex again causing the deadlock.

I will admit that the proposed fix below might not be the most idiomatic way to solve this issue, it is just a proposal. The unit test is derived from the behavior of my own test harness, which is where I saw this bug originally.

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -1454,7 +1454,8 @@ void Decoder::DecodeInstructionsAtEntry(FEXCore::Core::InternalThreadState* Thre
     if (BlockSuccIt != BlockInfo.Blocks.end() && BlockSuccIt->Entry < NextBlockStartAddress) {
       NextBlockStartAddress = BlockSuccIt->Entry;
     }
-    LOGMAN_THROW_A_FMT(NextBlockStartAddress > RIPToDecode, "unexpected");
+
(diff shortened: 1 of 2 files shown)


Sonicadvance1:
This isn't an all encompassing solution, multiblock can immediately cause this to break again if that mutex is being held. Looks like there will need to be some more investigation, since we're supposed to be tracking mapped pages accurately.

javelina-pkwy:
@Sonicadvance1 thanks for the feedback here. I think the new changes are more correct. There are two parts here:

1. A simple wraparound bug in CheckRangeExecutable() when called with address ~0ULL and size 1 byte.
2. An update to an assert, since ~0ULL was being used as "null" in DecodeInstructionsAtEntry().

I also updated the PR title to more accurately reflect that this bug only occurs for branches to 0xFFFFFFFFFFFFFFFF (or 0xFFFFFFFF for x86 processes) and no other address.

Sonicadvance1 (approved):
Okay, this looks reasonable enough to me now.

PR #5826, opened 2026-08-16
Windows/CRT: Add printf and puts stubs

cjacek:
Fixes PE builds with VIXL disassembler enabled.


lioncash (approved):
Makes sense, since we already have fputs and fprintf stubs

PR #5828, opened 2026-08-17
Steam/CompatTool: Become more picky about configs

Sonicadvance1:
If `STEAM_COMPAT_FEX_CONFIG` is missing options, then instead of having an opinion about what those options should be, just leave them unset. This allows FEX's regular default option handling to kick in for missing configuration options.

Where previously if an option was missing from the config, it would default to boolean false, which may or may not be the default depending on option.


Sonicadvance1:
Merging just to keep the python and cpp side in sync as much as possible.

PR #5831, opened 2026-08-18
Windows: Handle 0x2c interrupt in HandleGuestException

cjacek:
Fixes a Wine exception test.


Sonicadvance1 (approved):
Oh cool. Sounds good to me.

PR #5832, opened 2026-08-19
Disk Cache initial implementation

Plagman:
Serializes code blocks to disk - only blocks coming from known regions, for now

Disabled by default, key and versioning still needs work, but works for testing


Sonicadvance1 (approved):
Yea, sure. There's a bit of cleanup here and there needed to match the rest of the code base, but nothing outrageous enough to block this from being merged.

Looks good to me random generic person I've never heard of in my life.

Plagman:
Did some cleanup:
 - Made a DiskCache namespace and removed redundant naming prefixes from classes/structs therein
 - Reordered public/private
 - Borrowed code from Mesa goes into a namespace
 - Put some consts that were ugly global defines into their appropriate classes and renamed

Plagman:
https://github.com/Plagman/FEX/commit/80d2d0281e6dae0807a0895c38d0fa70ba8b6c0f if you'd like to see just the incremental diff

Sonicadvance1 (approved):
Even better with the cleanup. Still looking good here.

PR #5834, opened 2026-08-20
FEXCore/SharedCodeBufferManager: Pivot what tracks memory allocations

Sonicadvance1:
It's soon going to change how these buffers are managed, where the CodeBuffer is going to manage its own allocations soon once it changes over to the bitmap allocator. Additionally the Manager class is actually going to do proper management, pooling, and invalidation handling.

Split the task preemptively before we switch to the bitmap allocator to reduce churn. A little change in the CodeCache where it needs to query the codebuffer directly rather than the context, but fairly safe.

Shouldn't be any real behaviour change.

### DIFF
--- a/FEXCore/Source/Interface/Core/SharedCodeBufferManager.h
+++ b/FEXCore/Source/Interface/Core/SharedCodeBufferManager.h
@@ -22,7 +22,11 @@ struct GuestToHostMap;
 namespace FEXCore::CPU {
 struct CodeBuffer {
   uint8_t* Ptr;
+  uint8_t* CodeBufferEnd;
   size_t AllocatedSize; // including guard page; see UsableSize()
+
+  // Code buffer allocation information.
+  std::atomic<uint8_t*> CodeBufferOffset {};
 
   fextl::unique_ptr<GuestToHostMap> LookupCache;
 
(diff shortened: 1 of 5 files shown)


lioncash (approved):
Seems p. straightforward

FEXCore/Source/Interface/Core/SharedCodeBufferManager.h
```diff
@@ -98,14 +73,38 @@ class SharedCodeBufferManager {
...
+
+  // Allocate a new CodeBuffer with geometric growth up to an internal maximum.
+  // Subsequent calls to GetLatest will point to the returned buffer.
+  fextl::shared_ptr<CodeBuffer> StartLargerCodeBuffer();
+
+  // Allocate a new CodeBuffer with maximum internal size.
+  // Subsequent calls to GetLatest will point to the returned buffer.
+  fextl::shared_ptr<CodeBuffer> StartMaximalCodeBuffer();
 
   virtual void OnCodeBufferAllocated(const std::shared_ptr<CodeBuffer>&) {};
```
lioncash, line 111: Oh, one thing I guess. Should this also have a defaulted virtual dtor so there's always well-defined destruction semantics?

lioncash (approved):

PR #5835, opened 2026-08-21
DiskCache: offload Store to a WorkQueueThread

Plagman:
And some supporting helper commits


FEXCore/include/FEXCore/Utils/File.h
```diff
@@ -43,7 +45,7 @@ class File final {
 
   File() = default;
 
-  File(const char* Filepath, FileModes Modes) {
+  File(const char* Filepath, FileModes Modes, bool Seekable = true) {
```
lioncash: Seems like Seekable can be assigned in the initializer list instead of at the end like:

```cpp
File(const char* Filepath, FileModes Modes, bool Seekable = true)
    : Seekable{Seekable}
{
   // ...
}

lioncash (commented):
Just one minor nit, but looks good otherwise

FEXCore/include/FEXCore/Utils/File.h
Plagman: Done, and rebased while I was at it.

lioncash (approved):

PR #5836, opened 2026-08-21
Windows: Handle interrupt 3 in HandleGuestException

cjacek:
The current code handles only 1-byte 0xcc form of breakpoint, this PR adds support for 2-byte 0xcd,0x03. Fixes more Wine exception tests.


lioncash (approved):
LGTM

Sonicadvance1 (approved):

PR #5837, opened 2026-08-21
SharedCodeBufferManager: Leak less internal details about implementation

Sonicadvance1:
The various places that were using the CodeBuffer object were using internal implementation details that are changing as we move over to a bitmap allocator.

Preempt this by hiding some of the implementation details early without changing behaviour. `GetBufferBase` is still technically leaking some of the internal details, but it needs changes around how relocations are being handled and how the disk cache validation works in order to handle that right now.

Should be no functional change.


lioncash (approved):
Pretty straightforward API drop-in

PR #5840, opened 2026-08-22
Avoid double offset relocations

Sonicadvance1:
FEX Relocations now live at an offset from the `CodeData.BlockBegin` of the code. Regardless of where the relocation moves to, it should always be relative to that address. This is what makes it PIC compatible.

We were preemptively offsetting the relocation location to be relative to the memory base in the buffer, which is unnecessary and causes code caching to basically relocate twice to get the real location.

So in JIT.cpp, stop relocating the offsets, they're already relative to `BlockBegin`, which is offset 0.

Then when storing the relocation, stop relocating offsets AGAIN because it's already relative to the code being serialized.

Then when loading the relocations in `CodeCache::ApplyCodeRelocations` stop relocating offsets YET ANOTHER TIME.

All this is to say that relocation offsets are already PIC and relative to offset 0, so we don't need to do it three times.

### DIFF
--- a/FEXCore/Source/Interface/Context/Context.h
+++ b/FEXCore/Source/Interface/Context/Context.h
@@ -122,14 +122,12 @@ public:
    * Note that FEX relocations are unrelated to ELF/PE relocations.
    *
    * @param GuestDelta Guest address offset to apply to RIP-relative data
-   * @param RelocationOffset Offset to subtract from relocation target offsets
    * @param ForStorage True for serializing data (producing deterministic output); false for de-serializing it (resolving dynamic symbols)
    *
    * @return Returns true on success
    */
   [[nodiscard]]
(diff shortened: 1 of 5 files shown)


Plagman:
Confirmed it works with:

```
  static inline bool IsRelocationInBlock(const FEXCore::CPU::Relocation& Reloc, const CPU::CPUBackend::CompiledCode& CompiledCode) {
    return Reloc.Header.Offset < CompiledCode.Size;
  }
```
  
That being said, instead I think we don't `Store` if `CodeCache.IsGeneratingCache`, and instead get rid of `IsRelocationInBlock` fully and all three of its call sites.

Sonicadvance1:
Confirmed that with the `IsRelocationInBlock` that things are working here as well. Went through and removed the check and just skip the `DiskCache::Store` if `CodeCache.IsGeneratingCache` is enabled.

lioncash (approved):

PR #5842, opened 2026-08-22
unittests/ASM: Test CF of 16-bit SHLD with imm shift of 16

OFFTKP:


Sonicadvance1:
Thanks for bringing this up! It's now fixed.

PR #5844, opened 2026-08-23
DiskCache: add some SHM stats

Plagman:
I have a WTF local change that consumes that I could also push, and we could put something in MangoHUD as well. Is align OK here?

I'd imagine at some point we can replace some of the other cache stats but for now they're two separate things.

### DIFF
--- a/FEXCore/include/FEXCore/Utils/SHMStats.h
+++ b/FEXCore/include/FEXCore/Utils/SHMStats.h
@@ -70,6 +70,11 @@ struct ThreadStats {
   uint64_t AccumulatedCacheWriteLockTime;
 
   uint64_t AccumulatedJITCount;
+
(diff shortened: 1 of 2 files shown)


FEXCore/include/FEXCore/Utils/SHMStats.h
```diff
@@ -52,7 +52,7 @@ struct ThreadStatsHeader {
   uint32_t Pad;
 };
 
-struct ThreadStats {
+struct alignas(16) ThreadStats {
```
Sonicadvance1: Instead of `alignas(16)` just stick a pad member at the very end. We/I prefer explicit padding members and it'll remind us that can be removed later.

Sonicadvance1 (approved):

lioncash (approved):

PR #5845, opened 2026-08-23
DiskCache: make Writer thread low-priority

Plagman:
I could have an enum there or some other stronger-typed flag but I don't really expect more movement on that front, could do though, up to you all.


Plagman:
I plumb it as a member and public function to make sure the priority is guaranteed set _before_ the thread starts chewing through any of its workload, BTW.

Sonicadvance1 (approved):
Good 'nuff for me.

lioncash (approved):

PR #5852, opened 2026-08-25
DiskCache: bucket/key bookkeeping

Plagman:
Compute a bucket hash and use it in the cache path to avoid grouping entries that will never make sense together. Don't trust the path, though, and also lace it into the keys themselves, so that eg. a RO cache miss can never turn into corruption.

Keep a readable metadata entry at the beginning of the cache, with readable version, bitness and serialized config.

### DIFF
--- a/FEXCore/Source/Interface/Core/DiskCache.cpp
+++ b/FEXCore/Source/Interface/Core/DiskCache.cpp
@@ -1,5 +1,7 @@
 // SPDX-License-Identifier: MIT
 
+#include "FEXCore/Config/Config.h"
+#include "FEXCore/fextl/string.h"
 #include "FEXHeaderUtils/Filesystem.h"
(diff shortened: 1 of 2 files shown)


Plagman:
Will also add https://github.com/FEX-Emu/FEX/pull/5851 into BucketKey once it lands

Plagman:
Added a change to make the keys printable as well, as the tooling ecosystem for FOZ will expect that. Only the metadata entry isn't, if it breaks some tools we maybe can use 0s for the key but no sign of that so far.

Sonicadvance1 (approved):

lioncash:
5851 is landed, so the host features can be bucketed too

Plagman:
Marking WIP until I catch up to adding host stuff in there

lioncash (approved):

Issue #5336, opened 2026-02-27
steamwebhelper inconsistently crashes with multiblock enabled


Sonicadvance1:
This isn't a regression in FEX, but seemingly a change on the Steam side generating different code and causing multiblock to break.

Disabling multiblock seems to resolve the issue, but it's a bit inconsistent with how frequently it crashes.

Typical logs are
```
SP Desktop_uid0-'Steam': WARNING: cancelling DoClose() for bHideOnClose, but using Window.Close() is preferable
<...>
[2026-02-26 17:10:13] Browser - launching child process with: /proc/self/exe --type=renderer --crashpad-handler-pid=1696900 --enable-crash-reporter=, --change-stack-guard-on-fork=enable --enable-chrome-runtime --user-data-dir=/home/ryanh/steam_arm64/config/htmlcache --user-agent-product=Valve Steam Client --buildid=0 --steamid=0 --valve-initial-threadpool-size=188
[2026-02-26 17:10:13] Webhelper: An X Error occurred on display 0x7fff3410d0d0
[2026-02-26 17:10:13] X Error of failed request:  BadDamage (invalid Damage parameter)
<...>
```
The X errors are a red herring as this is just a side-effect of chrome killing the process and SDL not understanding the window going away.

Bisecting this isn't possible, as it's not a FEX regression, it has instead been a lurking bug in FEX's multiblock for many months.
Trying to figure out exactly what we're breaking in multiblock is going to be hard.

MitchellAugustin:
> seemingly a change on the Steam side generating different code and causing multiblock to break

@Sonicadvance1 I have a few interesting observations about this issue, stemming from how it interacts with the Steam snap, that I hope might be of help in pinpointing where this is coming from:

- It still does not reproduce at all on the version of the Steam snap that we released on 9 April 2026 (built with whichever FEX release was latest in the PPA at that time, either 2603 or 2604). This is interesting to me since, while the Steam deb itself will not have updated for any of these users, the Steam client's built-in auto-updater will have updated various components in the Steam home directory many times since then (AFAIK).
- It _does_ reproduce with all versions of the Steam snap that I have tried to build since the release of FEX 2605. This is interesting to me only because of how our build methodology works [0]
[...]

PR #5856, opened 2026-08-26
LookupCache: Clear L1 entries on dynamic cache growth

Hojun-Cho:
With the dynamic L1 cache, a translation invalidated between a grow and a later shrink  becomes reachable again, and FEX runs stale guest code.

The L1 index is `Address & L1PointerMask`. When the mask grows, an address with the new bit set maps to a different slot. The old entry remains, but `InvalidateCache` checks only the slot selected by the current mask. If the cache later shrinks, the old entry becomes reachable again:

```
call A        8192 entries, mask 0x1FFF        A & 0x1FFF = 7936
              +------------------------+
              |                     [A]|
              +------------------------+
                                    7936


grow          16384 entries, mask 0x3FFF       A & 0x3FFF = 16128 = 7936 + 8192 (bit  13)   
              +------------------------+------------------------+
              |                     [A]|                     [ ]|
              +------------------------+------------------------+
                                    7936                    16128
                  A's slot is no longer here      A's slot is here now     

rewrite A     guest overwrites A's code; InvalidateCache(A) looks only at L1[16128]
              +------------------------+------------------------+
              |                     [A]|                     [ ]|
              +------------------------+------------------------+
                            still there              nothing here, nothing cleared   

shrink        [8192, MAX_L1_ENTRIES) madvised away; mask is 0x1FFF again
              +------------------------+
              |                     [A]|  <- A's slot again: hit, old translation runs  
              +------------------------+
```

The shrink branch of UpdateDynamicL1Stats clears `[new, MAX)`; the grow branch now clears `[0, old)` ([old, new) is already zero).  This also evicts valid entries.
It avoids a full-table rehash at the cost of a burst of misses after growth.

### Testing

- The reproducer below takes about 15 seconds. It fails on FEX-2604 (Fedora package) and on main at 4ed80fd07; it passes with this change.   
- Without this change, it also passes with `FEX_DYNAMICL1CACHE=0` or `FEX_DYNAMICL1CACHEDECREASECOUNTHEURISTIC=0`. 
- Full ctest on c9b23eb0e: the same 48 pre-existing failures with and without the change.
- The code dates from f44cd9c54 (FEX-2511); 86acfb35a (FEX-2604) enabled it by default.
- Test environment: Apple M1 Pro, Fedora Asahi 44, muvm.

```
$ x86_64-linux-gnu-gcc -static -o repro repro.c
$ FEX ./repro                      # before
FAIL: target was rewritten to return 2 but returned 1
$ FEX ./repro                      # after
PASS: target was rewritten to return 2 and returned 2
```

<details><summary>repro.c</summary>

```c
#include <stdio.h>
#include <string.h>
#include <sys/mman.h>
#include <time.h>
#include <unistd.h>

typedef int (*fn)(void);

/* FEX: code rewritten between a grow and a shrink of the dynamic L1 cache runs stale */
static volatile fn target, miss0, miss1;	/* indirect calls: every call goes through the L1 lookup */

static void
emit(fn f, int v)	/* mov eax, v; ret */
{
	unsigned char op[] = {0xb8, v, 0, 0, 0, 0xc3};

	memcpy((void*)f, op, sizeof op);
}

int
main(void)
{
	char* p;
	time_t t;
	int r;

	p = mmap(NULL, 0x400000, PROT_READ|PROT_WRITE|PROT_EXEC, MAP_PRIVATE|MAP_ANONYMOUS, -1, 0);
	if (p == MAP_FAILED) {
		perror("mmap");
		return 2;
	}
	p = (char*)(((unsigned long)p + 0x1FFFFF) & ~0x1FFFFFUL);
	target = (fn)(p + 0x3F00);	/* bit 13: another L1 slot once the table grows past 8192 */
	miss0 = (fn)(p + 0x800);	/* bit 20 apart: the same L1 slot at every size */
	miss1 = (fn)(p + 0x100800);
	emit(target, 1);
	emit(miss0, 0);
	emit(miss1, 0);

	target();
	t = time(NULL) + 3;
	while (time(NULL) < t) {	/* >250 misses/s: L1 grows */
		miss0();
		miss1();
	}
	emit(target, 2);
	t = time(NULL) + 12;
	while (time(NULL) < t) {	/* <50/s: L1 shrinks back */
		miss0();
		miss1();
		usleep(100000);
	}
	r = target();

	if (r != 2) {
		printf("FAIL: target was rewritten to return 2 but returned %d\n", r);
		return 1;
	}
	printf("PASS: target was rewritten to return 2 and returned 2\n");
	return 0;
}
```
</details>

### DIFF
--- a/FEXCore/Source/Interface/Core/LookupCache.h
+++ b/FEXCore/Source/Interface/Core/LookupCache.h
@@ -242,6 +242,9 @@ public:
 
       if (AveragePerSecond >= DynamicL1CacheIncreaseCountHeuristic()) {
         if (CurrentL1Entries < MAX_L1_ENTRIES) {
+          // Entries whose address has the new mask bit set would be unreachable by InvalidateCache
+          FEXCore::Allocator::VirtualDontNeed(reinterpret_cast<void*>(L1Pointer), CurrentL1Entries * sizeof(LookupCacheEntry), false);
+
           CurrentL1Entries <<= 1;
           L1PointerMask = CurrentL1Entries - 1;
 


Hojun-Cho:
BTW, Found this chasing steamwebhelper crashes on Fedora Asahi (#5336).  
No crashes with `FEX_DYNAMICL1CACHEDECREASECOUNTHEURISTIC=0`, so I looked at the dynamic L1.

javelina-pkwy:
I think it would be useful to onboard repro.c as a proper CI test in the Linux test suite.

Hojun-Cho:
@javelina-pkwy It runs 15s as written. 
The sample period is 1s, so the busy phase grows the table three times and it takes three shrinks to come back. that is the 12s tail.
Holding the busy phase to one grow brings it to about 4s.

Sonicadvance1 (approved):
Oh jeez, that's pretty bad. It's going in.

MitchellAugustin:
This does not appear to solve the client crashing issue for me completely. It does seem to crash less, but still crashes after leaving the menu idle for about an hour, then interacting with it. (Might also happen sooner too, but just tested after the hour wait so far). Tested on 2 devices.

The other workaround discussed in #5336 has so far not had this issue, after 6+ hours of idle menu and then about 10 minutes of interacting with it, which reliably repro'd the issue in the past:

```
mitchell@spark-prod:~/FEX$ cat ~/snap/steam/common/fex_config/AppConfig/client.json
{
  "Config": {
    "HideHypervisorBit": "1",
    "DynamicL1CacheDecreaseCountHeuristic": "0"
  }
}
mitchell@spark-prod:~/FEX$ cat ~/snap/steam/common/fex_config/AppConfig/steamwebhelper.json
{
  "Comment": "Bypasses libGL's glX and instead sends GLX requests directly via xcb",
  "ThunksDB": {
    "GL": 0
  },
  "Config": {
    "DynamicL1CacheDecreaseCountHeuristic": "0"
  }
}
```

PR #5860, opened 2026-08-27
DiskCache: clean up bucket key building a bit

Plagman:
Less manual math like that.

I meant to add this chunk to the previous MR but forgot to commit before re-pushing, oops!


Sonicadvance1 (approved):
:feelsgoodman:

Plagman:
I wanted to put the config bytes inline there with a dynamic array size in the inline structure but C++ is provably inferior to C apparently!

lioncash (approved):

PR #5861, opened 2026-08-27
DiskCache: use memory-mapped IO for reads when possible

Plagman:
On Wine, use a unixlib call to get a quality mapping that can track a growing file.

### DIFF
--- a/FEXCore/include/FEXCore/Core/DiskCacheFileMapper.h
+++ b/FEXCore/include/FEXCore/Core/DiskCacheFileMapper.h
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "FEXCore/Utils/CompilerDefs.h"
+#include "FEXCore/Utils/File.h"
+
+namespace FEXCore::DiskCache {
+using FileMapperFunc = void* (*)(FEXCore::File::File::FileHandleType Handle, uint64_t MapSize);
+
+FEX_DEFAULT_VISIBILITY void SetFileMapper(FileMapperFunc Func);
+} // namespace FEXCore::DiskCache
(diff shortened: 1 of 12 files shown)


Source/Windows/UnixLib/FEXUnixLib.h
```diff
@@ -51,3 +52,10 @@ struct FEXUnixLib_GetSHMStatsVMA {
   uint32_t MaxSize;
 };
 static_assert(sizeof(FEXUnixLib_GetSHMStatsVMA) == 16);
+
+struct FEXUnixLib_MapFile {
+  int32_t FD;
```
Sonicadvance1, line 57: Do we want to stick the int32_t element at the end so we don't end up with any packing shenanigans? Changing it from 24 bytes to 20.

FEXCore/include/FEXCore/Core/DiskCacheFileMapper.h
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT
+#pragma once
+
+#include "FEXCore/Utils/File.h"
+
+namespace FEXCore::DiskCache {
+using FileMapperFunc = void* (*)(FEXCore::File::File::FileHandleType Handle, uint64_t MapSize);
+
+void SetFileMapper(FileMapperFunc Func);
```
Sonicadvance1: Weird, I would expect this to require `FEX_DEFAULT_VISIBILITY`. I think we messed up our default visibility somehow.

Sonicadvance1 (approved):

Source/Windows/UnixLib/FEXUnixLib.h
Plagman: It still wants it to be 24 like that, apparently.

Sonicadvance1: Yolo for tail padding I guess. oh well.

PR #5862, opened 2026-08-27
Fix pf2iw to saturate on 32-to-16-bit truncation

simon902:
The `pf2iw` instruction truncates the 32-bit integers to 16-bit signed integers.
When a value falls outside the 16-bit range, it is saturated to the smallest or largest 16-bit value instead.

FEX used `_VUnZip`, which correctly truncates but performs no saturation.
This PR instead uses `_VSQXTN`, which truncates and saturates to 16-bit in one step.


Sonicadvance1 (approved):
Sure, let's yolo some 3DNow! instructions. Eventually I'll run the unittests on my phenom again.

lioncash (approved):

PR #5865, opened 2026-08-27
FEXCore/Config:  Annotate all config options that can affect codegen

Sonicadvance1:
Instead serializing the world, allow the config to be data driven. A couple host feature options aren't serialized as they get explained elsewhere.


Sonicadvance1:
@Plagman For you

lioncash (approved):

PR #5867, opened 2026-08-28
Clear C1 for fcomi

simon902:
Clear the `C1` FPU flag for `fcomi` (per the SDM).


lioncash:
Needs a rebase

simon902:
Rebased and updated InstCountCI.

lioncash:
hostrunner failure seems legitimate

simon902:
Yes. I checked this on real x86 hardware: For `FCOMI/FCOMIP/FUCOMI/FUCOMIP` C1 is actually left untouched by these instructions and not cleared to 0.
However, the SDM states under `FPU Flags Affected` that `C1 Set to 0` for this instruction group.
I guess its a documentation artifact likely carried over from the FCOM/FUCOM entries which do genuinely clear C1.

Given that, should I close the PR and open a new one with the ASM test from here but fixed to assert the real unchanged C1 behavior?

lioncash:
That would likely be best. Particularly, the test should also have a comment documenting the discrepancy. But also on that note, you should also check and ensure that the behavior is described similarly in the AMD64 Architecture Programming Manual as well, and not solely the Intel SDM.

simon902:
Checked the AMD manual as well. It doesn't state an unconditional clear like the Intel SDM does and instead the entry for FCOMI/FCOMIP/FUCOMI/FUCOMIP reads:

> C1: 0, x87 stack underflow, if an x87 register stack fault was detected.

I tested this case (i.e., using a register operand whose tag is empty in FTW) on an Intel and AMD CPU and for both it cleared C1 to 0.
For the other case where both register operands have a valid tag in FTW, both CPUs left C1 unchanged.

PR #5870, opened 2026-08-28
External: Update fmt to upstream master to fix building with LLVM 23

cjacek:
Includes https://github.com/fmtlib/fmt/pull/4868.


cjacek:
It looks like recent fmt started using try/catch unconditionally, which breaks when used with `-fno-exceptions`. Interestingly, Clang 21 and newer is fine with it, which is why mingw builds work. It will need an additional upstream first.

cjacek:
Upstream PR: https://github.com/fmtlib/fmt/pull/4904

cjacek:
The fix for older Clang was merged upstream, I updated the PR to include that fix.

Sonicadvance1 (approved):

PR #5873, opened 2026-08-28
Try to honor mmap(address_hint, ..., MAP_32BIT)

### LINKED ISSUE #5872: mmap(address_hint, ..., MAP_32BIT) without MAP_FIXED never honors hint
The issue was observed when running x64 [sbcl](https://www.sbcl.org/) under emulation.
```
$ sbcl --version
mmap: wanted 1048576 bytes at 0x50000000, actually mapped at 0x7feff000
Dump of /proc/self/maps:
...
```
sbcl calls mmap(0x50000000, ..., MAP_32BIT), see
https://github.com/sbcl/sbcl/blob/master/src/compiler/x86-64/parms.lisp#L123
https://github.com/sbcl/sbcl/blob/master/src/runtime/linux-mman.c#L86

mcfi:
FEX should still try to honor the address hint for mmap callers that specify MAP_32BIT but not MAP_FIXED.

Fix #5872


lioncash:
Feel free to rebase this and the other draft PR on top of main. Config generation issues should be fixed now.

PR #5875, opened 2026-08-28
Config: Fix assertion being hit in config generation

lioncash:
Looks like we had one config option accidentally slip through


Sonicadvance1 (approved):
Woop.

PR #5876, opened 2026-08-28
Add SMCChecks=mtrack_proactive to unprotect code pages before read

### LINKED ISSUE #5874: SMCChecks=mtrack does not track code writes by the kernel through read-like syscalls
x64 [racket](https://racket-lang.org/) crashes under emulation:
```
$ racket hello.rkt
Error: error reading from ~a
("petite")
Aborted (core dumped)
```
Racket CS embeds its petite, scheme and racket.so boot images as sections in its own executable and hands Chez an fd plus an offset region, so the images are consumed with lseek/read on that fd. Chez's unbuffered fasl path reads code objects straight into the space_code segments it allocates for them, and those segments are recycled across the three boot files.

Under SMCChecks=mtrack FEX write-protects a guest page once it has translated code from it and relies on SIGSEGV to invalidate and unprotect. A host kernel write into guest memory cannot raise that signal, so the kernel reports EFAULT instead and Chez sees a short boot image.

mcfi:
Under SMCChecks=mtrack, FEX write-protects a guest page once it has translated code from it and relies on SIGSEGV to invalidate and unprotect. A host kernel write into guest memory (e.g., read syscall) does not raise that signal, so the syscall reports EFAULT and the program fails.

This problem could be solved with a co-operative kernel, which means that the kernel needs to be enlighted with FEX and can deliver a message to FEX when read hits EFAULT, then that would give FEX a chance to unprotect and allow the kernel to retry. However, this is impossible on vanilla kernel.

Considering that issuing read-like syscalls to modify code is rarely done by applications, this PP adds a new SMCCheck called mtrack_proactive. When the mode is selected, FEX always checks if a read's range overlaps protected code ranges and if so, it unprotects first. This might have negative perf impact if an application issues lots of reads, but compared to SMCChecks=full, this is most likely faster.

Fix #5874

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -468,7 +468,8 @@
           "Checks code for modification before execution.",
           "\tnone: No checks",
           "\tmtrack: Page tracking based invalidation (default)",
-          "\tfull: Validate code before every run (slow)"
+          "\tfull: Validate code before every run (slow)",
+          "\tmtrack_proactive: mtrack, and unprotect guest buffers before the kernel writes to them"
         ]
       },
       "TSOEnabled": {
@@ -646,4 +647,3 @@
     }
   }
 }
-
--- a/FEXCore/include/FEXCore/Config/Config.h
+++ b/FEXCore/include/FEXCore/Config/Config.h
@@ -28,6 +28,8 @@ namespace Handler {
       return "1";
     } else if (Value == "full") {
       return "2";
+    } else if (Value == "mtrack_proactive") {
+      return "3";
     }
     return "0";
   }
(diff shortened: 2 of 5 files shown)


mcfi:
@Sonicadvance1 , this PR seems hacky, but it's practical. I don't know the taste of FEX authors, so I'm not sure if this is an acceptable solution. Could you please enlighten? Thanks!

Sonicadvance1:
This is kind of similar to how the win32 side is handling it which is also still not correct. I have some ideas kicking around about how I'm going to improve this but I need to flesh them out. Since a cooperative invalidation with the kernel or even win32/wine side isn't a real option.

Thanks for the idea but we won't be pursuing this route. I'll be looking how to resolve this soon(tm)

PR #5886, opened 2026-08-31
Preserve previous value of IE exception flag 

simon902:
Per the SDM (8.1.3.3 x87 FPU Floating-Point Exception Flags): The x87 exception flags are "sticky" bits (once set, they remain set until explicitly cleared).

FEX did not preserve the old value of the `IE` exception flag for the `FCOMI` and `FTST` and `FIST` instructions.
This PR fixes this and adds a unittest.

I also updated InstcountCI and `FEXCore::DiskCache::FormatVersion` as i got a version mismatch for the binary cache. Not sure if this is correct.

Edit: Also added `FIST`.


lioncash:
needs a rebase

simon902:
Rebased and updated InstCountCI.

lioncash (approved):

PR #5887, opened 2026-08-31
Fix RestoreX87State to rotate MM registers by Top of Stack

simon902:
The `RestoreX87State` function wrote saved MM registers directly into the physical registers without account for the `Top` of the stack. This only worked when `Top` was equal to `0`. Otherwise the registers ended up being rotated.

This PR update  `RestoreX87State` to account for the `Top`  to match the functionality in `SaveX87State`.
Also mask off the reserved bytes above the 80-bit float matching `X87FRSTOR`.

I also updated InstcountCI and FEXCore::DiskCache::FormatVersion as i got a version mismatch for the binary cache. Not sure if this is correct.
**Note**: This also bumps the same version counter that #5886 updates.


Sonicadvance1:
Thanks for bringing up the issue! It took me a bit of time to get back to this since I wanted to create a unittest for this. I've cherry-picked your change, added the test, and rebased over in #5962 so we can get this change in. Thanks!

PR #5888, opened 2026-08-31
x87StackOptimizationPass: Mark ST(0) as empty after pop

simon902:
Popping the FPU stack should mark ST(0) as empty (see `PopRegisterStack` in the SDM).

The slow path in the x87 stack optimization pass never explicitly marked ST(0) as empty.
There was already a call to `SetX87ValidTag(0, false)` in the `OP_POPSTACKDESTROY` case.
However instructions like `fpatan` or `fyl2x` which also pop from the stack do not go through the `OP_POPSTACKDESTROY` case and hence do not empty ST(0).

This PR now marks ST(0) as empty for every pop in the `UpdateTopForPop_Slow` function.
I also updated InstcountCI and FEXCore::DiskCache::FormatVersion as i got a version mismatch for the binary cache. Not sure if this is correct.
Note: This also bumps the same version counter that #5886 and #5887 update.


lioncash:
closed in favour of #5900

PR #5891, opened 2026-09-01
DiskCache: hash guest code according to its real extents

Plagman:
Does less hashing, improves hit rate when there's data adjacent to code,
and/or when the SMC check makes us rebuild code that hasn't actually been
changed.


Plagman:
i forgot clang_format :frog:

lioncash (approved):

Sonicadvance1 (approved):

PR #5895, opened 2026-09-01
Utils/File: Properly handle move assignment/construction

### LINKED ISSUE #5894: FEX_DUMPIR=<folder> writes zero-byte files on the ARM64EC Windows build
## Summary

When `DumpIR` is set to a directory path, `IRDumper::Run` creates one file per block as expected, but every file is 0 bytes. The same build produces correct IR when the output is routed through `LogMan` instead, so the IR itself is fine — only the file write path fails.

## Environment

- `FEX-2608-158-ga6e74cd` (`a6e74cdbe6f454376e0a55ac567f73f45a8888e3`)
- Built as `libarm64ecfex.dll` via `Data/CMake/toolchain_mingw.cmake`, `MINGW_TRIPLE=arm64ec-w64-mingw32`, llvm-mingw 20260826, `RelWithDebInfo` + `-DENABLE_ASSERTIONS=True`
- Running under `proton-cachyos-11.0-arm64` on aarch64 Linux (Armada / AYN Thor)

## Reproduction

Launch any x86-64 Windows title with `FEX_DUMPIR=/some/directory`.

Result — files created with correct names, all empty:

```
-rw-r--r-- 1 armada armada 0 Sep  1 23:45 6fffe8fd0432-post.ir
-rw-r--r-- 1 armada armada 0 Sep  1 23:45 6fffe8fd07cc-post.ir
-rw-r--r-- 1 armada armada 0 Sep  1 23:45 6fffebfd0238-post.ir
...
```

This reproduced across three separate runs, at 29, 81 and 173,115 files. Not one had content.

## Where it goes wrong

In `FEXCore/Source/Interface/IR/Passes/IRDumperPass.cpp`, `DumpToFile` opens the file and `fextl::fmt::print(FD, ...)` writes to it:

```cpp
if (DumpToFile) {
  const auto fileName = fextl::fmt::format("{}/{:x}{}", DumpIR(), +HeaderOp->OriginalRIP,
                                           IR.PostRA() ? "-post.ir" : "-pre.ir");
  FD = FEXCore::File::File(fileName.c_str(),
                           FEXCore::File::FileModes::WRITE | FEXCore::File::FileModes::CREATE |
                           FEXCore::File::FileModes::TRUNCATE);
}
```

The file is created, so `FEXCore::File::File` is at least opening a handle. The `print` produces nothing. I did not narrow it further to open flags, the write call, or a missing flush before process teardown.

Forcing the same data down the `LogMan::Msg::IFmt` branch instead yields correct, complete IR in the log:

```
I 148 IR-post 0x6ffff7fd0f28:
        (%0) IRHeader %2, #0x6ffff7fd0f28, #1, #2
        (%2) CodeBlock %3, %4
                (%3 i0) BeginBlock %2
                (%8 i64) InlineConstant #0x28
                %9(R16) i64 = SubWithFlags R4, %8
                ...
@@@@@
```

So the pass, the filter and the IR serialisation all work; only the file output is broken on this platform.

## Possibly related

`OpcodeDispatcher::SetDumpIR` (`OpcodeDispatcher.h:335`) has no callers anywhere in the tree, so `ShouldDump` is permanently its `{false}` default and the `IRDumper` call sites in `Core.cpp` (lines 758 and 766) are unreachable. That looks like leftover code from a refactor rather than a bug in itself, but it made the dump path confusing to trace.

lioncash:
Ensures that whenever a file handle is transferred anywhere, that the moved from instance won't end up closing the file handle when the destructor runs.

Fixes: #5894

### DIFF
--- a/FEXCore/include/FEXCore/Utils/File.h
+++ b/FEXCore/include/FEXCore/Utils/File.h
@@ -7,6 +7,7 @@
 
 #include <chrono>
 #include <thread>
+#include <utility>
 
 #ifndef _WIN32
 #include <fcntl.h>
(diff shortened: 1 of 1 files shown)


Sonicadvance1 (approved):
Fancy

PR #5897, opened 2026-09-02
DiskCache: initial anon caching

Plagman:
Decode a few bytes in advance to get a hashable prefix to use as key.

Generate touched pages dynamically since they can be misaligned now, as the cached-hit guest code isn't necessarily in the same spot as the store was.

### DIFF
--- a/FEXCore/Source/Interface/Config/Config.json.in
+++ b/FEXCore/Source/Interface/Config/Config.json.in
@@ -168,6 +168,14 @@
           "Don't cache blocks with relocations pointing outside of any known region"
         ]
       },
+      "DiskCacheAnonCaching": {
+        "Type": "bool",
+        "Default": "true",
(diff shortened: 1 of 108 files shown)


Plagman:
Rebased and re-updated version/instcount.

lioncash (approved):

Sonicadvance1:
Spending a bit more time reviewing this, I want to make sure the pool buffer lifetime management is solid.

Plagman:
If LoadCachedCode fails we wouldn't call DelayedDisownBuffer(), that should be an easy fix but also basically never happen.

Sonicadvance1 (approved):
Alright, I reviewed and added some more buffer validation locally so I feel comfortable allowing this in now.
Definitely some cleanup to do afterwards but that's fine.

PR #5900, opened 2026-09-02
x87StackOptimizationPass: Generalize ST(0) tag invalidation on pop

simon902:
`FINCSTP` should only increment the x87 TOP pointer and not touch the tag word.
FEX's slow-path wired `FINCSTP` to the same helper used by a real stack pop which invalidates the tag of the register TOP.

Fixes it by making tag invalidation an explicit opt-in on that helper and adds an ASM unittest.

### DIFF
--- a/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
+++ b/FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
@@ -272,7 +272,7 @@ private:
   Ref LoadStackValueAtOffset_Slow(uint8_t Offset = 0);
   void StoreStackValueAtOffset_Slow(Ref Value, uint8_t Offset = 0, bool SetValid = true);
   // Update Top value in slow path for a pop
-  void UpdateTopForPop_Slow();
+  void UpdateTopForPop_Slow(bool InvalidateTag = true);
   void UpdateTopForPush_Slow();
   // Synchronizes the current simulated stack with the actual values.
   // Returns a new value for Top, that's synchronized between the simulated stack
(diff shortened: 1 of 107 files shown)


FEXCore/Source/Interface/IR/Passes/x87StackOptimizationPass.cpp
```diff
@@ -573,12 +573,16 @@ void X87StackOptimization::HandleBinopStack(IROps Op64, bool VFOp64, IROps Op80,
...
 }
 
-inline void X87StackOptimization::UpdateTopForPop_Slow() {
+inline void X87StackOptimization::UpdateTopForPop_Slow(bool InvalidateTag) {
   const auto PopContainer = [](auto& container) {
     const auto begin = std::begin(container);
     std::rotate(begin, std::next(begin), std::end(container));
   };
 
+  if (InvalidateTag) {
```
lioncash, line 582: Seems like the behavior for `OP_POPSTACKDESTROY` on line 1056 can have:

```cpp
if (SlowPath) {
  SetX87ValidTag(0, false);
}
```

removed, since this is now part of the behavior that `StackPop()` forwards into.

lioncash:
#5888 can likely be closed in favor of this one (moving its unit tests into this change), since this PR essentially just expands on that behavior

simon902:
Done. Removed the duplicated `SetX87ValidTag(0, false)` in `OP_POPSTACKDESTROY` and since it only became redundant because of the InvalidateTag change, I folded it into that same commit.
Also moved the unittests over from #5888 and regenerated InstcountCI.

lioncash (approved):

PR #5903, opened 2026-09-02
JIT: Assert out when debug features are enabled when caching is also enabled

Sonicadvance1:
These operations are mutually exclusive and can't be used with each other. Assert on them to ensure it doesn't occure.


Sonicadvance1:
Apparently the mono hack used this, so I need to actually figure out how to support this

PR #5904, opened 2026-09-02
Faster translation for *PMULHRSW

javelina-pkwy:
This instruction is in the hotloop for a couple of workloads in Geekbench 6.7.1

Here are some before and after results, as tested on X2 Elite:

| Workload | main | PR | % delta |
|---|---:|---:|---:|
| Background Blur | 808 | 1257 | +55.6% |
| Photo Filter | 1514 | 1837 | +21.3% |
| Structure from Motion | 1144 | 1596 | +39.5% |

The other workloads did not change outside of noise. These results were generated with ``--iterations 50`` to also reduce measurement noise.

Also a question: for changes like this that are not functional changes, just performance, will we be bumping the diskcache version every time? Or would it be better to do that opportunistically when there is a change that is really important?


Sonicadvance1:
> Also a question: for changes like this that are not functional changes, just performance, will we be bumping the diskcache version every time? Or would it be better to do that opportunistically when there is a change that is really important?

For now we're going to be incrementing the diskcache version for every change as caching is not really "shipped". Keeping everything together in lockstep is easier than picking and choosing right now.

Sonicadvance1:
I would have expected this PR to hit the diskcache version check and require incrementing the number. Please make sure to do that.

javelina-pkwy:
@Sonicadvance1 it did, that's why I reached out for clarification. I will bump the diskcache version and update the JSONs shortly.

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
```diff
@@ -3786,32 +3786,14 @@ void OpDispatchBuilder::VPMULHWOp(OpcodeArgs, bool Signed) {
...
 Ref OpDispatchBuilder::PMULHRSWOpImpl(OpSize Size, Ref Src1, Ref Src2) {
-  Ref Res {};
   if (Size == OpSize::i64Bit) {
     // Implementation is more efficient for 8byte registers
-    Res = _VSMull(Size << 1, OpSize::i16Bit, Src1, Src2);
-    Res = _VSShrI(Size << 1, OpSize::i32Bit, Res, 14);
-    auto OneVector = _VectorImm(Size << 1, OpSize::i32Bit, 1);
-    Res = _VAdd(Size << 1, OpSize::i32Bit, Res, OneVector);
-    return _VUShrNI(Size << 1, OpSize::i32Bit, Res, 1);
+    Ref Res = _VSMull(Size << 1, OpSize::i16Bit, Src1, Src2);
```
Sonicadvance1, line 3791: Got some comments from the peanut gallery.
Wondering if this operation could potentially use `SQRDMULH`.
Haven't had a chance to review myself. Maybe there is some edge case that matters?

Sonicadvance1: I think I know the answer to this is saturating behaviour is different, but haven't confirmed :)

javelina-pkwy: I'll take a look at this now, but I think this will behave differently in the case of overflow in the 16 bit intermediate values. I can play with this for a bit now while it's on our minds.

Sonicadvance1:
A1A results:
Before:
```
Test, Total Cycles, Iterations, Cycles Average, Iter Time Average, iterations/Second
vpmulhrsw_128bit, 585321820, 250000000, 2.34, 2.34, 427115462.74
vpmulhrsw_256bit, 587319820, 250000000, 2.35, 2.35, 425662461.04
```

After:
```
Test, Total Cycles, Iterations, Cycles Average, Iter Time Average, iterations/Second
vpmulhrsw_128bit, 707388040, 750000000, 0.94, 0.94, 1060238451.30
vpmulhrsw_256bit, 707577840, 750000000, 0.94, 0.94, 1059954053.96
```

TR7000 for comparison:
```
Test, Total Cycles, Iterations, Cycles Average, Iter Time Average, iterations/Second
vpmulhrsw_128bit, 1328349775, 5450000000, 0.24, 0.10, 10241701964.38
vpmulhrsw_256bit, 1330349800, 5450000000, 0.24, 0.10, 10226304765.86
```

Pretty nice uplift.

FEXCore/Source/Interface/Core/OpcodeDispatcher/Vector.cpp
javelina-pkwy: Some results. Emitting just SQRDMULH is _much_ faster than the implementation in this PR. For example, Background Blur is another 35% faster than the PR.

But it is wrong for a single input combination. Multiplying 0x8000 by 0x8000 returns 0x8000 on real hardware, and 0x7FFF when implemented with SQRDMULH. I wrote a small test program to try all 2^32 input combinations, and this is the only mismatch.

With all that said, I don't know if any game/app cares about this edge case, so it's hard to say if the faster implementation is worth the compat risk.

Sonicadvance1: Nah, let's just to it correctly. Also good to have unittests test that edge case if we don't already.

Maybe ARM ltd will be cool enough to give us a comparable instruction in the future to gain 35% perf.

javelina-pkwy: Existing unittests do catch this error already, fortunately.

javelina-pkwy:
Shoot, accidentally bumped a submodule. Will fix.

lioncash (approved):

PR #5905, opened 2026-09-02
FEXCore: Support ThreadRemoveCodeEntry as relocatable

Sonicadvance1:
Frontend just needs to generate a relocatable entry and pass it to the
backend.

DiskCache version needs to be bumped even though instcountCI didn't change.


lioncash (approved):

lioncash:
Needs version number fiddling to be updated, but lgtm otherwise

Sonicadvance1:
Rebased.

PR #5909, opened 2026-09-03
Mark both operands valid in the tag word for FXCH

simon902:
FXCH's slow path swapped register values but never updated the FPU tag word, so a register that started empty incorrectly stayed tagged empty after the swap.

### DIFF
--- a/unittests/ASM/FEX_bugs/fxch_tag_word.asm
+++ b/unittests/ASM/FEX_bugs/fxch_tag_word.asm
@@ -1,0 +1,40 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x3",
+    "RBX": "0x81"
+  }
+}
+%endif
+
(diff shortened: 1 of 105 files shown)


unittests/ASM/FEX_bugs/fxch_tag_word.asm
```diff
@@ -0,0 +1,26 @@
+%ifdef CONFIG
+{
+  "RegData": {
+    "RAX": "0x3"
+  }
+}
+%endif
+
+fxrstor [rel .fxsave_data]
+fxch st0, st1
```
lioncash, line 12: This test should probably also cover the case where a valid value is FXCH'ed around with an empty entry. e.g. something like:

```asm
FINIT
FLDZ
FXCH st0, st1
```

simon902:
Included the requested testcase change and rebased against main

lioncash (approved):

PR #5910, opened 2026-09-03
DiskCache: detect inline data, hash around it and patch it on Lookup

Plagman:
Initial framework with one common pattern seen in Mono that greatly improves hit rate there, but easy enough to expand on.

New validation mode will not load cache on successful Lookup and compare with live JIT output, and spew things like that: 

I DiskCache: lookup guest hash mismatch key=anon-f7930f9265341698 gsize=1275, ndiff=16, first diff: offset=83 live=[2a f3 66 0f 2e fe 0f 85 97 ee ff ff 0f 8a 91ee] cached=[2a f3 66 0f 2e fe 0f 85 81 62 f1 03 0f 8a 7b 62]

Examining those diff sites will inform what to add next to DetectDataMasks.

Test methodology: Red HK to in-game (slash first bush)

Without:
First run: 30% hit rate, 134M cache
Second run: 56% hit rate. 178M cache, hitch on slashing first bush
With:
First run: 46% hit rate, 129M cache
Second run: 96% hit rate, 138M cache, less hitch on slashing first bush

Based on https://github.com/FEX-Emu/FEX/pull/5897, will rebase / update CI when it gets merged (:crossed_fingers:)

### DIFF
--- a/FEXCore/Source/Interface/Core/Frontend.cpp
+++ b/FEXCore/Source/Interface/Core/Frontend.cpp
@@ -140,6 +140,8 @@ std::pair<uint64_t, bool> Decoder::ReadData(uint8_t Size) {
 
   uint64_t Res = 0;
   uint64_t Address = reinterpret_cast<uint64_t>(InstStream.InstStream + InstructionSize);
+  LastFieldReadOffset = (uint8_t)InstructionSize;
+  LastFieldReadSize = Size;
   if (CheckRangeExecutable(Address, Size)) {
     std::memcpy(&Res, &InstStream.AdjustedInstStream[InstructionSize], Size);
   } else {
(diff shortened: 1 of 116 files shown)


Plagman:
I split off the validation commit to https://github.com/FEX-Emu/FEX/pull/5918, and will rebase this MR once that lands and add InstCountCI. Still good for review.

Plagman:
Rebased with some fixups based on subsequent work, and with matching instcountCI, ready for more eyes now.

FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -1386,6 +1393,50 @@ bool Decoder::CheckIfCacheable(FEXCore::Core::InternalThreadState& Thread, const
...
+    return;
+  }
+
+  // remove this if we ever fixup ValidateCode crc constant after relocations
+  if (CTX->Config.SMCChecks == FEXCore::Config::CONFIG_SMC_FULL) {
+    return;
+  }
+
+  // the masks are needed for the anon lookup prefix decodes as well, not just stores
+  if (!CTX->DiskCache.IsReadingDiskCache() && !CTX->DiskCache.IsWritingDiskCache()) {
```
Sonicadvance1: Once these checks get lifted as we discussed then I think this PR will be fine to go.

Sonicadvance1 (approved):

PR #5911, opened 2026-09-03
Windows: Fix syscall regression

javelina-pkwy:
Fixes a regression introduced in https://github.com/FEX-Emu/FEX/pull/5889, causing crashes early in process start when using FEX on Windows.

I suspect that removing these lines in that PR was an accident, but I am not sure exactly what the intention was.


Sonicadvance1 (approved):
Oops. I think I was having some issues with vim freezing and it caught this file somehow.

lioncash (approved):

Issue #5912, opened 2026-09-03
FileID needs some work on Linux


Plagman:
Would be good to bring it up to parity with the Windows implementation, using something like GNU .build-id if it's available, and falling back to some objective measure of the file itself (like a hash of its timestamp/size/base name) that doesn't include a hash of its full path.

Right now the ELF parsing we do happens after the FileID is generated, so even if .build-id parsing was added there we likely need to reorganize things a bit.

Plagman:
@lioncash in case you're interested 👀 

Adding FileID to the code key for file-backed enries brings the hit rate for disk cache to pretty much 100% for 'normal' games, but I don't want to do that now as it would make keys dependent on your install path on Linux.

PR #5931, opened 2026-09-07
DiskCache: key fixes

Plagman:
Fixes Three Point Hospital. The FileId change gets Sekiro (the poster child for a 'normal' game) to 99.99% hit rate and gets in game with less than 40 JIT invocations, but will suffer from below-expected hit rates across installs until https://github.com/FEX-Emu/FEX/issues/5912 is fixed.

### DIFF
--- a/FEXCore/Source/Interface/Core/DiskCache.cpp
+++ b/FEXCore/Source/Interface/Core/DiskCache.cpp
@@ -17,6 +17,7 @@
 #include <cstdint>
 #include <cstring>
 #include <charconv>
+#include <atomic>
 
 namespace FEXCore {
 
@@ -279,7 +280,8 @@ namespace DiskCache {
     {
       std::lock_guard Guard(IndexMutex);
       if (Index.contains(Hash)) {
-        // shouldn't really happen.. assert or something?
+        // LogMan::Msg::IFmt("duplicate store {}", Hash);
+        //  shouldn't really happen.. assert or something?
         return true;
       }
     }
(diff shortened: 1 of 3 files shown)


Sonicadvance1 (approved):
Whew, spicy.

FEXCore/Source/Interface/Core/DiskCache.cpp
```diff
@@ -439,11 +441,25 @@ namespace DiskCache {
...
   }
 
-  uint64_t DiskCache::MakeBlobKey(const uint64_t CodeKey) {
-    struct {
+  uint64_t DiskCache::MakeBlobKey(Core::InternalThreadState* Thread, const uint64_t CodeKey, bool Writable, bool MonoBackpatcher) {
+    struct __attribute__((packed)) {
       uint64_t CodeKey;
       XXH128_hash_t BucketHash;
-    } BlobKeyBytes = {CodeKey, BucketHash};
+      uint8_t Flags;
```
Sonicadvance1, line 448: Does this need a DiskCache version bump since it's modifying the internal keys of the entries?

Sonicadvance1: Or no, I guess this is just a key hash, rather than a key that eventually gets taken from the cache. So it's fine.

Plagman: Yeah, the DiskCache bump is only the case where the effect on the cache can't just be described by the keys naturally rolling over from whatever change we're doing.

PR #5934, opened 2026-09-08
DiskCache: support multiple entries per lookup key

Plagman:
The LookupKey based on data available at Lookup time gives a list of possible candidate entries, which may have different guest code sizes/footprints.

If only one candidate, it's stored in-line in the map like before - if we grow past that, an additional (multi)map is allocated, sorted by footprint.

The footprint sorting lets us reduce the amount of hashes performed at Lookup to the strict minimum.

Anon entries keys are based on a limited size decoded prefix, so this is an important part of geting the best hit rate possible out of all the other work with anon keys, as there's often multiple candidates with the same prefix.

Cap the max amount of entries per bucket to limit growth.

### DIFF
--- a/FEXCore/Source/Interface/Core/DiskCache.cpp
+++ b/FEXCore/Source/Interface/Core/DiskCache.cpp
@@ -16,7 +16,6 @@
 #include "FEXCore/fextl/memory.h"
 #include <cstdint>
 #include <cstring>
-#include <charconv>
 #include <atomic>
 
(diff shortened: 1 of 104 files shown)


Sonicadvance1:
Oops, all constructors.

Sonicadvance1 (approved):
I love a self-contained PR that its blast radius is contained to a feature that isn't enabled by default.

PR #5940, opened 2026-09-09
Thunks: allow self-lookups in vkGetDeviceProcAddr and vkGetInstanceProcAddr

javelina-pkwy:
_Doom: The Dark Ages_ ships with XeSS which loads on game launch by default. XeSS attempts to query ``vkGetDeviceProcAddr()`` for ``"vkGetDeviceProcAddr"``, which at present returns ``NULL``. The Vulkan API docs suggest that instead this should return a valid function pointer: [docs](https://docs.vulkan.org/refpages/latest/refpages/source/vkGetDeviceProcAddr.html).

I am not entirely sure why they are doing this. This Vulkan doc suggests a pattern like this as an optimization technique to avoid dispatch cost: https://github.com/KhronosGroup/Vulkan-Loader/blob/main/docs/LoaderApplicationInterface.md#best-application-performance-setup

Regardless, with this fix, XeSS now works fine and _Doom: The Dark Ages_ makes it into the benchmark. While we are here, we might as well also fix the cousin function ``vkGetInstanceProcAddr``, even though XeSS doesn't seem to use it.

### DIFF
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -773,7 +773,6 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 #endif
 
     Thread->OpDispatcher->Finalize();
-
   }
 
   Thread->FrontendDecoder->DelayedDisownBuffer();
--- a/FEXCore/Source/Interface/Core/Core.cpp
+++ b/FEXCore/Source/Interface/Core/Core.cpp
@@ -774,8 +774,9 @@ ContextImpl::GenerateIR(FEXCore::Core::InternalThreadState* Thread, uint64_t Gue
 
     Thread->OpDispatcher->Finalize();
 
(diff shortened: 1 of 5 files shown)


javelina-pkwy:
Interesting, this test passed on my machines. Looking into the failure now...

Update: it didn't repro on my side because I hadn't compiled FEX with asserts. Looks like a pre-existing issue, but not sure what the deal is yet.

javelina-pkwy:
Figured it out, when ``HasCustomIRHandlers`` is true, the IR buffer was never "disowned" causing the assert. I guess none of the existing CI tests hit this path?

Sonicadvance1:
> Figured it out, when `HasCustomIRHandlers` is true, the IR buffer was never "disowned" causing the assert. I guess none of the existing CI tests hit this path?

Ah, yes that recently changed and our CI doesn't have 100% code coverage.

Sonicadvance1 (approved):

lioncash (approved):

PR #5943, opened 2026-09-10
WOW64: Use CHPEv2 for suspend and exception support

cjacek:
CHPE_V2_CPU_AREA_INFO is the mechanism used on ARM64EC to coordinate thread state, suspension and exception handling with the kernel. It's typically set up by ntdll for ARM64EC processes, but Wine uses the TEB pointer without restricting where it's coming from. Set ChpeV2CpuAreaInfo in FEX WOW64 thread initialization.

Fixes 32-bit debug events and a number of CPU context handling tests. Reduces the number of wineserver round-trips required for suspend and avoids the need for remote thread creation for inter-process suspends.


Sonicadvance1 (approved):
Certainly nothing could go wrong.

PR #5947, opened 2026-09-11
Implement ISA extension AVX-VNNI

javelina-pkwy:
This PR implements [AVX-VNNI](https://en.wikipedia.org/wiki/AVX-512#VNNI). This ISA extension came after AVX-512, but does not depend on it. This implements four opcodes in total:

``VPDPBUSD``
``VPDPBUSDS``
``VPDPWSSD``
``VPDPWSSDS``

Geekbench 6.7.1 uses ``VPDPBUSDS``, if CPUID advertises it, in the hot loop of "Object Detection" and "Photo Library" in its matrix multiplication kernel. These instructions map cleanly onto Neon instructions in FEAT_I8MM and FEAT_DotProd. For machines that don't support either of these features this PR also offers a naive fallback path, but it's quite slow, so I opted to not advertise AVX-VNNI unless one of these features are present on the machine.

| Workload         | Before | After w/ I8MM path | After w/ DotProd | After w/ naive fallback |
|------------------|-----:|------------------------:|--------------------------:|----------------------:|
| Object Detection |  422 |            981 (+132%) |               555 (+31%) |           409 (−3%) |
| Photo Library    |  982 |            1312 (+33%) |              1094 (+11%) |           972 (−1%) |

One other note is that the assembler we use encodes the AVX512 ``EVEX`` variants of these instructions by default, and I couldn't figure out how to trick it into emitting the ``VEX`` variants, so instead I encoded them by hand.


unittests/InstructionCountCI/AVX128/VEX_map2_I8MM.json
```diff
@@ -0,0 +1,189 @@
...
+    "BinaryCacheVersion": 20
+  },
+  "Instructions": {
+    "vpdpbusd xmm0, xmm1, xmm2": {
+      "x86InstructionCount": 1,
+      "ExpectedInstructionCount": 2,
+      "Comment": [
+        "Map 2 0b01 0x50 128-bit",
+        "vpdpbusd xmm0, xmm1, xmm2",
+        "NASM only emits the EVEX encoding, hand encode the VEX form"
```
Sonicadvance1, line 23: Are you sure these are hand encoded? :thonking:

Sonicadvance1: Oh, you did it in `x86Insts` json member. Nevermind I'm a derp.

Sonicadvance1 (approved):
Well, it passed in CI. Certainly nothing can go wrong here.
Luckily it doesn't seem like this is popular in games currently.

unittests/InstructionCountCI/AVX128/VEX_map2_I8MM.json
javelina-pkwy: I agree that this is perhaps a little unclear. Maybe there is a better way to express this type of thing? I'll think on it some. Thanks again for taking a look! :)

javelina-pkwy:
> Well, it passed in CI. Certainly nothing can go wrong here. Luckily it doesn't seem like this is popular in games currently.

I have heard, though not observed, that this type of thing _is_ in some games, since some games come bundled with AI SDKs these days. I'll report back if I see direct evidence of this.

unittests/InstructionCountCI/AVX128/VEX_map2_I8MM.json
Sonicadvance1: Technically the hand encoded instruction can just be in the string describing it. But as a new idiom this kind of works out regardless.

unittests/InstructionCountCI/AVX128/VEX_map2_I8MM.json
```diff
@@ -0,0 +1,189 @@
...
+    "BinaryCacheVersion": 21
+  },
+  "Instructions": {
+    "vpdpbusd xmm0, xmm1, xmm2": {
+      "x86InstructionCount": 1,
+      "ExpectedInstructionCount": 2,
+      "Comment": [
+        "Map 2 0b01 0x50 128-bit",
+        "vpdpbusd xmm0, xmm1, xmm2",
+        "NASM only emits the EVEX encoding, hand encode the VEX form"
```
Sonicadvance1, line 23: I have been told that apparently if you add `{vex}` to the front, these will choose the VEX coding instead of EVEX. A question will be if the nasm in CI is new enough.

Sonicadvance1: Looks like we still have jammy runners in CI actually, so we can't actually use that. nevermind.

PR #5949, opened 2026-09-13
DiskCache: patch a bunch of disp32s in anon code

Plagman:
And imm32 C7 movs

--
Reduces churn in NecroDancer. there's still a bunch, but there's probably a finite amount of those to handle (famous last words).


FEXCore/Source/Interface/Core/Frontend.cpp
```diff
@@ -1401,25 +1406,78 @@ void Decoder::DetectDataMasks(uint64_t OpAddress, DecodedBlocks& Block) {
   }
 
   FEXCore::X86Tables::DecodedOperand* LiteralToPatch = nullptr;
-  DataMaskType Type;
+  DataMaskType Type = DataMaskType::NONE;
 
   // mov reg,imm
-  if (DecodeInst->OP >= 0xB8 && DecodeInst->OP <= 0xBF) {
+  // somehow just OP doesn't work for C7?
```
Sonicadvance1: This is because 0xC7 is actually an escape code to the a secondary table at https://github.com/FEX-Emu/FEX/blob/main/FEXCore/Source/Interface/Core/X86Tables/PrimaryGroupTables.cpp#L199-L201

Luckily this secondary table only has a 3 instructions on that opcode. But we should probably just add a new flag to the instruction tables (like FLAGS_MODRM already has) and just tag all 9 instructions with something like `FLAGS_MOV_LITERAL_PATCHABLE` and check for that flag instead of raw ops.

Plagman:
Updated MR, split between disp32 and imm32 commits. Those together get the red-hollow-knight-get-ingame-and-whack-bush test to 99.90% hit rate on second run, with something like 150 JIT invocations.

Sonicadvance1 (approved):

lioncash (approved):

PR #5951, opened 2026-09-13
FEXCore: Removes three subflags for hardcoded registers

Sonicadvance1:
Still need to remove SF_DST_RAX but it requires more work for that one and I won't get back to this until Monday. This removes some burden in the frontend around displacement relocation and moves it to the dispatcher instead.

These hardcoded registers were kind of convenient in that the same dispatcher could handle something like `xchg rax, rbx` (0x90) and `xchg rax, rbx` (0x86 /r) in the some OpcodeDispatcher, but a little additional burden for these isn't a big deal. Speeds up the common case of the Frontend as well which is kind of nice.


Sonicadvance1:
I'm actually going to split this one in to multiple PRs, removing one flag at a time because doing them all in one swoop is a bit spicy.

lioncash:
Will close this, since it's already in the process of being split up.

PR #5953, opened 2026-09-14
JIT: Let DiskCache patch ValidateCode guest CRC (and other small fixes :angel:)

Plagman:
For CRC commit, maybe we can put crc32 helper in a clever spot shared with the other use, maybe we don't care?


Sonicadvance1 (approved):
I'll move the duplicated crc32 helper somewhere. No worries.

PR #5957, opened 2026-09-15
Frontend: Remove DST_RAX subflag

Sonicadvance1:
In total this gets us in the range of 2-6% faster frontend decode in the microbench. This'll be the final subflag removal for recapturing some of the CPU time that was added to the frontend.


Sonicadvance1:
I'll clean up the final bit for this tomorrow.

Sonicadvance1:
Turns out that I had forgot that FNSTSW can also store to memory, not just AX. Oops. All patched up now.

lioncash (approved):

PR #5958, opened 2026-09-15
Update functional.h

capitan-davide:
Update `move_only_function` to use `std::invoke` instead of regular C++ call syntax which is invalid for pointer-to-member-function call.

In fextl_function.cpp, a `move_only_function` is constructed from a pointer-to-member-function. For this use-case, `std::invoke` should be used over the regular call syntax. When building with `libc++`, the compiler errors out:

```sh
./FEX/FEXCore/include/FEXCore/fextl/functional.h:50:16: error: called object type 'Fnoref' (aka 'int (TrivialExampleFunctionObject::*)(int, int)') is not a function or function pointer
   50 |         return (*moved_lambda)(std::forward<Args>(args)...);
      |                ^~~~~~~~~~~~~~~
./FEX/FEXCore/include/FEXCore/fextl/functional.h:49:58: note: while substituting into a lambda expression here
   49 |       auto wrapped_lambda = [moved_lambda](Args... args) {
      |                                                          ^
./FEX/unittests/APITests/fextl_function.cpp:90:13: note: in instantiation of function template specialization 'fextl::move_only_function<int (TrivialExampleFunctionObject *, int, int), &TestAlloc, &TestDealloc>::move_only_function<int (TrivialExampleFunctionObject::*)(int, int)>' requested here
   90 |     REQUIRE(function<int(TrivialExampleFunctionObject*, int, int)> {&TrivialExampleFunctionObject::Multiply}(&obj, 10, 5) == 50);
      |             ^
./FEX/External/Catch2/src/catch2/../catch2/catch_test_macros.hpp:134:92: note: expanded from macro 'REQUIRE'
  134 |   #define REQUIRE( ... ) INTERNAL_CATCH_TEST( "REQUIRE", Catch::ResultDisposition::Normal, __VA_ARGS__  )
      |                                                                                            ^
./FEX/External/Catch2/src/catch2/../catch2/internal/catch_test_macro_impl.hpp:45:40: note: expanded from macro 'INTERNAL_CATCH_TEST'
   45 |         CATCH_INTERNAL_IGNORE_BUT_WARN(__VA_ARGS__); \
      |                                        ^
./FEX/External/Catch2/src/catch2/../catch2/internal/catch_compiler_capabilities.hpp:386:37: note: expanded from macro 'CATCH_INTERNAL_IGNORE_BUT_WARN'
  386 |         (void)__builtin_constant_p( __VA_ARGS__ ) /* NOLINT(cppcoreguidelines-pro-type-vararg, \
      |                                     ^
In file included from ./FEX/unittests/APITests/fextl_function.cpp:4:
```

Likely, in `libstdc++`, `std::is_nothrow_constructible<std::function<R(Args...)>, F>` evaluates to `true` and this branch was never hit.

In `libc++` this constructor is not `noexcept`: https://github.com/llvm/llvm-project/blob/main/libcxx/include/__functional/function.h#L646-L647


Sonicadvance1 (commented):
Seems reasonable enough to me. Thanks.

Sonicadvance1 (approved):

PR #5959, opened 2026-09-15
FEXCore: Moves crc32 helper to a shared location

Sonicadvance1:


FEXCore/Source/Utils/crc32.h
```diff
@@ -0,0 +1,46 @@
...
+#elif defined(ARCHITECTURE_x86_64)
+#include <nmmintrin.h>
+#endif
+
+namespace FEXCore::Utils {
+template<typename T>
+static inline uint32_t crc32(const T* Ptr, size_t Size) {
+  uint32_t Result {};
+#if defined(ARCHITECTURE_arm64)
+#define do_crc(type, suffix)                                               \
```
lioncash, line 16: These should probably be undef'ed when done with, so they don't leak out of header scope

Sonicadvance1: oop, you right. undeffed.

lioncash (approved):

PR #5960, opened 2026-09-15
unittests/ASM: Manually encode some smsw tests

Sonicadvance1:
Nasm 3.01 fails to encode these as we are expecting for our unittests, resulting in incorrect results. Manually encode to test everything we care about.


lioncash (approved):
nasm really does have the occasional tendency to try and beat cats when it comes to being fickle

Issue #5942, opened 2026-09-10
Segmentation fault when running /sbin/ldconfig


WhatAmISupposedToPutHere:
FEX-2609 segfaults when running /sbin/ldconfig (such as during steam startup). Bisected to 66cad978c39ef68bc39b9e2daf2c2df1cfaff3f2 , reverting just that commit causes fex to work again. Host machine is apple m1pro.

Sonicadvance1:
This doesn't seem to reproduce on my Ubuntu 26.04 host system running FEX's Ubuntu 24.04 rootfs image.
What is your host and guest rootfs environment?

WhatAmISupposedToPutHere:
Gentoo rootfs (and host), can be downloaded here https://github.com/WhatAmISupposedToPutHere/fex-rootfs/releases/tag/20260811

Sonicadvance1:
Reproduced, will see what's going on. Thanks!

PR #5963, opened 2026-09-16
Add regression test for consecutive syscalls

javelina-pkwy:
Issue #5942 was caused by not ending JITed blocks at ``syscall`` instructions, causing register corruption.

This simple regression test exercises this path by running ``getpid()`` twice in a row. Before https://github.com/FEX-Emu/FEX/pull/5944 this test crashes with ``ENOSYS`` when the second ``syscall`` is emulated, since the second ``mov  eax, %getPID%`` instruction is seemingly optimized out.

Admittedly, I don't think this test is perfect. I think that under the right circumstances a combination of multiple bugs at the same time could cause this scenario to succeed by accident, but this is still better than what exists right now.


Sonicadvance1 (approved):
Reasonable enough to me. Even a non-perfect test can ensure it doesn't regress the same way.

lioncash (approved):

PR #5964, opened 2026-09-16
FEXCore: Migrate internal DiskCache header to internal

Sonicadvance1:
Leave the smaller public header that is required to get the current DiskCache version variable still.

NFC


FEXCore/Source/Interface/Core/DiskCache.h
```diff
@@ -0,0 +1,230 @@
...
+#pragma once
+#include "FEXCore/Core/CodeCache.h"
+#include "FEXCore/Core/Context.h"
+#include "Interface/Core/JIT/Relocations.h"
+#include "Interface/Core/Frontend.h"
+#include "Interface/Core/CPUBackend.h"
+#include "FEXCore/Config/Config.h"
+#include "FEXCore/Utils/File.h"
+#include "FEXCore/Utils/WorkQueueThread.h"
+#include "FEXCore/fextl/memory.h"
```
lioncash: These fexcore ones (and the ones at the beginning of the includes) should probably be converted over to `<>` to be consistent

lioncash (approved):

PR #5970, opened 2026-09-18
DiskCache: optional memory buffer for faster lookups

Plagman:
May reserve a tiny area for small blocks in general, but will also be useful for LuaJIT in anon-diskless mode.


Sonicadvance1 (approved):
I love a fully self-contained change.

